PR #15747, opened 2016-03-24
FFT automatic phase correction

### LINKED ISSUE #15279: FFT: deal with phase when input signals do not start at X=0
This issue is the second stage of #15264 

Muon scientists would like `FFT` to handle input signals which do not start at X=0. The idea is to be able to add a phase to the data. Ask James Lord for further information.

James says:
> I've just noticed that an ordinary FFT will happily accept a workspace with uniform sized bins (tidied by Rebin) that doesn't start at x=0 or even have an exact x=0 point anywhere in the spectrum or extrapolated from either end. That's OK as it is, but the output phases seem not to depend on the x starting point. For example:

> ``` python
Rebin(InputWorkspace='PhaseQuadOutput', OutputWorkspace='Binned', Params='0.016')
FFT(InputWorkspace='Binned', OutputWorkspace='FFT1', Imaginary=1) 
ScaleX(InputWorkspace='Binned', OutputWorkspace='Offsetted', Factor='1', Operation='Add') # 1us offset is 62.5 bins
FFT(InputWorkspace='Offsetted', OutputWorkspace='FFT2', Imaginary=1)
```

>and then FFT1 and FFT2 compare equal, when they shouldn't! (there should be a phase difference in spectra 0 and 1, however spectrum 2=magnitude will still be equal)

>So any analysis looking for a peak in the FFT output and then setting up a sinewave to fit or subtract from the time dependent data is going to get its phase wrong. Maxent may suffer similarly. The test will be to Crop the data slightly differently, perhaps keeping the same number of points so the resolution is unchanged, and the phases of peaks should not vary by more than random statistical errors.

Also

> You need to check with a pulsed input signal centred at fixed x regardless of the extent of the X axis, and see how that behaves. For example `YRe=cos(w*x)*exp(-((x-x0)/sig)**2)` for a Gaussian burst.

### PR DESCRIPTION
**Pull request by tom-perkins**
Added an _AutoShift_ property to `FFT` at the request of scientists.

`FFT` assumes that the input workspace's X data is centred at zero.
If this is not true, the output will have the wrong phase.
This can be worked around by using the property _Shift_, but the requirement was to do this automatically.
- Added _AutoShift_ property, which automatically calculates and applies the correct shift
- Added unit tests and extended system test
- Updated algorithm documentation

**To test:**

_Code review:_
In particular, the changes to the documentation and the name of the new parameter.
I don't like the name "AutoShift" but couldn't think of anything better to call it - suggestions welcome!

_Functional tests:_
The new tests should cover most of this.
Using the script from the issue:

```
from math import pi, sin, cos
X = []
YRe = []
YIm = []
E = []
N = 1000
w = 10
for i in range(0,N):
       x = 2*pi*i/N
       X.append(x)
       YRe.append(cos(w*x))
       YIm.append(sin(w*x))
       E.append(0.1)
ws = CreateWorkspace(DataX=X+X,DataY=YRe+YIm,DataE=E+E,NSpec=2)

FFT(InputWorkspace='ws', OutputWorkspace='FFT1', Imaginary=1)

factor = pi/w
ScaleX(InputWorkspace='ws', OutputWorkspace='Offsetted', Factor=factor, Operation='Add')
FFT(InputWorkspace='Offsetted', OutputWorkspace='FFT2', Imaginary=1)
FFT(InputWorkspace='Offsetted', OutputWorkspace='FFT3', Imaginary=1, AutoShift=True)
```

The transforms `FFT1` and `FFT2` should compare equal (wrong) but `FFT1` and `FFT3` should have a phase difference.

Fixes #15279

---
#### Reviewer

##### Code Review
##### Functional Tests

### CODE REVIEW

File: Framework/Algorithms/src/FFT.cpp [hunk 1]
```diff
@@ -145,10 +154,8 @@ void FFT::exec() {
     }
   }
 
-  // centerShift == true means that the zero on the x axis is assumed to be in
-  // the data centre
-  // at point with index i = ySize/2. If shift == false the zero is at i = 0
-  const bool centerShift = true;
+  // Zero on the x axis is assumed to be in the centre,
+  // at point with index i = ySize/2.
```
raquelalvarezbanos: I never understood why we have `centerShift = true`. This causes the Fourier transform of:

signal

to be:

fftsignal

which does not seem right, as my data start at `x=0` and I would expect a positive peak at `w = w0`, being `w0` the frequency of the input signal.

I see why you've removed `centerShift` (it is set to `True` and never changes its value), I am just wondering if we should keep it to make it clear that for some reason we are considering that the x axis is assumed to be in the data center. This way, if at some point we want to set it to `false` (or maybe even add another input property to specify where our center should be) it should be just a matter of setting `const bool centerShift=false;`.
What do you think?

In any case your new property `AutoShift` fixes this behavior, which is good.

tom-perkins: Agreed, I've put `centerShift` back to make the code clearer. The default behaviour (to assume the data is centered at x=0 unless *Shift* or *AutoShift* is used) is a bit confusing but should be kept the same for compatibility. I've tried to make this a bit clearer in the documentation.

raquelalvarezbanos:
Code changes look good, unit tests were added to test the new functionality (as a bonus, existing unit tests were modified so that we don't need to use the ADS). 

`AutoShift` may not be the best name, but I can't think of anything better either, and given that this was the name suggested by muon scientists I would not change it... unless someone has a better suggestion :)

Works as expected. The new functionality is very useful for muon scientists and probably for other areas too.

:shipit:

FedeMPouzols:
I don't see any problem in calling this new property `AutoShift`. The documentation explains what it does, and I don't think there's any "well known" name for this.

PR #15756, opened 2016-03-28
15748 load nexus logs from entry

### LINKED ISSUE #15748: LoadNexusLogs looks only at the first entry
For REF_M, there are multiple entries in the file. The logs are split between these entries, but when using LoadEventNexus all I get is the logs from the first entry, even if in the LoadEventNexus I specified the name for a different entry. The entry name is not passed to LoadNexusLogs, since that algorithm does not take such parameter.

AndreiSavici:
LoadNexusLogs now allows an optional entry for an NXentryName. In LoadEventNexus, if you give an NXentryName, the logs are going to be read correctly, instead of being the logs from the first entry.

**To test:**

All unit and system tests should pass. Run LoadEventNexus on some files with multiple NXentries, and check if the corresponding logs are loaded

Fixes #15748.

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/DataHandling/src/LoadNexusLogs.cpp
```diff
@@ -106,10 +109,12 @@ void LoadNexusLogs::exec() {
   std::string filename = getPropertyValue("Filename");
   MatrixWorkspace_sptr workspace = getProperty("Workspace");
 
+  std::string entry_name = getPropertyValue("NXentryName");
```
quantumsteve: Now that the user can define the value of `NXentryName` the exception on line 123 may be misleading.

quantumsteve:
Changes look good. Unit tests ensure functionality. Logging errors have been updated to reflect that `NXentryName` is a user input and could potentially be typed incorrectly. 

:shipit:

PR #15763, opened 2016-03-29
Fix BinMD failure for file backed workspace of data originally from NXSPE file

### LINKED ISSUE #15397: Bug in SaveMD with incorrect instrument name
From Russell Ewings:
"SaveMD following CreateMD on a set of nxspe files loaded from disk. Seems to work, but then loading this back as a file-backed workspace (which also does not complain) and then taking a cut yields an error saying that the instrument "nxspe" is unknown."

matthew-d-jones:
The problem was that the original NXSPE files do not have instrument information in them and the instrument name is therefore set to "nxspe". This causes an error when BinMD tries to copy experiment info from the input workspace to the output workspace and it doesn't know about an instrument called "nxspe". For this PR I've simply changed BinMD to report a warning to the log rather than throwing the error and failing execution of the algorithm. This makes it somewhat consistent with the non file-backed case where a warning is given when the workspace is loaded if the instrument name is not found.

This is a very minor change which I'll report directly to the person who reported it, rather than putting something in the release notes.

**To test:**

Code review and the unit test may be sufficient, but if you'd prefer to do a manual test:
- Get the data to test with from //olympic/babylon5/Scratch/Matthew Jones/Fe_data.
- In the script below, replace the path with the correct path to the data, then run it.
- It should run BinMD successfully, but report a warning to the log:
  `BinMD was not able to copy experiment info to output workspace`

``` python
import numpy as np

# Create arrays of run numbers and corresponding values of psi
run_numbers = range(15052, 15062)
psi_array = np.arange(0.0, 20.0, 2)

input_runs = ['/path/to/Fe_data/map' + str(run_number) + '_ei400.nxspe' for run_number in run_numbers]

md_ws = CreateMD(input_runs, Emode='Direct', Alatt=[2.87, 2.87, 2.87], Angdeg=[90, 90, 90], u=[1, 0, 0, ], v=[0, 1, 0], Psi=psi_array, EFix=400.0, InPlace=True)

# Save, ignore warnings about missing instrument data
SaveMD(md_ws, Filename="test_save.nxs")
# Load with FileBackEnd option
md_ws_fb = LoadMD(Filename="test_save.nxs", FileBackEnd=True)
md_ws_fb_rebinned = BinMD(md_ws_fb, AlignedDim0="[H,0,0],-12.7244,12.7244,2", AlignedDim1="[0,K,0],-12.7244,12.7244,2", AlignedDim2="[0,0,L],-12.7244,12.7244,2", AlignedDim3="DeltaE,-4,360,2")
```

Fixes #15397.

---
#### Reviewer

##### Code Review
##### Functional Tests


tom-perkins:
The change catches the error from copying experiment infos and logs a warning. A new unit test exposes the bug and verifies the fix. 

On installed Mantid 3.6, the script above fails (failed to find an instrument 'NXSPE'). After the change, it succeeds and I see the new warning in the log.

(Also saw warnings about nan goniometer angle from CreateMD, but this has since been fixed in another PR #15774.)

:shipit:

PR #15774, opened 2016-03-30
Fix goniometer warning where two NaNs detected as not equal

### LINKED ISSUE #15740: CreateMD outputs warning about nan goniometer angle
### Expected behavior
Expected no warning. We are giving a value of psi to use, and default should be zero anyway I think, so warning is unexpected.

### Actual behavior
Get following warning: `Goniometer angle changed in psi log from nan to nan. Used mean = nan.`

Any use of CreateMD?
Known example:
```python 
import os
import numpy as np

#Make MD mega file (like gen_sqw / accumulate_sqw)
psi_arr=np.arange(0.0,92.0,2)  # 46.0
run_arr=range(15052,15098)  # range(15052,15075)

input_runs = list()
psi = list()
gs = list()
gl = list()
#note that gl, gs etc are not mandtory inputs. If we do not specify them then they are set to zero.

for i in range(len(psi_arr)):
    source_file = '/REPLACE/WITH/PATH/TO/Fe_data/map'+str(run_arr[i])+'_ei400.nxspe'
    input_runs.append(source_file)
    psi.append(float(psi_arr[i]))
    gl.append(0.0)
    gs.append(0.0)

print(psi)
print(input_runs)

md_ws = CreateMD(input_runs, Emode='Direct', Alatt=[2.87,2.87,2.87], Angdeg=[ 90, 90, 90], u=[1, 0, 0,], v=[0,1,0], Psi=psi, Gl=gl, Gs=gs, EFix=400.0)
```

matthew-d-jones:
A warning was being given that a goniometer value was changing when using CreateMD, in fact the value was always NaN. Instead of just testing `A != B` it now tests `A != B && !(isnan(A) && isnan(B))`.

**To test:**
Very minor change, code review should suffice and does not need to be in release notes.

Fixes #15740.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/API/src/Run.cpp
+++ b/Framework/API/src/Run.cpp
@@ -433,7 +433,8 @@ void Run::calculateGoniometerMatrix() {
     const double maxAngle =
         getLogAsSingleValue(axisName, Kernel::Math::Maximum);
     const double angle = getLogAsSingleValue(axisName, Kernel::Math::Mean);
-    if (minAngle != maxAngle) {
+    if (minAngle != maxAngle &&
+        !(boost::math::isnan(minAngle) && boost::math::isnan(maxAngle))) {
       const double lastAngle =
           getLogAsSingleValue(axisName, Kernel::Math::LastValue);
       g_log.warning("Goniometer angle changed in " + axisName + " log from " +


peterfpeterson:
Good catch.

:shipit:

PR #15785, opened 2016-03-30
Add make_cow for creation of cow_ptr

### LINKED ISSUE #15780: cow_ptr construction incomplete
the actual resources held by a `cow_ptr` are assumed to have a default constructor, that is invoked when the `cow_ptr` is created.

Things to handle:
* Should be a sink for `DataType*`
* Should be a sink for `ptr_type`, that is a `shared_ptr<DataType>`
* Should have a variadic `make_cow` to extend the allocating constructor case

Tests for the cow_ptr should be provided.

OwenArnold:
Description of work.

Added `make_cow` function to allow the in-place creation of the cow_ptr. Made it possible for the `cow_ptr` to accept an externally allocated resource. 

Manual testing is probably unnecessary. The best way to test this is to look at the `MakeCowTest` test. I added some tests for the `cow_ptr` as these where completely absent. All other testing should pass.

Fixes #15780.

Release notes update is not needed.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Kernel/inc/MantidKernel/cow_ptr.h
+++ b/Framework/Kernel/inc/MantidKernel/cow_ptr.h
@@ -65,7 +65,7 @@ private:
 
 public:
   cow_ptr(ptr_type &&resourceSptr);
-  cow_ptr(DataType *resourcePtr);
+  explicit cow_ptr(DataType *resourcePtr);
   cow_ptr();
   cow_ptr(const cow_ptr<DataType> &);
   cow_ptr(cow_ptr<DataType> &&) = default;
--- a/Framework/Kernel/inc/MantidKernel/cow_ptr.h
+++ b/Framework/Kernel/inc/MantidKernel/cow_ptr.h
@@ -64,6 +64,8 @@ private:
   std::mutex copyMutex;
 
 public:
+  cow_ptr(ptr_type &&resourceSptr);
+  cow_ptr(DataType *resourcePtr);
   cow_ptr();
   cow_ptr(const cow_ptr<DataType> &);
   cow_ptr(cow_ptr<DataType> &&) = default;
@@ -82,6 +84,18 @@ public:
   } ///< Based on ptr equality
   DataType &access();
 };
+
+/**
+ Constructor : creates a new cow_ptr around the resource
+ resource is a sink.
+ */
+template <typename DataType>
+cow_ptr<DataType>::cow_ptr(DataType *resourcePtr)
+    : Data(resourcePtr) {
+  if (resourcePtr == nullptr) {
+    throw std::invalid_argument("cow_ptr source pointer cannot be null");
+  }
+}
 
 /**
   Constructor : creates new data() object
@@ -146,6 +160,14 @@ template <typename DataType> DataType &cow_ptr<DataType>::access() {
   }
 
   return *Data;
+}
+
+template <typename DataType>
+cow_ptr<DataType>::cow_ptr(ptr_type &&resourceSptr) {
+  if (resourceSptr.get() == nullptr) {
+    throw std::invalid_argument("cow_ptr source pointer cannot be null");
+  }
+  this->Data = std::move(resourceSptr);
 }
 
 } // NAMESPACE Kernel
--- a/Framework/Kernel/inc/MantidKernel/cow_ptr.h
+++ b/Framework/Kernel/inc/MantidKernel/cow_ptr.h
@@ -65,6 +65,7 @@ private:
 
 public:
   cow_ptr(ptr_type &&resourceSptr);
+  cow_ptr(const ptr_type &resourceSptr);
   explicit cow_ptr(DataType *resourcePtr);
   cow_ptr();
   cow_ptr(const cow_ptr<DataType> &);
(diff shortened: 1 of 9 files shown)


Framework/Kernel/inc/MantidKernel/cow_ptr.h
```diff
@@ -64,6 +64,8 @@ template <typename DataType> class cow_ptr {
   std::mutex copyMutex;
 
 public:
+  cow_ptr(ptr_type &&resourceSptr);
+  cow_ptr(DataType *resourcePtr);
```
SimonHeybrock: Comparing this with http://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr (I guess it makes sense to follow this closely):

- `cow_ptr(DataType *resourcePtr);` should be `explicit`
- it would be convenient to also have `constexpr shared_ptr( std::nullptr_t );`. Would that break anything? The behavior of the default constructor would still be different, but at least we would have to option to create a `cow_ptr` that does not manage anything.

Framework/Kernel/inc/MantidKernel/cow_ptr.h
```diff
@@ -84,6 +86,18 @@ template <typename DataType> class cow_ptr {
...
+ */
+template <typename DataType>
+cow_ptr<DataType>::cow_ptr(DataType *resourcePtr)
+    : Data(resourcePtr) {
+  if (resourcePtr == nullptr) {
+    throw std::invalid_argument("cow_ptr source pointer cannot be null");
+  }
+}
+
+/**
```
SimonHeybrock: Good, unless we add the option of a `null` `cow_ptr`.

Framework/Kernel/test/CowPtrTest.h
```diff
@@ -0,0 +1,66 @@
...
+    TS_ASSERT_EQUALS(cow->value, value);
+    TSM_ASSERT("Resource should have been moved", resource.get() == nullptr);
+  }
+
+  void testCow() {
+
+    int value = 3;
+    cow_ptr<MyType> original{boost::make_shared<MyType>(value)};
+
+    MyType copy = original.access();
```
SimonHeybrock: I may be mistaken, but I think this test does not do what you want it to do:

1. Create: `cow_ptr<MyType> original{boost::make_shared<MyType>(value)};`
2. Access: `original.access()`. Reference count is 1, so no copy is made, I think.
3. Assignment: `MyType copy = original.access();`. `access()` returns a reference, you assign to value, so you are invoking the copy constructor of `MyType`.

That is, I think the `ASSERT` for the address below works because you copy the type. I think what we really want is a `cow_ptr` with reference count > 1, and then access and capture the result be reference.

SimonHeybrock:
Regarding my comments on `null_ptr` probably we should keep that in a separate PR if we go forward with it, so ignore those comments for now. Apart from my comment on the test of `cow_ptr` everything looks good.

martyngigg:
I don't know if this will raise hell but could we take this opportunity to standardize the class name? Every other class is CamelCase but this has always been different (for some reason). `CowPtr` would be more consistent with our coding standards.

Framework/Kernel/inc/MantidKernel/make_cow.h
```diff
@@ -0,0 +1,22 @@
+#ifndef Mantid_make_cow_h
+#define Mantid_make_cow_h
+
+#include "MantidKernel/cow_ptr.h"
+#include <boost/make_shared.hpp>
+#include <type_traits>
```
martyngigg: What's the `type_traits` header required for?

OwenArnold: Good catch. I was going to use it for `std::is_array`

OwenArnold:
@martyngigg  regarding your previous comment. I would prefer to leave the standardisation to a different issue. This needs thinking about anyway. Some things to consider 

* `CowPtr` is a better fit for our coding standards. `cow_ptr` is not a built-in type.
* `cow_ptr` has now std implementation, but should it come about it would surely be lower-case.

Personally, i'd prefer lower-case for everything, but that's not going to happen.

Would this make a useful maintenance task when agreement is reached?

SimonHeybrock:
@martyngigg I had considered the current name `cow_ptr` nicely consistent with `shared_ptr` and `unique_ptr`.

Framework/Kernel/test/CowPtrTest.h
```diff
@@ -0,0 +1,79 @@
...
+
+    int value = 3;
+    cow_ptr<MyType> original{boost::make_shared<MyType>(value)};
+    auto copy = original; // Now internal shared_ptr count should be at 2
+    MyType &copyResource =
+        original.access(); // The resource should now be copied.
+    copyResource.value = 4;
+
+    TSM_ASSERT_EQUALS("Value should now have changed", original->value,
+                      copyResource.value);
```
SimonHeybrock: I suggest adding an `ASSERT` to verify that `copy` was not changed. You could also test that the `Data` pointers are now different.

SimonHeybrock:
Looks good :shipit:.

PR #15788, opened 2016-03-30
Vesuvio commands to only run system tests on RHEL7 and Windows

### LINKED ISSUE #15784: Temporarily skip Vesuviocommands system test on all but win7 and Rhel7
The newly implemented system test for VesuvioCommands is failing on various operating systems (which the exclusion of windows7 and Rhel7) in order to allow nightly builds to complete, tests on other OSs shall be skipped until a permanent solution is found.

ElliotAOram:
The VesuvioCommands system test should now be skipped if the platform is not RHEL7 or windows. _This is a temporary change to ensure the nightly builds are not disrupted._

**To test:**
- Ensure the system tests pass.
- Would be good to have a few people test this to ensure it skips/doesn't skip on various OS

Fixes #15784.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Testing/SystemTests/tests/analysis/VesuvioCommandsTest.py
+++ b/Testing/SystemTests/tests/analysis/VesuvioCommandsTest.py
@@ -48,6 +48,20 @@ def _equal_within_tolerance(expected, actual, tolerance=0.05):
     abs_difference = abs(expected - actual)
     return abs_difference <= abs(tolerance_value)
 
+def skip_due_to_os():
+    """
+    Skip tests on some operating systems
+    Currenty tests are only being run on windows and RHEL7
+    @return true if tests SHOULD be skipped
+    """
+    dist = platform.linux_distribution()
+    is_rhel7 = (('Red Hat' in dist[0]) or ('CentOS' in dist[0])) and (dist[1].startswith('7'))
+    is_windows = (platform.system() == "Windows")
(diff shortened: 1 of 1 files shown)


Testing/SystemTests/tests/analysis/VesuvioCommandsTest.py
```diff
@@ -59,6 +59,16 @@ def runTest(self):
         runs = "15039-15045"
         self._fit_results = fit_tof(runs, flags)
 
+    def skipTests(self):
+        # Only run tests on Windows and RHEL7
+        dist = platform.linux_distribution()
+        is_rhel7 =  dist[0] == 'Red Hat Enterprise Linux Workstation' and dist[1] == '7.2'
```
martyngigg: We don't want to cut out CentOS 7 (the free rebuild of Red Hat) and I think the version check is a bit to strict. How about something like
```
is_rhel7 = (('Red Hat' in dist[0]) or ('CentOS' in dist[0])) and (dist[1].startswith('7'))
```
It might be an idea to extract this whole thing to a separate function at module level that each `def skipTests(self)` can call too.

martyngigg:
The changes look good now.

:shipit:

FedeMPouzols:
Let's merge this in for tonight.

PR #15790, opened 2016-03-30
Release 3.6.1 and 3.5.2 notes

peterfpeterson:
This is the release notes for both patches. Look it over and see that it isn't crazy.


quantumsteve:
LGTM :shipit:

PR #15801, opened 2016-04-01
Added missing suite methods that slowed down all Kernel tests.

SimonHeybrock:
StringTokenizerTestPerformance does lots of work in constructor and did not define `createSuite` or `destroySuite`. As a result all Kernel tests took of the order of 1 second longer.

**To test:**

Check that most tests from Kernel now run in much less than 1 second (as they used to).

No corresponding issue.
No release notes.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Kernel/test/StringTokenizerTest.h
+++ b/Framework/Kernel/test/StringTokenizerTest.h
@@ -195,6 +195,15 @@ private:
   std::size_t m_length = 50000000;
 
 public:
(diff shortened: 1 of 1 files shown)


KarlPalmen:
1240 tests in KernelTest ran OK in about one minute, rather than 20 minutes as would be for one second per test. :shipit:

PR #15817, opened 2016-04-04
MuonAnalysis: Fix bug with trailing space in workspace name

### LINKED ISSUE #15814: MuonAnalysis: spaces in workspace name mean fitting doesn't work
[This bug is not in the release version and hasn't been seen by users. Noticed during development.]

### Expected behaviour
Should be able to use MuonAnalysis to fit data

### Actual behaviour
Fitting fails with an error as below:
error

- Open MuonAnalysis interface
- Load some data
- Set up a fit function
- Attempt to run the fit

### Platforms affected
All

---

The reason is the recent implementation of string trimming in #15710 coupled with changes in #15615. The workspace name has a space at the end if overwrite is set in the Settings tab, but the space is trimmed by `Fit` and so the workspace can't be retrieved from the ADS. Since the interface lacks tests this was not noticed until now.

Need to change the workspace name so that it doesn't have a space at the end. (Better than setting auto trim to false as that might miss some places).

tom-perkins:
Removed trailing space from workspace name in MuonAnalysis

Always append a number to the workspace name.
If overwrite is on, this is always `#1`.
If overwrite is off, increment each time (same as before).

This fixes the bug caused when overwrite was on and the data was collected in only one period.
In that case the workspace name ended with a separator "; " and the trailing space was trimmed on passing to the `Fit` algorithm.

**To test:**
- Open interface (Interfaces/Muon/Muon Analysis)
- Load any muon data set (for example `MUSR00024580.nxs` from `\\olympic\babylon5\Scratch\TomPerkins\Data`)
- Go to the "Data Analysis" tab and set up a fit function
- Run the fit.

The fit should complete successfully and you should not see an error message that the workspace can't be found in the ADS.

(MuonAnalysis doesn't have automated tests as the interface is not written in a way that allows this easily.)

Fixes #15814 

_Does not need to be in the release notes_ as the bug was not in the 3.6 release and no users will have seen it - unless they are using the very latest nightly build from 1/4/16!

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidQt/CustomInterfaces/src/Muon/MuonAnalysis.cpp
+++ b/MantidQt/CustomInterfaces/src/Muon/MuonAnalysis.cpp
@@ -471,17 +471,21 @@ std::string MuonAnalysis::getNewAnalysisWSName(ItemType itemType, int tableRow,
     workspaceName << "Logs";
     break;
   }
-  
-  // Period(s)
-  workspaceName << sep << getPeriodLabels();
 
-  // Construct workspace name
+  // Period(s)
+  const auto periods = getPeriodLabels();
+  if (!periods.empty()) {
+    workspaceName << sep << periods;
+  }
+
+  // Version - always "#1" if overwrite is on, otherwise increment
+  workspaceName << sep << "#";
   std::string newName;
   if (isOverwriteEnabled()) {
+    workspaceName << "1"; // Always use #1
     newName = workspaceName.str();
   } else {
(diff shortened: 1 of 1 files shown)


ElliotAOram:
Fitted using the `ExpDecayMuon` fit function in Data Analysis interface all ran okay. Changes look good and I agree that no tests or docs need updating. :shipit:

PR #15842, opened 2016-04-05
Fix issues building with VS 2015 Update 2

quantumsteve:
Description of work.

This fixes all compiler warnings and failing unit tests the appeared when building Mantid with Visual Studio Update 2. I also tried modifying the switch statement in `GluGeometryHandler.cpp` so that a similar bug (using `.reserve(size_t)` instead of `.resize(size_t)`) is unlikely to happen again. This bug appeared after v3.6 was tagged and doesn't need to be in the release notes.

**To test:**

Code review is probably sufficient, though building with VS 2015 Update 2 wouldn't hurt. My greatest concern is with increasing the tolerance in `VesuvioTOFFitTest`.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/DataObjects/inc/MantidDataObjects/MDBoxBase.tcc
+++ b/Framework/DataObjects/inc/MantidDataObjects/MDBoxBase.tcc
@@ -112,7 +112,7 @@ TMDE(std::vector<Mantid::Kernel::VMD> MDBoxBase)::getVertexes() const {
     for (size_t d = 0; d < nd; d++) {
       // Use a bit mask to look at each bit of the integer we are iterating
       // through.
-      size_t mask = 1 << d;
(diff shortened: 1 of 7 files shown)


FedeMPouzols:
Code looks good, and as a bonus this also fixes #15857.
So I'm going to get update 2, double check and ship.
The tolerance increase looks acceptable. Please @ElliotAOram , @martyngigg raise any objections if you're not happy with that.

martyngigg:
That's fine with me.

FedeMPouzols:
I upgraded MSVC to update 2 (14.0.25123.00 Update 2). After a couple of reboots...
Mantid compiles and works well in debug/release modes. The tests also run and the issue in the geometry tests ( #15857) is gone.
:shipit:

PR #15852, opened 2016-04-06
H5Utils

peterfpeterson:
This is a central location to collect common usage of the [HDF high level API](https://www.hdfgroup.org/HDF5/doc/cpplus_RM/index.html). It is not intended to be complete.

**To test:**

Should be a code review. Make new issues for missing functionality. However, errors need to be fixed.

_Does not need to be in the release notes._ It is a developer facing change.

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/DataHandling/src/SaveDiffCal.cpp
```diff
@@ -108,77 +109,13 @@ std::map<std::string, std::string> SaveDiffCal::validateInputs() {
...
-} // anonymous
-
 void SaveDiffCal::writeDoubleFieldFromTable(H5::Group &group,
                                             const std::string &name) {
   auto column = m_calibrationWS->getColumn(name);
   std::vector<double> data;
   column->numeric_fill(data);
   data.erase(data.begin() + m_numValues, data.end());
-  writeArray(group, name, std::vector<double>(data));
+  H5Util::writeArray1D(group, name, std::vector<double>(data));
```
quantumsteve: Why are you creating another copy of `data`?

Framework/DataHandling/src/SaveDiffCal.cpp
```diff
@@ -187,7 +124,7 @@ void SaveDiffCal::writeIntFieldFromTable(H5::Group &group,
   std::vector<int32_t> data;
   column->numeric_fill(data);
   data.erase(data.begin() + m_numValues, data.end());
-  writeArray(group, name, std::vector<int32_t>(data));
+  H5Util::writeArray1D(group, name, std::vector<int32_t>(data));
```
quantumsteve: Why are you creating another copy of data?

Framework/DataHandling/src/SaveDiffCal.cpp
```diff
@@ -187,7 +124,7 @@ void SaveDiffCal::writeIntFieldFromTable(H5::Group &group,
   std::vector<int32_t> data;
   column->numeric_fill(data);
   data.erase(data.begin() + m_numValues, data.end());
```
quantumsteve: What do you think about passing `m_numValues` to numeric_fill and not calling erase?

Framework/DataHandling/src/H5Util.cpp
```diff
@@ -0,0 +1,306 @@
...
+    g_log.information() << "DataSet \"" << name << "\" should be double"
+                        << "\n";
+  }
+
+  return result;
+}
+
+namespace {
+// correct for widening - narrowing should use something else
+template <typename InputNumT, typename OutputNumT>
```
quantumsteve: Since `wideningRead` and `narrowingRead` only differ in the lambda expression, would you be better off writing one function and passing the lambda as a parameter?

Framework/DataHandling/src/H5Util.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+  if (getType<NumT>() == dataType) { // no conversion necessary
+    std::vector<NumT> result;
+    DataSpace dataSpace = dataset.getSpace();
+    result.resize(dataSpace.getSelectNpoints());
+    dataset.read(&result[0], dataType, dataSpace);
+    return result;
+  }
+
+  if (PredType::NATIVE_INT32 == dataType) {
```
quantumsteve: Is this if/else tree large enough to justify changing into a `std::map<DataType,std:::function<std::vector<NumT>(DataSet&,DataType&)>>`?

Framework/DataHandling/src/H5Util.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+namespace {
+// correct for widening - narrowing should use something else
+template <typename InputNumT, typename OutputNumT>
+std::vector<OutputNumT> wideningRead(DataSet &dataset,
+                                     const DataType &dataType) {
+  DataSpace dataSpace = dataset.getSpace();
+
+  std::vector<InputNumT> temp(dataSpace.getSelectNpoints());
+  dataset.read(&temp[0], dataType, dataSpace);
```
quantumsteve: use `temp.data()` instead of `&temp[0]`

Framework/DataHandling/src/H5Util.cpp
```diff
@@ -0,0 +1,306 @@
...
+}
+
+// should be correct for widening too
+template <typename InputNumT, typename OutputNumT>
+std::vector<OutputNumT> narrowingRead(DataSet &dataset,
+                                      const DataType &dataType) {
+  DataSpace dataSpace = dataset.getSpace();
+
+  std::vector<InputNumT> temp(dataSpace.getSelectNpoints());
+  dataset.read(&temp[0], dataType, dataSpace);
```
quantumsteve: use `temp.data()` instead of `&temp[0]`

Framework/DataHandling/src/H5Util.cpp
quantumsteve: Can/should we use something like [narrow](https://github.com/Microsoft/GSL/blob/a9f865900d28b854de5ead971aadb82e5ef9ed40/include/gsl_util.h#L104) (also see [this](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es46-avoid-lossy-narrowing-truncating-arithmetic-conversions)) in the GSL instead of setting narrowing or widening "by hand"?

Framework/DataHandling/test/H5UtilTest.h
```diff
@@ -0,0 +1,128 @@
...
+          H5Util::readArray1DCoerce<uint32_t>(group, "array1d_uint32"),
+          array1d_uint32);
+
+      //  with conversion
+      TS_ASSERT_EQUALS(
+          H5Util::readArray1DCoerce<double>(group, "array1d_float"),
+          array1d_double);
+      TS_ASSERT_EQUALS(
+          H5Util::readArray1DCoerce<int32_t>(group, "array1d_uint32"),
+          array1d_int32);
```
quantumsteve: Should we test that this throws an exception when an element > `std::numeric_limits<int32>::max()` is present?

Framework/DataHandling/src/H5Util.cpp
quantumsteve: Maybe we don't need `wideningRead`. [The documentation for numeric_cast](http://www.boost.org/doc/libs/1_60_0/libs/numeric/conversion/doc/html/boost_numericconversion/improved_numeric_cast__.html) says `The implementation must guarantee that for a conversion to a type that can hold all possible values of the source type, there will be no runtime overhead`.

Framework/DataHandling/src/H5Util.cpp
```diff
@@ -0,0 +1,285 @@
...
+template <typename NumT>
+void writeArray1D(Group &group, const std::string &name,
+                  const std::vector<NumT> &values) {
+  DataType dataType(getType<NumT>());
+  DataSpace dataSpace = getDataSpace(values);
+
+  DSetCreatPropList propList = setCompressionAttributes(values.size());
+
+  auto data = group.createDataSet(name, dataType, dataSpace, propList);
+  data.write(&(values[0]), dataType);
```
quantumsteve: use `values.data()`

Framework/DataHandling/src/H5Util.cpp
```diff
@@ -0,0 +1,285 @@
...
+template <typename InputNumT, typename OutputNumT>
+std::vector<OutputNumT> convertingingRead(DataSet &dataset,
+                                          const DataType &dataType) {
+  DataSpace dataSpace = dataset.getSpace();
+
+  std::vector<InputNumT> temp(dataSpace.getSelectNpoints());
+  dataset.read(temp.data(), dataType, dataSpace);
+
+  std::vector<OutputNumT> result;
+  result.resize(temp.size());
```
quantumsteve: I was going to suggest avoiding initializing the vector by calling reserve and using `std::back_inserter`. According to [this link](http://lemire.me/blog/2012/06/20/do-not-waste-time-with-stl-vectors/) what you've done is significantly faster.

quantumsteve:
All potential issues I found in code review have been resolved and the unit tests seem reasonable and pass. 

The extra copy/erase isn't part of this change and should be fixed as a separate maintenance task ( #15884 ).

:shipit:

FedeMPouzols:
All looks good. And yes, resize+read / resize+transform should be as fast as it can get.

PR #15855, opened 2016-04-07
Extended cow_ptr to be more like shared_ptr, in particular added support for NULL cow_ptr.

### LINKED ISSUE #15844: Add support for NULL cow_ptr.
For holding optional resources, and for making the semantics equivalent to `shared_ptr` and `unique_ptr`, `Kernel::cow_ptr` should be able to be equal to `nullptr`.

SimonHeybrock:
Extended cow_ptr to be more like shared_ptr, in particular added support for NULL cow_ptr. A few (related) missing interface functions from `shared_ptr` were also added to make the interfaces more similar.

**To test:**

Code review.

Fixes #15844.

No release notes

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Kernel/inc/MantidKernel/cow_ptr.h
+++ b/Framework/Kernel/inc/MantidKernel/cow_ptr.h
@@ -77,7 +77,7 @@ public:
   cow_ptr<DataType> &operator=(const ptr_type &);
 
   /// Returns the stored pointer.
-  DataType *get() const noexcept { return Data.get(); }
+  const DataType *get() const noexcept { return Data.get(); }
 
   /// Checks if *this stores a non-null pointer, i.e. whether get() != nullptr.
   explicit operator bool() const noexcept { return bool(Data); }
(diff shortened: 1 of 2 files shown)


OwenArnold:
@SimonHeybrock This all looks good. I have some slight reservations regarding the forwarding of methods belonging to the `shared_ptr` regarding uniqueness and instance counts. Looking at the original code, the author had gone through some trouble to hide the `smart_ptr` behind the `cow_ptr`.

SimonHeybrock:
@OwenArnold Yes, you may be right. On the other hand at the time this class was created `shared_ptr` and `unique_ptr` were maybe not so prevalent. Now everyone is familiar with these, so I thought a more uniform interface would actually be easier?

I think you are right and I should definitely remove `get()`, since it is dangerous. I would suggest keeping `unique` and `use_count`, but if you feel otherwise I can remove them as well.

OwenArnold:
@SimonHeybrock, my concern with the `unique` and `use_count` is that the client should never have to access those things. If there was a decision to replace the internals of `cow_ptr` with another type of smart pointer, it maybe impossible to keep providing the `use_count` and `unique` methods.

However, I'm not strongly adversed to having these methods, because I think it's unlikely that a different smart pointer type would be provided.

SimonHeybrock:
@OwenArnold I thought they might be useful in `MatrixWorkspace`, to determine whether `X` is shared, etc.

OwenArnold:
@SimonHeybrock. OK lets leave those methods in then.

OwenArnold:
:shipit:

PR #15860, opened 2016-04-07
Modified class_maker.py for better subfolder support.

SimonHeybrock:
Previous problems for subfolders (option `--subfolder`):
- include guards did not include subfolder name (in header and test)
- test file not placed in subfolder
- test name did not contain subfolder name

All this leads to potential naming clashes.

**To test:**

The changes here are a suggestion, please comment on whether you think they make sense:
- include guards are now, e.g., `MANTID_KERNEL_SUBFOLDER_CLASSNAME_H_`
- tests are placed in subfolders
- test names are `SubfolderClassnameTest`

Known issues:

`move_class.py` currently does not support moving _from_ the new style. Moves should automatically give the new style paths, however. Unless we intend to move all tests to their subdirectories, fixing `move_class.py` for support of both styles seems awkward and error prone, so I decided to leave it alone for now.

To test:
- make class
- make class in subfolder
- move existing class
- make sure the output files in all cases are ok

No corresponding issue.
No release notes.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/buildconfig/class_maker.py
+++ b/buildconfig/class_maker.py
@@ -24,7 +24,7 @@ def write_header(subproject, classname, filename, args):
     f = open(filename, 'w')
 
     subproject_upper = subproject.upper()
-    guard = "MANTID_%s_%s_H_" % (subproject_upper, classname.upper())
+    guard = "MANTID_%s_%s%s_H_" % (subproject_upper, args.subfolder.replace('/', '_').upper(), classname.upper())
 
     # Create an Algorithm header; will not use it if not an algo
     algorithm_header = """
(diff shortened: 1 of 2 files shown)


SimonHeybrock:
This does not quite work: cxxtest seems to expect test names in the format: "FILENAMETest". So the alternative would be to have test all tests in the top directory, with filenames "SubfolderClassnameTest.h" and the tests named "SubfolderClassnameTest".

martyngigg:
@SimonHeybrock We'd need to update [FindCxxTest.cmake](https://github.com/mantidproject/mantid/blob/master/buildconfig/CMake/FindCxxTest.cmake#L102) to be able to handle subfolders in the test names. I personally like the idea of the tests being in subfolders too.

Can I push this back to release 3.8?

SimonHeybrock:
@martyngigg Sure (I just did). I actually stopped working on this for now. We should consider whether or not this makes sense when we eventually do the code re-org? Maybe the script needs a bigger modification at that point anyways?

martyngigg:
Thanks!

True. The script would need a complete overhaul for the code reorg so maybe the improvements are better made at that point.

SimonHeybrock:
@martyngigg Ok, then I will close this completely for now.

PR #15867, opened 2016-04-08
Refactor ISpectrum to use X in Histogram

SimonHeybrock:
This is an intermediate refactoring step. This PR is just for testing on all platforms and running system tests.

Do not merge!


Framework/Algorithms/src/AbsorptionCorrection.cpp
```diff
@@ -188,8 +186,9 @@ void AbsorptionCorrection::exec() {
     MantidVec &Y = correctionFactors->dataY(i);
 
     // Loop through the bins in the current spectrum every m_xStep
+    const auto lambdas = m_inputWS->histogram(i).points();
```
SimonHeybrock: Note that this clunky interface is temporary: I am expecting changes of the `Histogram` interface, so I did not want to duplicate efforts and provide things like `MatrixWorkspace::points(size_t index)` until the `Histogram` interface has settled. So later this could read
```cpp
const auto lambdas = m_inputWS->points(i);
```

Framework/Algorithms/src/AsymmetryCalc.cpp
```diff
@@ -125,14 +125,12 @@ void AsymmetryCalc::exec() {
...
   assert(tmpWS->blocksize() == blocksize);
-  const bool isInputHistogram = inputWS->isHistogramData();
 
   // Create a point data workspace with only one spectra for forward
   API::MatrixWorkspace_sptr outputWS = API::WorkspaceFactory::Instance().create(
       inputWS, 1, blocksize, blocksize);
 
-  // Get the reference to the input x values.
-  auto &tmpX = tmpWS->readX(forward);
+  outputWS->setX(0, tmpWS->histogram(forward).points().cowData());
```
SimonHeybrock: This messy syntax is temporary and will go away once I introduced `Histogram::setPoints(const Points &points);`

SimonHeybrock:
I am now considering how to proceed with the interface to the `X` data. Consider the following: Many algorithms do not seem to care whether they deal with `BinEdges` or `Points`, they just modify the values (look, e.g., at `ScaleX`). As a consequence, they cannot use the part of the `Histogram` interface I described so far,

```cpp
BinEdges binEdges() const;
void setBinEdges(const BinEdges &);
Points points() consts;
void setPoints(const Points &);
```

since the algorithm does not know and does not care what the underlying data layout is. Therefore, we need some additional interface functions for `Histogram`. There are several options:

1. Work with references to the raw data. This would be very similar to what we are doing now, but works only if we decide once and for all that `Histogram` support switching the internal data type. Then, we could have
  ```cpp
  const std::vector<double> &dataX() const;
  std::vector<double> &dataX();
  const std::vector<double> &constDataX() const;
  ```
  The implementations would return a reference to either the data in `BinEdges` or the data in `Points`, depending on the current storage mode.

2. Do all type-agnostic modifications of data via a `Histogram::transformX` method, similar to `std::transform`. For read-only access we can consider adding
  ```cpp
  const std::vector<double> &dataX() const;
  ```
  however this reference could be invalidated if the `Histogram` is modified.

3. Add a new type `HistogramX` that contains `BinEdges` and `Points`. `Histogram` would then be assembled from `HistogramX` and `HistogramY`, etc. `HistogramX` could then support type-agnostic editing, by providing an interface similar to `std::vector`, but more limited.

PR #15883, opened 2016-04-11
Pythonize atom rmm in set sample material

### LINKED ISSUE #15882: Python interface to PhysicalConstants::Atom and SetSampleMaterial to use mass density to get number density
As noted in issue #15843, it would be nice to be able to access the database of cross-sections in PhysicalConstants::Atom from Python. This would be useful for e.g. writing python algorithms for absorption corrections, or for other data processing.

In addition, it was noted that the current absorption correction algorithms use information the Materials part of the workspaces, which is defined using the SetSampleMaterial algorithm. In particular, the absorption calculations needs the number density (in number of atoms/formula units per cubic Angstrom), and SetSampleMaterial either calculates this from the (inputted) number of atoms/formula units per unit cell and the unit cell dimensions (either input, or from the OrientedLattice property), or allows the user to define the number density. However, it would be useful to have an addition way to define the number density, by using the relative molecular mass (can be calculated from the chemical formula), and the measured mass-density.

mducle:
Changed Atom to be a class with all public members rather than a struct, and expose it to Python.
Allow SetSampleMaterial to calculate the number density from the mass density
Added python member functions to Materials - to output the constituent atoms as python objects and to calculate the relative molecular mass

**To test:**

```
import mantid
atom = mantid.kernel.Atom("Pr") 
print "%s-%i: %f amu" % (atom.symbol, atom.a_number, atom.mass)
atom = mantid.kernel.Atom(z_number=6,a_number=13)
print "%s-%i: %f amu" % (atom.symbol, atom.a_number, atom.mass)
atom = mantid.kernel.Atom(symbol="H",a_number=3)
print "%s-%i: %f amu" % (atom.symbol, atom.a_number, atom.mass)

ws = mantid.simpleapi.CreateSampleWorkspace(bankPixelWidth=1,binWidth=10)
mantid.simpleapi.SetSampleMaterial(ws,"Al2 O3",SampleMassDensity=4)
mater = ws.sample().getMaterial()
print mater.numberDensity
atoms = mater.chemicalFormula()
for i in range(len(atoms[0])):
    neut = atoms[0][i].neutron()
    print "%s_%i: sigma_scatt = %f barn" % (atoms[0][i].symbol, atoms[1][i], neut['tot_scatt_xs'])
print mater.relativeMolecularMass()
```

**Expected Output**

```
Pr-0: 140.907650 amu
C-13: 13.003355 amu
H-3: 3.016049 amu
0.023625211555806734
Al_2: sigma_scatt = 1.503000 barn
O_3: sigma_scatt = 4.232000 barn
101.961276
```

Fixes #15882.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/DataHandling/src/SetSampleMaterial.cpp
+++ b/Framework/DataHandling/src/SetSampleMaterial.cpp
@@ -62,9 +62,10 @@ void SetSampleMaterial::init() {
   declareProperty("UnitCellVolume", EMPTY_DBL(), mustBePositive,
                   "Unit cell volume in Angstoms^3. Will be calculated from the "
                   "OrientedLattice if not supplied.");
-  declareProperty("SampleMassDensity", EMPTY_DBL(), mustBePositive,
-                  "Measured mass density in g/cubic cm of the sample to be used "
-                  "to calculate the number density.");
+  declareProperty(
+      "SampleMassDensity", EMPTY_DBL(), mustBePositive,
+      "Measured mass density in g/cubic cm of the sample to be used "
+      "to calculate the number density.");
   declareProperty("CoherentXSection", EMPTY_DBL(), mustBePositive,
                   "Optional:  This coherent cross-section for the sample "
                   "material in barns will be used instead of tabulated");
@@ -244,7 +245,8 @@ void SetSampleMaterial::exec() {
 
     NeutronAtom neutron(0, 0., 0., 0., 0., 0.,
                         0.); // starting thing for neutronic information
-    if (CF.atoms.size() == 1 && isEmpty(zParameter) && isEmpty(rho) && isEmpty(rho_m)) {
+    if (CF.atoms.size() == 1 && isEmpty(zParameter) && isEmpty(rho) &&
+        isEmpty(rho_m)) {
       mat.reset(new Material(chemicalSymbol, CF.atoms[0]->neutron,
                              CF.atoms[0]->number_density));
       // can be directly calculated from the one atom
@@ -288,8 +290,8 @@ void SetSampleMaterial::exec() {
         if ((!isEmpty(zParameter)) && (!isEmpty(unitCellVolume)))
           rho = numAtoms * zParameter / unitCellVolume;
         // or from the relative molecular mass if the mass density is specified
-        else if(!isEmpty(rho_m)) {
-          rho = (rho_m / rmm) * N_A / 1e24;  // measured density in g/cm^3
+        else if (!isEmpty(rho_m)) {
+          rho = (rho_m / rmm) * N_A / 1e24; // measured density in g/cm^3
         }
       }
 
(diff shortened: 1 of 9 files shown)


Framework/DataHandling/src/SetSampleMaterial.cpp
```diff
@@ -308,8 +324,14 @@ void SetSampleMaterial::exec() {
...
       if ((!isEmpty(zParameter)) && (!isEmpty(unitCellVolume)))
         rho = zParameter / unitCellVolume;
+      // or from the relative molecular mass if the mass density is specified
+      else if (!isEmpty(rho_m)) {
+        rho = (rho_m / atom.mass) * N_A / 1e24; // measured density in g/cm^3
+      }
     }
-    mat.reset(new Material(chemicalSymbol, neutron, rho));
+    char symbolname[20];
+    sprintf(symbolname, "(%s%i)", atom.symbol.c_str(), a_number);
```
FedeMPouzols: Here you could use `stringstream` or other alternaives to make this more C++ish.
`snprintf` would be safer than `sprintf`. I wonder if coverity will complain about `sprintf`.
I don't think this is an issue, but maybe we should discourage old-char-style calls for forward-looking consistency.

FedeMPouzols:
@mducle : note I dropped the ticket number from the title of this pull request. It makes the title more readable, especially for the (long) release lists of changes.

Framework/DataHandling/src/SetSampleMaterial.cpp
mducle: Ok. I'll do that. Sorry, just too used to plain C. Stringstream should work.

FedeMPouzols:
@mducle : maybe you want to add a release note in framework / Python?
https://github.com/mantidproject/mantid/blob/master/docs/source/release/v3.7.0/framework.rst

FedeMPouzols:
Oh, maybe you also want to add a release note about the new properties of `SetSampleMaterial`, in framework / algorithms / improved.

FedeMPouzols:
Code looks good and it's documented and unit-tested. The new methods exposed to Python look good and come with their doc strings.
This works well for me and I think it is ready to "ship it", maybe after adding a release note and small changes. @martyngigg , @peterfpeterson I think that atom badly needed some "improvements", but this shouldn't do any harm. What do you think?

peterfpeterson:
Most of the improvements to `Atom` surround compilation and adding the ability to contain wavelength dependance of the absorption. This PR is all improvements.

FedeMPouzols:
:shipit:

martyngigg:
I agree with @peterfpeterson. These changes simply improve the interface for users but we can still change how we store the information internally without impacting on these changes.

Framework/DataHandling/src/SetSampleMaterial.cpp
```diff
@@ -62,6 +62,10 @@ void SetSampleMaterial::init() {
   declareProperty("UnitCellVolume", EMPTY_DBL(), mustBePositive,
                   "Unit cell volume in Angstoms^3. Will be calculated from the "
                   "OrientedLattice if not supplied.");
+  declareProperty(
```
martyngigg: I do have a slight concern about declaring this new parameter here. It will change the auto-generated Python function so that anyone calling it using positional arguments alone would be setting the wrong argument.

At time of writing we have about 100 uses in `v3.6` & 35 uses in `v3.5` that are not internal - http://reports.mantidproject.org/api/feature/1497 and could be affected. To be safe I think we should move the declaration to below the `ScatteringXSection` so that appears as a new positional argument.

mducle: Sure. I've just made the change.

martyngigg: Thanks!

PR #15908, opened 2016-04-14
TOFTOF refactoring1

mganeva:
TOFTOF data loader has been changed to create workspaces with tof_total=tof_source_sample + tof_sample_detector along the X axis. Now ConvertUnits algorithm can be applied to the TOFTOF workspaces. Unfortunately, I cannot get rid of the TOFTOFConvertTofToDeltaE algorithm yet. First I need to find out why ConvertUnits(Target='DeltaE') produces strange results for the TOFTOF data. For other targets (like 'Wavelength') conversion results from ConvertUnits are reasonable. 
I've adjusted dependant TOFTOF workflow algorithms, their unittests and documentation to the performed changes.

This is the first step of the whole refactoring. I am trying to keep changes small to simplify the review. 

**To test:**

Load the TOFTOF test dataset
ws = LoadMLZ(Filename='TOFTOFTestdata.nxs')
Plot and check that the elastic peak is near to 8000 microseconds.
Convert units:
wsL = ConvertUnits(ws, Target='Wavelength')
Check that elastic peak is at the incident wavelength (6 Angstrom)
Convert units to energy transfer:
wsE=TOFTOFConvertTofToDeltaE(ws,ChoiceElasticTof='Geometry')
Check that elastic peak is at the 0 meV.

Fixes #xxxx.

I am not sure, but I think

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/DataHandling/inc/MantidDataHandling/LoadMLZ.h
+++ b/Framework/DataHandling/inc/MantidDataHandling/LoadMLZ.h
@@ -107,7 +107,7 @@ private:
   double m_l1;
   double m_l2;
 
-  double m_t1;      // time of flight from source to sample
+  double m_t1; // time of flight from source to sample
 
   std::vector<std::string> m_supportedInstruments;
   LoadHelper m_mlzloader;
--- a/Framework/DataHandling/inc/MantidDataHandling/LoadMLZ.h
+++ b/Framework/DataHandling/inc/MantidDataHandling/LoadMLZ.h
@@ -107,6 +107,8 @@ private:
   double m_l1;
   double m_l2;
 
+  double m_t1;      // time of flight from source to sample
+
   std::vector<std::string> m_supportedInstruments;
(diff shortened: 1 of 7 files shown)


martyngigg:
@mganeva The changes seem sensible and I've confirmed the results of the conversions so I'm happy to :shipit:

Just a quick note about the conversion to `DeltaE` through `ConvertUnits`. You may be suffering from the same issue that the other SNS & ISIS instruments, namely defining `T0` correctly. `DgsReduction` has an example of what is done to the TOF values in order for ConvertUnits to get the right answer - https://github.com/mantidproject/mantid/blob/master/Framework/WorkflowAlgorithms/src/DgsConvertToEnergyTransfer.cpp#L229. That might be worth a look.

mganeva:
@martyngigg Thank you for the reviewing and for the advice. I will take a look into it.

PR #15926, opened 2016-04-18
Fix compiler warning in ObjCompAssemblyActor for debug build

martyngigg:
Fix a compiler warning in debug mode

**To test:**

Compile in debug mode using `gcc`

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidQt/MantidWidgets/src/InstrumentView/ObjCompAssemblyActor.cpp
+++ b/MantidQt/MantidWidgets/src/InstrumentView/ObjCompAssemblyActor.cpp
@@ -28,7 +28,7 @@ namespace MantidQt
 
           ObjCompAssembly_const_sptr objAss = getObjCompAssembly();
           mNumberOfDetectors = objAss->nelements();
-          assert(m_n == mNumberOfDetectors);
+          assert(static_cast<size_t>(m_n) == mNumberOfDetectors);
           m_data = new unsigned char[m_n * 3];
           m_pick_data = new unsigned char[m_n * 3];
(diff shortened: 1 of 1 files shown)


MantidQt/MantidWidgets/src/InstrumentView/ObjCompAssemblyActor.cpp
```diff
@@ -28,7 +28,7 @@ namespace MantidQt
 
           ObjCompAssembly_const_sptr objAss = getObjCompAssembly();
           mNumberOfDetectors = objAss->nelements();
```
quantumsteve: Could you move the rhs into the assert below? The variable `mNumberOfDetectors ` isn't otherwise used, so in release mode there's no reason to call `objAss->nelements();`.

quantumsteve:
Changing `m_n` to `size_t` and propagating that change to quickly makes this a major change with multiple classes affected and minimal benefit. The `static_cast` is the better option.  Other than minor comment above, LGTM.

MantidQt/MantidWidgets/src/InstrumentView/ObjCompAssemblyActor.cpp
martyngigg: I thought that originally too but it turns out to be a protected member `ICompAssemblyActor.h`! A quick search reveals that it is referenced by all of the other actor classes and I'm unsure of the consequences of changing that. 

I think a proper code review of the instrument view code would be wise at some point.

quantumsteve:
:shipit:

PR #15909, opened 2016-04-14
Add Rebin and Scale post processing steps to RRO

MattKing06:
Previously in the ISIS Reflectometry (Polref) interface there was an option to apply Rebinning and Scaling to the IvsQ workspaces after they had been processed. Since we are working on generalizing the interface so that it can be used by other Reduction algorithms we thought it was necessary to move these post-processing steps to the ReflectometryReductionOne algorithm itself so that the interface can be as generic as possible.

There are now 4 new properties in the ReflectometryReductionOne algorithm:
- **_ScaleFactor**_ - Used to apply scaling to the IvsQ workspace only.
- **_MomentumTransferMinimum**_ - The minimum Q value that you wish to see when the IvsQ workspace has been rebinned.
- **_MomentumTransferStep**_ - Analogous to BinWidth
- **_MomentumTransferMaximum**_ - The maximum Q value that you wish to see when the IvsQ workspace has been rebinned.

The names of these properties are not set in stone but they have been chosen to replicate the properties that already exist in the algorithm for Lambda rebinning, namely WavelengthMin, WavelengthStep and WavelengthStep.

**To test:** 
- This test will involve comparing results from the interface on the Master branch to the results produced by ReflectometryReductionOneAuto algorithm on this branch
## ScaleFactor Test

**_INTERFACE TEST**_
- Ensure you're on the Master branch for this test!
- Open the ISIS Reflectometry (Polref) interface (`Interfaces > Reflectometry > ISIS Reflectometry (Polref)`)
- Change the processing instrument to `INTER`
- Use run number 13460 (ensure that 'search the data archive' is ticked in your directory management settings)
- Change the `Scale` value to 0.5
- Process the workspace.
- Save this workspace as you will need it for comparison.

**_RROA TEST**_
- Load `INTER00013460.nxs` into mantid (this can be found at `Build\ExternalData\Testing\Data\UnitTest\.`
- Open the dialog for ReflectometryReductionOneAuto algorithm
- `INTER00013460` should already be selected as your InputWorkspace
- Give your OutputWorkspace and OutputWorkspacWavelength properties names.
- Set `Scale Factor` property to `0.5`
- Click `Run`
- Load in the workspace from the INTERFACE TEST section
- compare the workspaces using the CompareWorkspaces algorithm
- Ensure that they are equal and and empty `results` table is generated
## Rebinning Test

**_INTERFACE TEST**_
- Ensure you're on the Master branch for this test!
[...]

Issue #15864, opened 2016-04-08
Generalise LoadReflTBL/SaveReflTBL


raquelalvarezbanos:
`LoadReflTBL` and `SaveReflTBL` should be generalized as part of the work on the data processor user interface. At present, they are specific to Reflectometry in terms of the number of columns (which is fixed, and equal to 9) and the type of each column. We should be able to load/save any number of columns (bear in mind that the second- to-last column must be the only integer column). The algorithms should be renamed too, maybe to `LoadTBL` and `SaveTBL`.

MattKing06:
There are currently some obstacles stopping this from being implemented as smoothly as we had planned.

It seems that when the CSV values are read from a table they contain only the values for that row with no indication of what column they are to be assigned to. In our current code for LoadReflTBL we are predefining the column headings and types like this:

```
  auto colRuns = ws->addColumn("str", "Run(s)");
  auto colTheta = ws->addColumn("str", "ThetaIn");
  auto colTrans = ws->addColumn("str", "TransRun(s)");
  auto colQmin = ws->addColumn("str", "Qmin");
  auto colQmax = ws->addColumn("str", "Qmax");
  auto colDqq = ws->addColumn("str", "dq/q");
  auto colScale = ws->addColumn("double", "Scale");
  auto colStitch = ws->addColumn("int", "StitchGroup");
  auto colOptions = ws->addColumn("str", "Options");

  colRuns->setPlotType(0);
  colTheta->setPlotType(0);
  colTrans->setPlotType(0);
  colQmin->setPlotType(0);
  colQmax->setPlotType(0);
  colDqq->setPlotType(0);
  colScale->setPlotType(0);
  colStitch->setPlotType(0);
  colOptions->setPlotType(0);
```

Once we have split the comma-separated line into individual values, we then assign each value to it's predefined column i.e

"14456, 0.5, 1.000494, 2.9494, 0.1" is split into vector which would read like this `vector[0] = 14456`
`vector[1] = 0.5` etc. Once we have this vector, it is then easy to map the value to it's column heading as the value at the first index would go to the first column and so forth.

The difficulty we run into when trying to make this generic is that we should not be allowed to predefine the column headings as each scientific area may require different column headings.

_Proposed solution:_

- Make the first line of any .TBL file contain a comma-separated string of column headings such as
 "Runs,Angle,QMin,QMax"
- Read the number of entries in this line and that will give us the number of columns.
- Use each value from the string to set the heading of that column

[...]

PR #15943, opened 2016-04-20
Generalise ReflTableViewPresenter

### LINKED ISSUE #15672: Generalise ReflTableViewPresenter
Blocked by:

- [x] #14921

Implement steps 4, 5 and 6 in the [design document](https://github.com/mantidproject/documents/blob/master/Design/DataProcessorAlgorithmUI/DataProcessingUserInterface.md#implemenation-steps):

4. Refactor `ReflTableViewPresenter` so that it creates and configures the DataProcessingAlgorithm completely via it's arguments. Then rename it to `GeneralDataProcessorPresenter`. The `GeneralDataProcessorPresenter` should contain nothing that is relfectometry specific. This will also mean abstracting the post processing and pre processing steps.
5. Make `GeneralDataprocessorPresenter` take all arguments as abstract behaviours.
6. Move `GeneralDataProcessorPresenter`, the views and the behaviours into the MantidQtAPI package.

raquelalvarezbanos:
First version of the general Data Processor UI.

**To test:**

Fixes #15672.

---
#### Reviewer

##### Code Review
##### Functional Tests


raquelalvarezbanos:
Changes include:

- Classes have been renamed:
 - `ReflTableViewPresenter` -> `GenericDataProcessorPresenter`
 - `ReflTableView` -> `DataProcessorView`
 - `ReflCommand` -> `DataProcessorCommand`
 - etc

- There are three new classes to define pre-processing, processing and post-processing algorithms. I've named those as `DataPreprocessorAlgorithm`, `DataProcessorAlgorithm` and `DataPostprocessorAlgorithm` respectively, but they can be renamed (note that in Mantid we already use the name `DataProcessorAlgorithm`).

- The relevant methods that should be reviewed are:
 - `GenericDataProcessorPresenter::prepareRunWorkspace()`
 - `GenericDataProcessorPresenter::reduceRow()`
 - `GenericDataProcessorPresenter::postProcessRows()`

- There are still a few things that need to be generalised, for instance, the notebook and a couple of prefixes (`IvsQ`, `IvsLam`)

- We need to merge Matt's PR, in which `ReflectometryReductionOne` will be modify to `Scale` and `Rebin` the workspace `IvsQ`

- `LoadTBL`/`ImportTBL` won't work. We need to fix and merge issue #15864. It should not be too difficult.

raquelalvarezbanos:
Owen: there are still many things that need polishing, (those mentioned above plus new issues I may encounter) but let me know if you like the approach I've taken.

OwenArnold:
@raquelalvarezbanos I don't have a lot of time to look at this, but regarding your last question. Yes no logic in any view. Always refactor to put in presenters or other testable code.

OwenArnold:
@raquelalvarezbanos  from your description above. Sounds like you're on the right track.

raquelalvarezbanos:
One thing I noticed: 

When running the reduction using this table:

example

The following workspaces are created in the ADS:

`IvsLam_13460`
`IvsQ_13460`
`TOF_13460`
`TOF_13460_13462`
`TOF_13462`

This is because we use the `run_number` log of workspace `TOF_13460_13460`, which is just `13460` as the algorithm to add the runs, `Plus`, seems to keep the logs of the first run only, to name the output workspaces, thus giving `IvsLam_13460` and `IvsQ_13460`. This means that, although we have reduced runs `13460+13462`, our output workspaces `IvsLam` and `IvsQ` are named using the first run number only. I think this is confusing, and I would expect the following names:

`IvsLam_13460_13462`
`IvsQ_13460_13462`
`TOF_13460`
`TOF_13460_13462`
`TOF_13462`

This way, it would be clear that the `IvsLam` and `IvsQ` have been produced by running the reduction on `13460+13462` added together. Besides, with the original convention, we would not be able to distinguish between the following two cases:

example2

I made the necessary changes to use the second convention. This implies updating some of the existing unit tests (for instance `GenericDataProcessorPresenterTest::testProcessCustomNames()`). Hope this is OK, let me know if not, @OwenArnold

raquelalvarezbanos:
We now have a working version of the Data Processor UI. REHL7 and Ubuntu builds failed because of changes to `ReflectometryReductionOne` and `ReflectometryReductionOneAuto` (doc tests and system tests need to be updated, see Matt's PR #15909, but first I wanted to make sure with you, @OwenArnold, that changes are OK).

OwenArnold:
@raquelalvarezbanos. I'll look at this in detail when it's ready, but I've been keeping track as this has been going along and I think it's fine.

Framework/Algorithms/src/ReflectometryReductionOne.cpp
```diff
@@ -612,7 +645,59 @@ void ReflectometryReductionOne::exec() {
   }
 
   IvsQ = this->toIvsQ(IvsLam, correctDetectorPositions, theta, isPointDetector);
-
+  double momentumTransferMinimum = getProperty("MomentumTransferMinimum");
```
OwenArnold: This `exec` method has become far too large. I suggest creating methods (preferably non-member) for each sub-operation such as `Scale` like [here](https://github.com/mantidproject/mantid/blob/15672_Generalise_ReflTableViewPresenter/Framework/Algorithms/src/ReflectometryReductionOne.cpp#L691-L699)

Framework/Algorithms/src/ReflectometryReductionOne.cpp
```diff
@@ -612,7 +645,59 @@ void ReflectometryReductionOne::exec() {
...
+      if (!calcResAlg->isExecuted())
+        throw std::runtime_error("CalculateResolution failed. Please manually "
+                                 "enter a value in the dQ/Q column.");
+      momentumTransferStep = calcResAlg->getProperty("Resolution");
+    }
+    QParams.push_back(momentumTransferMinimum);
+    QParams.push_back(-momentumTransferStep);
+    QParams.push_back(momentumTransferMaximum);
+  }
+  IAlgorithm_sptr algRebin = this->createChildAlgorithm("Rebin");
```
OwenArnold: Looks like this final `Rebin` is mandatory. That's not backwards compatible.

Framework/Algorithms/src/ReflectometryReductionOneAuto.cpp
```diff
@@ -462,21 +497,59 @@ void ReflectometryReductionOneAuto::exec() {
...
     }
+    double scaleFactor = getProperty("ScaleFactor");
+    if (scaleFactor != 1.0) {
+      refRedOne->setProperty("ScaleFactor", scaleFactor);
+    }
+    auto momentumTransferMinimum = isSet<double>("MomentumTransferMinimum");
+    auto momentumTransferStep = isSet<double>("MomentumTransferStep");
+    auto momentumTransferMaximum = isSet<double>("MomentumTransferMaximum");
 
+    if (momentumTransferStep.is_initialized()) {
```
OwenArnold: For future reference `boost::optional::is_initialized` is now [deprecated](http://www.boost.org/doc/libs/1_34_0/libs/optional/doc/optional.html)

MantidQt/CustomInterfaces/CMakeLists.txt
```diff
@@ -76,20 +76,20 @@ set ( SRC_FILES
 	src/Muon/MuonAnalysisOptionTab.cpp
 	src/Muon/MuonAnalysisResultTableTab.cpp
 	src/ParseKeyValueString.cpp
+	src/Reflectometry/GenericDataProcessorPresenter.cpp
 	src/Reflectometry/MeasurementItem.cpp
-	src/Reflectometry/QReflTableModel.cpp
-	src/Reflectometry/QReflTableView.cpp
+	src/Reflectometry/QDataProcessorTableModel.cpp
+	src/Reflectometry/QDataProcessorWidget.cpp
```
OwenArnold: Good. Maybe as a later step we'd like to move this new `General` features outside the `Reflectometry` sub directory

MantidQt/CustomInterfaces/CMakeLists.txt
```diff
@@ -215,53 +215,59 @@ set ( INC_FILES
 	inc/MantidQtCustomInterfaces/Muon/MuonAnalysisResultTableTab.h
 	inc/MantidQtCustomInterfaces/ParseKeyValueString.h
 	inc/MantidQtCustomInterfaces/ProgressableView.h
+	inc/MantidQtCustomInterfaces/Reflectometry/DataPostprocessorAlgorithm.h
```
OwenArnold: I think we should have a sub-directory for these `Commands` at some point too.

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/DataPostprocessorAlgorithm.h
```diff
@@ -0,0 +1,120 @@
...
+
+      if (prop->direction() == 0 && prop->type() == "str list") {
+        // For now, we assume we receive the list of workspace to post-process
+        // as
+        // a 'str list'
+        m_inputProp = prop->name();
+
+        countInputWS++;
+      }
+      if (prop->direction() == 1 &&
```
OwenArnold: Looks to me to be an implicit conversion from an old-style enum to an int. Can you compare against `Kernel::Direction` enum (or whatever it is) directly instead.

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/DataPostprocessorAlgorithm.h
```diff
@@ -0,0 +1,120 @@
...
+          (prop->type() == "MatrixWorkspace" || prop->type() == "Workspace")) {
```
OwenArnold: Fine, but this looks like it's going to swallow errors for other types of output workspace. Can you instead implicitly catch that and throw some kind of exception if that happens instead of just passing through.

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/DataProcessorGenerateNotebook.h
```diff
@@ -0,0 +1,150 @@
...
+std::string DLLExport plot1DString(const std::vector<std::string> &ws_names);
+
+std::string DLLExport tableString(QDataProcessorTableModel_sptr model,
+                                  const DataProcessorWhiteList &whitelist,
+                                  const std::set<int> &rows);
+
+std::string DLLExport titleString(const std::string &wsName);
+
+boost::tuple<std::string, std::string> DLLExport postprocessGroupString(
+    const std::set<int> &rows,
```
OwenArnold: Too many arguments to function. Can you group these perhaps putting related arguments into a single `struct`.

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/DataPreprocessorAlgorithm.h
```diff
@@ -0,0 +1,135 @@
...
+
+#include <set>
+#include <string>
+
+namespace MantidQt {
+namespace CustomInterfaces {
+/** @class DataPreprocessorAlgorithm
+
+DataPreprocessorAlgorithm defines a pre-processor algorithm that will be
+responsible for pre-processsing a specific column in a Data Processor UI.
```
OwenArnold: Can you please elaborate on this. Specify the exact function it servers.

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/DataPreprocessorAlgorithm.h
```diff
@@ -0,0 +1,135 @@
...
+class DataPreprocessorAlgorithm {
+public:
+  /** Constructor
+  * @param name : The name of the pre-processing algorithm
+  * @param prefix : The prefix that will added to the output workspace name
+  * @param blacklist : The list of properties we don't want to show
+  * @param show : Whether or not to show the information associated with
+  * this pre-processor in the processed workspace's name
+  */
+  DataPreprocessorAlgorithm(
```
OwenArnold: Please create a corresponding `cpp` file to hold the definition for this and keep the header declaration-only.

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/DataPreprocessorAlgorithm.h
```diff
@@ -0,0 +1,135 @@
...
+      const std::string &name, const std::string &prefix = "",
+      const std::set<std::string> &blacklist = std::set<std::string>(),
+      bool show = true)
+      : m_name(name), m_prefix(prefix), m_blacklist(blacklist), m_show(show) {
+
+    Mantid::API::IAlgorithm_sptr alg =
```
OwenArnold: This is a code-smell to me. You have to have registered all algorithms in order to use this, or even test it. I would have the constructor take a `const IAlgorithm&` rather than a name.

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/DataPreprocessorAlgorithm.h
```diff
@@ -0,0 +1,135 @@
...
+
+    auto properties = alg->getProperties();
+    for (auto &prop : properties) {
+
+      if (prop->direction() == 0 &&
+          (prop->type() == "MatrixWorkspace" || prop->type() == "Workspace")) {
+        // For now, we restrict the workspaces to either MatrixWorkspace
+        // or Workspace, this condition can be relaxed if necessary
+
+        if (!countInputWS) {
```
OwenArnold: I don't like this. Implicit conversion from `int` to `bool`. Also this somehow depends on property declaration order inside the target algorithm doesn't it?

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/DataPreprocessorAlgorithm.h
```diff
@@ -0,0 +1,135 @@
...
+        Mantid::API::AlgorithmManager::Instance().create(m_name);
+
+    int countInputWS = 0;
+    int countOutputWS = 0;
+
+    auto properties = alg->getProperties();
+    for (auto &prop : properties) {
+
+      if (prop->direction() == 0 &&
+          (prop->type() == "MatrixWorkspace" || prop->type() == "Workspace")) {
```
OwenArnold: Again, issue about swallowing errors.

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/DataPreprocessorAlgorithm.h
```diff
@@ -0,0 +1,135 @@
...
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+File change history is stored at: <https://github.com/mantidproject/mantid>.
+Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+class DataPreprocessorAlgorithm {
+public:
+  /** Constructor
+  * @param name : The name of the pre-processing algorithm
+  * @param prefix : The prefix that will added to the output workspace name
```
OwenArnold: Specify that this is to the name in the `Analysis Data Service`

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/DataProcessorAlgorithm.h
```diff
@@ -0,0 +1,119 @@
...
+      const std::set<std::string> &blacklist = std::set<std::string>())
+      : m_name(name), m_prefix(prefix), m_blacklist(blacklist) {
+
+    Mantid::API::IAlgorithm_sptr alg =
+        Mantid::API::AlgorithmManager::Instance().create(m_name);
+
+    auto properties = alg->getProperties();
+    for (auto &prop : properties) {
+
+      if (prop->direction() == 0 &&
```
OwenArnold: Given my previous comments, I can't help but think that there's some code-sharing missing between these various `Data{}ProcessorAlgorithm` types. Such as here.

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/DataProcessorWhiteList.h
```diff
@@ -0,0 +1,74 @@
...
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+File change history is stored at: <https://github.com/mantidproject/mantid>.
+Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+class DataProcessorWhiteList final {
```
OwenArnold: I don't see why `final` is necessary. Maybe you have a reason? I never use `final`

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/GenericDataProcessorPresenterFactory.h
```diff
@@ -0,0 +1,48 @@
...
+public:
+  /**
+  Constructor
+  */
+  GenericDataProcessorPresenterFactory(){};
+  virtual ~GenericDataProcessorPresenterFactory() = default;
+  /**
+  Creates a GenericDataProcessorPresenter
+  */
+  virtual boost::shared_ptr<GenericDataProcessorPresenter> create() = 0;
```
OwenArnold: Good. We probably don't need this as a `shared_ptr` a `unique_ptr` would probably work as I don't see why `shared` ownership is required. Maybe I'm wrong about this? The `View` should be sole owner of the `Presenter`.

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/ReflGenericDataProcessorPresenterFactory.h
```diff
@@ -0,0 +1,147 @@
...
+#include "MantidQtCustomInterfaces/Reflectometry/DataPostprocessorAlgorithm.h"
+#include "MantidQtCustomInterfaces/Reflectometry/DataPreprocessorAlgorithm.h"
+#include "MantidQtCustomInterfaces/Reflectometry/DataProcessorAlgorithm.h"
+#include "MantidQtCustomInterfaces/Reflectometry/DataProcessorWhiteList.h"
+
+namespace MantidQt {
+namespace CustomInterfaces {
+/** @class ReflGenericDataProcessorPresenterFactory
+
+ReflGenericDataProcessorPresenterFactory creates a Reflectometry
```
OwenArnold: Nice! just like we discussed.

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/ReflGenericDataProcessorPresenterFactory.h
```diff
@@ -0,0 +1,147 @@
...
+class ReflGenericDataProcessorPresenterFactory
+    : public GenericDataProcessorPresenterFactory {
+public:
+  ReflGenericDataProcessorPresenterFactory() = default;
+  virtual ~ReflGenericDataProcessorPresenterFactory() = default;
+
+  /**
+  * Creates a Reflectometry Data Processor Presenter
+  */
+  boost::shared_ptr<GenericDataProcessorPresenter> create() override {
```
OwenArnold: Please change this and everything else that is not a templated type to separate implementaiton to a cpp file.

MantidQt/CustomInterfaces/src/Reflectometry/DataProcessorGenerateNotebook.cpp
```diff
@@ -0,0 +1,669 @@
...
+@param preprocessingOptionsMap : options to pre-processing algorithms
+specified via hinting line edits in the view
+@param processingOptions : options to the reduction algorithm specified via
+the corresponding hinting line edit in the view
+@param postprocessingOptions : options to the post-processing algorithm
+specified via the corresponding hinting line edit in the view
+@returns ipython notebook string
+*/
+DataProcessorGenerateNotebook::DataProcessorGenerateNotebook(
+    std::string name, QDataProcessorTableModel_sptr model,
```
OwenArnold: Can we group some of these related arguments together into a struct?

OwenArnold:
Two problems illustrated in the image below.

1. `Option` on each row needs to be explained better now. This is some kind of one-off options, as in only applies to that row right? Where as the options below are global. Otherwise my `WavelenthMax` entry in the image is going to really really confuse people.

2. The pop-up auto-complete doesn't die when that text box loses focus. You can see two of them still open. It works fine for the options against each row, but not for the new text boxes.

screen shot 2016-05-07 at 20 55 47

OwenArnold:
Try importing a TBL file. I get this...

screen shot 2016-05-07 at 21 01 38

OwenArnold:
@raquelalvarezbanos This is really good. Some very solvable issue to sort out, Maybe we should put together a white-paper on this?

Framework/Algorithms/src/ReflectometryReductionOne.cpp
raquelalvarezbanos: Yes, it is mandatory now. There are two main reasons:
1) It was mandatory in the interface, i.e. our old `ReflTableViewPresenter` would always apply this `Rebin` step.
2) Jos and Max mentioned that they want the interface to do exactly what the algorithm does, they don't want the interface to do extra steps. They also mentioned that they usually want their `IvsQ` workspace to be binned logarithmically.

OwenArnold: Very well. I defer to their judgement.

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/DataProcessorAlgorithm.h
raquelalvarezbanos: Yes, they do share code at the moment. In fact that was my impression too and when I first implemented this I did it using a shared base class. But at some point I decided to follow this approach instead to make the three classes more independent, so they would be easier to modify in the future (for instance other people may need their `DataPreprocessorAlgorithm` to allow other type of input properties, `InOut` properties instead of only `Input` or `Output`, etc), right now I don't know how different they may become...

OwenArnold: I don't think that the duplication should be left in, even if there is a longer term plan that something different might happen. You don't need a whole base class. A header defining a shared free-function would be enough. Composition is fine.

raquelalvarezbanos:
The last commits should fix the functionality that was broken (i.e., importing/exporting TBL files). I also fixed the issue with the hints not dying when the text box loses focus). I think at this point we have an interface with the same functionality as we had before.

OwenArnold:
@raquelalvarezbanos. OK i'll check the updated application. 

Are you OK to conduct the other aforementioned refactoring?

raquelalvarezbanos:
Yes, but first I will update the interface's documentation. Then I would say that moving the Reflectometry-independent class outside of the Reflectometry directory should be next thing to do. After that I will start with the refactoring, do you agree with this priority list?

OwenArnold:
I would consider the refactoring more important that the physical location of the files, but your ordering sounds reasonably sensible to me.

OwenArnold:
Just tried it out. The pop-up issue is fixed. But I still get problems importing an old style tbl file. I've sent you such a file.

OwenArnold:
I get a bit of a lag when I first start the Hinting Line Edits. I suspect this is because the hinting line edit information is calculated dynamically. I'm sure this could be done faster if it was calculated and cached. That's another issue to work on I think.

raquelalvarezbanos:
Loading of old TBL formats seems to work now, I was able to load `INTER_NR_test2.tbl` (which I found in `SampleData-ISIS`)

OwenArnold:
@raquelalvarezbanos. Great. I'll re-test the package.

raquelalvarezbanos:
I've added column titles for processing, pre-processing and post-processing options. I've tried using a QTableWidget but it did not work (I would have to set a delegate in order to include `HintingLineEdit`s in the table, but it seems that delegates can only be set to the entire column, not to a specific cell only).

How about something like this:

reflectometrywithheadings

OwenArnold:
Yes. Looks fine to me. Can you please use Matt for some of the testing today. I'm going to be quite tied up.

raquelalvarezbanos:
Added headers to global options. Matt and I found a problem with `ReflectometryReductionOneAuto`, we are getting wrong `Qmin` and `Qmax` values (after `ConvertUnits`, although it seems with that Matt's changes, as we `Rebin` we are fixing the problem) in the output workspace, we are still trying to figure out why.

OwenArnold:
I just did a quick test of the aforementioned table file. While it loads. Did you actually try reducing it? I get this:

```
Stitch1DMany started
Algorithm: Stitch1DMany v1
Execution Date: 1970-Jan-01 00:00:00
Execution Duration: -1 seconds
Parameters:
  Name: InputWorkspaces, Value: IvsQ_TOF_13460,IvsQ_TOF_13462, Default?: No, Direction: Input
  Name: OutputWorkspace, Value: IvsQ_TOF_13460_TOF_13462, Default?: No, Direction: Output
  Name: Params, Value: , Default?: Yes, Direction: Input
  Name: StartOverlaps, Value: , Default?: Yes, Direction: Input
  Name: EndOverlaps, Value: , Default?: Yes, Direction: Input
  Name: ScaleRHSWorkspace, Value: 1, Default?: Yes, Direction: Input
  Name: UseManualScaleFactor, Value: 0, Default?: Yes, Direction: Input
  Name: ManualScaleFactor, Value: 1, Default?: Yes, Direction: Input
  Name: OutScaleFactors, Value: , Default?: Yes, Direction: Output
Invalid value for Params: At least one parameter must be given.
```

raquelalvarezbanos:
Yes, that is because no options have been specified for `Stitch1DMany`

raquelalvarezbanos:
Just wanted to copy here some test results (copied from PR https://github.com/mantidproject/mantid/pull/15909), where I was testing reduction results running the interface on the master branch, and reduction results using this branch:

Reduction table:

intertest

Results (red is this branch, black is master):
intertestresults

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/DataPreprocessorAlgorithm.h
raquelalvarezbanos: I can change it to have the constructor take a `const IAlgorithm&`, however, I will still have to have registered all algorithms if I want to have a default constructor: `DataProcessorAlgorithm()` will need to register the algorithms, create the `Plus` algorithm (which is the default pre-processing algorithm) and pass it to the non-default constructor. Maybe there is a way to avoid this, but I can't see how, unless we get rid of the default constructor..

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/GenericDataProcessorPresenterFactory.h
raquelalvarezbanos: Both our table view (`QDataProcessorWidget`) and the `outer` presenter (`ReflMainViewPresenter`) need to interact with our `GenericDataProcessorPresenter`, the former because it is the presenter's view, and the latter because it needs to receive the commands. Can we do this with a `unique_ptr`?

OwenArnold:
Plots above look good. Did you also check with error bars?

raquelalvarezbanos:
With errors:

intertestresultswitherrors

OwenArnold:
@MattKing06 I'm going to check the code following Raquel's next round of changes. Could you please perform some detailed functional testing. Make sure that the interface documentation for PolRef all still works. Thanks!

raquelalvarezbanos:
I've created a new ticket https://github.com/mantidproject/mantid/issues/16203 to fix the remaining issues:
 - `shared_ptr` should be replaced with `unique_ptr`
 - `Data{}ProcessorAlgorithms` should not need to have all the algorithms registered (+ use of NullObject to define "null" behavior)
 - `DataProcessor{}` should be moved outside the `Reflectometry` directory.

raquelalvarezbanos:
@OwenArnold Matt performed detailed functional testing (which lead to a couple of bug fixes and improvements to the documentation). Is this ready to be moved to the next stage so that a second reviewer can test it? (once RHEL6 passes).

OwenArnold:
@raquelalvarezbanos. Looks good to me. Thanks for the refactoring steps.

If you don't already have an issue to configure a new "demo" data reduction based on something like [this](https://github.com/mantidproject/mantid/blob/a6ffb617a5c33fd63e94b4c4f14d1c5196c522db/MantidQt/CustomInterfaces/src/Reflectometry/ReflGenericDataProcessorPresenterFactory.cpp) could you please create one.

raquelalvarezbanos:
Thanks Owen, I've created a new issue (https://github.com/mantidproject/mantid/issues/16225)

MattKing06:
After performing functional tests in #15909 for the new interface, all of the results are as expected.
:shipit:

FedeMPouzols:
This is an important PR. Honestly I cannot review it much in detail but we can perfectly trust the quite thorough review and functional testing that has been done here. Let's merge it in so that it is ready for unscripted testing and beta testing.

PR #15953, opened 2016-04-20
Computes phonon density of states in the incoherent approximation.

### LINKED ISSUE #15843: Python algorithm to calculate the phonon density of states from powder inelastic data
It would be useful to have an algorithm to compute the (neutron weighted) density of states from S(q,w) in the incoherent approximation (from e.g. Skold and Price, Neutron Scattering, eqn. 1.70 http://bit.ly/1PTuc6a )

Z(E) \propto S(Q,E) / (Q^2/E) / exp(-2W) / <n+1>

for neutron energy loss; <n+1> should be <n> for neutron energy gain, with <n+1> = 1/(exp[E/kT]-1)+1, and <n> = 1/(exp[E/kT]).

This would be useful in lattice dynamics measurements in powder / polycrystalline samples, e.g. on Mari.

mducle:
This is a python algorithm to compute the phonon density of states from a powder S(q,w) dataset. In this case, all the scattering can be treated as incoherent (the coherent scattering averaged over all angles in a powder is similar to the isotropic incoherent scattering) and a simple formula can be applied to derive the neutron weighted partial density of states. The theory is partly detailed in the documentation rst.

It only operates on histogram workspaces with one axis being momentum transfer (Q) and one being energy transfer (in either meV or cm). By default it outputs the DOS g(E) in meV but can be specified to output in wavenumbers (cm^-1).

**To test:**

By constructing a (Q,w) workspace

```
ws = CreateSampleWorkspace(binWidth = 0.1, XMin = 0, XMax = 50, XUnit = 'DeltaE')
ws = ScaleX(ws, -25, "Add")
ws = SofQW(ws, [0, 0.05, 8], 'Direct', 25)
ws_DOS = ComputeIncoherentDOS(ws)
```

Alternatively, with a datafile (note the datafiles should be accessible in the archive, but the mapfile might need to be copied somewhere accessible. It is here: https://svn.isis.rl.ac.uk/InstrumentFiles/trunk/mari/mari_res2013.map 
The following uses a measurement of Aluminium on MARI at ISIS.

```
from Direct import DirectEnergyConversion
rd = DirectEnergyConversion.DirectEnergyConversion('MARI')
ws = rd.convert_to_energy(21334, 21335, 60, [-55,0.05,55], 'mari_res2013.map', 
    monovan_run=21347, sample_mass=106.4, sample_rmm=27, monovan_mapfile='mari_res2013.map')
ws_sqw = SofQW3(ws, [0,0.1,12], 'Direct', 60)
SetSampleMaterial(ws_sqw,'Al')
ws_dos = ComputeIncoherentDOS(ws_sqw, Temperature=5, StatesPerEnergy=True)

```

Compare to the data here: http://journals.aps.org/prb/abstract/10.1103/PhysRevB.77.024301
Too high by a factor of around 3, which I haven't figured out why... 
Possibly from the Debye-Waller factor. Or possibly due to multiple scattering.

Fixes #15843.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/PythonInterface/plugins/algorithms/ComputeIncoherentDOS.py
+++ b/Framework/PythonInterface/plugins/algorithms/ComputeIncoherentDOS.py
@@ -8,7 +8,7 @@ from scipy import constants
 def evaluateEbin(Emin, Emax, Ei, strn):
     return [eval(estr) for estr in strn.split(',')]
 
-def evaluateQRange(Qmin, Qmax, string):
+def evaluateQRange(Qmin, Qmax, strn):
     return [eval(qstr) for qstr in strn.split(',')]
 
 
--- a/Framework/PythonInterface/plugins/algorithms/ComputeIncoherentDOS.py
+++ b/Framework/PythonInterface/plugins/algorithms/ComputeIncoherentDOS.py
@@ -1,8 +1,9 @@
 #pylint: disable = no-init, invalid-name, line-too-long, eval-used, unused-argument, too-many-locals, too-many-branches, too-many-statements
-from mantid.kernel import *
-from mantid.api import *
-from mantid.simpleapi import *
+from mantid.kernel import CompositeValidator, Direction, FloatBoundedValidator
+from mantid.api import mtd, AlgorithmFactory, CommonBinsValidator, HistogramValidator, MatrixWorkspaceProperty, PythonAlgorithm
+from mantid.simpleapi import CreateWorkspace, DeleteWorkspace, Rebin2D
 import numpy as np
+from scipy import constants
 
 def evaluateEbin(Emin, Emax, Ei, string):
     return [eval(estr) for estr in string.split(',')]
@@ -16,14 +17,17 @@ class ComputeIncoherentDOS(PythonAlgorithm):
         measured powder inelastic neutron scattering spectrum.
 
     ComputeIncoherentDOS(InputWorkspace = iws, Temperature = T, MeanSquareDisplacement = msd,
-                         QSumRange = qrange, EnergyBinning = ebins, OutputWorkspace = ows)
+                         QSumRange = qrange, EnergyBinning = ebins, Wavenumbers = cm, StatesPerEnergy = states,
+                         OutputWorkspace = ows)
 
     Input:
             iws     -   An input MatrixWorkspace (must contain the reduced powder INS data)
             T       -   The sample temperature in Kelvin (default: 300 K)
-            msd     -   The average mean-squared displacement of all species in the sample in Angstrom^2 (default: 0.01A^2)
+            msd     -   The average mean-squared displacement of all species in the sample in Angstrom^2 (default: 0)
             qrange  -   Range in |Q| to integrate over (default: 0.5Qmax to Qmax)
             ebins   -   Energy bins in Energy transfer to rebin the data into (default: 0 to Emax in 50 steps)
+            cm      -   (True/False) whether the output should be in meV or cm^{-1}
+            states  -   (True/False) whether the output should be states/unit energy if the sample is a pure element.
             ows     -   An output spectrum workspace (1D)
     """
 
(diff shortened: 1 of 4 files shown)


NickDraper:
Code review is good, clear and well formatted.
Documentation looks good - There is one formula that is not parsing properly in github, but might be fine through sphinx - Still to check
Still Building at the moment, so have yet to test results, but I'm slightly concerned about this comment
"Too high by a factor of around 3, which I haven't figured out why... 
Possibly from the Debye-Waller factor. Or possibly due to multiple scattering."

Did the old PySlice routine show the same 3 fold factor increase? Otherwise we need to state that in the release notes.

mducle:
The old PySlice routine didn't calculate the DOS on an absolute scale (it left it in mb/sr/fu/meV). 

You're right that we should probably state this in the release notes - that the "StatesPerEnergy" option should be used with a pinch of salt. Or maybe it can be removed until it's verified.

docs/source/algorithms/ComputeIncoherentDOS-v1.rst
```diff
@@ -0,0 +1,88 @@
...
+Computes the phonon density of states from an inelastic neutron 
+scattering measurement of a powder or polycrystalline sample,
+assuming that all scattering is incoherent, using the formula 
+(Thermodynamic Properties of Solids, eds. Chaplot, Mittal, Choudhury,
+Chapter 3) for the 1-phonon incoherent scattering function:
+
+:math:`S^{(1)}_{\mathrm{inc}}(Q,E) = 
+   \exp\left(-2\bar{W}(Q)\right) \frac{Q^2}{E} 
+   \langle n+\frac{1}{2}\pm\frac{1}{2} \rangle
+   \left[ \sum_k \frac{\sigma_k^{\mathrm{scatt}}{2m_k} g_k(E) \right]`
```
NickDraper: There is a problem with this latex formula.

Firstly it is not parsing as you have it split over multiple lines, which the ln line :math: command does not like.
However there is also a problem with the last frac term the number of brackets do not work.

This corrects it so it will parse, but I've no idea if that is the correct formula once if made my guessed corrections.
:math:`S^{(1)}_{\mathrm{inc}}(Q,E) =    \exp\left(-2\bar{W}(Q)\right) \frac{Q^2}{E}    \langle n+\frac{1}{2}\pm\frac{1}{2} \rangle   \left[ \sum_k \frac{\sigma_k^{\mathrm{scatt}}{2m_k}} {g_k(E)} \right]`

image

NickDraper: bc21072a10520a1b4942f0319f703029dd487736

mducle: Sorry... 

:math:`S^{(1)}_{\mathrm{inc}}(Q,E) = \exp\left(-2\bar{W}(Q)\right) \frac{Q^2}{E} \langle n+\frac{1}{2}\pm\frac{1}{2} \rangle \left[ \sum_k \frac{\sigma_k^{\mathrm{scatt}}}{2m_k} g_k(E) \right]`

I should have checked this but I don't have DocUtils installed...

Framework/PythonInterface/plugins/algorithms/ComputeIncoherentDOS.py
```diff
@@ -0,0 +1,196 @@
...
+        # Convert to wavenumbers if requested (y-axis is in mbarns/sr/fu/meV -> mb/sr/fu/cm^-1)
+        if cm:
+            en = en*mev2cm
+            y = y/mev2cm
+            e = e/mev2cm
+            yunit = 'DeltaE_inWavenumber'
+        else:
+            yunit = 'DeltaE'
+
+        # Outputs the calculated density of states to another workspace
```
NickDraper: There is a problem here that causes a crash in debug mode on windows.  Your created workspace using CreateWorkspace is not setup correctly for Rebin2D to work on it.

The data on the old X edges is coming through as point data, not bin edges so the array overflows for the final boundary.

You will probably need to use the workspacefactory to create the workspace rather than CreateWorkspace to get one that is in the right form for Rebin2D.

mducle: Ah right. Ok. I'll make another commit. 

In future is it possible to get the bin center values of the axes values, as maybe a member function of the histogram workspace?

NickDraper:
All looking good now :shipit:

FedeMPouzols:
@NickDraper: please remember to tick the relevant boxes of the reviewer template.

FedeMPouzols:
I think we have a problem with RHEL6 and this code. Last nights docs-test build failed because apparently `constants.Boltzmann` is missing in the (probably ancient) scipy of RHEL6:
http://builds.mantidproject.org/job/master_doctests/279/label=rhel6-build/testReport/junit/docs/algorithms_ComputeIncoherentDOS-v1/ExGenerated/

FedeMPouzols:
On RHEL6 we seem to have scipy 0.7.2! The list of physical constants is much shorter: http://docs.scipy.org/doc/scipy-0.7.x/scipy-ref.pdf. But it seems that `k` is an alias for `Boltzmann` and is available in 0.7.2. @mducle: would you be happy with that change (`constants.Boltzmann` => `constants.k`?

mducle:
Yes. Please change it. Sorry for not checking!

FedeMPouzols:
No worries. It's impossible in practice to foresee all these incompatibilities with ancient RHEL6 Python packages.  We have issues like this from time to time. This one is easy to fix. Also the assert methods supported by unittest with Python 2.6 (which is what we have on RHEL6) is much more limited. See for example: #16055. There's been issues with numpy as well.

PR #15955, opened 2016-04-21
Refactor PeakShape

quantumsteve:
Description of work.

This refactors `Mantid::Geometry::PeakShape::radius()` to better support arbitrary shapes without casting back to it's derived type. I also made internal changes to `vtkPeaksFilter` to remove many unnecessary copies.

**To test:**

Open a `MDEventWorkspace` and a `PeaksWorkspace` in the VSI. View the peaks table to activate the PeaksFilter and verify that the points and glyphs still render properly.

This pull request has no related issue.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Geometry/inc/MantidGeometry/Crystal/PeakShape.h
+++ b/Framework/Geometry/inc/MantidGeometry/Crystal/PeakShape.h
@@ -47,6 +47,7 @@ public:
   virtual int algorithmVersion() const = 0;
   /// Shape name
   virtual std::string shapeName() const = 0;
+  /// For selecting different radius types.
   enum RadiusType { Radius = 0, OuterRadius = 1, InnerRadius = 2 };
   /// Radius
   virtual boost::optional<double> radius(RadiusType type) const = 0;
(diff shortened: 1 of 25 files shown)


Framework/Geometry/inc/MantidGeometry/Crystal/PeakShape.h
```diff
@@ -46,6 +47,10 @@ class DLLExport PeakShape {
   virtual int algorithmVersion() const = 0;
   /// Shape name
   virtual std::string shapeName() const = 0;
+  /// For selecting different radius types.
+  enum RadiusType { Radius = 0, OuterRadius = 1, InnerRadius = 2 };
```
AntonPiccardoSelg: We might want to consider using a strongly typed enum. At the interface boundaries, i.e. where `int`s are passed , we could have a simple translation function from `int` to the declared enum.

We could do this as a second step or for maintenance.

AntonPiccardoSelg:
Code Review:
* Code looks fine
* `enum class` might be something to consider in the next iteration
* good to remove passing by value

Functional Testing:
* Verified that the peaks table works for ellipsoidal and spherical peaks

Happy to :shipit:

Framework/Geometry/inc/MantidGeometry/Crystal/PeakShape.h
quantumsteve: Initially I stared with an enum class, but quickly reverted back due to ParaView's lack of support for them. I think strongly typed enums would work if, as you mention above, we one uses an adapter to maintain the C++03 interface.

PR #15873, opened 2016-04-08
Add algorithm to aggregate stacks images from wavelength dependend data

FedeMPouzols:
Add the algorithm as described in #13839. Earlier, python scripts had been added in #14095.
Once the algorithm is available, it will be used from the tomo GUI (#13467).

**To test:**
- Ensure that the new unit and system tests pass. As this is a purely  I/O algorithm the unit tests don't do any real functional test (would be slow). The system test does.
- Check that the documentation reads well.

Open up the algorithm dialog and set:
- InputPath to the directory where you put the files that (at ISIS)  you can grab a small one from \\olympic\Babylon5\Public\fedemp\IMAT\test_stack_test_larmor_201510_one_angle_smallish
- OutputPath to your favorite new/empty directory.
- Run the algorithm and see that what it generates makes sense.
- It has to add quite a few files so it will take some seconds. The progress bar should indicate progress
- There are some bogus directories and files included in the test data, these should be ignored and you might see warnings about it

More points to bear in mind:
- This produces outputs as fits images. You can open them with imageJ, fv, or with the Mantid Tomo GUI.
- If you set the "Log Level" to information or more, you should see messages like "Saved output aggregated image into: /home/fedemp/test/test_imgg_agg/sum_projection_00001_bands_indices_0_1000"
  by the end of the algorithm run.
- For a more intensive (and slow) test, you can grab these other stacks of images as input: 
  - \\olympic\Babylon5\Public\fedemp\IMAT\test_stack_test_larmor_201510_one_angle
  - \\olympic\Babylon5\Public\fedemp\IMAT\test_stack_test_phantom1_3angles

Fixes #13839.

Release notes: https://github.com/mantidproject/mantid/commit/3d2b31144a1434ea1a842e0300f64b1ab33397ae

---
#### Reviewer

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):
##### Code Review
- [x] Is the code of an acceptable quality?
- [x] Does the code conform to the coding standards? Is it well structured with small focussed classes/methods/functions?
- [x] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [x] If there are changes in the release notes then do they describe the changes appropriately?
##### Functional Tests
- [x] Do changes function as described? Add comments below that describe the tests performed?
- [x] How do the changes handle unexpected situations, e.g. bad input?
- [x] Has the relevant documentation been added/updated?
- [x] Is user-facing documentation written in a user-friendly manner?
[...]

Issue #15944, opened 2016-04-20
Add algorithm SaveFITS


FedeMPouzols:
The Save sibling of the already existing LoadFITS, needed and very convenient for format conversion utilities/interfaces.
Eventually it could be used as child of LoadImage/etc. (#6843), but as this is a very different format it makes sense to have it as a separate algorithm, as LoadFITS.
A basic implementation was included as part of #15873.
Later on it should support specific IMAT/other headers.

Issue #15942, opened 2016-04-19
Conversion between formats in tomography GUI


FedeMPouzols:
Implement a first working version of this new tab, as a widget.
Do it testable / MVP, but for now the the image format load/save code will have to be in the Qt view for any format other than FITS.
After this, when #6843 or similar is implemented the image load/save stuff should be taken out of the view.

PR #15964, opened 2016-04-21
Use byteswap intrinsics for changing endian.

quantumsteve:
Description of work.

This uses the byteswap intrinsics instead of reversing endianness by hand. I think casting from `double` (8 bytes) to `uint16_t` (2 bytes) is a bug, but will change it back if my assumption is incorrect.

**To test:**

see #15873

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/DataHandling/src/ImggAggregateWavelengths.cpp
+++ b/Framework/DataHandling/src/ImggAggregateWavelengths.cpp
@@ -954,7 +954,7 @@ void writeFITSHeaderBlock(const API::MatrixWorkspace_sptr img,
 }
 
 uint64_t endian_reverse(uint64_t value) {
-#if defined(_MSVC_VER)
+#if defined(_MSC_VER)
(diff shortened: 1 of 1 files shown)


FedeMPouzols:
The cast to 16 bits is not a bug. We are currently working with 16 bit FITS. Matrix workspaces are being misused for this because it is what we have at the moment, but the doubles are an overkill .What we effectively have is counts per pixel that will rarely go over 10000 or so. This is writing `BITPIX  = 16` in the FITS header and writing 64 bits per pixel would be insonsisten with that.
That block of code is now removed from here and has moved to a new algorithm (#15944) which is used here and in other places (#15942, etc.). It might be more effective to look at these details once the code stabilizes a bit.

quantumsteve:
OK, I reverted back to 16 bits. Keep the changes if you want, otherwise close.

PR #15968, opened 2016-04-22
Generalise LoadReflTBL and SaveReflTBL

### LINKED ISSUE #15864: Generalise LoadReflTBL/SaveReflTBL

MattKing06:
The aim of this PR is to allow the TBL format that is currently used by Reflectometry to be accessible to any scientific technique area that wishes to use the DataReduction Interface.

**To test:**

Fixes #15864 

---
#### Reviewer

##### Code Review
##### Functional Tests


MantidQt/CustomInterfaces/src/Reflectometry/ReflTableViewPresenter.cpp
```diff
@@ -1219,7 +1219,7 @@ void ReflTableViewPresenter::importTable() { m_tableView->showImportDialog(); }
 Export a table to TBL file
 */
 void ReflTableViewPresenter::exportTable() {
-  m_tableView->showAlgorithmDialog("SaveReflTBL");
+  m_tableView->showAlgorithmDialog("SaveTBL");
 }
```
raquelalvarezbanos: Don't forget to do the same for `importTable()`. This is currently done by the view, [here](https://github.com/mantidproject/mantid/blob/15864_generalise_loadReflTBL_and_saveReflTBL/MantidQt/CustomInterfaces/src/Reflectometry/QReflTableView.cpp#L385-L403), we should replace `LoadReflTBL` by `LoadTBL` otherwise the functionality is broken and we can not load TBL files from the interface.

Framework/DataHandling/src/SaveTBL.cpp
```diff
@@ -0,0 +1,126 @@
...
+  } else {
+    file << boost::lexical_cast<T>(val);
+  }
+  if (endsep) {
+    file << m_sep;
+  }
+  if (endline) {
+    file << std::endl;
+  }
+}
```
raquelalvarezbanos: Just a minor comment (you don't need to modify this, as I can see you have not changed the signature of this function): I think we could achieve the same behaviour with just one boolean parameter, for instance `endline`. As far as I can see, we only use either:

`writeVal(valueStr, file)`
or
`writeVal(valueStr, file, false, true)`

but never:
`writeVal(valueStr, file, false, false)`
`writeVal(valueStr, file, true, true)`
`writeVal(valueStr, file, true, false)`

if `endline` is set to true, we know we don't have to add a separator (i.e. a comma after the data), just an EOL at the end of this value.

docs/source/algorithms/LoadTBL-v1.rst
```diff
@@ -0,0 +1,49 @@
...
+|14455 |1.5  |0.333|1.44 |1.4   |1     |                            |
++------+-----+-----+-----+------+------+----------------------------+
+|14456 |1.5  |0.14 |3.522|1.6   |1     |ProcessingInstructions="1:2"|
++------+-----+-----+-----+------+------+----------------------------+
+|14457 |1.5  |0.83 |2.44 |1.2   |2     |                            |
++------+-----+-----+-----+------+------+----------------------------+
+
+
+The first line of the TBL file will define your column headings. As you can see in the table above
+the number of commas on each subsequent line must be equal to the number of column headings.
```
raquelalvarezbanos: I think the correct sentence is: "[...] As you can see in the table above the number of commas on each subsequent line must be equal to the number of column headings **minus one**"

Framework/DataHandling/src/SaveTBL.cpp
```diff
@@ -0,0 +1,126 @@
...
+          remove(filename.c_str());
+          throw std::runtime_error(groupHeading +
+                                   " Column must be of type \"int\"");
+        } else
+          writeVal<int>(row.Int(columnIndex), file);
+      } else if (columnIndex == columnHeadings.size() - 1)
+        writeVal<std::string>(row.String(columnIndex), file, false, true);
+      else
+        writeVal<std::string>(row.String(columnIndex), file);
+
```
raquelalvarezbanos: The above check:

`if (ws->getColumn(groupHeading)->type() != "int")`

is important, since the column "Group" should be an integer. However, I think we should also check the other columns. For instance, if I create the following table:

```
tab.addColumn("str", "Run")
tab.addColumn("double", "Angle")
tab.addColumn("int", "Group")
tab.addColumn("str", "Options")
tab.addRow(["100", 0.5, 1, ""])
```

and try to save it using `SaveTBL`, I get a segfault on Ubuntu, on Windows, I get:

*Error in execution of algorithm SaveTBL
bad allocation*

Something like this should work:

```
for (size_t columnIndex = 0; columnIndex < columnHeadings.size();
         columnIndex++) {

  auto heading = columnHeadings[columnIndex];
  auto headingType = ws->getColumn(heading)->type();

  // 'Group' column
  if (columnIndex == columnHeadings.size() - 2) {
    if (headingType != "int"){
      file.close();
      remove(filename.c_str());
      throw std::runtime_error("Column " + heading + " must be of type " + headingType);
    }
    writeVal<int>(row.Int(columnIndex), file);
  }
  // Any other column
  else{
    if (headingType != "str"){
      file.close();
      remove(filename.c_str());
      throw std::runtime_error("Column " + heading + " must be of type " + headingType);
    }
    if (columnIndex == columnHeadings.size() - 1){
      writeVal<std::string>(row.String(columnIndex), file, false, true);
    }
    else {
      writeVal<std::string>(row.String(columnIndex), file);
    }
  }
 }
}
```

This way we are providing more information about what users are doing wrong.

raquelalvarezbanos:
Changes look good, both algorithms work as expected. I am just going to wait until PR # 15943 is reviewed, then I will move this to the next stage so it can be reviewed by the second tester.

raquelalvarezbanos:
@MattKing06 see my [commit](https://github.com/mantidproject/mantid/pull/15943/commits/754fa3db9901806aa60afc70bf937807f269929d), I had to make a minor change to `LoadTBL`.

raquelalvarezbanos:
These changes have been merged into branch [15672_Generalise_ReflTableViewPresenter](https://github.com/mantidproject/mantid/tree/15672_Generalise_ReflTableViewPresenter) and conflicts were resolved. I would say that, to avoid resolving the same conflicts again, we can close this PR, changes will be merged into master when my branch is merged.

MattKing06:
@mantidproject/gatekeepers please review this but close this PR instead of merging it.

FedeMPouzols:
I see, this branch has been merged into another branch / separate pull request. You can close this I'd say (and it can be re-openend as well). Just explain in the other PR how the testing was done.

Issue #13950, opened 2015-10-13
Logically dead code in FindPeaks.cpp


NickDraper:
FindPeaks.cpp Line 1088 onwards

This was highlighted by coverity 1301341

A change you made in this commit
https://github.com/mantidproject/mantid/commit/d5970611f3c279d64d185ca7e1f2a8b7011fb774

meant that the change to the value of fitresult was hidden from the variable of the same name does not occur.
This means the large if statement lines 1101 - 1140 will never run.

Please verify that was your intention, and correct if it is not.

peterfpeterson:
A first attempt at this is in #13667.

FedeMPouzols:
I can confirm that if you try to fix this issue by not hiding/redefining the variable then several unit tests break (besides AlgorithmsTest_FindPeaksTest):

- PythonAlgorithms_EnggCalibrateTest
- PythonAlgorithms_EnggFitPeaksTest
- MDAlgorithmsTest_IntegratePeaksMD2Test
- MDAlgorithmsTest_IntegratePeaksMDTest
- PythonAlgorithms_TOSCABankCorrectionTest

And I'd anticipate that several or maybe many system tests will fail as well.

I know the first two tests (Engg ones) and they way the get broken doesn't look good (a totally wrong fit result because of wrong peaks found in one case, and lots of NaN/Infinity problems in the second case).

Refs #13667.

FedeMPouzols:
This is also a cppcheck (1.73) issue: #15978.

PR #13667, opened 2015-09-16
Fix incorrect reporting of background estimation failure in FindPeaks

martyngigg:
A variable had been redefined in an inner scope that shadowing the outer one.

PR #15978, opened 2016-04-25
Fix warnings from cpp-check 1.73

quantumsteve:
Description of work.

This fixes some of the warnings found with cpp-check 1.73. 

**To test:**

For most of these changes, code review should be sufficient. Also see the comments below about a "fix" that breaks several unit tests and using a smart pointer to eliminate an if statement cppcheck was warning about.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Algorithms/src/FindPeaks.cpp
+++ b/Framework/Algorithms/src/FindPeaks.cpp
@@ -1087,7 +1087,7 @@ int FindPeaks::findPeakBackground(const MatrixWorkspace_sptr &input,
         "No 7th column for use FindPeakBackground result or not. ");
 
   if (peaklisttablews->rowCount() > 0) {
-    int fitresult = peaklisttablews->Int(0, 6);
+    fitresult = peaklisttablews->Int(0, 6);
     g_log.information() << "fitresult=" << fitresult << "\n";
   }
 
--- a/Framework/Algorithms/src/FindPeaks.cpp
+++ b/Framework/Algorithms/src/FindPeaks.cpp
@@ -1087,7 +1087,9 @@ int FindPeaks::findPeakBackground(const MatrixWorkspace_sptr &input,
         "No 7th column for use FindPeakBackground result or not. ");
 
   if (peaklisttablews->rowCount() > 0) {
-    fitresult = peaklisttablews->Int(0, 6);
+    // setting fitresult currently breaks several tests
+    // see issues ##13667, 13950 and 15978
+    int fitresult = peaklisttablews->Int(0, 6);
     g_log.information() << "fitresult=" << fitresult << "\n";
   }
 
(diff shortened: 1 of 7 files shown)


Framework/Algorithms/src/FindPeaks.cpp
```diff
@@ -1087,7 +1087,7 @@ int FindPeaks::findPeakBackground(const MatrixWorkspace_sptr &input,
         "No 7th column for use FindPeakBackground result or not. ");
 
   if (peaklisttablews->rowCount() > 0) {
-    int fitresult = peaklisttablews->Int(0, 6);
+    fitresult = peaklisttablews->Int(0, 6);
```
quantumsteve: @peterfpeterson Would you please check if there is an issue setting `fitresult` and the above change is correct. Should this change be in the release notes?

MantidPlot/src/ApplicationWindow.cpp
```diff
@@ -16554,18 +16551,14 @@ void ApplicationWindow::CatalogLogin() {
 }
 
 void ApplicationWindow::CatalogSearch() {
-  if (catalogSearch == NULL || catalogSearch) {
```
quantumsteve: @ElliotAOram
Does switching to a smart pointer eliminate the check you added?

ElliotAOram: @quantumsteve 
I don't think I actually wrote this code. I think it's likely that my name is there in the git blame due to a clang format. Regardless, given the nature of the check I think a smart pointer should work.

Framework/Algorithms/src/FindPeaks.cpp
FedeMPouzols: That apparently innocent line is a can of worms: https://github.com/mantidproject/mantid/pull/13667, https://github.com/mantidproject/mantid/issues/13950 (from a coverity issue that caught this).

peterfpeterson: @FedeMPouzols is right. Unshadowing that variable is bad. I haven't had the time to track down why it is so important, but the fits are significantly better with it there.

quantumsteve: @FedeMPouzols @peterfpeterson Thanks. I reverted the change and added a comment referencing the issue number above.

MantidPlot/src/ApplicationWindow.cpp
quantumsteve: @ElliotAOram Thank you for checking.

martyngigg:
I agree with the all of the changes that have been made. I too have come across the bug in `FindPeaks` that breaks many tests. Should someone have a maintenance issue to investigate this?

martyngigg:
Other than that, let's :shipit:

FedeMPouzols:
We have this issue: #13950, which was tagged as "coverity". I've added "cppcheck". Maybe it can be tackled for the next maintenance period. At the moment we also have the GSL2 changes ongoing. I'd wonder how those changes would interplay with this issue.

Issue #15970, opened 2016-04-22
ISIS direct inelastic config script should work propertly on Express experiments


abuts:
Current direct inelastic config script copies users scripts requested to a RB folder  only for a user who is a primary user for this experiment (e.g. has only this experiment in his experiment list) This done for simplicity and performance and mainly works, but always fails for express experiments, as users on this experiment are always the people, who participate in other experiments. 

As we now schedule range of such experiments, some instrument scientists want users scripts to be copied to such experiments too.

PR #15982, opened 2016-04-25
ISIS autoconfig file copies multiple user's sample files

abuts:
Solves #15970 -- see details in the ticket description. 

ISIS direct inelastic users configuration script was copying sample reduction files for main experiment user only despite all experiments were added to the user profile. 

This is not-satisfactory for express experiments as no main user exist for this experiment, so after the changes, autoconfiguration script copies sample files for each experiment a user participates in. 

**To test:**
Test by code review. 

No much additional testing is necessary and actually possible as it relies and needs very specific computer configuration, provided by isiscompute. In addition to that, I've already deployed the script on isiscompute where it belongs and it works fine. The changes illustrated by two new unit tests, which verifies that for a user, participating in three experiments three configuration files were copied to three user' RB folders.

Changes are small and I hope obvious though clogged by pylint changes -- unfortunately can not do anything about it -- it does not let me further without decreasing pylint count.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/scripts/test/ISISDirecInelasticConfigTest.py
+++ b/scripts/test/ISISDirecInelasticConfigTest.py
@@ -480,8 +480,6 @@ class ISISDirectInelasticConfigTest(unittest.TestCase):
         # clean up
         if os.path.exists(os.path.join(self.userRootDir,'.mantid')):
             shutil.rmtree(os.path.join(self.userRootDir,'.mantid'),ignore_errors=True)
-        if os.path.exists(rbdir1):
-            shutil.rmtree(rbdir1,ignore_errors=True)
         if os.path.exists(rbdir2):
             shutil.rmtree(rbdir2,ignore_errors=True)
         if os.path.exists(rbdir3):
@@ -491,9 +489,9 @@ class ISISDirectInelasticConfigTest(unittest.TestCase):
 
 
 if __name__=="__main__":
-   test = ISISDirectInelasticConfigTest('test_copy_multiplpe_ucf')
-   test._set_up()
-   test.run()
-   test._tear_down()
+   #test = ISISDirectInelasticConfigTest('test_copy_multiplpe_ucf')
+   #test._set_up()
+   #test.run()
+   #test._tear_down()
 
-   #unittest.main()
+   unittest.main()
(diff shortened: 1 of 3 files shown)


FedeMPouzols:
@abuts : I assume this doesn't need a release note. Let me know if you'd like to add something.

abuts:
@FedeMPouzols :, thanks.  Its administrative ISISCompute script. I do not believe any release notes  are necessary due to very few people would be able to run it properly (ISISCOMPUTE administrators) and very specific application area (isiscompute itself). In addition to that, its a bugfix, made on request of the instrument scientists, and though they will be happy when it works nobody care about the details.

scripts/test/ISISDirecInelasticConfigTest.py
```diff
@@ -69,21 +71,31 @@ def _set_up(self):
...
-        fh.write('#Test reduction file\n')
-        fh.write('Contents={0}'.format(contents))
-        fh.close()
-        return full_file
+            file_name = mcf._sample_reduction_file(instr_name)
+            all_files.append(file_name)
+            full_file = os.path.join(file_path,file_name)
+            if os.path.isfile(full_file):
+                os.remove(full_file)
+            fh=open(full_file,'w')
```
FedeMPouzols: For this type of code I find very convenient the scoping
`with open(...) as fh:`, which tends to be safer and saves you from having to do `fh.close()`.

FedeMPouzols:
This PR is to update in the Mantid codebase inelastic scripts that are already deployed and used on the "isiscompute" cluster at ISIS.
- Changes the way files are looked for and copied to support so-called "express" experiments (handled by instrument scientists)
- Proper funcitonal tests are done in practice by the admins and users of the cluster, and cannot be replicated here. `ISISDirecInelasticConfigTest` has been updated.
- Code looks good. A pylint disable has been fixed. A "disable=too-many-instance-attributes" has been added which I think  is perfectly reasonable in its context.
- No release notes needed for this as it is a very local and specialized change.

:shipit:

scripts/test/ISISDirecInelasticConfigTest.py
abuts: good point, thanks -- unfortunately can not recall the exact form each time I deal with such issue in Python. Will try to use it next time.

PR #15988, opened 2016-04-26
Vesuvio - VesuvioCommands check height of peak rather than first/last bin

### LINKED ISSUE #15981: Vesuvio - VesuvioCommands system test failing on OSX and RHEL6
### Expected behavior
* VesuvioCommandsTest should pass on all OS

### Actual behavior
* This system test is failing on OSX and RHEL6 (potentially others)

* Run system tests on OSX or RHEL6

### Platforms affected
* OSX, RHEL6

ElliotAOram:
The VesuvioCommandsTest now checks the maximum height of the calculated data for comparison. This value was seen to be far more consistent across different platforms rather than the first/last bin which were seemingly sensitive to the noisy data at the start and end 

**To test:**
- Ensure the system test `VesuvioCommandsTest` passes ideally checking this on a few different OS. 
  - I have checked this locally on Windows, Centos 6 and Ubuntu 14.04 and it passes

Fixes #15981

---
#### Reviewer

##### Code Review
##### Functional Tests

Refs #15981

### DIFF
--- a/Testing/SystemTests/tests/analysis/VesuvioCommandsTest.py
+++ b/Testing/SystemTests/tests/analysis/VesuvioCommandsTest.py
@@ -4,6 +4,7 @@ and that mantid can be imported
 """
 import stresstesting
 import platform
+import numpy as np
 
 from mantid.api import (WorkspaceGroup, MatrixWorkspace)
 from mantid.simpleapi import *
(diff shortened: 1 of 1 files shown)


martyngigg:
Nice solution! 

I'm being picky but I wonder if we should also consider testing the bin index that the peak height falls in matches. What do you think?

martyngigg:
I hadn't realised this was holding up the nightly release on RHEL & OSX so I'm going to push this through without second review.

martyngigg:
@ElliotAOram I still think it's worth adding a check on the expected bin index of the peak but let's do that as a separate issue.

ElliotAOram:
@martyngigg I think you are right #16056

PR #15994, opened 2016-04-26
Live data prototype

lottiegreenwood:
Adds in method for instrument view that allows python to access the current detector ID the mouse is hovering over

**To test:**

When using mantid python call instrumentwidgetpicktab.get_currentPickID()

For use when using the live-data-prototype only, not for merging into master


lottiegreenwood:
@LamarMoore , @OwenArnold

Issue #15518, opened 2016-03-02
Multi-dataset fitting in MuonAnalysis interface


tom-perkins:
Created from #13096 item 9 and muon meeting on 25/2/16.

One use case for this is fitting the same function to all periods in a run, perhaps with some shared parameters, and then using the sequential fit to repeat this for different runs.

Have Chi^2 for each as well as a global value.

Circulate Balsamiq mockups to scientists to determine layout and how this will be presented.

tom-perkins:
Mantid roadmap: estimate release 3.8, 3-4 months including time for discussion with scientists

tom-perkins:
Meeting with scientists 23/3/16 to discuss mockups.
Comments contained in https://github.com/mantidproject/documents/pull/16

PR #16008, opened 2016-04-27
Fix result table regression in MuonAnalysis and add test

### LINKED ISSUE #16003: MuonAnalysis: fix regression and add test
[Bug is **not** in release 3.6 but has appeared more recently - not seen by users]

### Expected behavior
The "run number" field in MuonAnalysis should have period numbers as was the case in PR #15615.
If data is single-period then no period label should be displayed.

### Actual behavior
If the data is single-period then the "*#*1" label is displayed in place of the period label

(Only reproducible on recent builds since #15817 merged)
- Fit a single-period dataset in MuonAnalysis
- Generate a results table with the `run_number` field selected
- Run number is (e.g.) "62912:*#*1" rather than the expected "62912"

### Platforms affected
All

---

The bug was introduced by the fix in #15817 because there are no tests for this functionality.
To prevent this happening again, refactor the relevant functions so that tests can be added for them and regressions like this one can be detected.

tom-perkins:
Fixed a bug where the period label was being set wrongly for single-period data.
To prevent the bug reoccuring, refactored the method into a helper class and added tests.

The tested method generates a run/period label from the workspace name. (The function that generates the workspace name itself has not been refactored out and tested because it is closely coupled to the UI. However, this is anticipated to be done as part of the ongoing work for #15518, so testing this has been left until then.)

**To test:**
- _Code review:_ satisfy yourself that the unit tests cover the functionality and expose the bug
- Can also be tested with the files in `\\olympic\babylon5\Scratch\TomPerkins\Data\MUSR`:  
  - Open MuonAnalysis interface and set instrument to MUSR. Make sure data files are in your search path.
  - Type a single run number or a range in the box to load, e.g. "15189-91" or "24580". `MUSR00024580` and `MUSR00024581` are single-period and `MUSR00015189-91` are multi-period.
  - Go to Data Analysis tab and fit something (doesn't matter what)
  - Go to Results Table tab and click "Create Table"
  - The first column in the generated table should have the run number only for single-period data, and 'run number: period' for multiperiod data.

Fixes #16003. 

Does not need to be in the release notes as the bug is not in release 3.6.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Muon/MuonAnalysisHelper.h
+++ b/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Muon/MuonAnalysisHelper.h
@@ -82,6 +82,10 @@ appendTimeSeriesLogs(boost::shared_ptr<Mantid::API::Workspace> toAppend,
                      boost::shared_ptr<Mantid::API::Workspace> resultant,
                      const std::string &logName);
 
+/// Get "run: period" string from workspace name
+MANTIDQT_CUSTOMINTERFACES_DLL QString
+runNumberString(const std::string &workspaceName, const std::string &firstRun);
+
 /**
  * A class which deals with auto-saving the widget values. Widgets are
  * registered and then on any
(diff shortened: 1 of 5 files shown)


raquelalvarezbanos:
Works well: tested both multi-period and single-period datasets. In the first case, period numbers were added to the run number in the results table, whereas in the second case no period information was added to the run number. Good unit tests, hopefully this functionality will not break again.

:shipit:

FedeMPouzols:
Great to see those tests.

PR #16013, opened 2016-04-27
Incorrect convolution when uneven domains

### LINKED ISSUE #15954: Incorrect convolution when uneven domains
### Expected behavior
Correct convolution using a resolution function centered at `E=0` over a range `[-Emin, Emax] (Emin>0)` should be correctly computed when `Emin != Emax`

### Actual behavior
When `Emin != Emax`, the convolution shows a step decrease at `E=(Emin+Emax)/2`.

Example: convolution of [resolution](https://github.com/mantidproject/mantid/files/228851/resolution.zip) and [signal](https://github.com/mantidproject/mantid/files/228853/signal.zip) results in

convolution

### Solution when the signal originates from an analytic model
In the context of curve-fitting, convolving the resolution defined over `[-Emin, Emax]` with a model defined over `[-(Emin+Emax), (Emin+Emax)]` will solve this problem because the extended energy range guarantees complete overlap between the model and the resolution over the `[-Emin, Emax]` range.
 
[Example:](https://github.com/mantidproject/mantid/files/230297/synthetic_test.zip)

Here is a resolution and model signal evaluated in the [-0.12, 0.52] energy domain, along with a extended resolution (zero-padded) and extended signal model evaluated in the [-0.64, 0.64] range
synthetic_test
Hre is the result of convolving the resolution with the model, and convolving the extended resolution with the extended model. The dip starting at (0.52+0.12)/2=0.31 is gone
synthetic_test_convolved

### Platforms affected
All

jmborr:
Description of work.

**To test:**

Download [benchmark.zip](https://github.com/mantidproject/mantid/files/241212/benchmark.zip) and unzip. Run the python script in the python window of MantidPlot. The script will fit the signal against a Two-Lorentzian model in the FFT mode (when using a symmetric range) and in the Direct mode (when using an asymmetric range).

Fixes #15954.

[Release notes](https://github.com/mantidproject/mantid/blob/15954_convolution_extend_range/docs/source/release/v3.7.0/framework.rst) have been updated

---
#### Reviewer

##### Code Review
##### Functional Tests


jmborr:
@mantid-roman I was able to fix the problems by doing the convolution in direct space (avoiding the fourier transforms). The convolution is slower but gives the correct results. I used three benchmarks:

1. Convolution of two workspaces:
comparison

2. Fitting using an "artificial" resolution function and scattered signal:
comparison

3. Fitting to real data (taken at the BASIS beamline) using the instrument resolution function:
comparison

So my question is whether you are comfortable with these changes, meaning doing the convolution in direct rather than reciprocal space

mantid-roman:
@jmborr FFT should still be used when appropriate. We shouldn't lose efficiency unnecessarily. The function must be able to select the right algorithm for the data.

What about zero-padding? Didn't work?

jmborr:
@mantid-roman So for the moment I could set it up so that the algorithm will use the FFT method when the domain is symmetric with respect to inversion, and use the direct method otherwise. Uneven domains are a small percentage of the use cases. Then, in a future issue we can implement the FFT method for the asymmetric domain. Would that work for the moment?

mantid-roman:
@jmborr Absolutely. I just think that we shouldn't treat data where |Emin| is only slightly different form |Emax| as asymmetric.

jmborr:
@mantid-roman OK, I could do the switching when both differ by say, 5%. More than this and the effects are too noticeably

jmborr:
@mantid-roman Now the Convolution selects FFT or Direct mode of calculations based on how StartX and EndX differ from each other. If you are happy with this, I will start amending the documentation and release notes.

mantid-roman:
@jmborr The documentation and release notes will be on your branch (rst files) and will be a part of the pull request. I'll be happy to review the PR.

jmborr:
@mantid-roman Hi Roman, can you please give this a go?

mantid-roman:
@jmborr Sure

Framework/CurveFitting/src/Functions/Convolution.cpp
```diff
@@ -79,30 +81,55 @@ struct RealFFTWorkspace {
...
+  double dx =
+      (xValues[nData - 1] - xValues[0]) / static_cast<double>((nData - 1));
+  // positive x-values:
+  auto ixP = static_cast<size_t>(xValues[nData - 1] / dx);
+  auto ixN = nData - ixP - 1; // negative x-values (ixP+ixN=nData-1)
+
+  // determine wether to use FFT or Direct calculations
+  int assymmetry = abs(static_cast<int>(ixP - ixN));
+  if (xValues[0] * xValues[nData - 1] < 0 &&
+      assymmetry > 0.02 * static_cast<double>(ixP + ixN)) {
```
mantid-roman: 0.02 should be a named (upper case) constant defined in a anonymous namespace.

Framework/CurveFitting/src/Functions/Convolution.cpp
```diff
@@ -15,7 +17,7 @@
 #include <gsl/gsl_fft_halfcomplex.h>
 
 #include <sstream>
-
+#include <iostream>
```
mantid-roman: Is it being used?

Framework/CurveFitting/src/Functions/Convolution.cpp
```diff
@@ -270,7 +297,132 @@ void Convolution::function(const FunctionDomain &domain,
...
+    m_resolution.resize(nData);
+  }
+  resolution->function1D(m_resolution.data(), xValues, nData);
+
+  // Reverse the axis of the resolution data
+  std::reverse(m_resolution.begin(), m_resolution.end());
+
+  // check for delta functions
+  std::vector<boost::shared_ptr<DeltaFunction>> dltFuns;
+  double dltF = 0;
```
mantid-roman: Code dealing with the delta functions in this method duplicates code from functionFFTMode. Can this code be moved to a separate method(s) or standalone function(s)?

Framework/CurveFitting/src/Functions/Convolution.cpp
```diff
@@ -270,7 +297,132 @@ void Convolution::function(const FunctionDomain &domain,
...
+        tmp += outExt[i + j] * m_resolution[j];
+      }
+      out[i] = tmp * dx;
+    }
+  } else {
+    values.zeroCalculated();
+  }
+
+  if (dltF != 0.0 && !deltaShifted) {
+    // If model contains any delta functions their effect is addition of scaled
```
mantid-roman: This code is also duplicated.

Framework/API/inc/MantidAPI/FunctionValues.h
```diff
@@ -101,6 +101,9 @@ class MANTID_API_DLL FunctionValues {
   /// Return the calculated values as a vector
   std::vector<double> toVector() const { return m_calculated; }
 
+  /// Return a reference to the calculated values
+  std::vector<double> &getCalculated();
```
mantid-roman: If this method isn't used I would prefer to remove it. Non-virtual methods getCalculated(size_t) or operator[](size_t) can be used to access the calculated data.

jmborr:
@mantid-roman I'm implemented the comments you wrote about except those concerning duplicate code. I don't want  shared code between the two modes because in time I will write the necessary code to convolve an asymmetric domain using FFT. The current direct mode is meant to be a quick fix, and that's why I don't want any code to be shared with the FFT mode.

mantid-roman:
@jmborr Although this isn't my preferred way forward (because quite often temporary fixes become permanent solutions) but could you then comment in the code that the duplication will be dealt with mentioning the issue number. Could you also write unit tests for the delta-function convolution code in both cases (FFT and direct).

jmborr:
@mantid-roman I included the test for the convolution of the Delta Dirac. Also, can you please clarify for me which issue number are you referring to in your comment?

mantid-roman:
@jmborr Code duplication should be dealt with sooner rather than later. The best way not to forget about it is to open an issue for its fixing. And mention this issue in a comment in the code.

jmborr:
@mantid-roman Hi Romain. I opened an issue for next release in order to deal with code duplication. Can you please check if there is anything missing for this branch?

mantid-roman:
:shipit:

FedeMPouzols:
In the files changes I see a big modification in the release notes (`framework.rst`). From these there is the note about this branch and many other unrelated notes. Have the branches mixed-up?
The commit that introduces changes in `framework.rst` is: https://github.com/mantidproject/mantid/pull/16013/commits/180b5963805eb119ed93a218e96ec85ed1b9327e

jmborr:
@mantid-roman @FedeMPouzols @AndreiSavici Thank you all for your help! :ribbon:

PR #16016, opened 2016-04-28
Refactor LoadEventNexus

### LINKED ISSUE #15840: Apply hdf5 library to LoadEventNexus

wdzhou:
Description of work

Refactor LoadEventNexus for better code style in order to use hdf5 library.

**To test:**

Fixes #15840.
1. The refactored codes won't break any test.
2. Compare the speed to load a large Nexus file before and after the refactoring.

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/DataHandling/src/LoadEventNexus.cpp
```diff
@@ -1982,10 +1116,16 @@ void LoadEventNexus::loadEvents(API::Progress *const prog,
 
   for (size_t i = bank0; i < bankn; i++) {
     // We make tasks for loading
-    if (bankNumEvents[i] > 0)
-      pool.schedule(new LoadBankFromDiskTask(
+    if (bankNumEvents[i] > 0) {
+      LoadBankFromDiskTask *task = new LoadBankFromDiskTask(
```
martyngigg: What is the reason we are removing the parallelization? Is there evidence it does help with performance?

Framework/DataHandling/src/LoadEventNexus.cpp
```diff
@@ -1982,10 +1116,16 @@ void LoadEventNexus::loadEvents(API::Progress *const prog,
...
           this, bankNames[i], classType, bankNumEvents[i], oldNeXusFileNames,
-          prog2, diskIOMutex, scheduler, periodLogVec));
+          prog2, diskIOMutex, scheduler, periodLogVec, this->getLogger());
+      task->run();
+
```
martyngigg: There is now a memory leak as `task` is not deleted. If this structure is kept then we should use a `Mantid::Kernel::make_unique` and a `std::unique_ptr` to manage the memory.

Framework/DataHandling/src/LoadEventNexus.cpp
```diff
@@ -1982,10 +1116,16 @@ void LoadEventNexus::loadEvents(API::Progress *const prog,
...
+      // pool.schedule(new LoadBankFromDiskTask(
```
martyngigg: We shouldn't keep commented out code in the code base as it's confusing for future developers.

Framework/DataHandling/src/LoadEventNexus.cpp
peterfpeterson: Since the mutex(es) prevents things from reading in parallel this runs in series anyhow and complicates the code. [The tests](https://github.com/OwenArnold/hdf5_vs_nexus/blob/master/read_results.md) showed that using HDF's enumeration mechanism is significantly faster for the cases we are looking at.

Framework/DataHandling/src/LoadBankFromDiskTask.cpp
```diff
@@ -0,0 +1,909 @@
...
+        if ((tof >= alg->filter_tof_min) && (tof <= alg->filter_tof_max)) {
+          // Handle simulated data if present
+          if (have_weight) {
+            double weight = static_cast<double>(event_weight[i]);
+            double errorSq = weight * weight;
+            Mantid::DataHandling::LoadEventNexus::WeightedEventVector_pt
+                eventVector = alg->weightedEventVectors[periodIndex][detId];
+            // NULL eventVector indicates a bad spectrum lookup
+            if (eventVector) {
+#if !(defined(__INTEL_COMPILER)) && !(defined(__clang__))
```
quantumsteve: Would you also remove both this macro and the very similar one on line 219? `emplace_back` is available on all of our supported platforms.

Framework/DataHandling/src/LoadEventNexus.cpp
martyngigg: Ok, sounds good. Thanks for clarifying.

Framework/DataHandling/src/LoadBankFromDiskTask.cpp
```diff
@@ -0,0 +1,912 @@
...
+        alg(alg),
+        entry_name(entry_name), pixelID_to_wi_vector(alg->pixelID_to_wi_vector),
+        pixelID_to_wi_offset(alg->pixelID_to_wi_offset), prog(prog),
+        event_id(event_id), event_time_of_flight(event_time_of_flight),
+        numEvents(numEvents), startAt(startAt), event_index(event_index),
+        thisBankPulseTimes(thisBankPulseTimes), have_weight(have_weight),
+        event_weight(event_weight), m_min_id(min_event_id),
+        m_max_id(max_event_id) {
+    // Cost is approximately proportional to the number of events to process.
+    m_cost = static_cast<double>(numEvents);
```
martyngigg: Is `m_cost` necessary now that this is no longer a `Task`?

Framework/DataHandling/src/LoadBankFromDiskTask.cpp
wdzhou: Will do

Framework/DataHandling/src/LoadBankFromDiskTask.cpp
wdzhou: I will remove it.  Thanks.

peterfpeterson:
Did you have to name it `*Helper`?

wdzhou:
Not necessary.  Any name that you think better?

peterfpeterson:
It looks like it is slower now. I'm not sure what is the cause, or whether we want to worry about it since this is a temporary state.

* PG3_29571  - 974MB
  * before 11s, 11s, 2s, 2s
  * after 5s, 5s, 5s, 5s
* NOM_77533 - 5.9GB
  * 59s, 8s, 8s, 8s
  * 2m49s, 33s, 35s, 33s

wdzhou:
It does not look good.  I think that I should profile it.

peterfpeterson:
This version had serious performance degradation. A new attempt has been started.

Issue #15770, opened 2016-03-30
Add support for inelastic instruments to MonteCarloAbsorption algorithm 


martyngigg:
The current `MonteCarloAbsorption` algorithm only supports elastic data and needs to be extended to deal with wavelengths incident/final wavelengths.

A good test case would be to compare the results with the `FlatPlateAbsorption` algorithm, that already supports inelastic instruments.

Issue #15443, opened 2016-02-24
Document LoadEventNexus with a Table


KarlPalmen:
Add to the LoadEventNexus v1 documentation a table explain where various items are read from the Nexus and put into the Workspace.

This is part of the work for issue #14940 .

KarlPalmen:
The issue of logs will be dealt by issue #15607 for LoadNexusLogs, which is called to load logs.

PR #15923, opened 2016-04-18
Inelastic Support for MonteCarloAbsorption algorithm

martyngigg:
This work adds support for inelastic instruments to the `MonteCarloAbsorption` algorithm. I have chosen not to include this in the interface and require it to have been set on the workspace. Issue #15922 has been created to ensure `ConvertUnits` does this consistently for all modes.

**To test:**

The following script runs the MC absorption calculation against the flat plate & cylinder algorithms that use numerical integration. Plotting both against each other should show similar factors calculated in each case.

``` python
from mantid.simpleapi import *

# -------------------------------------
# Comment/uncomment the required mode
#emode = 'Indirect'
emode = 'Direct'

if emode == 'Indirect':
    sample_file = 'IRS26176.RAW'
    efixed = 1.845
else:
    sample_file = 'MAR11060.RAW'
    efixed = 12.0

# Vanadium
sample_formula = 'V'
# geometry in cm
sample_height = 3.0
sample_radius = 0.1
# atoms/A^3
sample_density = 0.073

# simulation options
nlambda_cyl = 20
nlambda_mc = 20

# ------------------- Setup the workspace ------------------------

red_ws = Load(Filename=sample_file)
# Just do a single spectrum
red_ws = CropWorkspace(red_ws, StartWorkspaceIndex=2, EndWorkspaceIndex=2)
SetInstrumentParameter(red_ws, ParameterName="deltaE-mode", Value=emode, ParameterType="String")
if emode == 'Indirect':
    SetInstrumentParameter(red_ws, ParameterName="Efixed",Value=str(efixed), ParameterType="Number")
elif emode == 'Direct':
    AddSampleLog(red_ws, LogName='Ei', LogText=str(efixed), LogType='Number')
red_ws = ConvertUnits(red_ws, Target='Wavelength',
                      EMode=emode)
SetSampleMaterial(red_ws, ChemicalFormula=sample_formula,
                  SampleNumberDensity=sample_density)

# --------------------- Cylinder absorption ---------------------

red_ws_ass_cylabs = CylinderAbsorption(InputWorkspace=red_ws,
                                       NumberOfWavelengthPoints=nlambda_cyl,
                                       CylinderSampleHeight=sample_height,
                                       CylinderSampleRadius=sample_radius,
                                       NumberOfSlices=100,
                                       NumberOfAnnuli=100,
                                       EMode=emode, EFixed=efixed)

# --------------------- Flat Plate absorption ---------------------

[...]

PR #16025, opened 2016-04-29
Implement non-zero beam size in MonteCarloAbsorption algorithm

### LINKED ISSUE #15887: Introduce concept of beam size to MonteCarloAbsorption algorithm
The current implementation of the `MonteCarloAbsorption` algorithm treats the source as a single point and all neutrons are emitted from this. In general the beam has some defined size & shape and the algorithm should be updated to allow for this.

martyngigg:
The `MonteCarloAbsortion` algorithm now has the concept of a beam size rather than all neutrons emanating from a zero-size point. The size of the beam is currently fixed to the width and height of the bounding box of the sample. This can be more flexible in the future.

**To test:**

The instructions for #15770 provide a script that can be used to drive the algorithm. The output no longer matches the integration algorithms as the average path lengths will be shorter as the beam is no longer divergent, resulting in slightly less absorption (higher transmitted factor)

Fixes #15887

Release Notes: I have decided not to update the release notes yet and wait until the general roadmap issue (#15443) is further along.

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/Algorithms/src/MonteCarloAbsorption.cpp
```diff
@@ -184,381 +206,57 @@ void MonteCarloAbsorption::exec() {
       } else {
         // elastic case already initialized
       }
-      doSimulation(detPos, lambdaIn, lambdaOut, yValues[bin], eValues[bin]);
+      std::tie(signal[j], std::ignore) =
```
FedeMPouzols: Is this intentionally throwing away those errors?
I see that the unit test checks the 1/N^2 errors returned from `MCAbsorptionStrategy::calculate()`, but with this the algorithm never returns those values, just zeros?

martyngigg: It is at the moment for consistency with the other absorption algorithms that do nothing with errors. It was unclear to me what to do with the errors here when interpolate. Should they stay the same or not. Do you have an opinion on that?

FedeMPouzols: I don't have any well informed opinion I'm afraid. Yes, the other algorithms produce all-zero errors so I'd also go for consistency.
Maybe this could be brought up to relevant expert users. Is the fact that the errors are usually set to zero a feature/well justified, or is it a limitation/improvable? Are they ever used in any other calculations?

martyngigg: I'm not sure to be honest. The original author of the algorithms is not around and it makes no mention of the errors anywhere.

FedeMPouzols:
- Works well for me, and I can see the expected decrease in the absorption factor compared to the previous version.

screenshot_mc_nz_beam_size

(compare with [previous version](https://github.com/mantidproject/mantid/pull/15923#issuecomment-215139811)). Maybe the sudden drop towards the maximum wavelength is a boundary issue in the method/calculations?

- Code looks very carefully organized. It refactors `MonteCarloAbsorption` and introduces `RectangularBeamProfile` and related base/support classes.
- Thoroughly tested, and it extends other related tests as well. Lovely tests, including mocked tests on the MC/random numbers mechanism.
- Release notes will be updated after some more work is done.

Let's :shipit:

PR #16020, opened 2016-04-28
Add algorithm SaveFITS

FedeMPouzols:
Adds a new algorithm, SaveFITS, to support other imaging/tomography algorithms and interfaces.

**To test:**
- Code review
- Check the documentation of the algorithm, and the usage examples. It should be easy to understand.
- The unit, docs and system tests should pass and look sensible.
- There's a few fixes in LoadFITS (other issues are open against it for more refactoring and fixes) which are covered by unit and system tests. It's been modified to not load all the headers at once before loading the file contents (this was causing user unfriendly delays without progress feedback when loading large stacks over the network).

Fixes #15944, 
Fixes #13887.

Release notes: https://github.com/mantidproject/mantid/commit/253929fd99b20ee2475f1f3d004b456cda08f254

---
#### Reviewer

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):
##### Code Review
- [x] Is the code of an acceptable quality?
- [x] Does the code conform to the coding standards? Is it well structured with small focussed classes/methods/functions?
- [x] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [x] If there are changes in the release notes then do they describe the changes appropriately?
##### Functional Tests
- [x] Do changes function as described? Add comments below that describe the tests performed?
- [x] How do the changes handle unexpected situations, e.g. bad input?
- [x] Has the relevant documentation been added/updated?
- [x] Is user-facing documentation written in a user-friendly manner?
- [x] Has developer documentation been updated if required?
- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

Issue #6843, opened 2012-10-24
Write a generic LoadImage algorithm


NickDraper:
This issue was originally [TRAC 5997](http://trac.mantidproject.org/mantid/ticket/5997)

Write an algorithm to load an image file to a workspace, and create a 'basic' instrument definition if we cannot match up to an existing one based on filename.  This should vote to open files like any other load algorithm but not normally return a particularly high preference so other more specific loaders can override it.

File type we absolutely have to support:
- TIFF
- FITS

It would make sense to use a library that allows us to generically handle multiple file formats.  Two that have spring to mind are:
1. BOOST Generic image library hereafter known as GIL
1. Freeimage

Both work by using the underlying specific libraries LibTIFF, libPNG etc, but give a consistent interface on top.  FreeImage looks simpler as it is just about loading and saving images, GIL is more complex as it also supports image algorithms in a generic way, but is header only and therefore already available on all platforms.

Neither have support for FITS out of the box, but both allow a plug in mechanism to support new formats.  A quick search has found a plugin for FITS in GIL.
http://ds33.ncsa.uiuc.edu/trac/browser/DMS/afw/trunk/include/lsst/afw/image/fits/fits_io.h?rev=13739

FedeMPouzols:
@NickDraper , @OwenArnold : this was an interesting ticket that I'd like to bring up and keep in mind for the discussions next weeks. 
From the imaging point of view, having support for common image formats (TIFF, PNG) well integrated in Mantid would be most convenient. I'd say it's a strong requirement for the near future. FITS is supported via `LoadFITS`.
My recent experience with FreeImage from Python is good. Sufficiently generic support for TIFF images can be challenging, but that's more a problem of the standard than the libraries.

FedeMPouzols:
GIL looks interesting for image processing. It has support for 3 formats: JPEG, TIFF and PNG, which is far less than what FreeImage provides but probably covers >95% of practical needs. But to have support for these file formats you need plugins/hpp files and linking against different libraries: libjpeg, libpng, libtiff.

A point in favor of FreeImage is that I noticed that one of our current dependencies, the `opencascade dev` package depends on `freeimage dev`. So for the CI infrastructure it would only require adding the windows libraries which always need to be added to the `ThirdParty` set anyway.

PR #16026, opened 2016-04-29
Conversion between image formats in tomography GUI

### LINKED ISSUE #11817: Documentation of the tomographic reconstruction system for IMAT
This issue was originally [TRAC 10978](http://trac.mantidproject.org/mantid/ticket/10978)

Original Reporter: @FedeMPouzols 

There are several disparate components involved and this is still work in progress, but better sooner than later there should be a piece of general documentation about the system (call it architecture, components, etc.). Include the following points:

- File formats and standards: FITS, NXtomo, the NeXus format for settings in DLS's SAVU.
- Reconstruction methods/algorithms and what you need to know to set them up (plugins, parameters, etc.).
- Autoreduction for IMAT.
- Running reduction jobs on SCARF, remote algorithm(s) to do this from Mantid.
- GUI and visualization.

 - - - - 
Keywords: IMAT

### LINKED ISSUE #15942: Conversion between formats in tomography GUI

FedeMPouzols:
New tab in the tomography GUI. This is a utility to convert between image formats.
Some input test stacks come in different formats, and sometimes users need to convert stacks in FITS or other formats to a different format before loading them into third party software for reconstruction/visualization/segmentation/etc.

This depends on: 
- [x] #16020, which should be merged first.

**To test:**
- Check the code
- New unit tests should pass
- Check that the documentation reads well.
- For a functional test, grab these files: (ISIS network) `\\olympic\babylon5\Public\fedemp\IMAT\convert_images_dummy_test_example`
- Convert the files to your favorite output format (from the list that the interface supports). For example:
  - Open the custom interface: "Diffraction" => "Tomographic Reconstruction". Tab Convert. Select "Pick Format" as "FITS. Then point the input path to the "dummy_larmor" subdirectory of the test data. Select output "Format" as "png". You can leave the maximum search depth as 3 by default. Enter an output path. See that it generates output images for the input images, reconstructing the same tree of subdirectories in the outputs. The interface will pop up a message when the conversion is finished. You can also use input path "dummy_stfc", input format "png", output format "tiff".
  - Open/close the interface and see that the settings are remembered between sessions.
  - NXTomo format is disabled as input, as we don't have a LoadNXTomo and we may not add it unless there's interest in it.
  - The interface can become unresponsive if the amount of images is large. This is unfortunately as it stands now, but this covers current needs and I wouldn't fix this with a temporary solution but rather add the Load/SaveImage algorithms, see more comments below.
- If you're an aggressive tester, you can check that if you create a directory inside the input path and use it as output path, the interface will not fall in the trap and will  not produce outputs for the output path itself.
- Still, if the input and output path are the same, the output files will be generated "in-place" next to the input files.
- You could also check that if you select an input or output path, and remove one or both of them before clicking "convert", the interface will behave well.

Fixes #15942
Also fixes #11817

(The second ticket was a long standing one about documentation, and now that all the tabs of the interface are documented I'd close it). We'll need one to add a worked out example or similar, which cannot be done at the moment.

For normal image formats (PNG, TIFF, etc.) this uses Qt classes (QImageWriter, QImageReader), which has some limitations but is the temporary solution until we have algorithms `Load/SaveImage` (#6843)

Load/Saving of FITS files is well testes in unit and system tests. NXTomo saving is also tested. Saving/loading other formats is unfortunately not well tested, as this is being done via Qt classes at the moment. The next step before we can improve on that is to have Load/SaveImage algorithms.

Release notes: https://github.com/mantidproject/mantid/commit/253929fd99b20ee2475f1f3d004b456cda08f254

---
#### Reviewer

##### Code Review
##### Functional Tests


MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Tomography/ImggFormats.h
```diff
@@ -0,0 +1,54 @@
...
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+File change history is stored at: <https://github.com/mantidproject/mantid>
+Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+namespace ImggFormats {
+enum Format { FITS = 0, TIFF = 1, PNG = 2, JPG = 3, NXTomo = 4 };
```
AntonPiccardoSelg: is an enum class possible or would you have to convert often in the GUI?

FedeMPouzols: I'd tend to prefer an enum class, but I was being lazy until we introduce the Load/SaveImage algorithms. When we know what underlying library is used for handling the image formats it will be more clear what would be more convenient. For now this is specific to the `Tomography/` sources.

docs/source/interfaces/Tomographic_Reconstruction.rst
```diff
@@ -385,23 +385,39 @@ algorithm processes the input directories and files.
...
+process can be lengthy and demanding in terms of disk space when
+processing more than one or a small number of experiments (RB
+reference numbers), and especially so for wavelength dependent
+experiments.
+
 Example
 -------
 
-TODO: there should be an example using a small data set.
+TODO: there should be a worked out example using a small data set.
```
AntonPiccardoSelg: Is this for another PR?

Framework/DataHandling/src/LoadFITS.cpp
```diff
@@ -32,9 +32,29 @@ template <typename InterpretType> double toDouble(uint8_t *src) {
...
+  double timeBin;
+  double scale;
+  std::string imageKey;
+  long int countsInImage;
+  long int numberOfTriggers;
+  std::string extension;
+  std::string filePath;
+  bool isFloat;
+};
+
```
AntonPiccardoSelg: good move into the `.cpp`

docs/source/interfaces/Tomographic_Reconstruction.rst
FedeMPouzols: Yes, definitely. There's several `TODO`s here that need further work as soon as we get data, and have a bit more experience with some of the tools so we can fill this in.

Framework/DataHandling/src/LoadFITS.cpp
```diff
@@ -150,170 +170,175 @@ void LoadFITS::exec() {
...
+        e.what());
+  }
+
+  // Get and convert specific MANDATORY standard header values which
+  // will are needed to know how to load the data: BITPIX, NAXIS,
+  // NAXISi (where i = 1..NAXIS, e.g. NAXIS2 for two axis).
+  try {
+    std::string tmpBitPix = header.headerKeys[m_headerBitDepthKey];
+    if (boost::contains(tmpBitPix, "-")) {
+      boost::erase_all(tmpBitPix, "-");
```
AntonPiccardoSelg: could use `std::find` and `std::replace` but this is actually easier to read :+1:

Framework/DataHandling/src/LoadFITS.cpp
```diff
@@ -150,170 +170,175 @@ void LoadFITS::exec() {
...
-            headers[i].headerKeys[m_headerAxisNameKeys[j]]));
-        // only debug level, when loading multiple files this is very verbose
-        g_log.debug() << "Found axis length header entry: "
-                      << m_headerAxisNameKeys[j] << " = "
-                      << headers[i].axisPixelLengths.back() << std::endl;
-      }
+    // Various extensions to the FITS format are used elsewhere, and
+    // must be parsed differently if used. This loader Loader
+    // doesn't support this.
+    header.extension = header.headerKeys["XTENSION"];
```
AntonPiccardoSelg: should `XTENSION` be defined somewhere as a `static const`?

Framework/DataHandling/src/LoadFITS.cpp
```diff
@@ -150,170 +170,175 @@ void LoadFITS::exec() {
...
+      header.offset =
+          boost::lexical_cast<int>(header.headerKeys[m_headerOffsetKey]);
+    } catch (std::exception & /*e*/) {
+      // still, second try with floating point format (as used for example
+      // by
+      // Starlight XPRESS cameras)
       try {
-        headers[i].scale = boost::lexical_cast<double>(
-            headers[i].headerKeys[m_headerScaleKey]);
+        double doff =
```
AntonPiccardoSelg: could use `auto`

Framework/DataHandling/src/SaveFITS.cpp
```diff
@@ -0,0 +1,252 @@
...
+      file.write(reinterpret_cast<const char *>(bytesPixel.data()), bytespp);
+    }
+  }
+}
+
+void SaveFITS::writeFITSHeaderEntry(const std::string &hdr,
+                                    std::ofstream &file) {
+  static const std::vector<char> blanks(g_maxLenHdr, 32);
+
+  size_t count = hdr.size();
```
AntonPiccardoSelg: `auto`

MantidQt/CustomInterfaces/src/Tomography/ImggFormatsConvertPresenter.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+  // What should be alg->getProperty("OutputWorkspace"); but this is a group
+  Mantid::API::WorkspaceGroup_sptr grp =
+      Mantid::API::AnalysisDataService::Instance()
+          .retrieveWS<Mantid::API::WorkspaceGroup>(wksGrpName);
+  Mantid::API::MatrixWorkspace_sptr imgWorkspace =
+      boost::dynamic_pointer_cast<Mantid::API::MatrixWorkspace>(
+          grp->getItem(grp->size() - 1));
+
+  return imgWorkspace;
```
AntonPiccardoSelg: Why do you go through the ADS here? Is the workspace used for something else?

MantidQt/CustomInterfaces/src/Tomography/ImggFormatsConvertViewQtWidget.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+  writeImgFile(img, outputName, outputFormat);
+}
+
+void ImggFormatsConvertViewQtWidget::writeImg(
+    MatrixWorkspace_sptr inWks, const std::string &outputName,
+    const std::string &outFormat) const {
+  if (!inWks)
+    return;
+  size_t width = inWks->getNumberHistograms();
```
AntonPiccardoSelg: `auto`, also below

AntonPiccardoSelg:
Code review:
 * Looks good. Comments are minor and don't warrant a change now.
 * Good unit tests are in place
 * Very extensive documentation is provided, making it clear how the scientists can use the converter.

MantidQt/CustomInterfaces/src/Tomography/ImggFormatsConvertPresenter.cpp
FedeMPouzols: Now that you mention this I'm going to double check. I think here it should be possible to do it the neat way with `getProperty()`.

AntonPiccardoSelg:
Funcitonal testing:
 * No pop up appears when the conversion has completed.
 * Converted from FITS to TIFF without any issues
 * Converted from FITS to JPEG --> Mantid terminates (Unexpected exception: map::at)
 * Converted from FITS to PNG without any issues
 * Converted from FITS to NXTOMO --> Mantid terminates (Unexpected exception: All arguments to init must be positive and non-zero)
   
The output files end with `.fits.tiff` and `fits.png`. Is this on purpose?

FedeMPouzols:
@AntonPiccardoSelg :  great that you caught these issues in time.
- The pop-up should show up now, I must have downgraded to a log message without noticing. This is now logging the message and also popping up.
- Saving to JPEG and NXTomo should be working well now. The JPEG issue was a silly mistake after an update in the name of the format. The NXTomo issue uncovered a bug in SaveNXTomo algorithm for which I've added a regression test.
- I've added better `try/catch` and more specific error messages for the scenarios where the load/save could fail. That should prevent those panic messages.
- And yes, the original extension is kept as it is useful information in practice. For example the images that were .fits original are most likely coming from the instrument cameras. As not all the image formats support headers this is useful in practice so it is better not to remove the old extension.

AntonPiccardoSelg:
Functional Testing:
* Converted from FITS to TIFF --> works well ( I assume that is only for compression)
* Converted from FITS to TIFF --> works well
* Converted from FITS to JPEG --> works well
* Converted from FITS to PNG --> works well
* Converted from FITS to NXTOMO --> works well

* Converted from PNG to PNG --> works well
* Converted from PNG to TIFF --> works well
* Converted from PNG to JPEG --> works well

* Converted from TIFF to PNG --> works well
* Converted from TIFF to TIFF --> works well
* Converted from TIFF to JPEG --> works well

* Converted from JPEG to PNG --> works well
* Converted from JPEG to TIFF --> works well
* Converted from JPEG to JPEG --> works well

* Converting from TIFF/PNG/JPEG to NXTOMO/FITS produces an error pop up (since this conversion is not well defined). It might make sense to disable it in the future.

Happy to :shipit:

PR #16031, opened 2016-04-29
MaxEnt: add property to correct for phase shift

### LINKED ISSUE #15280: MaxEnt: deal with phase when input data do not start at t=0
Have a look at the "Shift" property in `FFT`, we should do something similar here. `MaxEnt` should calculate this shift itself, if possible, so users don't not have to provide it as an input.

@tom-perkins I am assigning this to you, as you've been involved in the discussion with James about `FFT`. At present, we have a first version of `MaxEnt` and the current code is a bit untidy, needs refactoring etc, but I think this can be easily added at the end of the algorithm. Ask me any questions you have.

raquelalvarezbanos:
Added a new property, _AutoShift_, that automatically corrects for the phase shift when data do not start at x=0.

**To test:**

The following script:

```
from math import pi, sin, cos
X = []
YRe = []
YIm = []
E = []
N = 1000
w = 10
for i in range(0,N):
       x = 2*pi*i/N
       X.append(x)
       YRe.append(cos(w*x))
       YIm.append(sin(w*x))
       E.append(0.1)
ws = CreateWorkspace(DataX=X+X,DataY=YRe+YIm,DataE=E+E,NSpec=2)

_, _, image1, _ = MaxEnt(ws, ComplexData=True, A=1e-5, AutoShift=True)

factor = pi/w
ws = ScaleX(InputWorkspace='ws', Factor=factor, Operation='Add')
_, _, image2, _ = MaxEnt(ws, ComplexData=True, A=1e-5, AutoShift=True)
```

should produce two images, `image1` and `image2`, that have a phase difference (first spectrum corresponds to the real part of the Fourier transform, second corresponds to imaginary part, which should be close to 0).

Fixes #15280.
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Algorithms/inc/MantidAlgorithms/MaxEnt.h
+++ b/Framework/Algorithms/inc/MantidAlgorithms/MaxEnt.h
@@ -75,10 +75,14 @@ private:
                                            const QuadraticCoefficients &coeffs,
                                            const std::vector<double> &image,
                                            double background, double distEps);
-  /// Populates the output workspaces
-  void populateOutputWS(const API::MatrixWorkspace_sptr &inWS, size_t spec,
-                        size_t nspec, const std::vector<double> &result,
-                        API::MatrixWorkspace_sptr &outWS, bool isImage);
+  /// Populates the output workspace containing the reconstructed data
+  void populateDataWS(const API::MatrixWorkspace_sptr &inWS, size_t spec,
+                      size_t nspec, const std::vector<double> &result,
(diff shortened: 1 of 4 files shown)


tom-perkins:
Ran script as above, images have a phase difference as required. Also compared output (without *AutoShift*) to previous version - matches well.
Code changes look good and a test has been added to cover them. Release notes have been updated. :shipit:

Issue #16029, opened 2016-04-29
Convert a Python `dict` to a C++ PropertyManager


martyngigg:
#16028 introduces the `PropertyManagerProperty` type. In Python this is best represented as `dict` so 
when a `dict` is passed to a `PropertyManager` in C++ it should be converted to a new `PropertyManager` to be stored inside a `PropertyManagerProperty`.

PR #16034, opened 2016-05-02
Add a PropertyManagerProperty type

### LINKED ISSUE #16028: Introduce a PropertyManager property type
This is to support more flexible mapping type as an input property to an algorithm. The current support for passing a `PropertyManager` through to an algorithm is to store it in the `PropertyManagerDataService` and pass a string name in to the algorithm.

Introducing a property type would allow an anonymous object to be passed through the property mechanism.

martyngigg:
Adds a `PropertyManagerProperty` to allow a `PropertyManager` as a value in another `PropertyManager` without having to create a name and put it in the `PropertyManagerDataService`

**To test:**

There is limited external testing that can be done with this. Review the code and check you agree with how the property type is handled. The ultimate goal is to allow automatic conversion of a Python `dict` to a `PropertyManager` on input to an algorithm.

Fixes #16028

---
#### Reviewer

##### Code Review
- [ ] If there are changes in the release notes then do they describe the changes appropriately? -> N/A
##### Functional Tests
- [ ] Do changes function as described? Add comments below that describe the tests performed? ->N/A
- [ ] How do the changes handle unexpected situations, e.g. bad input? ->N/A
- [ ] Has the relevant documentation been added/updated? ->N/A
- [ ] Is user-facing documentation written in a user-friendly manner? ->N/A
- [ ] Has developer documentation been updated if required? ->N/A


peterfpeterson:
How would this change things like [PDDetermineCharacterizations](https://github.com/mantidproject/mantid/blob/master/Framework/Algorithms/src/PDDetermineCharacterizations.cpp)?

martyngigg:
@peterfpeterson If `ReductionProperties` was made a `PropertyManagerProperty` then in `exec` you could do `m_propertyManager = getProperty("ReductionProperties")`.

Once #16029 is complete I envisaged being able to do this in Python:

```python
PDDetermineCharacterizations(..., ReductionProperties={'frequency':10.0,'wavelength':5})
```
by converting the `dict` to a `PropertyManager` directly rather than having to create one, put it in the data service, give it a name and then retrieve it later. The current way of having the "global" manager would still be allowed though.

Framework/Kernel/test/PropertyManagerPropertyTest.h
```diff
@@ -0,0 +1,163 @@
...
+    using Mantid::Kernel::PropertyManager;
+    using Mantid::Kernel::PropertyManager_sptr;
+    // Create a top-level PropertyManager, add a PropertyManagerProperty
+    // as a nested PropertyManager
+    auto topMgr = boost::make_shared<PropertyManager>();
+    topMgr->declareProperty(
+        Mantid::Kernel::make_unique<PropertyManagerProperty>("Args"));
+    topMgr->setProperty("Args", createPropMgrWithInt());
+
+    PropertyManager_sptr args = topMgr->getProperty("Args");
```
AntonPiccardoSelg: Did you want to assert here?

martyngigg: You're right, that probably should assert something. It will throw so the test will fail but it's not obvious to a developer from reading it. I'll update it.

Framework/Kernel/test/PropertyManagerPropertyTest.h
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  void test_Empty_Property_Set_With_Json_String_Returns_Error() {
+    PropertyManagerProperty prop("Test");
+
+    auto json = createPropMgrWithInt()->asString(true);
+    std::string msg;
+    TS_ASSERT_THROWS_NOTHING(msg = prop.setValue(json));
+    TS_ASSERT(msg.length() > 0);
+  }
```
AntonPiccardoSelg: What is the reason for not allowing this scenario?

martyngigg: This is unfortunate, I was hoping this would be simpler to achieve but after I thought a little the implementation was going to be quite convoluted for little gain.

The problem with having an empty manager and it only being passed a string is that I have no idea what types all of the individual elements should be so it would have to be inferred which is difficult and error prone.

I see this being most useful from Python when translating an existing `dict` and in this case we will then know the types of the elements and can translate this properly. This is done in #16029.

AntonPiccardoSelg: Yep, makes sense!

AntonPiccardoSelg:
Performed mainly code review: 
- Changes look good to me. 
- All possible scenarios seem to be unit tested. 

Regarding the assertion in the Unit tests: It should be sufficient to change this in #16029

Functional testing was not performed. This will make more sense when the Python Interface is fully exposed.

Happy to :shipit:

martyngigg:
Thank you!

PR #16036, opened 2016-05-02
HistogramData module (X data only)

### LINKED ISSUE #16086: Phase 1 of implementing a Histogram type
See design document for details: https://github.com/mantidproject/documents/blob/histogram_type/Design/histogram_type.md.

The first step is the introduction of new types for x-data.

SimonHeybrock:
**Do not merge this PR, this is still in an early review stage.**

**Edit: Work has progressed, please see https://github.com/mantidproject/mantid/pull/16627 instead of this PR.**

Note that the documentation is still quite incomplete. I will work on this during the review process.

There is a `rst` file in `concepts` that tries to explain some details, it might be worth having a look at it before the review.

**Code review:**
This is a very large and critical change, I would like to suggest to split the review into several parts:
1. Review everything in the new module `HistogramData`. **<- current phase**
2. Feedback to me, I will then review the full set of changes myself.
3. Review of everything else (existing Mantid code).

I think we should have more than one reviewer for both 1) and 3).

**To test:**

This is a global change, there is no individual test. If you have access to any Python scripts using Mantid that are not part of our automated testing it would be good to check if they still work.

Fixes #16086.

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/HistogramData/inc/MantidHistogramData/FixedLengthVector.h
```diff
@@ -0,0 +1,124 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+template <class T> class FixedLengthVector {
```
quantumsteve: is class `T` ever used in `FixedLengthVector`?

quantumsteve:
Are `FixedLengthVector` and `VMDBase` similar enough that they should be a single class?

Framework/HistogramData/test/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+if ( CXXTEST_FOUND )
+  include_directories ( SYSTEM ${CXXTEST_INCLUDE_DIR} ${GMOCK_INCLUDE_DIR} ${GTEST_INCLUDE_DIR} )
+
+  cxxtest_add_test ( HistogramDataTest ${TEST_FILES} ${GMOCK_TEST_FILES})
+  target_link_libraries( HistogramDataTest LINK_PRIVATE ${TCMALLOC_LIBRARIES_LINKTIME}
+    HistogramData
+    Kernel
+    ${Boost_LIBRARIES}
+    ${POCO_LIBRARIES}
```
quantumsteve: Does `HistogramDataTest` use Poco?

Framework/HistogramData/inc/MantidHistogramData/FixedLengthVector.h
SimonHeybrock: `FixedLengthVector` is implemented using the curiously recurring template pattern (CRTP). `T` is the type of the derived type. It is not directly used in `FixedLengthVector`, but it makes derived types completely distinct. For example, we will later introduce `HistogramY`, inheriting from `FixedLengthVector<HistogramY>`. As a consequence, mixing `HistogramX` and `HistogramY` is simply impossible.

I think this more or less answers also your question regarding `VMDBase`. Furthermore `VMDBase` seems to be specificially for short vectors, whereas `FixedLengthVector` is for arbitrary length.
Additionally, I am using the CRTP to have something like "mixins". For example, `class Scalable` can be used to provide scaling for any class inheriting from `FixedLengthVector`. `Scalable` works only if there are iterators available. Later I want to add `class Addable`, to make `FixedLengthVectors` addable. However, `HistogramX` is not `Addable` (adding bin edges would not make sense), only `HistogramY` is. That is, the `operator+` cannot be implemented in the base type. Similarely other operations in `VMDBase` are not meaningful for `HistogramX` or `HistogramY`.

So, I don't think it makes sense to merge `VMDBase` and `FixedLengthVector`. Rather, `VMDBase` might be a candidate where the `Eigen` library would be useful. @MichaelWedel had a look at this recently.

PR #16050, opened 2016-05-03
GetEi version 1 does not work anymore

### LINKED ISSUE #16049: GetEi version 1 does not work any more.
### If one tries to run GetEi version 1, it fails, e.g: 
`
Load(Filename='MER30514.nxs', OutputWorkspace='MER30514', LoadMonitors=True, SpectrumMin=69631, SpectrumMax=69640)`
`
 GetEi('MER30514_monitors',69634,69638,100,Version=1)
`
should return Ei and  first peak time
### Actually:   it throws the error:
"Attempt to assign to property (StartWorkspaceIndex) of incorrect type"

### Tried on windows and RHEL7

abuts:
Description of work.

**To test:**
See ticket description -- GetEi version 1 fails before the fix and should run fine after. 
Changes are small and obvious so code review should be sufficient. 

Fixes #16049.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Algorithms/src/GetEi.cpp
+++ b/Framework/Algorithms/src/GetEi.cpp
@@ -7,6 +7,7 @@
 #include "MantidKernel/CompositeValidator.h"
 #include "MantidKernel/BoundedValidator.h"
 #include "MantidKernel/PhysicalConstants.h"
+#include "MantidKernel/Exception.h"
 
 #include <boost/lexical_cast.hpp>
 #include <cmath>
@@ -81,7 +82,7 @@ void GetEi::init() {
 * does not have common binning
 */
 void GetEi::exec() {
-  MatrixWorkspace_const_sptr inWS = getProperty("InputWorkspace");
+  MatrixWorkspace_sptr inWS = getProperty("InputWorkspace");
   const specnum_t mon1Spec = getProperty("Monitor1Spec");
   const specnum_t mon2Spec = getProperty("Monitor2Spec");
   double dist2moni0 = -1, dist2moni1 = -1;
@@ -139,6 +140,11 @@ void GetEi::exec() {
                  << " (your estimate was " << E_est << " meV)\n";
 
   setProperty("IncidentEnergy", E_i);
+  // store property in input workspace
+  Property *incident_energy =
+      new PropertyWithValue<double>("Ei", E_i, Direction::Input);
+  inWS->mutableRun().addProperty(incident_energy, true);
+
 }
 /** Gets the distances between the source and detectors whose IDs you pass to it
 *  @param WS :: the input workspace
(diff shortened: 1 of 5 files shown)


Framework/Algorithms/src/GetEi.cpp
```diff
@@ -325,10 +325,17 @@ void GetEi::extractSpec(int64_t wsInd, double start, double end) {
 
   childAlg->setProperty<MatrixWorkspace_sptr>("InputWorkspace",
                                               getProperty("InputWorkspace"));
+
   childAlg->setProperty("XMin", start);
   childAlg->setProperty("XMax", end);
-  childAlg->setProperty("StartWorkspaceIndex", wsInd);
-  childAlg->setProperty("EndWorkspaceIndex", wsInd);
+  if (wsInd < std::numeric_limits<int>::max()) {
```
quantumsteve: I don't understand why this function's signature takes an `int64_t` when the child algorithm needs and `int`. It looks like version 2 has the same issue but hides it with a `static_cast<int>`. 

Would changing `GetEi::extractSpec` so that `wsInd` is an `int` or changing `CropWorkspace` so that `StartWorkspaceIndex` and `EndWorkspaceIndex` are `int64_t` be better solutions? The former advertises a limitation, while the later may introduce other bugs but makes it work as a naive developer would assume.

quantumsteve:
Would you add a unit test for version 1 so that issues like these are more likely to be caught?

abuts:
Hmm.

abuts:
I do not know why its int_64 and have not looked at GetEi v2. 
The thing I did is minimal modification I needed to sort issue with experiment. 

I would better talk to Martyn who wrote both algorithms and discuss why this issue ever appear. Its possible it left from the time when we may have 2-bytes int on 32 bit system, but as these times have long gone, so we may change the GetEi V1. I do not think we will have instruments with ~4*10^9 detectors in forcible future, so no point in changing CropWorkspace. 

I will add unit test for version 1 if you think its needed.

quantumsteve:
If this version's worth fixing, then it's worth having a test to ensure the same issue doesn't happen again. If GetEi v1 hasn't been used in years, it should also inherit from `Mantid::API::DeprecatedAlgorithm`.

abuts:
Yes, it was not used but not I was actually needed it after all, so let it exist.

quantumsteve:
The change look good. Should this be in the release notes?

abuts:
I do not think release notes are necessary, as first its bugfix so people would expect it to work anyway, plus GetEi v1 can not be currently used in SNS reduction + reduction code needs changing if one wants to use it in ISIS reduction. GetEi v1 has very narrow area of applicability (like this time, when I was trying to identify why GetEi v1 returns incorrect results and appears that its sample problem) 

Let GetEi v1 to stay for emergencies and comparison reasons but its area of applicability belongs to advanced usage in very specific circumstances so only very special people in very special circumstances should care about it.

quantumsteve:
:shipit:

Issue #15856, opened 2016-04-07
Performance test LoadEventNexusTestPerformance affected by StructuredDetector


OwenArnold:
Looking at the performance test, I can see a significant jump in the execution time for these tests following the introduction of #15744 

![screen shot 2016-04-07 at 09 24 52](https://cloud.githubusercontent.com/assets/1109536/14345119/3ed148ca-fca3-11e5-9b1f-a64af740bb69.png)

**Also Affected**
x3 slowdown in DataHandlingTest.LoadEventPreNexusTestPerformance.testDefaultLoad

![screen shot 2016-04-07 at 09 32 09](https://cloud.githubusercontent.com/assets/1109536/14345238/c7d96cd8-fca3-11e5-8569-32650187974c.png)

### Steps to reproduce the behaviour

You should run the performance tests yourself before and after merge commit dc9168cc48abd3f5689c8371c0d7949dba674438 to see the timings for yourself and then identify and fix the issue.

### Platforms affected

Likely all platforms

OwenArnold:
@mantid-roman, @LamarMoore  is away for a few weeks, do you have time to investigate this before his return?

OwenArnold:
This particular performance test might be pointing more towards the root cause of the problem.
![screen shot 2016-04-07 at 09 43 23](https://cloud.githubusercontent.com/assets/1109536/14345536/50158cd4-fca5-11e5-9612-b735a0dbc8c7.png)

OwenArnold:
@MikeHart85 Could you please build between these change-sets on Ubuntu and see if you get the 2* speed change the performance tests indicate. If so, could you run callgrind and determine where the slow down might be so that Lamar can fix it.

MikeHart85:
I had the following results (in secs) when comparing commits `12b9eb521167b99810c127f3d688794867693e47` (before) and `dc9168cc48abd3f5689c8371c0d7949dba674438` (after):

Test | Class | Function | Before | After
----- | -------- | ----------- | --------- | ------
DataHandlingTest | LoadEventNexusTestPerformance | testDefaultLoad | 0.67 | 1.06
DataHandlingTest | LoadEventNexusTestPerformance | testDefaultLoadBankSplitting | 0.73 | 0.36
DataHandlingTest | LoadEventPreNexusTestPerformance | testDefaultLoad | 0.25 | 0.90
GeometryTest | QuadrilateralTestPerformance | test_Area_Calls | 3.92 | 6.18

[...]

PR #16062, opened 2016-05-04
Factor out constructor from test_Area_Calls

quantumsteve:
Description of work.

I noticed in #15856 that in `GeometryTest.QuadrilateralTestPerformance.test_Area_Calls` much of the time is spent constructing the Quadrilateral object. This constructor can be taken out of the loop so that most of the time is spent within the area() function. Clang does not optimize the loop away.

I also removed a copy constructor and copy assignment operator that is not different from the compiler-generated default and changed a few places to use `std::min` and `std::max` instead of the more confusing ternary conditional operator.

Unfortunately, this does not address the performance drop at commit dc9168c.   

**To test:**

`QuadrilateralTestPerformance` isn't one of unit tests run on the build servers and should be built and run locally.

This is a small change with no issue number

---
#### Reviewer

##### Code Review
##### Functional Tests


peterfpeterson:
Reviewed the code and builds pass. :shipit:

Issue #16057, opened 2016-05-04
Use Scoped Enum for Geometry types


LamarMoore:
Currently a weak enum is used to describe GeometryTypes.

```c++
enum GEOMETRY_TYPE {
    CUBOID,            ///< CUBOID
    HEXAHEDRON,        ///< HEXAHEDRON
    SPHERE,            ///< SPHERE
    CYLINDER,          ///< CYLINDER
    CONE,              ///< CONE
    SEGMENTED_CYLINDER ///< Cylinder with 1 or more segments (along the axis).
    /// Sizes of segments are important.
  };
```

this should be replaced with a more robust scoped enum e.g

```c++
enum class GEOMETRY_TYPE {
    CUBOID,            ///< CUBOID
    HEXAHEDRON,        ///< HEXAHEDRON
    SPHERE,            ///< SPHERE
    CYLINDER,          ///< CYLINDER
    CONE,              ///< CONE
    SEGMENTED_CYLINDER ///< Cylinder with 1 or more segments (along the axis).
    /// Sizes of segments are important.
  };
```

PR #16075, opened 2016-05-05
Cuboid/Hexahedron bug fix

### LINKED ISSUE #15999: hexahedron incorrectly treated as a cuboid
I found several places where hexahedron is being treated as a cuboid and giving incorrect results. The occurs because cuboid and hexadron return the same type in [`GluGeometryHandler::GetObjectGeom(...)`](https://github.com/mantidproject/mantid/blob/master/Framework/Geometry/src/Rendering/GluGeometryHandler.cpp#L92)

There are at least two places in Object.cpp (
[here](https://github.com/mantidproject/mantid/blob/master/Framework/Geometry/src/Objects/Object.cpp#L1019) and [here](https://github.com/mantidproject/mantid/blob/master/Framework/Geometry/src/Objects/Object.cpp#L1605)) that call [Object::SolidAngleHelper]( 
https://github.com/mantidproject/mantid/blob/master/Framework/Geometry/src/Objects/Object.cpp#L1158), which uses only four of the eight points.

Hexahedron given the type "box", and therefore doesn't throw an exception.
https://github.com/mantidproject/mantid/blob/master/Framework/Geometry/src/Instrument/ObjCompAssembly.cpp#L385

LamarMoore:
The GluGeometry handler no longer returns the same type for cuboids and hexahedra when the geometry is queried. SolidAngles for Hexahedra and Cuboids are now handled separately. A hexahedron special case will be developed as a part of another issue as this functionality is not urgent. Another issue to address the weakly typed enums for the glugeometryhandler/object can be found here #16057 

Code Review

Fixes #15999.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Geometry/inc/MantidGeometry/Rendering/GluGeometryHandler.h
+++ b/Framework/Geometry/inc/MantidGeometry/Rendering/GluGeometryHandler.h
@@ -48,7 +48,8 @@ class Object;
 class MANTID_GEOMETRY_DLL GluGeometryHandler : public GeometryHandler {
   /// the type of the geometry eg CUBOID,CYLINDER,CONE,SPHERE
   enum GEOMETRY_TYPE {
-    CUBOID = 1,        ///< CUBOID
+    NOSHAPE = 0,
+	CUBOID,        ///< CUBOID
     HEXAHEDRON,        ///< HEXAHEDRON
     SPHERE,            ///< SPHERE
     CYLINDER,          ///< CYLINDER
--- a/Framework/Geometry/inc/MantidGeometry/Rendering/GluGeometryHandler.h
+++ b/Framework/Geometry/inc/MantidGeometry/Rendering/GluGeometryHandler.h
@@ -48,7 +48,7 @@ class Object;
 class MANTID_GEOMETRY_DLL GluGeometryHandler : public GeometryHandler {
   /// the type of the geometry eg CUBOID,CYLINDER,CONE,SPHERE
   enum GEOMETRY_TYPE {
-    CUBOID,            ///< CUBOID
+    CUBOID = 1,        ///< CUBOID
     HEXAHEDRON,        ///< HEXAHEDRON
     SPHERE,            ///< SPHERE
     CYLINDER,          ///< CYLINDER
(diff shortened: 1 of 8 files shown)


quantumsteve:
At Framework/Geometry/src/Objects/Object.cpp#L1029, hexadrons are being treated as spheres.

Framework/Geometry/src/Rendering/GluGeometryHandler.cpp
```diff
@@ -95,9 +95,6 @@ void GluGeometryHandler::GetObjectGeom(int &mytype,
     vectors = m_points;
     switch (type) {
     case CUBOID:
-      mytype = 1;
-      break;
-    case HEXAHEDRON:
```
quantumsteve: Since `HEXAHEDRON` doesn't have a height or radius, should it use the default?

Framework/Geometry/inc/MantidGeometry/Rendering/GluGeometryHandler.h
```diff
@@ -48,7 +48,7 @@ class Object;
 class MANTID_GEOMETRY_DLL GluGeometryHandler : public GeometryHandler {
   /// the type of the geometry eg CUBOID,CYLINDER,CONE,SPHERE
   enum GEOMETRY_TYPE {
-    CUBOID,            ///< CUBOID
+    CUBOID = 1,        ///< CUBOID
     HEXAHEDRON,        ///< HEXAHEDRON
```
quantumsteve: Should we also have `NOSHAPE=0`?

Framework/Geometry/src/Rendering/GluGeometryHandler.cpp
LamarMoore: L1209 in Object.cpp is using ```type == 3``` however, hexahedron is 2. Could you explain in a little more detail?

LamarMoore:
I noticed there was no bounding box calculation for hexahedral types now I have made these changes. I created an issue [here](https://github.com/mantidproject/mantid/issues/16106) which depends on this merge.

quantumsteve:
The issues I found earlier have been addressed satisfactorily. Good to see issues created for further improvements to these classes.

:shipit:

PR #16076, opened 2016-05-05
Fix performance regression in ConvexPolygonTestPerformance 

quantumsteve:
Description of work.

This addresses the `ConvexPolygonlTestPerformance` and `QuadrilateralTestPerformance` regressions discussed in #15856. with GCC on linux, switching Vertices from a `std::deque` to a `std::vector` eliminates calls to `malloc_consolidate` and improves performance in the old test by about an 10x. 

I split up the performance tests into two, one that updates the points in a `ConvexPolygon` and another that tests calculating the area of the polyhedron. 

**To test:**

This PR is best verified on Linux since the performance regression occurs there. Since I modified the performance tests, use the test currently in master to verify that the performance regression is fixed. 

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/Geometry/test/QuadrilateralTest.h
```diff
@@ -152,10 +152,16 @@ class QuadrilateralTest : public CxxTest::TestSuite {
 //------------------------------------------------------------------------------
 class QuadrilateralTestPerformance : public CxxTest::TestSuite {
 public:
-  void test_Area_Calls() {
-    const size_t ntests(50000000);
+  void xtest_update_values() {
```
MikeHart85: Is this meant to be disabled (doesn't run since it starts with `xtest_` instead of `test_`)?

I'm guessing it was meant to be removed? Or updated?

quantumsteve: Thank you for catching that. It should now be enabled.

MikeHart85:
Tested on Linux. 

Significant improvement to both the `QuadrilateralTestPerformance` and the `ConvexPolygonTestPerformance` tests. Using the old tests, runtime on my machine went from 6.18 sec to 0.57 sec and from 9.5 sec to 7.4 sec, respectively, as compared to #15856. The new tests split testing construction performance from area calculation performance, and `ConvexPolygonTestPerformance` runs much faster since it uses fewer iterations.

:shipit:

PR #15744, opened 2016-03-23
Structured Detector Implementation

LamarMoore:
This represents the implementation of the StructuredDetector for irregular geometries #14965. 
### Testing

Code Review and run the LOKI IDF (you will need to change the file extension to .xml):
- Launch MantidPlot
- In Algorithms, type LoadEmptyInstrument then click execute
- Browse to the location of the LOKI IDF
- Select a workspace name e.g. "LOKI_Ws"
- Once the workspace has loaded, right click and select Show Instrument
- Ensure all views are correct and the instrument view works as expected with the new type of geometry.
#### LOKI IDF using structured detector

See attached
[LOKI_definition.txt](https://github.com/mantidproject/mantid/files/186328/LOKI_definition.txt)

Fixes #15542 .

---
#### Reviewer

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):
##### Code Review
- [x] Is the code of an acceptable quality?
- [x] Does the code conform to the coding standards? Is it well structured with small focussed classes/methods/functions?
- [x] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?
##### Functional Tests
- [x] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
- [x] Has the relevant documentation been added/updated?
- [x] Is user-facing documentation written in a user-friendly manner?
- [x] Has developer documentation been updated if required?
- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

PR #16083, opened 2016-05-06
Fix performance regressions caused by regex matches 

### LINKED ISSUE #15856: Performance test LoadEventNexusTestPerformance affected by StructuredDetector

MikeHart85:
This addresses #15856 by optimizing the regex matches introduced in #15744.

Originally, regex matches introduced a significant slowdown. 

Reverting to string comparisons was explored as an option, but ultimately it turned out that the regex code could easily be optimized to nearly the same performance levels. 

Just making the regex expression definition `static` had a profound effect. Should be considered for other usages as well, where the expression is constant.

**To test:**

Run `DataHandlingTest LoadEventNexusTestPerformance` and `DataHandlingTest LoadEventPreNexusTestPerformance` and observe differences before and after this PR. 

It would be best to compare performance against dc9168c, to avoid other changes influencing test results.

Fixes #15856.

_Does not need to be in the release notes. (Addresses regression that was never released)_

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Geometry/src/Instrument/InstrumentDefinitionParser.cpp
+++ b/Framework/Geometry/src/Instrument/InstrumentDefinitionParser.cpp
@@ -1563,7 +1563,7 @@ void InstrumentDefinitionParser::appendLeaf(Geometry::ICompAssembly *parent,
   if (pType->hasAttribute("is"))
     category = pType->getAttribute("is");
 
-  static boost::regex exp("Detector|detector|Monitor|monitor");
+  static const boost::regex exp("Detector|detector|Monitor|monitor");
 
   // do stuff a bit differently depending on which category the type belong to
   if (RectangularDetector::compareName(category)) {
--- a/Framework/Geometry/src/Instrument/InstrumentDefinitionParser.cpp
+++ b/Framework/Geometry/src/Instrument/InstrumentDefinitionParser.cpp
@@ -30,7 +30,6 @@
 #include <Poco/SAX/AttributesImpl.h>
 
 #include <boost/make_shared.hpp>
-#include <boost/regex.hpp>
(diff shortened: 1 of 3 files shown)


MikeHart85:
After testing an alternative approach, I've switched back to using regex again instead of string comparisons.

Simply adding a `static` keyword in front of the `boost::regex exp` definition improved performance by almost the same amount as using string comparisons, since this way the expression is only parsed once. Removing the capture groups around individual terms shaved it down a little more, bringing it even closer.

Overall, the regex matches are still a tiny bit slower than string comparisons, but the difference is so tiny that it should be negligible. Thus regex is preferred as being a bit cleaner and more flexible.

Updating PR description to match new approach.

MikeHart85:
It looks like Lamar checked everything but hasn't tagged it for shipping yet.

Since he's off sick and a code freeze is imminent, I'll go ahead and tag it.

:shipit:

OwenArnold:
@LamarMoore I've asked Michael to add the :shipit: since I don't know why you would approve the checklist and not raise a comment about any further problems. I therefore assume that you forgot to add :shipit:

OwenArnold:
Looks good to me.

PR #16114, opened 2016-05-09
Fix Coverity issues #1355157 and #1355158. 

quantumsteve:
Description of work.

This fixes Coverity issues #1355157 and #1355158, which found that some of our ParaView plugins were returning a `const char *` and destroying the string that it points to. The easiest fix is to return a `std::string`. I made the same change for the other plugins as well as a little code cleanup.

I'm not sure why changing the return type from `char *` to `const char *` triggered these warning, but I fear both functions were only working because of undefined behavior.

**To test:**

Changes should be clear from code review.

There is no issue associated with this pull request.

---
#### Reviewer

##### Code Review
##### Functional Tests


FedeMPouzols:
Code changes look good. I did a quick check and the vsi behaves well with histo, event and peaks data.
I'd also guess that we were just lucky with "undefined behavior" here, but this could definitely cause random crashes it seems.

:shipit:

PR #16122, opened 2016-05-10
Updates to Fitting concept page

### LINKED ISSUE #15901: Updates to Fitting concept page
Change http://docs.mantidproject.org/nightly/concepts/Fitting.html according to Nick Gould's (from UK SCD Numerical group) comments below.

> This looks very good. A few suggestions -
> 
> Instead of writing y_i, you might write y(x_i) where y(x) is your model (otherwise, the point of introducing x might be lost on the reader).
> 
> sigma_i is often thought of as the variance of the i-th data value.
> 
> I would put a comma so that
> 
>  the scientific problem in question or say a peak search
> 
> becomes
> 
>    the scientific problem in question, or say a peak search
> 
> as at first I didn't understand this sentence.
> 
> There is some confusion about what "minimizer" means. In some places it is the "consequence of minimizing", while in others it means "the minimization algorithm". The paragraph
> 
> "Local minimizers can be grouped into ..."
> 
> was confusing because of this; it should say "Local minimization algorithms ..."
> 
> I am not sure I know of any algorithm that can find the "nearest" local minimizer, since this is itself a global optimization problem. A strict downhill minimizer will find a 
> lower value, but that is all I think you can say (actually, for technical reasons, no method can even ensure that it finds a minimizer, but it will find a critical point of the cost function that with very high probability it is a
> minimizer.)
> 
> [Your description of constraints on the auxiliary page needs a tiny change, namely that "<" and ">" really have to be "<=" and ">=" since minimization is only well defined over closed domains.]

Anders-Markvardsen:
From comments from Nick Gould, see #15901, some smaller updates to Fitting concept page.

**To test:**

Look at changes to rst file and see if these changes provide improvement to reading

Fixes #15901

_Does not need update to release notes._

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/docs/source/concepts/Fitting.rst
+++ b/docs/source/concepts/Fitting.rst
@@ -28,7 +28,9 @@ a so-called least-squares cost function which has the form
 
 .. math:: \sum_i \left( \frac{y_i^{obs}-y_i}{\sigma_i} \right)^2
 
-where the :math:`y_i`'s are calculated from a model. 
+where :math:`y_i` is shorthand for :math:`y(x_i)`, i.e. the :math:`y`-value 
+calculated at the observed value :math:`x_i` from a model, and :math:`y_i^{obs}`
+and :math:`\sigma_i` the observed and error estimate values at :math:`x_i` respectively. 
 Mantid supports a number of other cost functions and these are described :ref:`here <algm-Fit>`. 
 The above cost function is the default Mantid cost function. If you use the 
 :ref:`Fit algorithm <algm-Fit>` this cost function goes under the name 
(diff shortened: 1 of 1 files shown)


Anders-Markvardsen:
All Gould's comments where considered except "[Your description of constraints on the auxiliary page needs a tiny change, namely that "<" and ">" really have to be "<=" and ">=" since minimization is only well defined over closed domains.]" which may not apply directly to the way we do constraint since ours are currently soft constraints so from that point of view remaining a Fit string syntax like *Constraints: "4.0 < c < 4.2"* seems ok for now to me, but Gould's comment is something we can keep in mind. @mantid-roman

NickDraper:
Simple rst doc changes. All seem good.

:shipit:

PR #16123, opened 2016-05-10
Allow log levels to match the lowest of all filter channels

### LINKED ISSUE #14524: Kernel::Logger::is() not working as expected
I'm not sure if this is a confusion but several algorithms send debug messages to the logger only if...
```
if (g_log.is(Kernel::Logger::Priority::PRIO_DEBUG)) ...
```
That doesn't seem to work as expected. Try to open Mantid, set for example `information` level, and check what you get from these ones for example:
```
auto prio1 = g_log.is(Kernel::Logger::Priority::PRIO_DEBUG);
auto prio2 = g_log.is(Kernel::Logger::Priority::PRIO_ERROR);
auto prio3 = g_log.is(Kernel::Logger::Priority::PRIO_INFORMATION);
auto prio4 = g_log.is(Kernel::Logger::Priority::PRIO_CRITICAL);
auto prio5 = g_log.is(Kernel::Logger::Priority::PRIO_FATAL);
```

All the `prio` values are `true` (checked on win7 and debian).

The documentation of `Logger::is` says `Returns true if at least the given log level is set`. Does that "set" mean "selected by the user" or "available"? It seems to behave as the second but the code elsewhere seems to be using as if it should behave as the first option.

Related to #14467, where the behavior of the logger here was really awkward: https://github.com/mantidproject/mantid/blob/master/Framework/CurveFitting/src/CostFunctions/CostFuncLeastSquares.cpp#L410

NickDraper:
The log level of all the loggers is now set to follow the lowest log level of all of the FilterChannels (File, Console and message display in mantidplot).  It used to track them going down to debug, but never back up again, now it does.

**To test:**
This is tested pretty well in the unit tests in ConfigServiceTest.  However you should also start up Mantidplot, change the log level of the log window and check the correct level filter is applied when running algorithms.

Fixes #14524 
##### RELEASE NOTES

---
#### Reviewer

##### Code Review
##### Functional Tests

This should allow the logger.is functionality to work correctly.

### DIFF
--- a/Framework/Kernel/inc/MantidKernel/ConfigService.h
+++ b/Framework/Kernel/inc/MantidKernel/ConfigService.h
@@ -224,6 +224,9 @@ public:
   /// Set the default facility
   void setFacility(const std::string &facilityName);
 
+  /// registers additional logging filter channels
+  void registerLoggingFilterChannel(const std::string &filterChannelName,
+                                    Poco::Channel *pChannel);
   /// Sets the log level priority for the File log channel
   void setFileLogLevel(int logLevel);
   /// Sets the log level priority for the Console log channel
@@ -270,6 +273,7 @@ private:
   void loadConfig(const std::string &filename, const bool append = false);
   /// Read a file and place its contents into the given string
   bool readFile(const std::string &filename, std::string &contents) const;
+
   /// Provides a string of a default configuration
   std::string defaultConfig() const;
   /// Writes out a fresh user properties file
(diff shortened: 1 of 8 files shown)


FedeMPouzols:
- This works well. I tested this in action by changing the log level in the GUI, and then also fiddling with the "channel" options in the `Mantid.properties` files.
- In MantidPlot, when you change the log level you get messages like `MessageDisplayPriority log channel set to PRIO_WARNING priority`
- Definitely fixes the issue. Now the logging filtering mechanism will find the lowest effective level across all the logging "channels".
- Well organized code, and also documented and unit-tested.

:shipit:

I checked this on debian and windows 7. I had some hopes this could speed up some algorithms on windows (calibration, and anything using `Fit`, `FindPeaks`, etc. algorithms which log massive amounts of debug information). Unfortunately it doesn't seem to make a big difference. It's still unclear what makes these algorithms much slower on windows. A long calibration that I remember could take 8-9h still takes approximately that. On debian it takes ~7 minutes.

PR #16131, opened 2016-05-10
Tomography GUI improvements in "ROI" tab for color maps

### LINKED ISSUE #16048: Miscellaneous improvements in the Tomography GUI
Several small improvements
- The convenient 'browse image' next to the 'browse stack' in the ROI selector.
- Selection of multiple color maps
- Possibly more details after tomorrows demo

FedeMPouzols:
Adds a couple of widgets in the "ROI etc" tab of the tomography GUI: Browse individual images, and color bar widget.

**To test:**
- Open "Custom Interfaces"=>"Diffraction"=>"Tomographic Reconstruction"
- Load an individual image with the "Image..." button that you find at the top. For example: `FITS_small_01.fits` from the unit test data.
- The unit test of the ROI widget has been updated and it should pass.
- Check that:
  - You can load maps by double-clicking on the colorbar (or the scale and numbers next to the bar)
  - You can change between linear/log/power scales, change the min/max, etc.
- This modifies the minimum sizes of the color bar widget, which is also used in the slice viewer. Open the slice viewer and check that the color bar still displays correctly there. For example open it for this file from the unit-test data: `SEQ_NDHW.nxs`.

The meaning of the two tick boxes at the bottom of the color bar can be a bit misleading. It might need some clarification/changes, but as this is also used in the slice viewer, I'm not changing it here.
- Autoscale to current slice: if it is on, the interface will get the min/max values from the image, on loading a new image or a new color map. But you can still change them manually in the spin boxes
- Autoscale on load: if t is on, you won't be able to edit the min/max manually. This option probably has little applicability in this context.

Fixes #16048.

Release note: https://github.com/mantidproject/mantid/commit/6e2d4ad6222e6907827d708704a227cb6d1a37db

---
#### Reviewer

##### Code Review
##### Functional Tests


tom-perkins:
Tested as above.
- New button added to load a single image instead of a stack, works correctly
- Color bar added to interface (and still displays correctly in slice viewer)
- Tests added for new functionality

I noticed the two checkboxes below the color bar worked a little differently to above, though. 
- With neither ticked, loading a new color map reset the scale to 1-65536 (different to behaviour in the sliceviewer?). 
- "Autoscale on load" automatically set the scale range to the image when a new color map was loaded.
- "Autoscale to current slice" didn't seem to do anything.
- The tooltips on these tick boxes seem more specific to the sliceviewer than the current interface.

If the tick box behaviour is as intended then I'm happy with this :+1:

FedeMPouzols:
Yes, the checkboxes are very confusing at the moment, not to mention the tooltip text which is so `SliceViewer`-centric. On first startup you'll get the `65536` hardcoded default which makes sense for the typical images that users would load here (16 bits pixels). Then changing the color bar and the min/max is all users would like to have for now.

For the two "autoscale" tick boxes we might add options to the `ColorBarWidget` to hide them, modify their behavior, and/or or modify the label and tooltip text of the two ticks, but I wouldn't do that at this stage. We have to check in a demo session with the IMAT crew what they'd prefer from the options available or any other suggestions. We might well go for a single and simple "Autoscale on load" tick box. This is useful when working with single images, but for stacks or especially when "playing" images a fixed range is preferrable (either default or user-provided) .

tom-perkins:
OK, thanks for explaining! In that case let's :shipit:

FedeMPouzols:
Thanks for bringing it up. I've open an issue for this #16154 so it doesn't get neglected.

PR #16135, opened 2016-05-11
Invalidate common-bins-flag in EventWorkspace.

### LINKED ISSUE #16133: EventWorkspace does not invalidate the "common bins flag" when accessing X data.
In `MatrixWorkspace` (and thus `Workspace2D`), accessing spectrum data, e.g., with `dataX(index)` forwards to `dataX()` of a spectrum, as follows:
```cpp
  MantidVec &dataX(std::size_t const index) {
    return getSpectrum(index)->dataX();
  }
``` 
`getSpectrum` calls `invalidateCommonBinsFlag();` since the bins of a single spectrum might get modified.

However, in `EventWorkspace`, `invalidateCommonBinsFlag();` is called neither directly nor indirectly since `getSpectrum` is not used in the implementation of, e.g., `dataX(index)`, but the access is direct:
```cpp
  return this->data[index]->dataX();
```

SimonHeybrock:
Fixed #16133 by using `getSpectrum` in `dataX` etc., which then invalidates the common bins flag.

**To test:**
- Check if the commons bins flag is now invalided. Try the Python example from the issue.
- Make sure the added regression test fails if the bug is reintroduced.

Unless we encounter previously hidden issues after the fix, this does not need to be in the release notes.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/DataObjects/src/EventWorkspace.cpp
+++ b/Framework/DataObjects/src/EventWorkspace.cpp
@@ -635,10 +635,7 @@ void EventWorkspace::deleteEmptyLists() {
 /// @param index :: the workspace index to return
 /// @returns A reference to the vector of binned X values
 MantidVec &EventWorkspace::dataX(const std::size_t index) {
-  if (index >= this->m_noVectors)
-    throw std::range_error(
-        "EventWorkspace::dataX, histogram number out of range");
-  return this->data[index]->dataX();
+  return getSpectrum(index)->dataX();
 }
 
 /// Return the data X error vector at a given workspace index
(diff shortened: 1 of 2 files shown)


FedeMPouzols:
@SimonHeybrock : you'd want this to go in release 3.7? I think it's worth. If so, I'll be happy to give it a check based on the examples from the issue. All looks good and worth including asap.

NickDraper:
If we are fixing this, don't we need to pay attention to the operations that should re-validate the common-bins-flag.

EventWorkspace->setAllX()

NickDraper:
I'd also like to see a unit test that shows the invalidation and re-validation of the common-bins-flag.

SimonHeybrock:
@FedeMPouzols I am not the right person to judge whether or not this should be in 3.7. Since it is not a big change, I don't see why not.

@NickDraper Once the flag has been invalidated, `MatrixWorkspace::isCommonBins()` will run a check and re-validate the flag, so I don't think it is necessary to do it manually in `setAllX()`.

I will add a unit test as you suggested.

NickDraper:
Unit test is in place and checks both invalidation and revalidation.
Code is good.

This is all under the hood stuff so does not need to be in release notes.

I'm going to assume that the system tests will pass, as they passed last time, and the only addition in between was a unit test.

:shipit:

FedeMPouzols:
And all the tests passed again.

PR #16138, opened 2016-05-11
Add <hollow-cylinder> as recognised shape in XML definitions

### LINKED ISSUE #16132: Add <hollow-cylinder> tag to list of known shapes
In the XML definition of shape an hollow-cylinder, of finite length, can be created from the difference of two cylinders. It would be more convenient to have a specific tag to create such a shape.

This will be more frequently used for defining cans in sample environments.

The suggested definition follows that of a [cylinder](http://docs.mantidproject.org/nightly/concepts/HowToDefineGeometricShape.html#howtodefinegeometricshape#cylinder):

```xml
<hollow-cylinder id="an-1">
  <centre-of-bottom-base x="0.0" y="0.0" z="0.0" />
  <axis x="0.0" y="1.0" z="0" />
  <inner-radius val="0.006"/>
  <outer-radius val="0.0065"/>
  <height val="1" />
</hollow-cylinder>
```

martyngigg:
The `ShapeFactory` has been extended to understand the concept of a `hollow-cylinder`. See the referenced issue for more detail.

**To test:**

Review the code and check the example in the unit tests makes sense.

As a test for visualization I modified the MARI instrument definition locally to use a hollow cylinder instead of a cylinder for the fermi-chopper. The instrument view then shows the expected shape:

hollowcylinderinsrtumentview

Try something similar to convince yourself it works as expected.

Fixes #16132

[Release notes updated](/mantidproject/mantid/commit/c7770f9c3b534949b08ce01c9129bbab32fc9425)

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/Geometry/src/Objects/ShapeFactory.cpp
```diff
@@ -504,6 +507,73 @@ std::string ShapeFactory::parseSegmentedCylinder(
...
+
+  // add bottom plane
+  auto pPlaneBottom = boost::make_shared<Plane>();
+  pPlaneBottom->setPlane(centreOfBottomBase, normVec);
+  prim[l_id] = pPlaneBottom;
+  retAlgebraMatch << "" << l_id << ")";
+  l_id++;
+
+  return retAlgebraMatch.str();
+}
```
mantid-roman: Are we checking for `innerRadius < outerRadius`? Or the largest becomes the outer?

martyngigg: We're not checking at the moment. None of the other definitions do either, which is equally bad. I was a little loathed to restructure the code too much at this stage as it's core stuff.

Do you think this is worth doing?

mantid-roman: Maybe in the new code only?

martyngigg: @mantid-roman I've added some checks for the new type and tests that they fail as expected.

mantid-roman: Zero height behaves OK in the instrument view: no visible shape.

Zero or negative radii give "Unknown exception".

martyngigg: Nice catch. I'll update the error checks.

martyngigg:
@mantid-roman I've added a few more checks and I don't see any exceptions in the instrument view any longer.

mantid-roman:
:shipit:

FedeMPouzols:
Nice checks and tests against bad inputs!

PR #16140, opened 2016-05-11
Fixed error in computation of population factor.

### LINKED ISSUE #15843: Python algorithm to calculate the phonon density of states from powder inelastic data

mducle:
There was an omission in the code for handling the neutron energy gain (upscattering) side in the thermal population (Bose) factor correction - the expressions for the corrections requires abs(E) rather than E.

**To test:**

With a fake dataset:

```
# Creates a fake dataset with an optic phonon mode at temperature T and energy en
T = 300
en = 5
fn = 'name=Gaussian, PeakCentre='+str(en)+', Height=1, Sigma=0.5;'
fn +='name=Gaussian, PeakCentre=-'+str(en)+', Height='+str(np.exp(-en*11.6/T))+', Sigma=0.5;'
ws = CreateSampleWorkspace(binWidth = 0.1, XMin = -25, XMax = 25, XUnit = 'DeltaE', Function = 'User Defined', UserDefinedFunction=fn)
ws = SofQW3(ws, [0, 0.05, 8], 'Direct', 25)
ws = Transpose(ws)
# We also need to get the Q dependence correct or the DOS calculation will be wrong.
qq = ws.getAxis(0).extractValues()
qq = (qq[1:len(qq)]+qq[0:len(qq)-1])/2
y = ws.extractY()
e = ws.extractE()
qqgrid = np.tile(np.array(qq), (np.shape(y)[0], 1))
y = y * qqgrid**2
e = e * qqgrid**2
for i in range(len(y)):
    ws.setY(i, y[i,:])
    ws.setE(i, e[i,:])
ws_DOS= ComputeIncoherentDOS(ws,Temperature=T,EnergyBinning='-20,0.1,20')
```

with a real dataset:

```
from Direct import DirectEnergyConversion
rd = DirectEnergyConversion.DirectEnergyConversion('MARI')
ws = rd.convert_to_energy(21334, 21458, 50, [-45,0.05,45], 'mari_res2013.map',
    monovan_run=21496, sample_mass=9, sample_rmm=249.48, monovan_mapfile='mari_res2013.map')
ws_sqw = SofQW3(ws, [0,0.1,10], 'Direct', 50)
ws_dos = ComputeIncoherentDOS(ws_sqw, Temperature=300, EnergyBinning='-45,0.5,45')
```

With the old algorithm the DOS in the up- and down- scattering sides (negative and positive energy transfer respectively) do not match up, but they should. This PR fixes this. E.g.

dos_popfactor_check

The black curve is without the bugfix, the red curve is with the bugfix. For the generated (fake) dataset, the peak at -en should be the same height as the peak at +en.

Fixes #15843 . 

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/PythonInterface/plugins/algorithms/ComputeIncoherentDOS.py
+++ b/Framework/PythonInterface/plugins/algorithms/ComputeIncoherentDOS.py
@@ -141,6 +141,9 @@ class ComputeIncoherentDOS(PythonAlgorithm):
         DWF = np.exp(-2*(qqgrid**2*msd))
         idm = np.where(engrid < 0)
         idp = np.where(engrid >= 0)
+        # The expression for the population (Bose) factor and phonon energy dependence below actually refer to the phonon
+        # energy (always defined to be positive) rather than the neutron energy transfer. So we need the absolute value here
+        engrid = np.abs(engrid)
         expm = np.exp(-engrid[idm]*mev2k/Temperature)
         expp = np.exp(-engrid[idp]*mev2k/Temperature)
         Bose = DWF*0
--- a/Framework/PythonInterface/plugins/algorithms/ComputeIncoherentDOS.py
+++ b/Framework/PythonInterface/plugins/algorithms/ComputeIncoherentDOS.py
@@ -107,8 +107,10 @@ class ComputeIncoherentDOS(PythonAlgorithm):
         mev2k = (constants.elementary_charge / 1000) / constants.k
 
         # Converts energy to meV for bose factor later
+        input_en_in_meV = 1
         if u0 == 'DeltaE_inWavenumber' or u1 == 'DeltaE_inWavenumber':
             en = en / mev2cm
+            input_en_in_meV = 0
 
         # Checks energy bins are ok.
         if EnergyBinning.count(',') != 2:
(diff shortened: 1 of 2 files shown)


mducle:
@martyngigg - there was a problem with the unit test - in the conversion of the energy transfer from meV to wavenumber and back (sorry I don't know how that got through the last unit test...) I've fixed it now, and also added a unit test for the bose factor correction.

martyngigg:
@mducle No problem, thanks for fixing it.

NickDraper:
Code review is good, tested both approaches, the two sides line up much better.
image

:shipit:

PR #16149, opened 2016-05-11
Simplify algorithm history comparison

quantumsteve:
Description of work.

This simplifies an overly expensive comparison function for `Mantid::API::AlgorithmHistory`. shared_ptrs are now passed by const reference instead of by value (I'd rather pass a simple reference, but the interface for `std::set` prevents that.) and removes the need to wrap the comparison function in a `boost::function`. I also updated a much simpler unnecessary `shared_ptr` copy in CompareWorkspaces.cpp. 

I found about a 10% performance improvement in WorkspaceHistoryTestPerformance ( before 0.53 sec, after 0.48 sec). 

The test itself spends most of its time parsing the defaultTime string.  Constructing a static object and calling the cheaper copy constructor brings the time to 0.17 sec.  Essentially, this performance test was primarily testing the DateAndTime constructor instead of WorkspaceHistory.

**To test:**

Code review: Can I (and if so should I) eliminate the template in CompareHistory? Any suggestions on what, if anything, to put in the release notes?

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/API/src/AlgorithmHistory.cpp
```diff
@@ -28,22 +28,14 @@ AlgorithmHistory::AlgorithmHistory(const Algorithm *const alg,
                                    const double &duration,
                                    std::size_t uexeccount)
     : m_name(alg->name()), m_version(alg->version()), m_executionDate(start),
-      m_executionDuration(duration), m_execCount(uexeccount),
-      m_childHistories(boost::bind(CompareHistory::compare, _1, _2)) {
+      m_executionDuration(duration), m_execCount(uexeccount) {
```
martyngigg: Should we still initialize `m_childHistories` for completeness?

Framework/API/src/AlgorithmHistory.cpp
```diff
@@ -60,8 +52,7 @@ AlgorithmHistory::AlgorithmHistory(const std::string &name, int vers,
                                    const double &duration,
                                    std::size_t uexeccount)
     : m_name(name), m_version(vers), m_executionDate(start),
-      m_executionDuration(duration), m_execCount(uexeccount),
-      m_childHistories(boost::bind(CompareHistory::compare, _1, _2)) {}
+      m_executionDuration(duration), m_execCount(uexeccount) {}
```
martyngigg: Same note as above.

quantumsteve:
I added `m_childHistories()` to the member initializer lists. It shouldn't change the code, but I don't see any harm in being explicit.

martyngigg:
I guess we could remove the template by having the comparator in a separate file, with the implementation in the `.cpp` and forward declares in the headers? While the template might be mildly confusing, the comment explains it well and moving introducing extra files seems unnecessary to me.

martyngigg:
This is a good clean up that comes with a measurable performance gain. I'm happy to :shipit:

PR #16155, opened 2016-05-12
Hexahedron Bounding Box Calculation Fix

### LINKED ISSUE #16106: Bounding Box Calculation for Hexahedral primitives
Currently there seems to be no bounding box calculation for hexahedral primitives:

https://github.com/mantidproject/mantid/blob/master/Framework/Geometry/src/Objects/Object.cpp#L1594

LamarMoore:
There was a bug in the bounding box calculation for hexahedra i.e it wasn't being calculated at all.

Code Review

Fixes #16106 .

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/Geometry/src/Objects/Object.cpp
```diff
@@ -1644,7 +1644,23 @@ void Object::calcBoundingBoxByGeometry() {
       maxZ = std::max(maxZ, vector.Z());
     }
   } break;
-
+  case 2: // HEXAHEDRON
```
SimonHeybrock: There is a comment on the `default:` label below, which says that `2` is a SPHERE. Is the comment wrong, it the `2` here?

Framework/Geometry/src/Objects/Object.cpp
```diff
@@ -1644,7 +1644,23 @@ void Object::calcBoundingBoxByGeometry() {
...
+    auto &lf = vectors[1];
+    auto &lb = vectors[0];
+    auto &rb = vectors[3];
+    auto &rf = vectors[2];
+    auto dz = vectors[4] - lf;
+
+    minX = std::min(lf.X(), lb.X());
+    maxX = std::max(rb.X(), rf.X());
+    minY = lb.Y();
+    maxY = rf.Y();
```
SimonHeybrock: Given the two lines for `minX` and `maxX` above, doesn't this imply that :
```cpp
minY = std::min(lb.Y(), rb.Y());
maxY = std::max(lf.Y(), rf.Y());
```

Framework/Geometry/src/Objects/Object.cpp
```diff
@@ -1644,7 +1644,23 @@ void Object::calcBoundingBoxByGeometry() {
...
+    minZ = 0;
+    maxZ = dz.Z();
```
SimonHeybrock: This implies that the bottom and top plane are always parallel to the x-y plane?

SimonHeybrock:
I am not familiar with this Geometry stuff at all, so my comments above are rather uninformed. Someone else apart from me should have a look, maybe @OwenArnold?

NickDraper:
I've got it

NickDraper:
Yes bounding boxes are always axis aligned for speed and efficiency reasons.

NickDraper:
Is there a reason we don't have any unit tests for this?

SimonHeybrock:
@NickDraper My question was about the alignment of the top and bottom plane of the Hexahedron, not about the planes of its bounding box.

Framework/Geometry/src/Objects/Object.cpp
LamarMoore: The comment is wrong, it needs to be changed. However, there is another pull request which uses a scoped enum instead of hard-coded integer values.

NickDraper:
@SimonHeybrock Sorry, my misunderstanding, Hexahedron surfaces can be any orientation, only bounding boxes are axis aligned.

Framework/Geometry/src/Objects/Object.cpp
LamarMoore: For the current StructuredDetector implementation, yes.

LamarMoore:
@NickDraper I can create an issue to address the unit tests unless you think I should include them now.

NickDraper:
We have bounding box tests for Sphere and Cylinder, I think we should at least add one for hexahedra in this PR.

NickDraper:
e.g. ObjectTest.h->testGetBoundingBoxForSphere()

NickDraper:
Unit test is good, code quality is good.

:shipit:

PR #16161, opened 2016-05-12
Usability fixes in ALC interface

### LINKED ISSUE #16145: ALC improvements to interface
Points raised at meeting with scientists on 11th May 2016.

- Is it possible to zoom in to the graphs? (something users have been asking)
   - It seems not, find out if this is an easy thing to add or not
- It's not clear what the "Function" dropdown does, or that it applies to the selected log.
  - Call it "Process log at" and default to "Mean" not "Last"
  - Move it closer to the Log dropdown so it's clear they go together - or put in their own groupbox?
- Integration range from/to: at present it starts at the first data point. 
  - It should default to starting from first good data (available in file). 
  - If there is a "last good data" then use that as the end, otherwise just use the last point
- If you set up all the options, then change the first run, everything is reset, in particular period combinations. It should leave the options alone unless they are invalid (the new first run doesn't have the right log/number of periods).
  - If you set a range of runs where the last one doesn't have the right number of periods (or the right log), it appears to work at the moment. It should ignore (but warn about) bad data.

tom-perkins:
Several usability fixes, requested by scientists, were made to the interface.

These fix some problems that made the interface inconvenient or confusing to use at times:
- The "Function" box was renamed "Take log value at", defaulted to Mean rather than Last, and moved next to the log to which it applies
- Integration start time is now initialised to the first good data, rather than the first time bin
- When a new first run is loaded, the period control is no longer reset (unless the selected periods don't exist in the new dataset)

**To test:**
_Data available in unit test folder_
- Open Interfaces/Muon/ALC
- In "First" box, select `MUSR00015189.nxs`
  - Note the integration start time ("From" in the bottom group box) is 0.106 us, not a negative number as it was previously. 
- In the "Periods" group box, set the selection to 2 - 1 (or anything other than default)
- Now load `MUSR00015190.nxs` in the "First" box
  - Verify the periods choice has not changed - nothing else should have either
- Load `EMU00006473.nxs` in the "First" box (this is single-period)
  - Now the periods box should change to "1" and the option to subtract will be disabled.

Fixes #16145 

[Release notes](https://github.com/mantidproject/mantid/blob/39ce0b9e4776720082476a9b06beb15e0a80536d/docs/source/release/v3.7.0/muon.rst#muon-alc) have been updated.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Muon/ALCDataLoadingView.h
+++ b/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Muon/ALCDataLoadingView.h
@@ -99,6 +99,10 @@ namespace CustomInterfaces
     // -- End of IALCDataLoadingView interface -----------------------------------------------------
 
   private:
+    /// Common function to set available items in a combo box
+    void setAvailableItems(QComboBox *comboBox,
+                           const std::vector<std::string> &items);
+
     /// UI form
     Ui::ALCDataLoadingView m_ui;
 
--- a/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Muon/ALCDataLoadingView.ui
+++ b/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Muon/ALCDataLoadingView.ui
@@ -107,22 +107,9 @@
              </widget>
             </item>
             <item>
-             <spacer name="logSpacer">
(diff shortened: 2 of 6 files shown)


AntonPiccardoSelg:
Code Review:
 * Code looks fine

Functional Testing:
  * Followed the test instructions and behaved as expected. 

Happy to :shipit:

PR #16193, opened 2016-05-12
Specialize toValue with tag dispatch.

quantumsteve:
Description of work.

This uses tag dispatch instead of macros to specialize `std::vector<integer type>`. The primary benefit is to developers. End-user behavior should not be changed.  

**To test:**

I think this is clearer than the current implementation, but would appreciate a second opinion from someone familiar with metaprogramming.

This has no associated issue number.

---
#### Reviewer

##### Code Review
##### Functional Tests


SimonHeybrock:
Looks good, but I have one question: `bool` is also an integral type, so the new code now takes the other branch. Is that correct?

quantumsteve:
Agreed. I created my own `is_range_type` check which is based on `boost::type_traits::is_integral` but with bool and char as `std::false_type`. 

A range of `bool` doesn't make much sense and operator++(bool) is deprecated anyway.

I can image creating a range of char from a string like `a:z`, but it would also be very easy to type in `aa:z` and get a runtime exception from `boost::lexical_cast`. Since the previous implementation didn't include char, I make it a std::false_type.

SimonHeybrock:
Recent changes look good. :shipit:

PR #16197, opened 2016-05-13
Use copy constructor when cloning TableColumn

### LINKED ISSUE #16136: Use copy constructor when cloning a `TableColumn`
In `TableColumn.h`:

```
  /// Clone
  TableColumn *clone() const override {
    TableColumn *temp = new TableColumn();
    temp->m_data = this->m_data;
    temp->setName(this->m_name);
    return temp;
  }
```

This causes that 'str' columns are converted to 'double'.

To fix this, it was suggested that we could use the copy constructor instead:

```
  TableColumn *clone() const override {
    return new TableColumn(*this);
  }
```

MattKing06:
When using the default constructor to create a TableColumn the types were being decided by finding certain characters in their type names i.e `m.find(`f`)` would set a type to `float`. Since there was no direct conversion from `std::string` to `str` and we were using `m.find('d')` to convert to `double`. the d in std::string was causing any std::string(s) to be converted to double types in the table.

**To test:**
- Ensure that Unit/Doc/System tests pass
- Code Review

Fixes #16136.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/DataObjects/inc/MantidDataObjects/TableColumn.h
+++ b/Framework/DataObjects/inc/MantidDataObjects/TableColumn.h
@@ -145,12 +145,7 @@ public:
     return static_cast<long int>(m_data.size() * sizeof(Type));
   }
   /// Clone
-  TableColumn *clone() const override {
-    TableColumn *temp = new TableColumn();
-    temp->m_data = this->m_data;
-    temp->setName(this->m_name);
-    return temp;
-  }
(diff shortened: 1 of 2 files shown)


raquelalvarezbanos:
I have tested these changes and now we are able to clone a table workspace containing `str` columns. This is the agreed solution discussed on slack a couple of days ago. Ready to merge when all tests pass.

:shipit:

PR #16202, opened 2016-05-13
GetIPTS algorithm

peterfpeterson:
This is a cleanup of an algorithm written by an instrument scientist on SNAP. There are no unit tests and the usage tests are commented out, because it requires ORNL servers to be reachable. The algorithm takes a run number and (optional) instrument to return the IPTS directory containing the run.

**To test:**
1. Be inside the ORNL firewall
2. Try some runs and see that the run is in that proposal.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/PythonInterface/plugins/algorithms/GetIPTS.py
+++ b/Framework/PythonInterface/plugins/algorithms/GetIPTS.py
@@ -32,11 +32,11 @@ class GetIPTS(PythonAlgorithm):
         # check for one after as well
         runIds.append(runnumber + 1)
 
-        runIds = [str(id) for id in runIds if id > 0]
+        runIds = [str(runId) for runId in runIds if runId > 0]
 
         # prepend non-empty instrument name for FileFinder
         if len(instrument) > 0:
-            runIds = ['%s_%s' % (instrument, id) for id in runIds]
+            runIds = ['%s_%s' % (instrument, runId) for runId in runIds]
 
         # look for a file
         for runId in runIds:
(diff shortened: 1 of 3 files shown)


Framework/PythonInterface/plugins/algorithms/GetIPTS.py
```diff
@@ -0,0 +1,84 @@
...
+
+        return instruments
+
+    def findFile(self, instrument, runnumber):
+        # start with run and check the five before it
+        runIds = range(runnumber, runnumber-6, -1)
+        # check for one after as well
+        runIds.append(runnumber + 1)
+
+        runIds = [str(id) for id in runIds if id > 0]
```
rosswhitfield: Overriding built-in function `id`.

docs/source/algorithms/GetIPTS-v1.rst
```diff
@@ -0,0 +1,55 @@
...
+.. summary::
+
+.. alias::
+
+.. properties::
+
+Description
+-----------
+
+Extracts the IPTS number from a run using FileFinde,findRuns. It
```
rosswhitfield: `FileFinde` -> `FileFind`

rosswhitfield:
👍 Works as described.

I did find one exception, when there isn't `IPTS` in the file path to match it returns `""`, _e.g._ `GetIPTS(78,Instrument="CORELLI")`. I don't think this a problem as that due to an inconsistency in our filesystem structure.

:shipit:

rosswhitfield:
Actually no, I'm revoking my :squirrel: from my previous comment. Returning an empty string is a bad idea, it should raise a RuntimeError. Otherwise it could cause data to be written an unexpected location, which could be really bad if it's autoreduction.

You need to protect if `FileFinder.findRuns` returns an unexpected value, we don't know how the file structure may change in the future.

I see two solutions, `filename.find('IPTS')` returns `-1` if it didn't match the string or check before the return that `direc` is not an empty string.

rosswhitfield:
LGTM :shipit:

PR #16207, opened 2016-05-15
Enable SpaceGroup to check if unit cell is compatible with symmetry

### LINKED ISSUE #16206: SpaceGroup should be able to check if a unit cell is compatible with symmetry
As requested in the [forum](http://forum.mantidproject.org/t/real-space-crystallography/118/) it would be useful if `SpaceGroup` was able to check whether the metric tensor of a unit cell is compatible with the symmetry operations.

The basis for this can be implemented in `Group`, a method needs to check whether a tensor is invariant with respect to the group's symmetry operations, i.e.

```
G_i = W_i^T * G * W_i | G_i == G
```
where `W_i` is the matrix of the i-th symmetry operation and `W_i^T` is its transpose. `G_i == G` must hold for all symmetry operations of the group.

This is quick to implement and a very useful addition so I suggest to include this in the current release.

MichaelWedel:
This fixes #16206.

The feature is implemented as described in the ticket. The base functionality is in `Group`, which now has a function that checks whether a tensor changes under any of the symmetry operations contained in the group.

SpaceGroup contains a convenience method that takes a unit cell and returns true if the cell metric is compatible with the group's symmetry. There is a system test that checks that this works for all space groups.

Both levels of functionality are exposed to Python.

**To test:**
Code review. Read the updated documentation and try the new usage examples and some variations.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Geometry/inc/MantidGeometry/Crystal/Group.h
+++ b/Framework/Geometry/inc/MantidGeometry/Crystal/Group.h
@@ -169,6 +169,9 @@ public:
 
   std::vector<Kernel::V3D> operator*(const Kernel::V3D &vector) const;
 
+  bool isInvariant(const Kernel::DblMatrix &tensor,
+                   double tolerance = 1e-8) const;
+
   bool operator==(const Group &other) const;
   bool operator!=(const Group &other) const;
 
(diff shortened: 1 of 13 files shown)


FedeMPouzols:
Nice to see the space group stuff growing and being adopted by more users in different contexts. This comes from some questions and suggestions given in the Mantid forum. It's definitely a good idea to include this in the upcoming release.

- Code: neat, just a small addition, exposed to Python
- Extensive tests and documentation
- I checked manually some examples based on the documentation, and also some other space groups from engineering materials that I had at hand. This behaves well, the true/false value makes sense when applicable, and you get informative messages for the symbols that are not supported/registered.

:shipit:

PR #16221, opened 2016-05-16
Store EMode on workspace during ConvertUnits operation

### LINKED ISSUE #15922: ConvertUnits should set the EMode as a property on the workspace
Running `ConvertUnits` for `Direct` mode currently stores the `Ei` in the experimental logs but does not set the EMode on the ExperimentInfo object. It would be useful to be able to detect this at a later stage rather than requiring each algorithm to pass in the same parameters.

martyngigg:
Stores the EMode as a run log on the workspace.

**To test:**

A code review of the actual change + unit test should suffice for this.

Fixes #15922 

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/API/src/ExperimentInfo.cpp
+++ b/Framework/API/src/ExperimentInfo.cpp
@@ -583,7 +583,7 @@ int ExperimentInfo::getRunNumber() const {
  * the instrument if one is not found. If neither exist then the run is
  * considered Elastic.
  * @return The emode enum for the energy transfer mode of this run. Currently
- * only checks the instrument
+ * checks the sample log & instrument in this order
  */
 Kernel::DeltaEMode::Type ExperimentInfo::getEMode() const {
   static const char *emodeTag = "deltaE-mode";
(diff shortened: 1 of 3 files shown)


FedeMPouzols:
Small but useful addition:
- No other tests were affected by this (which would have been bad).
- The relevant test cases of ConvertUnits have been updated to check the new `deltaE-mode` log entry is set and retrieved correctly.

:shipit:

stuartcampbell:
Why the name `deltaE-mode` and not just `EMode` as we do for `EFixed` ?  Also, how does this cope for instruments with different `EMode`s per detector ?

martyngigg:
@stuartcampbell I don't much like it either but I was following what was already there in a lot of the code. On the plus side all of the uses are internal so can be replaced pretty easily, I just didn't want to make that many changes at this point. Users should not need to interact with that string directly.

It won't cope with instruments with different `EMode`s per detector either but I `ConvertUnits` won't cope with that well will it?

PR #16254, opened 2016-05-17
Fix system test `EnginXCalibrateFullThenCalibrateTest` on windows

FedeMPouzols:
This test failed in the last nightly build: http://builds.mantidproject.org/job/master_systemtests-win7/183/testReport/junit/SystemTests/EnggCalibrationTest/EnginXCalibrateFullThenCalibrateTest/

**To test:**

Check the changes and that the builds pass. The system test should not fail on windows. The PR builds don't check that. You can run it locally with:
`systemtest.bat -C Release -R EnggCalibrationTest.EnginXCalibrateFullThenCalibrateTest` from the "command-prompt" bat. Now this doesn't guarantee much:
- I cannot reproduce this on my windows 7 machine (MSVC community 2015 update 2). For bank 1, the TZERO that I get is `5.24491783405`. The nightly system maintains that it got `5.141012`, which is slightly too far off the reference `5.246633173`. This is a `2.0131%` relative difference, bigger than the `2%` that the test was using as tolerance.
- As it is not the first time that I see slightly different results between different boxes with pretty similar configurations, I've increased the special tolerance of this TZERO parameter on windows to `2.5%`. This is to avoid potential headaches, as the test could fail some nights depending on what machine runs it, but  then pass other nights. A bigger tolerance would be acceptable, but I'd prefer not to increase it unncecessarily. I'd expect `2.5%` is enough.

No issue number for this quick fix.

_Does not need to be in the release notes_.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/PythonInterface/plugins/algorithms/EnggFitDIFCFromPeaks.py
+++ b/Framework/PythonInterface/plugins/algorithms/EnggFitDIFCFromPeaks.py
@@ -60,7 +60,7 @@ class EnggFitDIFCFromPeaks(PythonAlgorithm):
         out_tbl_name = self.getPropertyValue('OutParametersTable')
         self._produce_outputs(difa, difc, tzero, out_tbl_name)
 
-        self.log().information("Fitted {0} peaks in total. DIFA: {1}, DIFC: {2}, TZERP: {3}".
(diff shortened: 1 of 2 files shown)


raquelalvarezbanos:
Test passed on my machine :shipit:

PR #16292, opened 2016-05-18
Refactor checking of equal bin sizes in input workspaces

### LINKED ISSUE #15659: Validator for bin size in input workspaces
Several algorithms need to check that the input workspace has evenly-spaced bins as they won't work on workspaces with uneven bins, logarithmic bins etc. If the difference is small the user should be warned (could be rounding errors in X) but for large differences the workspace should be rejected.

`ApplyDeadTimeCorr`, `FFT` and `MaxEnt` use similar code to do this and it would be good to refactor this into a shared validator. Possibly other algorithms could benefit from this too. The validator should be made as general as possible with adjustable thresholds for warning/error. 
(see #15627)

tom-perkins:
Refactor common validation code in `ApplyDeadTimeCorr`, `FFT` and `MaxEnt` into a new class that can be reused in the future. 
- The thresholds for rejecting the bin sizes or printing a warning are both adjustable.
- Can also change bin width to compare to from "average" to "first", and type of error from cumulative to individual. These options are used to keep the existing behaviour of `FFT` when _AcceptXRoundingErrors_ is not set.

**To test:**
Unit tests have been added for the new class, and the existing unit tests for `FFT`, `MaxEnt` and `ApplyDeadTimeCorr` should all pass. There should be no functional changes as this is a refactoring only.

Fixes #15659.

_Does not need to be in the release notes_ - internal change only.

---
#### Reviewer

##### Code Review
##### Functional Tests


FedeMPouzols:
Good idea. Maybe it's worth adding a "EqualBinSizesValidator" based on this? We have a related ["CommonBinsValidator"](https://github.com/mantidproject/mantid/blob/master/Framework/API/inc/MantidAPI/CommonBinsValidator.h) but as far as I know we don't have any validator that checks the "equal bins" aspect.

tom-perkins:
I had intended to write this as a workspace validator at first, but the reason I didn't is because that would check the workspace as soon as the property `InputWorkspace` was set. 
For `FFT`, the check is different based on whether another property `AcceptXRoundingErrors` is set, so the check needs to come later on.

raquelalvarezbanos:
Changes look good to me: we have two new classes, `EqualBinChecker`, which checks if X values are evenly spaced (within a given tolerance), and a new `EqualBinSizesValidator` that uses `EqualBinChecker` to validate the input workspace. In addition, `CommonBins` are expected.

This branch has some conflicts, please @tom-perkins let me know when those are resolved.

tom-perkins:
@raquelalvarezbanos I've merged master into this branch to fix the conflicts, hopefully this should not affect whether the builds pass

raquelalvarezbanos:
Thanks for letting me know @tom-perkins.

:shipit:

PR #16304, opened 2016-05-19
Fix Coverity warning from SaveTBL algorithm

### LINKED ISSUE #16297: Fix Coverity Warning for SaveTBL
I received this warning from the Coverity Scan regarding SaveTBL.

```
>>>     CID 1355568:  Error handling issues  (CHECKED_RETURN)
>>>     Calling "remove(filename.c_str())" without checking return value. This library function may fail and return an error code.
93             remove(filename.c_str());
```

I believe there should be a check to see when `remove()` has failed and present the user with an associated warning.

MattKing06:
Report from Coverity scan:

```
*** CID 1355568:  Error handling issues  (CHECKED_RETURN)
/home/builder/jenkins-linode/workspace/coverity_build_and_submit/Framework/DataHandling/src/SaveTBL.cpp: 93 in Mantid::DataHandling::SaveTBL::exec()()
87               throw std::runtime_error(groupHeading +
88                                        " Column must be of type \"int\"");
89             } else
90               writeVal<int>(row.Int(columnIndex), file);
91           } else if (ws->getColumn(columnIndex)->type() != "str") {
92             file.close();
>>>     CID 1355568:  Error handling issues  (CHECKED_RETURN)
>>>     Calling "remove(filename.c_str())" without checking return value. This library function may fail and return an error code.
93             remove(filename.c_str());
94             throw std::runtime_error(columnHeadings[columnIndex] +
95                                      " columns must be of type \"str\"");
96           } else {
97             if (columnIndex == columnHeadings.size() - 1)
98               writeVal<std::string>(row.String(columnIndex), file, false, true);

```

Unfortunately the only way to truly test if this fix has worked is to merge it so coverity can be run again.

**To test:**
- Code Review

Fixes #16297.
#### Reviewer

##### Code Review
##### Functional Tests


FedeMPouzols:
If the chances of getting this error are not too remote it might make sense to include this fix in v3.7.

MattKing06:
@FedeMPouzols I discussed this with Nick this morning and we decided it might be a bit risky to merge this when we're so close to release. So I've marked it against 3.8 and hopefully this can be merged after the release :+1:

Framework/DataHandling/src/SaveTBL.cpp
```diff
@@ -83,16 +83,28 @@ void SaveTBL::exec() {
         std::string groupHeading = columnHeadings[columnIndex];
         if (ws->getColumn(groupHeading)->type() != "int") {
           file.close();
-          remove(filename.c_str());
-          throw std::runtime_error(groupHeading +
-                                   " Column must be of type \"int\"");
+          int error = remove(filename.c_str());
+          if (error == 0)
```
martyngigg: Is the error code check correct here? http://www.cplusplus.com/reference/cstdio/remove/ suggests that `0` marks success yet we are throwing an error on that scenario?

martyngigg: I've just reread that and I didn't take the error throwing in context with line 84 so I see now that the error relates to the column type being incorrect.

Please ignore my previous comment.

MattKing06: We throw in the event that remove was successful because the table was not formatted correctly, and we throw in the event remove was unsuccessful because we failed to remove the file. I hope this explains the code a bit more

Framework/DataHandling/src/SaveTBL.cpp
```diff
@@ -83,16 +83,28 @@ void SaveTBL::exec() {
...
+            throw std::runtime_error(groupHeading + " column must be of type "
+                                                    "\"int\". The TBL file "
+                                                    "will not be saved.");
+          else
+            throw std::runtime_error(
```
martyngigg: Can we add something in to the error comment about the file not being removed and it being incorrect? If it's not removed then someone may see it and think it is okay?

MattKing06: Yes, I was messing around with different error messages for both cases and I didn't really know how best to phrase it, but in the event that remove has failed I agree and I think we should add something extra about remove failing.

martyngigg:
Yes this looks fine now.

:shipit:

PR #16320, opened 2016-05-20
Fix connect call on AlgorithmDialog creation

### LINKED ISSUE #16319: Clicking close on algorithm dialog without running holds on to memory
### Expected behavior
Clicking close on an algorithm box does not hold on to memory any longer than clicking run.

### Actual behavior
Opening a dialog for an algorithm with an input workspace then closing it again does not allow memory to be freed when using `Delete`.

First use this script to produce a workspace ~4gb in size:
```python
ws = CreateSampleWorkspace(NumBanks=10, BankPixelWidth=100, Random=True, BinWidth=10)
```
and observe that MantidPlot is using this memory in the system monitor. Now:

* type `CropWorkspace` into the "Algorithms" box and click "Execute"
* hit "close" instead of "run"
* click "Delete" on the "Workspaces" list and see that the memory of the process is unchanged.

### Platforms affected

All

martyngigg:
The `close` button on `AlgorithmDialog` is now wired to the `reject()` slot so that other parts of the code assuming the `rejected()` signal was fired behave as intended.

**To test:**

Use the steps in the issue #16319 to reproduce the problem. I suggest first trying it without merging the fix to convince yourself that you understand the problem. After merging the problem should be fixed. 

Fixes #16319

---
#### Reviewer

##### Code Review
##### Functional Tests

On closing a dialog the algorithm must be removed from the algorithm manager. The removal
is keyed from the rejected() signal but this only fires on reject() calls not close()

### DIFF
--- a/MantidQt/API/src/AlgorithmDialog.cpp
+++ b/MantidQt/API/src/AlgorithmDialog.cpp
@@ -650,7 +650,7 @@ QLayout *AlgorithmDialog::createDefaultButtonLayout(
   m_okButton->setDefault(true);
 
   m_exitButton = new QPushButton(cancelText);
-  connect(m_exitButton, SIGNAL(clicked()), this, SLOT(close()));
+  connect(m_exitButton, SIGNAL(clicked()), this, SLOT(reject()));
 
   QHBoxLayout *buttonRowLayout = new QHBoxLayout;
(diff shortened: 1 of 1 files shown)


LamarMoore:
Works as described. I am happy to :shipit:

PR #16331, opened 2016-05-20
Engg GUI beta/demo fixes and improvements

### LINKED ISSUE #16252: Improvements and fixes in Engg GUI
A list of improvements and fixes that came up in the last Engg GUI demo and are worth fixing during beta testing:

- [x] Remove the spurious new/empty lines in the output .prm files.
- [x] Change the text label of the "force recalculate" tick box to "Force recalculate all existing Vanadium files"
- [x] Show important messages in the custom GUI status bar in addition to the log window
   - [x] Convenient: when using "Fit" in the fitting tab, the status should say how many peaks were fitted or at least if something was fitted, to prevent confusion.
- [x] Show a very visible message when the RB number is empty (and the widgets are grayed out) telling the user that it is required

- [x] As a safety measure, make sure that there's a catch and informative message for all the plot functions executed via the view/PythonRunner.

- [x] Make sure all output files are copied to the general output directory in addition to the RBNumber/user directory, for both calibration and focusing.

Other minor issues from the fitting tab:
- [x] Fitting tab: see if we can find a better min-max range for the peaks curves (for some peaks the fitted curve is cut too abruptly).
- The separate window plot could also show the spectrum, together with the peaks? We might want to check this with scientists.

This was in the end ready for 3.7:
- [x] TOF->dspacing conversion issue in the fitting tab. Use the GSAS-II style conversion rather than ConvertUnits, with the calibrated bank DIFA, DIFC, TZERO. The parameters are added to the focus workspace, and EnggFitPeaks uses them when available. A table workspace is also created, visible in the ADS window, with the per-bank GSAS parameters.

This may not be ready for 3.7: 
- a "Texture calibration" group in the Calibration tab.
- ~~TOF->dspacing conversion issue in the fitting tab. The preferred solution here is to use the GSAS-II style conversion rather than ConvertUnits, with the calibrated bank DIFA, DIFC, TZERO~~.

FedeMPouzols:
Fixes a handful of GUI issues identified in a demo/early beta testing (see issue for details). This should close all the issues that were deemed as doable and low risk during the beta fixes period.

**To test:**
- Check code changes:
  -  The changes in EnggUtils.py fix the issue of the spurious new lines in the calibration (.prm) files.
  - A couple of methods have been added to the view to handle the "very visible message when the RB number is empty/invalid" and the status messages.
  - The presenter takes care of updating the status when starting/finishing any significant work (calibrate, focus, rebin, fit). The relevant test cases have been updated to cover this as far as possible.
  - About half of the lines change where format changes. It seems that MSVC applies a slightly different clang formatting to multi-line stream in/out than my emacs package, both accepted by clang-format.
- Unit tests have been added/extended
- Documentation has been updated.
- Functional tests:
  - See what happens when you leave the RB Number field empty
  - Status messages: check that when you start, and then see finishing a calibration, focusing, rebinning, fitting, the status text at the bottom is updated correctly. When fitting peaks fails it should say that no peaks could be fitted.
  - There's a new table workspace: engggui_calibration_banks_parameters, as described in the documentation changes. See that it contains reasonable values and is updated with new calibrations.
  - Output directories: together with the other points, check that the output files are generated both in the `User/RBNumber` directory and in the general/all directory (directly under `EnginX_Mantid`). This includes the '.prm' files (calibration) and for focusing the '.nxs' file and, when enabling the "output files" option the '.dat', '.gss', and '.his' files. Log messages are also generated that say where the files are copied.
  - In the fitting tab, preview plot, check that the left and right tails of the peak shapes are plotted reasonably well. The startX/endX window around the peak centers has been increased to `X0-9*S`, `X0+12*S` when evaluating the peak. Hopefully this is enough.
  - Try to raise any remaining issues in the GUI.

Fixes #16252.

I think this does not deserve any specific release note, as it is just about polishing the interface, fixing recent issues, and a bit of eye candy.

---
#### Reviewer

##### Code Review
##### Functional Tests


FedeMPouzols:
@shahrozahmed1: I think this is now ready for testing. This addresses all the issues listed in #16252, except the ones that were "better to leave for the next releases". Many of them are small issues and simple changes, but the change of method for conversion of units in the fitting tab required some more in-depth changes to the algorithms.

FedeMPouzols:
@shahrozahmed1: I cannot reproduce the crash with empty RB number / clode /reopen, so it seems to be windows specific. I'm putting the "in progress" label.

FedeMPouzols:
@shahrozahmed1: I think the crashes on windows are now fixed now. It seems that the wrong root for Poco paths will cause all sorts of difficult to trace memory corruptions rather than failures. I was getting crashes in different places, but only sometimes, some other times I wouldn't get any issue.

Also, the second commit fixes another elusive potential crash that you could get in the fitting tab, which I could only get occasionally on windows. The method to enable/disable the buttons was being called from doFitting(), which is in another thread. I just moved it to the more external `processFitPeaks` method which runs in the GUI thread and can do those operations safely.

shahrozahmed1:
snipper
Other than Calibration spelling error here, small addition for future could be that to display `Calibration finished and ready as 'current calibration'` after all the `DeleteWorkspace` has successfully completed.

shahrozahmed1:
Also it seems that we have a slightly different clang-formatted because when I clang-format presenter file, it seems to change half of the file again?

shahrozahmed1:
The code look good and documentation and test have been updated. 
Carried out the functional testing and everything seems to work smooth now. New useful feature which displays the status messages on the interface according to the process being carried out. When `RB Number` is empty a splash message is displayed to avoid any confusion for the users. *(I am not sure if it is an intentional feature but further directories can also be created just by using `/` within the `RB Number` This could also be mentioned in the documentation at some point in future*)
The fitting peaks seem to also plot better as full curve can be displayed now instead of it just dropping down in to a straight line. The calibration file doesn't seem to producing new random lines new lines and the output files are now saved within the local (`RB Number`) and global directory (`EnginX_Mantid`). This should tick of all the issues which we came across during the demo to scientists. 
:shipit:

FedeMPouzols:
Oops, the last commit just fixes the "cablibration" typo.

shahrozahmed1:
:shipit:

Issue #16128, opened 2016-05-10
Remove Qt3Support Warnings in ApplicationWindow.cpp


quantumsteve:
I split the warnings in ApplicationWindow.cpp from [build #34](http://builds.mantidproject.org/job/master_clean-Qt3-warnings/34/) into 10 chunks of ~80 warnings and assigned myself the remainder. The actual line numbers may change slightly in later commits.

Please add your username to the list below and include this ticket number in your PR.

1. @AntonPiccardoSelg
2. @AntonPiccardoSelg
3. @tom-perkins
4. @AntonPiccardoSelg
5. @AntonPiccardoSelg
6. @tom-perkins
7. @AntonPiccardoSelg
8. @tom-perkins
9. @shahrozahmed1 
10. @FedeMPouzols
11. @quantumsteve

### No. 1

ApplicationWindow.cpp:454
ApplicationWindow.cpp:457
ApplicationWindow.cpp:469
ApplicationWindow.cpp:492
ApplicationWindow.cpp:1078
ApplicationWindow.cpp:1079
ApplicationWindow.cpp:1081
ApplicationWindow.cpp:1082
ApplicationWindow.cpp:1083
ApplicationWindow.cpp:1084
ApplicationWindow.cpp:1085
ApplicationWindow.cpp:1086
ApplicationWindow.cpp:1113
ApplicationWindow.cpp:1114
ApplicationWindow.cpp:1135
ApplicationWindow.cpp:1138
ApplicationWindow.cpp:1146
ApplicationWindow.cpp:1151
ApplicationWindow.cpp:1154
ApplicationWindow.cpp:1158
ApplicationWindow.cpp:1161
ApplicationWindow.cpp:1166
ApplicationWindow.cpp:1172
ApplicationWindow.cpp:1178
ApplicationWindow.cpp:1189
ApplicationWindow.cpp:1192
ApplicationWindow.cpp:1198
ApplicationWindow.cpp:1212
ApplicationWindow.cpp:1248
ApplicationWindow.cpp:1269
ApplicationWindow.cpp:1272
ApplicationWindow.cpp:1274
ApplicationWindow.cpp:1304
ApplicationWindow.cpp:1307
ApplicationWindow.cpp:1310
ApplicationWindow.cpp:1316
ApplicationWindow.cpp:1331
ApplicationWindow.cpp:1335
ApplicationWindow.cpp:1339
ApplicationWindow.cpp:1350
ApplicationWindow.cpp:1351
[...]

PR #16340, opened 2016-05-20
Fix Qt3Support warnings, group 10 of ApplicationWindow

FedeMPouzols:
This is for group 10 of ApplicationWindow: #16128.

**To test:**

Several places of MantidPlot are modified here:
- Review the changes
- Open MantidPlot. See that the main menus look ok
- See that you can create objects of all the different types from "File"->"New". Check that you can manipulate them normally.
- Check that your list of recent projects look normal. If you don't have any, save one or two projects and check the list.
- Similarly, check the list of recent files.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidPlot/src/ApplicationWindow.cpp
+++ b/MantidPlot/src/ApplicationWindow.cpp
@@ -13625,7 +13625,7 @@ void ApplicationWindow::translateActionsStrings() {
   conestyle->setToolTip(tr("Cones"));
   conestyle->setStatusTip(tr("Cones"));
   crossHairStyle->setText(tr("Crosshairs"));
-  crossHairStyle->setMenuText(tr("Crosshairs"));
+  crossHairStyle->setText(tr("Crosshairs"));
   crossHairStyle->setToolTip(tr("Crosshairs"));
   crossHairStyle->setStatusTip(tr("Crosshairs"));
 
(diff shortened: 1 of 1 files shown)


MantidPlot/src/ApplicationWindow.cpp
```diff
@@ -13625,47 +13625,47 @@ void ApplicationWindow::translateActionsStrings() {
   conestyle->setToolTip(tr("Cones"));
   conestyle->setStatusTip(tr("Cones"));
   crossHairStyle->setText(tr("Crosshairs"));
-  crossHairStyle->setMenuText(tr("Crosshairs"));
+  crossHairStyle->setText(tr("Crosshairs"));
```
tom-perkins: It looks as though `setText` is now called twice for each of the menu options in this section

MantidPlot/src/ApplicationWindow.cpp
```diff
@@ -15175,19 +15174,19 @@ void ApplicationWindow::windowProperties() {
   s += tr("Label") + ": " + static_cast<MdiSubWindow *>(w)->windowLabel() +
        "\n\n";
 
-  if (w->isA("Matrix")) {
+  if (QString(w->metaObject()->className()) == "Matrix") {
     mbox->setIconPixmap(getQPixmap("matrix_xpm"));
     s += tr("Type") + ": " + tr("Matrix") + "\n\n";
-  } else if (w->inherits("Table")) {
+  } else if (QString(w->metaObject()->className()) == "Table") {
```
tom-perkins: This one was `inherits` (rather than `isA` like the other lines) which is still ok in Qt5, I think. Would the changed version match something that derived from a Table e.g. MantidTable?

tom-perkins:
Tested all areas with changes, including menu options (3D graphs), *File > New* creation of objects, recent projects, recent files, command line arguments, save project, "Go to row/column", cascade windows, add user menu, toolbars and menus. Everything works just as before. :+1:

MantidPlot/src/ApplicationWindow.cpp
FedeMPouzols: Oh yes, this change is wrong. I'll fix/revert it. That line didn't need any fix, `inherits()` doesn't seem to be deprecated.

MantidPlot/src/ApplicationWindow.cpp
FedeMPouzols: True. Then this is totally redundant now. I wonder what the original code was meaning to do here. I'll just remove those lines, as the only replacement for `setMenuText()` is `setText()`.

tom-perkins:
All looks good now. :shipit: (once builds pass)

FedeMPouzols:
Thanks for the careful checking, @tom-perkins. I just removed the redundant `setText()`s and undid the change of that `inherits()` to isA-like comparison.

PR #16366, opened 2016-05-24
Add documentation page with comparison of performance of minimizers

### LINKED ISSUE #16361: Add minimizers documentation with comparison of performance
Add short documentation for minimizers and tables with summary and details on performance (accuracy and run time) for all minimizers, except FABADA.

Use the test problems from the NIST benchmark, and if they're ready in time some of the test problems from CUTEst. Organize the tables in three blocks, following the NIST categories: lower, average, higher difficulty.

This should be an RST page, included in `concepts`, with a subpage with detailed tables, produced by the scripts / system test from #15952.

FedeMPouzols:
A comparison of Mantid minimizers, using the NIST nonlinear regression benchmark.

**To test:**

Read and comment.

Fixes #16361

Release note in framework/CurveFitting: https://github.com/mantidproject/mantid/commit/2b72a6732e93a06c2041ba1ff8275211fef4ae76

---
#### Reviewer

##### Code Review
##### Functional Tests


FedeMPouzols:
@Anders-Markvardsen, @mantid-roman, @NickDraper: I'm opening the PR now so you have time to comment. The tables are now divided in three blocks as discussed las week.

I took a closer look at the "nan" rows as they looked a bit suspicious to me. I found that this had been introduced when I added Damping to the list of minimizers. The only minimizer producing `nan`s is Damping, and there was a bug (a non nan-safe minimum) that  was propagating the nan to the other minimizers in the row. This implies that the results are even better for Levenberg-Marquardt, as it gets a score of "1" for many of the cases where Damping produces 'nan' (and others tend to produce poor results / not really converge).

On the other hand, if I was to criticize this ranking, for the more noisy datasets the advantages of Levenberg-Marquardt are not so obvious. One could say that there is a bias towards "clean" datasets in these test problems. I'm going to try to add a fourth block using a handful of CUTEst problems. I'll also try to remove the jQuery dependency. The current javascript trick seems to be working well for both `docs-html` and `docs-qthelp`.

docs/source/concepts/FittingMinimizers.rst
```diff
@@ -0,0 +1,205 @@
...
+====================
+
+Minimizers play a central role when :ref:`Fitting a model <Fitting>`
+in Mantid. Given the following elements:
+
+- a dataset (e.g. a spectrum),
+- a model or function to fit (e.g. a peak or background function, with parameters),
+- an initial guess or starting point for the parameters of the function,
+- a cost function (e.g., squared errors weighted by the spectrum errors),
+- and a minimizer,
```
mantid-roman: . (full stop)

docs/source/concepts/FittingMinimizers.rst
```diff
@@ -0,0 +1,205 @@
...
+
+The minimizer is the algorith or method that adjusts the function
```
mantid-roman: algorithm (typo)

docs/source/concepts/FittingMinimizers.rst
```diff
@@ -0,0 +1,205 @@
...
+parameters so that the model fits the data as closely as possible. The
+cost function defines the concept of how close a fit is to the
+data. See the general concept page on :ref:`Fitting <Fitting>` for a
+broader discussion of how these components interplay when fitting a
+model with Mantid.
+
+Several minimizers are included with Mantid and can be seleced in the
```
mantid-roman: selected (typo)

docs/source/concepts/FittingMinimizers.rst
```diff
@@ -0,0 +1,205 @@
...
+`Fit Function property browser
+<http://www.mantidproject.org/MantidPlot:_Creating_Fit_Functions>`__
+or when using the algorithm :ref:`Fit <algm-Fit>` The following
+options are available:
+
+- `Simplex <https://en.wikipedia.org/wiki/Nelder%E2%80%93Mead_method>`__
+- `SteepestDescent <https://en.wikipedia.org/wiki/Gradient_descent>`__
+- `Conjugate gradient (Fletcher-Reeves imp.) <https://en.wikipedia.org/wiki/Nonlinear_conjugate_gradient_method>`__
+- `Conjugate gradient (Polak-Ribiere imp.) <https://en.wikipedia.org/wiki/Nonlinear_conjugate_gradient_method>`__
+- `BFGS (Broyden-Fletcher-Goldarfb-Shanno) <https://en.wikipedia.org/wiki/Broyden–Fletcher–Goldfarb–Shanno_algorithm>`__
```
mantid-roman: Goldfarb

docs/source/concepts/FittingMinimizers.rst
```diff
@@ -0,0 +1,205 @@
...
+The problems have been obtained from two different benchmarks:
+
+- `NIST nonlinear regression problems <http://itl.nist.gov/div898/strd/general/dataarchive.html>`__.
+
+- The `Constrained and Unconstrained Testing Environment <http://www.cuter.rl.ac.uk/Problems/mastsif.html>`__
+  set of test problems for linear and nonlinear optimization.
+
+As the NIST problems do not include measurement errors, assuming that
+these datasets would represent spectra from a typical Mantid workspace
+we introduce errors as the square root of the observations.
```
mantid-roman: I disagree with this approach. Reasons:

1. NIST problems don't represent typical mantid spectra.
2. They were selected to test how algorithms deal with certain difficulties presented by the cost functions landscapes. The introduced errors may distort the shape of the function and remove the challenging feature there.
3. Large errors reduce the absolute value of the chi squared making it look better than it is.

docs/source/concepts/FittingMinimizers.rst
```diff
@@ -0,0 +1,205 @@
...
+
+From the second source, which is a large collection of diverse
+problems (over a thousand) a subset of problems that are relevant for
+fitting in Mantid was selected. These problems are unconstrained, with
+a sum of squared objective function, and regular (first and second
+derivatives exist and are continuous everywhere), following the `CUTE
+classification scheme
+<http://www.cuter.rl.ac.uk/Problems/classification.shtml>`__ for
+optimization test problems.
```
mantid-roman: Will this be relevant and clear to the user?

FedeMPouzols: Definitely not as it stands now. This was here more for completeness. If I manage to add a decent table with cutest problems this would be moved to the detailed results, or it could be just removed. It would be just to explain what subset of problems was picked, as there are many others that are unrelated.

docs/source/concepts/FittingMinimizers.rst
FedeMPouzols: I agree that this is very arguable, and I'm myself uneasy with this, as it effectively means that we are not testing the real NIST  benchmark.
But the problem is that `Levenberg-Marquardt` doesn't behave if there are no errors. So that's the only alternative that I saw. Otherwise we would need to take out the best minimizer, or would there be an alternative?
On the other hand the comparison tables would look pretty similar.

mantid-roman: LM should behave fine with the default cost function (Least Squares). The parameter errors will be wrong but we are not testing them here. I understand that the chi squared is calculated independently?

FedeMPouzols: I'll try that. You mean setting all errors to zero and set the flag to ignore them all? We'd still have to see what do do about this.
Not using errors is good in that it keeps the original NIST problems, but it is bad in that it goes further in the direction of having "toy" problems that do not resemble our users' problems.
Yes, the chi squared is calculated directly from the residuals.

 If it works well without errors we could show that for simplicity, but I'd then at least mention somewhere how it changes with fake errors, or give the tables obtained with errors in yet one more subpage.

NickDraper:
Notes from my test:
1) The Todo lines need removing
2) The cell colouring does not work in IE (but does in Chrome and qtAssistant)
3) The Dan Wood link seems to be broken

FedeMPouzols:
There was another exception with the file names of `Rat42` and `Rat43` on the NIST website! They're also broken but it's easy to fix.

NickDraper:
removed myself as tester as I will not be in again until Thursday PM

docs/source/concepts/FittingMinimizers.rst
```diff
@@ -0,0 +1,205 @@
...
+- :ref:`FABADA <FABADA>`
+
+**TODO: Missing info: *Damping* differences between Lev-Mar and Lev-Mar "MD"**.
+
+All these algorithms are `iterative
+<https://en.wikipedia.org/wiki/Iterative_method>`__.  The *Simplex*
+algorithm, also known as Nelder–Mead method, belongs to the class of
+optimization algorithms without derivatives, or derivative-free
+optimization (another example of this class would be simulated
+annealing). Note that here simplex refers to downhill simplex
```
Anders-Markvardsen: I would suggest deleting "(another example of this class would be simulated annealing)" since it is referring to entirely different class of minimizers, i.e. a global minimizer.

docs/source/concepts/FittingMinimizers.rst
```diff
@@ -0,0 +1,205 @@
...
+optimization. *Steepest descent* and the two variants of Conjugate
+Gradient included with Mantid (*Fletcher-Reeves* and *Polak-Ribiere*)
+belong to the class of optimization or minimization algorithms
+generally known as conjugate gradient, which use first-order
+derivatives. The derivatives are calculated on the error of the fit to
```
Anders-Markvardsen: For "The derivatives are calculated on the error of the fit to drive the iterative process towards a local minimum" do you mean "The derivatives with respect to the cost function are used to drive the iterative process towards a local minimum"

docs/source/concepts/FittingMinimizers.rst
```diff
@@ -0,0 +1,205 @@
...
+drive the iterative process towards a local minimum.
+
+BFGS and the Levenberg-Marquardt algorithms belong to the second-order
+class of algorithms, in the sense that they use second-order
+information of the error function (second derivatives or the Hessian
```
Anders-Markvardsen: By "error function" I assume you here mean the "cost function"?

Anders-Markvardsen:
@mantid-roman if you have done this already, can you please provide a two liner description (or more if you wish) on "Damping" and "Levenberg-MarquardtMD"

docs/source/concepts/FittingMinimizers.rst
```diff
@@ -0,0 +1,205 @@
...
+
+- Referring to the accuracy of a minimizer, it produces a solution
+  with squared residuals 25% larger than the best solution in Mantid.
+- Referring to the run time, it takes 25% more time than the fastest
+  minimizer.
+
+All the minimizers available in Mantid 3.7 were compared, with the
+exception of FABADA which belongs to a different class of methods and
+would not be compared in a fair manner. Minimizers are used here as
+"black boxes", without any special initialization, constraints, or
```
Anders-Markvardsen: I would consider deleting or rewrite this sentence "Minimizers are used here as “black boxes”, without any special initialization, constraints, or rewriting of the equations to fit." since I don't understand what information it adds

Anders-Markvardsen:
@FedeMPouzols For sure this work must be mentioned in release notes. e.g. in  docs/source/release/v3.7.0/framework.rst below "Concept page for :ref:`Mantid Fitting <Fitting>` add new bullet point with something like: "Concept page for :ref:`Comparing fit minimizers <FittingMinimizers>` has been added to provide information, which compare how well different minimizers do against different fitting problems"

docs/source/concepts/FittingMinimizers.rst
```diff
@@ -0,0 +1,205 @@
...
+and different variants).  Run time is measured for the execution of
+the :ref:`Fit <algm-Fit>` algorithm for an equation and dataset
+previously created.
+
+.. There would be two alternative for the errors:
+   1. Without errors, as it is
+   2. With "simulated" sqrt(X) errors
+   Here we use 2 because Levenberg-Marquardt doesn't work with 1 ('Unweighted least squares' cost function
+
+The cost function used in all cases is `Least squares` (weighted by
```
FedeMPouzols: Note: clarify here that "Least squares" is the name in Mantid, and that this is the default choice of cost function. It is the weighted version.
The name of the unweighted version is "Unweighted least squares".

Anders-Markvardsen:
With the above suggestions for changes these tables provide new insight into the Mantid minimizers and provide a mechanism for how we can compare future proposed minimizers against existing minimizers. Also this can in the future be extended with more fitting problems and potentially to even be able to recommend different minimizers for differing classes of fitting problems.

FedeMPouzols:
I've just removed the "in progress" label and I think this is ready for testing.
The main results are now "unweighted", and there's an additional page with equivalent tables for the artificially "weighted" case.

NickDraper:
I think we just need a short one line description of 
•Levenberg-MarquardtMD
•Damping

I'll try to corner Roman

mantid-roman:
Should I do it on the PR branch? Or I can do it on master after the merge. There is no need to re-run all the checks.

FedeMPouzols:
It took a while to re-generate the runtime tables again, and the builds are still ongoing, but the last commit does the change to link to specific low/average/higher tables from the cells of the summary tables. It all displays well for me (with the IExplorer exception).

NickDraper:
This is good now, a real step up from the lack of docs had before this

:shipit:

NickDraper:
PS: No need to wait for the builds to finish, the last commit is docs only and has been tested locally.

PR #16125, opened 2016-05-10
Get FindNumpy working with python3

rosswhitfield:
FindNumpy.cmake doesn't work with python3 because the print syntax has changed.

```
CMake Error at buildconfig/CMake/FindNumpy.cmake:20 (if):
  if given arguments:

    "NOT" "File \"<string>\", line 1
      import numpy" " print numpy.get_include()
                              ^

  SyntaxError: invalid syntax" "MATCHES" "Traceback"

  Unknown arguments specified
Call Stack (most recent call first):
  Framework/PythonInterface/CMakeLists.txt:18 (find_package)


-- Configuring incomplete, errors occurred!
```

**To test:**
- Run cmake on a system with python3 and see if it finds numpy. You can run cmake defining which python to use like `cmake -DPYTHON_EXECUTABLE=/usr/bin/python3 .`
- Make sure it still works with python2, _i.e._ the builds pass

_Does not need to be in the release notes._

---
#### Reviewer

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):
##### Code Review
- [x] Is the code of an acceptable quality?
- [ ] Does the code conform to the coding standards? Is it well structured with small focussed classes/methods/functions?
- [ ] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?
##### Functional Tests
- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant documentation been added/updated?
- [ ] Is user-facing documentation written in a user-friendly manner?
- [ ] Has developer documentation been updated if required?
- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

PR #16381, opened 2016-05-24
CMake updates for building with with boost python 2 or 3

martyngigg:
This updates the cmake code for finding boost python so that it finds the version appropriate to the version of python being used.

The naming the library linked to python3 is not consistent across platforms. The standard form is `boost_pythonSUFFIX`. A list of known suffixes is used by default and a cmake user variable `BOOST_PYTHON_ADDITIONAL_SUFFIXES` has been defined to allow the user to supply additional ones to check. 

Note: This has had the branch from #16125 merged into it.

**To test:**

Run

```
cmake -DPYTHON_EXECUTABLE=/usr/bin/python3
```

and observe that the correct boost python library has been found.

No issue number.

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/PythonInterface/CMakeLists.txt
```diff
@@ -8,9 +8,38 @@
...
 ###########################################################################
 set ( old_boost ${Boost_LIBRARIES} )
-
 set ( Boost_LIBRARIES ) # Empty out the variable after previous use
-find_package ( Boost REQUIRED python )
+
+# We need to find the right boost python version for the given python version. Unfortunately
+# there is not standard naming convention across the platforms, for example:
+#  - fedora, arch: libboost_python.so (python2), libboost_python3.so (python3)
+#  - debian, ubuntu: libboost_python.so (python2), libboost_python34.so (python 3.4)
```
rosswhitfield: The code is correct but `libboost_python-py34.so` not `libboost_python34.so`.

martyngigg: Nice catch. I'll fix that.

buildconfig/CMake/FindSphinx.cmake
```diff
@@ -12,7 +12,7 @@
 #=============================================================
 
 find_program( SPHINX_EXECUTABLE NAME sphinx-build.cmd sphinx-build
-  PATHS ${CMAKE_LIBRARY_PATH}/python2.7/Scripts
+  PATHS ${CMAKE_LIBRARY_PATH}/python${PYTHON_MAJOR_VERSION}.${PYTHON_MINOR_VERSION}/Scripts
```
rosswhitfield: I don't think this works. The problem is the executable name is different, also depending on OS. I see `sphinx-build-3` on Fedora, `sphinx-build3` on Arch, and for Ubuntu it's located at `/usr/share/sphinx/scripts/python3/sphinx-build`.

But this is a different issues that can be fixed at another time.

martyngigg: Oh joy.

I think the PATHS are in addition to the standard system paths as that shouldn't work for Ubuntu now. Hopefully we can just detect whether we are using `3` and then have a list of known names for that. As you say, I'll leave that for another day.

rosswhitfield:
It works. 👍 
On Ubuntu 14.04 it found `/usr/lib/x86_64-linux-gnu/libboost_python-py34.so`
On Ubuntu 16.04 it found `/usr/lib/x86_64-linux-gnu/libboost_python-py35.so`
On Fedora 23 it found `/usr/lib64/libboost_python3.so`
On Arch it found `/usr/lib64/libboost_python3.so`
:shipit:

PR #16382, opened 2016-05-24
Remove locking mechanism from ISpectrum.

### LINKED ISSUE #16378: EventWorkspaceMRU can potentially leak memory.
```cpp
void EventWorkspaceMRU::clear() {
  std::lock_guard<std::mutex> _lock(this->m_toDeleteMutex);

  // FIXME: don't clear the locked ones!
  for (auto &marker : m_markersToDelete)
    if (!marker->m_locked)
      delete marker;
  m_markersToDelete.clear();
```

The MRU contains raw pointers and after the `clear()` call there is no code that deals with the clean up. Furthermore, this mechanism that was made for thread-safety is therefore actually broken: Thread X may clear the MRU and thus invalidates references that thread Y is holding.

On the plus side, the MRU does not seem to be locked in many places, and in those it actually does not seem be be necessary at all.

Suggested fix: Remove the locking mechanism.

Note that with the changes introduced as part of https://github.com/mantidproject/mantid/issues/16282 locking will also become less useful: Clients can simply use `counts()` which returns by-value instead of a potentially invalidated reference from `readY()`.

SimonHeybrock:
See #16378 for a description.

**To test:**

Code review, try to understand the MRU and locking mechanism thoroughly. In particular, try to verify that my understanding and reasoning is correct:
- Each thread has its own MRU.
- Therefore, locking is only needed if there is a risk that a particular thread kicks its own buffered data from the MRU, i.e., in code like this:
  
  ``` cpp
  auto &y0 = eventWS->readY(0);
  for(size_t i=0; i<eventWS->numberOfHistograms(); ++i) {
    auto &y = eventWS->readY(i);
    // y0 becomes an invalid reference after 50 iterations,
    // if we need it we would need to lock
  }
  ```
- Code like this does not seem to be present anywhere in the cases where `ISpectrum::lockData()` is used currently, so we can drop it completely.

Please also see below for my two comments regarding Poco threading.

Fixes #16378.

_Internal change, does not need to be in the release notes._

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/API/inc/MantidAPI/ISpectrum.h
+++ b/Framework/API/inc/MantidAPI/ISpectrum.h
@@ -115,11 +115,6 @@ public:
 
   void setSpectrumNo(specnum_t num);
 
-  // ---------------------------------------------------------
-  virtual void lockData() const;
-  virtual void unlockData() const;
-
-  //-------------------------------------------------------
   virtual bool hasDx() const;
   virtual void resetHasDx();
 
--- a/Framework/API/src/ISpectrum.cpp
+++ b/Framework/API/src/ISpectrum.cpp
@@ -207,17 +207,6 @@ specnum_t ISpectrum::getSpectrumNo() const { return m_specNo; }
  * @param num :: the spectrum number of this spectrum */
 void ISpectrum::setSpectrumNo(specnum_t num) { m_specNo = num; }
 
-// ---------------------------------------------------------
-/** Lock access to the data so that it does not get deleted while reading.
- * Does nothing unless overridden.
- */
-void ISpectrum::lockData() const {}
-
-/** Unlock access to the data so that it can again get deleted.
- * Does nothing unless overridden.
- */
-void ISpectrum::unlockData() const {}
-
 //---------------------------------------------------------
 /**
  * Gets the value of the use flag.
--- a/Framework/Algorithms/src/SmoothNeighbours.cpp
+++ b/Framework/Algorithms/src/SmoothNeighbours.cpp
@@ -643,7 +643,6 @@ void SmoothNeighbours::execWorkspace2D() {
       double weightSquared = weight * weight;
 
       const auto &inSpec = inWS->getSpectrum(inWI);
-      inSpec.lockData();
       const MantidVec &inY = inSpec.readY();
       const MantidVec &inE = inSpec.readE();
       const MantidVec &inX = inSpec.readX();
@@ -663,8 +662,6 @@ void SmoothNeighbours::execWorkspace2D() {
       if (inWS->isHistogramData()) {
         outX[YLength] = inX[YLength];
       }
-
-      inSpec.unlockData();
     } //(each neighbour)
 
     // Now un-square the error, since we summed it in quadrature
--- a/Framework/Crystal/src/AnvredCorrection.cpp
+++ b/Framework/Crystal/src/AnvredCorrection.cpp
@@ -189,7 +189,6 @@ void AnvredCorrection::exec() {
 
     // Copy over bin boundaries
     const auto &inSpec = m_inputWS->getSpectrum(i);
-    inSpec.lockData(); // for MRU-related thread safety
 
     const MantidVec &Xin = inSpec.readX();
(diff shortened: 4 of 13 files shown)


SimonHeybrock:
A unit test uncovered an issue:
```cpp
const MantidVec &EventList::constDataY() const {
  // ...
  // This is the thread number from which this function was called.
  int thread = PARALLEL_THREAD_NUMBER;
  mru->ensureEnoughBuffersY(thread);
```
The comment is only true if OpenMP threads are used. `Kernel::ThreadPool` uses `Poco::Thread`, `thread` will always be `0` and thus all threads in the pool share the same MRU. As a consequence, `EventWorkspace::readY`, `dataY() const`, `readE()`, ... are *not* *thread-safe* in combination with `Poco::Thread`. The same probalby holds for any other threading mechanism apart from OpenMP.

SimonHeybrock:
Regarding the Poco threading issue:
- `ConvertToDiffractionMDWorkspace` is the only place where this is used.
- There is a v2 of the algorithm, the issue in only in v1.

Given that there is a newer version of the algorithm, one option would have been to remove threading. However, I decided to instead replace it with OpenMP threading.

I verified that the run-time of the algorithm does not change significantly. In particular, I temporarily added a test
```cpp
void test_MINITOPAZ_long() { do_test_MINITOPAZ(TOF, 10000, 10000); }
```
to `ConvertToDiffractionMDWorkspaceTest.h`. This took around 30s on my 12 core workstation, in both the original Poco implementation as well as the new OpenMP version. Since this is an outdated version of the algorithm I think it does not make sense to keep this test as a performance test, so I removed it from the final implementation.

martyngigg:
Is the take home message here that references to data from things like `readY()` should not be kept around for too long or they could become invalidated.

Is the reason that we don't require the locking because we mostly process on a spectrum-by-spectrum basis and only require the reference for short period of time?

SimonHeybrock:
@martyngigg Correct, references to `Y` and `E` of an `EventWorkspace` can be invalidated. It might be that the locking mechanism was created at some point of the development to deal with this, however, as you can see it was only used in a handful of places.

We do not require locking for several reasons:
1. What you say.
2. It is relevant only for `EventWorkspace`.
3. Many algorithms that deal with `EventWorkspace` do not use `Y` and `E`, but just work with the `EventList`.

The safest solution would be to do away with the MRU and keep `Y` and `E` once they were generated (with a flag that marks them as valid/invalid to trigger updates if the `EventList` was modified). However, this will increase memory consumption, so I do not not whether or not that makes sense. I would be in favor of removing the MRU, but I do not have enough information to judge whether or not that will create problems.

SimonHeybrock:
@martyngigg As a side note, the `Histogram` type I have been working on makes this slightly better, but does not solve in all cases. The new interface has functions that return by value (objects containing `cow_ptr`), so if it is deleted from the MRU the returned object stays valid. However, we still have an interface equivalent to the current one which still suffers from the same issue:

```cpp
class Histogram {
public:
  // ok, return cow object by value
  Counts counts() const;
  template <typename... T> void setCounts(T &&... data) & ;

  // not ok, references internal data
  const HistogramY &y() const { return *m_y; }
  HistogramY &mutableY() & { return m_y.access(); }

  // ok, return cow object by value
  Kernel::cow_ptr<HistogramY> sharedY() const { return m_y; }
  void setSharedY(const Kernel::cow_ptr<HistogramY> &y) & ;

  // ...
};
```

martyngigg:
I also don't have enough knowledge at this point to say if removing the MRU would lead to memory problems. We would probably have to gather together the widely uses cases and explore them more in depth to get an idea

martyngigg:
At this point I think developers should understand what obtaining a reference to something means and have them be aware that is not guaranteed to be valid indefinitely in these cases.

martyngigg:
I agree with Simon on the changes here and the code looks fine. 

The parallel code in `ConvertToDiffractionMDWorkspace` version 1 looks neater now with OpenMP loop rather than the more verbose thread tasks approach.

Happy to :shipit:

Issue #16402, opened 2016-05-25
Refactor MantidCurve, MantidMatrixCurve, MantidMDCurve


FedeMPouzols:
Have a look at MantidMDCurve and MantidMatrixCurve. There's a lot in common between them, with code duplicated in  `MantidMDCurve.cpp` and `MantidMatrixCurve.cpp`. Especially look at the method `init()` called from the constructors, and the destructor, but also other methods. Much of that code is about setting general style options and is not specific to MD or Matrix curves. It could be in MantidCurve/shared.

The fix for #16388 doesn't improve this.

Ideally the code should be reorganized to prevent further issues in the future.

Note that changes in these clases might require some significant amount of manual testing of the 2D curves.

PR #16403, opened 2016-05-25
PlotMD: respect show all error bars flag

### LINKED ISSUE #16388: PlotMD: Does not respect the show all error bars flag
Beta test bug.

### Expected behavior

All error bars should be shown if the appropriate option is checked

### Actual behavior

A plot with `plotMD` does not respect the state. 

```python
#Some problems with MD algorithms / workflow

#Create an MD workspace, file backed because my system might not have enough memory:

import os
import numpy as np
#

#Create projections workspace for later cutting
from mantid.api import Projection
# Create an identity projection
proj_id = Projection([1,1,0], [-1,1,0], [0,0,1])
proj_id.setType(0, 'r') 
proj_id.setType(1, 'r')
proj_id.setType(2, 'r') 
ws_id = proj_id.createWorkspace()

#Make MD mega file (like gen_sqw / accumulate_sqw)
psi_arr=np.arange(0.0,92.0,2)
run_arr=range(15052,15098)

input_runs = list()
psi = list()
gs = list()
gl = list()
#note that gl, gs etc are not mandtory inputs. If we do not specify them then they are set to zero.

input_runs = list()
psi = list()
gs = list()
gl = list()
for i in range(len(psi_arr)):
    source_file = 'map'+str(run_arr[i])+'_ei400.nxspe'
    input_runs.append(source_file)
    psi.append(float(psi_arr[i]))
    gl.append(0.0)
    gs.append(0.0)

print(psi)
print(input_runs)

md_ws = CreateMD(input_runs, Emode='Direct', Alatt=[2.87,2.87,2.87], Angdeg=[ 90, 90, 90], u=[1, 0, 0,], v=[0,1,0], Psi=psi, Gl=gl, Gs=gs, EFix=400.0,Filename='test.nxs',FileBackEnd=True)
#Up to this point everything works as we would expect it to.

##Problem with PlotMD errorbars
out_md_1d_nopix = CutMD(md_ws, Projection=ws_id, PBins=([-3,0.05,3],[-1.1,-0.9],[-0.1,0.1],[130,150]), NoPix=True)
plotMD(out_md_1d_nopix,error_bars=True,normalization=2)
#After fettling the axes limits, can see that not all errorbars are plotted. Resizing window up and down shows more / fewer errorbars (old behaviour)
#Checkbox for "show all errorbars" was checked (since on by default)
```

I have the data for this script.

### Platforms affected
All

FedeMPouzols:
Fixes plotMD (MantidMDCurve) so that it respects the MantidPlot flag for showing all error bars.

**To test:**
- Make sure the option "draw all errors" is on. With v3.7 the option is enabled by default once. If it has been disabled it can be re-enabled in View->Preferences->2D plots->Curves->"Draw all errors".
- Do a plotMD. You can use the script from the issue description #16388 to generate a large MD workspace. Contact me if you need the input files. This will take some minutes, but the small files from test data won't probably be enough to test this issue as none of them seem to have enough points.
- Before this fix the option "draw all errors" was ignored and not all error bars are shown if they would be too close to each other on the plot (first image). After this fix you should see a very dense cloud of error bars if you adjust the axes conveniently (second image, towards the right). A trick to make this easy to see is to change the layer (double clicking) to Line->Connect->Sticks. Then you should see that every stick gets its error bar after the fix, regardless of how close they are to each other.

plotmd_sparse_colorbars 

 plotmd_dense_colorbars

This fix is a code horror in that it goes further in duplicating code between MantidMDCurve and MantidMatrixCurve. I've done this as bigger changes would require much more testing and can be risky when being so close to the release.
I've opened an issue for the duplicate code: #16402.

Fixes #16388.

This issue is not new, but it had not been reported before and was found during beta testing. Maybe it _does not need to be in the release notes._

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidPlot/src/Mantid/ErrorBarSettings.h
+++ b/MantidPlot/src/Mantid/ErrorBarSettings.h
@@ -64,6 +64,7 @@ private:
 
   friend class MantidCurve;
   friend class MantidMatrixCurve;
+  friend class MantidMDCurve;
 };
 
 #endif
--- a/MantidPlot/src/Mantid/MantidMDCurve.cpp
+++ b/MantidPlot/src/Mantid/MantidMDCurve.cpp
@@ -4,6 +4,7 @@
 #include <qpainter.h>
 #include <qwt_symbol.h>
 #include "MantidAPI/AnalysisDataService.h"
+#include "ErrorBarSettings.h"
 #include "../Graph.h"
 #include "../ApplicationWindow.h"
 #include "../MultiLayer.h"
@@ -88,6 +89,16 @@ void MantidMDCurve::init(Graph* g, bool distr, Graph::CurveType style)
     setStyle(QwtPlotCurve::Lines);
   }
   g->insertCurve(this,lineWidth);
+
+
+  // set the option to draw all error bars from the global settings
+  if (hasErrorBars()) {
+    setErrorBars(true, g->multiLayer()->applicationWindow()->drawAllErrors);
+  }
+  // Initialise error bar colour to match curve colour
+  m_errorSettings->m_color = pen().color();
+  m_errorSettings->setWidth(pen().widthF());
+
   connect(g,SIGNAL(axisScaleChanged(int,bool)),this,SLOT(axisScaleChanged(int,bool)));
   observePostDelete();
   connect( this, SIGNAL(resetData(const QString&)), this, SLOT(dataReset(const QString&)) );


AndreiSavici:
Looks good. The plotting of error bars follow the preference in `View -> Preferences... -> 2DPlots -> Curves -> DrawAllErrors`
:shipit:

PR #16406, opened 2016-05-25
Support loading older muon data

### LINKED ISSUE #16396: Old muon data cannot be loaded into MuonAnalysis
### Expected behavior
Should be able to load old muon data into MuonAnalysis interface

### Actual behavior
Some old data (see below) cannot be loaded

Try to load the datafile `01360.NXS`, which has been converted to NeXus. It is from 1994 and many archived data files from this time have used the same NeXus converter.

Note the file *will* load into Mantid using `LoadMuonNexus`, it just won't load into the MuonAnalysis GUI.

### Platforms affected
All

tom-perkins:
Fix `LoadMuonNexus` to deal with certain older files, and enable MuonAnalysis to load old muon data

The file in the issue was an old muon file that has been converted to Nexus and wouldn't load in MuonAnalysis.

There were several related issues here:
- `LoadMuonNexus` (version 1) failed to load the dead times and detector groupings for muon files like the one in the issue. Fixed this to load them correctly.
- There is no IDF for the instrument this data was recorded on (DEVA). As a short-term fix to enable loading this data into the muon interface, added a special case to allow this. (In the long term this should be removed and an IDF added, if this instrument is to be supported).
- If the dead times option in the interface is set to "None", it shouldn't try to load dead times from the file.

**To test:**

The data file is at `\\olympic\babylon5\Scratch\TomPerkins\Data\old_muon\01360.NXS`.
- Try loading this file with `LoadMuonNexus`. Set workspace names for the _DeadTimeTable_ and _DetectorGroupingTable_ properties and check these are loaded ok (the dead times are all zero and there is only one group, but check they load without error).
- Open the MuonAnalysis interface and try loading the file into it. It should load ok and you should be able to fit the data.

Fixes #16396

_I haven't added this to the release notes_ because DEVA still is not a supported instrument. The ability to load this old data is useful for instrument scientists, but the workspace doesn't have an instrument and this might cause problems if other algorithms are run on it.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/DataHandling/src/LoadMuonNexus.cpp
+++ b/Framework/DataHandling/src/LoadMuonNexus.cpp
@@ -7,6 +7,7 @@
 #include "MantidKernel/ConfigService.h"
 #include "MantidKernel/ArrayProperty.h"
 #include "MantidAPI/FileProperty.h"
+#include "MantidGeometry/Instrument.h"
 #include "MantidGeometry/Instrument/Detector.h"
 #include "MantidAPI/Progress.h"
 #include "MantidAPI/TableRow.h"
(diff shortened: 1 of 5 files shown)


KarlPalmen:
I tested the functionality and loaded the file directly worked as described, but loading it in the Muon Analysis interface failed to give a plot.

KarlPalmen:
It worked OK after restarting Visual Studio.

KarlPalmen:
I tried again and found Mantid hit an uncaught exception when I then went to the Results Table.

tom-perkins:
I see the same - it's because the stored run number in the file is zero. I've pushed a fix to deal with this case and added a test.

KarlPalmen:
Worked OK :shipit:

FedeMPouzols:
Good fix, and with a test for it.

Issue #16437, opened 2016-05-26
Input validation for ALC integration times


tom-perkins:
See #16393:

This issue is to add
- input validation for `PlotAsymmetryByLogValue` for integration time limits
- input validation for `MuonGroupDetectors` for making sure the supplied grouping is sensible (valid detector numbers)

tom-perkins:
This is a maintenance task to add `validateInputs` methods to two algorithms.

PR #16438, opened 2016-05-26
ALC: check for bad custom grouping before loading files

### LINKED ISSUE #16393: Check for bad grouping in ALC
*Reported by Steve Cottrell during beta testing for 3.7*

#### Expected behaviour
Should get an error in the ALC interface when using a custom grouping with detector numbers that don't appear in the instrument (e.g. 1-48, 49-96 if there are only 64 detectors)

#### Actual behaviour
No error until the data has already been loaded, which can take some time. It would be better to flag this up at the start.

- Use ALC interface to analyse (for example) EMU data (96 detectors). 
- Set a custom grouping using all these detectors, e.g. 1-48, 49-96.
- Then change to HIFI data (64 detectors), leaving the 96-detector grouping in place, and click "Load".   
  - No error shown until *after* the data has been loaded.

#### Suggested resolution
`PlotAsymmetryByLogValue` doesn't have a `validateInputs()` method - this would be a good place to put a check. The same check should be put in place for `MuonGroupDetectors`.

tom-perkins:
Added a check for bad custom grouping before loading files.

Previously, if a bad custom grouping (e.g. with detector numbers not present in the instrument) was supplied, no error was given until _after_ all the files had been loaded - which can take some time if there are a lot of files. 

Added a check before data is loaded and give a more understandable error message if the supplied grouping is bad. Added a unit test for this case.

The issue mentions validating integration times and validating inputs to `MuonGroupDetectors` - I will leave this to the next release and open an issue for it (#16437).

**To test:**
_Data available at \olympic\babylon5\Scratch\TomPerkins\Data\ALC_
- Open Interfaces/Muon/ALC
- Set first run to `HIFI00051636.nxs` and last run to `HIFI00051638.nxs`
- Set the grouping radio button to "Custom" and enter "1-48" in "Forward" and "49-96" in "Backward". (HIFI has only 64 detectors).
- Click "Load" and check you get an error explaining what went wrong.
  - The error should appear _before_ the data is loaded
- If you use "Auto" grouping or a sensible custom grouping with valid detector numbers, data should be loaded without error.

Fixes #16393.

_Does not need to be in the release notes_ - this only changes _when_ the error appears, there was still an error shown before. 

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Muon/ALCDataLoadingPresenter.h
+++ b/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Muon/ALCDataLoadingPresenter.h
@@ -84,6 +84,9 @@ namespace CustomInterfaces
     /// Start/stop watching directory
     void changeWatchState(bool watching);
 
+    /// Check custom grouping is sensible
+    bool checkCustomGrouping();
+
     /// View which the object works with
     IALCDataLoadingView* const m_view;
 
@@ -98,6 +101,9 @@ namespace CustomInterfaces
 
     /// ID of timer, if one is running
     int m_timerID;
+
+    /// Number of detectors for current first run
+    size_t m_numDetectors;
   };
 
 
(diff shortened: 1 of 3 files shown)


raquelalvarezbanos:
The error message is now shown before data is loaded. A unit test has been added, and code changes look good. :shipit:

PR #16441, opened 2016-05-26
Parallelize runpylint script

martyngigg:
The `run_pylint.py` script now has the option to run the checks in parallel. This is configured using the cmake variable `PYLINT_NTHREADS`, similarly to `cppcheck` 

**To test:**
- Run `cmake -DPYLINT_NTHREADS=X` and replace `X` with the required number
- Check the resource utilization and you should see X number of separate processes running.

On my machine the checks in serial took around 5 minutes. With `DPYLINT_NTHREADS=8` they completed in less than a minute.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/buildconfig/CMake/PylintSetup.cmake
+++ b/buildconfig/CMake/PylintSetup.cmake
@@ -41,8 +41,10 @@ if ( PYLINT_FOUND )
         scripts/test
         Testing/SystemTests/tests/analysis/reference
   )
-  set ( PYLINT_OUTPUT_DIR ${CMAKE_BINARY_DIR}/bin/Testing )
-  set ( PYLINT_NTHREADS 8 )
+
+  set ( PYLINT_OUTPUT_DIR "" CACHE STRING "Directory used to write the output from each pyint check." )
+  # Choose the number of threads used
+  set ( PYLINT_NTHREADS 8 CACHE STRING "Number of processes used for running pylint" )
 
   add_custom_target ( pylintcheck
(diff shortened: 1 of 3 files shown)


FedeMPouzols:
It works as it says. Here is a quick test of speed up, without any other significant workload on a debian machine with 4 cores available:

```
/usr/bin/time cmake --build . --target pylintcheck
```
- Without setting the parallel option (using its default of 8):
```
13.83user 25.78system 2:10.75elapsed 336%CPU (0avgtext+0avgdata 246920maxresident)k
```

- With -DPYLINT_NTHREADS=1:
```
332.45user 21.79system 6:43.42elapsed 87%CPU (0avgtext+0avgdata 247256maxresident)k
```

:+1:, I'll give the code a look (there seem to be interesting changes) and put the squirrel in a while.

FedeMPouzols:
:shipit:, this will save a few minutes in each pylint job. If only the system tests scaled as neatly as the pylint jobs!

martyngigg:
@FedeMPouzols The system tests are also run in separate processes so in theory would parallelize too. The complication there is that each test may start threads itself to use the available cores and then that would probably lead to things getting slower.

I guess one thing we could do is limit each process to half the number of cores on the system and run two at once?

FedeMPouzols:
Yes, I think there's not much scope for improvement with the system tests. Some of them can use all the cores, and many of them are probably contrained by I/O.

Issue #16387, opened 2016-05-24
SaveMD does not save workspace created with CreateMD & file-backed=true correctly


martyngigg:
### Expected behavior

Saving a file-backed MD workspace should produce a file that is identical to the file backend

### Actual behavior

The files are drastically different.

### Steps to reproduce the behavior

The following script should demonstrate the issue:

```python
#Some problems with MD algorithms / workflow

#Create an MD workspace, file backed because my system might not have enough memory:

import os
import numpy as np
#

#Create projections workspace for later cutting
from mantid.api import Projection
# Create an identity projection
proj_id = Projection([1,1,0], [-1,1,0], [0,0,1])
proj_id.setType(0, 'r') 
proj_id.setType(1, 'r')
proj_id.setType(2, 'r') 
ws_id = proj_id.createWorkspace()

#Make MD mega file (like gen_sqw / accumulate_sqw)
psi_arr=np.arange(0.0,92.0,2)
run_arr=range(15052,15098)

input_runs = list()
psi = list()
gs = list()
gl = list()
#note that gl, gs etc are not mandtory inputs. If we do not specify them then they are set to zero.

input_runs = list()
psi = list()
gs = list()
gl = list()
for i in range(len(psi_arr)):
    source_file = 'map'+str(run_arr[i])+'_ei400.nxspe'
    input_runs.append(source_file)
    psi.append(float(psi_arr[i]))
    gl.append(0.0)
    gs.append(0.0)

print(psi)
print(input_runs)

md_ws = CreateMD(input_runs, Emode='Direct', Alatt=[2.87,2.87,2.87], Angdeg=[ 90, 90, 90], u=[1, 0, 0,], v=[0,1,0], Psi=psi, Gl=gl, Gs=gs, EFix=400.0,Filename='test.nxs',FileBackEnd=True)
#Up to this point everything works as we would expect it to.

#Suppose I close my Mantid session. Rather than re-doing all of the above, I load the above file backed md workspace:
md_ws2=LoadMD('test.nxs',FileBackEnd=True)

#Then the following two operations should give identical results (MDHisto workspaces):
out_md2 = CutMD(md_ws, Projection=ws_id, PBins=([-3,0.05,3],[-3,0.05,3],[-0.1,0.1],[0,4,360]), NoPix=True)
[...]

PR #16468, opened 2016-05-27
Clarify conditions under which SaveMD does not function as expected

martyngigg:
Adds an error scenario when `SaveMD` is called on a file-backed workspace but the filename specified is different to the current file backend.

In theory this should work but at the moment the events are not saved properly and is a source of confusion to users so make it explicit that it does not work. The hint says just perform a copy outside of Mantid to achieve the say effect.

**To test:**

The following script demonstrates the new message:

``` python
nxspe = Load(Filename='CNCS_7860_coarse.nxspe')
md_ws = ConvertToMD(nxspe, QDimensions='Q3D', Q3DFrames='Q_sample', QConversionScales='HKL', OutputWorkspace='md_ws', Filename='file-backend.nxs', FileBackEnd=True)
SaveMD(md_ws, Filename='file-backend-different.nxs')
```

Refs #16387

---
#### Reviewer

##### Code Review
##### Functional Tests

Ultimately the code should be able to support saving a file-backed
workspace to a different file. For now we just warn the user that
this is not supported.
Refs #16387

### DIFF
--- a/Framework/MDAlgorithms/src/SaveMD.cpp
+++ b/Framework/MDAlgorithms/src/SaveMD.cpp
@@ -83,30 +83,41 @@ void SaveMD::init() {
  */
 template <typename MDE, size_t nd>
 void SaveMD::doSaveEvents(typename MDEventWorkspace<MDE, nd>::sptr ws) {
-  std::string filename = getPropertyValue("Filename");
-  bool update = getProperty("UpdateFileBackEnd");
-  bool MakeFileBacked = getProperty("MakeFileBacked");
-
-  bool wsIsFileBacked = ws->isFileBacked();
-  if (update && MakeFileBacked)
+  bool updateFileBackend = getProperty("UpdateFileBackEnd");
+  bool makeFileBackend = getProperty("MakeFileBacked");
+  if (updateFileBackend && makeFileBackend)
     throw std::invalid_argument(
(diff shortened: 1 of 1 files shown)


martyngigg:
I'm not sure what has happened to the ubuntu build. I'll restart it.

AntonPiccardoSelg:
Code looks fine.
Functional testing: Provides a clear error message when trying to save a file-backed workspace to a file which is not the at the backend. It gives the user a clear instruction what to do as a next step.
Please :shipit:

rosswhitfield:
I don't know why but this appears to cause some unit tests to randomly fail. If I run `ctest -j3 -R "LoadMD|SaveMD"` about 1 out of 5 times this fails on one or more of the tests. Can anyone else confirm?

SimonHeybrock:
@rosswhitfield Do you have the output of the failed tests?

rosswhitfield:
It looks like it's failing because the same name is used in LoadMD/SaveMD. `SaveMDTestHisto.nxs` and `SaveMD2TestHisto.nxs`. I just don't know why it didn't failed before.
[LoadMD_SaveMD_test.txt](https://github.com/mantidproject/mantid/files/291208/LoadMD_SaveMD_test.txt)

martyngigg:
The tests must have started overlapping for some reason. Shall I just change the names of those files? They really should be different.

PR #16472, opened 2016-05-27
Fix for reading string DASlogs

mdoucet:
`LoadNexusLogs` allows for string logs, but the current loader doesn't load them correctly.

The content of a string DASlog looks like this:

```
  NX Data  : device_id (NX_UINT32)
  NX Data  : device_name[15] (NX_CHAR)
  NX Data  : length[3] (NX_UINT32)
  NX Data  : offset[3] (NX_UINT32)
  NX Data  : time[3] (NX_FLOAT64)
  NX Data  : value[47] (NX_CHAR)
```

where `value` is the concatenation of all the string values (one for each time value). `offset` is the start position of each value within the concatenated string.

To read the string values, one must first read the `value` entry, parse the `offset` entry, then loop through the offset entries to extract the sub-string corresponding to each time value.

**To test:**

Load a file that contains string logs and make sure they are read properly.

There is not GitHub 'issue' for this PR.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/DataHandling/src/LoadNexusLogs.cpp
+++ b/Framework/DataHandling/src/LoadNexusLogs.cpp
@@ -552,7 +552,7 @@ LoadNexusLogs::createTimeSeries(::NeXus::File &file,
     std::transform(time_double.begin(), time_double.end(), time_double.begin(),
                    std::bind2nd(std::multiplies<double>(), 60.0));
   }
-  
+
   // String logs use offsets to cut the 'value' string into strings
   // corresponding to each individual time value. Get the offset now
   // before we have to enter the 'value' element.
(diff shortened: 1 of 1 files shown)


NickDraper:
@mdoucet  If this wants to make R3.7 then it needs to test clear today.  Otherwise we will have to move this to R3.8

mdoucet:
It turns out that string logs stored as 2D arrays were used in system tests. So there was precedent for storing them this way, and this was not an oversight. Instead of coding logic for two different ways of storing those logs, we decided to change the way those will be stored at SNS going forward. Those had never been used here. Since the Nexus format doesn't really address this issue, we decided to go with what already existed. I'm simply closing this PR.

PR #16477, opened 2016-05-30
Fixed cow_ptr move constructor and assignment

### LINKED ISSUE #16476: Kernel::cow_ptr is not movable
In `Kernel::cow_ptr` we have the move constructor and move assignment set as `= default`. However, the compiler does not generate them. The reason seems to be the `std::mutex` member.

Since `cow_ptr` wraps a `shared_ptr` the negative impact of this is probably rather small, but it should be cleaned up nevertheless.

SimonHeybrock:
Defaulted (compiler generated) move of `Kernel::cow_ptr` did not do what expected due to a non-movable class member. In this commit we therefore provide a custom implementation.

**To test:**

Code review & make sure the tests catch the regression if the bug is reintroduced.

Fixes #16476 .

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Kernel/inc/MantidKernel/cow_ptr.h
+++ b/Framework/Kernel/inc/MantidKernel/cow_ptr.h
@@ -72,10 +72,10 @@ public:
   constexpr cow_ptr(std::nullptr_t) : Data(nullptr) {}
   cow_ptr(const cow_ptr<DataType> &);
   // Move is hand-written, since std::mutex member prevents auto-generation.
-  cow_ptr(cow_ptr<DataType> &&other) : Data(std::move(other.Data)) {}
+  cow_ptr(cow_ptr<DataType> &&other) noexcept : Data(std::move(other.Data)) {}
   cow_ptr<DataType> &operator=(const cow_ptr<DataType> &);
   // Move is hand-written, since std::mutex member prevents auto-generation.
-  cow_ptr<DataType> &operator=(cow_ptr<DataType> &&rhs) {
+  cow_ptr<DataType> &operator=(cow_ptr<DataType> &&rhs) noexcept {
     Data = std::move(rhs.Data);
     return *this;
   }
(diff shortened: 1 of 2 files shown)


Framework/Kernel/inc/MantidKernel/cow_ptr.h
```diff
@@ -71,9 +71,14 @@ template <typename DataType> class cow_ptr {
   /// Constructs a cow_ptr with no managed object, i.e. empty cow_ptr.
   constexpr cow_ptr(std::nullptr_t) : Data(nullptr) {}
   cow_ptr(const cow_ptr<DataType> &);
-  cow_ptr(cow_ptr<DataType> &&) = default;
+  // Move is hand-written, since std::mutex member prevents auto-generation.
+  cow_ptr(cow_ptr<DataType> &&other) : Data(std::move(other.Data)) {}
```
quantumsteve: Can this be marked `noexcept`?

Framework/Kernel/inc/MantidKernel/cow_ptr.h
```diff
@@ -71,9 +71,14 @@ template <typename DataType> class cow_ptr {
...
   cow_ptr<DataType> &operator=(const cow_ptr<DataType> &);
-  cow_ptr<DataType> &operator=(cow_ptr<DataType> &&) = default;
+  // Move is hand-written, since std::mutex member prevents auto-generation.
+  cow_ptr<DataType> &operator=(cow_ptr<DataType> &&rhs) {
```
quantumsteve: Can this be marked `noexcept`?

SimonHeybrock:
@quantumsteve Yes you are right -- `shared_ptr` moves are `noexcept`, so I added it also here.

quantumsteve:
The move constructor and move assignment operator LGTM and two new tests verify that it functions correctly. We want a unique mutex per cow_ptr, so keeping it out of the member initializer list and relying on the default constructor makes sense.

:shipit:

FedeMPouzols:
I think this may be the second place in the code where we have `noexcept`! This may be a good area for a bit of modernization/maintenance.

quantumsteve:
@FedeMPouzols clang-tidy checks that move constructors and move assignment operators are marked `noexcept` and suggests 9 more places where we could use this keyword.

http://clang.llvm.org/extra/clang-tidy/checks/misc-noexcept-move-constructor.html

PR #16489, opened 2016-05-31
Remove API::MemoryManager

### LINKED ISSUE #16487: Remove MemoryManager
Quoting @martyngigg:
"MemoryManager used to be used for checking the available memory, primarily when we had the managed workspaces. It then gained powers to call tcmalloc functions when we compiled against that. As we don't  directly compile against it anymore I'm not certain that it needs to be around any more."

Most methods of `MemoryManager` do nothing, so the places where they are called are just clutter.

SimonHeybrock:
MemoryManager is a historic artefact and does nothing in the current implementation, so this PR removes it.

**To test:**

Code review & automatic tests should be sufficient.

Fixes #16487.

_Internal change, does not need to be in the release notes._
(There is a change of one menu item in MantidPlot, however it was compiled only when a certain TCMALLOC define was set, and I think it was thus not visible in any recent Manti release anyways).

---
#### Reviewer

##### Code Review
##### Functional Tests


FedeMPouzols:
All looks good and removing this makes perfect sense, as per discussions on slack.
This removes code that would give the very wrong impression that memory is being "released".

I had never seen that action "Release Free Memory" in . We have the "Clear All Memory" which is a different thing.

FedeMPouzols:
:shipit:

Issue #16464, opened 2016-05-27
Multi-dataset fitting: more beta bugs


tom-perkins:
*Reported by James Lord during beta testing*

- [ ] Set a parameter as *both* global and fixed. After fitting once, it loses the tie and is no longer fixed if you run the fit a second time --> see #16503
  - in general, ties and global flags can sometimes be lost after running several fits
- [x] Precision in "Edit parameter values": --> see #16492
  - Enter (e.g.) "20001.2345" and select "Fix" from the button on the right -- it chops off one decimal place
  - Enter an integer "23456" and do the same thing, it cuts off the last digit (2345)!
  - Values entered are rounded to fewer decimal places for display and the rounded values are used in the fit
- [x] "Add Workspace" dialog: workspaces are in random order, hard to find the one you want if there are a lot --> see #16462, #16496
- [ ] When a new workspace is added to the data selector, it should scroll to the bottom to show the new one added (makes it less likely to add one twice or skip one) --> see #16502
- [x] ~~Curves in graph: left/right go in the wrong order~~ (fixed in #16461)

tom-perkins:
The plot guess curve (purple) can sometimes appear and disappear:
> After doing a number of fits I get the data, red fit curve (which looks OK), and green difference (0+-noise), however on stepping through the spectra a purple curve appears, changes a few times, and disappears again. These curves don't fit the data well at all, but could perhaps be the remains of earlier guesses or fit iterations. For any given data set it's the same set of purple spectra each time I step to it. What's going on?

>(This seems to only happen when I have "Plot guess" ticked.)

tom-perkins:
The remaining issues here have been separated out into #16502, #16503 and so this issue can be closed.

PR #16492, opened 2016-05-31
Multi-dataset fit "edit parameter values": fix parameter truncation bug

tom-perkins:
Fix a bug in the "Edit parameter values" dialog in the multi-dataset fitting interface.

Previously, typing a value into the box and then clicking "fix" or another option, without first pressing return, would result in the final digit of the entered value being truncated. For example, "23456" would become "2345" and "20001.2345" would become "20001.234".

Ensure that, when user edits value, the internally stored value is kept up to date so that the correct value is used.

**To test:**
- Load a small number of workspaces
- Use "Add workspace" to add your workspaces
- Add a fit function. Click in one of the parameter values to edit it, and use this button to open the "edit parameter values" window:

editparamvalues
- In the dialog, type a number into one of the value boxes and then select "Fix" from the drop-down list to the right.
  - The value you entered should _not_ have its final digit chopped off
  - Try this with an integer (23456) and a decimal (20001.2345)
  - Try other options besides "fix"
  - Also try fixing with the Ctrl+F shortcut

Re #16464 - **does not** close this issue because there are other bugs contained in it.

[Release notes](https://github.com/mantidproject/mantid/blob/fc0cfdb350752ba5b0f4c6374a2385712f2c9d5c/docs/source/release/v3.7.0/ui.rst#bugs-resolved) have been updated.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/MultiDatasetFit/MDFLocalParameterEditor.h
+++ b/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/MultiDatasetFit/MDFLocalParameterEditor.h
@@ -40,6 +40,8 @@ private slots:
   void removeTie();
   void setTieAll();
   void removeAllTies();
+  void updateValue(const QString &value);
+
 private:
   bool eventFilter(QObject *widget, QEvent *evn) override;
   void setEditorState();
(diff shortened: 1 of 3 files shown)


ElliotAOram:
* Used all spectra in `irs26176_graphite002_red.nxs` as test data 
* Convolution of a linear background and Lorentzian as the function
* Set value to 0.00026 
* Tried `"Set to all", "Fix", "Fix all" and "Set tie"` none of these truncated the data

If trailing zeros are left in the data (e.g. `0.0120000`) these remain when you click off the field that it was entered into but are removed when the value is reselected. This is a very minor point and doesn't effect the functionality of this change. Code review and functionality looks good :shipit:

PR #16496, opened 2016-05-31
Sort names for workspace selection

### LINKED ISSUE #16462: Workspace names in random order
*Reported by James Lord in beta test for 3.7 in relation to multi-dataset fitting*

When selecting workspaces from a drop-down list, they appear in no particular order (not alphanumeric and not the order they were added in).
This is the same for a generic algorithm dialog and for the multi dataset fitting interface.

This makes it difficult to find and select the workspace you want if there are a lot of workspaces in the ADS - particularly when adding workspaces in the multi dataset fitting tool!

Example:
1. Generate workspaces:
``` python
a1 = CreateSampleWorkspace()
a2 = CreateSampleWorkspace()
a3 = CreateSampleWorkspace()
a4 = CreateSampleWorkspace()
b1 = CreateSampleWorkspace()
b2 = CreateSampleWorkspace()
b3 = CreateSampleWorkspace()
b4 = CreateSampleWorkspace()
c1 = CreateSampleWorkspace()
c2 = CreateSampleWorkspace()
c3 = CreateSampleWorkspace()
c4 = CreateSampleWorkspace()
```

2. Multi-dataset fitting:
mdf_outoforder

3. Generic algorithm dialog:
algoutoforder

---

If I generate workspaces called INST0001, INST0002 etc, they do (sometimes?) appear in order, but I think this might just be by chance. I think the ADS stores workspaces in an unordered container so they are not sorted when you get the names.

tom-perkins:
When presenting a drop-down list of workspaces to choose from in:
- generic algorithm dialog 
- multi-dataset fitting "add workspace" window,

ensure the list of workspaces is sorted. This makes it much easier to find the workspace(s) you want to select, especially when the ADS contains a lot of entries. (Point raised by beta testers)

Achieved this by sorting the list before populating the combo box. The alternative would be to use a `std::set` instead of a `std::unordered_set` as the return type for `DataService::getObjectNames()`, but this could have unintended effects elsewhere, especially this close to the release.

**To test:**
- Populate the ADS with a variety of workspaces with different names (see issue for example)
- Open an algorithm dialog for an algorithm that takes an _InputWorkspace_ and has a generic dialog (for example, `Rebin`). Look at the dropdown list for the workspace property and check the workspaces are sorted and easy to find.
- Open _Interfaces/General/Multi dataset fitting_ and click "Add workspace". In the dialog, check the dropdown list here too and make sure the workspaces are sorted and easy to find here.

Fixes #16462.

_Does not need to be in the release notes_ - too minor

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/API/inc/MantidAPI/WorkspaceProperty.h
+++ b/Framework/API/inc/MantidAPI/WorkspaceProperty.h
@@ -338,7 +338,9 @@ public:
         } else
           ++it;
       }
-      return std::vector<std::string>(vals.begin(), vals.end());
+      auto values = std::vector<std::string>(vals.begin(), vals.end());
+      std::sort(values.begin(), values.end());
+      return values;
     } else {
       // For output workspaces, just return an empty set
       return std::vector<std::string>();
--- a/MantidQt/CustomInterfaces/src/MultiDatasetFit/MDFAddWorkspaceDialog.cpp
+++ b/MantidQt/CustomInterfaces/src/MultiDatasetFit/MDFAddWorkspaceDialog.cpp
@@ -32,6 +32,7 @@ AddWorkspaceDialog::AddWorkspaceDialog(QWidget *parent):QDialog(parent),m_maxInd
       workspaceNames << QString::fromStdString( *name );
     }
   }
+  workspaceNames.sort();
   connect(m_uiForm.cbWorkspaceName,SIGNAL(currentIndexChanged(const QString&)),this,SLOT(workspaceNameChanged(const QString&)));
   m_uiForm.cbWorkspaceName->addItems( workspaceNames );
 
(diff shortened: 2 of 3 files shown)


AndreiSavici:
Works as described. :shipit:

FedeMPouzols:
This fixes the beta issue raised by users, and it is a great usability improvement. 

For the next release I'd really consider changing `getObjectNames()`. Probably performance is not an issue, but having to sort the list again feels untidy.
I'd need to check, but I think that the workspaces window also shows the workspaces sorted alphabetically. I'd guess there's another `sort()` there.

tom-perkins:
I've created a maintenance issue for changing `getObjectNames()`: #16500

PR #16497, opened 2016-05-31
Remove possibly dangerous exit() calls

martyngigg:
I came across a few calls to `exit()`, which if hit would cause the process to exit immediately without any explanation. I have updated these calls to what I think are more appropriate calls.

Note that I have not removed all calls to any function called `exit` as some are legitmate and some are not the `exit` function but an `exit` method on the Qt `Thread` class.

**To test:**

A code review should be sufficient.

No issue number.

---
#### Reviewer

##### Code Review
##### Functional Tests

Avoid calls to exit that could cause the program to just stop without
any warning.


Framework/Geometry/test/ObjectTest.h
```diff
@@ -1115,10 +1115,7 @@ class ObjectTest : public CxxTest::TestSuite {
     // Note that the testObject now manages the "new Plane"
     for (auto vc = SurfLine.cbegin(); vc != SurfLine.cend(); ++vc) {
       auto A = Geometry::SurfaceFactory::Instance()->processLine(vc->second);
-      if (!A) {
-        std::cerr << "Failed to process line " << vc->second << '\n';
-        exit(1);
-      }
+      assert(A);
```
FedeMPouzols: Would this be better off as a `TSM_ASSERT`?

martyngigg: You're right, it would be clearer when it failed. I'll update it.

MantidPlot/src/nrutil.cpp
```diff
@@ -5,24 +5,28 @@
...
 
-void nrerror(
-    const char *error_text) { // Numerical Recipes standard error handler
-  fprintf(stderr, "Numerical Recipes run-time error...\n");
-  fprintf(stderr, "%s\n", error_text);
-  fprintf(stderr, "...now exiting to system...\n");
-  exit(1);
-}
+// Numerical Recipes standard error handler
+void nrerror(const char *error_text) { LOGGER().error(error_text); }
```
FedeMPouzols: Oh so we had `fprintf`s from the old "good" numerical recipes code. Not really the topic of this PR, but there are several places in Framework (including Kernel and Geometry) where we also have `cout`s and `cerr`s. And we have plenty of commented out `cout`s and `cerr`s which should be removed and or replaced with Mantid logs or `TS_TRACE` or similar.
Maybe we could add a maintenance/easy issue for that?

Framework/Geometry/src/Objects/Object.cpp
```diff
@@ -725,15 +725,10 @@ int Object::procString(const std::string &Line) {
...
-
-  if (RuleList.size() != 1) {
-    std::cerr << "Map size not equal to 1 == " << RuleList.size() << '\n';
-    std::cerr << "Error Object::ProcString : " << Ln << '\n';
-    exit(1);
-    return 0;
+  while (procPair(Ln, RuleList, nullInt)) {
   }
+
+  assert(RuleList.size() == 1);
```
FedeMPouzols: So I understand that it is acceptable to move on even if the list size is not 1?

martyngigg: The logic of the code should mean that RuleList is always size 1 by this point. On second thoughts I'm thinking that this should probably throw rather than use an assert. I'll update this too.

FedeMPouzols:
@martyngigg: what do you think about this `exit()`: https://github.com/mantidproject/mantid/blob/master/Framework/Kernel/src/ANN/ANN.cpp#L172?
From a quick grep it looks like this code is only used from Geometry? In any case, an exception would be nicer here too.

martyngigg:
That's not great either - my grep missed it. I agree an exception would be better here too

martyngigg:
@FedeMPouzols I have updated those issues mentioned with the exception of the `ANN` code. It's third party code and the function containing the `exit(1)` call is never actually used so I'd prefer to leave it alone.

FedeMPouzols:
Good code improvement. This replaces with exceptions a few `exit()` calls that were hidden in `Geometry`.
It also replaces `fprintf` calls with Mantid log messages in MantidPlot when they use the numerical recipes routines.

:shipit:

PR #16505, opened 2016-06-01
Multi-dataset fitting: fix bug where global parameters lose their ties

### LINKED ISSUE #16503: Multi-dataset fitting: global parameter loses ties on fitting
*Created from #16464*

#### Expected behaviour
Ties should be retained when a fit is run

#### Actual behaviour
If a global parameter is fixed (or has ties), it is no longer fixed after a fit is run - the tie is lost.

Set a parameter as *both* global and fixed. Run the fit. The tie will have disappeared and, if you run the fit again, the parameter will not be fixed. (Local parameters do not lose their ties).

tom-perkins:
Fixed a bug in multi-dataset interface where global parameters lost their ties/fixes when a fit was run or the displayed dataset was changed.

When the current dataset is changed, all tie properties were removed, but only those for local parameters were replaced. By definition, global parameters are shared across all datasets so their ties shouldn't have be removed in the first place.

**To test:**
- Generate some data

``` python
import numpy as np

def func(x, a, b):
    return 0.2 + a*x + b

for i in range(1, 4):
    xData = []
    yData = []
    eData = []
    for x in np.arange(0.1, 10, 0.1):
        xData.append(x)
        yData.append(func(x, i, i/2.0))
        eData.append(0.1)

    outname = 'ws' + str(i)    
    CreateWorkspace(DataX=xData, DataY=yData, DataE=eData, NSpec=1, OutputWorkspace=outname)
```
- Add the workspaces you just generated
- Set up a function of _FlatBackground + LinearBackground_.
  - Set _A0_ for the flat background to 0.2 and check the box to set it global
  - Right-click and fix it as well
- Use the "<" and ">" buttons to change which dataset is displayed. 
  - The tie on _f0.A0_ should not disappear
- Run the fit
  - The tie on _f0.A0_ should not disappear

Fixes #16503.

[Release notes](https://github.com/mantidproject/mantid/blob/7e502eea91d0c3501635bce08cb999dc8d5626ee/docs/source/release/v3.7.0/ui.rst#bugs-resolved) have been updated - if this is retargeted at 3.8 then this will need to be removed from the 3.7 release notes.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidQt/MantidWidgets/src/FunctionBrowser.cpp
+++ b/MantidQt/MantidWidgets/src/FunctionBrowser.cpp
@@ -1766,9 +1766,12 @@ void FunctionBrowser::removeTie()
   }
 }
 
-/// Remove all tie properties
-void FunctionBrowser::removeAllTieProperties()
-{
+/**
+ * Remove all tie properties from local parameters
+ * (Called when changing datasets)
+ * Ties on global parameters are left as they are.
+ */
+void FunctionBrowser::removeAllLocalTieProperties() {
   QList<QtProperty*> tieProperties;
   for (auto p : m_properties)
   {
@@ -1777,9 +1780,11 @@ void FunctionBrowser::removeAllTieProperties()
       tieProperties << p.prop;
     }
   }
-  for (auto prop : tieProperties)
-  {
-    removeProperty(prop);
+  for (auto prop : tieProperties) {
+    if (prop->hasOption(globalOptionName) &&
+        !prop->checkOption(globalOptionName)) {
+      removeProperty(prop);
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


NickDraper:
Works well
:shipit:

MantidQt/MantidWidgets/src/FunctionBrowser.cpp
```diff
@@ -1777,9 +1780,11 @@ void FunctionBrowser::removeAllTieProperties()
       tieProperties << p.prop;
     }
   }
-  for (auto prop : tieProperties)
-  {
-    removeProperty(prop);
+  for (auto prop : tieProperties) {
```
FedeMPouzols: Maybe this could have been an `auto &` or even a `const auto &`?

PR #16528, opened 2016-06-03
Add sanity check on std::vector::shrink_to_fit.

quantumsteve:
Description of work.

C++11 adds `std::vector::shrink_to_fit()`, which is a non-binding request to reduce to the container's capacity to its size. Currently this is being done in Mantid using a copy and a swap (shrink-to-fit idiom).

Since the standard doesn't guarantee that every implementation reduces the capacity of `std::vector`, it is useful to have a test verifying that it does as we expect and reduces the capacity by at least as much as copy-and-swap.

http://clang.llvm.org/extra/clang-tidy/checks/modernize-shrink-to-fit.html
https://groups.google.com/a/isocpp.org/forum/#!topic/std-discussion/ts9WlqHtMsc

**To test:**

This pull request has no associated issue number.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Kernel/CMakeLists.txt
+++ b/Framework/Kernel/CMakeLists.txt
@@ -384,7 +384,7 @@ set ( TEST_FILES
 	RebinHistogramTest.h
 	RebinParamsValidatorTest.h
 	RegexStringsTest.h
-        ShrinkToFitTest.h
+	ShrinkToFitTest.h
 	SLSQPMinimizerTest.h
 	SobolSequenceTest.h
 	SpecialCoordinateSystemTest.h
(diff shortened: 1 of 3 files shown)


SimonHeybrock:
Looks good, as discussed on slack this is a first step for replacing the copy-and-swap idiom by shrink-to-fit. This test will serve as an early warning for the (unexpected) case that `std::vector::shrink_to_fit` does not shrink as expected. We would then need to consider changing back to copy-and-swap, but hopefully this will never happen.

:shipit:

FedeMPouzols:
Definitely, `shrink_to_fit` should be promoted. Maybe we could have some highglights/summary from these various modernization PRs in a Mantid review or similar?

PR #16541, opened 2016-06-03
use back() member function to get the last element of a container.

quantumsteve:
Description of work.

This uses [`std::vector::back()`](http://www.cplusplus.com/reference/vector/vector/back/) and [`std::string::back()`](http://www.cplusplus.com/reference/string/string/back/) to return a reference to the last element of a container instead of calculating `size() -1` and using `operator[]`.

The primary benefit is clearer code without compromising performance.

**To test:**

Review changes and check builds.

This is a small change with no associated issue number.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/API/inc/MantidAPI/BoxController.h
+++ b/Framework/API/inc/MantidAPI/BoxController.h
@@ -336,7 +336,7 @@ public:
    * */
   double getAverageDepth() const {
     double total = 0;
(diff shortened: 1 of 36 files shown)


raquelalvarezbanos:
Changes look good. `front` and `back` replace `vec[0]` and `vec[vec.size()-1]`. Also, `pop_back`, `std::min` and `std::max` were added to replace `resize(vec.size()-1)`, `if (localMin < min) min = localMin;` etc.

:shipit:

PR #16559, opened 2016-06-07
Consistent symbol policy

### LINKED ISSUE #15283: Set a consistent policy for symbol visibility on all platforms. Currently on MSVC hides symbols by default.
Currently on MSVC hides symbols by default.

Set [`CXX_VISIBILITY_PRESET`](https://cmake.org/cmake/help/v2.8.12/cmake.html#prop_tgt:LANG_VISIBILITY_PRESET) to `hidden` for gcc/clang and fix the builds.

AntonPiccardoSelg:
[pv_plugins_patch_for_hidden_symbol_visibility_in_gcc.txt](https://github.com/mantidproject/mantid/files/315792/pv_plugins_patch_for_hidden_symbol_visibility_in_gcc.txt)

This PR tires to create a consistent symbol visibility between `MSVC`, `gcc` and `clang` builds. The Windows have a hidden symbol visibility by default, whereas the `gcc` an `clang` builds are visible by default. 

**Items which had to be modified**
1. General `hidden` setting in `Cmake`
2. Redefinition of the export macros in `System.h` and the various `DllConfig.h` files (and `ANN.h`, etc.)
3. Explicit template instantiation was pushed into `Mantid::Kernel` namespace
4. The explicit template instantiations for the Singletons was modified to be `extern` when _imported_, see [here](https://support.microsoft.com/en-us/kb/168958)
   - Open question: Why does the implementation not break the ODR rule? It could be included several times within the same object file. --> need to understand this before merging!!!
5. Removed export macro from `typedef` (not clear why they were there in the first place)
6. Altered the position of the export macro for explicit template instantiations to be compatible with `gcc` (seems `MSVC` is more forgiving) => `template class MACRO classname`
7. fixed `Framework/DataObjects/src/generate_mdevent_declarations.py` which was not working any longer after `clang-format` had been applied to automatically-generated files. 
8. exported originally unexdported functions (in `Timer.h`) -- How did this ever work on `MSVC`?
9. Removed the export macro for `Matrix<int>`, `TimeSeriesProperty<Type> (in the .cpp only) and`NDArrayToVector` ( in .cpp onlyy), NDArrayTypeIndex -> ask me for details
10. Reordering of `MDEvent` and `MDLeanEvent` in order to avoid gcc warning. Same in `VMDBase`
11. export definitions in `qtpropertyBrowser`
12. Added export macro to `SignalBlocker.cpp` --> not fully clear yet why this was not an issue before
13. Visibility of `PeaksPalette` and `ADSWorkspaceProvider` had to be changed
14. We require a patch for `vtkPVPlugin.h` since this is used to automatically generated some code for our plugins. The visibility for the free function `pv_plugin_query_verification_data` needs to be changed, ie the `C_EXPORT` macro needs an addition for the non-win32 case.
15. Make `ScriptRepositoryImpl.h` visible
16. Add `CMP0063` to several packages including _GMock_ and _GTest_
17. Disable symbol export for clang

**To test:**
1. Very thorough code review
2. Perform unscripted testing. Ideally your tests should touch the Singletons.

Fixes #15283.

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/Kernel/src/Matrix.cpp
```diff
@@ -1613,7 +1613,7 @@ void fillFromStream(std::istream &is, Kernel::Matrix<T> &in,
 
 // Symbol definitions for common types
 template class MANTID_KERNEL_DLL Matrix<double>;
-template class MANTID_KERNEL_DLL Matrix<int>;
+template class Matrix<int>;
```
martyngigg: What's the reason that we don't have `MANTID_KERNEL_DLL` for the int definition?

Framework/Kernel/src/Matrix.cpp
```diff
@@ -1613,7 +1613,7 @@ void fillFromStream(std::istream &is, Kernel::Matrix<T> &in,
 
 // Symbol definitions for common types
 template class MANTID_KERNEL_DLL Matrix<double>;
```
martyngigg: Should be change `MANTID_KERNEL_DLL` to `DLLExport` here so that it's clear that this can only every be an export defintion? It's then also consistent with other cases.

Framework/Kernel/src/VMD.cpp
```diff
@@ -133,8 +117,24 @@ VMDBase<TYPE>::getNormalVector(const std::vector<VMDBase<TYPE>> &vectors) {
 }
 
 /// Instantiate VMDBase classes
-template MANTID_KERNEL_DLL class VMDBase<double>;
-template MANTID_KERNEL_DLL class VMDBase<float>;
+template class MANTID_KERNEL_DLL VMDBase<double>;
```
martyngigg: Should this also be `DLLExport` rather than the library-specific macro?

Framework/Kernel/src/Matrix.cpp
AntonPiccardoSelg: Having the ``MANTID_KERNEL_DLL` creates the warning: 
`warning: type attributes ignored after type is already defined [-Wattributes]
 template class MANTID_KERNEL_DLL Matrix<int>;`

I had similar warnings at other locations but could avoid them by rearranging the code. The reason for this seems to be the use of Matrix<int> in a template specialization: https://github.com/mantidproject/mantid/blob/master/Framework/Kernel/src/Matrix.cpp#L812

There are other parts of the code, where this had to be done:. `TimeSeriesProperty`, `NDArrayToVector` , `NDArrayTypeIndex`

Framework/Kernel/src/Matrix.cpp
AntonPiccardoSelg: I agree `MANTID_KERNEL_DLL` will be equivalent to `DLLExport` and should be changed. I am not sure though that we are consistent in general. We seem to have a mix of `DLL_Export` and library specific macros.

Framework/Kernel/src/Matrix.cpp
AntonPiccardoSelg: Performing `nm -C -D libMantidKernel.so | grep "Mantid::Kernel::Matrix<int>::"` on the Kernel library produces an equivalent output to `nm -C -D libMantidKernel.so | grep "Mantid::Kernel::Matrix<double>::"`

martyngigg: I see. It seems that the specialization of the class template for that method has implicitly defined the symbol and inherited the visibility from the class definition. I think it might be worth putting a comment in to that effect as on first glance it looks like something is missing.

Framework/ScriptRepository/inc/MantidScriptRepository/ScriptRepositoryImpl.h
```diff
@@ -12,6 +12,12 @@
 #else
 #define SCRIPT_DLL_EXPORT DLLImport
 #endif
+#elif defined(__GNUC__) && !defined(__clang__)
```
martyngigg: Can we move all of this to a `DllConfig.h` file to be consistent with everywhere else? This is all our own code and it would be better to be consistent.

CMakeLists.txt
```diff
@@ -17,6 +17,15 @@ if(POLICY CMP0053)
   cmake_policy(SET CMP0053 NEW)
 endif()
 
+if(POLICY CMP0063)
+  # The OLD behavior for this policy is to ignore the visibility properties
+  # for static libraries, object libraries, and executables without exports.
+  # The NEW behavior for this policy is to honor the visibility properties
+  # for all target types.
+  cmake_policy(SET CMP0063 NEW)
+endif()
```
quantumsteve: After PR #16646 is merged, this statement should be removed.

quantumsteve:
I added [this patch](https://github.com/mantidproject/paraview-build/blob/0875d240c3ad2cc5096d00cec540af06d00c84af/patches/hidden_symbol_visibility.patch) to the ParaView buildscript and updated the buildservers.

martyngigg:
Thanks, @quantumsteve. Does this mean we need to mail the devlist about requiring cmake 3.5?

quantumsteve:
@martyngigg Yes, we need to notify the devlist about CMake 3.5 and ParaView 5.1-RC1.

martyngigg:
I've reviewed the changes here. It is unclear in a couple of places why exports had to be added that weren't already required for MSVC, e.g. `Timer.h` (as mentioned in the original description). Nevertheless, what is there now seems to be correct as far as we understand it and all builders agree.

I'm happy to :shipit:

martyngigg:
@AntonPiccardoSelg It looks like removing the cmake policy has caused builds to fail but everything passed previous commit.

martyngigg:
I get warnings in my clean build without the cmake policy stuff...

Testing/Tools/gmock-1.7.0/CMakeLists.txt
```diff
@@ -38,7 +38,7 @@ endif()
 # ${gmock_BINARY_DIR}.
 # Language "C" is required for find_package(Threads).
 project(gmock CXX C)
-cmake_minimum_required(VERSION 2.6.2)
+cmake_minimum_required(VERSION 3.5)
```
quantumsteve: I don't see any benefit to changing the minimum version for gmock.

Testing/Tools/gmock-1.7.0/gtest/CMakeLists.txt
```diff
@@ -40,7 +40,7 @@ endif()
 # ${gtest_BINARY_DIR}.
 # Language "C" is required for find_package(Threads).
 project(gtest CXX C)
-cmake_minimum_required(VERSION 2.6.2)
+cmake_minimum_required(VERSION 3.5)
```
quantumsteve: I don't see any benefit to changing the minimum version for gtest.

Framework/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 # This is mainly here so you don't get a complaint when running cmake
-cmake_minimum_required (VERSION 2.8.12)
+cmake_minimum_required (VERSION 3.5)
```
quantumsteve: This should already be changed in [master](https://github.com/mantidproject/mantid/blob/master/Framework/CMakeLists.txt#L2)

AntonPiccardoSelg:
@quantumsteve  Either minimum requirement for cmake has to be set to 3.5 or CMP0063 needs to be applied to the gtest, else there are [warnings](http://builds.mantidproject.org/job/pull_requests-ubuntu/8143/consoleFull)

martyngigg:
As it's our embedded version I'm happy with leaving the gmock/gtest minimum version at 3.5 and :shipit: @quantumsteve Is that okay with you?

quantumsteve:
@ I'm ok with the changes to gmock and gtest. :Shipit:

PR #16563, opened 2016-06-07
Use '\n' instead of std::endl.

quantumsteve:
Description of work.

This removes most instances of `std::endl` from Mantid (See our discussion in the tips channel on Slack and [this video](http://articles.emptycrate.com/2016/04/18/stop_using_std_endl.html)). I tried appending `\n` to the end of a string literal and using  `'\n'` otherwise. 

**To test:**

I made most of these changes with find and replace, so they should be quite consistent and many potential issues will cause the build to fail. I would suggest reviewing until you are comfortable with the substitutions. The last few commits were done by hand and therefore more likely to contain errors. You may want to review them more carefully.

This PR has no associated GitHub issue

---
#### Reviewer

##### Code Review
##### Functional Tests


quantumsteve:
@martyngigg

Framework/CurveFitting/src/Algorithms/RefinePowderInstrumentParameters.cpp
```diff
@@ -295,8 +295,7 @@ void RefinePowderInstrumentParameters::fitInstrumentParameters() {
   for (size_t i = 0; i < m_dataWS->readX(0).size(); ++i)
     outss << m_dataWS->readX(0)[i] << "\t\t" << m_dataWS->readY(0)[i] << "\t\t"
           << m_dataWS->readE(0)[i] << '\n';
-  cout << "Input Peak Position Workspace To Fit: " << endl
-       << outss.str() << '\n';
+  cout << "Input Peak Position Workspace To Fit: \n" << outss.str() << '\n';
```
martyngigg: Should we get rid of any usages of  `cout`/`cerr` like this? They don't seem like a good idea in the middle of an algorithm.

martyngigg:
The changes look good. 

I notice now that we have a lot of places where we are using the the stream operators to log a single string literal. This would be far more efficient if it were passed directly to the appropriate `Logger` method, e.g `g_log.notice("My string literal")` rather than `g_log.notice() << "My string literal\n"`. Do you think this is worth addressing?

quantumsteve:
@martyngigg Yes, changes like that would be beneficial, but think it is outside the scope of this pull request. I added your comments and a few other ideas about logging to the maintenance tasks for another release.

martyngigg:
Agreed.

:shipit:

Issue #16562, opened 2016-06-07
MDNormSCD slows down when applied to workspace in a group


AndreiSavici:
### Expected behavior
For similar workspaces in a group, the time to run MDNormSCD on each workspace should be relatively constant

### Actual behavior
The time increases linearly with the number of workspaces in the group
![mdnorm](https://cloud.githubusercontent.com/assets/1128952/15873981/15643356-2ccf-11e6-96e2-05325d0b5dcc.png)

### Steps to reproduce the behavior
```
from mantid.simpleapi import *
import time
import numpy as np

N=60
w_in=Load("CNCS_7860")
w=DgsReduction(SampleInputWorkspace=w_in, SofPhiEIsDistribution='0',)
w=w[0]
SetUB(Workspace=w, a='4.48', b='4.48', c='10.8', u='1,0,0',v='0,0,1')
SetGoniometer(w,Axis0="0,0,1,0,1")
md=ConvertToMD(InputWorkspace=w, QDimensions='Q3D', Q3DFrames='HKL', QConversionScales='HKL', MinValues='-5,-5,-5,-5', MaxValues='5,5,5,5')
for i in range(N):
    SaveMD(md,Filename='/tmp/md_'+str(i)+'.nxs')

x=np.arange(2,60)
y=np.zeros(58)
for j in range(58):
    newN=j+2
    mdg=Load(','.join(['/tmp/md_'+str(i)+'.nxs' for i in range(newN)]))
    start=time.time()
    histoData,histoNorm=MDNormDirectSC(mdg[0],
                                           AlignedDim0="[H,0,0],-3,3,111",
                                           AlignedDim1="[0,0,L],-3,3,325",
                                           AlignedDim2="[0,K,0],-0.2,0.2,1",
                                           AlignedDim3="DeltaE,-1.5,1.5,1")
    y[j]=time.time()-start
    print "time for ",newN," group:", y[j]

tws=CreateWorkspace(x,y)
```
### Platforms affected
All

PR #16603, opened 2016-06-13
Speed up inputEnergyMode and add getPropertyValue to AlgHistory

### LINKED ISSUE #16591: Workspace history speedup
This issue is an in-depth look at the performance  problems seen in #16562. PR #16564 is a way to avoid the problem, but we need to find out why history lookup is slow. It seems related to finding the correct loader for files already loaded. The lookup time for a history of a workspace that was processed as part of the group exponentially increases with the number of workspaces in a group. This is unintuitive and has big performance penalty.

NickDraper:
This was a section of code `inputEnergyMode` that was slowing down as more complex worksaces were fed to it, and seemed to take too long anyway.

This came out to be that due to the special nature of he Load Algorithm when you reconsitute it by calling history.lastAlgorithm, It verifies and selects a loader for each file in the Filename property, and in this case there were several.

The main change here is to use the algorithm history objects directly rather than calling lastAlgorithm.  This takes the time to run that function from 500ms+ down to 3ms (windows debug timing).

To make the coding of this neater I have also added a getPropertyValue(name) method to the AlgorithmHistory class that operates like that of the Algorithm itself.

**To test:**

1. Code review
2. A unit test is covering the new getPropertyValue
3. Issue #16562 has a python snippet that runs MDNormDirectSC

Fixes #16591.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/API/src/AlgorithmHistory.cpp
+++ b/Framework/API/src/AlgorithmHistory.cpp
@@ -181,7 +181,7 @@ AlgorithmHistory_sptr AlgorithmHistory::operator[](const size_t index) const {
 */
 const std::string &
 AlgorithmHistory::getPropertyValue(const std::string &name) const {
-  for (auto &hist : m_properties) {
+  for (const auto &hist : m_properties) {
     if (hist->name() == name) {
       return hist->value();
     }
--- a/Framework/API/src/AlgorithmHistory.cpp
+++ b/Framework/API/src/AlgorithmHistory.cpp
@@ -148,7 +148,6 @@ size_t AlgorithmHistory::childHistorySize() const {
   return m_childHistories.size();
 }
 
-
 /**
   * Retrieve a child algorithm history by index
   * @param index ::  An index within the child algorithm history set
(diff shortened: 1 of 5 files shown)


Framework/API/src/AlgorithmHistory.cpp
```diff
@@ -174,6 +174,23 @@ AlgorithmHistory_sptr AlgorithmHistory::operator[](const size_t index) const {
...
+* @returns The string value of the property
+* @throw Exception::NotFoundError if the named property is unknown
+*/
+const std::string &
+AlgorithmHistory::getPropertyValue(const std::string &name) const {
+  for (auto &hist : m_properties) {
+    if (hist->name() == name) {
+      return hist->value();
+    }
+  }
```
quantumsteve: Is this the same as `std::find_if`?

NickDraper: For that I would need a unary fn that took the property element and compared it to the name, which I would have to pass in through a class member variable.

By the time I added that admittedly simple method, it would take more lines of code, and I think be harder to read.

http://www.cplusplus.com/reference/algorithm/find_if/

quantumsteve: https://github.com/mantidproject/mantid/compare/16591_workspace_history_speedup...16591_workspace_history_speedup_alt

NickDraper: Ah, I hadn't considered embedding an anonymous fn.

However, personally I think the original is easier to read (you may feel differently), and I would be suprised if there were any performance differences.

NickDraper: We have a slight disagreement on coding implementation, but it is not in violation of the coding standards, negative on performance, and 50/50 on readability.

If there isn't anything else could you pass this one please as it has been sat in limbo for almost 2 weeks.

quantumsteve:
Sorry for not responding quicker. This does substantially speed up a check and I am fine with the range-based for loop. 

Can you make the loop variable const? Other than that, I am happy to pass it.

quantumsteve:
LGTM :shipit:

PR #16606, opened 2016-06-13
Progress reporting for algs with group workspaces

### LINKED ISSUE #14249: Workspace groups: progress bar doesn't show any progress
This happens for example with `Rebin` or `SumSpectra`. If you run the algorithms on individual workspaces the progress bar works well. If you run the algorithms on a group the progress bar is created but it doesn't move at all.

For example any event and multi-period workspace should reproduce this issue. One particular example:
From \\isis\inst$\NDXENGINX\Instrument\data\cycle_12_4, ENGINX00197019.NXS

This was initially (and wrongly) identified as an issue with `RebinByPulseTimes` but it seems to happen in general with workspace groups.

We've seen this happening on linux and windows, with @LamarMoore .

@martyngigg : I just created this more general issue.

NickDraper:
Changed the group workspace handling of algorithms to route the dynamically created workspace progress reporting through the parent algorithm, correctly scaling on the way.

MergeRuns is a special case that is hadled seperately, but this has been done as well.

**To test:**
1. Run some algorithms with group workspaces.  They should work at least as well as the underlying algorithm reporting does.
2. look at the underlying issue, that has test python scripts for Rebin, and MergeRuns in the comments #14249 

Fixes #14249 .

RELEASE NOTES
docs/source/release/UI.rst

---
#### Reviewer

##### Code Review
##### Functional Tests


FedeMPouzols:
`MergeRunsTest.test_mixed_multiperiod_group_and_non_multiperiod_group_inputs_throws` didn't like this change. It is probably just a change of exception type that requires an update to the test.

NickDraper:
@FedeMPouzols Has this dropped of your radar?

FedeMPouzols:
Ops, sorry. My original plan was to test this together with some other fixes for groups but that got side-tracked. I'll give it a check by tomorrow.

Framework/API/src/MultiPeriodGroupWorker.cpp
```diff
@@ -203,12 +203,20 @@ bool MultiPeriodGroupWorker::processGroups(
...
-        sourceAlg->name(), sourceAlg->version());
+    // use create Child Algorithm that look like this one
+    Algorithm_sptr alg_sptr = sourceAlg->createChildAlgorithm(
+        sourceAlg->name(), progress_proportion * periodNumber,
+        progress_proportion * (1 + periodNumber), sourceAlg->isLogging(),
+        sourceAlg->version());
+    // Don't make the new algorithm a child so that it's workspaces are stored
+    // correctly
+    alg_sptr->setChild(false);
+    alg_sptr->setRethrows(true);
```
FedeMPouzols: I see the point in creating the algorithm with `createChildAlgorithm` so the behavior of the progress bar is adequate. But with this we have one more place with the funny sequence of events whereby a child algorithm is ripped off its childhood!
Would `setAlwaysStoreInADS(false)` serve the same purpose as `setChild(false)` and be a bit more clear? Or would there be any undesired difference? As far as I can see what `m_isChild` brings in addition to `m_alwaysStoreInADS` is that it:
- suppresses some logs
- disables workspace locking

NickDraper: In this case we create it as a child, using createChildAlgorithm and then revert it to a full algorithm using setChild(false).

This would seem clearer to me than createChildAlgorithm followed by setAlwaysStoreInADS(true).

docs/source/release/v3.8.0/ui.rst
```diff
@@ -29,10 +29,17 @@ Algorithm Toolbox
...
+##################
+
+- The progress reporting for algorithms has been improved, so that the progress is reported correctly when processin workspace groups or multi-period workspaces.
+
+
 Documentation
 #############
 * Added Ragged Workspace as a concept page
 
+>>>>>>> origin/master
```
FedeMPouzols: The last merge commit must have left this line here.

FedeMPouzols:
Works well. This fixes the progress bar when applying algorithms on a group. You can now see progress from 0 to 100% as it should be. This also fixes `MergeRuns` which is a special case implemented as a `MultiPeriodGroupAlgorithm`:

- tested with some multiperiod data rebin, and also with the example scripts found in the comments of the original issue.

I was going to write "squirrel" but I noticed there's a minor issue in the release notes file (see inline comment). @NickDraper: could you fix that? Then we don't need to wait for the builds again.

Framework/API/src/MultiPeriodGroupWorker.cpp
FedeMPouzols: That is fair enough. 

Now the code seems to be going in different directions in different places:
- We have `Algorithm` , `MultiPeriodGroupWorker` and one algorithm `ReflectometryReductionOneAuto`) that	use the `setChild(false)` trick as here, and then we also have [this place](https://github.com/mantidproject/mantid/blob/master/Framework/PythonInterface/mantid/api/src/Exports/BinaryOperations.cpp#L164-L166) with a `createUnmanaged` followed by a `setChild(false)`.
- And then we have `DataProcessorAlgorithm` and some 6 algorithms that prefer to use `setAlwaysStoreInADS()`.

NickDraper:
No problem I'll fix the release notes.

FedeMPouzols:
:shipit:
The last commit just fixes a leftover line in the release notes. No need to wait for the builds. See details [up there](https://github.com/mantidproject/mantid/pull/16606#issuecomment-228777317)

PR #16609, opened 2016-06-13
Fix CompressEventsTest on machines with > 50 threads.

quantumsteve:
Description of work.

`CompressEventsTest` currently fails on computers with > 50 threads. Because the first argument in `TS_ASSERT_DELTA` is unsigned, subtracting a number larger than 100 generates a very large positive number. Since CompressEvents currently [isn't run in parallel](https://github.com/mantidproject/mantid/blob/master/Framework/DataObjects/src/EventList.cpp#L1770), the simplest solution is to remove the if/else and just use the stricter check.

**To test:**

This PR has no associated issue.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/DataHandling/test/CompressEventsTest.h
+++ b/Framework/DataHandling/test/CompressEventsTest.h
@@ -74,12 +74,7 @@ public:
     }
 
     // Half the previous # of events
-    if (numPixels <= 1) {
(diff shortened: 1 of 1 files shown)


martyngigg:
This makes sense and the build servers agree. :shipit:

PR #16587, opened 2016-06-10
Move DataProcessorUI to its own directory

raquelalvarezbanos:
Moved all classes defining DataProcessor UI to a new directory under `MantidWidgets`.

**To test:**
- All tests should pass
- Code review (most changes are related to namespaces and includes, but note that I also had to modify `MantidQt/CustomInterfaces/test/CMakeLists.txt` and `MantidQt/MantidWidgets/test/CMakeLists.txt`to include additional dependencies and test helpers)
- Make sure you agree that `MantidWidgets` is the right location (in the [design document](https://github.com/mantidproject/documents/blob/master/Design/DataProcessorAlgorithmUI/DataProcessingUserInterface.md) it was suggested to move the UI into the MantidQtAPI package, but I think `MantidWidgets` makes more sense, as the `DataProcessorUI` is a new widget that can be added to any interface as any other widget).

Fixes #16413.

Internal change, does not need to be in the release notes.

---
#### Reviewer

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):
##### Code Review
- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)? Is it well structured with small focussed classes/methods/functions?
- [ ] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?
##### Functional Tests
- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant documentation been added/updated?
- [ ] Is user-facing documentation written in a user-friendly manner?
- [ ] Has developer documentation been updated if required?
- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

Issue #16415, opened 2016-05-25
DataProcessorUI: pre-processing and post-processing algorithms should be optional


raquelalvarezbanos:
The widget currently needs to be given a pre-processing and a post-processing algorithm. These should be optional. Use the NullObject pattern.

raquelalvarezbanos:
I am splitting this into two separate issues, as making post-processing optional will require more work.

This is now blocked by:
- [ ] https://github.com/mantidproject/mantid/issues/16631
- [ ] https://github.com/mantidproject/mantid/issues/16632

raquelalvarezbanos:
Closing this issue as #16631 and #16632 have now been merged.

PR #16616, opened 2016-06-14
DataProcessorUI demo (Poldi)

### LINKED ISSUE #16225: Demo for data reduction UI
Create a data reduction demo. Have a look here for an example of how the presenter is created for the Reflectometry interface (see [here](https://github.com/mantidproject/mantid/blob/a6ffb617a5c33fd63e94b4c4f14d1c5196c522db/MantidQt/CustomInterfaces/src/Reflectometry/ReflGenericDataProcessorPresenterFactory.cpp)).

raquelalvarezbanos:
Demonstrates how to use the new `DataProcessorWidget` for Poldi reduction.

**To test:**
1. Code changes: most commits come from branch https://github.com/mantidproject/mantid/pull/16587, relevant changes can be found in commits https://github.com/mantidproject/mantid/commit/cfc1e17d79d96c157667f8921ac980b0c7291d0b and https://github.com/mantidproject/mantid/commit/9ac240a3e1fc1ad28f5473209794380488fe1f0c
2. Functional testing: Open the interface (Interfaces -> Poldi -> Poldi Reduction). There is a "Demo" menu with an action "Populate Table", click and the interface will load some sample data and tell you how to run the reduction.

**Some comments:**
- The interface uses the MVP pattern, in this case the implementation is very simple as the interface is essentially the `DataProcessorUI` (I haven't added unit tests but some should be added)
- Note that **all** the table editing features are displayed in the tool bar (on another branch, open Interfaces -> Reflectometry -> ISIS Reflectometry (Polref) to note the difference. There, some of the options can only be found in the menu "Reflectometry". This is to illustrate that both choices are possible, if people prefer the Poldi display we will just have to make a few changes to the widget).
- Pre-processing and post-processing: in this case it may not make much sense to provide pre-processing and post-processing algorithms but at present they are mandatory (there is an issue to make them optional, I'll merge changes here when I fix it):
  - pre-processing: I've set `Plus` as the pre-processing algorithm, although the interface is using `PoldiLoadRuns` (which not only loads but also merges runs), so there is no need for the widget to merge runs again. This won't be used anyway if only one run is specified in the "Run(s)" column.
  - post-processing: I've set `GroupWorkspaces` as the post-processing algorithm, but again I don't think it makes much sense.
- I am showing the properties I though were more important, but we can add more/remove some.

Let me know your opinion/questions/comments/etc!

Fixes #16225 

---
#### Reviewer

##### Code Review
##### Functional Tests


raquelalvarezbanos:
@MichaelWedel This is an example of how the `DataProcessorUI` could be used for Poldi reduction, following [this](http://docs.mantidproject.org/nightly/algorithms/PoldiDataAnalysis-v1.html#usage) user example. Let me know what you think, we can make changes to the widget if needed (actually I have some issues that will change slightly the way in which the widget works, I'm guessing [one of them](https://github.com/mantidproject/mantid/issues/16415) could be really useful in this case).

@AntonPiccardoSelg Just in case you find this PR useful

@OwenArnold FYI

MichaelWedel:
As mentioned on Slack, one general thing I would suggest to change is to let `bool`-columns accept strings like `True` or `false`, this would make it probably a bit easier for users who don't know about conversion between `bool` and `int`.

I would like to add a dialog to this interface that simplifies loading data using `PoldiLoadRuns` and populating the table, I think this should be straightforward.

MantidQt/CustomInterfaces/src/Poldi/PoldiPresenter.cpp
```diff
@@ -0,0 +1,106 @@
...
+  switch (flag) {
+  case WorkspaceReceiver::ADSChangedFlag:
+    // Do nothing
+    break;
+  }
+}
+
+/** Slot to load a demo table following user example in PoldiDataAnalysis
+ * documentation */
+void PoldiPresenter::loadDemoTable() {
```
OwenArnold: Does it make sense to consider this as a `toy` demo? As you've pointed it at Poldi and have the help of @MichaelWedel. Would it make sense to extend the working functionality of this? Not sure I know the answer to this.

OwenArnold:
@MichaelWedel. My position on this is that @raquelalvarezbanos started out by making a completely fictitious demo because that's what I asked her to do. Fortunately she had the better idea of picking a real  technique/instrument. If we get a usable UI out of this rather than just a toy demo, I'd be much happier. So I think we should put the effort in to make this pleasant enough to use that we can point real instrument scientists at it. That would also make a very good story for Nobugs.

MichaelWedel:
@owenarnold I fully agree, especially since it really should not be too difficult to add that dialogue. I do however also think that the demo option should stay more or less as it is, it's a great help for users to get that example.

OwenArnold:
@MichaelWedel @raquelalvarezbanos In that case, we should try to ensure that the POLDI demo doesn't somehow get confused with the real thing. @raquelalvarezbanos, maybe there should be a new issue to take the demo and turn it into something nicely usable on POLDI.

MichaelWedel:
If this is enough to demo the UI that's great. The dialogue for loading actual data can be added as part of the other issue then.

raquelalvarezbanos:
I've just created a new issue https://github.com/mantidproject/mantid/issues/16744 and assigned it to me (just so that it does not get lost, but @MichaelWedel I think it makes more sense to assign it to you, so feel free to do so). Also, let me know if you need any help with the new dialog you plan to add, or with any other thing.

OwenArnold:
I think that True/false to bool conversion aforementioned by Michael should be added as part of this. I think that's a useful conversion.

MichaelWedel:
@raquelalvarezbanos thanks! The only thing I'm wondering about is how to proceed with this PR. Is there anything left to be done (except the bool thing)?

raquelalvarezbanos:
Well, yes: I still have to make the post-processing step optional: https://github.com/mantidproject/mantid/issues/16632 (but in order to do that I first have to fix issue https://github.com/mantidproject/mantid/issues/16734). 

Do we need to post-process Poldi runs? If we do, I would say we can start working on the new Poldi interface, but if we don't we should probably wait for those changes to be merged in, or alternatively use `GroupWorkspaces` as the post-processing algorithm, although it may not make much sense.

The post-processing step refers to how to group runs that have been reduced. In the case of Reflectometry, that means applying `Stitch1DMany` to runs that belong to the same group. For this demo I was using `GroupWorkspaces`

Issue #16298, opened 2016-05-19
Store Workspace Dx data for each bin instead of for each bin edge


SimonHeybrock:
During work on #16096 it became clear that `Dx` data is only ever needed for bin centers, not for bin edges, since it is storing a delta-Q value for each bin in, e.g., SANS. Currently basically all code using `Dx` somehow works around the extra unwanted value.

After discussions with @AntonPiccardoSelg, ORNL as part of ticket #16119 (https://github.com/mantidproject/mantid/pull/16141), and Andrew Jackson (ESS SANS) the consensus is that we never need `Dx` for bin edges.

Most changes are fairly simple (and simplify existing code), but we have to be able to load legacy files. This could be done by simply dropping the last value (if it is present).

PR #16627, opened 2016-06-15
HistogramData module

### LINKED ISSUE #16086: Phase 1 of implementing a Histogram type

### LINKED ISSUE #16096: Phase 2 of implementing a Histogram type: Dx
See design document for details: https://github.com/mantidproject/documents/blob/histogram_type/Design/histogram_type.md.

This is a follow up after #16086. This phase introduces new types for dx-data.

### LINKED ISSUE #16282: Phase 3 of implementing a Histogram type: Y and E
See design document for details: https://github.com/mantidproject/documents/blob/histogram_type/Design/histogram_type.md.

This is a follow up after #16096. This phase introduces new types for y-data and e-data.

SimonHeybrock:
There is a [new `rst` file tries to explain some details](https://raw.githubusercontent.com/mantidproject/mantid/a9e3891a567cce5b1f164b4ed2bb43b1f121a3bc/docs/source/concepts/HistogramData.rst) ([github-rendered version that shows the drawings, but unfortunately not the code examples](https://github.com/mantidproject/mantid/blob/a9e3891a567cce5b1f164b4ed2bb43b1f121a3bc/docs/source/concepts/HistogramData.rst), build `docs-html` locally for best reading experience instead), it might be worth having a look at it before the review.

**Code review:**
This is a very large and critical change, I would like to suggest to split the review into several parts:

1a. Review everything in the new module `HistogramData`.
1b. Review the interface of `ISpectrum` (including `Histogram1D`/`EventList`) and `MatrixWorkspace`.
2. Feedback to me, I will adapt where necessary, and then review the full set of changes myself. (**done**)
3. Review of everything else (existing Mantid code).

I think we should have more than one reviewer for both 1) and 3).

**To test:**

This is a global change, there is no individual test. If you have access to any Python scripts using Mantid that are not part of our automated testing it would be good to check if they still work.

Fixes #16086.
Fixes #16096.
Fixes #16282.

Note there there will be several follow-up PRs once this has been merged, in particular,
- #16298 will store `X` uncertainties for points instead of bin edges (removing parts of the `Dx` interface)
- #16661 might add some (yet to be determined) degree of validation for data in histograms, and also adds generators for easier creation of histograms, e.g., `setBinEdges(1000, LinearGenerator(0, 0.1));`
- @LamarMoore has started refactoring to replace uses of the old `X`, `Y`, `E` interface methods by the new ones.

---
#### Reviewer

##### Code Review
##### Functional Tests

This has been reviewed by several TSC members, namely @martyngigg and Owen Arnold. The changes are largely internal. This is well-structured, documented and tested.


Framework/API/inc/MantidAPI/ISpectrum.h
```diff
@@ -169,6 +169,12 @@ class DLLExport ISpectrum {
   HistogramData::HistogramDx &mutableDx() {
     return mutableHistogramRef().mutableDx();
   }
+  HistogramData::HistogramY &mutableY() {
+    return mutableHistogramRef().mutableY();
+  }
+  HistogramData::HistogramE &mutableE() {
+    return mutableHistogramRef().mutableE();
+  }
```
SimonHeybrock: I think you need a throwing implementation of `mutableY` and `mutableE` in `EventList`. Otherwise `mutableHistogramRef()` gives a reference to the histogram in `EventList` that only contains `X`.

Framework/DataObjects/inc/MantidDataObjects/EventWorkspace.h
```diff
@@ -127,6 +127,16 @@ class DLLExport EventWorkspace : public API::IEventWorkspace {
...
+  HistogramData::HistogramY &mutableY(const size_t index) override {
+    throw Mantid::Kernel::Exception::NotImplementedError(
+        "This method has not been implemented for EventWorkspace");
+  }
+
+  HistogramData::HistogramE &mutableE(const size_t index) override {
+    throw Mantid::Kernel::Exception::NotImplementedError(
+        "This method has not been implemented for EventWorkspace");
+  }
+
```
SimonHeybrock: @LamarMoore Probably we can make all methods on `MatrixWorkspace` non-virtual (those of the `Histogram` interface). The all just call `getSpectrum()`, and we can then handle things via virtual methods on `ISpectrum`.

Framework/API/inc/MantidAPI/WorkspaceFactory.h
```diff
@@ -108,6 +109,13 @@ template class MANTID_API_DLL
 typedef MANTID_API_DLL Mantid::Kernel::SingletonHolder<WorkspaceFactoryImpl>
     WorkspaceFactory;
 
+template <class T, class... InitArgs>
+boost::shared_ptr<T> createWorkspace(InitArgs... args) {
```
OwenArnold: Nice. Finally, proper initialize that doesn't depend on the caller.

Framework/API/inc/MantidAPI/ISpectrum.h
```diff
@@ -115,10 +98,117 @@ class DLLExport ISpectrum {
...
+    HistogramData::Histogram histogram(std::forward<T>(data)...);
+    checkHistogram(histogram);
+    mutableHistogramRef() = std::move(histogram);
+  }
+
+  HistogramData::BinEdges binEdges() const { return histogramRef().binEdges(); }
+  HistogramData::BinEdgeStandardDeviations binEdgeStandardDeviations() const {
+    return histogramRef().binEdgeStandardDeviations();
+  }
+  HistogramData::Points points() const { return histogramRef().binEdges(); }
```
OwenArnold: Is there a copy-paste error here? `histogramRef().binEdges();`. If `Histogram` is in `point` mode, we convert to bin edges and then back to points?

Framework/HistogramData/inc/MantidHistogramData/Histogram.h
```diff
@@ -0,0 +1,462 @@
...
+  Kernel::cow_ptr<HistogramE> m_e{nullptr};
+  // Mutable until Dx legacy interface is removed.
+  mutable Kernel::cow_ptr<HistogramDx> m_dx{nullptr};
+
+public:
+  enum class XMode { BinEdges, Points };
+  explicit Histogram(XMode mode)
+      : m_x(Kernel::make_cow<HistogramX>(0)), m_xMode(mode) {}
+
+  template <class TX, class TY = Counts, class TE = CountVariances>
```
OwenArnold: 👍

Framework/HistogramData/inc/MantidHistogramData/Points.h
```diff
@@ -0,0 +1,75 @@
...
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
```
OwenArnold: Really good. Nice use of CRTP.

Framework/API/inc/MantidAPI/ISpectrum.h
SimonHeybrock: Yes, you are right.

Framework/HistogramData/inc/MantidHistogramData/VectorOf.h
```diff
@@ -0,0 +1,158 @@
...
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+template <class T, class CowType> class VectorOf {
+public:
+  /// Default constructor, the stored object will be NULL.
+  VectorOf() = default;
+  /// Construct stored object of length count initialized to the constant value.
+  VectorOf(size_t count, const double &value) {
+    m_data = Kernel::make_cow<CowType>(count, value);
```
OwenArnold: Could be done in an initializer list.

Framework/HistogramData/inc/MantidHistogramData/VectorOf.h
```diff
@@ -0,0 +1,158 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+template <class T, class CowType> class VectorOf {
```
OwenArnold: Is the `T` type argument ever used?

SimonHeybrock: It is part of the CRTP. It ensures that subtypes like `Points` and `BinEdges` are different and incompatible types, by having separate base classes `VectorOf<Points>` and `VectorOf<BinEdges>.

OwenArnold: 👍 . That's a valid reason. Might be worth a comment about that in the class header.

Framework/HistogramData/src/BinEdges.cpp
```diff
@@ -0,0 +1,39 @@
+#include "MantidHistogramData/BinEdges.h"
+#include "MantidHistogramData/Points.h"
+
+namespace Mantid {
+namespace HistogramData {
+
+/// Constructs BinEdges from points, approximating each bin edges as mid-point
+/// between two points.
+BinEdges::BinEdges(const Points &points) {
+  if (!points)
```
OwenArnold: @SimonHeybrock: What overload is being called at this point? `!points`?

OwenArnold: Ok. I've found it in `VectorOf`

Framework/HistogramData/inc/MantidHistogramData/VectorOf.h
```diff
@@ -0,0 +1,158 @@
...
+  }
+  /// Move assignment to stored object from data.
+  VectorOf &operator=(CowType &&data) & {
+    if (!m_data || (&(*m_data) != &data))
+      m_data = Kernel::make_cow<CowType>(std::move(data));
+    return *this;
+  }
+
+  /// Checks if *this stores a non-null pointer.
+  explicit operator bool() const { return m_data.operator bool(); }
```
OwenArnold: Is this the right way to call the overloaded function operator on `m_data`? I haven't seen it in this form before?

Framework/HistogramData/inc/MantidHistogramData/Offsetable.h
```diff
@@ -0,0 +1,65 @@
...
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+template <class T> class Offsetable {
+public:
+  /// Offsets each element in the container by offset.
+  T &operator+=(const double offset) & {
+    auto &derived = static_cast<T &>(*this);
```
OwenArnold: Do we have to down_cast in the base class? Why not use `dynamic_cast`?

Framework/HistogramData/inc/MantidHistogramData/VectorOf.h
SimonHeybrock: I admit I have not seen it in many places either. But the compiler seems to like it.

Framework/HistogramData/inc/MantidHistogramData/VectorOf.h
SimonHeybrock: Will do.

Framework/HistogramData/inc/MantidHistogramData/Offsetable.h
SimonHeybrock: The CRTP provides compile-time polymorphism, so we can avoid the slow `dynamic_cast`. The CRTP guarantees that the `static_cast` will not fail. We need the cast since iterators are defined in the mixin `Iterable`, not in the base class.

OwenArnold: Agreed. the cast wont fail. Seems a little odd that the `Iterable` `begin`, `end` functionality used actually comes from an orthogonal CRTP class, but I don't see that as a major issue.

SimonHeybrock: Yes, it feels indeed a bit odd, at least in C++. In Python Mixins are more common, and the mechanism would not feel so unusual.
Having the iterators in `VectorOf` would not change anything though, since it is also just a mixin.

Framework/HistogramData/inc/MantidHistogramData/HistogramX.h
```diff
@@ -0,0 +1,75 @@
...
+
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+class MANTID_HISTOGRAMDATA_DLL HistogramX
+    : public detail::FixedLengthVector<HistogramX>,
+      public detail::Offsetable<HistogramX>,
+      public detail::Scalable<HistogramX> {
+public:
+  using detail::FixedLengthVector<HistogramX>::FixedLengthVector;
```
OwenArnold: If I'm being very picky. It would be nice to have a comment explaining that all constructors apart from the default come from the `FixedLengthVector`

Framework/HistogramData/inc/MantidHistogramData/HistogramX.h
```diff
@@ -0,0 +1,75 @@
...
+  HistogramX() = default;
+  // The copy and move constructor and assignment are not captured properly by
+  // the using declaration above, so we need them here explicitly.
+  HistogramX(const HistogramX &) = default;
+  HistogramX(HistogramX &&) = default;
+  HistogramX &operator=(const HistogramX &)& = default;
+  HistogramX &operator=(HistogramX &&)& = default;
+
+  // The classes are friends, such that they can modify the length.
+  friend class Histogram;
```
OwenArnold: I'd like to know better what the use-cases of having these `friend` declarations are? Can you point me to the client code that renders them necessary?

OwenArnold: I think one example is in `Histogram::dataX()` calling `FixedLengthVector::mutableRawData()`

OwenArnold: One way around this would be to have a virtual pointer to some mutable interface in `Histogram` of which `HistogramX` implements, but only `Histogram` sees. I imagine you thought about this already though?

Framework/HistogramData/inc/MantidHistogramData/HistogramX.h
SimonHeybrock: Good point. I am wondering if Doxygen is able to pick up the `using` declaration and will include the corresponding documentation? Do you know?

Framework/HistogramData/inc/MantidHistogramData/HistogramX.h
SimonHeybrock: The prime example is actually `VectorOf::mutableRawData()`, i.e., the `friend` declaration is not just for the legacy interface.

The point is, `Histogram` is not the only class that needs access: All we want to do is to forbid length modifications that can break a `Histogram`:

- `Histogram` returns `BinEdges`, `Counts`, etc. always by value, never by references. Thus changing the length of those types (all inheriting from `VectorOf`) is also allowed.
- `Histogram` returns references to `HistogramX`, `HistogramY`, etc., so changing the length directly on those types (inheriting from `FixedLengthVector`) must be forbidden.

Furthermore, I want to avoid virtuality in those low-level bits when possible.

Framework/HistogramData/inc/MantidHistogramData/HistogramX.h
OwenArnold: Not sure about that. I would suspect not. I know that doxygen has had problems parsing templates in the past based on my experience, not sure how up-to-date it is kept.

OwenArnold: There's an old bug recorded about this https://bugzilla.gnome.org/show_bug.cgi?id=333647

Framework/HistogramData/inc/MantidHistogramData/HistogramX.h
OwenArnold: I do however always query the use of `friend` as it breaks encapsulation, and is sometimes used where there are alternatives. I don't see a way around this if you want to avoid runtime polymorphism so I'm OK with this in this limited case.

SimonHeybrock: The alternative would be to re-allocate the underlying `HistogramX` on length modification of `BinEdges`, but I think we should avoid that performance penalty?

Framework/HistogramData/inc/MantidHistogramData/Histogram.h
```diff
@@ -0,0 +1,462 @@
...
+  // Temporary legacy interface to X
+  void setX(const Kernel::cow_ptr<HistogramX> &X) & { m_x = X; }
+  MantidVec &dataX() & { return m_x.access().mutableRawData(); }
+  const MantidVec &dataX() const & { return m_x->rawData(); }
+  const MantidVec &readX() const { return m_x->rawData(); }
+  Kernel::cow_ptr<HistogramX> ptrX() const { return m_x; }
+
+  // Temporary legacy interface to Y
+  void setY(const Kernel::cow_ptr<HistogramY> &Y) & { m_y = Y; }
+  MantidVec &dataY() & { return m_y.access().mutableRawData(); }
```
OwenArnold: @SimonHeybrock, so `HistogramY` is a `FixedLengthVector`, but exposing its held vector in this way removes the protection. Have you thought about what happens if the vector is extended and then subsequently used through the `FixedLengthVector` wrapper again?

SimonHeybrock: I think we cannot do anything about this as long as the old interface is around. Client code using the old interface simply requires length modifications, so without refactoring to use the new interface we are stuck with this.
If the vector is extended nothing bad happens to `FixedLengthVector`. The only risk is that the `Histogram` can get into an inconsistent state where `X` and `Y` length do not match, but this is the same issue we have right now. By using the new interface in more and more places we are reducing this risk step by step.

Framework/HistogramData/inc/MantidHistogramData/Histogram.h
```diff
@@ -0,0 +1,462 @@
...
+
+/** Sets the Histogram's bin edges.
+
+ Any arguments that can be used for constructing a BinEdges object are allowed,
+ however, a size check ensures that the Histogram stays valid, i.e., that x and
+ y lengths are consistent. */
+template <typename... T> void Histogram::setBinEdges(T &&... data) & {
+  BinEdges edges(std::forward<T>(data)...);
+  // If there is no data changing the size is ok.
+  // if(m_y)
```
SimonHeybrock: @OwenArnold You will be wondering about this. The problem here is that `EventWorkspace` ignores the `X` length it is being initialized with, and thus requires to be changed at a later point.
I had considered allowing modification of the `X` length if there is no data, but I am not sure. I am currently working around this in the `EventWorkspace` code, but maybe it would be better to fix `EventWorkspace` instead. Let me know what you think.

Framework/HistogramData/inc/MantidHistogramData/Histogram.h
```diff
@@ -0,0 +1,462 @@
...
+#include "MantidHistogramData/Points.h"
+#include "MantidHistogramData/PointStandardDeviations.h"
+#include "MantidHistogramData/PointVariances.h"
+#include "MantidHistogramData/Counts.h"
+#include "MantidHistogramData/CountVariances.h"
+#include "MantidHistogramData/CountStandardDeviations.h"
+#include "MantidHistogramData/Frequencies.h"
+#include "MantidHistogramData/FrequencyVariances.h"
+#include "MantidHistogramData/FrequencyStandardDeviations.h"
+
```
OwenArnold: Minor point, but based on "include what you" use best practices, I think you're incorrectly obtaining type definitions through loosely related includes. For example: No explicit include of `HistogramX.h`, but that comes through via `BinEdges.h`.

Framework/HistogramData/test/HistogramXTest.h
```diff
@@ -0,0 +1,20 @@
...
+using Mantid::HistogramData::HistogramX;
+
+class HistogramXTest : public CxxTest::TestSuite {
+public:
+  // This pair of boilerplate methods prevent the suite being created statically
+  // This means the constructor isn't called when running other tests
+  static HistogramXTest *createSuite() { return new HistogramXTest(); }
+  static void destroySuite(HistogramXTest *suite) { delete suite; }
+
+  void test_fake() {}
```
OwenArnold: I guess you haven't got round to writing these tests yet, but plan to?

Framework/HistogramData/test/HistogramYTest.h
```diff
@@ -0,0 +1,20 @@
...
+using Mantid::HistogramData::HistogramY;
+
+class HistogramYTest : public CxxTest::TestSuite {
+public:
+  // This pair of boilerplate methods prevent the suite being created statically
+  // This means the constructor isn't called when running other tests
+  static HistogramYTest *createSuite() { return new HistogramYTest(); }
+  static void destroySuite(HistogramYTest *suite) { delete suite; }
+
+  void test_fake() {}
```
OwenArnold: I guess you haven't got round to writing these tests yet, but plan to?

Framework/HistogramData/test/HistogramXTest.h
SimonHeybrock: Well, most things are tested as parts of the mixins, and I don't think we want to re-test everything (that is another nice feature of the Mixins, we write tests once and use them in many other classes). The only thing to test here are the basic constructors, but they are default, so that is not so useful either.

I think what might make sense is to have a test that verifies if the correct mixins are used. Maybe by a `dynamic_cast` to each of the expected base types (or I can figure out some other method)?

What do you think?

OwenArnold: I like that idea. We should also add a comment about the fact that the mixin functionality is tested in each of the base class tests.

Framework/DataObjects/inc/MantidDataObjects/EventList.h
```diff
@@ -357,7 +327,37 @@ class DLLExport EventList : public Mantid::API::IEventList {
...
+  void checkWorksWithPoints() const override;
+
 private:
+  const HistogramData::Histogram &histogramRef() const override {
+    return m_histogram;
+  }
+  HistogramData::Histogram &mutableHistogramRef() override;
+
+  /// Histogram object holding the histogram data. Currently only X.
+  HistogramData::Histogram m_histogram;
```
OwenArnold: So the other Histogram vectors `HistogramY` etc are left idle?

SimonHeybrock: Yes. I don't like it, but with the current `MatrixWorkspace` class tree there is not really another option.

Note that `EventList::histogram()` returns a fully initialized `Histogram` (generated on the fly similar to `readY()`).

There is also the ugly interaction with the MRU, but we cannot get rid of that, unless we accept a significant increase in memory consumption by keeping all `Y` and `E` in `m_histogram` (generated and updated based on the list of events).

Framework/HistogramData/inc/MantidHistogramData/Histogram.h
OwenArnold: I've traced this through now, and I'm satisfied with what you've done. I don't think it's a good idea to try to fix things like the `EventWorkspace` initialisation in one go, certainly not as part of this.

SimonHeybrock: @OwenArnold So we keep forbidding size changes, even if there is no data? If so, I'll remove those comments from the code.

OwenArnold:
I'm very happy with what's been done here. Providing that the test suites are passing, I'd be satisfied with this being made ready for a merge into master. I would however like to get another TSC member (ideally from SNS) to have a look over this too.

Framework/DataObjects/test/EventListTest.h
```diff
@@ -2488,6 +2488,36 @@ class EventListTest : public CxxTest::TestSuite {
     TS_ASSERT_THROWS(el.setPointStandardDeviations(1), std::runtime_error);
   }
 
+  void test_setCounts_fails() {
```
OwenArnold: Nice test additions.

Framework/API/inc/MantidAPI/ISpectrum.h
```diff
@@ -1,13 +1,17 @@
...
-#include "MantidKernel/System.h"
 #include "MantidGeometry/IDTypes.h"
+#include "MantidHistogramData/Histogram.h"
+#include "MantidKernel/System.h"
 #include "MantidKernel/cow_ptr.h"
 
 #include <set>
 
 namespace Mantid {
+namespace DataObjects {
```
martyngigg: Should we be forward declaring across modules like this? I can't see where it is used.

Framework/DataObjects/src/Workspace2D.cpp
```diff
@@ -68,19 +68,19 @@ void Workspace2D::init(const std::size_t &NVectors, const std::size_t &XLength,
...
+        new Histogram1D(HistogramData::getHistogramXMode(XLength, YLength));
     data[i] = spec;
     // Set the data and X
-    spec->setX(t1);
-    spec->setDx(t1);
-    spec->resetHasDx();
+    spec->setX(x);
     // Y,E arrays populated
-    spec->setData(t2, t2);
+    spec->setCounts(y);
```
martyngigg: Is the counts/error data no longer shared by default as there seemed little point in doing that as it would most likely be broken anyway?

Framework/API/inc/MantidAPI/ISpectrum.h
SimonHeybrock: Good catch. It is a leftover of a `friend` declaration I had to add a while ago, but managed to remove later.

Framework/DataObjects/src/Workspace2D.cpp
SimonHeybrock: Counts and errors now have distinct underlying types (`HistogramY` and `HistogramE`), so they cannot be shared any more.

On the plus side, we now have getters and setter for the shared pointers (or rather cow pointers) for `Y` and `E`. Those did not exist in the old interface. We should thus be able to improve the overall amount of sharing.

OwenArnold:
:shipit:

The system tests should pass because they did on the last set of changes. Simon only changed the release notes. If everything passes (and it should shortly), my recommendation is to merge this.

FedeMPouzols:
This looks pretty solid. It has been designed carefully and has gone through a thorough multi-review.
I particularly like the "transition" documentation. Let's merge it in. More work follows on this.

PR #16629, opened 2016-06-15
Moving the logic of single peak fitting to presenter

### LINKED ISSUE #16220: Engg GUI, single peak fitting: move the logic that builds the list of peaks to presenter, add regression tests
PR #16211 fixes a bug in the Engg GUI. The method modified in that PR is in the view and is manipulating the list with non-trivial operations. The view is taking care of bookkeeping the list of peaks. This is not the ideal separation. The view should just provide the inputs that the user gives, and take well digested results to just show them to the user.
The manipulation of the list of peaks would live better in the presenter, where it could be tested, so the bug fix of the PR could have a regression test to check that adding peaks always works well regardless of what's included in the list.

shahrozahmed1:
**Description of work.**

This PR moves most of the logic of single peak fitting code to presenter from the view. This will enable the code to be tested and maintained appropriately. The PR request now also removes the duplicates comma within the expected peak list, which seemed to have been causing a problem during the demo.

**To test:**
Code review
All builds pass successfully
Presenter test have been updated
Fitting tab still functions as expected
Any further functions or bit code which could be moved to presenter..?

Fixes #16220 

---
#### Reviewer

##### Code Review
##### Functional Tests


MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/EnggDiffraction/IEnggDiffractionView.h
```diff
@@ -387,6 +387,13 @@ class IEnggDiffractionView {
   virtual std::string fittingPeaksData() const = 0;
 
   /**
+  * Sets the peak list according to the QString given
+  *
+  * @param peakList list of expected peaks to be fitted as QString
+  */
+  virtual void setPeakList(QString peakList) const = 0;
```
FedeMPouzols: Just one small detail. This is an interface which does not use Qt classes for anything. It would be better to use `std::string` rather than adding `QString`, `QStringList`, etc. here. As the presenter is preparing the strings as `std::string`, this fits well between the view and the presenter. Only the Qt specific view classes should use Qt classes, as far as possible.

MantidQt/CustomInterfaces/src/EnggDiffraction/EnggDiffractionPresenter.cpp
```diff
@@ -863,27 +867,64 @@ void EnggDiffractionPresenter::inputChecksBeforeFitting(
...
+    if (expectedPeaks.at(strLength) == ',') {
+      expectedPeaks.erase(strLength, 1);
+    }
+
+    m_view->setPeakList(expectedPeaks);
+  }
+
+  return expectedPeaks;
+}
+
```
FedeMPouzols: Good that all this validation code is now in the presenter.

FedeMPouzols:
It's worth keeping in mind that some of the method names might need revisiting when/if we generalize this to be shared between the single-peak and the whole-pattern fitting tabs. At the same time more stuff from the view may need to go to the presenter. But this already moves most of the logic, error checks, etc.

MantidQt/CustomInterfaces/src/EnggDiffraction/EnggDiffractionPresenter.cpp
```diff
@@ -863,27 +867,64 @@ void EnggDiffractionPresenter::inputChecksBeforeFitting(
...
+    auto *comma = ",";
+
+    for (size_t i = 0; i < expectedPeaks.size() - 1; i++) {
+      size_t j = i + 1;
+
+      if (expectedPeaks[i] == *comma && expectedPeaks[i] == expectedPeaks[j]) {
+        expectedPeaks.erase(j, 1);
+        i--;
+
```
FedeMPouzols: For later, this is the type of processing that can be done with `find_if`, `remove_if`, etc.

MantidQt/CustomInterfaces/src/EnggDiffraction/EnggDiffractionPresenter.cpp
```diff
@@ -1371,6 +1412,132 @@ void EnggDiffractionPresenter::setDataToClonedWS(std::string &current_WS,
...
+                        "Could not add banks to "
+                        "combo-box or list widget; " +
+                            static_cast<std::string>(re.what()) +
+                            ". Please try again");
+  }
+}
+
+void EnggDiffractionPresenter::setRunNoItems(
+    std::vector<std::string> runNumVector, bool multiRun) {
+  try {
```
FedeMPouzols: runNumVector can be a `const &`.

MantidQt/CustomInterfaces/src/EnggDiffraction/EnggDiffractionPresenter.cpp
```diff
@@ -1371,6 +1412,132 @@ void EnggDiffractionPresenter::setDataToClonedWS(std::string &current_WS,
...
+  } catch (std::runtime_error &re) {
+    m_view->userWarning("Unable to insert items: ",
+                        "Could not add list widget; " +
+                            static_cast<std::string>(re.what()) +
+                            ". Please try again");
+  }
+}
+
+void EnggDiffractionPresenter::setDefaultBank(
+    std::vector<std::string> splittedBaseName, QString selectedFile) {
```
FedeMPouzols: `splittedBaseName` can also be a `const &`.
And also for later, `QString` should not be used here in the presenter.

MantidQt/CustomInterfaces/src/EnggDiffraction/EnggDiffractionPresenter.cpp
```diff
@@ -1371,6 +1412,132 @@ void EnggDiffractionPresenter::setDataToClonedWS(std::string &current_WS,
...
+  else if (!m_view->getFittingRunNo().empty())
+    m_view->setFittingRunNo(selectedFile);
+}
+
+bool EnggDiffractionPresenter::isDigit(std::string text) {
+  for (size_t i = 0; i < text.size(); i++) {
+    char *str = &text[i];
+    if (std::isdigit(*str)) {
+      return true;
+    }
```
FedeMPouzols: This is a `std::find_if`?

MantidQt/CustomInterfaces/src/EnggDiffraction/EnggDiffractionViewQtGUI.cpp
```diff
@@ -1405,6 +1405,40 @@ std::string EnggDiffractionViewQtGUI::getFittingRunNo() const {
...
+
+void MantidQt::CustomInterfaces::EnggDiffractionViewQtGUI::
+    setFittingListWidgetCurrentRow(int idx) const {
+  m_uiTabFitting.listWidget_fitting_run_num->setCurrentRow(idx);
+}
+
+int EnggDiffractionViewQtGUI::getFittingComboIdx(std::string bank) const {
+  return m_uiTabFitting.comboBox_bank->findText(QString::fromStdString(bank));
+}
+
```
FedeMPouzols: This block of methods is a perfect example of what the ideal view code should look like. All are pasively passing around values, to/from the widgets/user.

FedeMPouzols:
- Code changes look good.  Moves code to the presenter for better testability, and updates the mock definitions and tests. More changes and tests will follow in other issues.
- Tested a simple user session, doing a few fits, and the interface behaved well / as usual.

:shipit:

PR #16634, opened 2016-06-15
CalMuonDetectorPhases: frequency units and detector IDs

### LINKED ISSUE #16142: CalMuonDetectorPhases: frequency should be in MHz
[CalMuonDetectorPhases](http://docs.mantidproject.org/nightly/algorithms/CalMuonDetectorPhases-v1.html) takes a *Frequency* parameter. Users generally expect this to be in MHz, but it is actually in radians per second. 

The parameter description (tooltip in the input dialog) just says "Starting hint for the frequency" and gives no clue to the units.
(You can deduce the units by looking at the documentation, though it is never stated explicitly.)

- Scientists would like *Frequency* to be in MHz
- Units should be clear in parameter description

tom-perkins:
Fixed two issues in `CalMuonDetectorPhases`:
- _Frequency_ parameter is now supplied in MHz, as requested by scientists (previously was megaradians per second). The units are now stated in the help string for the parameter.
- The "Detector" column of the phase table produced shows detector IDs rather than workspace indices. This is more meaningful and consistent with the column heading.

**To test:**

Unit and doc tests have been updated to test the new changes.
To test manually:
- Load `MUSR00022725.nxs` from unit test data
- Run `CalMuonDetectorPhases` on it with _Frequency_ = 1.355 MHz: (should succeed without errors)

``` python
detectorTable, fittingResults = CalMuonDetectorPhases(InputWorkspace='MUSR00022725', Frequency=1.355)
```
- Check the `detectorTable` produced - the detector ids should go from 1-64 (NOT 0-63)
- Now try again with _Frequency_ = 8.514 (units of Mrad/s) - the fit should fail
- Open the `CalMuonDetectorPhases` dialog and hover over the "Frequency" field. Check the tooltip tells you what units to use.

Fixes #16142. 

[Release notes](https://github.com/mantidproject/mantid/blob/17c2f52ea1dcb8807cb8f3cdd9da25d71bc46d40/docs/source/release/v3.8.0/muon.rst#algorithms) have been updated.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Algorithms/src/CalMuonDetectorPhases.cpp
+++ b/Framework/Algorithms/src/CalMuonDetectorPhases.cpp
@@ -182,7 +182,7 @@ void CalMuonDetectorPhases::fitWorkspace(const API::MatrixWorkspace_sptr &ws,
     // Now we have our fitting results stored in tab
     // but we need to extract the relevant information, i.e.
     // the detector phases (parameter 'p') and asymmetries ('A')
-    extractDetectorInfo(tab, resTab, ispec);
+    extractDetectorInfo(tab, resTab, getDetectorIDFromSpectrumIndex(ws, ispec));
   }
 }
 
@@ -190,11 +190,11 @@ void CalMuonDetectorPhases::fitWorkspace(const API::MatrixWorkspace_sptr &ws,
 * and adds a new row to the results table with them
 * @param paramTab :: [input] Output parameter table resulting from the fit
 * @param resultsTab :: [input] Results table to update with a new row
-* @param ispec :: [input] Spectrum number
+* @param detid :: [input] Detector ID
 */
 void CalMuonDetectorPhases::extractDetectorInfo(
     const API::ITableWorkspace_sptr &paramTab,
-    const API::ITableWorkspace_sptr &resultsTab, const int ispec) {
+    const API::ITableWorkspace_sptr &resultsTab, const size_t detid) {
 
   double asym = paramTab->Double(0, 1);
   double phase = paramTab->Double(2, 1);
@@ -211,7 +211,7 @@ void CalMuonDetectorPhases::extractDetectorInfo(
   }
   // Copy parameters to new row in results table
   API::TableRow row = resultsTab->appendRow();
-  row << ispec << asym << phase;
+  row << static_cast<int>(detid) << asym << phase;
 }
 
 /** Creates the fitting function f(x) = A * sin( w*x + p) + B as string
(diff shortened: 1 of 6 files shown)


Framework/Algorithms/src/CalMuonDetectorPhases.cpp
```diff
@@ -539,5 +545,27 @@ void CalMuonDetectorPhases::reportProgress(const int thisSpectrum,
...
+      return entry.first;
+    }
+  }
+  // Couldn't find it
+  std::ostringstream error;
+  error << "Could not find detector ID corresponding to spectrum "
+        << spectrumIndex;
+  throw std::runtime_error(error.str());
+}
+
```
SimonHeybrock: Why not use `ISpectrum::getDetectorIDs`?

SimonHeybrock:
The algorithm loops over all spectra (https://github.com/mantidproject/mantid/blob/f1d6ad60c6dc7ce28388ccdda99de28fa2313c56/Framework/Algorithms/src/CalMuonDetectorPhases.cpp#L160), wouldn't it make more sense to have the spectrum number as a header instead of the detector ID?  I do not see a good reason why it should fail for spectra without detectors, since actually it does not seem to use any information from the detectors.

Framework/Algorithms/src/CalMuonDetectorPhases.cpp
tom-perkins: Good idea, that would be a better way. The function here is badly named too, it should be called `getDetectorIDFromWorkspaceIndex` not from "spectrum index" - I'll change this.

SimonHeybrock: @tom-perkins See my comment below first, I am not sure detector ID makes sense here at all.

tom-perkins:
The algorithm loops over workspace indices but these are referred to as spectra and `ispec` in the code - I've changed this to hopefully clarify it. 
The table that the algorithm produces gives the asymmetry and phase for each detector, so it's most useful to use detector ID or spectrum number here (they're equal in this case). Previously it used workspace index but labelled the column "detector", which was confusing.

SimonHeybrock:
But if I understand the code correctly, it *does not* give asymmetry and phase for each *detector*, but for each *spectrum*. If a spectrum has more than one detector the algorithm will only give results for the first detector of the spectrum.

tom-perkins:
You're right, the usual use case here is that each spectrum would correspond to one detector but to avoid confusion I think the heading for that column should be "Spectrum number" instead of "detector".

SimonHeybrock:
Code looks good now, and changes work as described in the PR. Once system tests pass, this is ready - :shipit:.

tom-perkins:
Thanks, have updated the docs to clarify this.

SimonHeybrock:
Great, also 👍 for mentioning spectrum/detector association in the documentation. :shipit:

Framework/Algorithms/test/CalMuonDetectorPhasesTest.h
```diff
@@ -124,6 +124,22 @@ class CalMuonDetectorPhasesTest : public CxxTest::TestSuite {
...
+    calc->initialize();
+    calc->setChild(true);
+    calc->setProperty("InputWorkspace", ws);
+    calc->setPropertyValue("Frequency", "25"); // Mrad/s, not MHz
+    calc->setPropertyValue("DataFitted", "fit");
+    calc->setPropertyValue("DetectorTable", "tab");
+    calc->setProperty("ForwardSpectra", std::vector<int>{1, 2});
+    calc->setProperty("BackwardSpectra", std::vector<int>{3, 4});
+
+    TS_ASSERT_THROWS(calc->execute(), std::runtime_error);
```
FedeMPouzols: This assert is failing on clean master builds: http://builds.mantidproject.org/job/master_clean-win7/222/testReport/
@tom-perkins: has there been any related changed that happened more or less in parallel? It would be strange that this passed on the PR builds but then fails on master.

SimonHeybrock: @FedeMPouzols Could this be due to a different GSL version? Maybe the fit succeeds even with this bad starting guess?

tom-perkins: @FedeMPouzols I can't think of a related change that would have affected this. I will try running the test a number of times on my PC and see how often it fails - if the test is unstable then we can remove it for now, and put in a better test later on.

FedeMPouzols: I think that we have the same GSL version on all the windows machines. Sometimes I've seen fitting results vary slightly between diferent boxes, with apparently the same versions of all the seemingly relevant libraries. So this issue may not be reproducible on some machines.

tom-perkins: Ok, I've removed this test in #16662 because it seems to depend on the machine it runs on - at a later time I can come back and add a more robust test.

PR #16650, opened 2016-06-17
Set Default of CorrectDetectorPositions to False in RRO and RROA

### LINKED ISSUE #16648: Change Default of 'CorrectionDetectorPositions' in RROA
This is a sub-issue of #16455 
### Expected behavior
- Reflectometry scientists have expressed that they often want the position of detectors to stay the same and not be corrected by the algorithm
### Actual behavior
- This is only possible if the default of `CorrectDetectorPositions` is `False` or the desired value is specified by the user.

MattKing06:
Scientists have expressed the need to change the default value of CorrectionDetectorPositions to `False` as it is not often they want to correct the positions of their detectors.

**To test:**
- Code Review

Fixes #16648

---
#### Reviewer

##### Code Review
##### Functional Tests

Unit and system tests have been adapted to cope with this change


raquelalvarezbanos:
Code changes look good to me. This change was requested by scientists, so I think this should be mentioned in the release notes, @MattKing06 could you add a line about this please?

MattKing06:
@raquelalvarezbanos Yes, you're right, I'll add this to the release notes.

raquelalvarezbanos:
Thanks @MattKing06 

:shipit:

martyngigg:
I assume the scientists were made aware that this means the behaviour of all scripts not specifying this value will change?

MattKing06:
@martyngigg I believe so, the scientists that we have contact with will definitely know as they requested this change. I can't say for definite whether the other scientists using these algorithms will know however we can always send a reminder around warning them of these changes.

raquelalvarezbanos:
Let's send them the installers to make sure they are happy with this change. As @MattKing06 says, this was requested by scientists here at ISIS.

martyngigg:
Thanks, both. I'm sure it will probably be fine but I just wanted to double check.

raquelalvarezbanos:
We asked scientists again about this and they confirmed that they would like this property to be set to false by default. However, I now realize I don't understand how this property works:

For instance, the documentation says: "Detector Position Correction is used for when the position of the detector is not aligned with the reflected beamline. The correction algorithm used is `SpecularReflectionPositionCorrect v1` which is a purely vertical position correction"

But if I run the following script:

```
offspec = Load('OFFSPEC00027575.raw')
trans = Load('OFFSPEC00027585.raw')

theta=0.5

_, IvsLam_true, _ = ReflectometryReductionOneAuto(offspec, ThetaIn=theta, FirstTransmissionRun=trans, CorrectDetectorPositions=1, MomentumTransferStep=1e-3)
_, IvsLam_false, _ = ReflectometryReductionOneAuto(offspec, ThetaIn=theta, FirstTransmissionRun=trans, CorrectDetectorPositions=0, MomentumTransferStep=1e-3)

print offspec.getItem(0).getInstrument().getDetector(4).getPos()
print IvsLam_false.getItem(0).getInstrument().getDetector(4).getPos()
print IvsLam_true.getItem(0).getInstrument().getDetector(4).getPos()
```

The output is:

```
[0,0,3.19]
[0,0,3.19]
[0,0.0278387,6.38]
```

which means that `SpecularReflectionPositionCorrect` is not only moving the detectors vertically but also horizontally (or there is another child algorithm moving them along the `Z` axis).

I first need to understand what is happening. In addition, we plan to write a second version of the algorithm (see https://github.com/mantidproject/mantid/issues/17429), so this change should probably go in this new version.

PR #16655, opened 2016-06-17
Implement ListProperty widget for workspaces

### LINKED ISSUE #16600: GroupWorkspaces: better default option for input
*Reported by James Lord*

The parameter *InputWorkspaces* for `GroupWorkspaces` defaults to whatever workspaces were grouped the last time the algorithm was run. These workspaces may not be present any more. It would be more helpful to use the selected workspaces, if any:

> I can select several workspaces in the ADS (workspaces) pane with 
> shift-click or ctrl-click. If I then run GroupWorkspaces, it would be 
> helpful if the "InputWorkspaces" box is filled in with them. But 
> instead it keeps an old default of workspaces I grouped at some point 
> in the past (which aren't there now). As I wanted to group a series of 
> sequentially numbered runs it was easier in the end to write a short 
> script in the Python Window than to type them all in.

NickDraper:
Simple implemtation of a list property widget for multiple selection of workspaces.

This is indicated by using a alg property std::vectorstd::string with an ADSValidator.
#### Notes:
1. The ListValidator could be extended later to also support this functiionality
#### Current Limitations :
1. Currently only for Workspaces in the ADS
2. No further validation of the workspaces can be tested other than presence in the ADS

**To test:**

Two algorithms have been modified to use this GUI approach, GroupWorkspaces and MergeRuns
1. try them both out and test drive the interface

Fixes #16600
### RELEASE NOTES

docs/source/release/v3.8/ui.rst

---
#### Reviewer

##### Code Review
##### Functional Tests


KarlPalmen:
I like the improvement in the user interface. It saves one from having to type the workspace names and possibly getting them wrong. However the documentation for the two algorithms does not reflect this well. It may be worth spending plenty of time on improving the documentation, because it set a precedent for other algorithms that will receive this widget in future.

KarlPalmen:
The change in documentation is just sufficient :shipit: .

Framework/API/src/ADSValidator.cpp
```diff
@@ -0,0 +1,75 @@
...
+  if (!m_isOptional && (value.size() == 0))
+    return "Select a value";
+  if (!m_AllowMultiSelection && (value.size() > 1)) {
+    return "Only one workspace was expected.";
+  }
+  auto &ads = AnalysisDataService::Instance();
+  std::ostringstream os;
+  for (std::string wsName : value) {
+    if (!ads.doesExist(wsName)) {
+      os << "The workspace \"" << wsName << "\" is not in the workspace list.";
```
FedeMPouzols: If I understand this correctly, when several workspaces are missing, users will get a long string without spaces or new lines between the "The workspace ... is not in the workspace list." sentences.

MantidQt/API/inc/MantidQtAPI/ListPropertyWidget.h
```diff
@@ -0,0 +1,63 @@
...
+#include <qlistwidget.h>
+#include <QLabel>
+
+namespace MantidQt {
+namespace API {
+
+/** Widget for displaying a Property that has a set of allowed values.
+ * The display is then a multi selection list box instead of a Text box.
+
+  @date 2012-02-17
```
FedeMPouzols: 2012?

FedeMPouzols:
This looks very useful. @NickDraper: let us know if you'd like to do a last second fix for those nitpicky comments. Otherwise this looks ready.

Framework/API/inc/MantidAPI/ADSValidator.h
```diff
@@ -0,0 +1,79 @@
...
+//----------------------------------------------------------------------
+// Includes
+//----------------------------------------------------------------------
+#include "MantidKernel/TypedValidator.h"
+#include "MantidAPI/DllConfig.h"
+
+namespace Mantid {
+namespace API {
+
+using namespace Mantid::Kernel;
```
martyngigg: `using` declarations should not be anywhere in the header. Please can we remove this before it gets to master.

martyngigg:
We shouldn't merge until the note about the using declaration is addressed.

FedeMPouzols:
@martyngigg : do we have any best practice with respect to using forward declarations of Qt classes in MantidQt/API headers? I see we use forward declarations sometimes, but not very consistently. It looks like they could be used in many more places than we currently have.

martyngigg:
@FedeMPouzols The standard rules apply really. Basically if you don't need to know the size of the object then you can use a forward declaration. For Qt pointers are used heavily so most of the time forward declarations will work very well.

FedeMPouzols:
Aha, thanks, so for example in this PR `ListPropertyWidget.h` could do with forward declarations of  `QListWidget` and `QLabel` I think. But then it will need an include of `QObject` or `QString`. This is an area where many other headers could be improved.

KarlPalmen:
Works OK :shipit:

FedeMPouzols:
I checked for `using` in headers. There seem to be a few in `MantidQt`, but `Framework` looks ok: #16755.

Issue #16455, opened 2016-05-27
Changes to ReflectometryReductionOne (Auto) requested by scientists


raquelalvarezbanos:
In a meeting with Max and Jos yesterday we looked at `ReflectometryReductionOne` in detail. We need to make some changes to the algorithm:
1. ~~In the workflow diagram, there is a `Scale` box at the top, it should not be there but at the end right after the question "Scale factor given?". Also, in the diagram it seems that if we `Rebin` in `Q` we won't apply the scaling, which is not true: the scaling is applied as long as a factor is given, independently of the `Rebin` in `Q`. #16671~~
2. ~~They don't want to have a `Rebin` in wavelength, `CropWorkspace` is fine but not the rebin, which should be removed. As it is going to be removed, the input property `WavelengthStep` is no longer needed, we should remove it too. #17162~~
3. ~~`ProcessingInstructions` need to be added to the flow chart and documented better. #17107~~
4. ~~The monitor should be optional. This was also requested by Tim and has been addressed in PR https://github.com/mantidproject/mantid/pull/15500~~
5. Dealt in #17085. We need an additional input property to apply flood corrections. This property was suggested to be called `DetectorEfficiencyCorrection`.
6. ~~`CreateTransmissionWorkspaceAuto` needs to be discussed at some point with them.~~
7. ~~Currently, if no transmission run is given no corrections are applied by default. Scientists would like the algorithm to read the type of corrections from the parameter file (see INTER for an example, where `polynomial` is the defined correction) and use it as a default. If no correction is found in this file, a warning message should be printed. #16518~~
8. **[Science validation required]** `CorrectDetectorPositions` should be defaulted to `False`. This has caused some confusion and it seems that in most of the cases they don't want to correct the detectors. #16650
9. ~~We need to document the difference between processing instructions specified with "-" and ",". Do they produce different results? We should add a section explaining this.  ##17161~~ 
10. ~~Add an option to specify whether the detector should be summed in lambda or Q.~~
11. To discuss with scientists: they may want the interface to be able to generate `QxQz` maps too. If that's the case, should this be added to `ReflectometryReductionOne` too?

~~The workflow diagram is getting really long, we may need to consider splitting it into different sections.~~

raquelalvarezbanos:
@MattKing06 Please have a look at this issue and split into different tasks. Let me know if you have any doubts.

MattKing06:
[...]

PR #16671, opened 2016-06-21
Include Scaling step in RRO Workflow Diagram

### LINKED ISSUE #16660: Fix workflow diagram for RRO
Currently, the workflow diagram for ReflectometryReductionOne (see [here](http://docs.mantidproject.org/nightly/_images/ReflectometryReductionOne-v1_wkflw.png)) is incorrect and in parts, incomplete.

Before we review this diagram as a whole in some of the issues listed in #16455 it should be fixed as it is currently.

MattKing06:
This is a sub-issue of #16455

**To test:**
- Build `ReflectometryReductionOne-v1` documentation
- Check the workflow diagram is now complete and Scaling step is included

Fixes #16660

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/docs/source/diagrams/ReflectometryReductionOne-v1_wkflw.dot
+++ b/docs/source/diagrams/ReflectometryReductionOne-v1_wkflw.dot
@@ -9,6 +9,7 @@ digraph ReflectometryReductionOne {
     secondTransRun    [label="SecondTransmissionRun"]
     outputWorkspaceWL [label="OutputWorkspaceWavelength"]
     outputWorkspaceMT [label="OutputWorkspace"]
+    outputWorksapceMTScale [label="OutputWorkspace"]
     thetaIn           [label="ThetaIn"]
     sourceRotationAngle  [label = "Rotation Angle"] 
    sourceRotationAngleBack [label = "-Rotation Angle"]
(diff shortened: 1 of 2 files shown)


raquelalvarezbanos:
@MattKing06 I think we should mention this in the release notes, as it was pointed out by scientists.

raquelalvarezbanos:
The Scaling step is displayed correctly. But now that I look at the diagram, I think the section between  `OutputWorkspace` and `Scale` is not correct. From the diagram it seems that only when `MomentumTransferMin` AND `MomentumTransferMax` AND `MomentumTransferStep` are given, we apply the `Rebin` step. But looking at the code, if the first two are not given we try to calculate them from the `IvsLam` workspace, and if the last one is not provided, we apply `CalculateResolution` (see [here](https://github.com/mantidproject/mantid/blob/master/Framework/Algorithms/src/ReflectometryReductionOne.cpp#L648-L669)). Could we update this section as part of this PR?

MattKing06:
@raquelalvarezbanos I agree, I'll see if I can fix this as part of this PR.

raquelalvarezbanos:
Thanks @MattKing06 I think the important thing is to clarify how these are calculated: `MomentumTransferMin` and `MomentumTransferMax` from the equation `Q = 4 \pi / \lambda * sin (\theta)` and `MomentumTransferStep` from `CalculateResolution`. It is also important to clarify that the `Rebin` in `Q` is always applied.

raquelalvarezbanos:
The workflow diagram is increasing in size but I think that is necessary if we want to document exactly what the algorithm does. Last section of the diagram (`Rebin` and `Scale` steps) looks more clear now in my opinion, it now documents how we calculate the necessary parameters for `Rebin` when those are not given by the user.

:shipit:

PR #16281, opened 2016-05-18
EnggGUI: Fixes and improvements for fitting tab

shahrozahmed1:
**Description of work.**

Various changes which were required on the `Fitting` tab after feedback from `Engin-X` scientists.
The plot of the focused workspace should now be plotting in crosses instead of lines, the focused workspace in the background should now be plotted in dark grey to improve to visibility of the single fit peaks. When fitting fails, the focused workspace should still be plotted in the background in `dSpacing` so that peaks on top can still be added. When file is selecting using browse button, the correct bank file should now be selected.
Improved layout of the tab; the peak tools to plot add new peaks have been differentiated in to separate group. `Save` and `Clear` button have been introduced next to the peak list.
A `Plot In Separate Window` button which will enable user to plot all peaks in `engggui_fitting_single_peaks` workspace

**To test:**

Go to: `Interfaces/Diffraction/Engineering Diffraction/Fitting`
`Browse` to the file and make sure the correct file selected is found within Focused Run# text-field

_Would be good idea testing invalid focused file e.g:_ (`ENGINX_228062_focused_bank_1`) _can be found under:_  `\\olympic\Babylon5\Public\Shahroz\FittingTest` 

Click `Clear` button to check whether the peaks list is cleared
Click `Save` button to check the peaks list is still being saved 
Input invalid peak and click `Fit`
Check that focused workspace is always displayed whether the fittings fails or not
Check that focused workspace is being plotted in `dSpacing` so that valid `dSpacing` peaks can be added to list
Check that `Plot In Separate Window` is disabled when fitting fails
This can be tested by: Selecting peak using clicking `Select Peak` and adding to list by clicking `Add Peak`
Click Fit and Check the peaks added have been fitted
Click `Plot In Separate Window` to check if `engggui_fitting_single_peaks` Workspace is plotted with appropriate title

Fixes #16194

---
#### Reviewer

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):
##### Code Review
- [x] Is the code of an acceptable quality?
- [x] Does the code conform to the coding standards? Is it well structured with small focussed classes/methods/functions?
- [ ] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [x] If there are changes in the release notes then do they describe the changes appropriately?
##### Functional Tests
- [x] Do changes function as described? Add comments below that describe the tests performed?
- [x] How do the changes handle unexpected situations, e.g. bad input?
- [x] Has the relevant documentation been added/updated?
[...]

PR #16211, opened 2016-05-16
Fittings peaks value can now be added when text-field is empty

shahrozahmed1:
**Description of work.**

On the fitting tab, when the `peaks` list/text-field is empty, user should now be able to add peaks to the list by clicking `Add Peaks` button. 

**To test:**
Go to: `Interfaces/Diffraction/Engineering Diffraction/Fitting`
Browse to: `ENGINX_241391_focused_texture_bank_1.nxs`
_can be found here:_ `\\olympic\Babylon5\Public\Shahroz\FittingTest`
Click `Fit`

Upon completion of process the plot should enable and click `Select Peak`
_(make sure that the `Peaks` text-field/list is empty if not then clear the text-field)_
Once desired peak selected click: `Add Peak`
- [x] Make sure that the selected peak/s has been added to the list by clicking `Fit` again

Fixes #16152
_Does not need to be in the release notes._

---
#### Reviewer

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):
##### Code Review
- [x] Is the code of an acceptable quality?
- [x] Does the code conform to the coding standards? Is it well structured with small focussed classes/methods/functions?
- [ ] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?
##### Functional Tests
- [x] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant documentation been added/updated?
- [ ] Is user-facing documentation written in a user-friendly manner?
- [ ] Has developer documentation been updated if required?
- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

PR #15991, opened 2016-04-26
Engg GUI: enable single peak fitting for consecutive run

shahrozahmed1:
**Description of work.**

Single peak fitting for consecutive run number's focused files has now been enabled. When a valid range of consecutive run numbers is given, the interface will automatically import and add the run number/s to the list view on the right side of the graph. The interface will then automatically update the Plot Bank combo-box according to the bank files found for each entered/selected run-number.

Also moved lot of Fitting logic related code to Presenter.

**To test:**

> Test 1

Files can be found at: `\\olympic\Babylon5\Public\Shahroz\FittingTest\`
Copy the files to focus output directory or set above directory as the focus output default directory via `settings` tab
Go to: `Interfaces/Diffraction/Engineering Diffraction/Fitting`
Enter `Focus Run #:` 194547-194551
The following run files should be found within the focused directory and imported on the right hand side of the interface within the list view widget
First run number and bank file should be selected by default
Make sure that all the banks are found within the combo-box upon selection each run number:

| Run | Bank |
| --- | --- |
| `194547` | `1,2,4,5,6` |
| `194548` | `1` |
| `194549` | `2` |
| `194550` | `1` |
| `194551` | `1,4` |

> Test 2

Enter `Focus Run #:` 194747
The focused file should still be imported on the right hand side of the interface within the list view widget but should be disabled
The bank files are still loaded correctly

> Other Tests

If one of the run file is missing, an appropriate warning should be displayed. 
Invalid run number range is handled appropriately
Other features still function as they should
Check the `EnggDiffractionPresenterTest` passes
Documentation has been updated

Fixes #15906 .
[Release Notes](https://github.com/mantidproject/mantid/blob/d0cd18a5f64db536eeaec47d1278f6d88acc4298/docs/source/release/v3.7.0/diffraction.rst)

---
#### Reviewer

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):
##### Code Review
- [x] Is the code of an acceptable quality?
- [x] Does the code conform to the coding standards? Is it well structured with small focussed classes/methods/functions?
- [x] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [x] If there are changes in the release notes then do they describe the changes appropriately?
##### Functional Tests
- [x] Do changes function as described? Add comments below that describe the tests performed?
- [x] How do the changes handle unexpected situations, e.g. bad input?
- [x] Has the relevant documentation been added/updated?
[...]

PR #16688, opened 2016-06-21
Engg GUI: split fitting tab into its own MVP widget

### LINKED ISSUE #16618: Engg GUI: extract the fitting tab into a widget of its own
As part of #15371.

This is purely about moving code around so that the fitting tab becomes a standalone widget.
The widget should then be usable from the Engg GUI.
A bit of communication from the Engg GUI to the "calibration tab widget" might be required to grab the current calibration parameters, as those are defined by the calibration tab.
Later on the widget should be more flexible/reusable so that it can be reused for the new multi-peak/whole pattern fitting tab.

FedeMPouzols:
This just moves code around. There should be no effective changes in functionality.

**To test:**

Check that the fitting tab still works as it should. Example files and test steps from previous PRs can be used (I can explain what files are required and the steps to to calibrate->focus->fit if you need to generate files from scratch):
- #16281 
- #16211 
- #15991

Fixes #16618.

---
#### Reviewer

##### Code Review
##### Functional Tests


MantidQt/CustomInterfaces/src/EnggDiffraction/EnggDiffFittingPresenter.cpp
```diff
@@ -0,0 +1,1126 @@
...
+                         singlePeaksWs + " or " + g_focusedFittingWSName +
+                         " workspace found.\n";
+    m_view->showStatus("Error while fitting peaks");
+    return;
+  }
+
+  try {
+    auto focusedPeaksWS =
+        ADS.retrieveWS<MatrixWorkspace>(g_focusedFittingWSName);
+    auto focusedData = ALCHelper::curveDataFromWs(focusedPeaksWS);
```
raquelalvarezbanos: I know this has not been introduced in this PR, but just wanted to say that, although this helper was initially written for the muon ALC interface, we should rename it and move it to a shared location if it is going to be useful for other interfaces.

MantidQt/CustomInterfaces/src/EnggDiffraction/EnggDiffFittingPresenter.cpp
```diff
@@ -0,0 +1,1126 @@
...
+                          "Please try again");
+    }
+  } else {
+    m_view->userWarning("Invalid Run Number",
+                        "The specfied range of run number "
+                        "entered is invalid. Please try again");
+  }
+}
+
+void EnggDiffFittingPresenter::processStart() {}
```
raquelalvarezbanos: Is this method intended to be used in the future?

FedeMPouzols: Yes. For example, at present you only get a preview plot after you run `Fit`. But a preview of the raw data before fitting is useful. The tab should show something on the preview plot if a run number is already selected (either in the current session or saved from previous sessions), without waiting for the user to run Fit. Functionality like that should be controlled/triggered from here.

MantidQt/CustomInterfaces/src/EnggDiffraction/EnggDiffFittingPresenter.cpp
FedeMPouzols: That could be done as part of the next reorganization of this code (move stuff to presenter and add tests) or as a separate issue/ticket. @shahrozahmed1.

raquelalvarezbanos:
Code looks good to me:

-  The original MPV pattern is now applied at a lower level: the "Fitting" tab has its own view and presenter (functionality is not completely split but current code will be re-organized in future issues).
- The "Fitting" presenter is able to get the calibration settings it needs to perform the fitting from the main view via the interface `IEnggDiffractionCalibration`.
- This PR is just the first step in a big refactoring in which the different tabs will use their own MVP pattern

:shipit:

shahrozahmed1:
`note:` while using this as a base branch I realised that the previous settings are no longer saved upon closure of the window

FedeMPouzols:
@shahrozahmed1: good catch. The shutdown call that saves the settings for the fitting tab was missing. It should be working well now. I merged the latest master in to sort the merge conflict out.

Issue #16480, opened 2016-05-31
Fit functions without documentation page


FedeMPouzols:
Every fit function should have a documentation page: http://docs.mantidproject.org/nightly/fitfunctions/index.html

There is no documentation for the following functions:
I've sectioned these out as best I can based on the history of the .h file.  Please create a pull request for each section.

- [ ] ComptonPeakProfile?, ComptonScatteringCountRate? GaussianComptonProfile? GramCharlierComptonProfile? (there is documentation for MultivariateGaussianComptonProfile)
- [ ] PoldiSpectrumConstantBackground, PoldiSpectrumDomainFunction, PoldiSpectrumLinearBackground, PoldiSpectrumPawleyFunction?
- [x] Polynomial #27412 
- [ ] MultiDomainFunction
- [ ] ReflectivityMulf
- [x] UserFunctionMD (there is documentation for UserFunction). #28328
- [x] FullprofPolynomial #28329
- [x] Bk2BkExpConvPV (but there is documentation for BackToBackExponential) #28331
- [x] ThermalNeutronBk2BkExpAlpha, ThermalNeutronBk2BkExpBeta, , ThermalNeutronBk2BkExpSigma, ThermalNeutronDtoTOFFunction (there is documentation for ThermalNeutronBk2BkExpConvPVoigt) #28330

You won't find the rst pages for these functions under: https://github.com/mantidproject/mantid/tree/master/docs/source/fitting/fitfunctions, so they are not found in the help browser or online pages.

Note that for some of these functions there might be old wiki pages.

For some of these functions (the groups with similar names) it may make sense as there is not much to say, but it would be good to have a documentation page that just states what is the exact difference with respect to the "base" function and link to it (if they are a product or convolution of the base with something else), or link to the important function if they are a component of a larger/more important function.

NickDraper:
Assigned to me to split out to authors and volunteers

ElliotAOram:
@NickDraper @FedeMPouzols I think Fick Diffusion is already there http://docs.mantidproject.org/nightly/fitfunctions/FickDiffusion.html

FedeMPouzols:
@ElliotAOram: true. I might have added it by mistake. I've taken it out.

FedeMPouzols:
Reopening this umbrella issue. I have marked in the description the functions that have been added.

PR #16698, opened 2016-06-22
Fix warnings and errors in docs build

martyngigg:
Fixes the warnings & errors that are present in a build of the documentation.

**To test:**

Remove the `build/docs/html` & `build/docs/doctrees` directories and build the documentation. Build the html documentation and check that there are no errors. Note that you will need `MAKE_VATES=ON` to avoid warnings around `SaveWorkspaceToVTK`.

No issue number.

---
#### Reviewer

##### Code Review
##### Functional Tests


AndreiSavici:
There are no warnings now. I've build the docs-qthelp instead.
I've set my Facility as `SNS`. I get several `IndirectDataReduction-[Warning] Failed to load instrument with error: Invalid value for property Filename (string) "/home/3y9/Mantid/mantid/instrument/VISION_Definition.xml": File "/home/3y9/Mantid/mantid/instrument/VISION_Definition.xml" not found. The current facility may not be fully supported.` and one `TomographyGUI-[Notice] Failed to initialize remote compute resource(s). This custom interface will not work. Error description: Failed to initialize because the facility is  SNS (and not ISIS). Facility not supported. This interface is designed to be used at ISIS. You will probably not be able to use it in a useful way because your facility is SNS. If you have set that facility by mistake in your settings, please update it.` messages. But they do not show up as sphinx warnings.

:shipit:

docs/source/algorithms/EnggFitPeaks-v1.rst
```diff
@@ -36,14 +36,13 @@ the Mantid :ref:`ConvertUnits <algm-ConvertUnits>`. See also
...
-:ref:`Back2BackExponential <func-Back2BackExponential>`. Other
+:ref:`Back2BackExponential <func-BackToBackExponential>`. Other
 alternatives might be added as optional in the future (for example the
 simpler :ref:`Gaussian <func-Gaussian>` or the more complex
-:ref:`Bk2BkExpConvPV <func-Bk2BkExpConvPV` or :ref:`IkedaCarpenterPV
-<func-IkedaCarpenterPV>`). To produce an initial guess for the peak
-function parameters this algorithm uses the :ref:`FindPeaks
-<algm-FindPeaks>` algorithm starting from the expected peaks list
-given in the inputs.
+:ref:`Bk2BkExpConvPV <func-NeutronBk2BkExpConvPVoigt>` or :ref:`IkedaCarpenterPV <func-IkedaCarpenterPV>`). To
```
FedeMPouzols: I just caught this by chance. We have a bit of confusion around these "back to back exponential" functions. Thanks for fixing these links!
Bk2BkExpConvPV is one of the undocumented functions: https://github.com/mantidproject/mantid/issues/16480.
NeutronBk2BkExpConvPVoigt has different parameters and can be more confusing than helpful. Maybe it's better to leave the text `Bk2BkExpConvPV` without a link?

martyngigg: Good catch. I thought I read in a file that Bk2BkExpConvPV had just been renamed but I must have misunderstood.

I think you are correct, it should not have a link if one does not exist. I'll take it out.

FedeMPouzols:
The rhel7 failure is totally unrelated to this.

PR #16707, opened 2016-06-23
ILL: Load multiple files

### LINKED ISSUE #16676: Multiple file loading for ILL instruments
### Expected behaviour
Loaders should be able to load multiple ILL files.
### Actual behaviour
Loaders are not able to load multiple ILL files, since the instrument prefix is absent from file names.
MantidPlot->Load->Select multiple ILL files->Run
### Platforms affected
All
### Facilities affected
ILL
### Instruments affected
All

gvardany:
ILL file names do not bear instrument prefix. Thus multiple file loading needs a special care.
Solution proposed by implementation of allowNoPrefix attribute in Facilities.xml, that is True for ILL.
Corresponding changes are delegated to MultiFileNameParser through FacilityInfo.

**To test:**

Load multiple files from an ILL instrument.
Load multiple files from non-ILL instrument.

Fixes #16676.

---
#### Reviewer

##### Code Review
##### Functional Tests


instrument/Facilities.xml
```diff
@@ -481,7 +481,7 @@
 
 </facility>
 
-<facility name="ILL" zeropadding="6" FileExtensions=".nxs,.hdf,.inx,.asc">
+<facility name="ILL" zeropadding="6" FileExtensions=".nxs,.hdf,.inx,.asc" allowNoPrefix="True">
```
martyngigg: These changes make sense but I'm not sure on the name of the attribute. How about `nofileprefix="True"`? The casing is more consistent (`FileExtensions` slipped through the net) and its slightly more direct about what it means.

Framework/Kernel/inc/MantidKernel/FacilityInfo.h
```diff
@@ -95,6 +95,9 @@ class MANTID_KERNEL_DLL FacilityInfo {
   /// Returns the catalogInfo class.
   const CatalogInfo &catalogInfo() const { return m_catalogs; }
 
+  /// Returns a bool indicating whether prefix is required in file names
+  const bool &allowNoPrefix() const { return m_allowNoPrefix; }
```
quantumsteve: Does returning a bool by const reference make sense? Further down, it looks like you are using the result immediately.

Framework/Kernel/src/FacilityInfo.cpp
```diff
@@ -59,6 +60,15 @@ void FacilityInfo::fillZeroPadding(const Poco::XML::Element *elem) {
   }
 }
 
+/// Called from constructor to fill the allowNoPrefix flag
+void FacilityInfo::fillAllowNoPrefix(const Poco::XML::Element *elem) {
+  std::string allowNoPrefixStr = elem->getAttribute("allowNoPrefix");
+  if (allowNoPrefixStr == "True")
+    m_allowNoPrefix = true;
+  else
+    m_allowNoPrefix = false;
```
quantumsteve: Instead of the branch, what do you think about just storing the result of the comparison?

Framework/Kernel/src/MultiFileNameParser.cpp
```diff
@@ -692,7 +692,10 @@ std::string getMatchingString(const std::string &regexString,
  */
 std::string pad(unsigned int run, const std::string &instString) {
   InstrumentInfo instInfo = ConfigService::Instance().getInstrument(instString);
-  std::string prefix = instInfo.filePrefix(run) + instInfo.delimiter();
+  FacilityInfo facilityInfo = instInfo.facility();
+  std::string prefix = "";
```
quantumsteve: This is identical to the default constructor.

Framework/Kernel/inc/MantidKernel/FacilityInfo.h
martyngigg: You're right. Returning by const reference for a primitive isn't necessary. Return by value is fine

Framework/Kernel/src/FacilityInfo.cpp
martyngigg: Yes, something like `m_allowNoPrefix = (allowNoPrefixStr == "True")` would eliminate the branch and still be readable,

martyngigg:
@gvardany Thanks for making those updates. It looks like the XML schema needs updating to specify that `nofileprefix` is now an optional attribute of the `facilty` tag.

gvardany:
@martyngigg @quantumsteve Thanks for the review, I have implemented the suggested changes.

Framework/DataHandling/test/LoadTest.h
```diff
@@ -233,6 +233,30 @@ class LoadTest : public CxxTest::TestSuite {
...
   }
 
+  /*
+   * This test loads and sums 2 IN4 runs from ILL
+   * without instrument prefix in the file names.
+   */
+  void test_ILLLoadMultipleFilesNoPrefix() {
+    Load loader;
+    loader.initialize();
+    loader.setPropertyValue("Filename", "ILL/IN4/IN4084446+084447.nxs");
```
martyngigg: Is this how we expect users would use this? Would is not be better to require that the directory containing the files is in the `datasearch.directories` path and then simply set `loader.setPropertyValue("Filename", "084446+084447.nxs");`

martyngigg:
Thanks, @gvardany. I've just made one more comment on the unit test.

Framework/Kernel/src/ConfigService.cpp
```diff
@@ -1582,6 +1582,44 @@ void ConfigServiceImpl::setDataSearchDirs(const std::string &searchDirs) {
...
+    subDirPath = Poco::Path(subdir);
+  } catch (Poco::PathSyntaxException &) {
+    return;
+  }
+
+  if (!subDirPath.isDirectory() || !subDirPath.isRelative()) {
+    return;
+  }
+
+  std::vector<std::string> newDataDirs = m_DataSearchDirs;
```
martyngigg: This is the right thing to do but I think we can tidy it up slightly with some C++11 syntax:
```c++
auto newDataDirs = m_DataSearchDirs;
for (const auto & path : m_DataSearchDirs) {
  Poco::Path newDirPath;
  try {
    newDirPath = Poco::Path(path);
    newDirPath.append(subDirPath);
    newDataDirs.push_back(newDirPath.toString());
  } catch (Poco::PathSyntaxException &) {
    continue;
  }
}
```

It saves having to worry about the loop termination and types of iterations etc.

martyngigg:
I'm happy with this now. When the tests pass I'll mark it ready to be merged.

martyngigg:
:shipit:

Framework/DataHandling/test/LoadTest.h
```diff
@@ -233,6 +254,31 @@ class LoadTest : public CxxTest::TestSuite {
...
+    std::string outputWS = "LoadTest_out";
+    loader.setPropertyValue("OutputWorkspace", outputWS);
+    TS_ASSERT_THROWS_NOTHING(loader.execute());
+
+    MatrixWorkspace_sptr output =
+        AnalysisDataService::Instance().retrieveWS<MatrixWorkspace>(outputWS);
+    MatrixWorkspace_sptr output2D =
+        boost::dynamic_pointer_cast<MatrixWorkspace>(output);
+
+    TS_ASSERT_EQUALS(output2D->getNumberHistograms(), 397);
```
FedeMPouzols: In this type of tests where settings like the search path, default instrument, facility, etc. in the config service are changed, I tend to like the test better if it resets those values to their defaults. In large `Test.h` files there could be some confusion. For example, if there is a clash of file names, etc. between different test cases, or if other tests were assuming different defaults.

FedeMPouzols:
Interesting exception. Maybe this should be documented/noted somewhere. I'm trying to remember where this implicit mechanism of appending instrument name to run number is explained, and only this wiki page came to mind: http://www.mantidproject.org/MultiFileLoading

Framework/DataHandling/test/LoadTest.h
martyngigg: @FedeMPouzols The `setUp` & `tearDown` methods should do that shouldn't they?

FedeMPouzols: Oh yes, they do. And they also clear the ADS. Perfect then, ignore my comment!

gvardany:
@FedeMPouzols I have added a note in limitations in http://www.mantidproject.org/MultiFileLoading

PR #16718, opened 2016-06-24
16612 histogram data rollout

SimonHeybrock:
**Do not merge, for review only**

### DIFF
--- a/Framework/Algorithms/src/AbsorptionCorrection.cpp
+++ b/Framework/Algorithms/src/AbsorptionCorrection.cpp
@@ -88,6 +88,38 @@ void AbsorptionCorrection::init() {
   defineProperties();
 }
 
+void linearlyInterpolateY(const Mantid::HistogramData::HistogramX &x,
+                          Mantid::HistogramData::HistogramY &y,
+                          const double stepSize) {
+  int specSize = static_cast<int>(y.size());
+  int xSize = static_cast<int>(x.size());
+  bool isHistogram(xSize == specSize + 1);
+  int step(static_cast<int>(stepSize)), index2(0);
+  double x1 = 0, x2 = 0, y1 = 0, y2 = 0, xp = 0, overgap = 0;
+
+  for (int i = 0; i < specSize - 1; ++i) // Last point has been calculated
+  {
+    if (step ==
+        stepSize) // Point numerically integrated, does not need interpolation
+    {
+      x1 = (isHistogram ? (0.5 * (x[i] + x[i + 1])) : x[i]);
+      index2 = static_cast<int>(
+          ((i + stepSize) >= specSize ? specSize - 1 : (i + stepSize)));
+      x2 = (isHistogram ? (0.5 * (x[index2] + x[index2 + 1])) : x[index2]);
+      overgap = 1.0 / (x2 - x1);
+      y1 = y[i];
+      y2 = y[index2];
+      step = 1;
+      continue;
+    }
+    xp = (isHistogram ? (0.5 * (x[i] + x[i + 1])) : x[i]);
+    // Linear interpolation
+    y[i] = (xp - x1) * y2 + (x2 - xp) * y1;
+    y[i] *= overgap;
+    step++;
+  }
+}
+
 void AbsorptionCorrection::exec() {
   // Retrieve the input workspace
   m_inputWS = getProperty("InputWorkspace");
@@ -144,8 +176,8 @@ void AbsorptionCorrection::exec() {
     PARALLEL_START_INTERUPT_REGION
 
     // Copy over bin boundaries
-    const MantidVec &X = m_inputWS->readX(i);
-    correctionFactors->dataX(i) = X;
+    auto &X = m_inputWS->x(i);
+    correctionFactors->mutableX(i) = X;
 
     // Get detector position
     IDetector_const_sptr det;
@@ -183,7 +215,7 @@ void AbsorptionCorrection::exec() {
     }
 
     // Get a reference to the Y's in the output WS for storing the factors
-    MantidVec &Y = correctionFactors->dataY(i);
+    auto &Y = correctionFactors->mutableY(i);
 
     // Loop through the bins in the current spectrum every m_xStep
     const auto lambdas = m_inputWS->points(i);
@@ -211,7 +243,8 @@ void AbsorptionCorrection::exec() {
         1) // Interpolate linearly between points separated by m_xStep,
            // last point required
     {
-      VectorHelper::linearlyInterpolateY(X, Y, static_cast<double>(m_xStep));
+	  //TODO Talk to Simon but this line was implemented in histogram
+      linearlyInterpolateY(X, Y, static_cast<double>(m_xStep));
     }
 
     prog.report();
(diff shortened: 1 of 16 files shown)


Framework/CurveFitting/src/FitMW.cpp
```diff
@@ -162,16 +163,16 @@ void FitMW::createDomain(boost::shared_ptr<API::FunctionDomain> &domain,
                          size_t i0) {
   setParameters();
 
-  const Mantid::MantidVec &X = m_matrixWorkspace->readX(m_workspaceIndex);
+  auto X = m_matrixWorkspace->points(m_workspaceIndex);
 
-  if (X.empty()) {
+  if (X.size() == 0) {
```
SimonHeybrock: I think I will add an `empty()` method, such that our new types are more compliant with `std::vector`.

Framework/Algorithms/src/CalculateFlatBackground.cpp
```diff
@@ -117,9 +118,7 @@ void CalculateFlatBackground::exec() {
     PARALLEL_FOR2(inputWS, outputWS)
     for (int i = 0; i < numHists; ++i) {
       PARALLEL_START_INTERUPT_REGION
-      outputWS->dataX(i) = inputWS->readX(i);
-      outputWS->dataY(i) = inputWS->readY(i);
-      outputWS->dataE(i) = inputWS->readE(i);
+      outputWS->setHistogram(i, inputWS->histogram(i));
```
SimonHeybrock: We have to be careful here. If `inputWS` is an `EventWorkspace`, does `histogram()` return the full histogram, or does it have `NULL` `Y` and `E`?

The question is not actually about this particular example, but in general;

- Should `MatrixWorkspace::histogram()` return a fully initialized `Histogram`? I think the answer should be yes.
- Given the above, do we provide a method for extracting the partial `Histogram` from an `EventWorkspace`?

Framework/Algorithms/src/CalculateFlatBackground.cpp
```diff
@@ -189,7 +189,7 @@ void CalculateFlatBackground::exec() {
             std::bind2nd(VectorHelper::AddVariance<double>(), variance));
       }
```
SimonHeybrock: It may be worth writing down places where you find this. Once we introduced `operator+` for our new types, we can then go back and replace this, without going through all the code once more.

Framework/Algorithms/src/CalculateFlatBackground.cpp
```diff
@@ -253,9 +253,9 @@ void CalculateFlatBackground::convertToDistribution(
                            ? 1
                            : workspace->getNumberHistograms();
 
-  MantidVec adjacents(workspace->readX(0).size() - 1);
+  MantidVec adjacents(workspace->x(0).size() - 1);
   for (std::size_t i = 0; i < total; ++i) {
-    MantidVec X = workspace->readX(i);
+    auto &X = workspace->x(i);
     // Calculate bin widths
     std::adjacent_difference(X.begin() + 1, X.end(), adjacents.begin());
```
SimonHeybrock: The following is another example of something we might provide in `HistogramData`, i.e., we should gather this in some sort of notes, so we do not forget.

Framework/CurveFitting/src/FitMW.cpp
```diff
@@ -198,15 +198,26 @@ void FitMW::createDomain(boost::shared_ptr<API::FunctionDomain> &domain,
...
+    std::vector<double> x(static_cast<size_t>(to - from));
+    auto it = from;
+    for (size_t i = 0; it != to; ++it, ++i) {
+      x[i] = (*it + *(it + 1)) / 2;
+    }
+    domain.reset(new API::FunctionDomain1DSpectrum(m_workspaceIndex, x));
+    x.clear();
+  } else {
+    domain.reset(new API::FunctionDomain1DSpectrum(m_workspaceIndex, from, to));
+  }
```
SimonHeybrock: So you did not trace down the problem?

LamarMoore: No I found that just replacing all references to the X,Y and E vectors with the HistogramX/Y/E types worked (without modifying the behaviour). Not sure why there is such a huge difference in the Fit results but I decided to make a note and move on to something else.

Framework/Algorithms/src/AbsorptionCorrection.cpp
```diff
@@ -88,6 +88,38 @@ void AbsorptionCorrection::init() {
...
+      continue;
+    }
+    xp = (isHistogram ? (0.5 * (x[i] + x[i + 1])) : x[i]);
+    // Linear interpolation
+    y[i] = (xp - x1) * y2 + (x2 - xp) * y1;
+    y[i] *= overgap;
+    step++;
+  }
+}
+
```
SimonHeybrock: Why did you duplicate this from `VectorHelpers` instead of changing the argument types there?

Framework/Algorithms/src/ApplyDeadTimeCorr.cpp
```diff
@@ -78,7 +78,7 @@ void ApplyDeadTimeCorr::exec() {
           boost::dynamic_pointer_cast<MatrixWorkspace>(temp);
 
       // Presumed to be the same for all data
-      double timeBinWidth(inputWs->dataX(0)[1] - inputWs->dataX(0)[0]);
+      double timeBinWidth(inputWs->mutableX(0)[1] - inputWs->mutableX(0)[0]);
```
SimonHeybrock: `x()`?

Framework/Algorithms/src/ApplyDeadTimeCorr.cpp
```diff
@@ -92,14 +92,15 @@ void ApplyDeadTimeCorr::exec() {
             for (size_t j = 0; j < inputWs->blocksize(); ++j) {
               double temp(
                   1 -
-                  inputWs->dataY(index)[j] *
+                  inputWs->mutableY(index)[j] *
```
SimonHeybrock: `x()`

Framework/Algorithms/src/ApplyDeadTimeCorr.cpp
```diff
@@ -92,14 +92,15 @@ void ApplyDeadTimeCorr::exec() {
...
                       (deadTimeRow.Double(1) / (timeBinWidth * numGoodFrames)));
               if (temp != 0) {
-                outputWs->dataY(index)[j] = inputWs->dataY(index)[j] / temp;
+                outputWs->mutableY(index)[j] =
+                    inputWs->mutableY(index)[j] / temp;
```
SimonHeybrock: `y()`

Framework/Algorithms/src/AbsorptionCorrection.cpp
LamarMoore: I wasn't sure if VectorHelpers were specifically used for Histogram type stuff. I imagined that if this is added to the underlying histogram datatype we could just remove the duplication regardless.

SimonHeybrock: I'd rather not add more duplication, just to remove it later. It is used in only 2 place. Either modify it in `VectorHelpers`, or move it to `HistogramData`.

PR #16724, opened 2016-06-24
Fix bug with MantidPlot test runner

### LINKED ISSUE #16721: MantidPlot tests returning incorrect error code
The MantidPlot test runner always returns a success error code. This error is a little subtle. Technically what it is doing is correct because the test runner (not the test) successfully completed. 

Discussed with @martyngigg what's the best solution to take here and the easiest fix is simply to call ```sys.exit(1)`` when the test runner fails instead of the Mantid exit call.

### Expected behaviour
MantidPlot python tests fail.

### Actual behaviour
MantidPlot python tests always succeed.

Easiest way would be to add ```self.fail()``` or an exception to any of the MantidPlot python tests. and run it with ctest and with the -V option. The test itself should show that it has failed, but ctest should report that the the test passed

### Platforms affected
All.

samueljackson92:
When the python tests for MantidPlot will currently always pass because the incorrect error code is being returned from within the bowls of ApplicationWindow. Technically it is doing the correct thing because the the _test runner_ finishes successfully, just not the actual tests.

Easiest solution here is to use the plain old python `sys.exit` to definitely close the program at this point. Everything we need to do including test clean ups should be done by this point so it shouldn't matter.

**To test:**
Easiest way to test is to pick a test case under `MantidPlot/test` and add a `self.fail()` to it. Run the unit test using ctest and using the verbose option. The verbose output should report that the test failed, but ctest will still say the test passed.

Test again with the changes applied and ctest should correctly report a failure.

Fixes #16721.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidPlot/test/mantidplottests.py
+++ b/MantidPlot/test/mantidplottests.py
@@ -58,10 +58,10 @@ def runTests(classname):
     QtCore.QCoreApplication.processEvents()
     QtCore.QCoreApplication.processEvents()
 
-    # Set Mantid exit code
+    # Close Mantid and set exit code
     if not res.wasSuccessful():
-        _qti.app.setExitCode(1)
+        sys.exit(1)
     else:
-        _qti.app.setExitCode(0)
+        sys.exit(0)
 
     return res


martyngigg:
@samueljackson92 This looks good. Just a small thing for when you create your next pull request - leave it unassigned. When a developer goes to test a PR they will assign it to themselves.

martyngigg:
The code changes are small & sensible.

I tried modifying a test first with a syntax error and then with a run-time error. Both causes MantidPlot to report a non-zero exit status as required.

:shipit:

PR #16711, opened 2016-06-23
Reorganize tests of fit functions, category background

FedeMPouzols:
For the background functions, move the unit test cases that run the Fit algorithm to FitTest, and reorganized/extended the function tests.

**To test:**
- Code review
- Tests should pass on all platforms
- Ensure that:
  - The algorithm `Fit` is not used in the function unit tests, and the old cases that were using `Fit` are now in `FitTest`
  - No additional dependencies/headers/using are added in `FitTest`, and many headers/using are removed from the function unit tests.
- This also added some basic tests to evaluate functions, and check that wrong attribute/parameter values are handled gracefully. Otherwise some unit tests would have been left empty. It also adds a test on the category(ies) wherever it was missing. Also removes calls to AnalysisDataService::Instance().addOrReplace() / AnalysisDataService::Instance().remove() that are unnecessary and not what these tests are about.

Fixes #16479
This is one more step of #16267

_Does not need to be in the release notes._

---
#### Reviewer

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):
##### Code Review
- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)? Is it well structured with small focussed classes/methods/functions?
- [ ] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?
##### Functional Tests
- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant documentation been added/updated?
- [ ] Is user-facing documentation written in a user-friendly manner?
- [ ] Has developer documentation been updated if required?
- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

Issue #16267, opened 2016-05-17
Change tests of `CurveFitting` "functions" to be actual unit tests


NickDraper:
### Expected behavior

We want to have unit tests that just test the unit aspect of the code. Including boundary conditions etc.

Complex tests of the complete fitting framework should move out tu the Fit unit testing or a system test.
### Actual behavior

Currently the unit tests for many of the functions test much more than the function itself, often performing optimised fits.

See [this comment below](#issuecomment-221586507) for a list of individual issues.

FedeMPouzols:
@raquelalvarezbanos

FedeMPouzols:
Dividing into blocks:

- Functions in the category "Muon": #16418.

raquelalvarezbanos:
Functions in the category "Peaks": https://github.com/mantidproject/mantid/issues/16452

FedeMPouzols:
@NickDraper: I just pushed this back to 3.9. This is 95% done, just the "QuasiElastic" functions need a bit of attention.

gemmaguest:
Closing as stale because we are unlikely to get around to this

PR #16725, opened 2016-06-24
Reorganize tests of fit functions, category general, chunk 2

### LINKED ISSUE #16484: Reorganize unit tests of functions, category "General", chunk 2
This is one chunk of #16267.
Move the test cases that use the algorithm Fit to the unit test file of Fit (FitTest). Add tests to test the function itself when necessary. Tidy up the includes wherever possible, as for example in #16420.

Functions included in this chunk:

- MultiDomainFunction, 
- MultivariateGaussianComptonProfile, 
- NeutronBk2BkExpConvPVoigt, 
- PawleyFunction, 
- PawleyParameterFunction, 
- PeakParameterFunction, 
- PoldiSpectrumConstantBackground, 
- PoldiSpectrumDomainFunction, 
- PoldiSpectrumLinearBackground, 
 -PoldiSpectrumPawleyFunction

FedeMPouzols:
This is now ready for testing:
- [x] Merge #16711 first.

Moves the unit test cases that run the Fit algorithm to FitTest, and reorganizes the function tests.

**To test:**
- Code review
- Tests should pass on all platforms. Some tests that were disabled have been re-enabled.
- Ensure that:
  - The algorithm `Fit` is not used in the function unit tests, and the old cases that were using `Fit` are now in `FitTest`
  - No additional dependencies/headers/using are added in `FitTest`, although there is an exception with `PawleyFunction` that I think is not easy to avoid.
  - Other changes remove unneeded headers, prevent leaks, use initialization lists, etc.

Fixes #16484.
This is the second last step of #16267

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/CurveFitting/test/Algorithms/FitTest.h
```diff
@@ -1028,6 +1054,209 @@ class FitTest : public CxxTest::TestSuite {
...
+      }
+      for (size_t i = 10; i < 20; ++i) {
+        x[i] = dx * double(i);
+        y[i] = A1;
+      }
+      for (size_t i = 20; i < 30; ++i) {
+        x[i] = dx * double(i);
+        y[i] = A0 + B0 * x[i];
+      }
+    }
```
samueljackson92: Why the extra set of braces here?

samueljackson92:
Looks good to me :shipit:

martyngigg:
It's good that the function tests are becoming slimmer but I'm slightly concerned that `FitTest` is now becoming a god-like test and is certainly not a unit test. Is there a plan to move these tests out? With our current testing layout they really are more along the lines of system tests.

It feels like `FitTest` should just use either mock of fake minimizers, cost functions etc to check it is working as expected - it shouldn't depend on the other functions too.

FedeMPouzols:
Yes, I think it is good that all this stuff has been moved from the individual function tests but now `FitTest` is bloated and definitely not a good example.

In an ideal world the next step would be to move all or nearly all the tests that fit a function to a system test, now that the code is reorganized and uses function strings instead of specific function objects, etc.

Now, knowing the platform dependence issues that we are usually facing with fitting (+ the GSL version issues), I'm not sure if that's a good idea in practice with our current setup. FitTest as it is runs relatively fast (1.6 seconds I think) and on all platforms, so it's effective to catch issues early.

martyngigg:
It maybe that we need to start thinking about integration tests that sit between unit and system tests for things like this. We can then still write complex C++ tests like this but we also can have the basic unit tests as they should be.

FedeMPouzols:
I think that this has introduced a compilation error in master (I'd suspect there must have been some clash with the big changes of the new Histogram data type).  It seems easy to fix, I'll make now a quick fix PR.

FedeMPouzols:
`CurveFitting` really looks like the kind of sub-system that asks for integration tests. Some tests need to check functions, constraints, cost functions, and a minimizer. LevenbergMarquardTest is an example of that. Then the tests of the Fit algorithm on top of that are/should be testing all those components together.

PR #16743, opened 2016-06-27
Fix method calls from C++ to Python for plugins in Python 3

martyngigg:
Fixes existence checks of Python methods from C++ in Python 3 by using a unicode object rather than a raw bytes object.

**To test:**

For Python 2 the build servers will tell you if this doesn't work.

For Python 3 there should be no warnings such as:

```
Python-[Warning] Error initializing SANSFitShiftScale on registration: 'SANSFitShiftScale has no function named 'PyInit'
Check the function exists and that its first argument is self.'
```

on start up. There will still be other warnings on start up but they will be addressed in the future and require this to be fixed first.

---
#### Reviewer

##### Code Review
##### Functional Tests


rosswhitfield:
Yes the warnings are gone. 😄  And now I'm able to start the python algorithms in Mantid that are already python 3 compatible. :shipit:

PR #16757, opened 2016-06-29
Remove use of workflow algorithm from CalculateEfficiency

### LINKED ISSUE #16710: Remove SANSMask from CalculateEfficiency
CalculateEfficiency now calls SANSMask, which is a workflow algorithm. 
Normal algorithms shouldn't be allowed to call workflow algorithms.

SANSMask also needs to be refactored. It's getting quite complex to understand.

ricleal:
Description of work.

Removed workflow algorithm that was being called by a non workflow algorithm.

**Test** 

``` python
Load(Filename='/HFIR/CG3/shared/data/CG3_mantid_test/BioSANS_exp318_scan0008_0001.xml', OutputWorkspace='BioSANS_exp318_scan0008_0001')
CalculateEfficiency(InputWorkspace='BioSANS_exp318_scan0008_0001', OutputWorkspace='effw', MinEfficiency=0.5, MaxEfficiency=1.5, MaskedFullComponent='wing_detector')
# See wing_detector masked in the instrument view

Load(Filename='/HFIR/CG3/shared/data/CG3_mantid_test/BioSANS_exp318_scan0008_0001.xml', OutputWorkspace='BioSANS_exp318_scan0008_0001')
CalculateEfficiency(InputWorkspace='BioSANS_exp318_scan0008_0001', OutputWorkspace='effw', MinEfficiency=0.5, MaxEfficiency=1.5, MaskedFullComponent='detector1')
# See detector1 masked in the instrument view
```

Fixes #16710.

---
#### Reviewer

##### Code Review
##### Functional Tests


mdoucet:
This is a small refactor. Everything looks good.
:shipit:

Framework/Algorithms/src/CalculateEfficiency.cpp
```diff
@@ -273,5 +264,37 @@ void CalculateEfficiency::normalizeDetectors(MatrixWorkspace_sptr rebinnedWS,
...
+/**
+ * Fully masks one component named componentName
+ */
+void CalculateEfficiency::maskComponent(MatrixWorkspace &ws,
+                                        const std::string &componentName) {
+  auto instrument = ws.getInstrument();
+  try {
+    boost::shared_ptr<const Geometry::ICompAssembly> component =
+        boost::dynamic_pointer_cast<const Geometry::ICompAssembly>(
+            instrument->getComponentByName(componentName));
```
FedeMPouzols: Would it be safer to check here that `component` is fine (as they do below for `detector`)?

Framework/Algorithms/src/CalculateEfficiency.cpp
```diff
@@ -273,5 +264,37 @@ void CalculateEfficiency::normalizeDetectors(MatrixWorkspace_sptr rebinnedWS,
...
+    std::vector<detid_t> detectorList;
+    for (int x = 0; x < component->nelements(); x++) {
+      boost::shared_ptr<Geometry::ICompAssembly> xColumn =
+          boost::dynamic_pointer_cast<Geometry::ICompAssembly>((*component)[x]);
```
FedeMPouzols: This dynamic cast would also need a check?

Framework/Algorithms/src/CalculateEfficiency.cpp
```diff
@@ -273,5 +264,37 @@ void CalculateEfficiency::normalizeDetectors(MatrixWorkspace_sptr rebinnedWS,
...
+      for (int y = 0; y < xColumn->nelements(); y++) {
+        boost::shared_ptr<Geometry::Detector> detector =
+            boost::dynamic_pointer_cast<Geometry::Detector>((*xColumn)[y]);
+        if (detector) {
+          auto detID = detector->getID();
+          detectorList.push_back(detID);
+        }
+      }
+    }
+    std::vector<size_t> indexList = ws.getIndicesFromDetectorIDs(detectorList);
```
FedeMPouzols: This could use `auto`?

AndreiSavici:
Why is @ricleal assigned to an issue that he worked on?

ricleal:
@FedeMPouzols I'm fixing some of the issues you mentioned. However I'm against exaggerated checking. When the function `maskComponent` is called, at this point, you know already that we are expecting a detector `CompAssembly`. A sanity initial check is enough in my opinion. Checking every tube and pixel will make the code slower... In my opinion the `if (detector)` is redundant but I left it there because I copied this code from somewhere else...

FedeMPouzols:
I see. I have the feeling that the design of these geometry classes force the client code to use too many casts.  Actually just the fact that you need to do a dynamic cast on every single element must be slowing down the code quite a bit. Let's merge it in as it looks good. I'm not sure if coverity will complain about the casts, though. Let's see.

PR #16761, opened 2016-06-30
Remove string comparison from method in PredictPeaks

### LINKED ISSUE #16758: Slightly change the crystallography convention check in PredictPeaks
For fixing #16722, a string comparison was added to `PredictPeaks:: calculateQAndAddToOutput`. In another place in the same file, a variable `qSign` is used. I think it would be better to only store this `qSign` in the first place instead of the `convention` and multiply `q` in that method with that factor. The method is called many times (as many times as there are HKLs, and that can be several ten- or even hundred thousand) and it's not necessary to do this comparison for each HKL separately, because the Q convention can't change during algorithm execution.

MichaelWedel:
Fixes #16758.

The `Q.convention` string is now used once to determine a multiplication factor (1 or -1) for going from `HKL` to `Q`. The check for the case of the PeaksWorkspace with a stored convention is still the same, but the logic has been moved into a small free function so that it can be reused for determining the factor in the algorithm's constructor.

The unit tests should still pass.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Crystal/src/PredictPeaks.cpp
+++ b/Framework/Crystal/src/PredictPeaks.cpp
@@ -22,13 +22,26 @@ using namespace Mantid::DataObjects;
 using namespace Mantid::Geometry;
 using namespace Mantid::Kernel;
 
+namespace {
+/// Small helper function that return -1 if convention
+/// is "Crystallography" and 1 otherwise.
+double get_factor_for_q_convention(const std::string &convention) {
+  if (convention == "Crystallography") {
+    return -1.0;
+  }
+
+  return 1.0;
+}
+}
+
 //----------------------------------------------------------------------------------------------
 /** Constructor
  */
 PredictPeaks::PredictPeaks()
-    : m_runNumber(-1), m_inst(), m_pw(), m_sfCalculator() {
+    : m_runNumber(-1), m_inst(), m_pw(), m_sfCalculator(),
+      m_qConventionFactor(get_factor_for_q_convention(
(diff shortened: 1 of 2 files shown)


SimonHeybrock:
Looks good, also code comments were improved. Probably the function to obtain the Q convention factor could be moved to a more central place where it can be used by other algorithms as well, but that is beyond the scope of this PR.

:shipit:

Framework/Crystal/src/PredictPeaks.cpp
```diff
@@ -398,16 +414,13 @@ void PredictPeaks::setStructureFactorCalculatorFromSample(
...
-void PredictPeaks::calculateQAndAddToOutput(V3D &hkl,
+void PredictPeaks::calculateQAndAddToOutput(const V3D &hkl,
                                             const DblMatrix &orientedUB,
                                             const DblMatrix &goniometerMatrix) {
   // The q-vector direction of the peak is = goniometer * ub * hkl_vector
   // This is in inelastic convention: momentum transfer of the LATTICE!
   // Also, q does have a 2pi factor = it is equal to 2pi/wavelength.
-  if (convention == "Crystallography") {
-    hkl = hkl * (-1.0);
-  }
```
VickieLynch: By changing hkl to a const and not changing the sign, the output of PredictPeaks has the wrong sign of hkl when using Crystallography Convention.  Fixing this in #18149.  Please ask me or Andrei to check any changes to Crystallography Convention.

PR #16765, opened 2016-06-30
Fix undefined reference when building with the undefined behavior sanitizer.

quantumsteve:
Description of work.

Building Mantid with ubsan fails with the following error

```
23:01:27 /usr/bin/ld: CMakeFiles/CustomInterfacesTest.dir/ALCPeakFittingPresenterTest.cpp.o: undefined reference to symbol '_ZTI7QwtData'
23:01:27 //usr/lib/libqwt-qt4.so.5: error adding symbols: DSO missing from command line
```

I think this is because ALCPeakFittingPresenterTest.h needs to know more about `&QwtData::size`. (One can verify this by removing `#include <qwt_data.h>` in ALCPeakFittingPresenterTest.h.) This can be fixed by adding `${QWT_LIBRARIES}` to the list of dependencies. 

I'm not sure why this error doesn't happen in the normal build. Perhaps ubsan is just a bit pickier about linking?

In doing this work, I also found a place where a forward declaration of `QwtData` was sufficient.

**To test:**

Instructions for building and running UBSAN are [here](http://www.mantidproject.org/Sanitizers)

Ideally this should be done on a system with GCC 5 or later.

There is no GitHub issue associated with the PR.

---
#### Reviewer

##### Code Review
##### Functional Tests


peterfpeterson:
This is a low risk change and the builds still pass. :shipit:

PR #16802, opened 2016-07-05
Algorithm to export ascii masks

abuts:
This is widely used in ISIS direct inelastic group algorithm which exports workspace masks, and saves them into ASCII file as special format file (.msk). The algorithm was used by group for ages and I just pull it out of qtiGenie and wrote Python algorithm wrapper, unit tests and documentation. As it is old working algorithm, I have not actually changed its core code.

**To test:**
Its small and simple algorithm, so look if the description is clear, all code conventions are satisfied and you can not break it by providing wrong data.

---
#### Reviewer

##### Code Review
##### Functional Tests


KarlPalmen:
When I run it from Mantidplot on the workspace created by the usage example. It fails to create a file in my default save directory, but does produce any error messages or warnings. I use Windows 7.

Framework/PythonInterface/plugins/algorithms/ExportSpectraMask.py
```diff
@@ -170,12 +170,11 @@ def category(self):
...
         """ Return name
         """
-        return "ExportASCIIMask"
+        return "ExportSpectraMask"
 
     def summary(self):
-        return "Exports workspace masks as legacy ASCII file with .msk extension, "\
-            "containing list of masked spectra numbers, "\
-            "in the ISIS ASCII format described in LoadMask algorithm"
+        return "Returns list of spectra mnumbers which are masksed"\
```
KarlPalmen: "numbers" and "masked" have typos.

Framework/PythonInterface/plugins/algorithms/ExportSpectraMask.py
```diff
@@ -184,10 +183,10 @@ def PyInit(self):
         self.declareProperty("Filename","","Name of the file to save mask to."\
                             " If empty, the name of the input workspace is used")
         self.declareProperty("ExportMaskOnly",False,"If true, algorithm will not save mask in a file"\
-                             "and returns only list with numbers of masted spectra.",\
+                             "and only returns the list containing numbers of masted spectra.",\
                               Direction.Input)
```
KarlPalmen: "masted" -> "masked".

abuts:
All fixed -- it was using wrong target directory if just filename is provided.

KarlPalmen:
Works OK and looks good. :shipit:

martyngigg:
This seems like a useful little routine for ISIS. Would it not be better as a different format option on `SaveMask`? It seems inconsistent that you use `SaveMask` for XML and then `ExportSpectraMasks` for ASCII.

abuts:
I am hoping to get rid of Save ASCII masks and change to xml eventually. Unfortunately, it brings up the range of issues related to the fact that the definition of spectra-detector map is not a part of an instrument definition but is always 1:1 and  **bound to some implicit order**, which is wrong for all ISIS instruments. Number of steps is necessary to overcome this and I am making some now. 

This is why I consider this algorithm as a temporary helper rather then permanent solution and as SaveMask is C++ and this one is existing Python, converting to C++ (or calling Python from C) at the moment seems too much hustle. Will see how people use it but its possible that it will just eventually die.

docs/source/algorithms/ExportSpectraMask-v1.rst
```diff
@@ -14,6 +14,12 @@ if requested, saves these spectra as legacy *.msk* file in ISIS ASCII format.
 
 The format is described on :ref:`LoadMask <algm-LoadMask>` algorithm pages.
 
+**Note** It is recommended not use exported *.msk* files for masking workspaces during reduction as
+the result of applying such mask will be different depending on the stage of reduction of the source and target files.
+You need to make additional efforts (see :ref:`LoadMask <algm-LoadMask>` algorithm description) to avoid this effect.
+It is better to use :ref:`SaveMask <algm-SaveMask>` algorithm instead. This algorithm produces *.hml* files, 
+containing masked detectors lists, which will produce identical results regardless of the workspace grouping.
```
KarlPalmen: Did you mean '.hml' files?  I thought they were '.xml' files.

abuts: yep, right note -- fixed...

martyngigg:
Now that the note has been added I think this can go in.

PR #16809, opened 2016-07-06
Replace QTableView with QTreeView

### LINKED ISSUE #16734: Reflectometry Interface: replace QTableView with QTreeView
Reflectometry scientists have requested a change in the way the data is displayed, they would like to use a `QTreeView` instead of the current `QTableView`. The reason is that at some point they are going to measure runs every few minutes and they may end up with hundreds of them in the table. They would like to hide sub-items (i.e. they would like to see the group only), and only inspect a group if something goes wrong with the reduction.

newtreeview

raquelalvarezbanos:
Description of work.

**To test:**
- Check code changes: relevant classes are `QDataProcessorTreeModel` (comes with unit tests), which replaces `QDataProcessorTableModel`. In `GenericDataProcessorPresenter`, check the actions (`insertRows()`, `insertGroups()`, `deleteRow()`, etc)
- Ensure that old TBL files can be loaded correctly (a sample TBL file and some sample datasets can be found here: [SampleData-ISIS.zip](https://github.com/mantidproject/mantid/files/361036/SampleData-ISIS.zip))
- Ensure that you can save a TBL file and load it again.
- Ensure runs can be reduced. For the reduction to work you will have to set `Params="-0.04"` in the textbox `Stitch1DMany`. Selecting runs (child items) will just reduce the selected runs. Selecting groups will reduce runs and post-process them.
- Check that actions work as you would expect. Actions can be accessed from the toolbar and from the outer menus. 
- Check changes to the documentation.

Fixes #16734 

Release notes were updated.

---
#### Reviewer

##### Code Review
##### Functional Tests


FedeMPouzols:
This branch seems to have developed a merge conflict with master.

raquelalvarezbanos:
Thanks @FedeMPouzols

Framework/DataHandling/src/LoadTBL.cpp
```diff
@@ -320,24 +320,24 @@ void LoadTBL::exec() {
...
+    colStitch->setPlotType(0);
     colRuns->setPlotType(0);
     colTheta->setPlotType(0);
     colTrans->setPlotType(0);
     colQmin->setPlotType(0);
     colQmax->setPlotType(0);
     colDqq->setPlotType(0);
     colScale->setPlotType(0);
-    colStitch->setPlotType(0);
     colOptions->setPlotType(0);
```
FedeMPouzols: If the aim here is to set PlotType as 0 for all the columns a loop would be neater and would prevent possible issues with future changes.

MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/DataProcessorUI/DataProcessorAppendGroupCommand.h
```diff
@@ -0,0 +1,48 @@
...
+#include "MantidQtMantidWidgets/DataProcessorUI/DataProcessorCommandBase.h"
+#include "MantidQtMantidWidgets/DataProcessorUI/DataProcessorPresenter.h"
+
+namespace MantidQt {
+namespace MantidWidgets {
+/** @class DataProcessorAppendGroupCommand
+
+DataProcessorAppendGroupCommand defines the action "Insert Group"
+
+Copyright &copy; 2011-14 ISIS Rutherford Appleton Laboratory, NScD Oak Ridge
```
FedeMPouzols: !

MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/DataProcessorUI/DataProcessorPresenter.h
```diff
@@ -1,10 +1,10 @@
 #ifndef MANTIDQTMANTIDWIDGETS_DATAPROCESSORPRESENTER_H
 #define MANTIDQTMANTIDWIDGETS_DATAPROCESSORPRESENTER_H
 
+#include <QVariant>
```
FedeMPouzols: Would a forward declaration of QVariant be enough here?

MantidQt/MantidWidgets/src/DataProcessorUI/DataProcessorGenerateNotebook.cpp
```diff
@@ -74,27 +74,27 @@ DataProcessorGenerateNotebook::DataProcessorGenerateNotebook(
...
   auto notebook = Mantid::Kernel::make_unique<Mantid::API::NotebookWriter>();
 
   notebook->markdownCell(titleString(m_wsName));
 
   notebook->markdownCell(tableString(m_model, m_whitelist, rows));
 
-  int groupNo = 1;
-  for (auto gIt = groups.begin(); gIt != groups.end(); ++gIt, ++groupNo) {
-    const std::set<int> groupRows = gIt->second;
+  for (auto gIt = rows.begin(); gIt != rows.end(); ++gIt) {
```
FedeMPouzols: Could this be a range based loop? And with `const auto &`?

MantidQt/MantidWidgets/src/DataProcessorUI/DataProcessorGenerateNotebook.cpp
```diff
@@ -258,24 +259,35 @@ std::string tableString(QDataProcessorTableModel_sptr model,
...
-          << model->data(model->index(*rowIt, col)).toString().toStdString()
-          << " | ";
+    auto groupId = groupIt->first;
+    auto rowSet = groupIt->second;
 
-    table_string
-        << model->data(model->index(*rowIt, ncols - 1)).toString().toStdString()
-        << "\n";
-  }
+    for (auto rowIt = rowSet.begin(); rowIt != rowSet.end(); ++rowIt) {
```
FedeMPouzols: Not sure if the interface of the Qt index() methods will allow it, but this could also be a `const auto &` range-based loop? Or use `cbegin()`, `cend()`?

MantidQt/MantidWidgets/src/DataProcessorUI/GenericDataProcessorPresenter.cpp
```diff
@@ -844,108 +841,172 @@ void GenericDataProcessorPresenter::reduceRow(int rowNo) {
...
 
-  // Now we just have to set the group id on the selected rows
-  for (auto it = rows.begin(); it != rows.end(); ++it)
-    m_model->setData(m_model->index(*it, m_columns - 2), groupId);
+  // Find if rows belong to the same group
+  // If they do, do nothing
+  // If they don't, remove rows from their groups and add them to a
+  // new group
+
+  auto selectedRows = m_view->getSelectedRows();
```
FedeMPouzols: Maybe this could be `const` if the loops below used also `const` iterators / `const` range-based.

MantidQt/MantidWidgets/src/DataProcessorUI/GenericDataProcessorPresenter.cpp
```diff
@@ -1186,51 +1250,43 @@ void GenericDataProcessorPresenter::afterReplaceHandle(
...
-
-    for (int i = 0; i < m_columns - 2; i++) {
-      m_model->setData(m_model->index(*row, i), "");
+  auto selectedRows = m_view->getSelectedRows();
+
+  for (auto group = selectedRows.begin(); group != selectedRows.end();
+       ++group) {
+    int groupIndex = group->first;
+    auto rows = group->second;
+    for (auto row = rows.begin(); row != rows.end(); ++row) {
```
FedeMPouzols: More opportunities for `const` in this loop and its parent?

MantidQt/MantidWidgets/test/DataProcessorUI/DataProcessorGenerateNotebookTest.h
```diff
@@ -387,6 +389,38 @@ class DataProcessorGenerateNotebookTest : public CxxTest::TestSuite {
...
+        "MomentumTransferMaximum = "
+        "2.9, "
+        "MomentumTransferStep = 0.04, ScaleFactor = 1)",
+        ""};
+
+    std::vector<std::string> notebookLines;
+    boost::split(notebookLines, boost::get<0>(output), boost::is_any_of("\n"));
+
+    int i = 0;
+    for (auto it = notebookLines.begin(); it != notebookLines.end();
```
FedeMPouzols: `const` range-based loop?

FedeMPouzols:
Looks very good to me. Generalizes the table model to a tree model. The tree model methods data(), index(), insertRows() etc. handle error conditions. This also adds a new "data processor" command to append a group. It also changes the "prepend row" command to "delete group".
This interface already had good mocked test coverage, and the suite of unit tests has been updated and extended. The new model is tested exhaustively.

Functionally (tested on debian):
- LoadTBL works well as standalone algorithm and when used from the "Polref" interface
- The new tree interface works well. All the runs given in the table are displayed in their groups.
- The reduction works well for the parameters given in the PR description. If I omit or change the parameters, the algorithms fail gracefully.
- All the actions behave normally as far as I can tell
- The documentation looks good and easy to follow

With the groups there may be some actions for which it is not clear what the result should be. For example let's see what happens if I add a few rows to the second group of the test table, then Ctrl-select some of those new rows and the rows of the first group, then use the action "group selected rows". The grouping happens but the new/extended group looses its name. Am I doing something forbidden here?

About the actions: can/should you be able to cut a group? If I remove all the runs in a group, should the group be removed? Maybe for the future, some actions should be disabled when the selection is a group, or an alternative behavior could be defined for groups. I'm thinking for example what should happen when I choose "plot selected rows" with a group as current selection.

Just noticed a minor glitch in the Refl/Polref interface. If you have an empty table / afresh open interface and you load a table the interface will tell you 
"your current table has unsaved changes. Are you sure...?"

Another minor glitch is that if you select the first run in the first group (which is listed in the table/tree as "1", the interface will tell you
"You have only selected 1 of the 2 rows in group 0". That could be confusing.

For later, was there a plan to give more informative names to the groups? Not sure how the interface would figure out those names, maybe from the run titles?

Also for the future, maybe it would be convenient to add the capability to remember the last parameters that the user had given in the "Post-process", "Pre-process", etc. fields below the table.

I'm also wondering if this interface should have a bottom band with the "question mark"/help button on the left and a close button on the right. I think that was some kind of standard or wish for harmonizing custom interfaces in Mantid.

raquelalvarezbanos:
@FedeMPouzols thank you very much for your comments, I will try to answer to your questions:

1. No, you are not doing anything wrong. The reason why the group loses its name is because it is replaced by a new one, and as it is new, it does not have a name defined. I am not sure how scientists want to define this action (it's been inherited from the `QTableView`), so I will leave it this way and change it in the future according to their feedback.

2. Option to cut a group and disabling/enabling options depending on selection: I agree it's a good idea, but I didn't want to introduce too many new options and features in this PR. When scientists test the interface and give their feedback we will re-arrange/re-define/add/remove options, depending on which are more important/convenient.

3. I'll fix that. It should not happen.

4. Thanks, I'll change the warning message.

5. Yes, there's a future plan to name groups. Scientists want to name them according to the sample and other parameters stored in the run. But that's for a future PR :)

6. Yes, scientists will definitely want to load previous reductions, but I am not sure how yet (ideally they should use the notebook that is optionally written once the reduction is complete, which already includes pre-processing, processing and post-processing options).

7. Probably... But I don't know why it was done this way, maybe there was a good reason.

MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/DataProcessorUI/DataProcessorPresenter.h
raquelalvarezbanos: Can I use a forward declaration here? I am using a `std::map<std::string, QVariant>`

FedeMPouzols: Then I'm afraid that `std::map` needs the complete definition. I'd leave it as it is then.

FedeMPouzols:
This looks good. I gave it a second check after the last polishing commits and it worked well for me.
Let's ship-it so users can start using this new feature and providing feedback.

:shipit:

PR #16823, opened 2016-07-08
User-defined background subtraction algorithm

### LINKED ISSUE #16798: User-defined background subtraction
*Requested by Adrian Hillier*

#### Use case:

Dataset consists of a number of peaks on top of a large background. The background is not flat and can have bumps and other features in it, so is difficult to fit. The end goal is to remove the background and leave just the peaks.

#### Partial solution:

- Using *Data/Draw data points*, double-click on points to define the background curve. The points are put into a GUI table called "Draw1" or similar.
- The table "Draw1" can be converted into a TableWorkspace or a MatrixWorkspace using the menu options under *Table*

#### To implement:
Write a Python algorithm that will take as input the original data workspace and the workspace with user-defined background points, and interpolate the user-defined points to create a background spectrum that can be subtracted. Use linear interpolation for the first attempt.

tom-perkins:
`CreateUserDefinedBackground` takes a set of points selected by the user in the GUI and returns a
workspace that can be subtracted from the data. The intended use case is described in the issue.

**To test:**
- Create or load a workspace that has a large, irregularly shaped background with several peaks on top of it.
- In MantidPlot, use _Data/Draw Data Points_ to define the background by double-clicking on a series of points. When you click outside of the graph window, a table will appear with your points in them.
  - (On Linux the table won't appear until you close the graph window - see #16874)
- With the table selected, convert to a workspace via _Table/Convert to TableWorkspace_
- Run the algorithm:

``` python
background = CreateUserDefinedBackground(InputWorkspace = "your input", BackgroundPoints = "your table")
```
- Ensure that the resulting workspace `background` can be subtracted from the original data workspace (using `Minus`) and the result contains just the peaks.

Fixes #16798.

[Release notes](https://github.com/mantidproject/mantid/blob/37607a4d6e6dd762cdec4b562170dcc628c2a5c1/docs/source/release/v3.8.0/framework.rst#new) have been updated.

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/Algorithms/src/CreateUserDefinedBackground.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+  const auto &xData = outputWS->readX(0);
+  MantidVec yBackground;
+  MantidVec eBackground(outputWS->blocksize(), 0);
+
+  // Interpolate Y data in the table
+  const auto &lerp = getInterpolator(background, data);
+  const bool isHisto = outputWS->isHistogramData();
+  for (size_t i = 0; i < data->blocksize(); i++) {
+    const double x = isHisto ? (xData[i] + xData[i + 1]) * 0.5 : xData[i];
```
AntonPiccardoSelg: Could this be done with the methods that the new `Histogram` type provides?

Framework/Algorithms/src/CreateUserDefinedBackground.cpp
```diff
@@ -0,0 +1,244 @@
...
+    double y = lerp.value(x);
+    if (isHisto) {
+      y *= (xData[i + 1] - xData[i]); // bin width
+    }
```
AntonPiccardoSelg: Similarly can the new `Frequency` and `Count` types be of help here?

Framework/Algorithms/src/CreateUserDefinedBackground.cpp
tom-perkins: Yes, that would be better. I will refactor to take advantage of the new `HistogramData` module.

Framework/Algorithms/src/CreateUserDefinedBackground.cpp
tom-perkins: Yes, I agree that would be clearer.

tom-perkins:
I've updated to use the new `HistogramData` module - This required merging in `master` so that the new module was available.

AntonPiccardoSelg:
@tom-perkins  Could you provide me with a workspace that has a large, irregularly shaped background with several peaks on top of it. Thanks

tom-perkins:
@AntonPiccardoSelg Spectrum 256 of `MAR11060.raw` (doc test data) looks a bit like the picture I have of the intended use case.

AntonPiccardoSelg:
@tom-perkins It seems that besides the spike there is a scaling issue. The data set is MARI11060 and teh spectrum is 256

original
original

background
background

tom-perkins:
The reason for the spike (and possibly the scaling) was in extending the background to the range of the data - it inserted an extra point using the original workspace's y data, but if it was multi-spectra then that would be spectrum 0 (a monitor in this case!) Changed this so the new point uses the last point of the background.

AntonPiccardoSelg:
@tom-perkins  The branch needs updating. Probably a release notes conflict.

AntonPiccardoSelg:
The spike is gone, but the scaling issue (and maybe shape issue) is still there

tom-perkins:
@AntonPiccardoSelg How do you get the scaling issue to appear? The results I get are as below:
results-linux

Top left: original data as loaded (histo)
Top right: background from histo data
Bottom left: data converted to points (`ConvertToPointData`)
Bottom right: background from the point data

I get the same results on Windows and Ubuntu (clean install of the PR build).

tom-perkins:
The difference was due to the "normalize histogram to bin width" option under *View/Preferences/2D Plots*, which plots the data as a distribution.

Added a fix for this - the algorithm should now work for histogram and point data, whether the option is turned on or off.

SimonHeybrock:
@tom-perkins @AntonPiccardoSelg It might be a good idea to wait with a merge until we have merged https://github.com/mantidproject/mantid/pull/16914. Basically this fixes a problem related to the `isDistribution` flag (we are moving it into `Histogram`). Maybe you could redo the related tests after a merge?
I am also happy to help if you run into trouble, I am right now working on cleaning up this distribution stuff a bit.

tom-perkins:
Now that #16914 has been merged into `master`, I have merged `master` into this branch. Hopefully the tests will still pass.

Framework/Algorithms/src/CreateUserDefinedBackground.cpp
```diff
@@ -0,0 +1,258 @@
...
+      data->isHistogramData() && (data->isDistribution() || distributionPlot);
+
+  // Apply Y and E data to all spectra in the workspace
+  for (size_t spec = 0; spec < outputWS->getNumberHistograms(); spec++) {
+    if (useFrequencies) {
+      outputWS->setFrequencies(spec, yBackground);
+      outputWS->setFrequencyStandardDeviations(spec, eBackground);
+    } else {
+      outputWS->setCounts(spec, yBackground);
+      outputWS->setCountStandardDeviations(spec, eBackground);
```
SimonHeybrock: @tom-perkins I think this block is making trouble in the unit tests now. For ensuring correctness, setting `Counts` when we have frequency data is now forbidden (and vice versa). The problem is basically the `useFrequencies` flag. I don't know what the best solution here is maybe changing the type of `outputWS` is the right way?

SimonHeybrock: @tom-perkins By changing the type I meant `outputWS->setDistribution(useFrequencies)`.

tom-perkins: @SimonHeybrock Thanks, I will take a look at this. At the moment the unit tests are dependent on what the config option `graph1d.autodistribution` is set to on the system where they are running - they should be independent of this, so I will modify the tests too.

tom-perkins:
Fixed and added extra tests to deal with config option being on or off.

Framework/Algorithms/src/CreateUserDefinedBackground.cpp
```diff
@@ -187,6 +180,61 @@ CreateUserDefinedBackground::createBackgroundWorkspace(
...
+    } else {
+      if ("On" ==
+          Kernel::ConfigService::Instance().getString(AUTODISTRIBUTIONKEY)) {
+        // User has selected the option "normalize histogram to bin width"
+        multiplyByBinWidth(data, yBackground);
+      }
+      return Histogram::YMode::Counts;
+    }
+  } else {
+    return Histogram::YMode::Counts;
```
SimonHeybrock: Is this correct? I thought point-data does often correspond to frequencies, not counts?

Framework/Algorithms/src/CreateUserDefinedBackground.cpp
```diff
@@ -187,6 +180,61 @@ CreateUserDefinedBackground::createBackgroundWorkspace(
...
+    std::vector<double> &yBackground) const {
+  if (!data->isHistogramData()) {
+    throw std::invalid_argument(
+        "Data is not histogram data: cannot multiply background by bin width");
+  }
+
+  const auto &X = data->binEdges(0);
+  for (size_t i = 0; i < yBackground.size(); ++i) {
+    const double width = X[i + 1] - X[i];
+    yBackground[i] *= width;
```
SimonHeybrock: Wouldn't you also need to modify `eBackground`?

tom-perkins: The errors of the background spectrum are set to zero so don't need to be multiplied. (The 'real' error would depend on how accurate the user was with their clicking on the data points)

Framework/Algorithms/src/CreateUserDefinedBackground.cpp
tom-perkins: In that case it may be best to keep the same Y mode as the input workspace has for this case.

SimonHeybrock:
@tom-perkins How about this?
```cpp
API::MatrixWorkspace_sptr
CreateUserDefinedBackground::createBackgroundWorkspace(
    const API::ITableWorkspace_const_sptr &background,
    const API::MatrixWorkspace_const_sptr &data) const {
  auto outputWS = data->clone();

  const auto &xPoints = outputWS->points(0);
  std::vector<double> yBackground;
  std::vector<double> eBackground(xPoints.size(), 0);

  // Interpolate Y data in the table to get y for each point
  const auto &lerp = getInterpolator(background, data);
  for (const auto &x : xPoints) {
    const double y = lerp.value(x);
    yBackground.push_back(y);
  }

  auto histogram = outputWS.histogram(0);
  if (histogram.yMode() == Histogram::YMode::Frequencies) {
    histogram.setFrequencies(spec, yBackground);
    histogram.setFrequencyStandardDeviations(spec, eBackground);
  } else {
    if ("On" == Kernel::ConfigService::Instance().getString(AUTODISTRIBUTIONKEY)) {
      // Background data is actually frequencies, we put it into temporary to benefit
      // from automatic conversion in setCounts(), etc.
      histogram.setCounts(spec, Frequencies(yBackground));
      histogram.setCountStandardDeviations(spec, FrequencyStandardDeviations(eBackground));
    } else {
      histogram.setCounts(spec, yBackground);
      histogram.setCountStandardDeviations(spec, eBackground);
    }
  }

  // Decide if Y data is counts or frequencies
  const Histogram::YMode yMode = getBackgroundYMode(data, yBackground);

  // Apply Y and E data to all spectra in the workspace
  for (size_t spec = 0; spec < outputWS->getNumberHistograms(); spec++) {
    // Setting same histogram for all spectra, data is shared, saving lots of memory
    outputWS->setHistogram(spec, histogram);
  }

  return API::MatrixWorkspace_sptr(std::move(outputWS));
}
```

tom-perkins:
The simplified version works well, the only thing is that for point data it shouldn't check `AUTODISTRIBUTIONKEY` and convert to counts. Other than that the results are the same and the new version is more readable (and efficient).

AntonPiccardoSelg:
The original is:
full

The background is:
background

The subtracted is:
full

Works as expected. This is a nice feature. Happy to :shipit:

Framework/Algorithms/inc/MantidAlgorithms/CreateUserDefinedBackground.h
```diff
@@ -0,0 +1,87 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+class MANTID_ALGORITHMS_DLL CreateUserDefinedBackground final
```
quantumsteve: Would you please remove `final` from the class definition and virtual function declaration? To avoid overuse of the `final`, we recently removed these from class_maker.py.

tom-perkins: Thanks, I have removed `final` now. I must have first created the class before class_maker.py was modified.

AndreiSavici:
In the documentation, it is not clear how do you create the background. Is it a liner fit? Or linear interpolation? Or spline?

tom-perkins:
It's linear interpolation (using `Kernel::Interpolation`) - I've added a note to the docs to explain this. If `Kernel::Interpolation` was ever extended to support other methods, a parameter could be added to this algorithm to choose the method, but it only supports linear at the moment.

quantumsteve:
@AndreiSavici Once you are satisfied with the documentation, would you please merge this?

Issue #16118, opened 2016-05-10
Umbrella: Quirks, questions, and ToDos during Histogram refactoring


SimonHeybrock:
This is a list of things I stumbled upon during introducing the `HistogramData` module. This includes need for cleanup, potential inefficiencies, ...

This is not necessarily anything that requires fixing, but may contain valuable information gained from scanning through a lot of code during the refectoring.
1. `ExtractSpectra.cpp`: If there are common `X` boundaries, sharing is set up for the output. However, `Dx` sharing is not. Is this a requirement, or an oversight? `commonBoundaries` does not check `Dx`, but maybe it is implied?
2. Cleanup: `Q1D2.cpp` lines 290ff can use `Histogram::operator=`.
3. `FitMW.cpp` lines 311 and 210 use of `MatrixWorkspace->isHistogramData()` slightly confusing.
4. `DiffractionFocussing2` line 200. Can only change this once `rebin` has been implemented in Histogram.
5. `FFT.cpp` line 361. `Kernel::binChecker` should be implemented in HistogramData as convenience function.
   See also #15880.
6. `AssymetricCalc.cpp` line 138 onwards may be useful when designing arithmetics for histogram types.
7. `CalculateTransmissionBeamSpreader.cpp` line 207 shows that we may need a RelativeStandardDeviation type
8. `ConvertAxesToRealSpace.cpp` line 261 dangerous use of E as variance.
9. `ConvertUnits.cpp` line 466 needs toTOF and fromTOF to be reimplemented outside of the kernel.
10. `ConvertUnits.cpp` line 714 `putBackBinWidth` emphasises the need to remove data storage abuse by other algorithms. This conversion should be internal to the histogram type only.
11. `CorrectToFile.cpp` line 98 more maths required in Histogram.
12. `DetectorEfficiencyCor.cpp` from line 261 another math example.
13. `AbsorptionCorrection.cpp` line 216 linearlyInterpolateY should be implemented in HistogramData. Until then we are forced to use the old interface.
14. `CrossCorrelate.cpp` line 180 onwards. More arithmetics for HistogramData
15. `FitPeak.cpp` line 1562 `getIndex()` might be worth adding to HistogramData. This also appears in `FindPeaks`.
16. `MuonGroupDetectors.cpp` line 106 would be a nice place to add Histogram arithmetic.
17. `ModeratorTzeroTest.h` maybe implement GaussianGenerator.

SimonHeybrock:
[...]

PR #16833, opened 2016-07-11
HistogramData Rollout for Framework/Algorithms A-F

### LINKED ISSUE #16612: HistogramData rollout Framework/Algorithm A-F
This issue relates to replacing the old interface for histogram data handling with the new interface developed by Simon Heybrock. See the [design document](https://github.com/mantidproject/documents/blob/master/Design/histogram_type.md) and first [pull request](https://github.com/mantidproject/mantid/pull/16036) for more information.

LamarMoore:
This PR concerns the refactoring of Framework/Algorithms with the new Histogram types as part of the plan to phase out the use of the legacy system (Sorry about the size). See [here](http://docs.mantidproject.org/nightly/concepts/HistogramData.html#rollout-status). In most cases, changes have resulted in increased efficiency which should hopefully be relflected in a performance improvement. There were a few cases where it was not yet possible to replace the legacy data access due to performance considerations (where more work needs to be done in the Histogram type with e.g arithmetics). These have been added to a global TODO list found [here](https://github.com/mantidproject/mantid/issues/16118).

**To test:**

Code Review

Fixes #16612 .

_Does not need to be in the release notes. The entire change to the new Histogram type will be reflected in the release notes._

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/Algorithms/src/DiffractionFocussing2.cpp
```diff
@@ -211,13 +212,16 @@ void DiffractionFocussing2::exec() {
...
-      const MantidVec &Xin = inSpec.readX();
-      const MantidVec &Yin = inSpec.readY();
-      const MantidVec &Ein = inSpec.readE();
+      auto &Xin = inSpec.x();
+      auto &Yin = inSpec.y();
+      auto &Ein = inSpec.e();
 
       outSpec.addDetectorIDs(inSpec.getDetectorIDs());
       try {
-        VectorHelper::rebinHistogram(Xin, Yin, Ein, Xout, Yout, Eout, true);
```
LamarMoore: DiffractionFocussing is partially updated because the use of `VectorHelper` methods restricts the use of the `HistogramData` types. This `VectorHelper::rebinHistogram` method should ideally be implemented in histogram.

Framework/Algorithms/src/ApplyTransmissionCorrection.cpp
```diff
@@ -61,26 +62,26 @@ void ApplyTransmissionCorrection::exec() {
...
   MatrixWorkspace_sptr outputWS = getProperty("OutputWorkspace");
 
-  MantidVec trans(inputWS->readY(0).size(), trans_value);
-  MantidVec dtrans(inputWS->readY(0).size(), trans_error);
-  MantidVec &TrIn = trans;
-  MantidVec &ETrIn = dtrans;
+  HistogramY trans(inputWS->y(0).size(), trans_value);
+  HistogramE dtrans(inputWS->y(0).size(), trans_error);
+  HistogramY &TrIn = trans;
+  HistogramE &ETrIn = dtrans;
```
SimonHeybrock: Why these references?

Framework/Algorithms/src/AsymmetryCalc.cpp
```diff
@@ -137,36 +137,33 @@ void AsymmetryCalc::exec() {
   Progress prog(this, 0.0, 1.0, blocksize);
   for (size_t j = 0; j < blocksize; ++j) {
     // cal F-aB
```
SimonHeybrock: Make a note to consult the following block of code when designing arithmetics for the histogram types.

Framework/Algorithms/src/CalculateTransmissionBeamSpreader.cpp
```diff
@@ -200,8 +200,8 @@ void CalculateTransmissionBeamSpreader::exec() {
 
       // Take the log of each datapoint for fitting. Preserve errors
       // percentage-wise.
-      MantidVec &Y = logTransmission->dataY(0);
-      MantidVec &E = logTransmission->dataE(0);
+      auto &Y = logTransmission->mutableY(0);
+      auto &E = logTransmission->mutableE(0);
       Progress progress(this, 0.4, 0.6, Y.size());
       for (size_t i = 0; i < Y.size(); ++i) {
         E[i] = std::abs(E[i] / Y[i]);
```
SimonHeybrock: Maybe we need a `RelativeStandardDeviation` type?

Framework/Algorithms/src/CalculateTransmissionBeamSpreader.cpp
```diff
@@ -289,9 +289,9 @@ CalculateTransmissionBeamSpreader::fitToData(API::MatrixWorkspace_sptr WS) {
...
     m = std::pow(10, m);
 
-    const MantidVec &X = result->readX(0);
-    MantidVec &Y = result->dataY(0);
-    MantidVec &E = result->dataE(0);
+    auto &X = result->x(0);
+    auto &Y = result->mutableY(0);
+    auto &E = result->mutableE(0);
     for (size_t i = 0; i < Y.size(); ++i) {
       Y[i] = b * (std::pow(m, 0.5 * (X[i] + X[i + 1])));
```
SimonHeybrock: This is computing bin centers, use `points()` instead of `x()`. Also add this to our list of potentially required math operations with histograms.

Framework/Algorithms/src/CalculateZscore.cpp
```diff
@@ -85,20 +85,20 @@ void CalculateZscore::exec() {
...
+    auto &histY = outWS->mutableY(i);
+    auto &histE = outWS->mutableE(i);
 
     vector<double> yzscores = getZscore(inpY);
     vector<double> ezscores = getZscore(inpE);
 
-    vecX = inpX;
-    vecY = yzscores;
-    vecE = ezscores;
+    histX = inpX;
```
SimonHeybrock: Share X!?

Framework/Algorithms/src/ChopData.cpp
```diff
@@ -136,12 +134,12 @@ void ChopData::exec() {
...
-        workspace->dataE(j)[k] = inputWS->readE(j)[oldbin];
-        workspace->dataX(j)[k] = inputWS->readX(j)[oldbin] - stepDiff;
+        workspace->mutableY(j)[k] = inputWS->y(j)[oldbin];
+        workspace->mutableE(j)[k] = inputWS->e(j)[oldbin];
+        workspace->mutableX(j)[k] = inputWS->x(j)[oldbin] - stepDiff;
       }
-      workspace->dataX(j)[nbins] =
-          inputWS->readX(j)[indexLow + nbins] - stepDiff;
+      workspace->mutableX(j)[nbins] =
+          inputWS->x(j)[indexLow + nbins] - stepDiff;
```
SimonHeybrock: Maybe this whole block can be written nicer with iterators? You can avoid the loops an the special treatment of the last bin edge.

Framework/Algorithms/src/ConvertAxesToRealSpace.cpp
```diff
@@ -258,7 +258,7 @@ void ConvertAxesToRealSpace::exec() {
   // loop over the data and sqrt the errors to complete the error calculation
   PARALLEL_FOR1(outputWs)
   for (int i = 0; i < nOutputHist; ++i) {
-    MantidVec &errorVec = outputWs->dataE(i);
+    auto &errorVec = outputWs->mutableE(i);
     std::transform(errorVec.begin(), errorVec.end(), errorVec.begin(),
                    static_cast<double (*)(double)>(sqrt));
```
SimonHeybrock: Another case of the dangerous use of E as variance (and subsequent conversion), put this on the list together with `rebinHistogram`.

Framework/Algorithms/src/ConvertEmptyToTof.cpp
```diff
@@ -509,20 +508,16 @@ void ConvertEmptyToTof::setTofInWS(const std::vector<double> &tofAxis,
                 << numberOfSpectra << '\n';
 
   auto axisPtr = Kernel::make_cow<HistogramData::HistogramX>(tofAxis);
+  HistogramData::BinEdges edges(tofAxis);
   Progress prog(this, 0.0, 0.2, numberOfSpectra);
-  PARALLEL_FOR2(m_inputWS, outputWS)
```
SimonHeybrock: Good :+1:, threading is probably pointless since we just manipulate a few pointers. Personally I would also remove the progress reporting from inside the loop, all it will do is slow things down.

Framework/Algorithms/src/ConvertEmptyToTof.cpp
```diff
@@ -509,20 +508,16 @@ void ConvertEmptyToTof::setTofInWS(const std::vector<double> &tofAxis,
                 << numberOfSpectra << '\n';
 
   auto axisPtr = Kernel::make_cow<HistogramData::HistogramX>(tofAxis);
+  HistogramData::BinEdges edges(tofAxis);
```
SimonHeybrock: You can now also directly construct `BinEdges` from `tofAxis`.

Framework/Algorithms/src/ConvertTableToMatrixWorkspace.cpp
```diff
@@ -58,12 +59,12 @@ void ConvertTableToMatrixWorkspace::exec() {
 
   MatrixWorkspace_sptr outputWorkspace =
       WorkspaceFactory::Instance().create("Workspace2D", 1, nrows, nrows);
-  outputWorkspace->dataX(0).assign(X.begin(), X.end());
-  outputWorkspace->dataY(0).assign(Y.begin(), Y.end());
+  outputWorkspace->mutableX(0) = HistogramX(X.begin(), X.end());
+  outputWorkspace->mutableY(0) = HistogramY(Y.begin(), Y.end());
```
SimonHeybrock: `outputWorkspace->mutableX(0) = std::move(X);` (etc.)?

Framework/Algorithms/src/ConvertUnits.cpp
```diff
@@ -266,13 +267,12 @@ void ConvertUnits::convertQuickly(API::MatrixWorkspace_sptr outputWS,
     // Only do the full check if the quick one passes
     if (commonBoundaries) {
       // Calculate the new (common) X values
-      MantidVec::iterator iter;
-      for (iter = outputWS->dataX(0).begin(); iter != outputWS->dataX(0).end();
-           ++iter) {
+      for (auto iter = outputWS->mutableX(0).begin();
+           iter != outputWS->x(0).end(); ++iter) {
```
SimonHeybrock: Careful! You are mixing iterators for `mutableX` and `x`!

Why not `for(auto &x : outputWS->mutableX(0))`?

Framework/Algorithms/src/ConvertUnits.cpp
```diff
@@ -297,24 +297,14 @@ void ConvertUnits::convertQuickly(API::MatrixWorkspace_sptr outputWS,
   for (int64_t k = 0; k < numberOfSpectra_i; ++k) {
     PARALLEL_START_INTERUPT_REGION
     if (!commonBoundaries) {
-      MantidVec::iterator it;
-      for (it = outputWS->dataX(k).begin(); it != outputWS->dataX(k).end();
+      for (auto it = outputWS->mutableX(k).begin(); it != outputWS->x(k).end();
```
SimonHeybrock: As above, use range-based `for`.

Framework/Algorithms/src/ConvertUnits.cpp
```diff
@@ -475,12 +465,16 @@ void ConvertUnits::convertViaTOF(Kernel::Unit_const_sptr fromUnit,
...
-      localFromUnit->toTOF(outputWS->dataX(i), emptyVec, l1, l2, twoTheta,
-                           emode, efixed, delta);
+      localFromUnit->toTOF(vals, emptyVec, l1, l2, twoTheta, emode, efixed,
+                           delta);
+
       // Convert from time-of-flight to the desired unit
-      localOutputUnit->fromTOF(outputWS->dataX(i), emptyVec, l1, l2, twoTheta,
-                               emode, efixed, delta);
+      localOutputUnit->fromTOF(vals, emptyVec, l1, l2, twoTheta, emode, efixed,
+                               delta);
```
SimonHeybrock: I think you should just refactor `toTOF` and `fromTOF` to accept `HistogramX` instead of `std::vector<double>`.

Framework/Algorithms/src/ConvertUnits.cpp
```diff
@@ -688,25 +682,19 @@ API::MatrixWorkspace_sptr ConvertUnits::removeUnphysicalBins(
...
-      MantidVec &Ynew = result->dataY(j);
-      MantidVec &Enew = result->dataE(j);
       int k;
       for (k = 0; k < lastBins[j] - 1; ++k) {
-        Xnew[k] = X[k];
-        Ynew[k] = Y[k];
-        Enew[k] = E[k];
+        result->mutableX(j)[k] = workspace->x(j)[k];
+        result->mutableY(j)[k] = workspace->y(j)[k];
+        result->mutableE(j)[k] = workspace->e(j)[k];
```
SimonHeybrock: Calling `mutable?()` in a tight loop should be avoided (expensive `cow_ptr` access).

Framework/Algorithms/src/ConvertUnitsUsingDetectorTable.cpp
```diff
@@ -422,9 +429,8 @@ void ConvertUnitsUsingDetectorTable::convertQuickly(
     // Only do the full check if the quick one passes
     if (commonBoundaries) {
       // Calculate the new (common) X values
-      MantidVec::iterator iter;
-      for (iter = outputWS->dataX(0).begin(); iter != outputWS->dataX(0).end();
-           ++iter) {
+      for (auto iter = outputWS->mutableX(0).begin();
+           iter != outputWS->x(0).end(); ++iter) {
```
SimonHeybrock: Many of my comments on `ConvertUnits` also apply to this file, I will not repeat them here.

Framework/Algorithms/src/CorrectFlightPaths.cpp
```diff
@@ -125,14 +115,12 @@ void CorrectFlightPaths::exec() {
 
     unsigned int j = 0;
     for (; j < numberOfChannels; ++j) {
-      xOut[j] = xIn[j] - deltaTOF;
-      // there's probably a better way of copying this....
-      yOut[j] = yIn[j];
-      eOut[j] = eIn[j];
+      m_outputWS->mutableX(i)[j] -= deltaTOF;
     }
```
SimonHeybrock: `m_outputWS->mutableX(i) -= deltaTOF;`?

Framework/Algorithms/src/CorrectKiKf.cpp
```diff
@@ -130,11 +130,11 @@ void CorrectKiKf::exec() {
...
-    MantidVec &eOut = outputWS->dataE(i);
+    auto &yOut = outputWS->mutableY(i);
+    auto &eOut = outputWS->mutableE(i);
     const auto &xIn = inputWS->points(i);
-    const MantidVec &yIn = inputWS->readY(i);
-    const MantidVec &eIn = inputWS->readE(i);
+    auto &yIn = inputWS->y(i);
+    auto &eIn = inputWS->e(i);
     // Copy the energy transfer axis
     outputWS->setX(i, inputWS->refX(i));
```
SimonHeybrock: New interface.

Framework/Algorithms/src/CorrectToFile.cpp
```diff
@@ -87,13 +87,13 @@ void CorrectToFile::exec() {
...
+      auto &yOut = outputWS->mutableY(i);
+      auto &eOut = outputWS->mutableE(i);
 
-      const MantidVec &yIn = toCorrect->readY(i);
-      const MantidVec &eIn = toCorrect->readE(i);
+      auto &yIn = toCorrect->y(i);
+      auto &eIn = toCorrect->e(i);
 
       for (size_t j = 0; j < nbins; ++j) {
         const double currentX = xIn[j];
```
SimonHeybrock: More examples for potential required math operations follow.

Framework/Algorithms/src/CreatePSDBleedMask.cpp
```diff
@@ -227,10 +227,10 @@ bool CreatePSDBleedMask::performBleedTest(
...
-    const MantidVec &botY = inputWS->readY(botIndex);
-    const MantidVec &topX = inputWS->readX(topIndex);
-    const MantidVec &botX = inputWS->readX(botIndex);
+    auto &topY = inputWS->y(topIndex);
+    auto &botY = inputWS->y(botIndex);
+    auto &topX = inputWS->x(topIndex);
+    auto &botX = inputWS->x(botIndex);
     for (int j = 0; j < numBins; ++j) {
       double topRate(topY[j]), botRate(botY[j]);
       if (isRawCounts) {
```
SimonHeybrock: This is doing a conversion from counts to frequencies when necessary. I think you could maybe use `frequencies()` here.

Framework/Algorithms/src/CrossCorrelate.cpp
```diff
@@ -175,17 +175,18 @@ void CrossCorrelate::exec() {
     size_t wsIndex = indexes[i]; // Get the ws index from the table
     // Copy spectra info from input Workspace
     out->getSpectrum(i).copyInfoFrom(inputWS->getSpectrum(wsIndex));
-    out->dataX(i) = XX;
+    out->mutableX(i) = XX;
```
SimonHeybrock: You could share X, I think.

Framework/Algorithms/src/DetectorEfficiencyCor.cpp
```diff
@@ -124,12 +124,12 @@ void DetectorEfficiencyCor::exec() {
...
-    m_outputWS->setX(i, m_inputWS->refX(i));
+    m_outputWS->setSharedX(i, m_inputWS->sharedX(i));
     try {
       correctForEfficiency(i);
     } catch (Exception::NotFoundError &) {
       // zero the Y data that can't be corrected
-      MantidVec &outY = m_outputWS->dataY(i);
+      auto &outY = m_outputWS->mutableY(i);
       std::transform(outY.begin(), outY.end(), outY.begin(),
                      std::bind2nd(std::multiplies<double>(), 0));
```
SimonHeybrock: `m_outputWS->mutableY(i) *= 0;`

Framework/Algorithms/src/DetectorEfficiencyCor.cpp
```diff
@@ -258,7 +258,7 @@ void DetectorEfficiencyCor::correctForEfficiency(int64_t spectraIn) {
     auto einItr = eValues.cbegin();
     auto youtItr = yout.begin();
     auto eoutItr = eout.begin();
-    auto xItr = m_inputWS->readX(spectraIn).cbegin();
+    auto xItr = m_inputWS->x(spectraIn).cbegin();
```
SimonHeybrock: Another math example, please take a note.

Framework/Algorithms/test/CalculateTransmissionTest.h
```diff
@@ -352,8 +352,8 @@ class CalculateTransmissionTest : public CxxTest::TestSuite {
                 outputWS + "_unfitted")));
 
     {
-      const Mantid::MantidVec &fitted_y = fitted->readY(0),
-                              &fitted_x = fitted->readX(0);
+      auto &fitted_y = fitted->y(0);
+      auto &fitted_x = fitted->x(0);
```
SimonHeybrock: :+1: for getting rid of these multi-line comma-separated variable declarations.

Framework/Algorithms/test/ConvertToMatrixWorkspaceTest.h
```diff
@@ -5,13 +5,15 @@
...
 #include "MantidAlgorithms/ConvertToMatrixWorkspace.h"
+#include "MantidTestHelpers/HistogramDataTestHelper.h"
 #include "MantidAPI/MatrixWorkspace.h"
 #include "MantidDataObjects/EventWorkspace.h"
 #include "MantidGeometry/Instrument.h"
 #include "MantidTestHelpers/WorkspaceCreationHelper.h"
 
 using namespace Mantid;
 using namespace Mantid::Kernel;
+using namespace Mantid::HistogramData::detail;
```
SimonHeybrock: Why?

Framework/Algorithms/src/AbsorptionCorrection.cpp
```diff
@@ -88,6 +88,41 @@ void AbsorptionCorrection::init() {
...
+      continue;
+    }
+    xp = (isHistogram ? (0.5 * (x[i] + x[i + 1])) : x[i]);
+    // Linear interpolation
+    y[i] = (xp - x1) * y2 + (x2 - xp) * y1;
+    y[i] *= overgap;
+    step++;
+  }
+}
+
```
SimonHeybrock: We had discussed this before (on the previous PR), you should not duplicate that code.

Framework/Algorithms/src/CalculateTransmissionBeamSpreader.cpp
```diff
@@ -289,11 +289,11 @@ CalculateTransmissionBeamSpreader::fitToData(API::MatrixWorkspace_sptr WS) {
     b = std::pow(10, b);
     m = std::pow(10, m);
 
-    auto &X = result->x(0);
+    auto &X = result->points(0);
     auto &Y = result->mutableY(0);
     auto &E = result->mutableE(0);
     for (size_t i = 0; i < Y.size(); ++i) {
-      Y[i] = b * (std::pow(m, 0.5 * (X[i] + X[i + 1])));
+      Y[i] = b * (std::pow(m, 0.5 * X[i]));
```
SimonHeybrock: The `0.5` is part of the bin center computation.

Framework/Algorithms/src/CalculateTransmissionBeamSpreader.cpp
```diff
@@ -289,11 +289,11 @@ CalculateTransmissionBeamSpreader::fitToData(API::MatrixWorkspace_sptr WS) {
     b = std::pow(10, b);
     m = std::pow(10, m);
 
-    auto &X = result->x(0);
+    auto &X = result->points(0);
```
SimonHeybrock: I would remove the `&` to decrease the risk of confusion, since `points()` returns by value.

Framework/Algorithms/src/ChopData.cpp
```diff
@@ -131,16 +131,22 @@ void ChopData::exec() {
...
+
+      PARALLEL_START_INTERUPT_REGION;
+      std::transform(edges.cbegin() + indexLow,
+                     edges.cbegin() + indexLow + nbins + 1,
+                     workspace->mutableX(j).begin(),
+                     [stepDiff](const double &a) { return a - stepDiff; });
+
+      std::copy(inputWS->y(j).cbegin() + indexLow,
+                inputWS->y(j).cbegin() + indexLow + nbins,
+                workspace->mutableY(j).begin());
```
SimonHeybrock: Maybe we should give `HistogramY` (etc.) an `assign` method, like `std::vector`? What do you think?

Framework/Algorithms/src/ConvertUnits.cpp
```diff
@@ -465,21 +464,18 @@ void ConvertUnits::convertViaTOF(Kernel::Unit_const_sptr fromUnit,
 
       /// @todo Don't yet consider hold-off (delta)
       const double delta = 0.0;
-      std::vector<double> vals(outputWS->x(i).begin(), outputWS->x(i).end());
-      // Convert the input unit to time-of-flight
-      localFromUnit->toTOF(vals, emptyVec, l1, l2, twoTheta, emode, efixed,
-                           delta);
 
+      // TODO toTOF and fromTOF need to be reimplemented outside of kernel
```
SimonHeybrock: Are you keeping track of these `TODOs` somewhere? Otherwise I would recommend to choose a more unique label, something like `TODOHISTOGRAM`.

Framework/Algorithms/src/ConvertUnits.cpp
```diff
@@ -682,21 +678,30 @@ API::MatrixWorkspace_sptr ConvertUnits::removeUnphysicalBins(
...
+      auto &resX = result->mutableX(j);
+
+      auto k = lastBins[j];
+
+      std::copy(edges.cbegin(), edges.cbegin() + lastBins[j],
+                result->mutableX(j).begin());
+
+	  // If necessary, add on some fake values to the end of the X array (Y&E
+	  // will be zero)
       if (k < maxBins) {
```
SimonHeybrock: I think you can drop the `if`. It is implicit by the range that you give to `std::transform`, I think?

Framework/Algorithms/src/ConvertUnitsUsingDetectorTable.cpp
```diff
@@ -618,9 +617,10 @@ API::MatrixWorkspace_sptr ConvertUnitsUsingDetectorTable::removeUnphysicalBins(
       auto &X = workspace->x(i);
       auto &Y = workspace->y(i);
       auto &E = workspace->e(i);
-      result->mutableX(i) = HistogramX(X.begin() + first, X.end());
-      result->mutableY(i) = HistogramY(Y.begin() + first, Y.end());
-      result->mutableE(i) = HistogramE(E.begin() + first, E.end());
+
+      result->mutableX(i) = std::move(HistogramX(X.begin() + first, X.end()));
+      result->mutableY(i) = std::move(HistogramY(Y.begin() + first, Y.end()));
+      result->mutableE(i) = std::move(HistogramE(E.begin() + first, E.end()));
```
SimonHeybrock: The `std::move` is redundant, you are assigning an rvalue anyway. Also, I think we should really add `FixedLengthVector::assign`.

Framework/Algorithms/src/CorrectFlightPaths.cpp
```diff
@@ -113,13 +113,11 @@ void CorrectFlightPaths::exec() {
...
-      m_outputWS->mutableX(i)[j] -= deltaTOF;
-    }
-    // last bin
-    m_outputWS->mutableX(i)[numberOfChannels] =
-        m_inputWS->x(i)[numberOfChannels] + deltaTOF;
+    auto &edges = m_outputWS->binEdges(i);
+    auto &X = m_outputWS->mutableX(i);
+
+    std::transform(edges.cbegin(), edges.cbegin() + numberOfChannels + 1,
+                   X.begin(), [=](const double &x) { return x - deltaTOF; });
```
SimonHeybrock: `m_outputWS->mutableX(i) -= deltaTOF`.

Framework/Algorithms/src/CreatePSDBleedMask.cpp
```diff
@@ -217,26 +217,20 @@ bool CreatePSDBleedMask::performBleedTest(
...
   size_t top = 0, bot = bottomBegin;
   for (; top < topEnd; ++top, ++bot) {
     const int topIndex = tubeIndices[top];
     const int botIndex = tubeIndices[bot];
-    auto &topY = inputWS->y(topIndex);
-    auto &botY = inputWS->y(botIndex);
+    // Provide frequencies as opposed to calculating frequencies on the fly for
+    // count daya
+    auto &topY = inputWS->frequencies(topIndex);
+    auto &botY = inputWS->frequencies(botIndex);
```
SimonHeybrock: No, as discussed yesterday you cannot do these changes as long as `ConvertToDistribution` and `WorkspaceHelpers::makeDistribution` exist!

Framework/Algorithms/src/CrossCorrelate.cpp
```diff
@@ -175,7 +176,10 @@ void CrossCorrelate::exec() {
     size_t wsIndex = indexes[i]; // Get the ws index from the table
     // Copy spectra info from input Workspace
     out->getSpectrum(i).copyInfoFrom(inputWS->getSpectrum(wsIndex));
-    out->mutableX(i) = XX;
+
+    if (i > 0) {
```
SimonHeybrock: You do not need this `if`. The objects in `HistogramData` check for self-assignment and handle it correctly.

Framework/Algorithms/src/CrossCorrelate.cpp
```diff
@@ -175,7 +176,10 @@ void CrossCorrelate::exec() {
...
+      out->setSharedX(i, out->sharedX(0));
+    }
```
SimonHeybrock: Note: below is another good example for histogram arithmetics, please add to your list.

Framework/Algorithms/src/FFT.cpp
```diff
@@ -110,18 +109,201 @@ void FFT::exec() {
...
+  gsl_fft_complex_workspace_free(workspace);
+}
+
+void FFT::transformBackward(boost::shared_array<double> &data,
+                            const bool addPositiveOnly, bool isComplex,
+                            const int iReal, const int iImag, const double df,
+                            const double dx) {
+
+  const int ySize = static_cast<int>(m_inWS->blocksize());
+  const int xSize = static_cast<int>(m_inWS->x(iReal).size());
```
SimonHeybrock: There are things that a duplicated in the forward and backward functions. I would move it outside the scope of these methods to avoid that. Maybe put it in the `exec()`, before and after the call to the forward/backward methods.

Framework/Algorithms/src/ConvertUnits.cpp
LamarMoore: I am making a note of these things [here](https://github.com/mantidproject/mantid/issues/16118). The these are just for other developers who may encounter this code before we get the chance to change it.

Framework/Algorithms/src/ConvertUnits.cpp
```diff
@@ -683,15 +683,6 @@ API::MatrixWorkspace_sptr ConvertUnits::removeUnphysicalBins(
...
 
-      // If necessary, add on some fake values to the end of the X array (Y&E
-      // will be zero)
-      if (k < maxBins) {
-        auto l = k;
-        std::transform(resX.begin() + lastBins[j], resX.end(),
-                       resX.begin() + lastBins[j],
-                       [=](double &x) mutable { return x + 1 + (++l) - k; });
-      }
-
```
SimonHeybrock: You need the `transform`, what I meant is that you do not need to wrap it in the `if`!

Framework/HistogramData/inc/MantidHistogramData/VectorOf.h
```diff
@@ -116,7 +116,14 @@ template <class T, class CowType> class VectorOf {
...
   }
-
+  /// Assigns the contents of the range [first, last]
+  template <class InputIt> void assign(InputIt first, InputIt last) & {
+    m_data.access().assign(first, last);
+  }
+  /// Assigns data of size count with all elements initialized to value.
+  void assign(size_t count, const double &value) & {
+    m_data.access().assign(count, value);
+  }
```
SimonHeybrock: Good.
Could you also add tests for these and those in `FixedLengthVector`? `HistogramData` is a low level library that is used everywhere, so I am trying to be really thorough and test everything.

LamarMoore: Will do

SimonHeybrock:
@LamarMoore I will keep quickly reviewing your incremental changes. Once you think this is basically stable, I will review the whole thing again. Can you let me know when you think it is mostly ready? Maybe we can then get this merged next week.

Framework/HistogramData/test/FixedLengthVectorTest.h
```diff
@@ -107,6 +107,30 @@ class FixedLengthVectorTest : public CxxTest::TestSuite {
...
+    std::vector<double> src{3.6, 9.7, 8.5};
+    FixedLengthVectorTester dest;
+
+    dest.assign(src.cbegin(), src.cend());
+
+    TS_ASSERT_EQUALS(dest.size(), 3);
+    TS_ASSERT_EQUALS(dest[0], src[0]);
+    TS_ASSERT_EQUALS(dest[1], src[1]);
+    TS_ASSERT_EQUALS(dest[2], src[2]);
+  }
```
SimonHeybrock: I overlooked this earlier: Of course you need to verify the assignment size in the case of `FixedLengthVector` (as the name implies). See other cases like `operator=`.

Framework/HistogramData/test/VectorOfTest.h
```diff
@@ -108,6 +108,29 @@ class VectorOfTest : public CxxTest::TestSuite {
...
+    VectorOfTester dest(1);
+
+    dest.assign(3, 9.8);
+
+    TS_ASSERT_EQUALS(dest.size(), 3);
+    TS_ASSERT_EQUALS(dest[0], 9.8);
+    TS_ASSERT_EQUALS(dest[1], 9.8);
+    TS_ASSERT_EQUALS(dest[2], 9.8);
+  }
+
```
SimonHeybrock: These are good, changing length is allowed for `VectorOf`. 👍

Framework/HistogramData/inc/MantidHistogramData/FixedLengthVector.h
```diff
@@ -107,11 +108,16 @@ template <class T> class FixedLengthVector {
...
 private:
-  template <class Other> void checkAssignmentSize(const Other &other) {
+  template <class Other> void checkAssignmentSize(const Other &other) const {
     if (size() != other.size())
       throw std::logic_error("FixedLengthVector::operator=: size mismatch");
   }
 
+  void checkAssignmentSize(const size_t &size) const {
+    if (this->size() != size)
+      throw std::logic_error("FixedLengthVector::operator=: size mismatch");
```
SimonHeybrock: Not `operator=` in this case.

Framework/Algorithms/test/DetectorEfficiencyCorTest.h
```diff
@@ -123,8 +123,8 @@ class DetectorEfficiencyCorTest : public CxxTest::TestSuite {
     std::vector<double> e{sqrt(5.0), sqrt(5.0), sqrt(5.0), 0.0};
 
     space2D->setBinEdges(0, x);
-    space2D->mutableY(0) = y;
-    space2D->mutableE(0) = e;
+    space2D->mutableY(0).assign(y.begin(), y.end());
+    space2D->mutableE(0).assign(e.begin(), e.end());
```
SimonHeybrock: Why assign if it is the full range?

You could also just do everything directly, `space2D->setHistogram(0, BinEdges{0.0, ...}, Counts{10, ...}, CountVariances{5.0, ...})`.

Framework/Algorithms/test/DetectorEfficiencyCorTest.h
```diff
@@ -117,14 +119,9 @@ class DetectorEfficiencyCorTest : public CxxTest::TestSuite {
...
-    // Fill a couple of zeros just as a check that it doesn't get changed
-    std::vector<double> y{10, 11, 12, 0};
-    std::vector<double> e{sqrt(5.0), sqrt(5.0), sqrt(5.0), 0.0};
-
-    space2D->setBinEdges(0, x);
-    space2D->mutableY(0).assign(y.begin(), y.end());
-    space2D->mutableE(0).assign(e.begin(), e.end());
+    space2D->setHistogram(0, BinEdges{0.0, 0.0, 0.0, 0.0, 4.0},
+                          Counts{10, 11, 12, 0},
+                          CountVariances{sqrt(5.0), sqrt(5.0), sqrt(5.0), 0.0});
```
SimonHeybrock: If you use `CountVariances` you have to remove the `sqrt`. `E` is standard deviations be default.

Framework/Algorithms/src/AbsorptionCorrection.cpp
```diff
@@ -144,8 +144,8 @@ void AbsorptionCorrection::exec() {
     PARALLEL_START_INTERUPT_REGION
 
     // Copy over bin boundaries
-    const MantidVec &X = m_inputWS->readX(i);
-    correctionFactors->dataX(i) = X;
+    auto &X = m_inputWS->x(i);
```
SimonHeybrock: Move this closer to where it is used.

Framework/Algorithms/src/ConvertUnits.cpp
```diff
@@ -686,27 +675,24 @@ API::MatrixWorkspace_sptr ConvertUnits::removeUnphysicalBins(
...
+
+      // If the entire X range is not covered, generate fake values.
+      auto l = lastBins[j];
+
+      std::generate(result->mutableX(j).begin() + l, result->mutableX(j).end(),
+                    [=]() mutable {
+                      ++l;
+                      return result->mutableX(j)[lastBins[j]] + 1 +
+                             static_cast<double>(l - lastBins[j]);
+                    });
```
SimonHeybrock: I think something here is messed up with the indices. You are always using `result->mutableX(j)[lastBins[j]]` as base value, but it used to be `X[k]`.

Framework/Algorithms/src/ConvertUnitsUsingDetectorTable.cpp
```diff
@@ -354,12 +355,18 @@ void ConvertUnitsUsingDetectorTable::convertViaTOF(
...
-                             emode, efixed, delta);
+        localFromUnit->toTOF(values, emptyVec, l1, l2, twoTheta, emode, efixed,
+                             delta);
         // Convert from time-of-flight to the desired unit
-        localOutputUnit->fromTOF(outputWS->dataX(wsid), emptyVec, l1, l2,
-                                 twoTheta, emode, efixed, delta);
+        localOutputUnit->fromTOF(values, emptyVec, l1, l2, twoTheta, emode,
+                                 efixed, delta);
+
+        outputWS->mutableX(wsid) = HistogramX(values);
```
SimonHeybrock: `std::move`?

Framework/Algorithms/src/ConvertUnitsUsingDetectorTable.cpp
```diff
@@ -521,15 +524,15 @@ const std::vector<double> ConvertUnitsUsingDetectorTable::calculateRebinParams(
...
-    std::reverse(WS->dataY(0).begin(), WS->dataY(0).end());
-    std::reverse(WS->dataE(0).begin(), WS->dataE(0).end());
+    std::reverse(WS->mutableX(0).begin(), WS->mutableX(0).end());
+    std::reverse(WS->mutableY(0).begin(), WS->mutableY(0).end());
+    std::reverse(WS->mutableE(0).begin(), WS->mutableE(0).end());
 
-    auto xVals = WS->refX(0);
+    auto xVals = WS->sharedX(0);
     for (size_t j = 1; j < m_numberOfSpectra; ++j) {
       WS->setX(j, xVals);
```
SimonHeybrock: `setSharedX`

Framework/Algorithms/src/ConvertUnitsUsingDetectorTable.cpp
```diff
@@ -638,27 +642,24 @@ API::MatrixWorkspace_sptr ConvertUnitsUsingDetectorTable::removeUnphysicalBins(
...
+
+      // If the entire X range is not covered, generate fake values.
+      auto l = lastBins[j];
+
+      std::generate(result->mutableX(j).begin() + l, result->mutableX(j).end(),
+                    [=]() mutable {
+                      ++l;
+                      return result->mutableX(j)[lastBins[j]] + 1 +
+                             static_cast<double>(l - lastBins[j]);
+                    });
```
SimonHeybrock: Same index issue here (see `ConvertUnits`)?

Framework/Algorithms/src/DetectorEfficiencyCorUser.cpp
```diff
@@ -113,8 +113,10 @@ void DetectorEfficiencyCorUser::exec() {
  * @param eOut corrected spectrum errors
  */
 void DetectorEfficiencyCorUser::applyDetEfficiency(
-    const size_t numberOfChannels, const MantidVec &yIn, const MantidVec &eIn,
-    const MantidVec &effVec, MantidVec &yOut, MantidVec &eOut) {
+    const size_t numberOfChannels, const Mantid::HistogramData::HistogramY &yIn,
+    const Mantid::HistogramData::HistogramE &eIn, const MantidVec &effVec,
+    Mantid::HistogramData::HistogramY &yOut,
+    Mantid::HistogramData::HistogramE &eOut) {
```
SimonHeybrock: Personally, I would refactor this function to have a signature like this: `Histogram applyXYZ(const Histogram &histogram)`. THat would also make the call site look cleaner, I think.

Framework/Algorithms/src/FFT.cpp
```diff
@@ -110,18 +108,183 @@ void FFT::exec() {
...
-  MatrixWorkspace_sptr outWS =
-      WorkspaceFactory::Instance().create(inWS, nOut, xSize, ySize);
+  m_outWS = WorkspaceFactory::Instance().create(m_inWS, nOut, xSize, ySize);
 
   const double dx = X[1] - X[0];
   double df = 1.0 / (dx * ySize);
 
   // Output label
-  outWS->getAxis(0)->unit() = UnitFactory::Instance().create("Label");
+  outputLabel(df);
```
SimonHeybrock: It is not clear what this does, can you find a better name?

Framework/Algorithms/test/CalculateTransmissionTest.h
```diff
@@ -461,8 +461,9 @@ class CalculateTransmissionTest : public CxxTest::TestSuite {
...
         source = boost::dynamic_pointer_cast<MatrixWorkspace>(
             Mantid::API::AnalysisDataService::Instance().retrieve(m_transWS));
-    Mantid::MantidVec &Xfiddle0 = dir->dataX(0), &Xsource = source->dataX(0);
-    Mantid::MantidVec &Xfiddle1 = dir->dataX(1);
+    auto &Xfiddle0 = dir->mutableX(0);
+    auto &Xsource = source->x(0);
+    auto &Xfiddle1 = dir->mutableX(1);
     for (unsigned int i = 0; i < Xfiddle0.size(); ++i) {
       Xfiddle0[i] = Xsource[i];
       Xfiddle1[i] = Xsource[i];
```
SimonHeybrock: I guess you can just set up sharing of X?

Framework/HistogramData/inc/MantidHistogramData/VectorOf.h
```diff
@@ -116,7 +116,14 @@ template <class T, class CowType> class VectorOf {
       m_data = Kernel::make_cow<CowType>(std::move(data));
     return *this;
   }
-
+  /// Assigns the contents of the range [first, last]
```
SimonHeybrock: `last` is not included, so this should read `[first, last)`, see also http://en.cppreference.com/w/cpp/container/vector/assign.

Framework/Algorithms/src/DetectorEfficiencyCorUser.cpp
```diff
@@ -105,24 +103,22 @@ void DetectorEfficiencyCorUser::exec() {
...
-    // xOut[j] = xIn[j];
-    yOut[j] = yIn[j] / effVec[j];
-    eOut[j] = eIn[j] / effVec[j];
+Histogram DetectorEfficiencyCorUser::applyDetEfficiency(
+    const size_t nChans, const MantidVec &effVec, const Histogram &histogram) {
+  Histogram outHist(histogram);
+
+  for (unsigned int j = 0; j < nChans; ++j) {
+    outHist.mutableY()[j] = histogram.y()[j] / effVec[j];
+    outHist.mutableE()[j] = histogram.e()[j] / effVec[j];
```
SimonHeybrock: `mutable?()` in inner loop, 👎  (please change ;) )

Apart from that, considering performance (maybe not a major issue here, but in the long run it pays off the pay attention to these things), consider what happens here:

1. `mutableY()` triggers a copy.
2. You set the values in the copy to the value in the original divided by some value. The resulting data movement is thus: load original, load copy, load scale, store copy, i.e., 4 * 8 Byte for each value in the Y output. If you instead scale the copy directly you avoid loading the original, i.e., the CPU has to move only 3 * 8 Byte.

Framework/Algorithms/src/ConvertUnits.cpp
```diff
@@ -686,27 +675,22 @@ API::MatrixWorkspace_sptr ConvertUnits::removeUnphysicalBins(
...
-      }
+      auto edges = workspace->binEdges(j);
+      auto &X = workspace->x(j);
+      auto k = lastBins[j];
+
+      result->mutableX(j).assign(edges.cbegin(), edges.cbegin() + lastBins[j]);
+
+      // If the entire X range is not covered, generate fake values.
+      std::generate(result->mutableX(j).begin() + k, result->mutableX(j).end(),
+                    [ l = k, k, j, X ]() mutable {
```
SimonHeybrock: You don't need `j`, do you?

Framework/Algorithms/src/ConvertUnits.cpp
```diff
@@ -686,27 +675,22 @@ API::MatrixWorkspace_sptr ConvertUnits::removeUnphysicalBins(
...
+                      return X[k] + 1 + static_cast<double>(l++ - k);
+                    });
```
SimonHeybrock: Looking at this again, can't we write slightly simpler? Consider this `std::generate` in combination with the assign above:

```cpp
result->mutableX(j).assign(edges.cbegin(), edges.cbegin() + lastBins[j]);
```
If we just extend that until the end of the result X, you do not need to have `X[k]` in the generate.

OwenArnold:
@SimonHeybrock has already given me a :shipit: thumbs up and has reviewed it. All looks give, I think that some of these examples highlight the need for `Histogram` level arithmetic. System tests look to be running in about the same amount of time as master clean, although not easy to compare as not run on the same machine, even though same OS.

PR #16836, opened 2016-07-11
16810 spectrum detector indexing

SimonHeybrock:
Do not merge, for review and discussion only.

@OwenArnold These are the essential indexing bits. We would have basically the same also for `DetectorId`.
- `SpectrumNumberTranslator` would be associated to a workspace (probably via a `shared_ptr` or `cow_ptr` since it changes rarely).
- Algorithm properties yield spectrum numbers, translated via `SpectrumNumber::makeIndexSet(...)`.
- Algorithms use
  
  ``` cpp
  #pragma omp parallel for
  for (size_t i=0; i<indexSet.size(); ++i) {
    // use indexSet[i] to access workspace
  }
  ```

### DIFF
--- a/Framework/Indexing/inc/MantidIndexing/IPartitioning.h
+++ b/Framework/Indexing/inc/MantidIndexing/IPartitioning.h
@@ -1,15 +1,16 @@
-#ifndef MANTID_INDEXING_SPECTRUMNUMBER_H_
-#define MANTID_INDEXING_SPECTRUMNUMBER_H_
+#ifndef MANTID_INDEXING_IPARTITIONING_H_
+#define MANTID_INDEXING_IPARTITIONING_H_
 
 #include "MantidIndexing/DllConfig.h"
+#include "MantidIndexing/SpectrumNumber.h"
 
 namespace Mantid {
 namespace Indexing {
 
-/** SpectrumNumber : TODO: DESCRIPTION
+/** IPartitioning : TODO: DESCRIPTION
 
-  @author Simon Heybrock
-  @date 2016
+  The main intention of this class is defining partitioning of all spectrum
+  numbers into subsets for an MPI-based Mantid run.
 
   Copyright &copy; 2016 ISIS Rutherford Appleton Laboratory, NScD Oak Ridge
   National Laboratory & European Spallation Source
@@ -32,22 +33,16 @@ namespace Indexing {
   File change history is stored at: <https://github.com/mantidproject/mantid>
   Code Documentation is available at: <http://doxygen.mantidproject.org>
 */
-class MANTID_INDEXING_DLL SpectrumNumber {
+class MANTID_INDEXING_DLL IPartitioning {
 public:
-  explicit SpectrumNumber(int64_t number) noexcept : m_number(number) {}
-  explicit operator int64_t() const noexcept { return m_number; }
-  template <class T> bool operator==(const T &other) const noexcept {
-    return m_number == SpectrumNumber(other).m_number;
-  }
-  template <class T> bool operator!=(const T &other) const noexcept {
-    return m_number != SpectrumNumber(other).m_number;
-  }
+  virtual ~IPartitioning() = default;
 
-private:
-  int64_t m_number;
+  virtual int numberOfPartitions() const = 0;
+  virtual int partitionIndex() const = 0;
+  virtual int partitionIndexOf(const SpectrumNumber &spectrumNumber) const = 0;
 };
 
 } // namespace Indexing
 } // namespace Mantid
 
-#endif /* MANTID_INDEXING_SPECTRUMNUMBER_H_ */
+#endif /* MANTID_INDEXING_IPARTITIONING_H_ */
--- a/Framework/Indexing/inc/MantidIndexing/IndexSet.h
+++ b/Framework/Indexing/inc/MantidIndexing/IndexSet.h
@@ -45,7 +45,7 @@ template <class T> class IndexSet {
 public:
   IndexSet(size_t fullRange);
   IndexSet(int64_t min, int64_t max, size_t fullRange);
-  IndexSet(const std::vector<size_t> indices, size_t fullRange);
+  IndexSet(const std::vector<size_t> &indices, size_t fullRange);
 
   /// Returns the size of the set.
   size_t size() const { return m_size; }
@@ -91,15 +91,12 @@ IndexSet<T>::IndexSet(int64_t min, int64_t max, size_t fullRange) {
 /// Constructor for a set containing all specified indices. Range is verified at
 /// construction time and duplicates are removed.
 template <class T>
-IndexSet<T>::IndexSet(const std::vector<size_t> indices, size_t fullRange)
+IndexSet<T>::IndexSet(const std::vector<size_t> &indices, size_t fullRange)
     : m_isRange(false) {
   // We use a set to create unique and ordered indices.
-  std::set<size_t> index_set;
-  for (const auto &index : indices) {
-    if (index >= fullRange)
-      throw std::out_of_range("IndexSet: specified index is out of range");
-    index_set.insert(index);
-  }
+  std::set<size_t> index_set(indices.cbegin(), indices.cend());
+  if (!index_set.empty() && *(index_set.rbegin()) >= fullRange)
+    throw std::out_of_range("IndexSet: specified index is out of range");
   m_indices = std::vector<size_t>(begin(index_set), end(index_set));
   m_size = m_indices.size();
 }
--- a/Framework/Indexing/inc/MantidIndexing/IndexSet.h
+++ b/Framework/Indexing/inc/MantidIndexing/IndexSet.h
@@ -70,7 +70,9 @@ private:
 };
 
 /// Constructor for a set covering the full range from 0 to fullRange-1.
-template <class T> IndexSet<T>::IndexSet(size_t fullRange) : m_size(fullRange) {}
+template <class T>
+IndexSet<T>::IndexSet(size_t fullRange)
+    : m_size(fullRange) {}
 
 /// Constructor for a set covering the range from min to max. Range is verified
 /// at construction time.
--- a/Framework/Indexing/inc/MantidIndexing/IndexSet.h
+++ b/Framework/Indexing/inc/MantidIndexing/IndexSet.h
@@ -1,0 +1,109 @@
+#ifndef MANTID_INDEXING_INDEXSET_H_
+#define MANTID_INDEXING_INDEXSET_H_
+
+#include "MantidIndexing/DllConfig.h"
+
+#include <set>
+#include <stdexcept>
+#include <vector>
+
+namespace Mantid {
+namespace Indexing {
+namespace detail {
+
+/** IndexSet : TODO: DESCRIPTION
+  CRTP
+  We want DetectorIndexSet and SpectrumIndexSet, with the same functionality,
+  but the types should be incompatible. Thus, we use IndexSet<DetectorIndexSet>
+  and IndexSet<SpectrumIndexSet>.
+
+  @author Simon Heybrock
+  @date 2016
+
+  Copyright &copy; 2016 ISIS Rutherford Appleton Laboratory, NScD Oak Ridge
+  National Laboratory & European Spallation Source
+
+  This file is part of Mantid.
+
+  Mantid is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 3 of the License, or
+  (at your option) any later version.
+
+  Mantid is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+template <class T> class IndexSet {
+public:
+  IndexSet(size_t fullRange);
+  IndexSet(int64_t min, int64_t max, size_t fullRange);
+  IndexSet(const std::vector<size_t> indices, size_t fullRange);
+
+  /// Returns the size of the set.
+  size_t size() const { return m_size; }
+
+  /// Returns the element at given index (range 0...size()-1).
+  size_t operator[](size_t index) const {
+    // This is accessed frequently in loops and thus inlined.
+    if (m_isRange)
+      return m_min + index;
+    return m_indices[index];
+  }
+
+protected:
+  ~IndexSet() = default;
+
+private:
+  bool m_isRange = true;
+  // Default here to avoid uninitialized warning for m_isRange = false.
+  size_t m_min = 0;
+  size_t m_size;
+  std::vector<size_t> m_indices;
+};
+
+/// Constructor for a set covering the full range from 0 to fullRange-1.
+template <class T> IndexSet<T>::IndexSet(size_t fullRange) : m_size(fullRange) {}
+
+/// Constructor for a set covering the range from min to max. Range is verified
+/// at construction time.
+template <class T>
+IndexSet<T>::IndexSet(int64_t min, int64_t max, size_t fullRange) {
+  if (min < 0 || min > max)
+    throw std::logic_error("IndexSet: specified min or max values are invalid");
+  if (max >= static_cast<int64_t>(fullRange))
+    throw std::out_of_range("IndexSet: specified max value is out of range");
+
+  // Bounds checked, cast should be fine in all cases.
+  m_min = static_cast<size_t>(min);
+  m_size = static_cast<size_t>(max - min + 1);
+}
+
+/// Constructor for a set containing all specified indices. Range is verified at
+/// construction time and duplicates are removed.
+template <class T>
+IndexSet<T>::IndexSet(const std::vector<size_t> indices, size_t fullRange)
+    : m_isRange(false) {
+  // We use a set to create unique and ordered indices.
+  std::set<size_t> index_set;
+  for (const auto &index : indices) {
+    if (index >= fullRange)
+      throw std::out_of_range("IndexSet: specified index is out of range");
+    index_set.insert(index);
+  }
+  m_indices = std::vector<size_t>(begin(index_set), end(index_set));
+  m_size = m_indices.size();
+}
+
+} // namespace detail
+} // namespace Indexing
+} // namespace Mantid
+
+#endif /* MANTID_INDEXING_INDEXSET_H_ */
--- a/Framework/Indexing/inc/MantidIndexing/Partitioning.h
+++ b/Framework/Indexing/inc/MantidIndexing/Partitioning.h
@@ -1,5 +1,5 @@
-#ifndef MANTID_INDEXING_IPARTITIONING_H_
-#define MANTID_INDEXING_IPARTITIONING_H_
+#ifndef MANTID_INDEXING_PARTITIONING_H_
+#define MANTID_INDEXING_PARTITIONING_H_
 
 #include "MantidIndexing/DllConfig.h"
 #include "MantidIndexing/SpectrumNumber.h"
@@ -7,7 +7,7 @@
 namespace Mantid {
 namespace Indexing {
 
-/** IPartitioning : TODO: DESCRIPTION
+/** Partitioning : TODO: DESCRIPTION
 
   The main intention of this class is defining partitioning of all spectrum
   numbers into subsets for an MPI-based Mantid run.
@@ -33,9 +33,9 @@ namespace Indexing {
   File change history is stored at: <https://github.com/mantidproject/mantid>
   Code Documentation is available at: <http://doxygen.mantidproject.org>
 */
-class MANTID_INDEXING_DLL IPartitioning {
+class MANTID_INDEXING_DLL Partitioning {
 public:
-  virtual ~IPartitioning() = default;
+  virtual ~Partitioning() = default;
 
   virtual int numberOfPartitions() const = 0;
   virtual int partitionIndex() const = 0;
@@ -45,4 +45,4 @@ public:
 } // namespace Indexing
 } // namespace Mantid
 
-#endif /* MANTID_INDEXING_IPARTITIONING_H_ */
+#endif /* MANTID_INDEXING_PARTITIONING_H_ */
--- a/Framework/Indexing/inc/MantidIndexing/Partitioning.h
+++ b/Framework/Indexing/inc/MantidIndexing/Partitioning.h
@@ -5,6 +5,8 @@
 #include "MantidIndexing/PartitionIndex.h"
 #include "MantidIndexing/SpectrumNumber.h"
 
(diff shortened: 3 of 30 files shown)


Framework/Indexing/inc/MantidIndexing/IPartitioning.h
```diff
@@ -0,0 +1,48 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+class MANTID_INDEXING_DLL IPartitioning {
```
OwenArnold: Minor point. I would rename to `Partitioning` over `IPartitioning`

SimonHeybrock:
@OwenArnold The more complex case, i.e., not just using plain index translation is as follows, and I am still looking for a nice way to do this:

- We want to construct `SpectrumInfo` (and lower-level bits of the instrument) and need a `std::vector<Spectrum>` for all spectra on this MPI rank, where `Spectrum` contains a `std::vector<DetectorID>`.

- From a Nexus file we can get a *global* vector of spectra, i.e., a vector containing spectra for all MPI ranks.

- We could extend `SpectrumNumberTranslator` to translate this into the local vector we need, for example:
  ```cpp
  std::vector<Spectrum> SpectrumNumberTranslator::makeVectorForThisPartition(
      std::vector<std::pair<SpectrumNumber, Spectrum>>);
  ```

Do you like these semantics (apart from naming)?

Framework/Indexing/inc/MantidIndexing/RoundRobinPartitioning.h
```diff
@@ -0,0 +1,65 @@
...
+      throw std::logic_error(
+          "RoundRobinPartitioning: There must be at least 1 partition");
+    if (m_partition < 0 || m_partition >= m_partitions)
+      throw std::out_of_range("RoundRobinPartitioning: Index of this partition "
+                              "is out of range 0...numberOfPartitions-1");
+  }
+
+  int numberOfPartitions() const override { return m_partitions; }
+  int partitionIndex() const override { return m_partition; }
+  int partitionIndexOf(const SpectrumNumber &spectrumNumber) const override {
```
OwenArnold: I think this naming could lead to confusion. `partitionIndex` and `partitionIndexOf` are doing very different and returning very different things.

How about `indexOfThisPartition` and `indexOfSpectrumNumber`? Or some other way to highlight the difference?

Framework/Indexing/inc/MantidIndexing/IndexSet.h
```diff
@@ -0,0 +1,111 @@
...
+  m_size = static_cast<size_t>(max - min + 1);
+}
+
+/// Constructor for a set containing all specified indices. Range is verified at
+/// construction time and duplicates are removed.
+template <class T>
+IndexSet<T>::IndexSet(const std::vector<size_t> indices, size_t fullRange)
+    : m_isRange(false) {
+  // We use a set to create unique and ordered indices.
+  std::set<size_t> index_set;
```
OwenArnold: Good move. I was going to check that you'd done this.

Framework/Indexing/inc/MantidIndexing/RoundRobinPartitioning.h
SimonHeybrock: You are right, still not happy about the naming though, will think some more.

SimonHeybrock: @OwenArnold I think the naming difficulty is due to having two different concepts in one class, `Partitioning` and `Partition`. I am considering splitting, but then the `SpectrumNumberTranslator` cosntructor gets more messy.

Framework/Indexing/inc/MantidIndexing/SpectrumNumberTranslator.h
```diff
@@ -0,0 +1,91 @@
...
+      auto number = spectrumNumbers[i];
+      auto partition = partitioning.partitionIndexOf(number);
+      m_partitions[number] = partition;
+      if (partition == m_partition)
+        m_indices[number] = currentIndex++;
+    }
+  }
+
+  // Full set
+  SpectrumIndexSet makeIndexSet() { return SpectrumIndexSet(m_indices.size()); }
```
OwenArnold: @SimonHeybrock Is this right? The `SpectrumIndexSet` you get back will only have the size set

SimonHeybrock: @OwenArnold Yes, if only the size is set we basically have a low-overhead `IndexSet` that represents "all indices". In that case `IndexSet` is just a 1:1 mapping from input to output, without the need to access any index vectors.

Framework/Indexing/inc/MantidIndexing/SpectrumNumberTranslator.h
```diff
@@ -0,0 +1,91 @@
...
+
+private:
+  struct SpectrumNumberHash {
+    std::size_t operator()(const SpectrumNumber &spectrumNumber) const {
+      return std::hash<std::int64_t>()(
+          static_cast<const int64_t>(spectrumNumber));
+    }
+  };
+
+  int m_partition;
```
OwenArnold: Is it worth making this `const`?

Framework/Indexing/test/RoundRobinPartitioningTest.h
```diff
@@ -0,0 +1,47 @@
...
+    return new RoundRobinPartitioningTest();
+  }
+  static void destroySuite(RoundRobinPartitioningTest *suite) { delete suite; }
+
+  void test_constructor_failures() {
+    TS_ASSERT_THROWS(RoundRobinPartitioning(-1, 0), std::logic_error);
+    TS_ASSERT_THROWS(RoundRobinPartitioning(0, 0), std::logic_error);
+    TS_ASSERT_THROWS(RoundRobinPartitioning(1, -1), std::logic_error);
+    TS_ASSERT_THROWS(RoundRobinPartitioning(1, 1), std::logic_error);
+  }
```
OwenArnold: Good testing.

Framework/Indexing/inc/MantidIndexing/SpectrumNumber.h
```diff
@@ -0,0 +1,53 @@
+#ifndef MANTID_INDEXING_SPECTRUMNUMBER_H_
+#define MANTID_INDEXING_SPECTRUMNUMBER_H_
+
+#include "MantidIndexing/DllConfig.h"
+
+namespace Mantid {
+namespace Indexing {
+
+/** SpectrumNumber : TODO: DESCRIPTION
```
OwenArnold: Could use a description.

Framework/Indexing/inc/MantidIndexing/SpectrumNumberTranslator.h
```diff
@@ -0,0 +1,91 @@
...
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+class MANTID_INDEXING_DLL SpectrumNumberTranslator {
+public:
+  SpectrumNumberTranslator(const std::vector<SpectrumNumber> &spectrumNumbers,
+                           const IPartitioning &partitioning)
+      : m_partition(partitioning.partitionIndex()) {
+    size_t currentIndex = 0;
+    for (size_t i = 0; i < spectrumNumbers.size(); ++i) {
+      auto number = spectrumNumbers[i];
```
OwenArnold: You loose your type safety here. I would consider having explicit rather than implicit conversion methods on things like `SpectrumNumber`.

SimonHeybrock: Could you elaborate? `SpectrumNumber::operator int64_t ()` is `explicit`?

Framework/Indexing/inc/MantidIndexing/SpectrumNumberTranslator.h
```diff
@@ -0,0 +1,84 @@
...
+
+private:
+  struct SpectrumNumberHash {
+    std::size_t operator()(const SpectrumNumber &spectrumNumber) const {
+      return std::hash<std::int64_t>()(
+          static_cast<const int64_t>(spectrumNumber));
+    }
+  };
+
+  int m_rank{0};
```
OwenArnold: Deliberate fixing of rank?

Framework/Indexing/test/SpectrumNumberTranslatorTest.h
```diff
@@ -0,0 +1,76 @@
...
+    TS_ASSERT_EQUALS(set1.size(), 2);
+    TS_ASSERT_EQUALS(set1[0], 0);
+    TS_ASSERT_EQUALS(set1[1], 1);
+    auto set2 = translator.makeIndexSet(makeSpectrumNumbers({4, 5}));
+    TS_ASSERT_EQUALS(set2.size(), 2);
+    TS_ASSERT_EQUALS(set2[0], 2);
+    TS_ASSERT_EQUALS(set2[1], 3);
+  }
+};
+
```
OwenArnold: These tests are pretty good. I wonder (again about naming) if the *thing* that the `SpectrumNumberTranslater` is translating into should be specified in the name. This type is converting `SpectrumNumber` to `SpectrumIndex` (via the `IndexSet` proxy.

Framework/Indexing/inc/MantidIndexing/RoundRobinPartitioning.h
SimonHeybrock: @OwenArnold See my latest commit. I basically split this class into two, `Partition` and `PartitionIndex`.

Framework/Indexing/test/SpectrumNumberTranslatorTest.h
SimonHeybrock: @OwenArnold `SpectrumNumberToIndexTranslator`?

Framework/Indexing/inc/MantidIndexing/Partitioning.h
```diff
@@ -38,8 +39,11 @@ class MANTID_INDEXING_DLL Partitioning {
   virtual ~Partitioning() = default;
 
   virtual int numberOfPartitions() const = 0;
-  virtual int partitionIndex() const = 0;
-  virtual int partitionIndexOf(const SpectrumNumber &spectrumNumber) const = 0;
+  virtual PartitionIndex
+  indexOf(const SpectrumNumber &spectrumNumber) const = 0;
```
OwenArnold: Good solution.

OwenArnold:
@SimonHeybrock, Regarding your question above about making the Specta... 

`std::vector<Spectrum> SpectrumNumberTranslator::makeVectorForThisPartition(
    std::vector<std::pair<SpectrumNumber, Spectrum>>);`

* Yes. I like this. As you said, naming is not sorted. but I think `makeSpectraForThisPartition` or `makeSpectrumsForThisPartition` would be slightly better names. 
* Method should be `const` (minor point)
* Method should probably take argument as `lvalue` ref. (minor point)

What about lifetimes? I don't yet have a sense of what owns what, and how long-lived some of these things are. Given that `SpectrumNumberTranslator` holds a `Partitioning` strategy, does that also mean that a change in `Partitioning` involves rebuilding the `SpectrumNumberTranslator` too? What does the translator attach to?

SimonHeybrock:
@OwenArnold 
- I do not know if the `Spectrum` case that we need for constructing `SpectrumInfo` is the only use case, so I had kept the name more general. Can you think of other scenarious?
- I have ignored changing partitioning for now. It is not a given that it will become necessary, and even if, I do not think we can make a qualified guess at the precise requirements.
- Translators would basically have the same lifetime as workspaces (or live until the last owning workspace is deleted). Basically this is just wrapping the index translation that is currently part of `MatrixWorkspace`.

OwenArnold:
@SimonHeybrock.

There are cases where you of course end up with a different mapping, but these should always be mirrored in the `SpectrumInfo` ultimately. So I think the answer is no.

The rest of this sounds good.

Framework/Indexing/inc/MantidIndexing/IndexSet.h
```diff
@@ -0,0 +1,111 @@
...
+
+  // Bounds checked, cast should be fine in all cases.
+  m_min = static_cast<size_t>(min);
+  m_size = static_cast<size_t>(max - min + 1);
+}
+
+/// Constructor for a set containing all specified indices. Range is verified at
+/// construction time and duplicates are removed.
+template <class T>
+IndexSet<T>::IndexSet(const std::vector<size_t> indices, size_t fullRange)
```
quantumsteve: Did you intend to pass `indices` by value?

Framework/Indexing/inc/MantidIndexing/IndexSet.h
```diff
@@ -0,0 +1,111 @@
...
+    : m_isRange(false) {
+  // We use a set to create unique and ordered indices.
+  std::set<size_t> index_set;
+  for (const auto &index : indices) {
+    if (index >= fullRange)
```
quantumsteve: This check would be faster after the indices are sorted.

Framework/Indexing/inc/MantidIndexing/IndexSet.h
SimonHeybrock: No I did not, thanks, I'll fix that.

Framework/Indexing/inc/MantidIndexing/IndexSet.h
SimonHeybrock: Good point, I can just create the set and check the last element!

Framework/Indexing/inc/MantidIndexing/IndexSet.h
```diff
@@ -0,0 +1,111 @@
...
+      throw std::out_of_range("IndexSet: specified index is out of range");
+    index_set.insert(index);
+  }
+  m_indices = std::vector<size_t>(begin(index_set), end(index_set));
```
quantumsteve: How does using `std::set` compare with using  `std::sort`, `std::unique`, and `std::erase` to do this in place? Are the number of indices large enough for the higher memory consumption to matter?

SimonHeybrock: @quantumsteve Looking at how it is currently used, I think no. For most algorithms (and their properties), we either use all indices, a range (`min...max`, I still have to think of a good solution how that can be done with MPI), or a list of indices entered by the user. Being entered by the user implies that the list is typically small, I think.

Let my know if you are aware of other cases.

Framework/Indexing/inc/MantidIndexing/SpectrumNumberTranslator.h
```diff
@@ -0,0 +1,154 @@
...
+    }
+    if (spectrumNumbers.size() != m_partitions.size())
+      throw std::logic_error("SpectrumNumberTranslator: The vector of spectrum "
+                             "numbers contained duplicate entries.");
+  }
+
+  // Full set
+  SpectrumIndexSet makeIndexSet() { return SpectrumIndexSet(m_indices.size()); }
+
+  SpectrumIndexSet makeIndexSet(SpectrumNumber min, SpectrumNumber max) {
```
SimonHeybrock: @OwenArnold The implementation of this min/max variant is really nasty/ugly. It would be much simpler if we would use a `std::map` instead of `std::unordered_map`, but I am not sure if that would cause trouble with performance. Do you have a feeling of how frequently different translation requests would occur?

SimonHeybrock:
@OwenArnold Important point regarding (user) interfaces:

- (Local) workspace indices lose their meaning in an MPI run of Mantid.
- We could force all algorithm interfaces to use only `SpectrumNumber`, which can can uniquely translate with what has been done in this PR so far.
- It is not clear to me if this is a step we want to take. If we assume that we do not, we need to introduce the notion of a *global* workspace index.
- Global workspace indices would also be used when displaying data, e.g., a table representation of the data in a `MatrixWorkspace`, as we have ti now. This would involve (temporarily) transferring data from all ranks to the rank on which the visualization is done. This is something far beyond the scope of what is discussed here, but we have to ensure that our indexing mechanism can cope with this requirement. Potentially this case can be dealt with without any changes. We simply need to create a non-distributed workspace, one rank has all the data and thus the local indices on that rank are equivalent to the global indices.

Question: Do you think we need to introduce a global workspace index type, or do you see other solutions, such as strictly enforcing the use of only `SpectrumNumber`?

SimonHeybrock:
@OwenArnold Another requirement: The `MatrixWorkspace` interface provides various methods for translating from `DetectorID` to workspace indices (*not* detector indices). This is used in various algorithms that accept detector IDs as input properties but actually work based on spectra. I am not sure what the best option is in this case:

1. Get rid of such properties. I don't think this is realistically possible/sensible.
2. Provide such means of translation, for example `SpectrumIndexSet DetectorIDTranslator::makeSpectrumIndexSet(const std::vector<DetectorID> &)`.
3. other?

I think there are probably quite a few subtleties hidden here. For example, `ExtractSpectra` extracts a spectrum if at least one of the specified detectors that is part of the spectrum, even if all the other detectors are not. This may make sense, but I fear there may be other similar cases where it does not.

Framework/Indexing/inc/MantidIndexing/SpectrumNumberTranslator.h
OwenArnold: @SimonHeybrock. I'm not sure what this will do to performance. I don't have a feel for how frequently this would be performed.

OwenArnold:
@SimonHeybrock 
```
Question: Do you think we need to introduce a global workspace index type, or do you see other solutions, such as strictly enforcing the use of only SpectrumNumber?
```
I think that indexes are the wrong tool when we get to specifying partitions. My personal preference would be to identify spectra in a way that was inherently unique and unchanging across ranks.

SimonHeybrock:
@OwenArnold A global workspace index would be unique and unchanging, it is equivalent to our current (non-MPI) definition of workspace index. For example:

```cpp
SpectrumNumber       1 2 3 4 5 6 7 8
Rank/Partition       0 1 2 0 1 2 0 1
workspace index      0 0 0 1 1 1 2 2
GlobalWorkspaceIndex 0 1 2 3 4 5 6 7
```

OwenArnold:
@SimonHeybrock I think your proposal 2. sounds reasonable 
```
SpectrumIndexSet DetectorIDTranslator::makeSpectrumIndexSet(const std::vector<DetectorID> &)
```
If we want to allow for the `ExtractSpectrum` case then maybe we should include some additional notation in the name. For example `makeSpectrumIndexInclusiveSet` etc.

SimonHeybrock:
@OwenArnold So we could have two variants?

```cpp
enum class PartialSpectrumMode { Include , Exclude };
SpectrumIndexSet DetectorIDTranslator::makeSpectrumIndexSet(
    const std::vector<DetectorID> &,
    PartialSpectrumMode mode);
```

OwenArnold:
@SimonHeybrock. That signature looks good to me.

SimonHeybrock:
@OwenArnold There is yet another problem: In the simplest case of an MPI-based Mantid run, monitors should be present as cloned copies on all ranks.

However, the concept of a *monitor* does not exist in this `Indexing` library and thus in the `Partitioning` class. Do you see a solution for this? Do we simply need to make *monitor* something that `Indexing` knows about?

OwenArnold:
@SimonHeybrock. Yes! excellent spot. Monitors must be present on all ranks. (I'll make a note of this for Instrument 2.0). An interesting aside of this is that for load-balancing is what we will do about the additional data rates we'll see for processing monitors in event mode. One of our early discussions centred around the fact that we would want to process the Monitor data on a dedicated rank. I guess that this is still possible, but I'm not an MPI expert.

I guess that making the `Indexing` library aware of monitors is a reasonable solution.

SimonHeybrock:
@OwenArnold I had discussed monitors also in my talk at the developer meeting (Jan. 2016), see some slide towards the end: https://github.com/mantidproject/documents/blob/master/Presentations/DevMeetings/2016-01/mantid-mpi-new.pdf.

The easiest (by far) solution is to have monitors on all ranks. Most existing algorithms will then work without modification.

Only if that is not sufficient we may need a dedicated rank or even several dedicated ranks for monitors. However, that comes with huge extra development cost. There needs to be a fast an efficient mechanism for getting information about monitors on all ranks, e.g., for a normalization.

Currently I do need see a reason for going with the latter. We don't even know yet if the ESS monitors will run in event mode.

OwenArnold:
I'm not sure that the ESS will be able to determine whether the monitors will run in either event mode or histogram mode as a rule. Most of the time you can run in histogram mode, but for stop-flow experiments where time slicing is required, it becomes necessary if monitor normalisation is also required as part of the workflow.

SimonHeybrock:
@OwenArnold I implemented a first version of what we discussed regarding monitors, I don't see any major issues for now.

SimonHeybrock:
What I do not like is that we have to route all those monitor related bit from the child classes like `RoundRobinPartitioning` into `Partitioning`. The former object is actually completely independent of this concept. Maybe we should restructure this a bit?

How about reverting `Partitioning` and its derived classes to how they were before adding the monitor functionality, and moving that functionality into a new class:

```cpp
template <class T>
class PartitionInfo {
public:
  PartitionInfo(const int numberOfPartitions, const PartitionIndex partition,
               const MonitorStrategy monitorStrategy,
               std::vector<SpectrumNumber> monitors)
        : T(monitorStrategy == CloneOnEachPartition ? numberOfPartitions : numberOfPartitions -1) {
    // init other bits as now in Partitioning
  }

  // interface as Partitioning has now

private:
  const T m_partitioning;
  const std::vector<SpectrumNumber> m_monitors;
  // ...
};
```

OwenArnold:
Looks reasonable to me.

SimonHeybrock:
Maybe templating this is not necessary, a base class pointer to `Partitioning` should be sufficient.

SimonHeybrock:
I have browsed some code that uses the `MatrixWorkspace` interface to deal with detector IDs. This uncovered two more requirements, or problems:

1. Some algorithms, e.g., `LoadEventNexus` need to translate a series of detector IDs into workspace indices. For example, `LoadEventNexus` translates one ID per event. The implications are:
  - `std::map` may be too slow.
  - So far I always assumed that the output of the translator could be a set (e.g., `SpectrumIndexSet`). When using such a set we can silently drop anything that is not on this rank and client code can just iterate over the set (the typical scenario here is iteration over all spectra in a workspace, or a subset specified by algorithm properties). Now, in this case there is data attached to each detector ID, e.g., a TOF, and an ID will show up multiple times. This could we solved by passing, e.g., `std::vector<std::pair<DetectorID, T>>` to a new method `DetectorIDTranslator::makeIndexedVector()`. The problem is that this would require significant restructuring of quite a lot of client code.
  - If we want to avoid the significant restructuring, we could instead provide methods for translating individual detector IDs. However, the ID can be on another rank, so the client code would need to check for that first, or handle exceptions. This would imply that those algorithms would need to be aware of the fact that Mantid is running with MPI.

2. Actually very similar: Some algorithms get detectors at certain positions and then the corresponding spectrum. Again, due to distribution onto MPI ranks, it can now happen that there is no spectrum at that position. We would either need to restructure or deal with the case that there is no spectrum and skip.

SimonHeybrock:
This has been solved in (several) other PRs.

PR #16837, opened 2016-07-11
Serialise composite function's local attributes in string

### LINKED ISSUE #16835: MultiDomainFunction loses domains when serialised
#### Expected behavior
Converting a fit function to its string representation and back again should not lose any information

#### Actual behavior
Domains are lost when serialising a `MultiDomainFunction`

#### Steps to reproduce the behavior
Attempt to clone a `MultiDomainFunction` and note that the domains are lost (as cloning involves going via a string)

---

The issue here is that local attributes are not serialised in the string, and the domains are stored as a local attribute.

tom-perkins:
When converting a `CompositeFunction` to a string, make sure that any local attributes are included
in the string.

This is useful for a `MultiDomainFunction`, where the domains are stored as a local attribute.
Now cloning a `MultiDomainFunction`, or going to and from its string representation, should not lose the domains.

**To test:**
- Code review
- Unit tests have been added to check that cloning preserves the number of domains.
  Check that the tests fail before this change and pass after it.
- Try the following:

``` python
string = "composite=MultiDomainFunction;name=FlatBackground,A0=0,$domains=i; \
name=FlatBackground,A0=1,$domains=(0,1);name=FlatBackground,A0=2,$domains=(0,2)"

func = FunctionFactory.createInitialized(string)
print func
```

The function string printed out should look like the input, in particular it should have the domains in it.

Fixes #16835.

_Release notes:_ I think this does not need to be in the release notes as it is an 'internal' change?

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/API/src/CompositeFunction.cpp
+++ b/Framework/API/src/CompositeFunction.cpp
@@ -68,12 +68,21 @@ std::string CompositeFunction::asString() const {
     ostr << ';';
   }
   for (size_t i = 0; i < nFunctions(); i++) {
+    const auto localAttr = this->getLocalAttributeNames();
     IFunction_sptr fun = getFunction(i);
     bool isComp =
         boost::dynamic_pointer_cast<CompositeFunction>(fun) != nullptr;
     if (isComp)
       ostr << '(';
     ostr << fun->asString();
+    for (const auto &localAttName : localAttr) {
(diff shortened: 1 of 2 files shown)


KarlPalmen:
Test described works as expect. Unit test added looks good. :shipit:

Issue #15602, opened 2016-03-08
Fix and Speed up project saving


NickDraper:
Fix and Speed up project saving
- Save project as Python script (all graphs, viewers to expose python)
- Performance options?
- Partial Save functionality (select graphs / workspaces then only save what is needed)
- Express Save - Only save History from a workspace since it was last loaded, then rerun on Load Project, then regenerate graphs
- Complete Save (as currently)

sub tickets
- [x] #3383
- [x] #16672
- [x] #16719 
- [x] #16768
- [x] #16960
- [x] #16961
- [x] #16962
- [x] #16963
- [x] #16964
- [x] #16965
- [x] #17060 
- [x] #17228
- [ ] #17445
- [x] #17969

NickDraper:
Specific notes from unscripted testing:

1D plot: quite good, retained zooming errors and text lables.
Waterfall plot: retaind appearance, but lost water fall plot option boxes
2D plot: lost arrows and lines (2D would probably too as well), Lost axis scaling and colour bar log and scaling
qtiplot table: good
3D plot basic recreation ok

samueljackson92:
Some more notes on this:
 - Looks like most of the code for this is in the god-like ApplicationWindow. It would be nice to pull it out of there instead of just adding to that class.
 - There appears to be a single exposed method from the ApplicationWindow class (```saveProjectAs()```) which causes mantid to crash if you give it the tilde on OSX, i.e. ```saveProjectAs("~/test.mantid")``` but appears to work otherwise.

samueljackson92:
I'm still currently working on #16719. I've managed to get wrangle Mantid so that we can actually test loading and saving automatically. This afternoon I got serious with unscripted testing and manually ran though a whole bunch of scenarios. I've compiled a list of tests below and noted what does/doesn't work. I'm sure there's even more cases which have been missed, but we can add to it as they are found.

All of the below were carried out on Mac OSX 10.11.5 (El Capitan). 

| No. | Description                                                          | Pass/Fail | Notes                                                                                                                                                                                                                                                              | Tickets |
|-----|----------------------------------------------------------------------|-----------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|
| 1   | Save project with workspaces without plotting                        | Fail      | Project is saved, but the open workspace is not included in the project.                                                                                                                                                                                           |         |
[...]

PR #16846, opened 2016-07-12
Convert MdiSubWindow to use dynamic factory

### LINKED ISSUE #16768: Convert project serialising to use factory
@NickDraper would like to see project serialisation implemented following a factory pattern. We already have a IProjectSerialisable class which defines methods which GUI components use to convert to and from a string.

The real challenge is that a lot of the GUI components currently depend on having an instance of application window during instantiation. Ideally this should not be the case. GUI components should not depend on knowing about application window and application window should edit it to connect signals, etc.

samueljackson92:
This converts classes derived from MdiSubWindow which are serialised to Mantid project files to be created by a new dynamic factory, drastically simplifying how each object is loaded at a high level. 

Each window type can now be created using a default constructor from the factory. The `loadFromProject` function will then initialise and load setting for each window. The loading logic for each type of window is now contained within the class itself, i.e. MdiSubWindows now completely know how to load themselves from file. 

Unfortunately there is still a large dependance on ApplicationWindow, but reducing that would be getting beyond the scope of this ticket.

This PR should not add any new functionality or fix any bugs, but hopefully should not break anything either.

**To test:**

A lot of code for project loading/saving has been moved around, but on the surface it should appears as if nothing has changed.
- Check all the tests pass. (especially MantidPlotProjectSerialise!)
- Check that all types of windows can be saved and loaded back. For a list of what should and shouldn't work see issue #15602. Types of windows should include:
  - MantidMatrix
  - Matrix
  - Note
  - Table
  - TableStatistics
  - 1D plots
  - 2D plots
  - 3D plots
  - InstrumentWindow
  - ScriptWindow
- Code review.

Fixes #16768.

---
#### Reviewer

##### Code Review
##### Functional Tests


samueljackson92:
Talking to Martyn, we should keep the default contractors and switch to using a [variadic function templates](http://en.cppreference.com/w/cpp/language/parameter_pack) to create a factory that allows construction of objects with non-default constructors. 

Unfortunately the dynamic factory implementation does not support this already and modifying it to achieve this would:
 1. Not work because the ```createInstance``` method(s) on the factory are virtual and cannot be templated.
 2. Touch too many things beyond the scope of this ticket.

Therefore after much discussion we'll proceed by creating a new, light weight dynamic factory implementation that functions similarly to the standard dynamic factory.

MantidPlot/src/Graph3D.cpp
```diff
@@ -2483,120 +2494,437 @@ Graph3D::~Graph3D() {
...
+  boost::split(lineVec, lines, boost::is_any_of("\n"));
+
+  // First line is name\tdate
+  const std::string firstLine = lineVec[0];
+  boost::split(valVec, firstLine, boost::is_any_of("\t"));
+
+  if (valVec.size() < 2)
+    return nullptr;
+
+  QString caption = QString::fromStdString(valVec[0]);
```
AntonPiccardoSelg: you can use `auto` here and above/below

MantidPlot/src/Graph3D.h
```diff
@@ -77,9 +81,43 @@ class Graph3D : public MdiSubWindow, public Mantid::IProjectSerialisable {
     Cones = 4
   };
 
+  enum SurfaceFunctionType {
```
AntonPiccardoSelg: is an enum class an option here?

samueljackson92: @AntonPiccardoSelg Yes. My C++ language knowledge grows daily.

MantidPlot/src/WindowFactory.h
```diff
@@ -0,0 +1,203 @@
...
+
+/// Forward declaration of a specialisation of SingletonHolder for
+/// AlgorithmFactoryImpl (needed for dllexport/dllimport) and a typedef for it.
+#ifdef _WIN32
+// this breaks new namespace declaraion rules; need to find a better fix
+template class Mantid::Kernel::SingletonHolder<WindowFactoryImpl>;
+#endif /* _WIN32 */
+
+typedef Mantid::Kernel::SingletonHolder<WindowFactoryImpl> WindowFactory;
+}
```
AntonPiccardoSelg: This has changed recently: https://github.com/mantidproject/mantid/wiki/Exporting-Symbols-of-Shared-Libraries-in-Mantid

MantidPlot/src/WindowFactory.h
```diff
@@ -0,0 +1,203 @@
...
+
+/* Cross platform string case insensitive string comparator functor.
+ */
+struct CaseInsensitiveStringComparator {
+  bool operator()(const std::string &s1, const std::string &s2) const {
+#ifdef _MSC_VER
+    return _stricmp(s1.c_str(), s2.c_str()) < 0;
+#else
+    return strcasecmp(s1.c_str(), s2.c_str()) < 0;
+#endif
```
AntonPiccardoSelg: * Can this be reused from above (or the other way around)?
* Can you use the `boost::iequals`?

AntonPiccardoSelg:
There are quite a few opportunities for the usage of `auto`. If they were added during this work, it might be a good opportunity to change that, if they were there before, then don't worry too much about it.

samueljackson92:
@AntonPiccardoSelg a lot of this is code that has simply been moved around. Future tickets will be bug fixes & improvements now loading code is in the same place, so it might be nicer to catch it then. However I have replaced some which I was aware that I had added (e.g. the Graph3D* you pointed out above).

MantidPlot/src/WindowFactory.h
samueljackson92: @AntonPiccardoSelg Yes. This code is also in ```DynamicFactory.h``` which would be worth replacing as well (although perhaps not in this issue). Is there a sensible common place we can put this comparator?

AntonPiccardoSelg:
@samueljackson92 Yes, then don't worry about it.

MantidPlot/src/WindowFactory.h
AntonPiccardoSelg: Maybe it is worth checking with the person who wrote this in the first place, why they did it. It could well be that it is much more performant, although chances are that the boost version does the same thing just abstracted away.

I am not aware of a sensible place, but you could create one. If DynamicFactory uses the same thing it would have to be in `Kernel`, but it might be worth getting Martyn's opinion on this.

MantidPlot/src/WindowFactory.h
samueljackson92: As discussed this isn't applicable because it's in MantidPlot.

MantidPlot/src/TSVSerialiser.h
```diff
@@ -33,6 +33,17 @@
...
   File change history is stored at: <https://github.com/mantidproject/mantid>
 */
+
+struct CaseInsensitiveStringComparator {
+  bool operator()(const std::string &s1, const std::string &s2) const {
+#ifdef _MSC_VER
+    return stricmp(s1.c_str(), s2.c_str()) < 0;
+#else
+    return strcasecmp(s1.c_str(), s2.c_str()) < 0;
+#endif
```
samueljackson92: I've changed my mind on this one:
 1. Note that be above is not equality, but checking if s1 is less than s2. 
 2. Boost versions add locale support, which we are not bothered about and the C lib ones should therefore be more efficient in this case.
 3. Yes this isn't the nicest code, but after the next few commits there should be only one instance.

What do you think? If you don't agree I could find the equivalent compare function for boost?

AntonPiccardoSelg:
@samueljackson92 There seems to be a merge conflict.

AntonPiccardoSelg:
*Code look good to me.
*Saved all types of windows in a project and reloaded them. Worked fine on Windows and Ubuntu. 
Happy to :shipit:

MantidPlot/src/WindowFactory.h
martyngigg: You're correct that the symbol exporting doesn't apply to MantidPlot but this code suggests, with the comments etc that it is involved in that mechanism somehow.

I think we should clean up the code at the bottom here so that it more closely follows that as defined in the link provided by @AntonPiccardoSelg. At the very least the `SingletonHolder` template needs to be specialized in the Mantid::Kernel namespace. I think all we need to do is remove the `EXTERN_MANTID_API` from the example.

Framework/Kernel/inc/MantidKernel/case_insensitive_map.h
```diff
@@ -0,0 +1,31 @@
+#ifndef CASE_INSENSITIVE_MAP_H
```
martyngigg: Is there a reason that the filename doesn't follow the standard conventions without underscores?

samueljackson92: Not really. Wasn't sure if we used camel case for just classes and lower case underscore for stuff that isn't e.g. ```make_*.h``` or whether that was just due mimicking std lib support.

martyngigg: I think given that the alias class name is CamelCase then it would be better if the filename reflected that. Could you change it to `CaseInsensitiveMap`?

samueljackson92: Yes, seems like a silly decision in hindsight. Already made the change, just checking it builds.

martyngigg:
Thanks for making that update. Happy to merge now.

Issue #16601, opened 2016-06-13
Multi-dataset fitting: global flag lost when adding new function


tom-perkins:
_Reported by James Lord_
#### Expected behaviour

If a function parameter is set as global, this flag should not be unset
#### Actual behaviour

Global flag is lost on existing function parameters when a new function is added to the composite function
#### Steps to reproduce the behaviour
- Generate some workspaces
- Open multi-dataset fitting interface and load these workspaces
- Set up a function e.g. `FlatBackground` and set _f0.A0_ global
- Add a new function e.g. `LinearBackground`
  - _f0.A0_ from the flat background is no longer global

tom-perkins:
This is more than just the global flag:

- Ties are also lost when adding a new function, although you can't see this until you change the spectrum or open the "edit parameters" dialog as it stays on screen.

- Values of non-global parameters are reset to those of the first spectrum, rather than keeping their individual values.

This problem is often seen when you run a fit and then want to adjust the fitting function and fit again.

tom-perkins:
Other ways to lose the ties and global flags include:
- Loading a new dataset
- Checking or unchecking the NumDeriv box

tom-perkins:
I think [this](https://github.com/mantidproject/mantid/blob/6a3e45f40997de8022656fd816172f8c92cb0515/MantidQt/MantidWidgets/src/FunctionBrowser.cpp#L406-L409) is where the global flags/ties are lost - it is called from `attributeChanged`.

tom-perkins:
This bug annoys a number of users, aim to fix for 3.9.

Anders-Markvardsen:
Hi @AnthonyLim23 moved some tom perkins assigned 3.9 into your space. This one might be for the generic multiple scattering interface it which case it would be worth discussing it with roman. I have put it in temporary holding since I don't recall the muon group bringing this up as urgent in out last meeting with them

PR #16855, opened 2016-07-13
Muon Analysis multi-dataset fitting

### LINKED ISSUE #15518: Multi-dataset fitting in MuonAnalysis interface

tom-perkins:
The Muon Analysis interface has been extended to enable multi-dataset fitting. Changes have been made to the "Data Analysis" and "Results Table" tabs.
The document of the meeting with scientists to discuss requirements is [here](https://github.com/mantidproject/documents/blob/110a96335752282cce54ad0623a495f9c3a4bd05/Design/Muon/MuonAnalysis_multi_fitting.md).

Fixes #15518

**To test:**
A new [unscripted testing guide](https://github.com/mantidproject/mantid/blob/54495bb2b985f63b04870ea1a3d5d77ec35164d7/docs/source/interfaces/Muon_Analysis_unscripted_testing_guide.rst) has been written, which exercises all of the new functionality. Working through this should be enough to test the new/changed features. The data is available at `\\olympic\babylon5\Scratch\TomPerkins\test_data_16855`, with built packages from the build servers.

Automated testing has been improved as part of this work, and much previously untested code now has unit tests.

Instrument scientists have tested alpha versions of this branch, and further feedback will be sought on the current version before the release.

**Description of changes:**
- The overriding aim was to reuse existing, working code - the main changes just put a new 'skin' on the Data Analysis tab.
- Fitting is still done by the `MuonFitPropertyBrowser` (a specialised `FitPropertyBrowser`), just as before.
- Extra widgets are added to control the function and the data:
  - _Function_: used the existing `FunctionBrowser` from the general multi-dataset fitting interface
  - _Data_: a new widget was created, which acts as a 'humble view' as far as possible
- These widgets act like "views", and each has a "presenter". The "model" is the existing FitPropertyBrowser. Note that this is not strictly MVP, but is somewhat like it (reusing existing code  was prioritised over strict adherence to the pattern). Both presenters are unit tested.
- Data loading and processing code was refactored out of the MuonAnalysis GUI into a separate class, where it is now tested.
- After a simultaneous fit is completed, the output of `Fit` is processed into the same form as the output of a regular fit, so that the existing code for generating results tables can cope with it.
- The code to generate results tables on the "Results Table" tab has also been refactored out of the GUI and tested. It has been extended to support tables for the new kinds of fits now supported.

The "Files changed" view looks like there have been a lot of changes at first, but much of this is just refactoring existing code into new classes, testing previously untestable code, or adding abstract base classes for mocking purposes. 

**Feature key:**
There is an option on the "Settings" tab of the interface called "Compatibility mode".
This makes the UI appear and behave as in previous versions of Mantid.
The point of this setting is so that users will not be blocked from fitting data if there turns out to be a bug in the new UI. It should reduce the risk of merging this branch.
(The default value for this setting will be changed once scientists are happy to do so)

**Documentation:**
[User documentation](https://github.com/mantidproject/mantid/blob/b509a79efbab8a1a74b1872e4aa1baeee6c3faed/docs/source/interfaces/Muon_Analysis.rst#data-analysis) has been updated for the "Data analysis" and "Results table" sections. 
The help buttons on the interface will still bring up the wiki instead of this new documentation - this is fixed in a separate PR (~~#17325~~ merged).

[Release notes](https://github.com/mantidproject/mantid/blob/b509a79efbab8a1a74b1872e4aa1baeee6c3faed/docs/source/release/v3.8.0/muon.rst#multi-dataset-fitting) have been updated.

**Future work not covered here:**
This PR does not deal with plotting (~~PR #17434~~ merged) or planned extensions to setting fit parameters from log values (#17184). Some known bugs with the `FunctionBrowser` remain, e.g. #16601. There is still a lot of refactoring and improvement that could be done - hopefully the new tests will make this easier.


raquelalvarezbanos:
I am getting an exception when trying to open the `DataAnalysis` tab with a dataset loaded into the interface (both on Linux and Windows). The problem seems to be in line L411 in `qtpropertybrowser.cpp`, in the method:

```
L409 void QtProperty::setEnabled(bool enable)
L410 {
L411     if (d_ptr->m_enabled == enable)
L412        return;
L413
L414    d_ptr->m_enabled = enable;
L415    propertyChanged();
L416}
```
However, switching to that tab without any data loaded is OK, and I can see the new display of the tab.

I've checked your calendar and seen that you are aware until Thursday, but I'll be out of office that day running on of the Mantid training courses, so won't be able to show you the issue. Therefore I think it's better if I leave this unassaigned so another developer can test it.

tom-perkins:
Thanks for testing - I think the error is related to the option recently added in #17264 for "evaluation type". This option was added to the base `FitPropertyBrowser` but not the derived `MuonFitPropertyBrowser`, I've pushed a fix for this and now I don't see the error.

KarlPalmen:
I'll test the Data Analysis tab.

KarlPalmen:
I can switch to data analysis tab with one data file loaded.

KarlPalmen:
After a chat with @raquelalvarezbanos , it became evident that there is plenty of more work to do in testing this and doing the code review.

KarlPalmen:
When fitting the Pawley function, I got an unexpected exception: ``Unknown function parameter f0.a``.

KarlPalmen:
This is not robust I got it to crash again by loading RAW files, which the file browser listed.

KarlPalmen:
Tom, for the code review, it would be helpful, if you gave a complete list of files you modified, then the code review could be done by examining these files, rather than over 62 commits.

KarlPalmen:
I attempted the sequential fit as in the script of the unscripted test (skipping test 1) and it crashes when I do the sequential fit.

tom-perkins:
Hi Karl, thanks for testing.

It shouldn't be possible to load RAW files into the interface (it's Nexus only) - when I try loading one on the front tab there is a message box `Unable to load the file[s]`. On the "Data Analysis" tab, the browse button is disabled. Which steps did you follow to see the crash?

Having looked at the documentation for the Pawley function, it seems it needs a special setup from the `PawleyFit` algorithm and can't be used like a normal fit function, which might explain that error. I think it is for powder diffraction, so wouldn't be used in this interface.

I don't see the crash on the sequential fit - which point do you see it at? Is it before the dialog comes up, or after?

I will put together a list of which files have been changed, and in which way.

tom-perkins:
Changes to files:
- `MuonAnalysis.h/.cpp`: Mainly code has been refactored out of here into other files. Now MuonAnalysis owns two new presenters:
  - `MuonAnalysisFitDataPresenter` (.h, .cpp, test) deals with the data to be fitted
  - `MuonAnalysisFitFunctionPresenter` (.h, .cpp, test) deals with the fit model to use
- `MuonAnalysisHelper.h/.cpp/test`: Some new helper functions added, plus tests
- `MuonSeqentialFitDialog.h/.cpp`: Few actual changes, it's just been moved from MantidWidgets -> CustomInterfaces. It doesn't load data itself any more, instead it makes use of the fit data presenter to do that.
- `MuonAnalysisDataLoader.h/.cpp/test`: Data loading code that's been refactored out of MuonAnalysis and had tests added
- `MuonAnalysisResultTableCreator.h/.cpp/test`: Code to generate results table that has been refactored out of MuonAnalysis and had tests added
- `MuonFitPropertyBrowser`: extended to deal with simultaneous fits
- `MuonFunctionBrowser`: A specialisation of `FunctionBrowser` - the only change is that a restricted set of functions are available (just Background, General and Muon categories)
- `MuonFitDataSelector`: A new widget that acts as a humble view, just GUI get/set code
- `*.rst/*.png`: Documentation and testing guide
- Abstract base classes have been used for mocking purposes in tests, these are `IFunctionBrowser.h`, `IMuonFitDataSelector.h`, `IMuonFitFunctionControl.h`, `IWorkspaceFitControl.h`.
- `MDFLogValueFinder.h/.cpp`: Refactored out of the multi-dataset fit interface, so that the "Edit Local Parameter" dialog can be used with a FunctionBrowser even when the parent isn't the MDF interface.

KarlPalmen:
Thank you Tom for the information. I'm now working at removing an imperfection from the build, in case this is the cause of the crashes, I've had.

KarlPalmen:
A look at muonAnalysis.cpp shows it now has just under 3000 lines of code. Most of its numerous functions are short. A few are long with three with over 100 lines.

KarlPalmen:
I found one more function of over 100 lines . Most are short and compact.

KarlPalmen:
This is far from robust. Correct usage needs to be explained very carefully. If usage deviates from correct a crash is likely.

tom-perkins:
Two problems here:
- the test instructions are misleading for "test 2" (I'll clarify them)
- the sequential fit dialog seems to be enabled when it shouldn't be (I will investigate)

tom-perkins:
- Testing guide has been updated
- The check for multiple runs has been fixed so it shouldn't be possible to launch the sequential fit dialog when more than one run is specified in the interface

KarlPalmen:
Fitting tests 2-7 (before reordering 7) work OK.

KarlPalmen:
I confirmed with Tom, that only the fitting tests are necessary :shipit: .

PR #16856, opened 2016-07-13
Rename workspaces check if out name exists

### LINKED ISSUE #16263: RenameWorkspaces should only overwrite if forced
### Expected behaviour

RenameWorkspace should fail if an existing workspace of the output name is defined.  This check needs to be for anything in the ADS and also members of groups in the ADS.

We should add a new overwite Boolean property (default false) that would delete the outputworkspace if it exists before renaming.

### Actual behaviour
Currently renameWorkspace can overwite workspaces if the outputworkspace name is already in use.
This can make a particular mess if you name two members within a workspace group the same (see #8652).

```python 
CreateSampleWorkspace(OutputWorkspace='bob', Function='User Defined', UserDefinedFunction='name=LinearBackground,A0=3.22542,A1=-0.0633193')
CreateSampleWorkspace(OutputWorkspace='bob1', Function='User Defined', UserDefinedFunction='name=LinearBackground,A0=3.22542,A1=-0.0633193')
CreateSampleWorkspace(OutputWorkspace='bob2', Function='User Defined', UserDefinedFunction='name=LinearBackground,A0=3.22542,A1=-0.0633193')
CreateSampleWorkspace(OutputWorkspace='bob3', Function='User Defined', UserDefinedFunction='name=LinearBackground,A0=3.22542,A1=-0.0633193')
GroupWorkspaces(InputWorkspaces='bob,bob1,bob2,bob3', OutputWorkspace='NewGroup')
RenameWorkspace(InputWorkspace='bob1', OutputWorkspace='bob')
```
This will give you a group with two members with the same name, there is lots of trouble from this point on.

### Platforms affected
All

Once this is solved we should check with Rob Dalgleish to make sure this does not mess up his scripts.

DavidFair:
Description of work.
Modified RenameWorkspace and RenameWorkspaces to check if any of the output names are used in other workspaces
**To test:**
Renaming a workspace to the same name as an existing workspace should throw an error
Renaming workspaces should throw an error if any of the output names match an existing name

Fixes #16263 .

Release notes have been updated

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/Algorithms/src/RenameWorkspaces.cpp
```diff
@@ -86,6 +96,18 @@ void RenameWorkspaces::exec() {
     }
   }
 
+  // Check all names are not used already before starting rename
+  for (size_t i = 0; i < nWs; ++i) {
```
FedeMPouzols: I think here you could use a range-based loop: `for (const auto &name : newWsName) { ...`

Framework/Algorithms/src/RenameWorkspace.cpp
```diff
@@ -44,13 +52,28 @@ void RenameWorkspace::exec() {
...
   // get the output workspace name
   std::string outputwsName = getPropertyValue("OutputWorkspace");
+  // check if we are overriding existing workspaces
+  bool overrideWorkspaces = getProperty("OverrideExisting");
 
   if (getPropertyValue("InputWorkspace") ==
       getPropertyValue("OutputWorkspace")) {
     throw std::invalid_argument(
         "The input and output workspace names must be different");
   }
```
FedeMPouzols: This check could be done at the `validateInputs()` step.

Framework/Algorithms/src/RenameWorkspace.cpp
```diff
@@ -44,13 +52,28 @@ void RenameWorkspace::exec() {
...
+    // Output name already exists - either remove or error
+    if (!overrideWorkspaces) {
+      // If we try to delete the workspace here a subtle bug is introduced
+      // Where the workspace group handle is deleted if we are renaming
+      // Its last workspace member, then when we add (or rename) that member
+      // undefined behavior happens usually Python Unit tests breaking
+      throw std::runtime_error("The workspace " + outputwsName +
+                               " already exists");
+    }
+  }
```
FedeMPouzols: And probably this too can go in `validateInputs()`?

FedeMPouzols:
@DavidFair: I've made some comments for possible improvements,. let me know what you think. But I think this looks good and ready to go. It doesn't break backwards compatibility, code looks good, and the documentation and examples are very good.

DavidFair:
I'll move the examples into validateInputs .
Previously the code also deleted the target workspace at execution time to make sure however we later discovered this caused a race condition where Poco would delete the workspace handle which we then returned so now we let the ADS handle it.

DavidFair:
@FedeMPouzols I've moved all tests (except one) into validate inputs and reworked the unit tests to use these now.
In RenameWorkspaces the check to see if names are in use has to be done at execution time as we need to join suffix + name + prefix then query the ADS for each one.

docs/source/release/v3.8.0/framework.rst
```diff
@@ -49,6 +49,10 @@ Improved
 - :ref:`SetSample <algm-SetSample>`: Fixed a bug with interpreting the `Center` attribute for cylinders/annuli 
 
 
+- :ref:`RenameWorkspace <algm-RenameWorkspace>` and `RenameWorkspaces <algm-RenameWorkspaces>`
+  now check if a Workspace with that name already exists in the ADS and gives
+  the option to override it.
```
FedeMPouzols: override => overwrite? If you modify this we don't need to wait for the builds again.

FedeMPouzols:
Looks good. It implements the plan in the issue description, code looks good, it is well covered with new/extended unit tests, and the documentation reads well.
Keeping a default of this new option as "yes, overwrite" makes perfect sense for backwards compatibility.

:shipit:

martyngigg:
@DavidFair This branch has developed a merge conflict

PR #16857, opened 2016-07-13
Store Workspace Dx data for each bin instead of for each bin edge

### LINKED ISSUE #16298: Store Workspace Dx data for each bin instead of for each bin edge

SimonHeybrock:
Title says it all, also see the issue description.

The only major and potentially breaking change is loading old Nexus files. I went with the assumption that delta-Q varies slowly and we can thus simply drop the last value (see also the additional unit test in `LoadNexusProcessed`). Please comment on whether or not that makes sense.

**To test:**

Apart from code review,
- Test loading an viewing workspaces with Dx data.
- Make sure that old processed Nexus files (with 1 Dx value too many) can be loaded correctly, and that a warning is printed.

Fixes #16298.

I added this to the SANS release notes, or should it be in a more general section?

@AntonPiccardoSelg You might be a good candidate for reviewing this. 

---
#### Reviewer

##### Code Review
##### Functional Tests


AntonPiccardoSelg:
@SimonHeybrock Could you merge master into this branch? Thank you.

SimonHeybrock:
@AntonPiccardoSelg Done (I did not build/test locally, I hope the merge went fine, we will see).

SimonHeybrock:
@AntonPiccardoSelg Have you had a look at this? If not, I don't mind pushing this to 3.9, it is nothing critical.

AntonPiccardoSelg:
There seems to be a difference in the DX values when I run the same sample script in for Mantid3.7.2 and this branch.

M3.7.2
old_version


This branch:
new_version

AntonPiccardoSelg:
The issue was with the normalization of the plot. Once the bin normalization is turned off, the displayed data is the same. Checking the actual values also reveals that this change has not altered the reduction result.

Code changes look good. Tested with the sample script above. All worked well and is correct. This new implementation of DX makes much more sense for how we store QResolution results in SANS. Happy to :shipit:

PR #16839, opened 2016-07-11
Remove final keyword from class_maker.py

quantumsteve:
Description of work.

This removes the final keyword from algorithms generated by `class_maker.py`. As discussed in #tech-steering, this is looking like a premature optimization and best reserved only for when deriving from a class or overriding a member function will cause bad things to happen.

I tried to clean up the damage and fixed all other warnings from clang-tidy's `modernize-use-override` check. 

**To test:**

There is no GitHub issue associated with this PR.

_Does not need to be in the release notes._

---
#### Reviewer

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):
##### Code Review
- [x] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)? Is it well structured with small focussed classes/methods/functions?
- [ ] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?
##### Functional Tests
- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant documentation been added/updated?
- [ ] Is user-facing documentation written in a user-friendly manner?
- [ ] Has developer documentation been updated if required?
- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

PR #16859, opened 2016-07-13
Create CropToComponent algorithm

### LINKED ISSUE #16101: Create CropToDetector algorithm
__Context__

For the current and future SANS reduction we need to crop a workspace to a specific detector. Currently this is set by specifying a detector range. Since we will be dealing in the future with detectors which don't have a continuous spectrum number range, we require an algorithm which extracts the correct spectra for a specified component

__Requirements__

* Should work for event-type and histogram-type detectors
* Should extract everything which is associated with a specified detector
* Should throw if a non-existant detector has been specified
* Should allow for cropping to multiple detectors (maybe 2nd iteration)
*Should allow for inplace manipulation (check if that is possible)

__Basic Implementation__

1. Check in the IDF of the workspace if the specified detector(s) are valid
1. Extract the spectrum numbers from the IDF (this is the unknown part)
1. Apply ExtractSpectra

__Tests__

* Unit tests should cover the above requirements
* A performance tests should be added

AntonPiccardoSelg:
## Description of work.

For the current and future SANS reduction we need to crop a workspace to a specific detector. Currently this is set by specifying a detector range. Since we will be dealing in the future with detectors which don't have a continuous spectrum number range, we require an algorithm which extracts the correct spectra for a specified component

Fixes #16101 
## To test:

Run the script from the doc tests. Play with the different paramete setting.
## Release Notes

See [here](https://github.com/mantidproject/mantid/pull/16859/commits/323949ff84042f2b0604a1cac0a2be76103b9c2f)

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/Algorithms/src/CropToComponent.cpp
```diff
@@ -0,0 +1,129 @@
...
+                        "InputWorkspace", "", Direction::Input),
+                    "An input workspace.");
+    declareProperty(Kernel::make_unique<WorkspaceProperty<Mantid::API::MatrixWorkspace>>(
+                        "OutputWorkspace", "", Direction::Output),
+                    "An output workspace.");
+    declareProperty(
+        Kernel::make_unique<Mantid::Kernel::ArrayProperty<std::string>>(
+            "ComponentNames"),
+        "List of component names to which the workspace is goign to be cropped "
+        "to.");
```
KarlPalmen: typo

Framework/Algorithms/src/CropToComponent.cpp
```diff
@@ -0,0 +1,129 @@
...
+void CropToComponent::exec()
+{
+    // Get the names of the components
+    std::vector<std::string> componentNames = getProperty("ComponentNames");
+    Mantid::API::MatrixWorkspace_sptr inputWorkspace
+        = getProperty("InputWorkspace");
+    Mantid::API::MatrixWorkspace_sptr outputWorkspace
+        = getProperty("OutputWorkspace");
+
+    // Check if the component names are empty
```
KarlPalmen: Should comment say, something like, // If component names are empty, make output workspace the same as input workspace. ?

AntonPiccardoSelg:
@KarlPalmen Your suggestion was appplied.

KarlPalmen:
I find the algorithm functions as suggested by its name rather than its description.
The description says that I need to supply a list of detectors, but it works with a list of components.
The description and release notes need to be modified to indicate that components may be listed (as suggested by algorithm name).

KarlPalmen:
I tested with the GUI and used show instrument to see that it cropped as expected.

KarlPalmen:
Looks good :shipit: .

Framework/Algorithms/inc/MantidAlgorithms/CropToComponent.h
```diff
@@ -0,0 +1,50 @@
...
+  int version() const override final;
+  const std::string category() const override final;
+  const std::string summary() const override final;
+
+protected:
+  std::map<std::string, std::string> validateInputs() override;
+
+private:
+  void init() override final;
+  void exec() override final;
```
FedeMPouzols: Note these `final` specifiers are not being added by the class maker any longer, only since very recently: #16839.
I don't think this needs fixing, it's just that I recalled this recent change.
@AntonPiccardoSelg: let us know if you want to change that, otherwise I or any gatekeeper will merge
this a bit later.

AntonPiccardoSelg: @FedeMPouzols  I think we can leave them as they are for now. Thanks for pointing it out.

Issue #16661, opened 2016-06-20
Validation of data in HistogramData module


SimonHeybrock:
The `Histogram` type will make certain promises about data validity. This requires validation of input data values.
Upon implementation, this will cause many failures in unit tests that are current using broken histogram data, so this also needs to be fixed as part of this issue.

PR #16861, opened 2016-07-13
Cleanup and preparations for better validation of data in histograms

SimonHeybrock:
This is a set of changes related to validation of data in histograms. For example, we might require that bin edges are always strictly increasing, such that bin widths are always positive.

The bulk of these changes come from an attempt to make all unit tests pass with a stricter validation for bin edges. These validation have been disabled in the last step, since further details need to be worked out, and especially since it is not clear if the performance impact would be acceptable.

Nevertheless, what remains are a few changes that are worthwhile to include into master now:
- `WorkspaceCreationHelpers` now create sensible X data. That is, our unit tests no longer use degenerate bin edges that all have the same value. This should make this more useful for unit tests, and allow for more realistic and safer test.
- Quite a few cases where "bad" bin edges have been used were cleanup up. These are cases where the histogram types would throw exceptions if validation were enabled.
- Routines for manual validation are available in `HistogramData`.
- `LinearGenerator` and `LogarithmicGenerator` together with new constructors for objects in `HistogramData` make creating sensible data easier, especially for testing. For example
  
  ``` cpp
  BinEdges edges(numBins + 1, LinearGenerator(start, step));
  ```

**To test:**

There is no corresponding issue. This is a first step for #16661, but it is incomplete.

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/Algorithms/test/GetEiV1Test.h
```diff
@@ -170,17 +172,15 @@ class GetEiV1Test : public CxxTest::TestSuite {
...
     const double peakOneHeight(3000.), peakTwoHeight(1000.);
-    for (int i = 0; i <= numBins; ++i) {
-      const double xValue = 5.0 + 5.5 * i;
-      if (includePeaks && i < numBins) {
+    for (int i = 0; i < numBins; ++i) {
+      if (includePeaks) {
         testWS->dataY(0)[i] =
             peakOneHeight *
-            exp(-0.5 * pow(xValue - peakOneCentre, 2.) / sigmaSqOne);
+            exp(-0.5 * pow(xdata[i] - peakOneCentre, 2.) / sigmaSqOne);
```
quantumsteve: Should this be `pow(xdata[i] - peakOneCentre, 2)` instead of `pow(xdata[i] - peakOneCentre, 2.)`? Same change on line 182?

They appears to be two [different specializations](http://en.cppreference.com/w/cpp/numeric/math/pow).

SimonHeybrock: @quantumsteve The `int` overloads seem to be removed with c++11. The new variant (7) claims it casts `int` to `double`. If you think there might be a worthwhile speedup, we can rewrite the `pow` as a product.

quantumsteve: Nevermind, making the exponent integral looks like outdated advice. The assembly generated for  `x*x` and std::pow(x,2.) is identical.

https://godbolt.org/g/CuL895

abuts:
There is number of useful changes here (especially defined binning) but some are questionable. 

E.g. : 
Suggested changes to move constructor --
at:
https://github.com/mantidproject/mantid/compare/16661_histogram_validation#diff-67d04efe0816aee0906060fb6f039f24L52:
 
```
-  FixedLengthVector(std::vector<double> &&other) : m_data(std::move(other)) {}
+  FixedLengthVector(std::vector<double> &&other) { m_data = std::move(other); }
```
where the first one is the standard possible move constructor implementation, I do not understand the benefits of the new one. What does it mean?

In addition, http://en.cppreference.com/w/cpp/utility/move states that
```C++
std::vector<int> v = {2, 3, 3};
v = std::move(v); // undefined behavior
```
Why the new form is any way better then the old one?

Another suggestion:
https://github.com/mantidproject/mantid/compare/16661_histogram_validation#diff-67d04efe0816aee0906060fb6f039f24R65:
```
+  FixedLengthVector(size_t count, Generator g)
+      : m_data(count) {
+    std::generate(m_data.begin(), m_data.end(), g);
+  }
```
Should Generator be transferred by (const) reference? std::generate will copy it anyway, I do not see any need to copy it twice (may be modern compilers are smart enough to do it themselves -- if this is the case then fine but -- just not sure its ok. 

Changes are big and I am not sure I've looked through everything carefully enough, but otherwise all looks fine and mainly useful.

SimonHeybrock:
@abuts Regarding the `FixedLengthVector` move: The change is just a left over after deactivating the data validity check (which had to come before changing `m_data`, i.e., I could not use the initializer list). I will revert that.

Good point about `std::generate`, I will change that!

abuts:
Should not copy & move constructors check against self-assignment? (i.e. default copy-move constructors are not acceptable for such classes?)

SimonHeybrock:
@abuts Yes, since these classes are just wrappers around `std::vector` and `Kernel::cow_ptr` I am relying on the self-assignment checks in there. There are also some unit tests that check self assignment.

abuts:
Unit test are of course ultimate checks. As I have not seen them in this ticket, its difficult to judge, but I would expect them to be there.

SimonHeybrock:
@abuts 
1. I assume you mean assignment operators, not constructors.
2. They are not defaulted in this case?
3. Can you give an example what could go wrong with the default? I cannot think of a case where the behavior would be worse than those of any of the class members.
4. My implementations are bases on the assignment operators of `std::vector`. As you linked above, self move-assignment is not supported. I am not sure why we should provide more functionality than the standard library in this case. What is the motivation?

abuts:
1) yes, its abut assignment, sorry not being clear about it
2) I mean construct like the one mentioned above:
```
FixedLengthVector(10,1) v;
v = FixedLengthVector(v);
```
3) It was a while when I Last time looking at good practice in writing all this stuff, so I have some reminiscent memory, not the fresh knowledge. As this is part of the bigger work, I would accept this as it is. 
4) I think the whole solution would need to test operation like 2)

abuts:
as the tests are passing, there are number of trivial changes and non-trivial one are the part of further work, I would suggest to :shipit:

SimonHeybrock:
@abuts Regarding your code example, there is no self-assignment, I think. You are creating a temporary with the copy constructor, and then you have a copy assignment?

abuts:
to be fair, I have no idea what is happening in this situation. I know this may often occur unintentionally, and extremely difficult to debug when you do operations with custom classes like the one you are building now. Probably easiest solution around would be to prohibit default assignment until somebody  needs it, or look through a text which says what should be implemented in this case. For myself, I am just remembering that its not entirely trivial, so in my code I am trying to avoid such stuff. 

You probably cannot if this job has been decided upon.

SimonHeybrock:
Well, we are using `std::vector` everywhere and I am not aware of any such problems. This class behaves basically identically, so I don't see this as an issue.

PR #16870, opened 2016-07-13
Handle spectrum properties when loading processed nexus

### LINKED ISSUE #15958: Enable LoadNexus to respond to Spectrum Properties when LoadNexusPrecessed is to be run.
At present, LoadNexus responds to the properties that define which spectra are to be loaded only if the algorithm that does the loading is not LoadNexusProcessed.

In the case of LoadNexusProcessed, the code for passing on spectrum properties is commented out. So the task is to uncomment out this code and make sure it works. 

If there is no need to modify a unit test, a new unit test may be needed.

I think the code was commented out, because at the time of writing, LoadNexusProcessed has no spectrum properties.

DavidFair:
Description of work.
This PR removes commented out code out of `LoadNexus`. Previously `LoadNexusProcessed` could not handle a spectrumList and spectrumMin/Max so a check was in place to prevent that. 
However since then the ability to handle these properties has been added so the check needs to be removed. Additionally some additional simple validation was added. 

**To test:**
Open a processed nexus file. A test file called 'GEM3870_focussed_legacy' within the unit test files can be used. Try to enter some invalid input such as reversing Min/Max or setting them to negative numbers - it should not allow this.
Fixes #15958 .

Release notes do not need to be updated

---
#### Reviewer

##### Code Review
##### Functional Tests


FedeMPouzols:
This looks very good and well tested. Now there's a caveat I think. @DavidFair: the fact that the chunk of commented out code was in `LoadNexus.cpp` might have been misleading. I'd say that the best place for the input checks that you've added is  `LoadNexusProcess.cpp`. See for example what happens in `LoadNexus::runLoadMuonNexus()`: the just pass the properties to the algorithm `LoadMuonNexus`. If you look into `LoadMuonNexus2.cpp` you'll see that they do `checkOptionalProperties()` which is similar to the checks that you do here (note that that method is defined elsewhere). 

The downside of checking the spectrum index properties in `LoadNexus` is that if users or scripts call `LoadNexusProcess` directly that would bypass your checks. This happens in quite a few algorithms, scripts and tests.

`LoadNexus` is a special algorithm in the sense that it can call a few alternative `LoadXXXNexus`. On top of it `Load` is even more special, as it can call may different `LoadXXX` algorithms.

I can come by later and explain if this looks too confusing.

FedeMPouzols:
I am also wondering if those checks could be done in `validateInputs()` (see [this wiki page on writing algorithms](http://www.mantidproject.org/Writing_an_Algorithm), last point). I think that's the ideal place. `validateInputs()` checks are used to warn the user at an early stage (it for example produces "red blob" warnings in the algorithm dialog). 

`validateInputs()` is not used as much as it should be, but you can find many examples under Framework/Algorithms.

DavidFair:
@FedeMPouzols So just to ensure I understand to make sure people don't work around this check by calling `LoadNexusProcessed` directly move this check into that algorithm and use `validateInputs` method on it?
Would validate inputs be able to check that certain inputs are mutually exclusive?

FedeMPouzols:
Yes, you have full freedom in `validateInputs()`, and you can report multiple issues by adding several messages into the output map of errors. See for example the `validateInputs()` methods of `ImggTomographicReconstruction`, `WorkspaceJoiners`  or `CorelliCrossCorrelate`

FedeMPouzols:
By the way @DavidFair the documentation string that is used in the `declareProperty(...)`s in the algorithm `init()` is grabbed and added automatically to the table of properties of the algorithm documentation pages: http://docs.mantidproject.org/nightly/algorithms/LoadNexusProcessed-v1.html

Here you could add a note saying that these options are mutually exclusive, either in the body of the algorithm documentation (`LoadNexusProcessed.rst`) or in the table. Probably the body is a better place, where they say "A specific list of spectra to load can also be given...".

DavidFair:
@FedeMPouzols I've moved the method into ValidateInputs of `LoadNexusProcessed` and tested it. 
I also had to remove some unit tests which were checking combinations of Min/Max with a list however only on a contiguous range. This were not throwing the exceptions which you get when you use Min/Max and List with a non-contiguous range.

Would you be able to check this over again?

FedeMPouzols:
@DavidFair: your changes look good. But I'm not sure if forbidding min/max + list combinations is consistent with how other algorithms work. If you look at LoadNexusProcess itself (in `LoadNexusProcessed::loadNonEventEntry()`) you'll see that it can happily process for example Min=1, Max=3, List=7,8,9. And I think that is a relatively common practice. Other Load***Nexus algorithms also behave like that, see for example `LoadMuonNexus2`.

You say that if you try to load a processed nexus file with both SpectrumMin/SpectrumMax and SpectrumList specified it fails. How does it fail?

It look to me as if it was enough to check that the min/max/list are consistent with the maximum number of spectra (and between the min and max). For example in [`LoadMuonNexus.cpp: LoadMuonNexus::checkOptionalProperties()` ](https://github.com/mantidproject/mantid/blob/master/Framework/DataHandling/src/LoadMuonNexus.cpp#L109-L139) or in [`LoadNexusProcessed` itself](https://github.com/mantidproject/mantid/blob/master/Framework/DataHandling/src/LoadMuonNexus.cpp#L109-L139), in the method `checkOptionalProperties()`. Probably that method can be removed, if all the checks done there can be moved to `validateInputs()`.

Does this make sense?

DavidFair:
@FedeMPouzols I've just tested against 3.7.1 loading with Min=1 Max=3 List=7,8,9 and it throws the following error loading `GEM38370_Focussed_Legacy.nxs` :

```
Invalid list of spectra
Error in execution of algorithm LoadNexusProcessed:
Inconsistent properties defined
Error in execution of algorithm Load:
Inconsistent properties defined
```

When I initially went through the ticket with Karl and looked at the commented out code it looked like it was there to check for and prohibit SpectrumList and SpectrumMin/Max. This is why we went ahead and assumed that was correct behaviour and put the check back in.
Are we potentially covering a bug deeper in the execution or should it not be able to mix SpectrumList with SpectrumMin/Max?
If this is intended behaviour I'm quite happy to merge the checks into a single method.

FedeMPouzols:
Aha I see `GEM38370_Focussed_Legacy.nxs` has 6 spectra (numbered from 2 to 7). The error message is confusing, I think it should say that the list given has some numbers higher than the maximum spectrum number found in the file (7) (or similar).

Try for example `LoadNexusProcessed` with that same file, but giving:
- SpectrumMin=2
- SpectrumMax=3
- SpectrumList=5,6
That works well for me. And parameters like that work well in general for different Load**Nexus algorithms. This is as far as I know, the way it works, I cannot confirm if that is always the intended/best/most consistent behavior. That goes beyond the scope of this issue. These checks on the min/max/list options of the loaders should be better harmonized (and the code shared), ideally.

DavidFair:
Ahhh me miscounting in HDF view is the source of all of this confusion, not a mistake I'm going to be making again soon. 

So going back to the original issue if 3.7.1 is working as intended is it worth reverting the changes and just remove the commented out check in `LoadNexus` as per the original issue?

FedeMPouzols:
Always trust Mantid better than HDFView ;)
I think that the checks that you added in `validateInputs()` (relaxed to accept min/max + list if they're consistent) and the unit tests are worth.

DavidFair:
@FedeMPouzols I've completed the changes we discussed and I'll make a start on refactoring the unit tests and removing the int64_t at a later date. 
Would you be able to look this code over the make sure I haven't missed anything obvious?

FedeMPouzols:
@DavidFair : do you need to update the description of what you did in this PR?

FedeMPouzols:
Looks good. This tidies up a bit `LoadNexusProcess`: removes a very confusing block of dead/commented out code, and adds `validateInputs()`.

To be picky, I'll wait until the PR description is updated, but this is ready for a "squirrel" stamp on it.

DavidFair:
@FedeMPouzols I was so busy looking through the code for something I missed I completely forgot about the PR descrption. I've since updated it to reflect the changes performed and why.

FedeMPouzols:
Let's :shipit:

PR #16881, opened 2016-07-14
ConvertUnits could corrupt inplace workspaces on error

### LINKED ISSUE #16794: ConvertUnits Sets EMode on failed run
### Expected behavior
ConvertUnits Should work fail each time if the emode is not specified for a conversion that needs it.

### Actual behavior
ConvertUnits fails on first trial, but sets deltaE-mode when working on in in place change, so it  Succeeds on Second trial. 

Terminal log Containing Behaviour 
File is 1642000.nxs is part of sample training data
[164200.zip](https://github.com/mantidproject/mantid/files/347442/164200.zip)
```
>>>ws3 = Load('164200.nxs')

>>>ws3 = ConvertUnits(ws3,Target = 'DeltaE')
RuntimeError                              Traceback (most recent call last)
<ipython-input-82-0c2849963340> in <module>()
----> 1 ws3 = ConvertUnits(ws3,Target = 'DeltaE')

C:\MantidInstall\bin\mantid\simpleapi.pyc in algorithm_wrapper(InputWorkspace,Target,EMode=None,EFixed=None,AlignBins=None, Version=1)
    879 
    880         try:
--> 881             algm.execute()
    882         except RuntimeError, e:
    883             if e.args[0] == 'Some invalid Properties found':

RuntimeError: ConvertUnits: error (see log)

>>>ws3 = ConvertUnits(ws3,Target = 'DeltaE')
>>>
```

### Platforms affected
Tested MantidPlot 3.7.1 on Windows 7

NickDraper:
Several validation checks were taking place after the output workspace had already been altered.  for In place workspaces this meant the workspace was altered even though the algorithm failed.

The changes delay the creation of the output workspace and changes until all of the validation has passed.

**To test:**
1. All unit tests and system tests must pass
2. Create or load a workspace
3. Try to convert to DeltaE, or SpinEchoLength with an Elastic Emode and not setting eFixed.
4. Check that the input workspace has not changed (check the units of the ws).

Fixes #16794
### RELEASE NOTES

Either edit the file in docs/source/release/ui.rst

---
#### Reviewer

##### Code Review
##### Functional Tests


SimonHeybrock:
@NickDraper Looking at this right now. Could you check if `ConvertUnitsUsingDetectorTable` suffers from the same issue? The relevant part in `exec` looks basically identical (unfortunately there is a lot of copy and paste between those algorithms), so I think it might be.

Framework/Algorithms/src/ConvertUnits.cpp
```diff
@@ -322,32 +315,93 @@ void ConvertUnits::convertQuickly(API::MatrixWorkspace_sptr outputWS,
...
+* @param l2 :: The returned sample - detector distance
+* @param twoTheta :: the returned two theta angle
+* @returns true if lookup successful, false on error
+*/
+bool ConvertUnits::getDetectorValues(
+    Kernel::Unit_const_sptr outputUnit, Geometry::IComponent_const_sptr source,
+    Geometry::IComponent_const_sptr sample, double l1, int emode,
+    MatrixWorkspace_const_sptr ws,
+    function<double(const Geometry::IDetector &)> thetaFunction,
+    int64_t wsIndex, double &efixed, double &l2, double &twoTheta) {
```
SimonHeybrock: I like that this was extracted into a function. Most of the arguments should not be `shared_ptr` however, since the function is of course not taking ownership. So `source`, `sample`, and `ws` should be passed by `const &` instead.
Since `getDetectorValues` is called once for each spectrum this may actually also make a (very small) performance difference.

Framework/Algorithms/src/ConvertUnits.cpp
```diff
@@ -402,69 +456,63 @@ void ConvertUnits::convertViaTOF(Kernel::Unit_const_sptr fromUnit,
...
+  size_t checkIndex = 0;
+  if (getDetectorValues(outputUnit, source, sample, l1, emode, inputWS,
+                        thetaFunction, checkIndex, checkefixed, checkl2,
+                        checktwoTheta)) {
+    const double checkdelta = 0.0;
+    // copy the X values for the check
+    auto checkXValues = inputWS->readX(checkIndex);
+    // Convert the input unit to time-of-flight
+    Unit *checkFromUnit = fromUnit->clone();
+    Unit *checkOutputUnit = outputUnit->clone();
```
SimonHeybrock: Could be simply dropped into a `std::unique_ptr`, then we can avoid the manual `deletes` below.

Framework/Algorithms/src/ConvertUnits.cpp
```diff
@@ -402,69 +456,63 @@ void ConvertUnits::convertViaTOF(Kernel::Unit_const_sptr fromUnit,
...
-        }
-      }
+    // Now get the detector object for this histogram
+    double l2;
+    double twoTheta;
+    if (getDetectorValues(outputUnit, source, sample, l1, emode, outputWS,
+                          thetaFunction, i, efixed, l2, twoTheta)) {
 
       // Make local copies of the units. This allows running the loop in
       // parallel
```
SimonHeybrock: Also the `Unit *` below could be replaced by `std::unique_ptr<Unit>`.

SimonHeybrock:
@NickDraper Looks all good apart from those minor comments.

I also reverted the changes in `ConvertUnits`. The new unit test catches the regression. 👍

NickDraper:
Yes, ConvertUnitsUsingDetectorTable   has the same problems - good spot there. I've created a separate issue for that #16935

SimonHeybrock:
@NickDraper Great, latest changes look good. If Jenkins gives his ok this is ready to merge. :shipit:

PR #16884, opened 2016-07-14
Fix LoadRKH for data with second header and 2D data

### LINKED ISSUE #16879: LoadRKH crashes when loading valid RKH file
Stephen King noticed that LoadRKH is not working properly in v3.7.1. It appears to be fine in v.3.6

### Expected behavior
`LoadRKH` should be able to load RKH files.

### Actual behavior

`LoadRKH` crashes when trying to load RKH files.

Load and RKH file.

### Platforms affected
All

AntonPiccardoSelg:
Fixes #16879 

This fixes a bug which is caused by the way `LoadRKH` switches between 1D and 2D loading. Up until now the switch looked if there is an entry in the second header line. This is bad. There is no reason to not have a multiple entries in there. 

Yet the second line (and potentially some subsequent lines) is the only way to figure out if the data is 1D or 2D. The 2D line will typically look something like
6 q (Angstrom^-1)

We implement a check to try and pattern-match the second line of an RKH file to the the above format. This is far from perfect, but currently the only option we have and considerably more robust than the status quo.

**To test:**

Please find sample files here: \olympic\Babylon5\Scratch\Anton\Testing\Fix_RKH
Make sure that the path is in your Mantid search directories.

**Test  RKH file which was crashing Mantid**
- Run 

``` python
ws = LoadRKH("M3_42_3rd-MT_rear.txt")
```
- Confirm that the file loads.
- Confirm that it is 1D by using Show Data

**Test with 2D RKH file**
- Run 

``` python
ws = LoadRKH("34481rear_2D_1.75_16.5.txt")
```
- Confirm that the file loads.
- Confirm that it is 2D by using Show Data

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/DataHandling/src/LoadRKH.cpp
+++ b/Framework/DataHandling/src/LoadRKH.cpp
@@ -20,6 +20,26 @@
 #include <MantidKernel/StringTokenizer.h>
 
 #include <istream>
+#include <numeric>
+#include <boost/regex.hpp>
+
+
+namespace {
+  // Check if we are dealing with a unit line
+  bool isUnit(const Mantid::Kernel::StringTokenizer& codes) {
+    // The unit line needs to have the format of
+    //  1. Either 0 or 6 [06]
+    //  2. Then several characters [\w]
+    //  3. Open bracket
+    //  4. Several characters
+    //  5. Close bracket  
+    std::string input = std::accumulate(codes.begin(), codes.end(), std::string(""));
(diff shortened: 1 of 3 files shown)


KarlPalmen:
Works OK for me. I later tried it with 'Load' , but that worked only with the 1D file, it selected the wrong algorithm for the 2D file. I take this to be a different issue.  :shipit:

PR #16891, opened 2016-07-15
Fix deselecting a first subcategory in Pref->Mantid->Options unticks main category 

### LINKED ISSUE #16871: Switching default instruments changes displayed algorithm categories
Reported by Spencer Howells at ISIS

### Expected behavior

Changing the default instrument in the MantidPlot "View->Preferences" menu should not affect the algorithm categories that are displayed in the Algorithms pane.

### Actual behavior

If a user has deselected some sub-categories in "View->Preferences->Mantid->Options" and changes the default instrument the parent category disappears from the algorithm view.

1. Clear (or comment out) any `algorithms.categories.hidden` in the `Mantid.user.properties` files
1. Go to the `View->Preferences` dialog and select the `Mantid->Options` tab
1. Expand the `Workflow` category and deselect all but `MIDAS` and click OK. Confirm that `MIDAS` is the only category visible in the `Algorithms` pane
1. Go back to the `View->Preferences` dialog and on the `Mantid->Instrument` tab select a different instrument and click OK
1. Observe that the `Workflow` category is now missing from the `Algorithms pane`  

### Platforms affected

All

DavidFair:
Description of work.
Certain categories under "View->Preferences->Mantid->Options" would take the value of their first subcategory depending on if they themselves were a category. 

When we get the list of categories and subcategories we now recurse backwards checking for any parent categories which were first seen from a subcategory and default them to a tick on as they do not have a visibility property associated.

**To test:**
Open "View->Preferences->Mantid->Options" all categories and subcategories should render correctly.
Categories such as "Algorithm" will preserve the tick option as it is standalone. The "Crystal" category is derived from its sub categories so will always appear ticked. 
Ensure the users selection is preserved across applying and re-opening the settings window.

Fixes #16871 .

Release notes have been updated

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidPlot/src/ConfigDialog.cpp
+++ b/MantidPlot/src/ConfigDialog.cpp
@@ -983,9 +983,9 @@ void ConfigDialog::initMantidOptionsTab() {
   treeCategories->setColumnCount(1);
   treeCategories->setSortingEnabled(false);
   treeCategories->setHeaderLabel("Show Algorithm Categories");
-  //Connect this item's signal to check box SLOT
-  connect(treeCategories, SIGNAL(itemClicked(QTreeWidgetItem*, int)),
-	 this, SLOT(tickBoxClickedSlot(QTreeWidgetItem*, int)));
+  // Connect this item's signal to check box SLOT
+  connect(treeCategories, SIGNAL(itemClicked(QTreeWidgetItem *, int)), this,
+          SLOT(tickBoxClickedSlot(QTreeWidgetItem *, int)));
   grid->addWidget(treeCategories, 2, 0);
   refreshTreeCategories();
 
@@ -1187,21 +1187,22 @@ void ConfigDialog::populateProgramTree() {
 
 void ConfigDialog::updateProgramTree() {
   // Store into a map ready to go into config service when apply is clicked
-  std::map<std::string, std::map<std::string, std::string>>::const_iterator
-      itr = m_sendToSettings.begin();
-  for (; itr != m_sendToSettings.end(); ++itr) {
+  for (const auto itr : m_sendToSettings) {
     // creating the map of kvps needs to happen first as createing the item
     // requires them.
-    std::map<std::string, std::string> programKeysAndDetails = itr->second;
+    std::map<std::string, std::string> programKeysAndDetails = itr.second;
 
     Qt::CheckState checkStatus = Qt::Unchecked;
-    if (programKeysAndDetails.find("visible")->second == "Yes") {
+    auto checkStatusIter = programKeysAndDetails.find("visible");
+    // Check that found the iterator before dereferencing it
+    if (checkStatusIter != programKeysAndDetails.end() &&
+        checkStatusIter->second == "Yes") {
       checkStatus = Qt::Checked;
     }
 
     // Populate list
     QTreeWidgetItem *program =
-        createCheckedTreeItem(QString::fromStdString(itr->first), checkStatus);
+        createCheckedTreeItem(QString::fromStdString(itr.first), checkStatus);
     treePrograms->addTopLevelItem(program);
     updateChildren(programKeysAndDetails, program);
   }
@@ -1212,14 +1213,12 @@ void ConfigDialog::updateChildren(
     QTreeWidgetItem *program) {
   program->takeChildren();
   // get the current program's (itr) keys and values (pItr)
-  std::map<std::string, std::string>::const_iterator pItr =
-      programKeysAndDetails.begin();
-  for (; pItr != programKeysAndDetails.end(); ++pItr) {
+  for (const auto pItr : programKeysAndDetails) {
     QTreeWidgetItem *item = new QTreeWidgetItem(program);
     QString itemText = QString("   ")
-                           .append(QString::fromStdString(pItr->first))
+                           .append(QString::fromStdString(pItr.first))
                            .append(" --- ")
-                           .append(QString::fromStdString(pItr->second));
+                           .append(QString::fromStdString(pItr.second));
     item->setText(0, itemText);
     program->addChild(item);
   }
@@ -1230,31 +1229,26 @@ void ConfigDialog::updateSendToTab() {
       Mantid::Kernel::ConfigService::Instance();
 
   // Add new values to the config service
-  std::map<std::string, std::map<std::string, std::string>>::const_iterator
-      itr = m_sendToSettings.begin();
   std::vector<std::string> programNames =
       cfgSvc.getKeys("workspace.sendto.name");
 
-  for (; itr != m_sendToSettings.end(); ++itr) {
+  for (const auto itr : m_sendToSettings) {
     for (size_t i = 0; i < programNames.size(); ++i) {
-      if (programNames[i] == itr->first) {
+      if (programNames[i] == itr.first) {
         // The selected program hasn't been deleted so set to blank string (all
         // those left without blank strings are to be deleted
         programNames[i] = "";
       }
     }
 
-    cfgSvc.setString("workspace.sendto.name." + itr->first, "0");
+    cfgSvc.setString("workspace.sendto.name." + itr.first, "0");
 
-    std::map<std::string, std::string> programKeysAndDetails = itr->second;
+    std::map<std::string, std::string> programKeysAndDetails = itr.second;
 
-    std::map<std::string, std::string>::const_iterator pItr =
-        programKeysAndDetails.begin();
-
-    for (; pItr != programKeysAndDetails.end(); ++pItr) {
-      if (pItr->second != "")
-        cfgSvc.setString("workspace.sendto." + itr->first + "." + pItr->first,
-                         pItr->second);
+    for (const auto pItr : programKeysAndDetails) {
+      if (pItr.second != "")
+        cfgSvc.setString("workspace.sendto." + itr.first + "." + pItr.first,
+                         pItr.second);
     }
   }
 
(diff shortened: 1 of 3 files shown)


mantid-roman:
The PR doesn't seem to fix issue #16871. Is it possible that it is fixing a different issue?

DavidFair:
@mantid-roman On my end it seems to be resolving the issue - its a bit confusing as the cause of the bug isn't related to default instruments unlike what the ticket says.

Previously if you deselected the first subcategory for Crystal or Workflow for example it would untick those categories when you next loaded the options window. 

The steps to reproduce are:
1. Currently if you untick Alignment under Workflow. 
2. Apply or OK
3. Open options window again and notice Workflow is now unticked regardless of its subcategories.
4. If the user hits apply the category Workflow now is hidden as it was unticked on load.

Hopefully with this PR it preserves that tick resolving the bug however if you can find a test where it doesn't I would be glad to look further into it.

mantid-roman:
The expected behaviour is that changing the default instrument doesn't change the displayed categories. And this is what I am testing.

1. Open Options.
2. Untick all subcategories of Workflow except MIDAS.
3. Click OK.
4. The algorithm pane shows only MIDAS under Workflow.
4. Change the default instrument.
5. Open Options again. Now both MIDAS and MLZ are visible.
6. The pane also shows them both now.

DavidFair:
Ahhh I was looking for algorithms becoming hidden not visible. I'll look further into this.

DavidFair:
@mantid-roman I've gone back and tested it and can confirm this happens. Unfortunately due to the nature of the bug this isn't easily resolved. 

MLZ is found from its subcategories loaded in. So the GUI loading code first sees `MLZ//DNS` instead of `MLZ` then `MLZ//DNS`.  Because of this `MLZ` doesn't have its own attributes for if it should be checked so we need to default to something which this code does. 

I originally consulted @martyngigg for what the tick should default to and we decided that if a subcategory exists we make the parent category ticked. This is the behaviour we are seeing with `MLZ` as there are subcategories associated with it however I am happy to change the behaviour if there are better approaches?

mantid-roman:
@martyngigg @DavidFair When you untick a root category (Workflow for example) it disappears from the Algorithms pane. But when you open the options you see that Workflow is ticked back but still invisible in the pane. If you hit Cancel Workflow category is ticked but invisible which is confusing. If you hit OK/Apply instead the category will appear without your explicitly ticking it which is unexpected.

martyngigg:
@DavidFair We discussed seeing if Qt support a "partial" tick status, i.e grayed out to suggest that not all is ticked but some. Did you look at that?

DavidFair:
I've implemented partial ticks within the last commit which I completely forgot about. Due to the many edge cases involved (explained below) with these categories the best solution was to walk the widget tree after it is generated setting the parent categories based on their children's status. 
This should make the behaviour a bit more consistent however AFAIK there isn't any methods called when a user ticks or unticks a box so the partial ticks can only be generated when the options window is refreshed. If there is one I am missing its just as simple as adding a few lines of code to it so I am still looking for it. 

@mantid-roman Unfortunately Workflow does not have its own category, instead it is derived from its child categories which are returned from the `AlgorithmFactory`. Because of this Workflow has to figure out from these sub-categories whether to be ticked, unticked or partially ticked. 
If any subcategories of workflow are on then workflow will take the checked or partial option on loading the options windows which makes it reappear when hitting apply as you describe. 
Without Workflow being returned from the `AlgorithmFactory` as a stand-alone category with a state of on or off I can't save the tick status hence the workarounds in these commits in the mean time.

mantid-roman:
In this case the check box of Workflow (and similar entries) should be disabled because editing them manually only adds confusion.

martyngigg:
@DavidFair What about signals such as http://doc.qt.io/qt-4.8/qtreewidget.html#itemClicked, are they not of any use?

MantidPlot/src/ConfigDialog.cpp
```diff
@@ -1192,10 +1192,14 @@ void ConfigDialog::updateProgramTree() {
     // requires them.
     std::map<std::string, std::string> programKeysAndDetails = itr->second;
 
+    Qt::CheckState checkStatus = Qt::Unchecked;
+    if (programKeysAndDetails.find("visible")->second == "Yes") {
```
mantid-roman: Can `find` return an invalid iterator?

MantidPlot/src/ConfigDialog.cpp
```diff
@@ -1267,59 +1271,176 @@ void ConfigDialog::updateSendToTab() {
...
+}
+
+void ConfigDialog::buildTreeCategoryStructure(
+    const QString *catNames, const bool isHidden,
+    QMap<QString, QTreeWidgetItem *> *seenCategories) {
+  /* Some categories come as standalone and some are discovered
+  // via their subcategory. E.g. Arithmetic vs Crystal\\Cell
+  // If standalone it has its own isHidden attribute otherwise
+  // we have to assume it to be visible (or ticked) as that information
+  // is not stored. */
```
mantid-roman: Use only `//` for comments

MantidPlot/src/ConfigDialog.cpp
```diff
@@ -1267,59 +1271,176 @@ void ConfigDialog::updateSendToTab() {
...
+  for (auto i = categoryMap.cbegin(); i != categoryMap.cend(); ++i) {
+
+    QString catNames = QString::fromStdString(i->first);
+    // Start recursion down building tree from names
+    buildTreeCategoryStructure(&catNames, i->second, &categories);
+  }
+}
+
+void ConfigDialog::buildTreeCategoryStructure(
+    const QString *catNames, const bool isHidden,
```
mantid-roman: Can `catNames` be a `nullptr`? Use a reference if not.

MantidPlot/src/ConfigDialog.cpp
```diff
@@ -1267,59 +1271,176 @@ void ConfigDialog::updateSendToTab() {
...
+    QMap<QString, QTreeWidgetItem *> *seenCategories) {
```
mantid-roman: `seenCategories` should probably be a reference too.

MantidPlot/src/ConfigDialog.cpp
```diff
@@ -1267,59 +1271,176 @@ void ConfigDialog::updateSendToTab() {
...
+
+  // Split the categories whose name is delimited by '\\'
+  QStringList splitCats = catNames->split('\\');
+
+  // Create a unique key for storing into the mapping
+  QString catKey(*catNames);
+  // We need to remove delimiter as top level categories do not come with it
+  catKey.remove('\\');
+
+  int splitCatIndex = splitCats.size();
```
mantid-roman: Can `splitCatIndex` be 0?

MantidPlot/src/ConfigDialog.cpp
```diff
@@ -1267,59 +1271,176 @@ void ConfigDialog::updateSendToTab() {
...
+  if (splitCatIndex == 0) {
+    // This is a top level category and standalone
+    QTreeWidgetItem *catWidget =
+        createCheckedTreeItem(splitCats[0], tickStatus);
+    treeCategories->addTopLevelItem(catWidget);
+    seenCategories->insert(catKey, catWidget);
+  } else {
+    // We need to determine if we have seen this category before
+    // and get any pointers to its parent(s) and set their tick status
+    int prevElement = splitCatIndex - 1;
```
mantid-roman: Let's hope that `splitCatIndex` isn't 0.

MantidPlot/src/ConfigDialog.cpp
```diff
@@ -1267,59 +1271,176 @@ void ConfigDialog::updateSendToTab() {
...
+  // We have built tree now fix any ticks or make partial ticks
+  int topLevelCats = treeCategories->topLevelItemCount();
+  for (int i = 0; i < topLevelCats; i++) {
+    correctTreePatrialTicks(treeCategories->topLevelItem(i));
+  }
+}
+
+QTreeWidgetItem *ConfigDialog::walkBackwardsThroughCategories(
+    const QString *catNames, int elementToCheck,
+    QMap<QString, QTreeWidgetItem *> *seenCategories,
```
mantid-roman: References?

DavidFair:
@mantid-roman I've gone through and changed the parameters to use references and also converted the loops to for ranged loops. Also the tick box behaviour should be more consistent throughout now. 
Would you be able to give it another once over to see if there is anything else that could be improved?

mantid-roman:
@DavidFair No problem

mantid-roman:
The Options and the Algorithms pane seem consistent now. One more little thing to consider. There is no way to hide all sub-categories and keep only the top level algorithms. For example, if you untick Diffraction\UsesPropertyManger all Diffraction disappears.

DavidFair:
Ok following a discussion with Roman I am happy with it in its current state. 
If unticking subcategories such as UsesPropertyManager causes issue for users we can look into modifying the `AlgorithmFactory` however that is out of the scope of this ticket.

mantid-roman:
I am happy to :shipit:

MantidPlot/src/ConfigDialog.cpp
```diff
@@ -715,8 +715,8 @@ void ConfigDialog::initMantidPage() {
 
   // Populate boxes
   auto faclist = cfgSvc.getFacilityNames();
-  for (auto it = faclist.begin(); it != faclist.end(); ++it) {
-    facility->addItem(QString::fromStdString(*it));
+  for (auto it : faclist) {
```
FedeMPouzols: In this and other loops below you probably can use `const auto &`. It's not going to do any difference here, but I'd say it's good practice in general to try to use all those specifiers whenever possible.

PR #16901, opened 2016-07-18
Fix uninitialized pointer in  EnggDiff interface

FedeMPouzols:
Small fix for an unitialized member issue that Coverity caught. I also fixed another uninit member in ListValidator.

**To test:**
- check code changes.

No issue number. 

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/EnggDiffraction/EnggDiffractionViewQtGUI.h
+++ b/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/EnggDiffraction/EnggDiffractionViewQtGUI.h
@@ -109,7 +109,7 @@ public:
   std::vector<std::string> newCeriaNo() const override;
 
   int currentCropCalibBankName() const override {
-    return m_currentCropCalibBankName;
+    return g_currentCropCalibBankName;
   }
 
   std::string currentCalibSpecNos() const override;
@@ -165,9 +165,9 @@ public:
 
   bool saveFocusedOutputFiles() const override;
 
-  int currentPlotType() const override { return m_currentType; }
+  int currentPlotType() const override { return g_currentType; }
 
-  int currentMultiRunMode() const override { return m_currentRunMode; }
+  int currentMultiRunMode() const override { return g_currentRunMode; }
 
 signals:
   void getBanks();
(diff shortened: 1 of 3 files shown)


abuts:
Looks reasonable :shipit:

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/EnggDiffraction/EnggDiffractionViewQtGUI.h
```diff
@@ -257,15 +257,15 @@ private slots:
...
+  EnggDiffFittingViewQtWidget *m_fittingWidget{nullptr};
   Ui::EnggDiffractionQtTabSettings m_uiTabSettings;
 
   /// converts QList to a vector
   std::vector<std::string> qListToVector(QStringList list,
                                          bool validator) const;
 
   /// instrument selected (ENGIN-X, etc.)
-  std::string m_currentInst;
+  std::string m_currentInst{"ENGINX"};
```
martyngigg: I'm not sure I see the need to use member initializers **and** constructor initializer lists here?

We have to write the constructor since the QWidget instance needs to be accepted. Could it not cause confusion if we have the initialization in two places?

FedeMPouzols: I see, in trying to use in-class initializers as much as possible probably this creates some confusion. And also the `IEnggDiffractionView` shouldn't have been left there.

Could we come up with a general best practice/recommendation from now on? If in cases like this that require a constructor we go for initialization in the constructor init list, but when the constructor is not required we prefer in-class initializers that could also cause some confusion. My assumption is that we want/wish to gradually increase the use of in-class initializers?

martyngigg: I think it's got to be about having a true single source of information about what ia member will be initialized to. If someone changed the default of `m_currentInst` in the class header but not the constructor then it's not obvious at first glance which one is used.

I think this is described well by http://stackoverflow.com/a/13662599. If a constructor allows overriding the default then initialize it in the class and similiar if we are using `= default` otherwise I would stick to the constructor init.

Issue #16913, opened 2016-07-19
Eliminate manual handling of `MatrixWorkspace::isDistribution` (Y storage mode) flag


SimonHeybrock:
This is a follow up to https://github.com/mantidproject/mantid/issues/16902. Currently `MatrixWorkspace::setDistribution` is used in various places. Since this simply sets a flag without any actual conversion it is error prone.

A better solution would be to set the correct storage mode on creation of a workspace, and remove `MatrixWorkspace::setDistribution` as wells as `Histogram::setYMode`. Changing in storage mode can then only happen by calling the conversion routines.

To accomplish this, we need to adapt `MatrixWorkspace::initialize` as well as some `WorkspaceFactory::create` routines. It must be possible to specify the desired Y storage mode (counts or frequencies) in the function calls.

PR #16914, opened 2016-07-19
Replace MatrixWorkspace isDistribution flag by Y-storage mode in Histogram

### LINKED ISSUE #16902: Add Y-storage-mode to Histogram
Currently `Histogram` does not know if it is storing counts or frequencies. This implies that the corresponding setters and getters will not give the correct result when storing frequencies, e.g., `Histogram::frequencies()` will scale `Histogram::m_y` with the inverse bin width, based on the expectation that `m_y` corresponds to counts. If this is not the case due to an external conversion, the resulting data will be garbage.

`MatrixWorkspace` has a flag tracking the storage mode (used for `MatrixWorkspace::isDistribution()`). Effectively this flag should be moved into `Histogram`.

SimonHeybrock:
See https://github.com/mantidproject/mantid/issues/16902 for a description.

This PR implements the storage mode for Y in histogram. `MatriWorkspace::isDistribution` is now based on the storage mode of the first spectrum (all spectra must have the same storage mode, and this is the case in current Mantid code).

In this PR no fundamental functionality has been changed. As a next step, https://github.com/mantidproject/mantid/issues/16913 will deal with cleaning up the storage mode mechanism.

**To test:**

Code review.

Fixes #16902 .

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/API/inc/MantidAPI/ISpectrum.h
+++ b/Framework/API/inc/MantidAPI/ISpectrum.h
@@ -105,7 +105,7 @@ public:
     HistogramData::Histogram histogram(std::forward<T>(data)...);
     // Check for the special case EventList, it only accepts histograms without
     // Y and E data.
-    checkHistogram(histogram);
+    checkAndSanitizeHistogram(histogram);
     mutableHistogramRef() = std::move(histogram);
   }
 
@@ -251,7 +251,7 @@ public:
   }
 
 protected:
-  virtual void checkHistogram(const HistogramData::Histogram &) const {}
+  virtual void checkAndSanitizeHistogram(HistogramData::Histogram &) {}
   virtual void checkWorksWithPoints() const {}
   virtual void checkIsYAndEWritable() const {}
 
--- a/Framework/API/inc/MantidAPI/ISpectrum.h
+++ b/Framework/API/inc/MantidAPI/ISpectrum.h
@@ -114,7 +114,7 @@ public:
   }
   void setYMode(HistogramData::Histogram::YMode ymode) {
     checkIsYAndEWritable();
-    return mutableHistogramRef().setYMode(ymode);
+    mutableHistogramRef().setYMode(ymode);
   }
   void convertToCounts() {
     checkIsYAndEWritable();
--- a/Framework/API/inc/MantidAPI/ISpectrum.h
+++ b/Framework/API/inc/MantidAPI/ISpectrum.h
@@ -113,12 +113,15 @@ public:
     return histogramRef().yMode();
   }
   void setYMode(HistogramData::Histogram::YMode ymode) {
+    checkIsYAndEWritable();
     return mutableHistogramRef().setYMode(ymode);
   }
   void convertToCounts() {
+    checkIsYAndEWritable();
     mutableHistogramRef().convertToCounts();
   }
   void convertToFrequencies() {
+    checkIsYAndEWritable();
     mutableHistogramRef().convertToFrequencies();
   }
 
--- a/Framework/API/inc/MantidAPI/ISpectrum.h
+++ b/Framework/API/inc/MantidAPI/ISpectrum.h
@@ -109,6 +109,19 @@ public:
     mutableHistogramRef() = std::move(histogram);
   }
 
+  HistogramData::Histogram::YMode yMode() const {
+    return histogramRef().yMode();
+  }
+  void setYMode(HistogramData::Histogram::YMode ymode) {
+    return mutableHistogramRef().setYMode(ymode);
+  }
+  void convertToCounts() {
+    mutableHistogramRef().convertToCounts();
+  }
+  void convertToFrequencies() {
+    mutableHistogramRef().convertToFrequencies();
+  }
+
   HistogramData::BinEdges binEdges() const { return histogramRef().binEdges(); }
   HistogramData::BinEdgeStandardDeviations binEdgeStandardDeviations() const {
     return histogramRef().binEdgeStandardDeviations();
--- a/Framework/API/inc/MantidAPI/ISpectrum.h
+++ b/Framework/API/inc/MantidAPI/ISpectrum.h
@@ -113,7 +113,6 @@ public:
     return histogramRef().yMode();
   }
   void setYMode(HistogramData::Histogram::YMode ymode) {
-    checkIsYAndEWritable();
     mutableHistogramRef().setYMode(ymode);
   }
   void convertToCounts() {
--- a/Framework/API/src/MatrixWorkspace.cpp
+++ b/Framework/API/src/MatrixWorkspace.cpp
@@ -41,9 +41,8 @@ const std::string MatrixWorkspace::yDimensionId = "yDimension";
 MatrixWorkspace::MatrixWorkspace(
     Mantid::Geometry::INearestNeighboursFactory *nnFactory)
     : IMDWorkspace(), ExperimentInfo(), m_axes(), m_isInitialized(false),
-      m_YUnit(), m_YUnitLabel(), m_isDistribution(false),
-      m_isCommonBinsFlagSet(false), m_isCommonBinsFlag(false), m_masks(),
-      m_indexCalculator(),
+      m_YUnit(), m_YUnitLabel(), m_isCommonBinsFlagSet(false),
+      m_isCommonBinsFlag(false), m_masks(), m_indexCalculator(),
       m_nearestNeighboursFactory(
           (nnFactory == nullptr) ? new NearestNeighboursFactory : nnFactory),
       m_nearestNeighbours() {}
@@ -57,7 +56,6 @@ MatrixWorkspace::MatrixWorkspace(const MatrixWorkspace &other)
   m_isInitialized = other.m_isInitialized;
   m_YUnit = other.m_YUnit;
   m_YUnitLabel = other.m_YUnitLabel;
-  m_isDistribution = other.m_isDistribution;
   m_isCommonBinsFlagSet = other.m_isCommonBinsFlagSet;
   m_isCommonBinsFlag = other.m_isCommonBinsFlag;
   m_masks = other.m_masks;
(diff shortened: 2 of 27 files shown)


Framework/API/inc/MantidAPI/ISpectrum.h
```diff
@@ -109,6 +109,22 @@ class DLLExport ISpectrum {
     mutableHistogramRef() = std::move(histogram);
   }
 
+  HistogramData::Histogram::YMode yMode() const {
+    return histogramRef().yMode();
+  }
+  void setYMode(HistogramData::Histogram::YMode ymode) {
+    checkIsYAndEWritable();
+    return mutableHistogramRef().setYMode(ymode);
```
SimonHeybrock: I'll remove this `return`.

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -926,13 +924,21 @@ void MatrixWorkspace::setYUnitLabel(const std::string &newLabel) {
 * TODO: For example: ????
 * @return whether workspace is a distribution or not
 */
-const bool &MatrixWorkspace::isDistribution() const { return m_isDistribution; }
+bool MatrixWorkspace::isDistribution() const {
+  return getSpectrum(0).yMode() == HistogramData::Histogram::YMode::Frequencies;
```
OwenArnold: Is it worth having a size check here? I don't know if we can have zero spectrum in a MatrixWorkspace, but this might break? I haven't checked, but maybe you are relying on a size check in `getSpectrum`.

SimonHeybrock: A `MatrixWorkspace` can never have zero spectra, the `initialize` method throws otherwise. Also many other things rely on this.

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -926,13 +924,21 @@ void MatrixWorkspace::setYUnitLabel(const std::string &newLabel) {
...
 *  @return whether workspace is now a distribution
 */
 void MatrixWorkspace::setDistribution(bool newValue) {
-  m_isDistribution = newValue;
+  if(isDistribution() == newValue)
+    return;
+  HistogramData::Histogram::YMode ymode =
+      newValue ? HistogramData::Histogram::YMode::Frequencies
+               : HistogramData::Histogram::YMode::Counts;
+  for (size_t i = 0; i < getNumberHistograms(); ++i)
```
OwenArnold: Maybe consider making `setDistribution` noexcept? We would want this to be atomic as it's happening in place.

Framework/DataHandling/src/LoadAscii.cpp
```diff
@@ -177,7 +177,9 @@ API::Workspace_sptr LoadAscii::readData(std::ifstream &file) const {
   // could be blank lines and comment lines
   int numBins(0), lineNo(0);
   std::vector<DataObjects::Histogram1D> spectra(
-      numSpectra, HistogramData::Histogram::XMode::Points);
+      numSpectra,
+      DataObjects::Histogram1D(HistogramData::Histogram::XMode::Points,
+                               HistogramData::Histogram::YMode::Counts));
```
OwenArnold: I much prefer this explicit `YMode` specification. I think this is a really good step to forcing people to think about the nature of their data at the point of import.

Framework/API/src/MatrixWorkspace.cpp
SimonHeybrock: I think there are two reasons not to:
1. I want to remove `setDistribution`, if possible (see follow-up issue).
2. `noexcept` is a promise that is true right now, but this is on a rather high level and things are still a bit in the flow, so I would not want to do that currently, since we would risk breaking other code if we remove it at a later point.

Framework/HistogramData/src/Histogram.cpp
```diff
@@ -140,6 +159,9 @@ void Histogram::setSharedX(const Kernel::cow_ptr<HistogramX> &x) & {
 
   Throws if the size does not match the current size. */
 void Histogram::setSharedY(const Kernel::cow_ptr<HistogramY> &y) & {
+  if (yMode() == YMode::Uninitialized)
+    throw std::logic_error(
+        "Histogram::setSharedY: YMode is not set and cannot be determined");
```
OwenArnold: 👍

SimonHeybrock: I have actually considered putting the `YMode` flag into `HistogramY`. That would allows us to increase the number of points where we have safety checks. What do you think? This could be done as a follow up at a later point.

Framework/HistogramData/src/Histogram.cpp
```diff
@@ -208,6 +254,20 @@ void Histogram::setUncertainties(const FrequencyStandardDeviations &e) {
   setFrequencyStandardDeviations(e);
 }
 
+void Histogram::checkAndSetYModeCounts() {
+  if (yMode() == YMode::Frequencies)
+    throw std::logic_error("Histogram: Y is storing Counts, modifying "
+                           "Frequencies is not possible.");
+  m_yMode = YMode::Counts;
+}
```
OwenArnold: So I think this is going to allow you to set the `YMode` to counts if uninitialised as well as if already `YMode::Counts`, which I think is correct.

Framework/HistogramData/test/HistogramTest.h
```diff
@@ -837,6 +851,72 @@ class HistogramTest : public CxxTest::TestSuite {
...
+    Histogram countHist(Points(1), Counts(1));
+    TS_ASSERT_THROWS(countHist.setFrequencies(1), std::logic_error);
+    TS_ASSERT_THROWS(countHist.setFrequencyVariances(1), std::logic_error);
+    TS_ASSERT_THROWS(countHist.setFrequencyStandardDeviations(1),
+                     std::logic_error);
+    Histogram freqHist(Points(1), Frequencies(1));
+    TS_ASSERT_THROWS(freqHist.setCounts(1), std::logic_error);
+    TS_ASSERT_THROWS(freqHist.setCountVariances(1), std::logic_error);
+    TS_ASSERT_THROWS(freqHist.setCountStandardDeviations(1), std::logic_error);
+  }
```
OwenArnold: Good, I would probably have broken this up into more individual test methods to test each scenario.

Framework/HistogramData/src/Histogram.cpp
OwenArnold: It's more logically associated with the `HistogramY`, so would probably best live there. However, I don't think it matters too much at this stage. I guess the conversion code counts->frequencies, frequencies->counts still makes most sense on `Histogram` as it touches all three Histogram vectors.

SimonHeybrock: @OwenArnold Yes, and there is another reason against putting the flag on `HistogramY`: We introduced that type to have something the client can use when they do not care what the underlying data actually is. If we now put the flag there, we could ask why do we not just have to distinct types and remove `Histogram::y()` altogether.

So I am not sure about this, I will think about it some more.

SimonHeybrock:
@OwenArnold I pushed an update with the changes we discussed, please have a careful look.

Framework/API/inc/MantidAPI/ISpectrum.h
```diff
@@ -105,7 +105,7 @@ class DLLExport ISpectrum {
     HistogramData::Histogram histogram(std::forward<T>(data)...);
     // Check for the special case EventList, it only accepts histograms without
     // Y and E data.
-    checkHistogram(histogram);
+    checkAndSanitizeHistogram(histogram);
     mutableHistogramRef() = std::move(histogram);
```
OwenArnold: Is this right, Will this potentially move `HistogramY` too?

OwenArnold:
Looks good to me otherwise 👍

Framework/API/inc/MantidAPI/ISpectrum.h
SimonHeybrock: It will move all of the `Histogram`, I am not sure I understand the question?

OwenArnold: The question is. For an `EventList` is it right to potentially take anything other than `HistogramX` and the mode flags?

SimonHeybrock: `checkAndSanitizeHistogram` takes care of that. For `EventList` it thows unless the `Y` and `E` in the `Histogram` are `nullptr`.

SimonHeybrock: And we would not just want to silently drop Y and E data. Raising an exception is safer.

OwenArnold: Yep, I've seen the code that's doing this now. No problems then. I just wanted to check that that situation was being dealt with.

OwenArnold:
@SimonHeybrock @martyngigg  I'm wondering if the changes to `EventList`, particularly the new methods added, should warrant some additional unit tests being provided at that level? I didn't see any changes to `EventListTest.h`.

Sorry for not mentioning that yesterday. I had a hand-written note about that that I never followed up on.

SimonHeybrock:
@OwenArnold Yes you are right, I'll do that.

SimonHeybrock:
@OwenArnold @martyngigg Ok, I'll just do the test as a separate PR then :)

martyngigg:
The page didn't update until after I'd hit merge :(

OwenArnold:
@SimonHeybrock @martyngigg I don't think it's worth restoring the branch or PR status for this. It can safely be done as a new addition.

SimonHeybrock:
@martyngigg Yes, github seemed to be updating automatically until a while ago, but recently it never does for me any more :(

martyngigg:
I just noticed there was a little "(1)" in the tab title bar on another PR that got edited. It's really easy to miss that though.

PR #16922, opened 2016-07-19
Chained BinMD slice viewer issues.

### LINKED ISSUE #16860: Chained BinMD in SliceViewer
Pascal and Users found an issue while reducing WISH data on a recent experiment. I have been able to reproduce the issue with a simple python script. While a single `BinMD` call works fine, **Instrument Scientists will run an initial `BinMD` to improve statistics and identify features before running `BinMD` a second time to create a slice of interest, all using the SliceViewer.**

**This is a high-priority fix for the next Mantid Version**

1) Run this in MantidPlot
```python
# Lets create a fake MDWorkspace
mdworkspace = CreateMDWorkspace(Dimensions=3, Extents='-10,10,-10,10,-10,10', Names='H,K,L', Units='U,U,U')
# Add A Peak at H=0, K=0, L=1
FakeMDEventData(InputWorkspace=mdworkspace, PeakParams='100000,0,0,0,1')
# Add another peak at H=2, K=2, L=0
FakeMDEventData(InputWorkspace=mdworkspace, PeakParams='100000,2,2,0,1')
```
2) Now open the SliceViewer
```python
# Open SliceViewer and set the slice point to K=1, half-way between the two Peaks.
plotSlice(mdworkspace, xydim=['H','L'], slicepoint=[0,1,0]) 
```

3) Now Integrate over a region in K of thickness 1.00. I'm in Rebin Mode. Here's the plot I get. I can correctly see both Peaks (image below). 
image
Furthermore, looking at the history of the generated slice `mdworkspace_rebinned` I can see that the OutputExtents are correct `-1.2,3,6,0,2,-1.2,1.2`, exactly as I specified in the SliceViewer.

4) Now open the SliceViewer again, but this time on the `mdworkspace_rebinned`. Cosmetically I get the correct view of the slice (below). However, note that **K=0, which is wrong/inconsistent, because in the first rebin, I took my slice at **K=1**

image

5) Now hit the Rebin button. I now get the correct slice at K=0 (we can only see the first peak we added), but this does not correspond to the original view I had loaded.
image
Also, although I have been allowed to enter Rebin mode, on the SliceViewer, it has not provided me with the ability to set the slice position or thickness I wanted. The thickness seems to be hard-coded to +/- 0.1 in this case looking at the history of the BinMD call.

## How I suspect this should be fixed.

1. We know that `BinMD` on the second-call will look at it's `MDHistogramWorkspace`, establish the original `MDWorkspace`, and perform the `BinMD` on that. For the second rebin, therefore in the SliceViewer, it seems to forget the extents that have already been applied in the first `BinMD`. I think the sliceviewer should ensure that `OutputExtents` are taken from the input `MDHistoWorkspace`, when performing a second BinMD call.  
2. In the case of performing this type of rebin, it should be possible for the user to specify the thickness, as they where able to in the first Rebin. That would stop the +/- 0.1 hardcoded thickness being fixed.
3. SliceViewer python tests should be added to cover these use cases.

samueljackson92:
**To Test**
See the issue description for more detail.

 A useful script was created by Owen which will generate some data, setup, and show the slice viewer:

``` python
# Lets create a fake MDWorkspace
mdworkspace = CreateMDWorkspace(Dimensions=3, Extents='-10,10,-10,10,-10,10', Names='H,K,L', Units='U,U,U')
# Add A Peak at H=0, K=0, L=1
FakeMDEventData(InputWorkspace=mdworkspace, PeakParams='100000,0,0,0,1')
# Add another peak at H=2, K=2, L=0
FakeMDEventData(InputWorkspace=mdworkspace, PeakParams='100000,2,2,0,1')
# Open SliceViewer and set the slice point to K=1, half-way between the two Peaks.
plotSlice(mdworkspace, xydim=['H','L'], slicepoint=[0,1,0]) 
```

From there set the following options:
- Toggle robin mode on
- Make sure that H is the X and L is the Y so we're slicing on K
- Slice point to 1.0
- Thickness to 1.0

Now right click on the newly generated workspace (called mdworkspace_rebinned) and open the slice viewer and toggle dynamic rebin mode. The screen should look something like the following:

screen shot 2016-07-19 at 12 17 18

Note that two peaks are visible and that the bin size of the K axis is 1.

Fixes #16860.

---
#### Reviewer

##### Code Review
##### Functional Tests


mantid-roman:
Works as described. :shipit:

MantidQt/SliceViewer/src/SliceViewer.cpp
```diff
@@ -670,6 +670,13 @@ void SliceViewer::updateDimensionSliceWidgets() {
     if (w > maxUnitsWidth)
       maxUnitsWidth = w;
 
+    // if the workspace is already binned, update the interface with the
+    // existing number of bins
+    if (boost::dynamic_pointer_cast<IMDHistoWorkspace>(m_ws)) {
```
martyngigg: @OwenArnold Is there any better way of testing for a histo vs event WS than a dynamic cast?

martyngigg:
@OwenArnold Is it worth including this in the patch once it's merged?

MantidQt/SliceViewer/src/SliceViewer.cpp
OwenArnold: Currently not that I'm aware. Facilities for this could be added at the `IMDWorkspace` level though.

OwenArnold:
@martyngigg. It is a bug. We need to make sure that it gets properly tested though by people at different facilities as part of any beta-testing.

MantidQt/SliceViewer/src/SliceViewer.cpp
samueljackson92: @martyngigg @OwenArnold I couldn't see an existing method for it either. I'd be happy to create an issue for it?

martyngigg: @samueljackson92 I think it might worth it. More `dynamic_cast`s only encourage further use of them. Thanks!

martyngigg:
@OwenArnold The patch doesn't usually go out to beta testers so if it's too high risk I wouldn't include it.

OwenArnold:
Id rather not risk this going out without it being beta-tested across multiple facilities. The SliceViewer is widely used, and I think this is one interface where we do not know all the usage scenarios our facility users apply to it.

MantidQt/SliceViewer/src/SliceViewer.cpp
samueljackson92: Issue number is #16977

FedeMPouzols:
The new method that tells you whether the `MDWorkspace` is a `MDHisto` one has been added elsewhere.
This looks ready to go (although not safe for the patch release)?

martyngigg:
As this hasn't been merged yet is it worth updating the code to remove the casts now? Saves us merging, creating another issue and retesting the same thing?

samueljackson92:
@martyngigg Sounds like a plan.

PR #16927, opened 2016-07-19
ResampleX bugfix

peterfpeterson:
NOMAD's autoreduction is failing and needs to be working again. The issue was tracked down to the particular way that `ResampleX` was being used.

**To test:**

The following script (using `DocTest` data) exposes the bug. Make sure that it works now.

``` python
LoadEventNexus(Filename='PG3_4871', OutputWorkspace='PG3_4871')
ConvertUnits(InputWorkspace='PG3_4871',OutputWorkspace='PG3_4871',
             Target='dSpacing', EMode='Elastic')
ResampleX(InputWorkspace='PG3_4871',OutputWorkspace='PG3_4871',
          NumberBins=6000)
```

_Does not need to be in the release notes_ because it was not a bug in a release version.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Algorithms/src/ResampleX.cpp
+++ b/Framework/Algorithms/src/ResampleX.cpp
@@ -346,9 +346,8 @@ void ResampleX::exec() {
         for (int wkspIndex = 0; wkspIndex < numSpectra; ++wkspIndex) {
           PARALLEL_START_INTERUPT_REGION
           BinEdges xValues(0);
-          double delta = this->determineBinning(xValues.mutableRawData(),
-                                                xmins[wkspIndex],
-                                                xmaxs[wkspIndex]);
+          double delta = this->determineBinning(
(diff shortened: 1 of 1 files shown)


SimonHeybrock:
Changes look good (will give it a shipit once I built this branch and ran the test).

The problem here is basically due to a quirk of `EventWorkspace`: It *ignores* the `X` length passed to `initialize`. When `EventWorkspace::setBinEdges` is called at a later point (with the correct length), the length check will fail.

The workaround used here is to instead use `setHistogram`. Since an `EventWorkspace` has no `Y` and `E` data, those can be left empty. Since `setHistogram` implies that any potential previous `Y` and `E` data is also replaced there is no length check, and thus does not cause an exception. This is the standard way to deal with this quirk, and is also done already in several other places.

PR #16941, opened 2016-07-20
Split Reflectometry interface into tabs

### LINKED ISSUE #16416: DataProcessorUI: make the view be the sole owner of the presenter
[This](https://github.com/mantidproject/mantid/blob/master/MantidQt/CustomInterfaces/src/Reflectometry/QDataProcessorWidget.cpp#L21) should be a `unique_ptr`.

### LINKED ISSUE #16869: Move global options to separate tab
This is how scientists would like global options to `ReflectometryReductionOneAuto` to be displayed in the new version of the GUI. Currently global options are displayed by the `DataProcessor` widget and therefore are in the same tab as the reduction table. This issue is to move them to a separate tab.

settingstab

raquelalvarezbanos:
Description of work: This PR refactors the existing functionality into two different tabs. The `Runs` tab contains the `Search` list and `Process` table. The `Settings` tab has text boxes to enter global options for the reduction.

**To test:**

Owen had a look at the MVP pattern (now we have the MVP at three different levels) and he is happy with how the code is organized. Therefore, tester only needs to ensure that existing functionality still works. This means reading carefully the [documentation](http://docs.mantidproject.org/nightly/interfaces/ISIS_Reflectometry.html) and ensuring that the functionality is the same as on the master branch.

Please pay special attention to:
- Catalog: make sure that you can use the catalog to search for runs: in "Search Runs" select "POLREF" and enter 1520249 for the investigation id. You should be prompted with a dialog requesting login details. If you have access to the ICat a list of runs should be populated on the left. Make sure runs are transferred to the table by clicking on the "Transfer" button.
- Make sure you can Import a TBL file and reduce data: a sample TBL file can be found here [SampleData-ISIS.zip](https://github.com/mantidproject/mantid/files/361036/SampleData-ISIS.zip). You should be able to load that file into the interface ("Import TBL") and reduce the data by clicking on "Process" (if nothing is selected or if groups are selected you will need to provide a value for "Stitch1DMany" in the "Settings" tab, you can just enter "Params="-0.04"" and the reduction should work.
- Make sure you can export your table to TBL format.
- Ensure table editing features work (those can be accessed from the two menus or from the toolbar). 

Fixes #16869.
Also Fixes #16416.

Release notes were updated in this commit: https://github.com/mantidproject/mantid/pull/16941/commits/f8ce4cc0d6c481a20a6a4c4b75c577a5ab1bb199

---
#### Reviewer

##### Code Review
##### Functional Tests


raquelalvarezbanos:
This is how I have structured the code:

- The main window uses MVP. Classes that form the MVP are:
 - Views: `IReflMainWindowView` (interface), `ReflMainWindowView` (implementation)
 - Presenters: `IReflMainWindowPresenter` (interface) and `ReflMainWindowPresenter` (implementation.

- GUI has two tabs, `Runs` and `Settings`. Settings provide some widgets to select global options for processing and post-processing (more options, including options for pre-processing will be added in the future, but I first want to have a meeting with Jos and Max to get a list of the options they want to have in this tab). Each tab uses MVP. Relevant classes are:
 - 'Runs' tab: `IReflRunsTabView` and `ReflRunsTabView` (view), `IReflRunsTabPresenter` and `ReflRunsTabPresenter` (presenter). `ReflRunsTabPresenter` also inherits from `DataProcessorMainPresenter` (our old `WorkspaceReceiver`) to receive the list of commands and table workspaces.
 - 'Settings' tab: `IReflSettingsTabView` and `ReflSettingsTabView` (view), `IReflSettingsTabPresenter` and `ReflSettingsTabPresenter`.

reflmvp

- Information is exchanged by presenters: when the `DataProcessorWidget` needs global options, `GenericDataProcessorPresenter` requests it to `ReflRunsTabPresnter` which requests it to `ReflMainWindowPresenter`, which finally requests it to `ReflSettingsTabPresenter`. 

- Messages/warnings/user input messages have been moved to the `MainWindow` (they were managed by the `DataProcessorWidget` when we had no tabs). When a tab wants to print a message the presenter notifies the `MainWindow` presenter, which acts on the `MainWindow` view.

- Each view is the owner of its presenter. The `MainWindow` presenter keeps a pointer to tab presenters so that information can be exchanged between them. An alternative approach would be that the `MainWindow` presenter creates and owns views and presenters (something similar to what is done in the Muon ALC interface). Not sure which approach is better and why?

- In addition I have also tried to fix #16416, to make the concrete `DataProcessor` view be the sole owner of the `DataProcessor` presenter.

raquelalvarezbanos:
@OwenArnold This is the work on the Reflectometry interface I mentioned last week. Would you be able to test this PR and let me know if I am following the right approach using the MVP at different levels? There are still many things to do (fix existing unit tests and add unit tests for new classes mainly), but I would like to make sure the approach is correct first.

OwenArnold:
@raquelalvarezbanos. Sure. If I hope to look at this tomorrow, but please do give me a nudge if I forget.

raquelalvarezbanos:
Thanks @OwenArnold I'll work on unit tests in the meantime.

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/IReflSettingsTabView.h
```diff
@@ -0,0 +1,57 @@
...
+  /// Constructor
+  IReflSettingsTabView(){};
+  /// Destructor
+  virtual ~IReflSettingsTabView(){};
+  /// Returns the presenter managing this view
+  virtual IReflSettingsTabPresenter *getPresenter() const = 0;
+
+  /// Processing
+  /// Returns selected analysis mode
+  virtual std::string getAnalysisMode() const = 0;
```
OwenArnold: Minor point. Rather than passing back strings that have no type safety. Should we perhaps have new types for the `AnalysisMode` which might have a method called `asPropertyName()`, which can be invoked if the `AnalysisMode` is being passed directly to one of the `DataProcessor` algorithms?

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/IReflSettingsTabView.h
```diff
@@ -0,0 +1,57 @@
...
+
+  /// Post-processing
+  /// Returns selected resolution (dQ/Q)
+  virtual std::string getResolution() const = 0;
```
OwenArnold: Same kind of argument as above, but slightly more confusing. `dQ/Q` is represented as a string? I would have expected this to be a `double`

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/QtReflSettingsTabView.h
```diff
@@ -0,0 +1,64 @@
...
+*/
+class QtReflSettingsTabView : public QWidget, public IReflSettingsTabView {
+  Q_OBJECT
+public:
+  /// Constructor
+  QtReflSettingsTabView(QWidget *parent = 0);
+  /// Destructor
+  ~QtReflSettingsTabView() override;
+  /// Returns the presenter managing this view
+  IReflSettingsTabPresenter *getPresenter() const;
```
OwenArnold: This is OK, but I started wondering if we ought to return a `WeakPtr`. I haven't figured that out properly yet.

In any case at least a ptr can be assigned and checked for a `nullptr`

MantidQt/CustomInterfaces/src/Reflectometry/QtReflRunsTabView.cpp
```diff
@@ -46,13 +46,13 @@ void QtReflMainView::initLayout() {
   ui.splitterTables->setStretchFactor(0, 0);
   ui.splitterTables->setStretchFactor(1, 1);
 
+  // Create the DataProcessor presenter
   ReflGenericDataProcessorPresenterFactory presenterFactory;
```
OwenArnold: Not sure about this. I like the `PresenterFactory`, but the concrete type has effectively been hard-coded. I think it would be better to inject the Factory, probably into the constructor of the `QtReflRunsTabView`

MantidQt/CustomInterfaces/src/Reflectometry/QtReflRunsTabView.cpp
```diff
@@ -46,13 +46,13 @@ void QtReflMainView::initLayout() {
...
+  auto processorPresenter = presenterFactory.create();
```
OwenArnold: This temporary `processorPresenter` is not required. I would suggest invoking `presenterFactory.create()` at line 54 directly into the constructor of the `QDataProcessorWidget`. 

Secondly, and it won't matter if you fix this above by avoiding the temporary, but don't use processorPresenter.release(), when you could use `std::move`

MantidQt/CustomInterfaces/src/Reflectometry/QtReflRunsTabView.cpp
```diff
@@ -72,15 +72,12 @@ void QtReflMainView::initLayout() {
...
 
-  m_presenter = boost::make_shared<ReflMainViewPresenter>(
-      this /*main view*/,
-      this /*currently this concrete view is also responsibile for prog reporting*/,
-      processorPresenter /*the table presenter*/);
+  // Create the presenter
+  m_presenter = std::make_shared<ReflRunsTabPresenter>(
+      this /* main view */,
+      this /* Currently this concrete view is also responsible for prog reporting */,
+      qDataProcessorWidget->getPresenter() /* The data processor presenter */);
```
OwenArnold: If you're not already, I would make sure that the pointers being passed back by `qDataProcessorWidget->getPresenter()` are of the form `Presenter * const`, this will prevent any client code from accidentally trying to call delete on a resource that is already being managed.

MantidQt/CustomInterfaces/src/Reflectometry/QtReflSettingsTabView.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+//----------------------------------------------------------------------------------------------
+/** Constructor
+* @param parent :: [input] The parent of this widget
+*/
+QtReflSettingsTabView::QtReflSettingsTabView(QWidget *parent) {
+
+  initLayout();
+
+  m_presenter.reset(new ReflSettingsTabPresenter(this));
```
OwenArnold: I would move this into the initializer list. That way you shouldn't need the reset.

MantidQt/CustomInterfaces/src/Reflectometry/QtReflSettingsTabView.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+Initialise the Interface
+*/
+void QtReflSettingsTabView::initLayout() { m_ui.setupUi(this); }
+
+/** Returns the presenter managing this view
+* @return :: A pointer to the presenter
+*/
+IReflSettingsTabPresenter *QtReflSettingsTabView::getPresenter() const {
```
OwenArnold: As I mentioned above, should be `IReflSettingsTabPresenter * const`

MantidQt/CustomInterfaces/src/Reflectometry/ReflMainWindowPresenter.cpp
```diff
@@ -0,0 +1,130 @@
...
+* @param settingsPresenter :: [input] A pointer to the 'Settings' tab presenter
+*/
+ReflMainWindowPresenter::ReflMainWindowPresenter(
+    IReflMainWindowView *view, IReflRunsTabPresenter *runsPresenter,
+    IReflSettingsTabPresenter *settingsPresenter)
+    : m_view(view), m_runsPresenter(runsPresenter),
+      m_settingsPresenter(settingsPresenter) {
+
+  // Tell the tab presenters that this is going to be the main presenter
+  m_runsPresenter->acceptMainPresenter(this);
```
OwenArnold: Good solution!

MantidQt/CustomInterfaces/src/Reflectometry/ReflMainWindowPresenter.cpp
```diff
@@ -0,0 +1,130 @@
...
+* @param view :: [input] The view we are managing
+* @param runsPresenter :: [input] A pointer to the 'Runs' tab presenter
+* @param settingsPresenter :: [input] A pointer to the 'Settings' tab presenter
+*/
+ReflMainWindowPresenter::ReflMainWindowPresenter(
+    IReflMainWindowView *view, IReflRunsTabPresenter *runsPresenter,
+    IReflSettingsTabPresenter *settingsPresenter)
+    : m_view(view), m_runsPresenter(runsPresenter),
+      m_settingsPresenter(settingsPresenter) {
+
```
OwenArnold: One alternative approach is to get the `IReflMainViewWindow` to return to you all sub presenters:

```cpp
view->getRunsPresenter();
view->getSettingsPresenter();
```
I think this form is slightly more testable, but I don't have a strong opinion.

MantidQt/CustomInterfaces/src/Reflectometry/ReflMainWindowPresenter.cpp
```diff
@@ -0,0 +1,130 @@
...
+  return std::map<std::string, std::string>();
+}
+
+/** Returns global processing options
+* @return :: Global processing options
+*/
+std::string ReflMainWindowPresenter::getProcessingOptions() {
+
+  // Request global processing options to 'Settings' presenter
+  return m_settingsPresenter->getProcessingOptions();
```
OwenArnold: Strictly, I think you should do a `nullptr` check on the `m_settingsPresenter` prior to this call. We have no guarantees that the pointer is valid, the sub `View` is the owner. While the owning `unique_ptr` shouldn't have gone out of scope, we can't guarantee it.

MantidQt/CustomInterfaces/src/Reflectometry/ReflSettingsTabPresenter.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+/** Get global post-processing options
+* @return :: Post-processing options as a string
+*/
+std::string ReflSettingsTabPresenter::getPostprocessingOptions() const {
+
+  std::vector<std::string> options;
+
+  // Resolution (dQ/Q)
+  options.push_back("Params=\"" + m_view->getResolution());
```
OwenArnold: Ok. This answers the previous question about why you return a string. I still think that a new type `QResolution` would be a better return type.

OwenArnold:
Generally this is excellent. I like the approach. I've made a few suggestions. 

I'm wondering if we ought to go back to having the `views` store their `presenters` as shared_pointers after all? That way we could do this:

```cpp

class view{
public: 
// The view owns the presenter, we cannot guarantee that it is available
virtual std::weak_ptr<PresenterX> getPresenterX() const = 0;
};


class PresenterY{
public:

void doSomething(){

  if(auto presenterX = m_view->getPresenterX().lock()){
  // We have a guarantee that presenterX is available.
    presenterX->doX(); 
  } else {
   throw std::runtime_error("No PresenterX on view X");
  }
}

};
```

On the other hand, it's more complex and leaks knowledge about the fundamental storage type of the presenter on the view. Any strong options?

MantidQt/CustomInterfaces/src/Reflectometry/ReflSettingsTabPresenter.cpp
raquelalvarezbanos: OK, I see. I guess it should be `ReflRunsTabPresenter` the one that converts global options to the right `key=value` format?

raquelalvarezbanos: I'm thinking that, for this PR, I will probably go back to the `HintingLineEdit`s, that is, I'll replace the resolution and polarization constants in the `Settings` tab by the `HintingLineEdit`s we currently have. I am having a meeting with scientists next week and would like to wait until then to have a more concrete list of the global options that should be shown in the `Settings` tab.

In future PR, I'll follow your suggestion and make the `ReflSettingsPresenter` unaware of how the specific algorithm is run.

OwenArnold: 👍

raquelalvarezbanos:
Regarding your last comment, can't we just do `if(!m_presenterX)` prior to calls to `m_presenterX` (as I think you suggest above)? What would be the benefit of the alternative approach with `shared_ptr`s? I'm slightly confused because I though we shouldn't use `shared_ptr` unless strictly necessary.

OwenArnold:
@raquelalvarezbanos. Yep. There's no **strong** shared usage, and from that perspective a `unique_ptr` is a far better option. However, in order to return a `weak_ptr`, you need to create one from a `shared_ptr` you cannot make one from a `unique_ptr`, that would not make sense.

The advantage of `weak_ptr` over a `nullptr` check is that it locks the resource so that it cannot be deleted between your check and the time you actually use it. You don't have such a strong guarantee with a `nullptr` check.

On balance, I think that the `nullptr` check is the better option in these scenarios. I suggest checking that the `ptr` is not null and throwing prior to using it. so for example:

```cpp
if(ptr == nullptr){
    throw std::runtime_error("");
}
// Safe to dereference pointer here now.
```

You might want to wrap the above code into a `checkPtrValid` function because you'll need it in a lot of places.

MantidQt/CustomInterfaces/src/Reflectometry/ReflMainWindowPresenter.cpp
raquelalvarezbanos: I'm not sure about this. In the `initLayout()` method, I have:

```
void QtReflMainWindowView::initLayout() {
  m_ui.setupUi(this);

  // Create the tabs
  auto runsPresenter = createRunsTab();
  auto settingsPresenter = createSettingsTab();

  // Create the presenter
  m_presenter.reset(
      new ReflMainWindowPresenter(this, runsPresenter, settingsPresenter));
}
```

If I don't pass `runsPresenter` and `settingsPresenter` as arguments to the `ReflMainWindowPresenter`'s constructor, the `QtReflMainWindowView` would need to keep pointers to both `IReflRunsTabPresenter` and `IReflSettingsTabPresenter`, so that I can later do:

```
IReflRunsTabPresenter * const QtReflMainWindowView::getRunsPresenter() {
    return m_runsPresenter;
}
```
 Wouldn't this imply that my view is slightly more aware of what's outside? I know it is already using the pointers to create the presenter, but it does not need to "keep track" of them. I guess that is the reason why I initially did this.

OwenArnold: Fair point. I think you're method is better.

MantidQt/CustomInterfaces/src/Reflectometry/QtReflRunsTabView.cpp
raquelalvarezbanos: OK, I can inject the factory into the constructor of `QtReflRunsTabView`, but `QtReflMainWindowView` will still need to create an object of type `ReflGenericDataProcessorPresenterFactory` and do something like:

```
void QtReflMainWindowView::initLayout() {
  m_ui.setupUi(this);

  ReflGenericDataProcessorPresenterFactory presenterFactory;

  // Create the tabs
  auto runsPresenter = createRunsTab(presenterFactory);
  auto settingsPresenter = createSettingsTab();

  // Create the presenter
  m_presenter.reset(
      new ReflMainWindowPresenter(this, runsPresenter, settingsPresenter));
}
```

Then:

```
IReflRunsTabPresenter *QtReflMainWindowView::createRunsTab(const ReflGenericDataProcessorPresenterFactory &factory) {

  QtReflRunsTabView *runsTab = new QtReflRunsTabView(this, factory);
  m_ui.mainTab->addTab(runsTab, QString("Runs"));

  return runsTab->getPresenter();
}
```

Why do it at this level and not at the `RunsTab` level?

OwenArnold: The best solution (most flexible) will be to ultimately push the creation/decision of the concrete factories right up to the outside of your top-most view. So that you can do something like this:
```cpp
new QtReflMainViewWindow(new StarndardISISPresenterFactory{}); 
```

Where you have an `AbstractFactory` (read AbstractFactory pattern)

```cpp
class AbstractPresenterFactory{

virtual IReflGenericDataProcessorPresenter* createDataProcessorPresenter() = 0;

virtual IReflRunTabPresenter* createRunTabPresenter() = 0;

...

};

class StarndardISISPresenterFactory : public AbstractPresenterFactory {
// Implement all concrete presenter types here.
};

```
At this point this is a minor problem to solve, so I wouldn't worry about changing what you have if you are not convinced.

raquelalvarezbanos:
I've added your suggestions and replaced global options with our current `HintingLineEdit`s in the `SettingsTab`. In issue #16983 I will go back to having normal `QWidget`s (I'll make sure the presenter knows as little as possible about how algorithms are run).

The only change I've not been able to made is making views return `IPresenterX * const` as I was getting a compiler warning about type qualifiers being ignored (see [here](http://builds.mantidproject.org/job/pull_requests-ubuntu/8365/console)).

PranavBahuguna:
On analysis of the reflectometry interface, I have found the following:

**Working:**
- I was prompted with a dialog requesting login details.
- ICat run list displayed on Search Runs interface after inserting login details and investigation id.
- Runs are transferred to table on clicking transfer button.
- Can import a TBL file.
- Data is reduced on clicking on process. Also works when none or groups only selected and params="-0.04" value is provided for Stich1DMany.
- Can export table to TBL format correctly.
- 'Process' works.
- 'Expand Selection' works.
- 'Plot Selected Rows' works. 
- 'Plot Selected Groups' works.
- 'Insert Row After' works. 
- 'Insert Group After' works.
- 'Delete Row' works. 
- 'Delete Group' works.
- 'Group Rows' works, though with issues described below.
- 'Copy Rows' works, though with issues described below.
- 'Cut Rows' works.
- 'Paste Rows' works.
- 'Clear Rows' works.

**Not working:**
- Clicking transfer button with no runs selected shows activity in progress bar to the left of it - to be fixed in #17057.
- Selecting multiple rows from within a single group and clicking the 'Group rows' button does not insert the rows into a new group - to be fixed in #17058.
- One can select a group, click the 'Copy Selected' then 'Paste Selected' buttons, which displays an error message - to be fixed in #17058.

Overall, this issue has been satisfactorily dealt with :shipit:

raquelalvarezbanos:
Thanks for testing this @PranavBahuguna :+1:

PR #16943, opened 2016-07-20
ConvertUnitsUsingDetectorTable emode bug

### LINKED ISSUE #16935: ConvertUnitsUsingDetectorTable sets EMode on failed run
See #16794 

This algorithms has a lot of copy pasted code from convertunits and suffers from the same problem.

NickDraper:
## Important

This branch is formed from #16881, and contains those changes - it should not be merged until that one has been merged

**To test:**

Several validation checks were taking place after the output workspace had already been altered. for In place workspaces this meant the workspace was altered even though the algorithm failed.

The changes delay the creation of the output workspace and changes until all of the validation has passed.

To test:
1. All unit tests and system tests must pass
2. Including the new unit test that checks this

Fixes #16935.

RELEASE NOTES

Either edit the file in docs/source/release/ui.rst

---
#### Reviewer

##### Code Review
##### Functional Tests


NickDraper:
@stuartcampbell As the original author of this alg I though I'd give you a heads up in case you wanted to check this PR has not broken anything.

Framework/Algorithms/src/ConvertUnits.cpp
```diff
@@ -333,7 +340,7 @@ ConvertUnits::convertQuickly(API::MatrixWorkspace_const_sptr inputWS,
 */
 bool ConvertUnits::getDetectorValues(
     const Kernel::Unit &outputUnit, const Geometry::IComponent &source,
-    const Geometry::IComponent &sample, double l1, int emode,
+    const Geometry::IComponent &sample, double &l1, int &emode,
```
SimonHeybrock: Why non-`const` reference? In any case, I would prefer passing things like `double` and `int` by value, after all an `int` is smaller than a pointer, and you avoid some pointer overhead.

Framework/Algorithms/src/ConvertUnitsUsingDetectorTable.cpp
```diff
@@ -276,29 +120,52 @@ void ConvertUnitsUsingDetectorTable::convertViaTOF(
...
-  EventWorkspace_sptr eventWS =
-      boost::dynamic_pointer_cast<EventWorkspace>(outputWS);
-  assert(static_cast<bool>(eventWS) == m_inputEvents); // Sanity check
+  // Now let's take a reference to the vectors.
+  const auto &l1Column = paramWS->getColVector<double>("l1");
+  const auto &l2Column = paramWS->getColVector<double>("l2");
+  const auto &twoThetaColumn = paramWS->getColVector<double>("twotheta");
+  const auto &efixedColumn = paramWS->getColVector<double>("efixed");
+  const auto &emodeColumn = paramWS->getColVector<int>("emode");
+  auto &spectraColumn = paramWS->getColVector<int>("spectra");
```
SimonHeybrock: This does not seem to be modified (lucky for us, otherwise it would edit the table since we now have a reference), so it should be `const`, I think.

Framework/Algorithms/src/ConvertUnitsUsingDetectorTable.cpp
```diff
@@ -276,29 +120,52 @@ void ConvertUnitsUsingDetectorTable::convertViaTOF(
...
 
   // Get the unit object for each workspace
-  Kernel::Unit_const_sptr outputUnit = outputWS->getAxis(0)->unit();
-
+  Kernel::Unit_const_sptr outputUnit = m_outputUnit;
   std::vector<double> emptyVec;
   int failedDetectorCount = 0;
 
-  // ConstColumnVector<int> spectraNumber = paramWS->getVector("spectra");
+  // Perform Sanity Validation before creating workspace
```
SimonHeybrock: @NickDraper Just for my understanding: I could not find an equivalent for the code block performing these checks in the old version, i.e., these are new checks you added?

Framework/Algorithms/src/ConvertUnitsUsingDetectorTable.cpp
```diff
@@ -91,168 +80,23 @@ void ConvertUnitsUsingDetectorTable::init() {
                   "http://www.mantidproject.org/ConvertUnits).");
 }
```
SimonHeybrock: `ConvertUnits::init` uses some validators for `InputWorkspace`. Since `ConvertUnitsUsingDetectorTable` is basically doing the same thing, maybe it makes sense to also put the stricter validators here?

SimonHeybrock:
@NickDraper This looks all good to me. According to my difftool `ConvertUnitsUsingDetectorTable` was identical to `ConvertUnits` apart from `ConvertViaTOF`, which is now the only remaining part in the former algorithm, i.e., a lot of duplicated code was removed 👍 .

After the minor comments have been fixed and @stuartcampbell gave his ok this is ready for shipping.

Framework/Algorithms/src/ConvertUnitsUsingDetectorTable.cpp
NickDraper: The previous code had some checks but they are down in the heart of the ToTOF and FromTOF of the unit classes themselves.  As such they were the cause of the problem as they threw exceptions after the output workspaces had already been modified.  Here, as in ConvertUnits I perform a trial unit conversion of a single spectra (from the input workspace) before I create the output workspace, so any errors from the unit classes can fire safely.

A better design would be to separate this validation into a more explicit method, but it would still need pretty much the same inputs.  Given the Core nature of ConvertUnits, I went for the smaller change.

Framework/Algorithms/src/ConvertUnits.cpp
NickDraper: This was a hang over from a an earlier refactoring attempt where I was hoping to only have the algorithms differ by implementations of getDetectorValues.  However it turned out to be more complex than that.

I'll revert this change.

SimonHeybrock:
@NickDraper This branch has developed merge conflicts.
@stuartcampbell Did you have a look at this, can I put the ship-it?

NickDraper:
This isn't going to make the release, moved to 3.9

stuartcampbell:
@SimonHeybrock - Sorry for some reason I missed the notification on this - yes I would say that it can be shipped as this is basically a bug fix and I don't think we should ship knowingly buggy code :shipit:

FedeMPouzols:
The branch has merge conflicts in 3 files apparently.

stuartcampbell:
@NickDraper or @SimonHeybrock will one of you resolve the merge conflicts ? I think that you are in the best position to work them out.. :smile:

SimonHeybrock:
@NickDraper I did the merge, but had some trouble afterwards. `ConvertUnitsUsingDetectorTableTest` passes now, but please have a close look to see if the merge went fine.

martyngigg (approved):

Issue #16808, opened 2016-07-06
Mask detectors with loaded xml mask does not work on grouped workspaces.


abuts:
Majority of workspaces on ISIS instruments come grouped into 4-1 groups where every spectra is related to 4 engineering detectors.

If you mask such workspace and SaveMask, the mask is saved as xml file, containing masked detectors id-s. 

When one loads such mask using LoadMask algorithm, the resulting mask workspace comes ungrouped, containing 4 times more spectra then the initial workspace with each spectra corresponding to a single detector. 

MaskDetectors algorithm than refuses to apply such mask workspace to the initial workspace, as number of spectra in the mask workspace and the initial workspace are different. 

As the result, ISIS direct inelastic group can not use xml masks and standard masking procedure.

Probably the easiest way to resolve this issue is to change MaskDetectors algorithm to analyse masks detector's id rather than number of spectra while applying mask workspace to another workspace.

PR #16945, opened 2016-07-20
Load spectra mask file applying it to the detector map.

### LINKED ISSUE #16848: Modify loadMask algorithm to accept workspace with spectra-detector maps
The ticket is the precondition to solve #16808 

Apparently, ISIS Workspaces came from instrument having a spectra-detector map fully unrelated to IDF Spectra-detector map. When one loads a mask, to apply it to a workspace correctly , spectra-detector map, used by this workspace is necessary.

abuts:
This is part of wider ticket #16808 to allow ISIS direct inelastic group to use xml-based detector maps.

**To test:**
The changes in functionality are probably best described by changed LoadMask algorithm description and its doctest so the part of the testing is to understand if the documentation is clear

Check if new functionality works without side effects and as intended. Check if the description of the new functionality is clear enough. 

Fixes #16848.

The changes are not particularly useful without #16808 so all release notes will be there. 

---
#### Reviewer

##### Code Review
##### Functional Tests


mantid-roman:
It looks like it's doing what is expected. But the code is very inconsistent in terms of coding standards. Names should be `camelCase`, start with a lower case letter and not contain underscores except for class members `g_...`, `m_...`.

abuts:
It was not in my scope to rewrite old algorithm and this is strange algorithm, containing a lot of things which does not seems have been completed. I believe, I made my small modifications according to modern coding standards but is reluctant to change everything else (except fixing copying vectors of strings by value) 

It seems that half of the old code can be thrown away, as the development went in a bit different direction but I  am not ready to make this decision without some discussion. I am also not sure what xml types supported by this algorithm are implemented and used. 

This is why I have left the majority of the old code untouched. If you think it is important, I would accept another ticket to modify the old algorithm according to modern code standards and cut some slack out of it. After discussion on what there is dead and what is used or intended to be used by somebody. 

But all this counts for a fair additional ticket, requesting some additional investigations. I think my changes are self-containing without this.

mantid-roman:
OK. Maybe you are right. Too much work for very little benefit. I am just saying that when there is a chance it's worth bringing at least some of the old code to the standard. With this comment we can :shipit:

Issue #16575, opened 2016-06-09
Add usage tracking to interfaces and plotting tools


NickDraper:
We should track custom interfaces and the more niche plotting tools like sliceviewer, vsi and spectrumviewer.

Add a usage report on startup of these.

PR #16946, opened 2016-07-20
Register VSI usage with UsageService

martyngigg:
The VSI was missed when the interfaces were updated to start tracking their usage. These changes plug that hole.

Each view is registered separately to get a better picture of the actual usage. 

**To test:**
1. Load the `SEQ_MDEW.nxs` from the system test data into MantidPlot
2. Run this line of Python to turn usage reporting on  UsageService.setEnabled(True) 
3. Right click on the workspace and show the VSI
4. Switch between all of the views
5. Close down MantidPlot
6. Go to http://reports.mantidproject.org/api/feature?page=31 (31 may need to be updated if this is no longer the last page in the results. If "next": null is at the top then it is the last page) in your browser and search on the page for VSI. There should be records for all of the view types.
7. Let me know so I can delete your testing data records from the DB.

No issue number.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Vates/VatesSimpleGui/ViewWidgets/src/MdViewerWidget.cpp
+++ b/Vates/VatesSimpleGui/ViewWidgets/src/MdViewerWidget.cpp
@@ -9,6 +9,7 @@
 #include "MantidKernel/Logger.h"
 #include "MantidKernel/ConfigService.h"
 #include "MantidKernel/InstrumentInfo.h"
+#include "MantidKernel/UsageService.h"
 #include "MantidQtAPI/InterfaceManager.h"
 #include "MantidQtAPI/MdConstants.h"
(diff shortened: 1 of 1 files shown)


quantumsteve:
@martyngigg Would this be a good candidate for the 3.7.1 patch release?

martyngigg:
@quantumsteve Good point, I think it would. Tracking of the other interfaces was also only added recently, #16575 , so we could pick that up to.

AntonPiccardoSelg:
@martyngigg There don't seem to be any entries regarding the VSI on the reports page. Isn't the reporting strategy such that Mantid bundles the reports and sends them out once a day?

AntonPiccardoSelg:
* Code looks fine
* The useage service works as described

happy to :shipit:

PR #16947, opened 2016-07-20
Fix bug creating MultiDomainFunction from string

### LINKED ISSUE #16942: Creating multi-domain function from string can sometimes fail
#### Expected behaviour
`FunctionFactory` should be able to create a MultiDomainFunction from its string representation (and so it should be possible to clone any MultiDomainFunction without problems).

#### Actual behaviour
Works for simple cases but falls over for more complicated functions - see below.

```python
stringSimple = "composite=MultiDomainFunction;name=FlatBackground,A0=0,$domains=i; \
name=FlatBackground,A0=1,$domains=(0,1);name=FlatBackground,A0=2,$domains=(0,2)"

stringComplex = "composite=MultiDomainFunction,NumDeriv=true;(name=FlatBackground,A0=0;\
name=MuonFInteraction,Lambda=0.2,Omega=0.5,Beta=1,A=1,ties=(Beta=1),$domains=i);\
(name=FlatBackground,A0=0;name=MuonFInteraction,Lambda=0.2,Omega=0.5,Beta=1,A=1,$domains=i);ties=(f1.f1.Beta=f0.f1.Beta)"

funcSimple = FunctionFactory.createInitialized(stringSimple)
print funcSimple

funcComplex = FunctionFactory.createInitialized(stringComplex)
print funcComplex
```

The "simple" function is created from string OK, but the "complex" one is not. The error seen is:

`ValueError: Attribute domains not found in function CompositeFunction`

(Note: the `stringComplex` above came from the `asString()` method of a real function, so it should be possible to turn it back into a clone of this function)

tom-perkins:
Fixes a bug found when creating a `MultiDomainFunction` from its string representation in certain cases (specifically, when the individual functions were themselves composite functions).

The `$domains` parent attribute is passed up to the parent function, but this only worked when the MultiDomainFunction contained simple functions. If it contained composite functions that themselves contained other functions, the attribute was only passed up one level. Now it is passed up as many levels as needed until it reaches the (grand)parent MultiDomainFunction.

**To test:**
- New unit test has been added for this case.
- Run the script in the issue - should work with no exceptions thrown.

Fixes #16942.

[Release notes](https://github.com/mantidproject/mantid/blob/a9718d8aa6094522c061172bbd949a42adc0f2d2/docs/source/release/v3.8.0/framework.rst#bug-fixes) have been updated.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/API/src/FunctionFactory.cpp
+++ b/Framework/API/src/FunctionFactory.cpp
@@ -216,7 +216,14 @@ CompositeFunction_sptr FunctionFactoryImpl::createComposite(
     cfun->addFunction(fun);
     size_t i = cfun->nFunctions() - 1;
     for (auto &pAttribute : pAttributes) {
-      cfun->setLocalAttributeValue(i, pAttribute.first, pAttribute.second);
+      // Apply parent attributes of the child function to this function. If this
+      // function doesn't have those attributes, they get passed up the chain to
+      // this function's parent.
+      if (cfun->hasLocalAttribute(pAttribute.first)) {
+        cfun->setLocalAttributeValue(i, pAttribute.first, pAttribute.second);
+      } else {
+        parentAttributes[pAttribute.first] = pAttribute.second;
+      }
     }
   }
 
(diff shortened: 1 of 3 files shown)


raquelalvarezbanos:
I've tried the script in the description and it works fine. Code changes look also good and a unit test was extended and a new one was added.

:shipit:

PR #16949, opened 2016-07-20
MatrixWorkspace::maskBins  Replace NaN with 0.

### LINKED ISSUE #16948: Masking to replace also NaN with 0
### Expected behaviour
Masking should 0 down also the NaN values/errors, if present.

### Actual behaviour
Masking is not replacing NaN with 0.

* Load/create a workspace with some NaN values.
* Mask the bins containing NaN with MaskBins or MaskDetectors.

### Platforms affected
ALL

gvardany:
This PR replaces NaN/Inf values/errors with 0 during the masking with non-0 weight.

**To test:**

- Load/create a workspace with some NaN values.
- Mask the bins containing NaN with MaskBins or MaskDetectors.

Fixes #16948.

[RELEASE NOTES](https://github.com/mantidproject/mantid/blob/16948_MatrixWorkspace_maskBins_replaceNaNwithZero/docs/source/release/v3.8.0/framework.rst#id17) 

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/API/src/MatrixWorkspace.cpp
+++ b/Framework/API/src/MatrixWorkspace.cpp
@@ -1061,8 +1061,11 @@ void MatrixWorkspace::maskBin(const size_t &workspaceIndex,
   // this is the actual result of the masking that most algorithms and plotting
   // implementations will see, the bin mask flags defined above are used by only
   // some algorithms
-  this->dataY(workspaceIndex)[binIndex] *= (1 - weight);
-  this->dataE(workspaceIndex)[binIndex] *= (1 - weight);
+  // NaN values are set to 0, other values are scaled by (1 - weight)
+  double &y = this->dataY(workspaceIndex)[binIndex];
+  isnan(y) ? y = 0 : y *= (1 - weight);
+  double &e = this->dataE(workspaceIndex)[binIndex];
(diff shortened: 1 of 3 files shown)


tom-perkins:
Tested with both `MaskBins` and `MaskDetectors`. In both cases NaNs were changed to zeros in bins that were masked.

- What should happen to Inf values? With `MaskDetectors` they are converted to zero too, but with `MaskBins` the Inf was changed into a NaN.
- Is it possible to add a test to `MatrixWorkspaceTest` for this case?

gvardany:
@tom-perkins Thanks for pointing this out. I have added a similar treatment also for `Inf` values.
Moreover, when the weight is 0, `maskBins` should be just equivalent to `flagMasked`, so the values should be conserved, including `NaN` and `Inf`. I have added the test. Please test again, now there should not be difference between what you get from `MaskBins` and `MaskDetectors`.

tom-perkins:
Changes look good, now NaN and Inf are both zeroed (except when the weight is zero) and the new test covers the changes. Happy to :shipit:

peterfpeterson:
This should be added to the release notes.

gvardany:
I have added a line in release notes.

PR #16951, opened 2016-07-20
Indirect - IndirectResolution unit test

### LINKED ISSUE #16892: Indirect Data Reduction - Indirect Resolution unit test
### Expected behavior
* IndirectResolution algorithm (like all other algorithms) should have a unit test.
### Actual behavior
* IndirectResolution.py currently does not have a unit test

### Platforms affected
* All

louisemccann:
Add a unit test for the algorithm IndirectResolution. Includes a test for normal execution as well as scale and log. Additionally ensured that sample rebinning params were added to the sample log (these are also now tested).

**To test:**

Run unit test locally and ensure they pass on the build servers

Fixes #16892

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/PythonInterface/test/python/plugins/algorithms/IndirectResolutionTest.py
+++ b/Framework/PythonInterface/test/python/plugins/algorithms/IndirectResolutionTest.py
@@ -5,51 +5,51 @@ import mantid
 from mantid.simpleapi import IndirectResolution
 
 class IndirectResolutionTest(unittest.TestCase):
-    
+
     def test_simple(self):
         res_ws = IndirectResolution(InputFiles='IRS26173.raw',
-                                Instrument='IRIS',
-                                Analyser='graphite',
-                                Reflection='002',
-                                DetectorRange=[3, 53],
-                                BackgroundRange=[-0.16, -0.14],
-                                RebinParam='-0.175, 0.002, 0.175')
-                                
+                                    Instrument='IRIS',
+                                    Analyser='graphite',
+                                    Reflection='002',
+                                    DetectorRange=[3, 53],
+                                    BackgroundRange=[-0.16, -0.14],
+                                    RebinParam='-0.175, 0.002, 0.175')
+
         self.assertEqual(res_ws.getNumberHistograms(), 1)
         self.assertEqual(res_ws.blocksize(), 175)
         run = res_ws.run()
         self.assertEqual(run.getProperty('res_back_end').value, -0.14)
         self.assertEqual(run.getProperty('rebin_high').value, 0.175)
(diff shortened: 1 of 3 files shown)


Framework/PythonInterface/test/python/plugins/algorithms/IndirectResolutionTest.py
```diff
@@ -0,0 +1,55 @@
...
+                                Analyser='graphite',
+                                Reflection='002',
+                                DetectorRange=[3, 53],
+                                BackgroundRange=[-0.16, -0.14],
+                                RebinParam='-0.175,0.002,0.175',
+                                ScaleFactor=10)
+                                
+        self.assertEqual(res_ws.getNumberHistograms(), 1)
+        self.assertEqual(res_ws.blocksize(), 175)
+        self.assertEqual(res_ws.run().getProperty('res_scale_factor').value, 10)
```
ElliotAOram: This is a good check to see that the scale factor has been added to the sample logs. I think it might also be worth checking that the scale has actually been applied to the workspace. You could do this by checking a bin matches an expected value or (even better) checking the scale factor vs. a non scaled workspace.

Framework/PythonInterface/test/python/plugins/algorithms/IndirectResolutionTest.py
```diff
@@ -0,0 +1,55 @@
...
+        
+    def test_logs(self):
+        res_ws = IndirectResolution(InputFiles='IRS26173.raw',
+                                Instrument='IRIS',
+                                Analyser='graphite',
+                                Reflection='002',
+                                DetectorRange=[3, 53],
+                                BackgroundRange=[-0.16, -0.14],
+                                RebinParam='-0.175,0.002,0.175',
+                                LoadLogFiles=True)
```
ElliotAOram: Minor change: Might be better to align the white space so that, for example, `InputFiles` lines up with `Instrument`:
```
res_ws = IndirectResolution(InputFiles='IRS26173.raw',
                            Instrument='IRIS',
                            Analyser='graphite',
                            ...)
```

ElliotAOram:
@louisemccann looks good on the whole. Just a couple of minor changes (see above) then we should be good to go!

ElliotAOram:
Looks good now :shipit:

FedeMPouzols:
Longer term I'd worry that if the algorithm loads multiple files in multiple test cases, or the same file many times, this unit test could be a bit slow. Maybe the slower/bigger file cases could go into a system test if needed.

ElliotAOram:
@FedeMPouzols I agree, I think perhaps we should look into a way of generating data instead for this and other algorithms in Indirect.

Issue #16739, opened 2016-06-27
Update Python algorithms & functions for compatibility with both Python 2 & 3 


martyngigg:
A cheat sheet is available at http://python-future.org/compatible_idioms.html that details the differences between Python 2 & 3 code. Most of our incompatibilities fall in to one of three categories:

* `print` is now a function and requires parentheses in Python 3
* commas not allowed in `except` clauses.
* `dict.iteritems()` no longer exists 

We need to maintain compatibility with Python 2. See the [Python 3](http://www.mantidproject.org/Python_3) page on the wiki for more detail.

This is an umbrella issue. Please add your username to the list below and include this ticket number in your PR.

1. ~~@martyngigg #16814~~
2. ~~@FedeMPouzols #16993~~
3. ~~@rosswhitfield #16952~~
4. ~~@rosswhitfield #16953~~
5. ~~@MichaelWedel #17052~~
6. ~~@DavidFair #17575~~
7. ~~@DavidFair #17579~~
8. ~~@DavidFair #17595~~
9.  ~~@rosswhitfield #16976~~
10. @rosswhitfield

**Group 1**
PythonInterface/plugins/algorithms/AlignComponents.py
PythonInterface/plugins/algorithms/ApplyNegMuCorrection.py
PythonInterface/plugins/algorithms/BASISReduction311.py
PythonInterface/plugins/algorithms/BASISReduction.py
PythonInterface/plugins/algorithms/CalculateSampleTransmission.py
PythonInterface/plugins/algorithms/CalibrateRectangularDetectors.py
PythonInterface/plugins/algorithms/CheckForSampleLogs.py
PythonInterface/plugins/algorithms/CleanFileCache.py
PythonInterface/plugins/algorithms/CollectHB3AExperimentInfo.py
PythonInterface/plugins/algorithms/CompareSampleLogs.py
PythonInterface/plugins/algorithms/ComputeCalibrationCoefVan.py
PythonInterface/plugins/algorithms/ComputeIncoherentDOS.py
PythonInterface/plugins/algorithms/ConjoinFiles.py
PythonInterface/plugins/algorithms/ConjoinSpectra.py
PythonInterface/plugins/algorithms/ConvertSnsRoiFileToMask.py
PythonInterface/plugins/algorithms/CorrectLogTimes.py
PythonInterface/plugins/algorithms/CorrectTOF.py
PythonInterface/plugins/algorithms/CreateCacheFilename.py
[...]

PR #16953, opened 2016-07-20
Update algorithms for Python3 - Group 4

rosswhitfield:
This is Group 4 of #16739 

**To test:**
- Code review
- Make sure the unittests still pass with python2.
- Test that the related unittests pass when using python3.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/PythonInterface/plugins/algorithms/LRAutoReduction.py
+++ b/Framework/PythonInterface/plugins/algorithms/LRAutoReduction.py
@@ -15,6 +15,7 @@ from mantid.simpleapi import *
 from mantid.kernel import *
 from reduction_gui.reduction.reflectometer.refl_data_series import DataSeries
 from reduction_gui.reduction.reflectometer.refl_data_script import DataSets
+from six import string_types
 
 
 class LRAutoReduction(PythonAlgorithm):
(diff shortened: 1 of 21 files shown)


peterfpeterson:
After merging this, tests that fail in python2 are:
```
$ ctest -R Python
...
469 - PythonInterfaceAPI_MultipleFilePropertyTest (Failed)
539 - PythonAlgorithms_LoadLogPropertyTableTest (Failed)
```
which are both broken on my system with nexus 4.4.1 anyhow. See #17001.

Lots more fail in python3. However, none are relevant to this PR.

I also reviewed the changes (very unexciting) and it is fine.

:shipit:

Framework/PythonInterface/plugins/algorithms/LRAutoReduction.py
```diff
@@ -425,7 +426,7 @@ def _get_direct_beam(self, meta_data_run, data_set):
         direct_beam_runs_str = self._read_property(meta_data_run, "direct_beam_runs",
                                                    _direct_beam_runs, is_string=True)
         # The direct runs in the DAS logs are stored as a string
-        if isinstance(direct_beam_runs_str, (str, unicode)):
+        if isinstance(direct_beam_runs_str, str):
```
MichaelWedel: Will this still do the right thing in Python 2 for unicode strings?

rosswhitfield: No it doesn't. `isinstance(u"foobar",(str,unicode)) = True` while `isinstance(u"foobar",str) = False`. But `unicode` isn't a type in python 3 so something else is needed here.

How about `isinstance(u"foobar",(str,type(u'')))`? It works for python 2 and 3. Unless there is a better way?

MichaelWedel: I just found this:

```
from six import string_types

if isinstance(direct_beam_runs_str, string_types):
    ....
```
But haven't tested it. From the docs:

"six.string_types: Possible types for text data. This is basestring() in Python 2 and str in Python 3."

rosswhitfield: Yeah that works. Thanks!

Issue #15370, opened 2016-02-18
C++11 Use unordered_map where possible


matthew-d-jones:
Check for places where we should be using `std::unordered_map` instead of `std::map`.

The same procedure for other containers was carried out for #15287

NickDraper:
completed by #15382

SimonHeybrock:
That PR does not seem to include changes for `unordered_map`, so I guess this should be reopened.

FedeMPouzols:
Note that one very core place where a change `std::map` => `std::unordered_map` probably makes sense is `DataService`, as discussed in #16967.

NickDraper:
archived

PR #16967, opened 2016-07-21
ADS GetObjectNames method - Add params to get sorted or hidden object names

### LINKED ISSUE #16500: ADS should return sorted object names
See discussion in #16496.

At present, the ADS returns workspace names (via `getObjectNames()`) as an unordered set, which then has to be sorted. Sorting occurs in several separate places:
- WorkspaceProperty
- WorkspaceSelector
- "Add workspace" dialog in Multi dataset fitting interface
(and maybe more?)

It would be nicer if `getObjectNames()` returned an ordered collection (already sorted) instead.

DavidFair:
Description of work.
Previously the ADS `getObjectNames` method returned an unordered set of names. Many of the callers of this method subsequently sorted these names. 

The method has been rewritten to allow users to specify if they want the names sorted whilst the `getObjectNamesInclHidden()` has been merged in as an additional hidden parameter. 
The default parameters still return a unsorted list with hidden items included if set in the configuration as per the original behaviour.

All callers of this method have been checked and any which subsequently sorted the values have been refactored.

**To test:**

All units tests should pass. 
WorkspaceSelector and the Add workspace dialogue in multi dataset fitting should return sorted lists still.

Fixes #16500 .

Doesn't need to be added to release notes.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Kernel/inc/MantidKernel/DataService.h
+++ b/Framework/Kernel/inc/MantidKernel/DataService.h
@@ -19,6 +19,15 @@
 
 namespace Mantid {
 namespace Kernel {
+
+/// Flag for whether to sort items before returning
+enum class DataServiceSort { Sorted, Unsorted };
+/**
+ * Flag for whether to include hidden items when returning,
+ * Auto queries the class to determine this behavior
+ */
+enum class DataServiceHidden { Auto, Include, Exclude };
+
 /** DataService stores instances of a given type.
     This is a templated class, designed to be implemented as a
     singleton. For simplicity and naming conventions, specialized classes must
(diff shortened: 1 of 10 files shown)


raquelalvarezbanos:
Looks good to me: `getObjectNames()` is now able to return unordered/ordered vector of workspace names (it is also possible to specify whether or not hidden workspaces should be included). `WorkspaceProperty`, `ADSValidator`, `MDFAddWorkspaceDialog` and `WorkspaceSelector` have been updated accordingly. There's also a new unit test testing the changes.

:shipit:

Framework/Kernel/inc/MantidKernel/DataService.h
```diff
@@ -19,6 +19,15 @@
...
 namespace Mantid {
 namespace Kernel {
+
+/// Flag for whether to sort items before returning
+enum class DataServiceSort { Sorted, Unsorted };
+/**
+ * Flag for whether to include hidden items when returning,
+ * Auto queries the class to determine this behavior
+ */
+enum class DataServiceHidden { Auto, Include, Exclude };
```
FedeMPouzols: When you have an enum that is specific to a class, an alternative is to declare that enum inside the class. That's a way of limiting its scope, similarly as you would do with namespaces. So you would have something like `Mantid::Kernel::DataService::SortOption` or `Mantid::Kernel::DataSerive::HiddenMode`.
I don't necessarily have a strong preference, but maybe this is a very clear-cut case? Notice the names of the enums are `DataService...`.

FedeMPouzols:
I think that the changes in this PR are good and tidy up the code, but I'm very confused by the traditional use of `map` and `unordered_map` in the `DataService`. The underlying structure is an ordered `std::map`  (typedef `svcmap`) but `getObjectNames()`  used to take the non-hidden objects from the map and insert them in an `unordered_map`, which would then need sorting.

I wouldn't be very concerned with performance as I think we normally have smallish collections of workspaces or instruments, but has there been any thought in the past as to whether the `DataService`s should keep its elements sorted, and what data structure it is supposed to be? @martyngigg, @NickDraper 

I'm wondering if, related to [this maintenance issue](https://github.com/mantidproject/mantid/issues/15370), the map underlying the analysis service should be an `unordered_map`.

Also, when I see that we now have the `Sorted` and `Unsorted` alternatives in getObjectNames(), I wonder if anyone willingly uses the `Unsorted` option, other than tests.

martyngigg:
@FedeMPouzols I think you are right to question the conflicted uses of these types. In C++03 days without tr1 then `map`  was the only `std` structure that could be used. These days I think `unordered_map` is the way to go in most situations as it's actually really a hash table and you're getting what you really want which is just a way to associated a value with a key.

I agree that part of the referenced maintenance issue should convert the underlying storage type from `map` to `unordered_map`.

On the point about the `Unsorted` option - I don't think it will be used much either but as it was simple to do we thought why not include it.

Framework/Kernel/inc/MantidKernel/DataService.h
martyngigg: @FedeMPouzols David chatted to me about this and usually I would agree with you. What we didn't like much was that the templated type would need to be included when calling the enum e.g.  `DataService<Workspace_sptr>::Sorted`, so this seemed like a reasonable compromise. What do you think?

FedeMPouzols: Ah, I hadn't considered that here there's the additional complexity of the template. That makes sense as a good compromise.

martyngigg: Thanks. I'll get this merged now.

FedeMPouzols:
Thanks @martyngigg, that makes sense. I've added a comment in the maintenance issue.

PR #16985, opened 2016-07-22
Use system python even if anaconda is in the path

peterfpeterson:
Some users at ORNL have been installing python which breaks `mantidpython` in a variety of ways. This starts the python that mantid was built with. This is only a solution for linux.

**To test:**
1. Activate anaconda (may need to install it first)
2. Start `mantidpython` and see that it happily tries to startup with that python. You can see which version is started by running `import sys;print sys.argv[0]". This cannot be passed via the command line as a`-c` option though.
3. Rerun `cmake to` get the templates filled in
4. Start `mantidpython` again to see the warning
5. Deactivate anaconda, start `mantidpython` and don't see the warning

Also try out `mantidpython --classic` (vanilla python) and `mantidpython notebook` to start a jupyter notebook.

---
#### Reviewer

##### Code Review
##### Functional Tests


rosswhitfield:
How about changing `command -v ipython` to just `/usr/bin/ipython`? Is there an OS that this doesn't work?

peterfpeterson:
This is a stopgap until we can find a more reliable way to figure out which `ipython` we were compiled against. CMake gives us `PYTHON_EXECUTABLE` but doesn't currently give out `IPYTHON_EXECUTABLE`. Once it does we can get rid of this warning and the `if` at the top of the script. Looking closer at the windows version of this file, it already implements something like that.

peterfpeterson:
Most of the previous comments can be ignored as this is a completely new solution that just uses the python mantid was built with.

rosswhitfield:
It works. I tested it will python 2 and 3. :shipit:

PR #16991, opened 2016-07-22
Add IMDWorkspace isHistogramData function

### LINKED ISSUE #16977: Add a function for check whether a IMDWorkspace is actually a IMDHistoWorkspace
There is currently (as far as I can see) no easy way of checking if a IMDWorkspace is a IMDHistoWorkspace without using a dynamic cast. This leads to selection statements like the following in SliceViewer.cpp:

```cpp
if(boost::dynamic_pointer_cast<IMDHistoWorkspace>(m_ws) {
  ...
}
``` 
The goal of this issue is to:
 - Add a function to provide this check
 - Replace code like the line above with the new function.

samueljackson92:
This PR is a spin off from PR #16922. Currently the only way to tell if a workspace was a histogram workspace was to dynamically cast it. 

The aim of this PR was to create a function that would allow us to avoid doing that by modifying IMDWorkspace. As we already have a function called isHistogramData() one way to do this is to move this one level up and implement it for MDEventWorkspace and MDHistoWorkspace. 

**To test:**
There's not much to test other than that the unit tests & system tests pass and that the code is good.

Fixes #16977.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/DataObjects/inc/MantidDataObjects/MDEventWorkspace.tcc
+++ b/Framework/DataObjects/inc/MantidDataObjects/MDEventWorkspace.tcc
@@ -1011,13 +1011,6 @@ TMDE(void MDEventWorkspace)::setDisplayNormalization(
 }
 
 /**
-  Return if the workspace is a histogram.
-*/
-TMDE(bool MDEventWorkspace)::isHistogramData() const {
-  return false;
-}
-
-/**
 Return the preferred normalization to use for visualization.
 */
 TMDE(API::MDNormalization MDEventWorkspace)::displayNormalization() const {
--- a/Framework/DataObjects/inc/MantidDataObjects/MDEventWorkspace.tcc
+++ b/Framework/DataObjects/inc/MantidDataObjects/MDEventWorkspace.tcc
@@ -1011,6 +1011,13 @@ TMDE(void MDEventWorkspace)::setDisplayNormalization(
 }
 
 /**
+  Return if the workspace is a histogram.
+*/
+TMDE(bool MDEventWorkspace)::isHistogramData() const {
+  return false;
+}
+
+/**
 Return the preferred normalization to use for visualization.
 */
 TMDE(API::MDNormalization MDEventWorkspace)::displayNormalization() const {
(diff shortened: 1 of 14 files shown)


Framework/DataObjects/inc/MantidDataObjects/MDEventWorkspace.tcc
```diff
@@ -1011,6 +1011,13 @@ TMDE(void MDEventWorkspace)::setDisplayNormalization(
 }
 
 /**
+  Return if the workspace is a histogram.
+*/
+TMDE(bool MDEventWorkspace)::isHistogramData() const {
+  return false;
```
SimonHeybrock: Why? Isn't an MD workspace always histogram data, in the sense of the `MatrixWorkspace::isHistogramData()` method?

SimonHeybrock:
@samueljackson92 If I understand the issue description correctly the method you need is *NOT* equivalent to `MatrixWorkspace::isHistogramData()`. The latter indicates whether the X data corresponds to bin edges or the data points (same length of X and Y vectors). In that sense `EventWorkspace` is also histogram data and so is, I think, every MD workspace including `MDEventWorkspace`. The grid boxes correspond to bin edges.

samueljackson92:
@SimonHeybrock My mistake. I've misunderstood the meaning of this method. You are correct, this should be separate method for differentiating MDEvent from MDHisto and that is not equivalent.

Framework/PythonInterface/mantid/api/src/Exports/MatrixWorkspace.cpp
```diff
@@ -230,8 +230,6 @@ void export_MatrixWorkspace() {
       .def("getAxis", &MatrixWorkspace::getAxis,
            (arg("self"), arg("axis_index")), return_internal_reference<>(),
            "Get a pointer to a workspace axis")
-      .def("isHistogramData", &MatrixWorkspace::isHistogramData, arg("self"),
-           "Returns True if this is considered to be binned data.")
```
SimonHeybrock: I think this change should also be reverted?

MantidQt/API/test/SignalRangeTest.h
```diff
@@ -44,6 +44,7 @@ class SignalRangeTest : public CxxTest::TestSuite {
     MOCK_METHOD0(clearMDMasking, void());
     MOCK_CONST_METHOD0(getSpecialCoordinateSystem,
                        Mantid::Kernel::SpecialCoordinateSystem());
+    MOCK_CONST_METHOD0(isHistogramData, bool());
```
SimonHeybrock: Is this correct?

SimonHeybrock:
Changes look ok now. This helps to avoid some `dynamic_casts`. If Jenkins gives his ok I would say this is ready for :shipit:.

PR #16952, opened 2016-07-20
Update algorithms for Python3 - Group 3

rosswhitfield:
This is Group 3 of #16739 

**To test:**
- Code review
- Make sure the unittests still pass with python2.

Does not need to be in the release notes.

---
#### Reviewer

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):
##### Code Review
- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)? Is it well structured with small focussed classes/methods/functions?
- [ ] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?
##### Functional Tests
- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant documentation been added/updated?
- [ ] Is user-facing documentation written in a user-friendly manner?
- [ ] Has developer documentation been updated if required?
- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

PR #16993, opened 2016-07-22
Update algorithms for Python 2/3 compatibility, group 2

FedeMPouzols:
Group 2 of #16739.

**To test:**
- check code changes
- tests should still pass (Python 2)
- If someone has a Python 3 Mantid setup, the tests should also pass there

---
#### Reviewer

##### Code Review
##### Functional Tests


rosswhitfield:
All the Engg tests fail because they use EnggUtils.py which hasn't been converted. If you fix [this one except line](https://github.com/mantidproject/mantid/blob/master/scripts/Engineering/EnggUtils.py#L56) the Engg tests pass.

FedeMPouzols:
Thanks for the tips. I had missed `EnggUtils`. It should be fixed now. For the other algorithms I'll have to setup a proper Python 3 environment and see.

Framework/PythonInterface/plugins/algorithms/DNSFlippingRatioCorr.py
```diff
@@ -67,10 +68,10 @@ def _flipper_valid(self):
         """
         # sort workspaces for sf and nsf
         nsf = []
-        for key in self.input_workspaces.keys():
+        for key in list(self.input_workspaces.keys()):
```
AndreiSavici: No need to transform to list when you iterate over keys

AndreiSavici:
There are several places where one iterates over some keys. You do not need to transform that into a list

Framework/PythonInterface/plugins/algorithms/DNSFlippingRatioCorr.py
FedeMPouzols: 2to3 does that all the time.

FedeMPouzols: Should we add this to the recipes given here: http://www.mantidproject.org/Python_3? 2to3 adds `list()` everywhere it sees `keys()` and it is easy to miss that.

AndreiSavici: Sure

FedeMPouzols:
@rosswhitfield, @martyngigg: I'm experiencing a glitch with one of the tests. It seems that with Python 3, when an algorithm raises an exception the simple API raises RuntimeError with an empty message string.
This [DakotaChiSquaredTest](https://github.com/mantidproject/mantid/blob/master/Framework/PythonInterface/test/python/plugins/algorithms/DakotaChiSquaredTest.py#L64) checks the exception **and the message** which I think is not very common, and it is failing because it gets empty strings.
Have you seen anything like this? I cannot find anything wrong in simpleapi.py related to this.

rosswhitfield:
I have `GetEiT0atSNSTest` failing for the reason in #16952 I wasn't able to fix it.

rosswhitfield:
LGTM All the related test (except `DakotaChiSquaredTest`) pass with python 3. I'm happy to leave `DakotaChiSquaredTest` for another time, maybe fix `GetEiT0atSNSTest` at the same time in another issue. :shipit:

martyngigg:
I've located the problem with the exception handling and it's in the C layer. It should be relatively easy to fix.

PR #17000, opened 2016-07-22
Avoid copying values into std::vector only to erase them

### LINKED ISSUE #15884: Copy only part of a column.
There are several places in [SaveDiffCall.cpp](https://github.com/mantidproject/mantid/blob/master/Framework/DataHandling/src/SaveDiffCal.cpp#L180) where `column::numeric_fill` is called followed by `std::vector::erase`. It would be better to just not add the extra elements in the previous step.

I think this best done by adding an optional parameter to [`column::numeric_fill`](https://github.com/mantidproject/mantid/blob/aeb269d5c61754eaaaf6d10c492fc7ef2cb84ebd/Framework/API/inc/MantidAPI/Column.h#L162) which specifies that maximum number of values to copy.

quantumsteve:
Description of work.

This adds an optional parameter to `Column::fill_numeric` to specify a maximum number of elements. Without this, there are two locations where we are inserting values into a vector only to erase them shortly after. 

**To test:**

Fixes #15884.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/API/inc/MantidAPI/Column.h
+++ b/Framework/API/inc/MantidAPI/Column.h
@@ -166,6 +166,7 @@ public:
     for (size_t i = 0; i < vec.size(); ++i) {
       vec[i] = static_cast<T>(toDouble(i));
     }
+    return vec;
   }
 
 protected:
--- a/Framework/API/inc/MantidAPI/Column.h
+++ b/Framework/API/inc/MantidAPI/Column.h
@@ -157,10 +157,12 @@ public:
 
   /**
    * Fills a std vector with values from the column if the types are compatible.
-   * @param vec :: The vector to fill in.
+   * @param maxSize :: Set size than the full column.
    */
-  template <class T> void numeric_fill(std::vector<T> &vec) const {
-    vec.resize(size());
+  template <class T = double>
+  std::vector<T>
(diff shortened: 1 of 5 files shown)


Framework/DataHandling/src/SaveDiffCal.cpp
```diff
@@ -121,8 +117,6 @@ void SaveDiffCal::writeIntFieldFromSVWS(
     H5::Group &group, const std::string &name,
     DataObjects::SpecialWorkspace2D_const_sptr ws) {
   auto detidCol = m_calibrationWS->getColumn("detid");
-  std::vector<detid_t> detids;
-  detidCol->numeric_fill(detids);
```
quantumsteve: Removing the in/out parameter LAO found an unused parameter :smile:

Framework/API/inc/MantidAPI/Column.h
```diff
@@ -157,10 +157,12 @@ class MANTID_API_DLL Column {
 
   /**
    * Fills a std vector with values from the column if the types are compatible.
-   * @param vec :: The vector to fill in.
+   * @param maxSize :: Set size than the full column.
```
KarlPalmen: Doesn't make sense. Missing word(s)?

martyngigg:
Nice set of changes that make this particular method far easier to use.

PR #17004, opened 2016-07-25
Histogram Rollout - StripPeaks

### LINKED ISSUE #16877: HistogramData Rollout: FrameWork/Algorithms/StripPeaks
This issue relates to the broader HistogramData rollout see [here](http://docs.mantidproject.org/nightly/concepts/HistogramData.html#rollout-status) and [here](https://github.com/mantidproject/mantid/issues/16872). We are currently refactoring code in order to make use of the new interface by replacing references to the old interface and making a few improvements where possible. In this issue you will be modifying StripPeaks.cpp.

### Examples

Examples of changes include:
 - `MantidVec &X = inputWS->dataX(0);` becomes `auto &X = inputWS->mutableX(0);`
 - `outputWS->dataX(i) = inputWS->readX(i);` becomes `outputWS->setSharedX(i, inputWS->sharedX(i));`
 - 
```c 
outputWS->dataY(i) = inputWS->readY(i);
outputWS->dataE(i) = intputWS->readE(i);
```
becomes
```
outputWS->setHistogram(i, inputWS->histogram(i));
```
### Tasks
There are a few tasks which should be completed when working on these files:
1.Add performance tests where they may be missing. Read [this](http://www.mantidproject.org/Writing_Performance_Tests) document carefully for information on how to create and run performance tests.
2. Run the SystemTest/PerformanceTest for the algorithm before making any changes and make a note of the execution time. How to run unit tests can be found [here](http://www.mantidproject.org/Running_the_unit_tests).
3. Modify StripPeaks as well as its unit test (StripPeaks.cpp/StripPeaksTest.cpp) to conform to the new Histogram standard.
3. After making changes run the SystemTest/Performance test and make sure the execution time has not increased significantly.

DavidFair:
Description of work.
StripPeaks has been converted to use the new histrogram interface and a performance test has additionally been added.

The performance test locally before changes took 1.7s-1.8s. With the changes the performance test completes in 1.75-1.85s 

**To test:**
The unit tests should all pass and the time taken before and after should not degrade significantly. Strip peaks output should not have changed. 

Fixes #16877 
Does not need to be added to the relase notes

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Algorithms/src/StripPeaks.cpp
+++ b/Framework/Algorithms/src/StripPeaks.cpp
@@ -252,7 +252,7 @@ StripPeaks::removePeaks(API::MatrixWorkspace_const_sptr input,
     }
 
     // Get central bin values using points
-    auto binCenters = outputWS->points(i);
+    auto binCenters = outputWS->points(peakslist->getRef<int>("spectrum", i));
     const int spectrumLength = static_cast<int>(Y.size());
     for (int j = 0; j < spectrumLength; ++j) {
       double x = binCenters[j];
--- a/Framework/Algorithms/src/StripPeaks.cpp
+++ b/Framework/Algorithms/src/StripPeaks.cpp
@@ -188,8 +188,6 @@ StripPeaks::removePeaks(API::MatrixWorkspace_const_sptr input,
     // but Y needs to be mutable
     auto &X = outputWS->x(peakslist->getRef<int>("spectrum", i));
     auto &Y = outputWS->mutableY(peakslist->getRef<int>("spectrum", i));
-    // Also create point data for later
-    auto &pointData = outputWS->points(i);
     // Get back the gaussian parameters
     const double height = peakslist->getRef<double>("Height", i);
     const double centre = peakslist->getRef<double>("PeakCentre", i);
@@ -253,10 +251,11 @@ StripPeaks::removePeaks(API::MatrixWorkspace_const_sptr input,
                           << " x + " << a2 << " x^2\n";
     }
 
+	// Get central bin values using points
+	auto binCenters = outputWS->points(i);
     const int spectrumLength = static_cast<int>(Y.size());
     for (int j = 0; j < spectrumLength; ++j) {
-      // If this is histogram data, we want to use the bin's central value
-      double x = pointData[j];
+      double x = binCenters[j];
       // Skip if not anywhere near this peak
       if (x < centre - 5.0 * width)
         continue;
--- a/Framework/Algorithms/src/StripPeaks.cpp
+++ b/Framework/Algorithms/src/StripPeaks.cpp
@@ -188,8 +188,8 @@ StripPeaks::removePeaks(API::MatrixWorkspace_const_sptr input,
     // but Y needs to be mutable
     auto &X = outputWS->x(peakslist->getRef<int>("spectrum", i));
     auto &Y = outputWS->mutableY(peakslist->getRef<int>("spectrum", i));
-	// Also create point data for later
-	auto &pointData = outputWS->points(i);
+    // Also create point data for later
+    auto &pointData = outputWS->points(i);
     // Get back the gaussian parameters
     const double height = peakslist->getRef<double>("Height", i);
     const double centre = peakslist->getRef<double>("PeakCentre", i);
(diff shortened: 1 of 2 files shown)


Framework/Algorithms/src/StripPeaks.cpp
```diff
@@ -187,9 +185,10 @@ StripPeaks::removePeaks(API::MatrixWorkspace_const_sptr input,
   prg = 0.3;
```
SimonHeybrock: The combination of the `isHistogramData` flag two lines above together with this:
https://github.com/mantidproject/mantid/pull/17004/files#diff-dbacdcd002cca75b9a2d1ad07830e4c4R259

.. can be replaced by `outputWS->points(index)`. Instead of getting `X` which may be bin centers (points or bin edges, you can simply get points directly.

Framework/Algorithms/test/StripPeaksTest.h
```diff
@@ -106,4 +106,50 @@ class StripPeaksTest : public CxxTest::TestSuite {
...
+    auto &X = WS->x(1);
+    auto &Y1 = WS->mutableY(1);
+    auto &E1 = WS->mutableE(1);
+    auto &Y0 = WS->mutableY(0);
+    for (size_t i = 0; i < Y1.size(); ++i) {
+      // Spectrum 0
+      Y0[i] = 5000;
+
+      // Spectrum 1
+      const double x = (X[i] + X[i + 1]) / 2;
```
SimonHeybrock: Above, use `auto X = WS->points(1)`, then you do not need this line.

Framework/Algorithms/src/StripPeaks.cpp
```diff
@@ -175,21 +175,21 @@ StripPeaks::removePeaks(API::MatrixWorkspace_const_sptr input,
...
   for (size_t i = 0; i < peakslist->rowCount(); ++i) {
-    // Get references to the data
-    const MantidVec &X = outputWS->readX(peakslist->getRef<int>("spectrum", i));
-    MantidVec &Y = outputWS->dataY(peakslist->getRef<int>("spectrum", i));
+    // Get references to the data - X can be const
+    // but Y needs to be mutable
+    auto &X = outputWS->x(peakslist->getRef<int>("spectrum", i));
+    auto &Y = outputWS->mutableY(peakslist->getRef<int>("spectrum", i));
+	// Also create point data for later
+	auto &pointData = outputWS->points(i);
```
SimonHeybrock: I would prefer to declare such variables close to where they are used, i.e., right before the `for` loop further down. That makes the code more readable and such code comments are not needed. Ideally the variable name should tell everything, e.g., `points` (which can mean many things, but it is commonly used in this context in Mantid) or maybe better something like `centerPoints` or `binCenters`?

Framework/Algorithms/src/StripPeaks.cpp
```diff
@@ -253,11 +253,11 @@ StripPeaks::removePeaks(API::MatrixWorkspace_const_sptr input,
                           << " x + " << a2 << " x^2\n";
     }
 
-    // Loop over the spectrum elements
+	
     const int spectrumLength = static_cast<int>(Y.size());
     for (int j = 0; j < spectrumLength; ++j) {
       // If this is histogram data, we want to use the bin's central value
```
SimonHeybrock: That comment was describing the computation based on `X`, which is now gone, so can be removed as well.

Framework/Algorithms/src/StripPeaks.cpp
SimonHeybrock: Probably this will not compile. `points()` returns by value, not by reference (the type`Point` is wrapping a pointer, so return by value is reasonably cheap).

DavidFair:
@SimonHeybrock I've added the changes suggested - would you be able to check for anything else which could be improved?

SimonHeybrock:
@DavidFair Looks good! Did you try running the new performance test before and after the changes? We might see a speedup in this case, and it would be interesting to know how big it is.

All changes look good to me, the performance test is a nice addition. :shipit:

DavidFair:
I've re-ran the performance tests with a newer machine and got the following results which confirms the new interface is faster.

Old interface : 1.55-1.60s
New interface : 1.30-1.35s

FedeMPouzols:
Excellent. Let's see how much speed-up the "performance" build machine sees.

SimonHeybrock:
@FedeMPouzols The performance test was added as part of this PR, so we do not have data from before the change.

If we want to do this, maybe in future we could split up the changes into two PRs, one that adds the performance tests, and a follow up with the actual changes?

FedeMPouzols:
Yes, splitting the performance test from the rollout, so we can see the before and after, sounds like a good idea for the algorithms where the rollout is expected to have a bigger impact. I'm not sure how many relevant algorithms don't have a performance test though.

PR #17008, opened 2016-07-25
LoadNexusProcessed use int instead of int64_t

### LINKED ISSUE #16923: In LoadNexusProcessed use int instead of int64_t
Currently `LoadNexusProcessed` uses int64_t for all of it's properties. However it checks the boundaries up to `Mantid::EMPTY_INT()` which returns the maximum value for an int meaning we cannot use the additional range provided.

We should go through and change these int64_t to int in the algorithm properties.

DavidFair:
Description of work.
Previously `LoadNexusProcessed` used int64_t for its spectra properties and throughout the file. However the upper bound was capped to the range of an int wasting all the additional space of the 64 bit int. 
Most of the int64_t has now been converted to either int or size_t correctly with one exception in a function where it is needed. 
**To test:**
All unit tests should pass. Loading processed nexus files should still work. 

Fixes #16923 .

---
#### Reviewer

##### Code Review
##### Functional Tests


peterfpeterson:
Why is there a file `Framework/DataHandling/src/LoadNexusProcessed.cpp.orig`? Isn't that what revision control systems do by default?

DavidFair:
@peterfpeterson This was an unwanted backup from the merge tool as the settings weren't set correctly. Thanks for spotting it.

Framework/DataHandling/src/LoadNexusProcessed.cpp
```diff
@@ -289,13 +289,15 @@ Workspace_sptr LoadNexusProcessed::doAccelleratedMultiPeriodLoading(
...
+      static_cast<int>(nHistograms) /
+      blockSize; // Truncated number of full blocks to read. Remainder removed
   const int readOptimumStop = (nFullBlocks * blockSize);
   const int readStop = m_spec_max - 1;
   const int finalBlockSize = readStop - readOptimumStop;
 
   int wsIndex = 0;
-  //This cast is safe as m_spec_min must be >= 1
+  // This cast is safe as m_spec_min must be >= 1
   int histIndex = m_spec_min - 1;
```
KarlPalmen: which cast?

KarlPalmen:
I found one comment that may be referring to a cast that has been removed. I suggest you check the comments and ensure they are up to date.

DavidFair:
@KarlPalmen I've gone through and checked for any additional comments which may be out of date. Thanks for spotting that.

KarlPalmen:
Works OK with a few processed next files I have. :shipit:

PR #17016, opened 2016-07-26
Enable use of start and end times as log values

### LINKED ISSUE #13195: Add start and end times to the list of log values
From James Lord:

> Another item to add to the shopping list for Muon Analysis: It would be good to be able to put run start and end times in the fitted Results Table, so we can plot something versus time in a kinetic experiment. At the moment the times aren't in the list of log values I can choose. I had to plot versus Run Number which is only acceptable if the beam current is steady and all runs are the same length.

> The times of runs are there in the Sample Logs dialog and accessible from Python.

> ALC versus time may be useful too.

tom-perkins:
Enable use of `run_start` and `run_end` in muon interfaces.

`run_start` and `run_end` have been added to the list of log values that can be used in the results table in Muon Analysis. They are available either as strings or as numeric values (seconds since start time of first run).

In the ALC interface, using the log values `run_start` and `run_end` now works as it should - the numeric values are seconds since the start of the first run.
- [Online documentation](http://www.mantidproject.org/MuonAnalysisResultsTable#Values) updated for Muon Analysis
- RST documentation updated for ALC
- RST documentation updated for `PlotAsymmetryByLogValue`

**To test:**
1. _Muon Analysis_
   - Open _Interfaces/Muon/Muon Analysis_
   - Load a dataset, for example `MUSR00015189.nxs` from test data (or any muon data)
   - Fit the data on the "Data Analysis" tab. Repeat with a couple of other datasets, using the same fitting model for each.
   - Go to "Results table" tab. In the top table, there should be options to select `run_start` and `run_end` as either text or seconds.
   - Select these options and generate the results table
   - Text should be ISO-formatted date/time string as a label column
   - Seconds should be number of seconds since first run start as an X column
2. _ALC_
   - Open _Interfaces/Muon/ALC_
   - Set first and last datasets, for example `MUSR00015189` and `MUSR00015191` (test data)
   - Set the log value dropdown to either `run_start` or `run_end`
   - Click "Load" and check data is loaded and plotted. The X axis should be number of seconds since the first run start.

Fixes #13195.

[Release notes](https://github.com/mantidproject/mantid/blob/cca8fc32bfbf63955490420561a23676fba5b114/docs/source/release/v3.8.0/muon.rst) have been updated.

---
#### Reviewer

##### Code Review
##### Functional Tests


KarlPalmen:
In running Muon Analysis. I find that if for the results table, one selects all four of the run_start and nun_end in both text and seconds. They come in the order 
```
run_start(text), run_start(s), run_end(s), run_end(text). 
```
Do you really want them to occur in this order?

KarlPalmen:
When running the ALC, run_start, it plots the first run at 0 as expected. If run_end is selected, it plots the first run at somewhere between 50 and 100.

tom-perkins:
In the Muon Analysis table, the order of columns in the table is not defined when the columns are from the same log (like `run_start (text)` and `run_start (s)`) -  I'll add this to `logNameLessThan`. 

For the ALC, the times are always relative to the start of the first run - even the end times. So the first run's end time is plotted at the length of the first run in seconds. (This is so the scales agree if you plot graphs versus start and end times).

KarlPalmen:
In the MuonAnalysis, the order the four appear in the results table is the same as the order their check boxes appear in the GUI.

tom-perkins:
Yes, it sorts the names putting the non-timeseries logs first based on a predefined order:

*[run number, run start, run end, sample temp, sample magnetic field]*

then the timeseries logs sorted alphabetically. When two non-timeseries entries are from the same log (`run_start (text)` and `run_start (s)` both come from the log `run_start`) then the order wasn't defined - I've pushed a change to compare alphabetically in this case. The order of the checkboxes and the order in the table should be the same, and now should go

*[run number, run start (s), run start (text), run end (s), run end (text), sample temp, sample magnetic field]*

KarlPalmen:
Works well :shipit:

PR #17050, opened 2016-07-27
Remove Material.h from ObjComponent.h and Object.h

peterfpeterson:
It was discovered that `Material.h` was included directly into `Object.h` and `ObjComponent.h`. Half of the fix was to have `ObjComponent` use its `Object->sample` to hold the material rather than explicitly have its own copy, The other half was to return `Material` by value rather than reference since they are relatively small. The effect of this work is to reduce compile times and source file dependencies.

**To test:**

_There was no associated issue_

_Since this is not a user-facing change, it does not need to be in the release notes._

---
#### Reviewer

##### Code Review
##### Functional Tests

This is part of the ongoing effort to decrease build times.


martyngigg:
Do we need this in `3.7.2`?

peterfpeterson:
I must have just selected the nearest milestone. Fixed.

Framework/Geometry/inc/MantidGeometry/Instrument/SampleEnvironmentSpecParser.h
```diff
@@ -3,6 +3,7 @@
 
 #include "MantidGeometry/DllConfig.h"
 #include "MantidGeometry/Instrument/SampleEnvironmentSpec.h"
+#include "MantidKernel/Material.h"
```
martyngigg: Was this necessary? I notice it was also included in the `.cpp` too?

peterfpeterson:
MSVS doesn't seem to like a `std::unordered_map<string, unique_ptr<Material>)`. Do you know how to get around it?

martyngigg:
Maybe it's not possible with `unique_ptr` and the `std` implementation of `unordered_map`. I can't actually get it to work with `gcc` at https://gcc.godbolt.org/. As the `SampleEnvironmentParser` is not low level I think we'll have to live with the header file there.

martyngigg:
Everything looks good here now.

:shipit:

FedeMPouzols:
Nice. This reduced the average number of includes in Framework cpp files from 116.6 down to 114.9.

PR #17053, opened 2016-07-27
Fix loading old muon NeXus files without frames_good

### LINKED ISSUE #16723: Muon Analysis: load dead times for older data
#### Expected behaviour
Loading data in MuonAnalysis should succeed, no matter which combination of options are chosen.
If it does fail, this shouldn't mess up the next attempt.

#### Actual behaviour
Cannot load some older data (converted to NeXus from older format) under certain conditions - see below

- Dataset: `\\olympic\babylon5\Scratch\TomPerkins\Data\keren_data\53590.NXS` (old MUT data)
- Open Muon Analysis interface
- Set "Dead Time Correction" dropdown to *None*
- Load the data
  - Data loads OK and a plot is shown
- Now set "Dead Time Correction" to *From Data File*
  - Data cannot be loaded and an error dialog is shown. Details in log:
mutlog
- Change the dropdown back to *None*. This doesn't help and you get an exception `add : Unable to insert Data Object : 'MuonAnalysisGrouped'`. Closing and reopening the interface also doesn't help, you have to restart Mantid to continue. 

#### Cause
The number of good frames (goodfrm) is not stored in the file, so the data can't be loaded with this option. However the error should be handled more gracefully and the message given to the user should be clearer.

tom-perkins:
Fix a bug seen in MuonAnalysis where attempting to load certain old muon NeXus files failed and caused the interface to lock up.

Cause: the file did not contain number of good frames, so no value was written to `goodfrm` in the workspace's run object. This made subsequent algorithms (`ApplyDeadTimeCorr`) fail.
- If `run/instrument/beam/frames_good` is not present, use `run/instrument/beam/frames` instead and assume all were good. Warn the user in the log that this is happening. 
- Added a test and updated documentation
- In `MuonProcess`, ensured algorithm fails fast if dead time correction fails. This avoids the interface locking up as was seen, and gives the user a clearer error message.

**To test:**
- The data files from the issue can be found at `\\olympic\babylon5\Scratch\TomPerkins\Data\keren_data`. These are old files that have been converted to NeXus and have no `run/instrument/beam/frames_good` field.
- An alternative data file to try is `EMU30604.nxs` from the system test data.
1. _From script:_
   - Load one of these files: 
   
   ``` python
   LoadMuonNexus(Filename='path/to/53590.NXS', OutputWorkspace='ws', DeadTimeTable='dt')
   ```
   - Look in the sample logs of the loaded workspace `ws` and check `goodfrm` is present. 
   - Check the log for a warning that `run/instrument/beam/frames` was used instead of   `run/instrument/beam/frames_good`.
   - Attempt to apply dead time correction (should not give an error):
   
   ``` python
   ApplyDeadTimeCorr(InputWorkspace='ws', DeadTimeTable='dt', OutputWorkspace='corr53590')
   ```
2. _Using GUI:_
   - Set facility to ISIS, open _Interfaces/Muon/Muon Analysis_ and set the "Dead Time Correction" dropdown to "From data file"
   - Select one of the files with "Browse" and load it
   - No error should be seen (should see a plot appear)

Fixes #16723.

[Release notes](https://github.com/mantidproject/mantid/blob/a147f4699abc655164f7cbf94bfb5976bdaf9832/docs/source/release/v3.8.0/muon.rst#algorithms) have been updated.

---
#### Reviewer

##### Code Review
##### Functional Tests


tom-perkins:
The reason for the system test failure is that the test file, `EMU30604.nxs`, also does not have number of good frames in it. Rather than giving an error, `CalMuonDeadTime` uses number of good frames = 1 in this case and continues... this means that the old system test results were not meaningful and the changes here have corrected them. 

To do:
- Update system test with correct results
- `CalMuonDeadTime` should throw if there is no `goodfrm` in the input workspace. At the moment it writes an error to the log but then continues regardless and gives incorrect results.

tom-perkins:
Updated `CalMuonDeadTime` and `ApplyDeadTimeCorr`, tests and docs and also updated system test results.

KarlPalmen:
:shipit:

PR #17078, opened 2016-07-29
WorkspaceListProperty Implementation

LamarMoore:
**Do no merge this PR. It is currently only for discussion**

Description of work.

**To test:**

Fixes #xxxx.

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/API/test/WorkspaceListPropertyTest.h
```diff
@@ -0,0 +1,304 @@
...
+    MyAlgorithm() { this->setRethrows(true); }
+
+    virtual int version() const { return 1; }
+
+    virtual const std::string name() const { return "MyAlgorithm"; }
+
+	const std::string summary() const override { return "MyAlgorithm helper."; }
+
+    virtual void init() {
+      declareProperty(std::make_unique<WorkspaceListProperty<Workspace>>(
```
OwenArnold: If we don't find a way to move `Workspace` into Kernel, then I suggest we name this to be a `DataItemListProperty` to avoid confusion about what is actually being stored.

Framework/API/test/WorkspaceListPropertyTest.h
```diff
@@ -0,0 +1,304 @@
...
+    return new WorkspaceListPropertyTest();
+  }
+  static void destroySuite(WorkspaceListPropertyTest *suite) { delete suite; }
+
+  //------------------------------------------------------------------------------
+  // Functional Testing
+  //------------------------------------------------------------------------------
+
+  void test_construction_minimalistic() {
+    AnalysisDataServiceImpl &ads = AnalysisDataService::Instance();
```
OwenArnold: I think we suggested that the ADS stuff was going to go (you're still working on this I think) in which case the constructor taking a workspace name must also be removed.

Framework/API/test/WorkspaceListPropertyTest.h
```diff
@@ -0,0 +1,304 @@
...
+    MyAlgorithm alg;
+    alg.initialize();
+
+    auto vecIn = std::vector<Workspace_sptr>();
+    vecIn.push_back(a);
+    vecIn.push_back(b);
+
+    alg.setProperty("MyProperty", vecIn);
+    //Now fetch the property.
+    std::string outStr = alg.getPropertyValue("MyProperty");
```
OwenArnold: I'm really not sure what we ought to use to represent the workspace "value". Name would make sense if all workspaces resided in the ADS, but is `value == std::string::empty()` the correct measure when the workspace has no name?

Framework/API/inc/MantidAPI/WorkspaceListProperty.h
```diff
@@ -0,0 +1,184 @@
...
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+
+template <typename TYPE = MatrixWorkspace>
+class DLLExport WorkspaceListProperty
+    : public Mantid::Kernel::PropertyWithValue<
+          std::vector<boost::shared_ptr<TYPE>>> {
+public:
+  /// Typedef the value type of this property with value.
```
OwenArnold: minor point. I think it's better to use `using` over `typedef`

Framework/API/inc/MantidAPI/WorkspaceListProperty.h
```diff
@@ -0,0 +1,184 @@
...
+
+    for (auto &wksp : workspaces) {
+      auto group = boost::dynamic_pointer_cast<WorkspaceGroup>(wksp);
+
+      if (group) {
+        auto names = group->getNames();
+        if (std::any_of(names.cbegin(), names.cend(), [](const std::string &s) {
+              return AnalysisDataService::Instance().doesExist(s);
+            })) {
+          error = "WorkspaceGroups which exist in the ADS are not allowed.";
```
OwenArnold: Good. Can you put a code comment around this to explain why we are doing it. Someone reading this might be wondering why you would enforce such a restriction.

Framework/API/test/WorkspaceListPropertyTest.h
```diff
@@ -0,0 +1,293 @@
...
+      std::vector<Workspace_sptr> val = getProperty("MyProperty");
+    }
+
+    virtual ~MyAlgorithm() {}
+  };
+
+  /**
+  * Helper class. Algorithms are instances of IPropertyManager
+  */
+  class MyAlgorithm2 : public Mantid::API::Algorithm {
```
OwenArnold: I don't see how `MyAlgorithm` type id different fro `MyAlgorithm2`

Framework/API/test/WorkspaceListPropertyTest.h
```diff
@@ -0,0 +1,293 @@
...
+                                         Direction::Input,
+                                         PropertyMode::Optional)));
+  }
+
+  void test_copy_construction() {
+    auto list = createWorkspaceList();
+
+    WorkspaceListProperty<Workspace> a("PropA", list, Direction::Input,
+                                       PropertyMode::Optional);
+    WorkspaceListProperty<Workspace> b(a);
```
OwenArnold: If `WorkspaceListProperty<T>` does not overload `operator==` it might be a good idea to add it so that you can use `TS_ASSERT_EQ` directly in the tests rather than re-imagining the equality operator here.

Framework/API/test/WorkspaceListPropertyTest.h
```diff
@@ -0,0 +1,293 @@
...
+  }
+
+  void test_clone() {
+    auto list = createWorkspaceList();
+    WorkspaceListProperty<Workspace> *propOne =
+        new WorkspaceListProperty<Workspace>("PropA", list, Direction::Input,
+                                             PropertyMode::Optional);
+    WorkspaceListProperty<Workspace> *propTwo = propOne->clone();
+    TS_ASSERT_DIFFERS(propOne, propTwo);
+    delete propOne;
```
OwenArnold: Minor detail. There's no reason to create `propOne` on the heap. You can just create it on the stack and then you don't need the `delete`

Framework/Kernel/inc/MantidKernel/PropertyWithValue.h
```diff
@@ -525,7 +525,7 @@ template <typename TYPE> class DLLExport PropertyWithValue : public Property {
    *  N.B. Problems found in validator are written to the log
    *  if you override this function to do checking outside a validator may want
    * to do more logging
-   *  @returns "" if the value is valid or a discription of the problem
+   *  @returns "" if the value is valid or a discription of the problemz
```
OwenArnold: typo problemz

Framework/API/src/ITableWorkspace.cpp
```diff
@@ -137,6 +137,25 @@ IPropertyManager::getValue<API::ITableWorkspace_const_sptr>(
...
+      dynamic_cast<
+          PropertyWithValue<std::vector<Mantid::API::ITableWorkspace_sptr>> *>(
+          getPointerToProperty(name));
+  if (prop) {
+    return *prop;
+  } else {
+    std::string message = "Attempt to assign property " + name +
+                          " to incorrect type. Expected type "
+                          "std::vector<boost::shared_ptr<Mantid::Kernel::"
+                          "DataItem>>";
```
OwenArnold: Is this right? "DataItem"

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -2090,6 +2090,25 @@ IPropertyManager::getValue<Mantid::API::MatrixWorkspace_const_sptr>(
...
+      dynamic_cast<
+          PropertyWithValue<std::vector<Mantid::API::MatrixWorkspace_sptr>> *>(
+          getPointerToProperty(name));
+  if (prop) {
+    return *prop;
+  } else {
+    std::string message = "Attempt to assign property " + name +
+                          " to incorrect type. Expected type "
+                          "std::vector<boost::shared_ptr<Mantid::Kernel::"
+                          "DataItem>>";
```
OwenArnold: Is this right? "DataItem"

Framework/API/src/WorkspaceGroup.cpp
```diff
@@ -438,6 +438,25 @@ IPropertyManager::getValue<Mantid::API::WorkspaceGroup_const_sptr>(
...
+      dynamic_cast<
+          PropertyWithValue<std::vector<Mantid::API::WorkspaceGroup_sptr>> *>(
+          getPointerToProperty(name));
+  if (prop) {
+    return *prop;
+  } else {
+    std::string message = "Attempt to assign property " + name +
+                          " to incorrect type. Expected type "
+                          "std::vector<boost::shared_ptr<Mantid::Kernel::"
+                          "DataItem>>";
```
OwenArnold: Is this right? "DataItem"

Framework/Kernel/src/IPropertyManager.cpp
```diff
@@ -23,6 +24,9 @@ DEFINE_IPROPERTYMANAGER_GETVALUE(std::vector<double>)
 DEFINE_IPROPERTYMANAGER_GETVALUE(std::vector<std::string>)
 DEFINE_IPROPERTYMANAGER_GETVALUE(std::vector<std::vector<std::string>>)
 
+typedef std::vector<boost::shared_ptr<Mantid::Kernel::DataItem>>
```
OwenArnold: Is this typedef used?

Framework/Kernel/inc/MantidKernel/IPropertyManager.h
```diff
@@ -467,6 +476,24 @@ class MANTID_KERNEL_DLL IPropertyManager {
...
   }
+
+  template <typename T>
+  IPropertyManager *setTypedProperty(const std::string &name,
+                                     const std::vector<T> &value,
+                                     const boost::true_type &) {
+    std::vector<boost::shared_ptr<DataItem>> tmp(value.size());
+
+    for (size_t i = 0; i < value.size(); i++)
+      tmp[i] = boost::static_pointer_cast<DataItem>(value[i]);
```
OwenArnold: What happens if the cast fails. Have you tried a vector of say `std::vector<int>` anywhere in testing?

OwenArnold:
@LamarMoore. I think this looks pretty good. 

One thing I would like to see more of in the tests is checking of how bad assignments (wrong type), wrong types of vectors etc. are handled.

Framework/API/inc/MantidAPI/WorkspaceListProperty.h
```diff
@@ -0,0 +1,184 @@
...
+  std::string setDataItems(
+      const std::vector<boost::shared_ptr<Kernel::DataItem>> &items) override {
+    std::string error;
+
+    std::vector<boost::shared_ptr<TYPE>> tmp(items.size());
+
+    for (size_t i = 0; i < items.size(); i++)
+      tmp[i] = boost::dynamic_pointer_cast<TYPE>(items[i]);
+
+    auto valid =
```
OwenArnold: Nice use of the lambda, however it might be more efficient to do the validation as part of the loop above. You could also break processing at the point that any null pointer is found.

Framework/API/test/WorkspaceListPropertyTest.h
LamarMoore: The issue with this is that the `PropertyWithValue<T>operator==` checks whether or not the properties being compared have identical names. `PropertyWithValue<T>operator=` does not copy the property name, just it's contents. So assignment will not create an identical copy of `a` into `b`. So `TS_ASSERT_EQUAL` will fail which is counter intuitive. Therefore, I have to check the contents.

Framework/Kernel/inc/MantidKernel/IPropertyManager.h
LamarMoore: The boost::true_type is only allowed where T is `boost::is_convertible` to DataItem. The false type defaults to the `setTypedProperty` method which will fail when attempting the dynamic_cast.

Framework/API/inc/MantidAPI/WorkspaceListProperty.h
LamarMoore: The for look loops over the items held by m_value. Once a valid WorkspaceGroup is discovered, it is tested for validity.

gvardany:
@LamarMoore Is this still relevant?

NickDraper:
@LamarMoore This PR has been stale since 2016, and fails build and is heavily conflicted.  I'm closing the PR for now, but leaving the branch open, we can create a new PR if this becomes ready.

PR #17079, opened 2016-07-29
Set parameter from log value in multi dataset fitting interface

### LINKED ISSUE #16494: Multi-dataset fitting: initialise parameters from log
Enhancement asked for by James Lord:

Add option to initialise function parameters from a log value in each workspace.

> I should be able to give a log name, e.g. Field_NMR or sample_temp, a choice of first/last/mean etc, and it will fill in the values for all the workspaces from that. I could then choose to have those fixed or as initial guesses. At present I have to type them all in manually. Even being able to paste a column of numbers copied from a table elsewhere would help.

tom-perkins:
Added a feature requested by scientists: in the "edit local parameter values" dialog, add options to set parameter values from a sample log.

The value can be set for one spectrum, or for all spectra. The log used can be chosen via a drop-down list, and, if it is a time series, there is the choice of taking the mean, min, max, first or last value.

Also noticed and fixed another problem: the "paste" function pasted data into the wrong column. 

**To test:**
- Load some workspaces that contain sample logs (or use the script in the comment below).
- Add workspaces with "Add workspaces" button
- Set up a fit function
- Click on a non-global parameter. A small button will appear, use this to get the _Edit local parameter values_ dialog:

mdfbutton

1) _"Set to log" options_
- In the dialog, open the drop-down menu next to one of the parameter values. Confirm new options are present to "set to log" and "set all to log", but they are not yet enabled.
- Tick the checkbox next to "Log" at the bottom and verify the options are now enabled.
- In the log combo box, select a log that is either a numeric value or a numeric time series.
- Try "Set to log" on one of the parameters. The value should be set to the value of the chosen sample log in that workspace.
- Try "Set all to log". All spectra should have their parameter values set to the sample log value from their respective workspaces.
- (If you chose a non-numeric log, you will get an exception, likewise if you selected a log that isn't shared by all workspaces. The user is required to select a sensible option...)
  mdfsettolog

2) _Paste values into table_
- Copy a list of numbers (same length as number of spectra) to the clipboard. They can be separated by newlines or commas. For example, if you loaded 3 workspaces, you could copy "1, 2, 3"
- Right-click on the header of the "Value" column and select "Paste"
- The values should be pasted into the "Value" column (and not the "Role" column)

Fixes #16494.

[Release notes](https://github.com/mantidproject/mantid/blob/d624e14c6474b1d4164035bc41dda2002e458e03/docs/source/release/v3.8.0/ui.rst#multi-dataset-fitting-interface-improvements) have been updated, with an image. There is no user documentation/tests for this interface.

---
#### Reviewer

##### Code Review
##### Functional Tests


tom-perkins:
Example script to generate test workspaces:
```python
import numpy as np

def func(x, a, b):
    return 0.2 + a*x + b

for i in range(1, 4):
    xData = []
    yData = []
    eData = []
    for x in np.arange(0.1, 10, 0.1):
        xData.append(x)
        yData.append(func(x, i, i/2.0))
        eData.append(0.1)

    outname = 'ws' + str(i)    
    CreateWorkspace(DataX=xData, DataY=yData, DataE=eData, NSpec=1, OutputWorkspace=outname)
    AddSampleLog(Workspace=outname, LogName='simple_log', LogText=str(i*i), LogType='Number', NumberType="Double")
    AddTimeSeriesLog(Workspace=outname, Name='time_log', Time="2010-01-01T00:00:0"+str(i), Value=str(i*i), Type="double")
    
group = GroupWorkspaces(['ws1','ws2','ws3'])
```
(The `GroupWorkspaces` is so you can add the whole group to the interface all at once, instead of one at a time)

tom-perkins:
Doesn't work for integer logs yet (`getLogAsSingleValue` is double only) - marked as "in progress"

tom-perkins:
At present this only works for double logs (or time series doubles). Opened issue #17087 to extend this to integers too by extending `getLogAsSingleValue` rather than implementing a solution here.

tom-perkins:
Merged master in to include the fix in #17105 - now should work with integer logs as well as doubles.

AntonPiccardoSelg:
Code looks good.
Followed the test instructions above. There are some things to consider:
* Setting the log value to a string causes an unhandled exception instead of a handled one.
* The log setup vanishes when I close and reopen the window. Is this intentional?

tom-perkins:
Thanks for reviewing - for the first point, I've changed this so that you see an error message and the value is set to NaN, rather than getting an unhandled exception.

At the moment the window just sets the value and doesn't remember that it came from a log value. In the future, it's intended that the log values will be included in the fit string for greater flexibility (#17184), and when that's done the same log can be set up when the window is reopened.

AntonPiccardoSelg:
The crash issue described above has been fixed and it behaves as expected. 
The copy-paste functionality also works as described above.
Happy to :shipit:

Issue #17077, opened 2016-07-29
flattenContainer() convenience function


soininen:
A function which flattens for example vector of vectors into a single vector is needed in a few places in Mantid. Currently, such function exists, namely `MultipleFileProperty::flattenFileNames()`, but it accepts only `std::vector<std::vector<std::string>>`s while treatment for vectors of `int`s would also be needed. 

To make the method more general, it could be converted into a function template and moved to a more suitable place, such as VectorHelper.h.

PR #17080, opened 2016-07-29
Moved `MultipleFileProperty::flattenFileNames()` to `VectorHelper::flattenVector()`

soininen:
Templatized `MultipleFileProperty::flattenFileNames()` to work on vectors of whatever type and moved the method to `VectorHelper` namespace under the name `flattenVector()`. The more generalized version will be needed in another branch I'm working on (issue 17009).

Changed the usages of the original method to use the new function.

Implements #17077.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Kernel/test/VectorHelperTest.h
+++ b/Framework/Kernel/test/VectorHelperTest.h
@@ -142,9 +142,8 @@ public:
 
   void test_flattenContainer_SingleSubvectorWithMultipleValues() {
     std::vector<std::vector<int>> input;
-    auto values = {3, 1, -1, -3, -5};
-    input.emplace_back(values);
-    const std::vector<int> expected(values);
+    input.emplace_back(std::vector<int>{3, 1, -1, -3, -5});
+    const std::vector<int> expected{3, 1, -1, -3, -5};
     const auto result = VectorHelper::flattenVector(input);
     TS_ASSERT_EQUALS(result.size(), expected.size());
     for (size_t i = 0; i < result.size(); ++i) {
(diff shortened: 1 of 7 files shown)


Framework/Kernel/test/VectorHelperTest.h
```diff
@@ -134,6 +134,53 @@ class VectorHelperTest : public CxxTest::TestSuite {
...
+    const auto result = VectorHelper::flattenVector(input);
+    TS_ASSERT_EQUALS(result.size(), expected.size());
+    for (size_t i = 0; i < result.size(); ++i) {
+      TS_ASSERT_EQUALS(result[i], expected[i]);
+    }
+  }
+
+  void test_flattenContainer_MultipleSubvectorsWithSingleValues() {
+    std::vector<std::vector<int>> input;
+    input.emplace_back(std::vector<int>{3});
```
quantumsteve: I don't think this is doing what you expect. Using `std::vector<>::emplace_back` like this will construct a temporary, then use the copy constructor to construct a `std::vector<int>` in place. You'd be better off using the a different constructor (for example `input.emplace_back(1,3);` or using `std::vector<std::vector<int>>::push_back`.

soininen: Actually, it is doing what I expect: I get an `input` vector with a single sub-vector containing the same values as in `expected`. Anyway, I have now refactored the test to resemble more the other ones.

DimitarTasev:
Works as expected, tests run successfully. :shipit:

PR #17089, opened 2016-08-01
Engg: Fix engineering diffraction interface crashes on new install

### LINKED ISSUE #17088: Engg:: Engineering diffraction interface crashes on new install
### Expected behavior
Opening the Engineering Diffraction interface under custom interfaces on a new install of Mantid should open the interface and create a pop up prompting the user to enter an RB number
### Actual behavior
The RB number pop up box appears then Mantid crashes. 
Install a fresh copy of Mantid, ensure the RB number box does not have an older value (from a previous entry). Open custom interfaces -> Diffraction -> Engineering Diffraction. This should cause the crash.
### Platforms affected
Windows 7 (Depends from machine to machine)
Windows 10

DavidFair:
Description of work.
On initialisation of the interface a notification should check and display a pop up informing the user they should enter a RB number if they have never entered one before (not used the interface before). 
However when this check happens and the window is set to visible QT framework gets confused between the main interface and the pop up window and deferences a null pointer crashing Mantid. 
To avoid this we should set the notification to appear after the main interface has loaded.
**To test:**
On a copy of Mantid where no RB number has been set open Custom Interfaces -> Diffraction -> Engineering Diffraction. If the window opens and a notification appears asking the user to enter an RB number the bug has been fixed

Fixes #17088 .

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidQt/CustomInterfaces/src/EnggDiffraction/EnggDiffractionViewQtGUI.cpp
+++ b/MantidQt/CustomInterfaces/src/EnggDiffraction/EnggDiffractionViewQtGUI.cpp
@@ -116,11 +116,11 @@ void EnggDiffractionViewQtGUI::initLayout() {
   doSetupTabSettings();
 
   m_presenter->notify(IEnggDiffractionPresenter::Start);
-  // Never do this here: if the RB-number checks show a pop-up (splash
-  // message), it will be shown very early (before the interface
+  // We need to delay the RB-number check for the pop-up (splash message)
+  // as it will be shown very early (before the interface
   // window itself is shown) and that will cause a crash in Qt code on
   // some platforms (windows 10, and 7 sometimes).
-  // m_presenter->notify(IEnggDiffractionPresenter::RBNumberChange);
+  // so perform the check in the showEvent method to check on start up
 }
 
 void EnggDiffractionViewQtGUI::doSetupTabCalib() {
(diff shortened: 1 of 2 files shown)


shahrozahmed1:
works fine for me ..  :shipit:

MantidQt/CustomInterfaces/src/EnggDiffraction/EnggDiffractionViewQtGUI.cpp
```diff
@@ -116,7 +116,11 @@ void EnggDiffractionViewQtGUI::initLayout() {
   doSetupTabSettings();
 
   m_presenter->notify(IEnggDiffractionPresenter::Start);
-  m_presenter->notify(IEnggDiffractionPresenter::RBNumberChange);
+  // Never do this here: if the RB-number checks show a pop-up (splash
+  // message), it will be shown very early (before the interface
+  // window itself is shown) and that will cause a crash in Qt code on
+  // some platforms (windows 10, and 7 sometimes).
+  // m_presenter->notify(IEnggDiffractionPresenter::RBNumberChange);
```
martyngigg: Can we remove the commented out code and update the comment to say that we delay notifying about the RB number change until the window is show?

DavidFair:
I've updated the comment to reflect it's behaviour and removed the commented out code as well.

PR #17100, opened 2016-08-01
PDFFourierTransform Q-range bugfix

peterfpeterson:
There was an issue with data being supplied that had `NaN` near the ends and it turned into a `G(r)` that was entirely `NaN` as a result. The solution is to quietly move in the boundaries to avoid the issue.

**To test:**

Since a bug was found and a unit test was added to test for it, this should just be a code review and make sure that the builds pass.

_There was no associated issue_

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/Algorithms/src/PDFFourierTransform.cpp
```diff
@@ -35,6 +36,8 @@ const string S_OF_Q("S(Q)");
 const string S_OF_Q_MINUS_ONE("S(Q)-1");
 /// Kernel of the Fourier transform
 const string Q_S_OF_Q_MINUS_ONE("Q[S(Q)-1]");
+
+const double TWO_OVER_PI(2. / M_PI);
```
SimonHeybrock: `constexpr`

Framework/Algorithms/src/PDFFourierTransform.cpp
```diff
@@ -137,6 +140,89 @@ std::map<string, string> PDFFourierTransform::validateInputs() {
...
+
+  // get index for the Qmin from the Q-range
+  auto q_iter = std::upper_bound(Q.begin(), Q.end(), qmin);
+  size_t qmin_index = std::distance(Q.begin(), q_iter);
+  if (qmin_index == 0)
+    qmin_index += 1; // so there doesn't have to be a check in integration loop
+
+  // go to first non-nan value
+  q_iter = std::find_if(std::next(FofQ.begin(), qmin_index), FofQ.end(),
+                        boost::math::isnormal<double>);
```
SimonHeybrock: Is this what you want? `0.0` is *not* normal. Maybe `!isfinite`? Also note that you do not need the `boost` versions, C++11 has them in `cmath`.

Framework/Algorithms/src/PDFFourierTransform.cpp
```diff
@@ -137,6 +140,89 @@ std::map<string, string> PDFFourierTransform::validateInputs() {
...
+    qmax = Q.back();
+  }
+
+  // get pointers for the data range
+  auto q_iter = std::lower_bound(Q.begin(), Q.end(), qmax);
+  size_t qmax_index = std::distance(Q.begin(), q_iter);
+
+  // go to first non-nan value
+  auto q_back_iter =
+      std::find_if(FofQ.rbegin(), FofQ.rend(), boost::math::isnormal<double>);
```
SimonHeybrock: See above, is `isnormal` correct?

Framework/Algorithms/src/PDFFourierTransform.cpp
```diff
@@ -220,50 +306,24 @@ void PDFFourierTransform::exec() {
...
-    auto qmin_ptr = std::upper_bound(inputQ.begin(), inputQ.end(), qmin);
-    qmin_index = std::distance(inputQ.begin(), qmin_ptr);
-    if (qmin_index == 0)
-      qmin_index += 1; // so there doesn't have to be a check below
-    auto qmax_ptr = std::lower_bound(inputQ.begin(), inputQ.end(), qmax);
-    qmax_index = std::distance(inputQ.begin(), qmax_ptr);
+    size_t qmi_out = qmax_index;
+    if (qmi_out == inputQ.size())
+      qmi_out--; // prevent unit test problem under windows (and probably other
+    // hardly identified problem)
```
SimonHeybrock: Is this mess still necessary? Your new `determineQMaxIndex` seems to return only valid `qmax_index` values?

Framework/Algorithms/src/PDFFourierTransform.cpp
```diff
@@ -296,12 +358,14 @@ void PDFFourierTransform::exec() {
     double fs = 0;
     double error = 0;
     for (size_t q_index = qmin_index; q_index < qmax_index; q_index++) {
```
SimonHeybrock: Should this be including `qmax_index`?

Framework/Algorithms/src/PDFFourierTransform.cpp
```diff
@@ -296,12 +358,14 @@ void PDFFourierTransform::exec() {
...
-      double deltaq = inputQ[q_index] - inputQ[q_index - 1];
+      const double q = inputQ[q_index];
+      const double deltaq = inputQ[q_index] - inputQ[q_index - 1];
       double sinus = sin(q * r) * deltaq;
+
       // multiply by filter function sin(q*pi/qmax)/(q*pi/qmax)
       if (filter && q != 0) {
-        sinus *= sin(q * rdelta) / (q * rdelta);
+        const double lorchKernel = q * M_PI / inputQ[qmax_index];
+        sinus *= sin(lorchKernel) / lorchKernel;
```
SimonHeybrock: The old version had a bug?

peterfpeterson: The old version was using the full data Q-range rather than the range used in the Fourier transform. That is why the tests had to change.

SimonHeybrock: I was referring to the factor \pi.

peterfpeterson: That factor was hiding in the default value of `rdelta`. This version gets rid of the assumption that the user didn't override `rdelta`.

Framework/Algorithms/src/PDFFourierTransform.cpp
peterfpeterson: `std::isfinite` produces the wrong behavior. Thank you new unit test.

SimonHeybrock:
@peterfpeterson Apart from my comment regarding the range of the `q_index` loop this looks ok. Did you verify whether the range is correct or not?

peterfpeterson:
The Q-range is the traditional `[min, max)`.

SimonHeybrock:
Ok, then I'd say this is ready for :shipit:.

PR #17105, opened 2016-08-02
Extend getLogAsSingleValue for other numeric types

### LINKED ISSUE #17087: Extend getLogAsSingleValue to work with integers
Issue raised in #17079 and required for this feature to work correctly.

#### Expected behaviour
The function `getLogAsSingleValue` should work on any sample log of numeric type, or a numeric time series. It should do what the name suggests - return one `double` value from the named log using the given statistic type.

#### Actual behaviour
The function only works with logs of type `double` or `TimeSeriesProperty<double>`. If the log is an integer value or time series of integers, an exception is thrown. The caller would reasonably expect that the function should work with any numeric log.

#### Resolution
Extend the function to deal with integer types too. It could even be clever and deal with strings that can be parsed to numeric values. See [here](https://github.com/mantidproject/mantid/blob/1458e3fd2a9b26e565f46c654ecd37fd5095e04a/Framework/Algorithms/src/PlotAsymmetryByLogValue.cpp#L34-L60) and [here](https://github.com/mantidproject/mantid/blob/1458e3fd2a9b26e565f46c654ecd37fd5095e04a/Framework/Algorithms/src/PlotAsymmetryByLogValue.cpp#L838-L861) for an example.

tom-perkins:
Extend `LogManager::getLogAsSingleValue` to work with logs of integer (or time series integer) types as well as doubles and time series doubles.

`LogManager::getLogAsSingleValue` previously only worked for log values stored as doubles or time series doubles and threw an exception if used on an integer log. It is less surprising to the caller if the function works on any numeric, or time series numeric, log and does what is suggested by the name.

**To test:**
Unit tests have been written to cover all new cases. Since this function is not exposed to Python it is difficult to test manually - code review and check the new tests cover the behaviour.

Fixes #17087. (This feature is also required for #17079.)

_Does not need to be in the release notes_ (internal change not visible to users)

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/API/src/LogManager.cpp
+++ b/Framework/API/src/LogManager.cpp
@@ -356,8 +356,8 @@ double LogManager::getPropertyAsSingleValue(
         return std::stod(stringLog->value());
       } catch (const std::invalid_argument &) {
         throw std::invalid_argument(
-            "Run::getPropertyAsSingleValue - Property\"" + name +
-            "\"cannot be converted to a numeric value.");
+            "Run::getPropertyAsSingleValue - Property \"" + name +
+            "\" cannot be converted to a numeric value.");
       }
     } else {
       throw std::invalid_argument(
--- a/Framework/API/src/LogManager.cpp
+++ b/Framework/API/src/LogManager.cpp
@@ -335,6 +335,16 @@ double LogManager::getPropertyAsSingleValue(
     if (convertSingleValue(log, singleValue) ||
         convertTimeSeriesToDouble(log, singleValue, statistic)) {
       return singleValue;
+    } else if (const auto stringLog =
+                   dynamic_cast<const PropertyWithValue<std::string> *>(log)) {
+      // Try to lexically cast string to a double
+      try {
+        return std::stod(stringLog->value());
+      } catch (const std::invalid_argument &) {
+        throw std::invalid_argument(
+            "Run::getPropertyAsSingleValue - Property\"" + name +
+            "\"cannot be converted to a numeric value.");
+      }
     } else {
       throw std::invalid_argument(
           "Run::getPropertyAsSingleValue - Property \"" + name +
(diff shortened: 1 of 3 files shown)


Framework/API/src/LogManager.cpp
```diff
@@ -268,16 +346,23 @@ double LogManager::getPropertyAsSingleValue(
...
+        throw std::invalid_argument(
+            "Run::getPropertyAsSingleValue - Property\"" + name +
+            "\"cannot be converted to a numeric value.");
+      }
     } else {
       throw std::invalid_argument(
           "Run::getPropertyAsSingleValue - Property \"" + name +
-          "\" is not a single double or time series double.");
+          "\" is not a single numeric value or numeric time series.");
     }
```
ianbush: I think you are now missing a test for this second exception branch, should that be checked too? You may also want to check the message given when throwing for the string, I think it is missing some spaces :smile: .

Otherwise all looks good.

Framework/API/src/LogManager.cpp
```diff
@@ -268,16 +346,23 @@ double LogManager::getPropertyAsSingleValue(
...
+    } else if (const auto stringLog =
+                   dynamic_cast<const PropertyWithValue<std::string> *>(log)) {
+      // Try to lexically cast string to a double
+      try {
+        return std::stod(stringLog->value());
+      } catch (const std::invalid_argument &) {
+        throw std::invalid_argument(
+            "Run::getPropertyAsSingleValue - Property\"" + name +
+            "\"cannot be converted to a numeric value.");
+      }
```
ianbush: I think you are now missing a test for this second exception branch, should that be checked too? You may also want to check the message given when throwing for the string, I think it is missing some spaces :smile: .

Otherwise all looks good.

tom-perkins: Good point, I've added a test for both exception branches now (and fixed the spaces!)

ianbush:
Looks good, with unit tests in place for testing the behaviour :shipit:.

PR #17111, opened 2016-08-02
Update weights used during CorelliCrossCorrelate

### LINKED ISSUE #11381: Documentation for CORELLICrossCorelate
This issue was originally [TRAC 10539](http://trac.mantidproject.org/mantid/ticket/10539)

Need some formulas about the way the weight is calculated

rosswhitfield:
This is done at the request of the Corelli instrument scientists to be better inline with what the users expect the algorithm to do. This simply changes the scale of the weight by a factor of the duty cycle (c≈0.498).

**To test:**
- Load in a file from Corelli and apply the CorelliCrossCorrelation, compared the scale before and after this change, they should differ by a factor of ≈0.498.

Release notes have been updated!!!

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Algorithms/src/CorelliCrossCorrelate.cpp
+++ b/Framework/Algorithms/src/CorelliCrossCorrelate.cpp
@@ -131,10 +131,9 @@ void CorelliCrossCorrelate::exec() {
 
   // Calculate the duty cycle and the event weights from the duty cycle.
   double dutyCycle = totalOpen / sequence.back();
-  float weightTransparent = static_cast<float>(1.0 / dutyCycle);
-  float weightAbsorbing = static_cast<float>(-1.0 / (1.0 - dutyCycle));
+  float weightAbsorbing = static_cast<float>(-dutyCycle / (1.0 - dutyCycle));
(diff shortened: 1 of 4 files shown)


Framework/Algorithms/src/CorelliCrossCorrelate.cpp
```diff
@@ -131,8 +131,8 @@ void CorelliCrossCorrelate::exec() {
 
   // Calculate the duty cycle and the event weights from the duty cycle.
   double dutyCycle = totalOpen / sequence.back();
-  float weightTransparent = static_cast<float>(1.0 / dutyCycle);
-  float weightAbsorbing = static_cast<float>(-1.0 / (1.0 - dutyCycle));
+  float weightTransparent = 1.0;
```
AndreiSavici: For all the times you use weightTransparent you could use a string with value "42", and it would not change anything. You just use it on line 137 to print it.

rosswhitfield: It was used but not anymore. Does it matter?

AndreiSavici:
:shipit:

PR #17121, opened 2016-08-03
Framework/Algorithms/L-M HistogramData Rollout

### LINKED ISSUE #17037: HistogramData Rollout Framework/Algorithm L-M
HistogramData rollout based on new Histrogram interface introduced to mantid. This issue deals with files in Framework/Algorithm starting with the letters L-M and their associated unit tests. See here.

Files to be checked (and their associated unit tests):
Logarithm
LorentzCorrection
MaskBins
MaskBinsFromTable
MaskDetectorsIf
MatrixWorkspaceAccess
Max
MaxEnt
MaxMin
MedianDetectorTest
MergeRuns
Min
Minus
ModeratorTzero
ModeratorTzeroLinear
MonitorEfficiencyCorUser
MonteCarloAbsorption
MultipleScatteringCylinderAbsorption
Multiply
MultiplyRange
MuonGroupDetectors

LamarMoore:
Framework/Algorithms files starting with L-M modified to use HistogramData types. 

**To test:**
- Code Review
- Ensure Unit tests pass and systems tests

Fixes #17037 .

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/Algorithms/src/LorentzCorrection.cpp
```diff
@@ -73,10 +73,10 @@ void LorentzCorrection::exec() {
   for (int64_t i = 0; i < int64_t(numHistos); ++i) {
     PARALLEL_START_INTERUPT_REGION
 
-    const MantidVec &inY = inWS->readY(i);
+    auto &inY = inWS->y(i);
 
-    MantidVec &outY = outWS->dataY(i);
-    MantidVec &outE = outWS->dataE(i);
+    auto &outY = outWS->mutableY(i);
+    auto &outE = outWS->mutableE(i);
```
SimonHeybrock: This block should probably be moved below the `try/catch` block below. That way we avoid triggering a copy in case of an exception.

Framework/Algorithms/src/MaskBins.cpp
```diff
@@ -139,7 +141,7 @@ void MaskBins::exec() {
       else
         wi = i;
 
-      const MantidVec &X = outputWS->readX(wi);
+      auto X = outputWS->binEdges(wi);
```
SimonHeybrock: Move into `if` body below? Not used outside?

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -342,13 +343,13 @@ void MaxEnt::exec() {
...
-    }
+    std::iota(outEvolChi->mutableX(s).begin(),
+              outEvolChi->mutableX(s).begin() + niter, 0.0);
+    std::iota(outEvolTest->mutableX(s).begin(),
+              outEvolTest->mutableX(s).begin() + niter, 0.0);
     // Y values
-    outEvolChi->dataY(s).assign(evolChi.begin(), evolChi.end());
-    outEvolTest->dataY(s).assign(evolTest.begin(), evolTest.end());
+    outEvolChi->mutableY(s).assign(evolChi.begin(), evolChi.end());
+    outEvolTest->mutableY(s).assign(evolTest.begin(), evolTest.end());
```
SimonHeybrock: Why not `outEvolChi->setCounts(s, std::move(evolChi))` (etc.)?

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -342,13 +343,13 @@ void MaxEnt::exec() {
...
     // Populate workspaces recording the evolution of Chi and Test
     // X values
-    for (size_t it = 0; it < niter; it++) {
-      outEvolChi->dataX(s)[it] = static_cast<double>(it);
-      outEvolTest->dataX(s)[it] = static_cast<double>(it);
-    }
+    std::iota(outEvolChi->mutableX(s).begin(),
+              outEvolChi->mutableX(s).begin() + niter, 0.0);
+    std::iota(outEvolTest->mutableX(s).begin(),
+              outEvolTest->mutableX(s).begin() + niter, 0.0);
```
SimonHeybrock: X seems to be the same for all spectra, share? And use `LinearGenerator`.

Framework/Algorithms/src/ModeratorTzero.cpp
```diff
@@ -133,8 +133,8 @@ void ModeratorTzero::exec() {
   // iterate over the spectra
   for (int i = 0; i < static_cast<int>(numHists); ++i) {
     PARALLEL_START_INTERUPT_REGION
-    MantidVec &inbins = inputWS->dataX(i);
-    MantidVec &outbins = outputWS->dataX(i);
+    auto &inbins = inputWS->mutableX(i);
```
SimonHeybrock: ?

Framework/Algorithms/src/ModeratorTzero.cpp
```diff
@@ -133,8 +133,8 @@ void ModeratorTzero::exec() {
...
+    auto &outbins = outputWS->mutableX(i);
```
SimonHeybrock: I think we can do better here: Look at all the code blocks below, there are several places that set `outbins = inbins`, i.e., we can set up sharing. Probably just moving variables closer to where they are used will show the right way to do this.

Framework/Algorithms/src/ModeratorTzero.cpp
```diff
@@ -218,8 +218,8 @@ void ModeratorTzero::exec() {
     }   // end of if(L2 >= 0)
 
     // Copy y and e data
-    outputWS->dataY(i) = inputWS->dataY(i);
-    outputWS->dataE(i) = inputWS->dataE(i);
+    outputWS->setSharedY(i, inputWS->sharedY(i));
+    outputWS->setSharedE(i, inputWS->sharedE(i));
```
SimonHeybrock: It would be nice to track the performance of algorithms that have significantly improved sharing. That would probably also make these often rather boring refactoring more  rewarding...? If you don't want to do it for all algorithms that are changed, maybe pick those that have a significant number of uses in the algorithm-usage-statistics?

Framework/Algorithms/src/ModeratorTzero.cpp
```diff
@@ -371,13 +371,10 @@ void ModeratorTzero::execEvent(const std::string &emode) {
...
           evlist.setTofs(tofs);
           evlist.setSortOrder(Mantid::DataObjects::EventSortType::UNSORTED);
-
-          MantidVec tofs_b = evlist.getTofs();
-          MantidVec xarray = evlist.readX();
         } // end of else if(emode=="Elastic")
         else if (emode == "Direct") {
           // fix the histogram bins
-          MantidVec &x = evlist.dataX();
+          auto &x = evlist.mutableX();
```
SimonHeybrock: ` += -1.0 * t0_direct`?

Framework/Algorithms/src/ModeratorTzeroLinear.cpp
```diff
@@ -144,17 +144,17 @@ void ModeratorTzeroLinear::exec() {
     {
       const double scaling = L_i / (L_i + m_gradient);
       const double offset = (1 - scaling) * t_f - scaling * m_intercept;
-      const MantidVec &inbins = inputWS->readX(i);
-      MantidVec &outbins = outputWS->dataX(i);
+      auto &inbins = inputWS->x(i);
+      auto &outbins = outputWS->mutableX(i);
       for (unsigned int j = 0; j < inbins.size(); j++) {
         outbins[j] = scaling * inbins[j] + offset;
```
SimonHeybrock: `outbins = scaling * inbins + offset` would work, but is in theory slightly slower. Would be interesting to test the performance?

Framework/Algorithms/src/ModeratorTzeroLinear.cpp
```diff
@@ -144,17 +144,17 @@ void ModeratorTzeroLinear::exec() {
...
       }
     } else {
-      outputWS->dataX(i) = inputWS->readX(i);
+      outputWS->setSharedX(i, inputWS->sharedX(i));
     }
     // Copy y and e data
-    outputWS->dataY(i) = inputWS->readY(i);
-    outputWS->dataE(i) = inputWS->readE(i);
+    outputWS->setSharedY(i, inputWS->sharedY(i));
+    outputWS->setSharedE(i, inputWS->sharedE(i));
```
SimonHeybrock: Again, performance results would be cool.

Framework/Algorithms/src/MonitorEfficiencyCorUser.cpp
```diff
@@ -97,8 +100,8 @@ void MonitorEfficiencyCorUser::exec() {
  * @param eOut corrected spectrum errors
  */
 void MonitorEfficiencyCorUser::applyMonEfficiency(
-    const size_t numberOfChannels, const MantidVec &yIn, const MantidVec &eIn,
-    const double effVec, MantidVec &yOut, MantidVec &eOut) {
+    const size_t numberOfChannels, const HistogramY &yIn, const HistogramE &eIn,
+    const double effVec, HistogramY &yOut, HistogramE &eOut) {
 
   for (unsigned int j = 0; j < numberOfChannels; ++j) {
     yOut[j] = yIn[j] / effVec;
```
SimonHeybrock: This could be made much more efficient by computing `1 / effVec` in advance (why is it called `Vec`??). And actually this is just `operator*=` (you can also implement `operator/=` could be useful.

And we can just go one step further, and finally implement arithmetics for `Histogram`, then we can completely remove this function and the whole `yOut`, `yIn`, ... bla bla above:

```cpp
Histogram operator*(Histogram lhs, const double rhs) {
  lhs.mutableY() *= rhs;
  lhs.mutableE() *= rhs;
  return lhs;
}
```

Then the whole bunch of code in `MonitorEfficiencyCorUser` can be replaced by:

```cpp
m_outputWS->setHistogram(i, m_inputWS->histogram(i) * (1.0/eff0));
```

I think.

SimonHeybrock: And again, you can also provide `operator/` for convenience.

Framework/Algorithms/src/MonteCarloAbsorption.cpp
```diff
@@ -42,12 +45,12 @@ inline double toWavelength(double energy) {
...
 /// Get ith wavelength point for point data
 /// Assumes all checks on sizes have been done before calling
-double getWavelengthPointData(int i, const std::vector<double> &lambdas) {
+double getWavelengthPointData(int i, const HistogramX &lambdas) {
   return lambdas[i];
 }
 
 /// Get ith wavelength point for histogram data
-double getWavelengthHistogramData(int i, const std::vector<double> &lambdas) {
+double getWavelengthHistogramData(int i, const HistogramX &lambdas) {
```
SimonHeybrock: You can get rid of these two functions and the whole mess with `LambdaPointProvider lambda` below, I think.

Framework/Algorithms/src/MonteCarloAbsorption.cpp
```diff
@@ -175,12 +178,12 @@ MonteCarloAbsorption::doSimulation(const MatrixWorkspace &inputWS,
   for (int64_t i = 0; i < nhists; ++i) {
     PARALLEL_START_INTERUPT_REGION
 
-    const auto &xvalues = outputWS->readX(i);
-    auto &signal = outputWS->dataY(i);
-    auto &errors = outputWS->dataE(i);
+    auto &xvalues = outputWS->x(i);
```
SimonHeybrock: As I said above, just use `points()` and use that to obtain `lambdaStep` below.

SimonHeybrock: It seems to be skipping bins though, maybe you should first figure out if this is performance relevant.

Framework/Algorithms/src/MultipleScatteringCylinderAbsorption.cpp
```diff
@@ -182,13 +185,17 @@ void MultipleScatteringCylinderAbsorption::exec() {
...
       eventList.getWeightErrors(err_vec);
 
-      apply_msa_correction(tth_rad, radius, coeff1, coeff2, coeff3, tof_vec,
-                           y_vec, err_vec);
+      HistogramX tof(std::move(tof_vec));
+      HistogramY y(std::move(y_vec));
+      HistogramE err(std::move(err_vec));
+
+      apply_msa_correction(tth_rad, radius, coeff1, coeff2, coeff3, tof, y,
+                           err);
```
SimonHeybrock: These vectors are *not* X, Y, and E. They have nothing to do with histograms, I think we should keep the old code here.

Framework/Algorithms/src/MultipleScatteringCylinderAbsorption.cpp
```diff
@@ -201,7 +208,7 @@ void MultipleScatteringCylinderAbsorption::exec() {
   {
     // Create the new workspace
     MatrixWorkspace_sptr out_WS = WorkspaceFactory::Instance().create(
-        in_WS, NUM_HIST, in_WS->readX(0).size(), in_WS->readY(0).size());
+        in_WS, NUM_HIST, in_WS->x(0).size(), in_WS->y(0).size());
```
SimonHeybrock: I think these are all just default values and can probably be omitted?

Framework/Algorithms/src/MultipleScatteringCylinderAbsorption.cpp
```diff
@@ -312,8 +314,8 @@ double calculate_msa_factor(const double radius, const double Q2,
  */
 void MultipleScatteringCylinderAbsorption::apply_msa_correction(
     const double angle_deg, const double radius, const double coeff1,
-    const double coeff2, const double coeff3, const vector<double> &wavelength,
-    vector<double> &y_val, std::vector<double> &errors) {
+    const double coeff2, const double coeff3, const HistogramX &wavelength,
+    HistogramY &y_val, HistogramE &errors) {
```
SimonHeybrock: - Use `Histogram` as argument (and return)?
- Use `points()` below.

Framework/Algorithms/src/MuonGroupDetectors.cpp
```diff
@@ -97,14 +97,15 @@ void MuonGroupDetectors::exec() {
...
     for (auto &wsIndex : wsIndices) {
       for (size_t i = 0; i < inWS->blocksize(); ++i) {
         // Sum the y values
-        outWS->dataY(groupIndex)[i] += inWS->dataY(wsIndex)[i];
-
+        Y[i] += inWS->y(wsIndex)[i];
         // Sum the errors in quadrature
-        outWS->dataE(groupIndex)[i] = sqrt(pow(outWS->dataE(groupIndex)[i], 2) +
-                                           pow(inWS->dataE(wsIndex)[i], 2));
+        E[i] = sqrt(pow(E[i], 2) + pow(inWS->e(wsIndex)[i], 2));
```
SimonHeybrock: Don't do this by hand. If you can figure out whether the workspaces store counts or frequencies, you can use, e.g., `countVariances()` to do the squaring for you, and the respective setter will to the `sqrt`.

SimonHeybrock: But maybe it is time to also implement `operator+=` for `Histogram`?

Framework/Algorithms/test/LorentzCorrectionTest.h
```diff
@@ -68,12 +68,9 @@ class LorentzCorrectionTest : public CxxTest::TestSuite {
     auto workspace = Create2DWorkspaceBinned(nSpectra, nBins, startX,
                                              deltaX); // Creates histogram data
 
-    auto &ydata = workspace->dataY(0);
-    auto &edata = workspace->dataE(0);
-    for (size_t i = 0; i < ydata.size(); ++i) {
-      ydata[i] = 1;
-      edata[i] = 1;
-    }
+    auto size = workspace->mutableY(0).size();
```
SimonHeybrock: `y`

Framework/Algorithms/test/LorentzCorrectionTest.h
```diff
@@ -68,12 +68,9 @@ class LorentzCorrectionTest : public CxxTest::TestSuite {
...
+    workspace->mutableY(0).assign(size, 1.0);
+    workspace->mutableE(0).assign(size, 1.0);
```
SimonHeybrock: How about adding `operator=(const double)`? You can do this via a new mixin (similar to `Scalable` and `Offsetable`)? This would be useful in many places, I think.

Framework/Algorithms/test/MaxEntTest.h
```diff
@@ -516,43 +519,43 @@ class MaxEntTest : public CxxTest::TestSuite {
...
+        CountStandardDeviations{0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1,
+                                0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1,
+                                0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1,
+                                0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1,
+                                0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1,
+                                0.1, 0.1, 0.1, 0.1, 0.1, 0.1});
+
+    // Imaginary
+    ws->setSharedX(1, ws->sharedX(0));
+    ws->setSharedE(1, ws->sharedE(0));
```
SimonHeybrock: You can save one more line by using `ws->setHistogram` instead.

Framework/Algorithms/test/ModeratorTzeroTest.h
```diff
@@ -244,9 +244,11 @@ class ModeratorTzeroTest : public CxxTest::TestSuite {
         Mantid::Kernel::UnitFactory::Instance().create("TOF");
     BinEdges xdata(numBins + 1, LinearGenerator(0.0, 4.0));
     const double peakHeight(1000.), peakCentre(7000.), sigmaSq(1000 * 1000.);
+
+    auto &Y = testWS->mutableY(0);
     // tof ranges from 0 to 16000 (units assumed micro-seconds
     for (int ibin = 0; ibin < numBins; ++ibin) {
-      testWS->dataY(0)[ibin] =
+      Y[ibin] =
           peakHeight * exp(-0.5 * pow(xdata[ibin] - peakCentre, 2.) / sigmaSq);
```
SimonHeybrock: It must be the 10th time I see this gaussian somehwhere, maybe we should add a `GaussianGenerator`, or some other way to generator this?

Framework/Algorithms/src/MuonGroupDetectors.cpp
```diff
@@ -98,17 +99,13 @@ void MuonGroupDetectors::exec() {
...
-      for (size_t i = 0; i < inWS->blocksize(); ++i) {
-        // Sum the y values
-        Y[i] += inWS->y(wsIndex)[i];
-        // Sum the errors in quadrature
-        E[i] = sqrt(pow(E[i], 2) + pow(inWS->e(wsIndex)[i], 2));
-      }
-
-      // Detectors list of the group should contain all the detectors of it's
+	  outWS->mutableY(groupIndex) += inWS->y(wsIndex);
+	  inWS->mutableE(groupIndex) += inWS->e(wsIndex);
```
SimonHeybrock: `HistogramE` does not add the squares automatically! Add the histograms instead!

LamarMoore: Seems like this part has to be reverted to what was originally there. The algorithm unit test `PlotAssymetryByLogValueTest` is expecting to be able to use `MuonGroupDetectors` and add the Y and E values even though the X values are not the same...

SimonHeybrock:
@LamarMoore I don't quite remember, where there more changes to make here, or should I give it a final review pass?

Framework/Algorithms/src/MonteCarloAbsorption.cpp
```diff
@@ -163,24 +155,17 @@ MonteCarloAbsorption::doSimulation(const MatrixWorkspace &inputWS,
...
-    const auto &xvalues = outputWS->readX(i);
-    auto &signal = outputWS->dataY(i);
-    auto &errors = outputWS->dataE(i);
+    auto xvalues = outputWS->points(i);
+    auto &signal = outputWS->mutableY(i);
+    auto &errors = outputWS->mutableE(i);
     // The input was cloned so clear the errors out
     // Y values are all overwritten later
-    std::fill(errors.begin(), errors.end(), 0.0);
+    errors.assign(errors.size(), 0.0);
```
SimonHeybrock: `errors = 0.0;` (I think that PR has been merged, right?)

Framework/Algorithms/src/MultipleScatteringCylinderAbsorption.cpp
SimonHeybrock: @LamarMoore I think you overlooked this comment, could you clean this up as suggested (similar to how you did somewhere above, returning `Histogram` by value).

Framework/Algorithms/test/LorentzCorrectionTest.h
```diff
@@ -111,8 +106,8 @@ class LorentzCorrectionTest : public CxxTest::TestSuite {
 
   void test_throws_if_wavelength_zero() {
     auto ws_lam = this->create_workspace(2 /*nBins*/);
-    ws_lam->dataX(0)[0] = 0; // Make wavelength zero
-    ws_lam->dataX(0)[1] = 0; // Make wavelength zero
+    ws_lam->mutableX(0)[0] = 0; // Make wavelength zero
+    ws_lam->mutableX(0)[1] = 0; // Make wavelength zero
```
SimonHeybrock: `ws_lam->mutableX(0) = 0.0;`

Framework/Algorithms/test/MergeRunsTest.h
```diff
@@ -645,7 +645,7 @@ class MergeRunsTest : public CxxTest::TestSuite {
     TS_ASSERT(!merge2.isExecuted());
     MatrixWorkspace_sptr badIn =
         WorkspaceCreationHelper::Create2DWorkspace123(3, 10, 1);
-    badIn->dataX(0) = std::vector<double>(11, 2.0);
+    badIn->mutableX(0).assign(11, 2.0);
```
SimonHeybrock: `=`

Framework/Algorithms/test/ModeratorTzeroLinearTest.h
```diff
@@ -153,40 +178,63 @@ class ModeratorTzeroLinearTest : public CxxTest::TestSuite {
     for (size_t ihist = 0; ihist < numHists; ++ihist) {
       EventList &evlist = testWS->getSpectrum(ihist);
       BinEdges xdata(numBins + 1, LinearGenerator(0.0, 4.0));
-      for (int ibin = 0; ibin <= numBins; ++ibin) {
-        double tof = 4 * ibin;
-        TofEvent tofevent(tof);
-        evlist.addEventQuickly(tofevent); // insert event
+      for (auto &tof : xdata) {
+        evlist.addEventQuickly(TofEvent(tof));
```
SimonHeybrock: Missing scale factor?

Framework/HistogramData/test/HistogramTest.h
```diff
@@ -932,4 +934,37 @@ class HistogramTest : public CxxTest::TestSuite {
...
+  static HistogramTestPerformance *createSuite() {
+    return new HistogramTestPerformance;
+  }
+  static void destroySuite(HistogramTestPerformance *suite) { delete suite; }
+
+  HistogramTestPerformance() : copy(histSize, LinearGenerator(0, 2)) {
+    for (size_t i = 0; i < nHists; i++)
+      hists.push_back(Histogram(BinEdges(histSize, LinearGenerator(0, 1))));
+
+    histsCopy = hists;
```
SimonHeybrock: Maybe add some comment that explains this. If I remember correctly it was to avoid benchmarking de-allocation?

SimonHeybrock:
Looks good apart from the minor comments above. Have you recording before/after results for the new performance tests that you have added?

LamarMoore:
@SimonHeybrock  yes I have. Will try to get these changes in by tomorrow.

SimonHeybrock:
Latest changes look good. :shipit: once Jenkins gives his ok.

LamarMoore:
@SimonHeybrock  In places where sharing was improved within the code base. I ones worth mentioning are:
ModeratorTZero - 29% faster execution
ModeratorTZeroLinear - 38% faster execution

SimonHeybrock:
@LamarMoore Right, I forgot about that. Could you add all those results to the release notes?

SimonHeybrock:
@LamarMoore What about `MaxEntTestPerformance`, `MonitorEfficiencyCorUserTestPerformance`, and `MonteCarloAbsorptionTestPerformance`, which were also added in this PR?

LamarMoore:
@SimonHeybrock The performance improvements were either very slight or about the same as before so I did not mention them.

SimonHeybrock:
@LamarMoore The performance tests added here seem to make trouble: http://builds.mantidproject.org/job/master_performancetests/3442/console. Had a quick look, but cannot see a reason right now.

SimonHeybrock:
@LamarMoore I think you are simply running out of memory (using more than 24 GB, if my early-morning math is right). Also, the test is taking a bit too long.

PR #17123, opened 2016-08-03
Initial steps for improved index handling in MatrixWorkspace

### LINKED ISSUE #17114: Refactor initialization of MatrixWorkspaces with spectrum numbers and detector IDs.
This is a prerequisite for #16810 and is discussed as part of the design document https://github.com/mantidproject/documents/pull/13.

The goal is to make workspace creation more atomic and reduce the number of violations of the law of Demeter, which is causing trouble in planning the implementation of improved indexing and Instrument 2.0.

SimonHeybrock:
A few examples of how `MatrixWorkspace::setIndexInfo` could replace `ISpectrum::setSpectrumNo` and `ISpectrum::setDetectorIDs`.

Notable changes:
- Improved factory functions for MatrixWorkspaces available in `MantidDataObjects/WorkspaceCreation.h`.
- `MatrixWorkspace::cloneEmpty` with covariant return type, basically a polymorphic default constructor, mainly as helper for new factory.
- Added `IndexInfo` object. This will eventually replace direct access to spectrum numbers and detector IDs via `ISpectrum`. Currently this acts as a wrapper that just forwards to data stored in `MatrixWorkspace`, but can also exist as a stand-alone object for basic manipulation. Apart from manual initialization of spectrum numbers and detector IDs, `extract` and `group` functions are available, for common spectrum manipulation functionality used by several algorithms.

I am not entirely happy with all aspects of this implementation, but for now I did not see a nicer way. I'd be happy to hear suggestions for improvements!

What is not included in this initial implementation, but will be added in follow-up stages:
- MPI support. `IndexInfo` will internally deal with most of the index translation and other messy things that a required when partitioning workspaces for an MPI run of Mantid.
- Validation of indices: We will want to forbid duplicate spectrum numbers and  detector IDs that are not part of the instrument. Both will be handled internally in `IndexInfo` or in combination with changes in Instrument-2.0.

Fixes #17114.

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -83,6 +84,31 @@ MatrixWorkspace::~MatrixWorkspace() {
   }
 }
 
+void MatrixWorkspace::setIndexTranslator(
```
OwenArnold: Is there a good reason for having this method? It's going to make the `MatrixWorkspace` interface sensitive to any future changes we might want to make to the `IndexTranslator` constructor.

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -83,6 +84,31 @@ MatrixWorkspace::~MatrixWorkspace() {
...
+  auto &spectrumNumbers = translator.spectrumNumbers();
+  auto &detectorIDs = translator.detectorIDs();
+  for (size_t i = 0; i < getNumberHistograms(); ++i) {
+    auto &spectrum = getSpectrum(i);
+    spectrum.setSpectrumNo(spectrumNumbers[i]);
+    if (!detectorIDs.empty())
+      spectrum.setDetectorIDs(
+          std::set<detid_t>(detectorIDs[i].begin(), detectorIDs[i].end()));
+  }
+}
```
OwenArnold: For some reason I thought that we might keep the translator. What is the default for spectrum numbers and detector ids prior to this method being called?

Framework/Indexing/test/DetectorIDsTest.h
```diff
@@ -0,0 +1,37 @@
...
+  // This pair of boilerplate methods prevent the suite being created statically
+  // This means the constructor isn't called when running other tests
+  static DetectorIDsTest *createSuite() { return new DetectorIDsTest(); }
+  static void destroySuite(DetectorIDsTest *suite) { delete suite; }
+
+  void test_constructor() { TS_ASSERT_THROWS_NOTHING((DetectorIDs{1, 2, 3})); }
+
+  void test_size() { TS_ASSERT_EQUALS(DetectorIDs({1, 2, 3}).size(), 3); }
+
+  void test_detectorIDs() {
```
OwenArnold: You're testing de-duplication here. maybe that should be in the test name.

OwenArnold:
@SimonHeybrock. Apart from my minor points, this looks OK to me.

Framework/API/src/MatrixWorkspace.cpp
SimonHeybrock: I guess you are right. I had also considered having
```cpp
template <class... T>
void setIndexTranslator(T &&... args);
```
and then forward to the constructor. However, that would imply that we need to include `IndexTranslator.h` in `MatrixWorkspace.h`. What is your opinion on this?

Framework/API/src/MatrixWorkspace.cpp
SimonHeybrock: Keeping it will be a second step. If we would keep it right now, we would have two locations where spectrum numbers and detector IDs are stored. As long as `ISpectrum::setSpectrumNo` and `ISpectrum::setDetectorIDs` are still available we cannot keep the translator, since the information in the translator will run out of sync with the information in the spectra.

SimonHeybrock: Basically what I am doing so far is a first step of refactoring, with the goal to move all `setSpectrumNo` and `setDetectorID` calls into the body of `setIndexTranslator`. Once that is done we can remove to respective members (`ISpectrum::m_specNo`, ...) and instead store the translator.

SimonHeybrock: Regarding default: That is being set by the `init` methods of the respective workspace types. Eventually that will be refactored to a `setIndexTranslator` call.

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -83,6 +84,31 @@ MatrixWorkspace::~MatrixWorkspace() {
...
+    throw std::runtime_error("MatrixWorkspace::setIndexTranslator: Translator "
+                             "size does not match number of histograms in "
+                             "workspace");
+
+  auto &spectrumNumbers = translator.spectrumNumbers();
+  auto &detectorIDs = translator.detectorIDs();
+  for (size_t i = 0; i < getNumberHistograms(); ++i) {
+    auto &spectrum = getSpectrum(i);
+    spectrum.setSpectrumNo(spectrumNumbers[i]);
+    if (!detectorIDs.empty())
```
SimonHeybrock: @OwenArnold This is a point that I am not quite happy with: In many places algorithms only update the spectrum numbers, keeping the detector IDs as they were. Here I solve this be setting n `IndexTranslator` with an empty vector of detector IDs. However, it is a bit weird to let `setIndexTranslator` overwrite only parts of the current indices.

What do you think? Is there a way to make this more explicit?
- Should there be two routines, `setSpectrumNumberTranslator` and `setDetectorIDTranslator`?
- Should we just explicitly specify it in the `IndexTranslator` constructor, with some sort of flag (`DetectorIDs::KeepCurrent`)?

Framework/API/src/MatrixWorkspace.cpp
OwenArnold: Yes, as you say, this is also a possible and good option for us. I think that forward declarations rather than includes are probably the better option when it comes to such a well used interfaces as `MatrixWorkspace`. I think we can live with having both signatures and maybe refactoring the one above out as part of the second step?

Framework/API/src/MatrixWorkspace.cpp
OwenArnold: I quite like the second explicit option that you've suggested with the `DetectorIDs::KeepCurrent` flag.

Framework/API/src/MatrixWorkspace.cpp
SimonHeybrock: I think we should decide this based on my question below, regarding cases where only spectrum numbers and not detector IDs get updated.

Framework/API/src/MatrixWorkspace.cpp
SimonHeybrock: @OwenArnold Example for the first: Remember the `Indexing` prototyping I did, there was a separate `SpectrumNumberTranslator` and `DetectorIDTranslator`. We could then have

```cpp
MatrixWorkspace::setSpectrumNumberTranslator(...);
MatrixWorkspace::setDetectorIDTranslator(...);

// update spectrum numbers in workspace
ws.setSpectrumNumberTranslator(SpectrumNumbers{1,2,3,4,5});

// update detector IDs in workspace
ws.setDetectorIDs(DetectorIDs{100,101,102,103,105});
```

Basically there would be two independent setters, and probably also two independent translator objects.

OwenArnold: @SimonHeybrock. I'm not sure about this option. I quite liked the idea about having a single `IndexTranslator` that could manage all self-consistency internally. 

Somehow I think that what you really want to do from the algorithm perspective is:

1. Take the existing `IndexTranslator` on the `InputWorkspace`. Possibly by cloning the input workspace to make your `OutputWorkspace`.
2. Overwrite just the spectrum numbers on the `IndexTranslator` on the `OutputWorkspace`. This should invoke COW. but we would keep the `DetectorID` parts.

This supposes that the `IndexTranslator` is carried by the `MatrixWorkspace`. So this is something that you've already mentioned you'd only do in a later refactoring step. It then requires (unless you use a `smart_ptr`) that you are going to include `IndexTranslator` in `MatrixWorkspace.h`, so (as an aside) with your previous question you may as well have:
```
template <class... T>
void setIndexTranslator(T &&... args);
```

SimonHeybrock: @OwenArnold So if we want to change only the spectrum numbers and want to avoid partially initialized translator objects, we could do:

```cpp
auto translator = ws.translator; // return by value, cow_ptr inside?
translator.setSpectrumNumbers({1,2,3,4,5});
ws.setTranslator(translator);
```

OwenArnold: @SimonHeybrock I prefer this. It allows us to just take the `DetectorIndexes` from the original.

SimonHeybrock:
@OwenArnold I implemented the changes we discussed. I think functionality-wise this will allow us to do everything we need. In particular, returning the `IndexTranslator` by-value will permit changing the partitioning, since the workspace can inspect the new translator object when setting it:

```cpp
auto translator = ws.indexTranslator();
translator.setPartioning(MyNewPartitioning());
ws.setIndexTranslator(translator);

// MatrixWorkspace.cpp:
void MatrixWorkspace::setIndexTranslator(const IndexTranslator &translator) {
  if(translator.partitioning() != m_translator.partitioning())
    this->repartition(translator); // updates SpectrumInfo, moves spectrum data to other ranks
  else
    m_translator = translator;
}
```

The main downside I see currently is that the new interface is quite a bit more verbose in some cases, adding a few lines of code around every current use of `ISpectrum::setSpectrumNo()`. This might be alleviated by improved factory methods, at least in some cases.

Comments would be appreciated.

SimonHeybrock:
@OwenArnold There is also one open issue I am not sure how to solve in a nice way: `EventList::operator+=`, which is used in quite a few places, also merges the sets of detector IDs. This cannot be done on that level anymore, since the detector IDs will not be part of `ISpectrum`.

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -83,6 +84,41 @@ MatrixWorkspace::~MatrixWorkspace() {
...
+  t.setSpectrumNumbers(std::move(specNums));
+  t.setDetectorIDs(std::move(detIDs));
+  return t;
+}
+
+void MatrixWorkspace::setIndexTranslator(
+    const Indexing::IndexTranslator &translator) {
+  // Comparing the *local* size of the translator.
+  if (translator.size() != getNumberHistograms())
+    throw std::runtime_error("MatrixWorkspace::setIndexTranslator: Translator "
```
OwenArnold: Rather than throwing if setIndexTranslator fails. What about returning a boolean? Does it always make sense to throw if the translator cannot be set? 

I guess throwing does seem to be the policy already with the `MatrixWorkspace`.

Framework/Algorithms/src/DiffractionFocussing.cpp
```diff
@@ -133,14 +134,17 @@ void DiffractionFocussing::exec() {
...
     int64_t i = resultIndeces[hist];
     outputW->setHistogram(hist, tmpW->histogram(i));
-    auto &inSpec = tmpW->getSpectrum(i);
-    outputW->getSpectrum(hist).setSpectrumNo(inSpec.getSpectrumNo());
-    inSpec.setSpectrumNo(-1);
+    specNums.push_back(tmpTranslator.spectrumNumber(i));
   }
+  auto outputTranslator = outputW->indexTranslator();
+  outputTranslator.setSpectrumNumbers(std::move(specNums));
+  outputW->setIndexTranslator(outputTranslator);
```
OwenArnold: The semantics of this look right to me. I can't see how one might accidentally get this wrong.

Framework/API/src/MatrixWorkspace.cpp
SimonHeybrock: There should rarely be a size mismatch. Once an algorithm is implemented correctly this should not happen anymore. I am not sure what a boolean return type could be used for?

OwenArnold: You where thinking about checking `partitioning`. But maybe it always makes sense to throw.

SimonHeybrock: Well, the partitioning is allowed to be changed, I think (unless we figure out another way for dealing with repartitioning, e.g., by simply creating a new workspace if we need a new partitioning), so there would be no error in that case.

SimonHeybrock:
@OwenArnold There is another major question that needs resolving in order to make this compatible with MPI:

There is quite a large number of algorithms that does one or more of the following with a workspace:

1. Reorder spectra (`ConvertSpectrumAxis`)
2. Take only some spectra from a workspace to create a new workspace (`ExtractSpectra`)
3. Group detectors differently (`SumSpectra`, `GroupDetectors`, `DiffractionFocussing`)

In any of these cases the partition a spectrum belongs to may change:

- If we use `GlobalWorkspaceIndex` to define partitioning, partitions can change in any of the three cases.
- If we use `SpectrumNumber` to define partitioning, partitions stay the same in case 1., but change in 2. and 3. (they can change in 2. since the total number of spectra can affect the target partition, even if the spectrum numbers themselves are unchanged).

Let us now consider this in combination with the `IndexTranslator`. We would like to do the following:

```cpp
const auto translator = inputWS.indexTranslator();
// getSpectrumNumberProperty() gives algorithm property
auto indices = translator.makeSpectrumIndexSet(getSpectrumNumberProperty());
const auto newTranslator = extractSubset(translator, indices);
auto outputWS = WorkspaceFactory::create(inputWS, newTranslator);
```

Apart from some tweaks, I think this could work in the non-MPI case. I **think** we can also make it work in the MPI case **if we allow `IndexTranslator` and `MatrixWorkspace` to be in a state where spectra are not distributed as defined by the initial `Partitioning`**.

What does this imply?

- `IndexTranslator` must be the only place that knows about the current `Partitioning`. In particular, any code that needs to determine what partition a spectrum number belongs to must query `IndexTranslator` and **not** an instance of `Partitioning`. This should be no problem, we would want it that way in any case.
- A workspace can get into an "unbalanced" state, e.g., be extracting all spectra on rank 0. We could provide a `rebalance` method or free function. The additional advantage here is that this might reduce the number of unintentional (costly) re-shuffling of data between MPI ranks.
- Operations that have two input workspaces must check that the `IndexTranslator` (inluding `Partitioning`) is the same **and** that the **current** distribution of spectra in those workspaces is the same. Effectively, we could triggers calls to `rebalance` if this is not the case.

Does this make any sense to you? As usual, this also has implications for `SpectrumInfo`, i.e., there might be fewer cases where it needs redistribution.

Maybe this is a longer discussion, worth having next week.

SimonHeybrock:
I overlooked one issue: `extractSubset` cannot determine the global workspace indices for the output translator. :(

Not sure how to solve this. We could either only accept spectrum numbers as input (i.e., information about what to extra for **all* ranks), or we would need to add MPI communication, such that the function can figure out information about other indices itself. The first option is probably better, but would require dealing with spectrum numbers (etc.) in the affected algorithms, instead of just having the `SpectrumIndexSet` abstraction.

OwenArnold:
@SimonHeybrock. Yes we should definitely talk about this more next week.

I think that your conclusions for the first message looked sensible. I'm not clear about your `extractSubset` problem.

Framework/API/inc/MantidAPI/HistoWorkspace.h
```diff
@@ -0,0 +1,54 @@
+#ifndef MANTID_API_HISTOWORKSPACE_H_
+#define MANTID_API_HISTOWORKSPACE_H_
+
+#include "MantidAPI/DllConfig.h"
+#include "MantidAPI/MatrixWorkspace.h"
+
+namespace Mantid {
+namespace API {
+
+/** HistoWorkspace is an abstract base class for MatrixWorkspace types that are
```
OwenArnold: An analogue to `IEventWorkspace`?

Framework/API/inc/MantidAPI/WorkspaceFactory.h
```diff
@@ -77,8 +78,8 @@ class MANTID_API_DLL WorkspaceFactoryImpl
                               const size_t &NVectors, const size_t &XLength,
                               const size_t &YLength) const;
 
-  void initializeFromParent(const MatrixWorkspace_const_sptr parent,
-                            const MatrixWorkspace_sptr child,
+  void initializeFromParent(const MatrixWorkspace &parent,
```
OwenArnold: Good catch changing `parent` to be a `const ref` can we do the same with `child`?

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -153,6 +181,37 @@ void MatrixWorkspace::initialize(const std::size_t &NVectors,
...
+  m_indexCalculator = MatrixWSIndexCalculator(this->blocksize());
+  // Indicate that this workspace has been initialized to prevent duplicate
+  // attempts.
+  m_isInitialized = true;
+}
+
+void MatrixWorkspace::initialize(const Indexing::IndexInfo &indexInfo,
+                                 const HistogramData::Histogram &histogram) {
+  initialize(indexInfo.size(), histogram);
+  setIndexInfo(indexInfo);
```
OwenArnold: Is this the reason we need to have the cache and call_once on the return. i.e. we can't rely on this at construction time?

Framework/Indexing/inc/MantidIndexing/IndexInfo.h
```diff
@@ -0,0 +1,79 @@
...
+
+  specnum_t spectrumNumber(const size_t index) const;
+  std::vector<detid_t> detectorIDs(const size_t index) const;
+
+  void setSpectrumNumbers(std::vector<specnum_t> &&spectrumNumbers) & ;
+  void setDetectorIDs(const std::vector<detid_t> &detectorIDs) & ;
+  void setDetectorIDs(std::vector<std::vector<detid_t>> &&detectorIDs) & ;
+
+private:
+  bool m_isLegacy{false};
```
OwenArnold: Could definitely use some documentation in `IndexInfo` about this `m_isLegacy` . I agree that it is currently needed, but this flag binds together quite a few other methods in the context of the MatrixWorkspace that could use some explanation for future readers.

OwenArnold:
This looks pretty good to me.

Framework/API/inc/MantidAPI/HistoWorkspace.h
SimonHeybrock: Yes, needed for the factory functions. I tried to explain this better in a slightly extended Doxygen.

Framework/API/inc/MantidAPI/WorkspaceFactory.h
SimonHeybrock: Yes. I did not bother with it since `initializeFromParent` is basically made obsolete by the new factory functions. That is, we can either change also `child` (wasting effort), or use the new factory right away. I'll think about what is better now.

Framework/API/src/MatrixWorkspace.cpp
SimonHeybrock: No, the reason was that the size of a workspace is unknown at construction time, and I had some trouble with doing it in `initialize()` (worked in most cases, but there were some odd ones where workspaces were not initialized).
But actually I *can* do it in the constructor, I'll just have to pass a `std::function` also for the size.

Framework/DataObjects/inc/MantidDataObjects/WorkspaceCreation.h
```diff
@@ -0,0 +1,207 @@
...
+  //   supported for EventWorkspace
+  boost::shared_ptr<T> ws;
+  if (std::is_base_of<API::HistoWorkspace, T>::value &&
+      parent.id() == "EventWorkspace") {
+    // Drop events, create Workspace2D or T whichever is more derived.
+    ws = detail::createHelper<T>();
+  } else {
+    try {
+      // If parent is more derived than T: create type(parent)
+      ws = dynamic_cast<const T &>(parent).cloneEmpty();
```
OwenArnold: Not sure that this makes sense with the comment here. I agree with the comment.

"If parent is more derived than T: create type(parent)"

which implies that P is declared something like:

```cpp
class Parent : public T {};
```
But then you are upcasting as part of the `dynamic_cast` to `T&` can calling `cloneEmpty()` on the base class reference.

I think there is mismatch between comment and implementation.

OwenArnold:
@martyngigg. Could you take a look at this PR? It contains some important new features and we'd appreciate a third pair of eyes on the changes.

Framework/DataObjects/inc/MantidDataObjects/WorkspaceCreation.h
SimonHeybrock: Let me give an example:

```cpp
const MatrixWorkspace &parent = ... // say Workspace2D
auto child = create<MatrixWorkspace>(parent);
```

The *dynamic* type of parent is more derived than `T` (which is `MatrixWorkspace`), `cloneEmpty()` will create the correct type.

The other case:

```cpp
const MatrixWorkspace &parent = ... // say Workspace2D
auto child = create<SpecialWorkspace2D>(parent);
```

T is more derived than the parent, so the cast will fail, and we enter the other code branch via `throw` + `catch`. This is not particularly nice, but the alternative would be to attempt some string matching based on the workspace ID, which is also ugly, I think?

OwenArnold: My main problem was with the use of the dynamic_cast, for a seemingly upcast, but reading the code again I see that this is actually required.

Framework/API/src/MatrixWorkspace.cpp
martyngigg: I agree with the decision here too. The concept of mapping from spectrum numbers->detector IDs causes enough confusion that I think being able to explain that it is a single type that defines the mapping is helpful.

martyngigg:
Regarding the `EventList::operator+=(EventList)` problem. This is essentially a remapping of the current translation so can the `IndexInfo` have a `merge`-type method that would accept a list of spectrum numbers to merge together or was the idea to keep `IndexInfo` immutable?

SimonHeybrock:
@martyngigg: Yes, basically the newly added `group` should be used, I think. The main problem with the current `EventList::operator+=` is that it is incremental, which will be very hard to deal with with MPI.

SimonHeybrock:
@martyngigg To elaborate a bit more:
`IndexInfo` itself can be mutable, but modifying the `IndexInfo` that is part of a `MatrixWorkspace` would be difficult. Think of modifications of spectrum numbers or detector IDs for an MPI-based Mantid run:

- We want to validate the spectrum numbers (avoid duplicates), and would need to communicate with all MPI ranks.
- Similarly, we need to verify that detector IDs that are added are part of the instrument.
- Maybe the most costly part would be adding detector IDs to a spectrum. Instrument-2.0 will store information for detectors only on the MPI rank that holds the corresponding spectrum (or spectra). Adding a detector to a spectrum will potentially require removing that detector from the instrument on one rank, send the information to another rank via MPI, and adding the detector to the instrument on that rank.

All this would be possible to implement, but I think @OwenArnold and me agree that this should be avoided for two reasons:

- Runtime cost (performance).
- Development effort.

Basically, what I am saying is that we try to eliminate any incremental modifications of these things. The control flow I have in mind right now is to create a new `IndexInfo` based on an existing `IndexInfo` (that is part of the input workspace). Constructing the new `IndexInfo` can happen incrementally. In a final step it is then used to construct the output workspace.

So in summary I would like to make the indexing in a workspace immutable.

martyngigg:
@SimonHeybrock That make sense. I actually think it will be beneficial for developers to have to think how the mapping will end up looking at the end of a given algorithm rather than "something" just happening under the hood.

At the moment there is also a disparity between `EventList` and `Histogram1D` where you can't sum the latter and have to take care of the mapping changes by hand anyway.

OwenArnold:
I like the semantics of the new design. Soon after merging I think some education for the development community will be needed on how to use these features (as you did for `HistogramData`)

SimonHeybrock:
@OwenArnold I have found (so far) 2 cases where the new factory style is not sufficient: Some algorithms want to force crrating a certain workspace type, even if the parent is more derived. For example, for creating a `Workspace2D`, even if the parent is, e.g., `SpecialWorkspace2D`. Currently this is done in two steps, via `WorkspaceFactory::create("Workspace2D")` (no parent) and `initializeFromParent`.

Any thoughts on how we can do this in the new style?

OwenArnold:
@SimonHeybrock. Can we have some *creational* function variant in which you provide a template argument for the desired object type?

SimonHeybrock:
@OwenArnold Any ideas on how to make a clear / obvious distinction between the two?

SimonHeybrock:
Regarding the comments from 10 days ago: (As discussed on Slack) We will probably add another factory method `createAndInit` for that case.

@OwenArnold I did some minor changes here, I think this should be ready after a final review. As a next step I am considering rolling out the new factory. I hope that will reduce the amount of direct access to spectrum numbers and detector IDs in `ISpectrum`, i.e., gets us closer to the point where we can make `IndexInfo` handle spectrum numbers and detector IDs.

OwenArnold:
👍 I should be able to review this soon.

Framework/Indexing/test/IndexInfoTest.h
```diff
@@ -15,7 +15,31 @@ class IndexInfoTest : public CxxTest::TestSuite {
...
+
+  void test_vector_constructor_sets_correct_indices() {
+    IndexInfo info({3, 2, 1}, {{}, {10}, {20, 30}});
+    TS_ASSERT_EQUALS(info.spectrumNumber(0), 3);
+    TS_ASSERT_EQUALS(info.spectrumNumber(1), 2);
+    TS_ASSERT_EQUALS(info.spectrumNumber(2), 1);
+    TS_ASSERT_EQUALS(info.detectorIDs(0), (std::vector<detid_t>{}));
+    TS_ASSERT_EQUALS(info.detectorIDs(1), (std::vector<detid_t>{10}));
+    TS_ASSERT_EQUALS(info.detectorIDs(2), (std::vector<detid_t>{20, 30}));
+  }
```
OwenArnold: Looks good.

Framework/Indexing/src/Group.cpp
```diff
@@ -0,0 +1,25 @@
+#include "MantidIndexing/Group.h"
+#include "MantidIndexing/IndexInfo.h"
+
+namespace Mantid {
+namespace Indexing {
+
+IndexInfo group(const IndexInfo &source, std::vector<specnum_t> &&specNums,
```
OwenArnold: Doxygen for this function would be useful.

Framework/Indexing/inc/MantidIndexing/MakeRange.h
```diff
@@ -0,0 +1,47 @@
...
+
+#include "MantidIndexing/DllConfig.h"
+
+#include <numeric>
+#include <type_traits>
+
+namespace Mantid {
+namespace Indexing {
+
+/** MakeRange : TODO: DESCRIPTION
```
OwenArnold: Description missing

OwenArnold:
@SimonHeybrock. Sorry I've let this get stale. Once the merge conflicts have been fixed, is there anything that would stop us shipping this?

SimonHeybrock:
@OwenArnold I resolved the merge conflicts. I don't see anything to add here right now, this PR is already too big. The rollout is far from complete, of course.

Framework/Indexing/src/IndexInfo.cpp
```diff
@@ -0,0 +1,122 @@
...
+  auto &detIDs = m_detectorIDs.access();
+  detIDs = std::move(detectorIDs);
+  for (auto &ids : detIDs) {
+    std::sort(ids.begin(), ids.end());
+    ids.erase(std::unique(ids.begin(), ids.end()), ids.end());
+  }
+}
+
+} // namespace Indexing
+} // namespace Mantid
```
OwenArnold: This looks solid to me. I guess partitioning is something that can be added later?

Framework/Indexing/inc/MantidIndexing/IndexInfo.h
```diff
@@ -0,0 +1,97 @@
...
+public:
+  explicit IndexInfo(const size_t globalSize);
+  IndexInfo(std::vector<specnum_t> &&spectrumNumbers,
+            std::vector<std::vector<detid_t>> &&detectorIDs);
+  IndexInfo(
+      std::function<size_t()> getSize,
+      std::function<specnum_t(const size_t)> getSpectrumNumber,
+      std::function<const std::set<specnum_t> &(const size_t)> getDetectorIDs);
+
+  IndexInfo(const IndexInfo &other);
```
OwenArnold: The copy constructor has been overloaded, but not the assignment operator. Given the fact that vectors need to be filled in "legacy" mode, I that assignment will not work the way one might expect. One option may be to disable assignment altogether. It might not make sense to allow it.

Framework/Indexing/inc/MantidIndexing/IndexInfo.h
```diff
@@ -0,0 +1,97 @@
...
+  /// legacy mode.
+  std::function<size_t()> m_getSize;
+  /// Function bound to MatrixWorkspace::spectrumNumber(), used only in legacy
+  /// mode.
+  std::function<specnum_t(const size_t)> m_getSpectrumNumber;
+  /// Function bound to MatrixWorkspace::detectorIDs(), used only in legacy
+  /// mode.
+  std::function<const std::set<specnum_t> &(const size_t)> m_getDetectorIDs;
+
+  Kernel::cow_ptr<std::vector<specnum_t>> m_spectrumNumbers;
```
SimonHeybrock: Indeed, we did not.

OwenArnold:
:shipit:

PR #17125, opened 2016-08-03
RRO documentation overhaul

### LINKED ISSUE #17107: ReflectometryReductionOne documentation overhaul
Edit: This issue has been changed from simply altering the workflow diagram to overhauling the whole documentation.

**Workflow Diagram**

There are several issues that must be addressed with the workflow diagram:

- 'No monitors' should not be separate from 'monitors' - a decision diamond should split the two, rather than splitting from 'X axis in Y'

- The ConvertUnits formula performed is not the same for Monitors and Detectors. For Detectors, it is performed with parameter `AlignBins = True`. This difference should be shown.

- In each case, more actions than 'ConvertUnits' are being performed, as shown in the formulas `toLamDetector` and `toLamMonitor` from `ReflectometryWorkflowBase::toLam`. These as well as appropriate inputs should be included.

- The resultant monitor and detector workspaces are used in another. The detector WS is rebinned to the monitor WS. This must also be represented.

- The 'Perform Direct Beam Normalisation' block should be split into three blocks:
  - Sum over direct beam
  - Rebin to detector workspace
  - Normalize by the direct beam

Finally, a high level version of the existing diagram should be produced, the original being a 'low-level' version.

Edit: In addition, you should also check the documentation for the rest of the tree and make similar improvements if they are lacking.

Edit: The low-level diagram will now be split into four smaller diagrams as follows:
  - ConvertToWavelength
  - TransmissionCorrection
  - CorrectDetectorPositions
  - ConvertToMomentum

**Documentation**

As suggested by @raquelalvarezbanos the following will be included:

* First section is a brief description of the algorithm.

* Second section shows the high-level workflow diagram.

* Third section contains subsections for each of the five smaller diagrams that the low-level diagram is split into.

PranavBahuguna:
Description of work.

Made several improvements and provided better clarification of the various functions in the workflow diagram for the algorithm. I have also rearranged much of the documentation to accompany the new workflow diagram. Many of these improvements are described in the issue it fixed.

Fixes #17107.

---
#### Reviewer

##### Code Review
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)? Is it well structured with small focused classes/methods/functions?
##### Functional Tests


raquelalvarezbanos:
I have a few comments about the new workflow diagrams:

Regarding the **high level** diagram, I think we could make the diagram more readable, I find it too complicated for a high level diagram. For instance, the first part of the diagram looks like this (I am using an online graphviz viewer because my PC is not working today, therefore colors and shapes won't match what we get when building the documentation, but the idea is the same):

highlevelworkflowdiagram1

- I would say there is no reason to use abbreviations. `CheckXUnits` could simply be something similar to what we had before: `Is X axis in \lambda?`. Otherwise, someone who looks at the diagram will not know what we are checking for, they could think that `No` means that we are not checking anything.
- Other examples are: `CheckTransRun` (which could be replaced with `Are transmission runs given?`), `CheckCorAlg`, `CorrAlgProc` (I don't know what this stands for, Correction Algorithm Processing?), etc.
- When we have the two possibilities, `Yes` and `No`, for `CheckXUnits`, we have a `convertUnits` box if the answer is `No`. I think we should be very careful when using algorithm names, as someone could think that we are just running `ConvertUnits`. I think we should not use names of algorithms (unless we are actually running them) to avoid this confusion.

I think something as simple as the following should be fine for a high level diagram:

highlevelworkflowdiagram2


For the **low level** diagram, as it is getting so big, I really think we should split it into different sections, matching what we have in the high level diagram. For instance, the first section would be "Convert to wavelength" and would correspond to "Convert to wavelength and normalize by monitors if needed" in the high level diagram (note there is no need to use the full name if it is too long). The second section would be "Transmission corrections" and would contain all the steps related to transmission runs and correction algorithms. The other sections would be "Calculate ThetaIn", "Correct detector positions", and "Convert to momentum transfer, rebin and scale". 

Let me know what you think.

raquelalvarezbanos:
Another comment: for the **low level** diagram, I see you are using abbreviations too, for instance, `wavMin`, `wavMax`, `procCommands`, etc. I think we should use the names of the input properties (`WavelengthMin`, `WavelengthMax`, `ProcessingInstructions`, etc), this way users can see where exactly each input property is being used. In general, try to use the full word to make the diagram more clear (for instance, I don't know what `valMon` or `valUnity` mean).

raquelalvarezbanos:
Thinking about the diagram I suggested above, I would even probably replace the boxes `Are transmission runs given?`, `Divide by transmission run` and `Apply correction algorithm if specified` with just a single box named `Transmission corrections`. Then in the low level diagram we can specify what we do when transmission runs are given or correction algorithms specified.

PranavBahuguna:
Regarding both of the diagram labelling, could you check this again? Those abbreviations are supposed to be replaced with labels assigned to each one that provide better descriptions (e.g. `Is X axis in \lambda?`). You can see these label descriptions in the new files added.

raquelalvarezbanos:
Apologies, my mistake, I missed that part :S. 

I'll review the low level diagram. Regarding the high level one, could we simplify it to have something similar to what I suggested above? What do you think? I think something like that will make it easier for users (and future developers) to understand the main steps in the algorithm. Then they can refer to the specific sections for a more detailed description. I also think we should split the low level diagram into completely separate sections, what do you think?

raquelalvarezbanos:
Another suggestion is that currently, both diagrams (low level and high level) are displayed one after the other and there's no description about what they represent. Can I suggest adding just a few lines before each diagram? For instance something like: "Below is displayed a high level workflow diagram illustrating the main steps in the reduction. For a more detailed description, please refer to the low level diagram". And something similar for each of the bits in the low level diagram.

docs/source/diagrams/ReflectometryReductionOne_HighLvl-v1_wkflw.dot
```diff
@@ -12,37 +12,31 @@ subgraph params {
...
 }
 
 subgraph algorithms {
  $algorithm_style
-  divideTrans   [label="Divide by\ntransmission run"]
-  corrAlg       [label="Apply correction algorithm\nif specified"]
   calcTheta     [label="Calculate ThetaIn"]
   corrDetPos    [label="Correct detector\npositions if necessary"]
   convertMom    [label="Convert to momentum\ntransfer, rebin and scale"]
 }
```
raquelalvarezbanos: The high level diagram looks very good now with the last commit. However, could we, just for consistency, define the above as `processes`? They are not algorithms any more.

docs/source/diagrams/ReflectometryReductionOne_LowLvl-v1_wkflw.dot
```diff
@@ -55,17 +66,28 @@ $algorithm_style
   rebin           [label="Rebin"]
   scale           [label="Scale"]
   calcResolution  [label="CalculateResolution"]
+  convertDet      [label="Convert Units\n(AlignBins = True)"]
+  convertMon      [label="Convert Units"]
+  groupDet        [label="Group Detectors"]
+  cropDetWS       [label="Crop Detector\nWorkspace"]
+  rebinDet        [label="Rebin Detectors"]
```
raquelalvarezbanos: We should use the names of the algorithms above. For instance, `Rebin Detectors` should be just `Rebin` (there's no algorithm called like that. I don't think it is necessary here to explicitly say we are rebinning the detector workspace, as we already have two separate paths for detectors and monitors).

The same for the rest of the algorithms:

- `Convert Units\n(AlignBins = True)` -> `ConvertUnits\n(AlignBins=True)` (In this case I think it's OK to leave the `AlignBins = True` there).
- `Convert Units` -> `ConvertUnits`
- `Group Detectors` -> `GroupDetectors`
- `CropDetector\nWorkspace` -> `CropWorkspace`

raquelalvarezbanos:
For the low level diagram, the first half of the diagram is:

top-half

I think the bit related to the `DirectBeam` normalization is not reflecting exactly what the algorithm does. We shouldn't assume users will know what algorithms we are running and with what parameter values when we say `Sum Direct Beam`, `Rebin By Direct Beam` and `Normalize By Direct Beam`. I think a more exact description would be:

directbeamnormalization

@PranavBahuguna: Could you have a look and let me know if you agree that this is exactly what we are doing?

docs/source/diagrams/ReflectometryReductionOne_LowLvl-v1_wkflw.dot
```diff
@@ -1,27 +1,36 @@
...
   corrAlg           [label="CorrectionAlgorithm"]
   momentumTransferMin  [label="MomentumTransferMinimum"]
   momentumTransferStep [label="MomentumTransferStep"]
   momentumTransferMax  [label="MomentumTransferMaximum"]
+  procCommands      [label="ProcessingCommands"]
+  wavMax            [label="WavelengthMax"]
+  wavMin            [label="WavelengthMin"]
+  wavStep           [label="WavelengthStep"]
+  monStartWSIndex   [label="StartWorkspaceIndex"]
+  monEndWSIndex     [label="EndWorkspaceIndex"]
```
raquelalvarezbanos: What are `StartWorkspaceIndex` and `EndWorkspaceIndex`? They are not listed as input properties. I think these should be replaced by `I0MonitorIndex`, which is the input property we are using to run `CropWorkspace` on monitors, if I am correct.

docs/source/diagrams/ReflectometryReductionOne_LowLvl-v1_wkflw.dot
```diff
@@ -32,17 +41,19 @@ $decision_style
...
   checkSourceRotation [label="Is Source in Correct Position?"]
   checkSourceRotationBack [label = "Was Source Rotated?"]
-  checkRebinMin  [label="MomentumTransferMinimum given?"]
-  checkRebinMax [label="MomentumTransferMaximum given?"]
+  checkRebinMin   [label="MomentumTransferMinimum given?"]
+  checkRebinMax   [label="MomentumTransferMaximum given?"]
   checkRebinStep  [label="MomentumTransferStep given?"]
   checkScale      [label="ScaleFactor given?"]
+  checkMonUnity   [label="Monitors?"]
+  checkBgMinMax   [label="Min or Max\nnon-zero?"]
```
raquelalvarezbanos: What do `Min` and `Max` refer to? We have `WavelengthMin` and `WavelengthMax`, `MonitorInitegrationWavelengthMin` and `MonitorIntegrationWavelengthMax`, `MonitorBackgroundWavelengthMin` and `MonitorBackGroundWavelengthMax`, etc. I think here we are referring to the latter two.

docs/source/diagrams/ReflectometryReductionOne_LowLvl-v1_wkflw.dot
```diff
@@ -55,17 +66,28 @@ $algorithm_style
...
   scale           [label="Scale"]
   calcResolution  [label="CalculateResolution"]
+  convertDet      [label="Convert Units\n(AlignBins = True)"]
+  convertMon      [label="Convert Units"]
+  groupDet        [label="Group Detectors"]
+  cropDetWS       [label="Crop Detector\nWorkspace"]
+  rebinDet        [label="Rebin Detectors"]
+  cropMonWS       [label="Crop Monitor\nWorkspace"]
+  calcFlatBg      [label="Calculate Flat\nBackground"]
+  rebinToWS       [label="Rebin To\nWorkspace"]
```
raquelalvarezbanos: Again, those should be `CropWorkspace`, `CalculateFlatBackground`, and `RebinToWorkspace`, which are the exact names of the algorithms we are running.

PranavBahuguna:
@raquelalvarezbanos I agree with this proposition for the above diagram, and will implement this later.

docs/source/diagrams/ReflectometryReductionOne_LowLvl-v1_wkflw.dot
```diff
@@ -32,17 +41,19 @@ $decision_style
...
   checkSourceRotationBack [label = "Was Source Rotated?"]
-  checkRebinMin  [label="MomentumTransferMinimum given?"]
-  checkRebinMax [label="MomentumTransferMaximum given?"]
+  checkRebinMin   [label="MomentumTransferMinimum given?"]
+  checkRebinMax   [label="MomentumTransferMaximum given?"]
   checkRebinStep  [label="MomentumTransferStep given?"]
   checkScale      [label="ScaleFactor given?"]
+  checkMonUnity   [label="Monitors?"]
+  checkBgMinMax   [label="Min or Max\nnon-zero?"]
+  monIntWavInterval [label="Monitor Integration\nWavelength Interval"]
```
raquelalvarezbanos: I think this should be split into two different input properties (also, note that they should be defined as `param` instead of `decisions`), and called `MonitorIntegrationWavelengthMin` and `MonitorIntegrationWavelengthMax`, which are the names of the input properties we are using.

docs/source/diagrams/ReflectometryReductionOne_LowLvl-v1_wkflw.dot
```diff
@@ -75,23 +97,45 @@ $value_style
...
-valMon            -> checkNormDetMon
+checkXUnit        -> convertDet       [label="Detectors"]
+checkXUnit        -> checkMonUnity    [label="Monitors"]
+checkMonUnity     -> convertMon       [label="Yes"]
+checkMonUnity     -> valUnity         [label="No"]
+convertDet        -> valDet
+valDet            -> groupDet
+procCommands      -> groupDet
+convertMon        -> valMon
+monIntWavInterval -> intMon
```
raquelalvarezbanos: `intMon` is currently defined as:

`intMon [label="Integrate"]`

but the algorithm is `Integration`, so I think it should be defined as:

`intMon [label="Integration"]`

docs/source/diagrams/ReflectometryReductionOne_LowLvl-v1_wkflw.dot
```diff
@@ -55,17 +66,28 @@ $algorithm_style
...
+  groupDet        [label="Group Detectors"]
+  cropDetWS       [label="Crop Detector\nWorkspace"]
+  rebinDet        [label="Rebin Detectors"]
+  cropMonWS       [label="Crop Monitor\nWorkspace"]
+  calcFlatBg      [label="Calculate Flat\nBackground"]
+  rebinToWS       [label="Rebin To\nWorkspace"]
+  sumDirectBeam   [label="Sum Direct\nBeam"]
+  rebinByDirectBeam [label="Rebin By\nDirect Beam"]
+  normalizeByDirectBeam [label="Normalize By\nDirect Beam"]
+  rebinByTransWS  [label="Rebin By Transmission\nWorkspace"]
```
raquelalvarezbanos: The same for `Rebin By Transmission\nWorkspace`, it should be `RebinToWorkspace`. The workspace we are taking as a reference for the rebinning is the input workspace in wavelength (with all normalizations applied), therefore it could be worth adding a `param` item called `InputWorkspace in wavelength` between `Divide (Detectors/Monitors/Unity)` and `FirstTransmissionRun given?`. This can be then linked to `RebinToWorkspace` to indicate that it is the input property to the algorithm.

raquelalvarezbanos:
I think the bit from `ThetaIn given?` to `OutputWorkspaceWavelenth` also needs to be updated. Looking at the code, I'd say it should be like this:

thetaingiven

Do you agree?

raquelalvarezbanos:
The final bit between `OutputWorkspaceWavelength` and `OutputWorkspace` is correct, I think. @PranavBahuguna could you review my comments and let me know if you disagree with any of them? 

In addition, I suggest rearranging the existing documentation:

- The first section should be "Description", which is fine as it is now, as it is just a brief summary of the algorithm.
- Next we should show a section called something like "High level workflow" containing the high level diagram.
- The next section could be a section called "Low level workflow" that contains subsections for each of the main bits in the high level workflow. Those subsections could be:
 - "Conversion to Wavelength". It should include the first two sentences in the current "Analysis Modes" sections (note that the rest of the section refers to other calculations, therefore I think it shouldn't go here).
 - "Transmission Corrections", we can copy and paste what we currently have plus section "Polynomial corrections".
 - "Detector Position Correction". Use what we have in the existing section with the same name. In addition, I think it is important to stress that detector positions are corrected in terms of the `ThetaIn` given and that, in general, they should always be corrected (otherwise `ConvertUnits` won't calculate the right `Q` values), unless the `InputWorkspace` has already the detectors in the right positions, which can be achieved by running `MoveInstrumentComponent` before `ReflectometryReductionOne`. Let me know if you have any questions about this.
 - "Conversion to Momentum Transfer (Q)". Section "Source Rotation" should be a subsubsection here. Then copying and pasting what we have in "Post-processing Options" (please don't include this section title) should be enough.

PranavBahuguna:
* On analysis of the `thetaIn` section of code, I agree that the new diagram seems to reflect whats really going on inside the code. I will include this as well.

* With "Conversion to Wavelength", while I agree with including the sentences from "Analysis Modes", I don't see why we shouldn't include the 3rd and 4th sentences. They may refer to other calculations, but they potentially offer better clarity and context regarding its operation and input,

* "Transmission Corrections" - agreed.

* "Detector Position Correction" - I agree as well with this, especially as the original description is very sparse.

* "Conversion to Momentum Transfer (Q)" - seems OK. I assume you wish to keep the "Rebinning" and "Scaling" subsections as well?

raquelalvarezbanos:
- Regarding the new "Conversion to wavelength" section. Yes, you're right that the third sentence should be included too. I think 4th sentence ("The normalization by transmission runs...") should go in the "Transmission corrections" sentence.

- For "Conversion to Momentum Transfer", yes, I also think we should keep the "Rebinning" and "Scaling" subsections.

- Regarding the last question, it's up to you, but I would probably update the documentation as part of this issue (so you can avoid creating a new branch, building, etc). You can always edit the original issue to include this task and change the title if necessary.

docs/source/diagrams/ReflectometryReductionOne_ConvertToWavelength-v1_wkflw.dot
```diff
@@ -0,0 +1,105 @@
...
+
+subgraph algorithms {
+ $algorithm_style
+  convertDet    [label="ConvertUnits\n(AlignBins = True)"]
+  convertMon    [label="ConvertUnits"]
+  groupDet      [label="GroupDetectors"]
+  intMon        [label="Integration"]
+  cropDetWS     [label="CropWorkspace"]
+  rebinDet      [label="Rebin"]
+  rebinToWS     [label="Rebin"]
```
raquelalvarezbanos: Shouldn't this be a `RebinToWorkspace`? Please check

docs/source/diagrams/ReflectometryReductionOne_ConvertToWavelength-v1_wkflw.dot
```diff
@@ -0,0 +1,105 @@
+digraph ReflectometryReductionOne {
+label = "ReflectometryReductionOne ConvertToWavelength Flowchart"
```
raquelalvarezbanos: I've been trying a few things to make the workflow look nicer and I've noticed that adding:

`rankdir = TB;`

here and
```
{rank = same; convertDet; convertMon}
{rank = same; valUnity; valDet; valMon}
{rank = same; cropDetWS; cropMonWS}
```

and the end of the `digraph` seems to improve the result.

Also replacing:

```
checkBgMinMax   [label="Monitor background wavelength\nmin/max non-zero?"]
```
with 
```
checkBgMinMax   [label="Monitor background\nwavelength\nmin/max non-zero?"]
```

and
```
checkMultRDB    [label="MultidetectorAnalysis and\nRegionOfDirectBeam given?"]
```
with
```
checkMultRDB    [label="MultidetectorAnalysis\nand\nRegionOfDirectBeam\ngiven?"]
```
also improves the result.

docs/source/diagrams/ReflectometryReductionOne_TransmissionCorrection-v1_wkflw.dot
```diff
@@ -0,0 +1,59 @@
...
+subgraph processes {
+ $process_style
+}
+
+subgraph values {
+ $value_style
+  valTrans        [label="It(&lambda;) / &Sigma;It(&lambda;)"]
+}
+
+inputWorkspace    -> checkTransRun
```
raquelalvarezbanos: We must be careful with using `InputWorkspace` here (the same applies to subsequent diagrams). Strictly speaking this is not our `InputWorkspace` but our `DetectorWorkspace` after the normalization. I don't mind keeping this name as long as we add a sentence somewhere in this section stating that this is the `InputWorkspace` once it has converted to wavelength and normalized by monitors.

docs/source/diagrams/ReflectometryReductionOne_TransmissionCorrection-v1_wkflw.dot
```diff
@@ -0,0 +1,59 @@
...
+polyCorr          -> outputWorkspace
+expCorr           -> outputWorkspace
+firstTransRun     -> checkTransUnits
+checkTransUnits   -> valTrans         [label="Yes"]
+checkTransUnits   -> createTransWS    [label="No"]
+secondTransRun    -> createTransWS
+createTransWS     -> valTrans
+valTrans          -> rebinByTransWS
+rebinByTransWS    -> divideTrans
+divideTrans       -> outputWorkspace
```
raquelalvarezbanos: Something similar happens with `OutputWorkspace`, we should rename this to maybe just `Output` or keep it but adding a sentence saying that this is not the output of the algorithm, and that there will be more steps applied as explained in the next sections.

raquelalvarezbanos:
Looks very good, thanks for your last commit @PranavBahuguna 

:shipit:

PR #17127, opened 2016-08-03
Add CMake option to build only the Framework

peterfpeterson:
This is intended to consolidate which `CMakeLists.txt` is used for configuring mantid.

**To test:**

For all of these steps, none of the gui should appear in the output area.
1. Create a fresh build area
2. `cmake -DENABLE_MANTIDPLOT=False ...` with the normal other arguments set including the location of the upper level `CMakeLists.txt`
3. `cmake --build . -- -j <numthread>`
4. `cmake --build . --target FrameworkTests -- -j <numthread>`
5. `ctest -j <numthread>`

_There is no associated issue._

---
#### Reviewer

##### Code Review
##### Functional Tests

The additional keyword is ENABLE_MANTIDPLOT and defaults to True.


rgmiller:
I just tried the MPI build and it worked, but the system tests still fail to link because we’re not including libmpi.so.

rgmiller:
Should ENABLE_MANTIDPLOT and MPI_BUILD be mutually exclusive?  About the only thing that MPI_BUILD enables are the BroadcastWorkspaces and GatherWorkspaces algorithms, but you’d need to start up with mpirun to use them, and that’s probably not something you want to do with MantidPlot.

VickieLynch:
My first thought was exactly the same as Ross's, but then I thought about shared memory machines with hundreds/thousands of cores in the future. Maybe then a user would want to use MantidPlot and have MPI reduce the data with each bank of each run on a separate core and still have the window to look at the results.

peterfpeterson:
@yxqd said in an email
```
I tried it but it failed at finding PyQT. I wonder if it is OK to ignore everything
related to graphics such as OpenGL, Qt, Paraview, etc? Or the framework also needs
some of them?
```

I say, ignore away!

rgmiller:
> ...shared memory machines with hundreds/thousands of cores in the future.

In that case, you still wouldn't want to start MantidPlot with mpirun:  you'd end up with hundreds/thousands of MantidPlot windows.  Besides, on a shared memory machine, you don't need message passing since all threads can just access the same address space.  On a shared memory machine, OpenMP or similar is probably the way to go (and which Mantid already uses). Admittedly, I don't know if OpenMP scales to thousands of cores yet, but I suspect it will by the the time we actually have thousands of cores to work with...

rgmiller:
Minor quibble: After building everything, the bin directory contains a "launch_mantidplot.sh" script which gets included in the RPM when you package everything up.  Running that script fails, of course, because MantidPlot itself doesn't exist.

rgmiller:
The RPM's dependencies need to be updated: it's still requiring GUI things such as qt4, PyQt4 & qwt.  Also libGL.so and libGLU.so.   Not sure if fixing the RPM dependencies should be part of this issue, though.

rgmiller:
`cpack -G RPM` works, but `cpack -G TGZ` fails with two errors:
>CMake Error at /home/xmr/mantid-build/cmake_install.cmake:63 (file):
>  file INSTALL cannot copy file
>  "/home/xmr/git/mantid/installers/LinuxInstaller/mantidplot.desktop" to
>  "/usr/share/applications/mantidplot.desktop".
>
>
>CMake Error at /home/xmr/mantid-build/cmake_install.cmake:75 (file):
>  file INSTALL cannot copy file
>  "/home/xmr/git/mantid/images/MantidPlot_Icon_32offset.png" to
>  "/usr/share/pixmaps/mantidplot.png".

This is on both CentOS6 and CentOS7.  (On CentOS6, I wrapped the cpack calls with `scl enable devtoolset-2`.)

rgmiller:
We've decided clean up the packaging in a separate pull request (or possibly multiple requests), so this request is ready to go. :shipit:

PR #17145, opened 2016-08-04
Use PARALLEL_FOR_IF with threadSafe function

quantumsteve:
Description of work.

This adds a variadic function `ThreadSafe(...)` to determine whether one or more workspaces thread-safe. We can then remove  `PARALLEL_FOR1`, `PARALLEL_FOR2`, and `PARALLEL_FOR3` macros.

**To test:**

---
#### Reviewer

##### Code Review
##### Functional Tests


quantumsteve:
@martyngigg I hid the function name inside `PARALLEL_FOR_IF(...)` and am debating whether this is better or worse than a macro-only solution. Does access to the type system justify the slightly uglier syntax?

martyngigg:
@quantumsteve I don't know if it does. I think you first version, with the `threadSafe` part also appearing in the call to `PARALLEL_FOR_IF`, reads better as then you can see that the `if` is dependent on thread-safety rather than having to dig into the macro definition.

It would certainly be nice if `threadSafe` could accept references but `.get` is not horrendous.

Framework/Algorithms/inc/MantidAlgorithms/SpectrumAlgorithm.h
```diff
@@ -88,7 +88,7 @@ class MANTID_ALGORITHMS_DLL SpectrumAlgorithm : public API::Algorithm {
     auto size = static_cast<int64_t>(indexSet.size());
     API::Progress progress(this, 0.0, 1.0, size);
 
-    PARALLEL_FOR1((&workspace))
+    PARALLEL_FOR_IF(Kernel::threadSafe(workspace))
```
martyngigg: I like this. It's nice that it now conveys that there is actually a `if` going on and it also describes the check too.

Framework/Algorithms/src/BinaryOperation.cpp
```diff
@@ -492,7 +492,7 @@ void BinaryOperation::doSingleValue() {
 
   if (m_eout) {
     // ---- The output is an EventWorkspace ------
-    PARALLEL_FOR3(m_lhs, m_rhs, m_out)
+    PARALLEL_FOR_IF(Kernel::threadSafe(m_lhs.get(), m_rhs.get(), m_out.get()))
```
martyngigg: The single parameter variant now uses a reference. Is there a reason why the 3 parameter version has to call `.get()`?

quantumsteve: One should be able to use any combination of pointers and references. Pointers are checked for  `nullptr`, while references must never be NULL.

I changed the functions that were generating warnings with GCC 6.  Some of the three parameter versions can probably also be changed.

Framework/Algorithms/src/Rebin.cpp
```diff
@@ -179,7 +179,7 @@ void Rebin::exec() {
       Progress prog(this, 0.0, 1.0, histnumber);
 
       // Go through all the histograms and set the data
-      PARALLEL_FOR3(inputWS, eventInputWS, outputWS)
+      PARALLEL_FOR_IF(Kernel::threadSafe(*inputWS, *outputWS))
       for (int i = 0; i < histnumber; ++i) {
```
quantumsteve: I removed `eventInputWS` since it points to the same workspace as `inputWS`.

martyngigg:
@quantumsteve Is this ready to go now?

quantumsteve:
@martyngigg Yes, I think so.

Framework/Algorithms/src/Q1D2.cpp
```diff
@@ -146,7 +146,7 @@ void Q1D2::exec() {
   Progress progress(this, 0.05, 1.0, numSpec + 1);
 
   const auto &spectrumInfo = m_dataWS->spectrumInfo();
-  PARALLEL_FOR3(m_dataWS, outputWS, pixelAdj)
+  PARALLEL_FOR_IF(Kernel::threadSafe(*m_dataWS, *outputWS, pixelAdj.get()))
```
martyngigg: Should this be `*pixelAdj`?

quantumsteve: No, `pixelAdj` can be `NULL`.  I tried dereferencing it and the unit test failed.

martyngigg: I see, that make sense.

martyngigg (approved):
I'm happy with this now. It's a definite improvement in the readability and usability of these macros.

:shipit:

PR #17147, opened 2016-08-05
Basic math operations for Histogram

### LINKED ISSUE #17137: Add more basic math operations for Histogram
Add more frequently used math bits for `Histogram` and sub-types. See comments on this PR for examples: https://github.com/mantidproject/mantid/pull/17121.

SimonHeybrock:
This PR adds a few basic math operations for `Histogram` and some of the X, Y, and E types.
- `-` and `/` with `double` argument for Y and E types (previously only `+` and `*` were implemented)
- `=` with `double` argument for Y and E types.
- `*` and `/` with `double` argument for `Histogram`
- `+` and `-` for two `Histograms`.
- Where applicable, also the corresponding `+=` operators (etc.) have been added.

**To test:**
- Thorough code review.
- Check that the tests leave no holes.
- Please comment whether you think the amount of testing of input values is reasonable and correct.

This partially resolves #17137.

No release notes (we still have no Python interface for `HistogramData`).

Once code review confirmed that these changes are as we want them I will update the developer documentation with a list of newly available operations.

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/HistogramData/src/HistogramMath.cpp
```diff
@@ -0,0 +1,107 @@
...
+/// Subtracts data in other Histogram from this Histogram, propagating
+/// uncertainties.
+Histogram &operator-=(Histogram &histogram, const Histogram &other) {
+  checkSameXYMode(histogram, other);
+  checkSameX(histogram, other);
+  histogram.mutableY() -= other.y();
+  std::transform(histogram.e().cbegin(), histogram.e().cend(),
+                 other.e().begin(), histogram.mutableE().begin(),
+                 [](const double &lhs, const double &rhs)
+                     -> double { return std::sqrt(lhs * lhs + rhs * rhs); });
```
LamarMoore: Is this correct? Should this be `lhs * lhs - rhs * rhs`

SimonHeybrock: Should be correct:
https://en.wikipedia.org/wiki/Propagation_of_uncertainty#Example_formulas

Framework/HistogramData/test/HistogramMathTest.h
```diff
@@ -0,0 +1,184 @@
...
+    const Histogram hist1(BinEdges{1, 2, 3}, Counts{4, 9});
+    const Histogram hist2(BinEdges{1, 2}, Counts{1});
+    TS_ASSERT_THROWS(hist1 + hist2, std::runtime_error);
+  }
+
+  void test_plus_histogram_fail_x_value_mismatch() {
+    const Histogram hist1(BinEdges{1, 2.0, 3}, Counts{4, 9});
+    const Histogram hist2(BinEdges{1, 2.1, 3}, Counts{1, 2});
+    TS_ASSERT_THROWS(hist1 + hist2, std::runtime_error);
+  }
```
LamarMoore: I don't understand why this should fail

Framework/HistogramData/test/ScalableTest.h
```diff
@@ -60,6 +78,14 @@ class ScalableTest : public CxxTest::TestSuite {
     TS_ASSERT_DELTA(lhs[1], 0.6, 1e-14);
   }
 
+  void test_divide_reverse_order() {
```
LamarMoore: I am conflicted about whether or not a divide in reverse order should be allowed. I like the idea but not sure if it really makes sense to allow this. What do you think @OwenArnold ?

LamarMoore:
Looks good to me. These are very useful additions which should hopefully make this cause become more popular. I think the unit testing is quite comprehensive and it doesn't seem like anything is missing. Unless @OwenArnold  can spot anything else.

Framework/HistogramData/test/ScalableTest.h
SimonHeybrock: Right, it makes sense in some cases, but not in others. It would be correct for, e.g., `Frequencies`, but wrong for `Counts` (or rather the output would not be `Counts` in that case). Should we remove it?

martyngigg: I would say we shouldn't allow this is it's going to be wrong on certain cases. If it's there it will be assumed it does the correct thing.

martyngigg:
Looks good to me too.

:shipit:

OwenArnold:
I'm going to take a quick look with the intention of merging this very soon.

Framework/HistogramData/inc/MantidHistogramData/Counts.h
```diff
@@ -46,12 +47,22 @@ class Frequencies;
 */
 class MANTID_HISTOGRAMDATA_DLL Counts
     : public detail::VectorOf<Counts, HistogramY>,
+      public detail::Addable<Counts>,
```
OwenArnold: 👍

Framework/HistogramData/src/HistogramMath.cpp
```diff
@@ -0,0 +1,107 @@
...
+}
+
+/** Dividies data in histogram by constant (positive) factor.
+ *
+ * Uncertainties are divided by the same factor, such that the *relative*
+ * uncertainties remain unchanged. */
+Histogram operator/(Histogram histogram, const double factor) {
+  return histogram *= 1.0 / factor;
+}
+
```
OwenArnold: 👍  Good that you remembered to do that. Adding the non-member functions that is. Very important.

Framework/HistogramData/test/AddableTest.h
```diff
@@ -0,0 +1,92 @@
...
+#include "MantidHistogramData/Iterable.h"
+#include "MantidHistogramData/VectorOf.h"
+
+using Mantid::HistogramData::detail::Addable;
+using Mantid::HistogramData::detail::FixedLengthVector;
+using Mantid::HistogramData::detail::Iterable;
+using Mantid::HistogramData::detail::VectorOf;
+using Mantid::HistogramData::HistogramX;
+
+class AddableTester : public VectorOf<AddableTester, HistogramX>,
```
OwenArnold: As a very minor point. I'm not a great fan of seeing that testing the Addable functionality requires brining in so many other mixins. Doesn't this feel more like an integration test rather than a test of `Addable`?

SimonHeybrock: Well, in principle I could just test it with `std::vector<double>`, I think. We would then have no test of the functionality with `FixedLengthVector` and `VectorOf`, but basically that simply uses iterator access.

OwenArnold: I think having the minimal derived (concrete) type as well as fully `Iterable`, `Scaleable` etc derived type is important for unit tests. If someone messes up another one of the base classes, it would be easier to track down the issue if it didn't affect all tests. 

I'm not suggesting you change this now. I'm going to merge what you've done as Lamar wants to use it and as there's no reason to hold things up.

Issue #17146, opened 2016-08-04
Test failures on Fedora 24


quantumsteve:
### Expected behavior

Mantid builds on Fedora 24 and passes all tests 

### Actual behavior

EqualBinSizesValidatorTest failed to build
```
[1592/3133] Building CXX object Framework/API/test/CMakeFiles/APITest.dir/EqualBinSizesValidatorTest.cpp.o
FAILED: /usr/bin/c++   -DBOOST_ALL_DYN_LINK -DBOOST_ALL_NO_LIB -DBOOST_DATE_TIME_POSIX_TIME_STD_CONFIG -DENABLE_OPENCASCADE -DHAVE_STDINT_H -isystem /usr/include/nexus -I../Framework/Kernel/inc -I../Framework/HistogramData/inc -I../Framework/Geometry/inc -I../Framework/API/inc -isystem ../Testing/Tools/cxxtest -isystem ../Testing/Tools/gmock-1.7.0/include -isystem ../Testing/Tools/gmock-1.7.0/gtest/include -I../Framework/API/test/../../TestHelpers/inc -I../Framework/API/test/../../Nexus/inc -fopenmp -Wsuggest-override -Wall -Wextra -Wconversion -Winit-self -Wpointer-arith -Wcast-qual -Wcast-align -fno-common -Wno-deprecated -Wno-write-strings -Wno-unused-result -pedantic -Woverloaded-virtual -fno-operator-names -O3 -DNDEBUG -fvisibility=hidden   -std=gnu++14 -MMD -MT Framework/API/test/CMakeFiles/APITest.dir/EqualBinSizesValidatorTest.cpp.o -MF Framework/API/test/CMakeFiles/APITest.dir/EqualBinSizesValidatorTest.cpp.o.d -o Framework/API/test/CMakeFiles/APITest.dir/EqualBinSizesValidatorTest.cpp.o -c Framework/API/test/EqualBinSizesValidatorTest.cpp
In file included from ../Framework/API/inc/MantidAPI/MatrixWorkspaceValidator.h:6:0,
                 from ../Framework/API/inc/MantidAPI/EqualBinSizesValidator.h:5,
                 from /home/svh/Documents/mantid/Framework/API/test/EqualBinSizesValidatorTest.h:6,
                 from Framework/API/test/EqualBinSizesValidatorTest.cpp:15:
[...]

PR #17154, opened 2016-08-05
Work around internal compiler error on Fedora 24

quantumsteve:
Description of work.

This addresses an internal compiler error when building `EqualBinSizesValidatorTest::test_null`. The problem appears to be [obtaining the address of a `nullptr_t`](https://github.com/mantidproject/mantid/blob/27ca913c1f2e640916cdcdfd65b15f80669523d5/Framework/Kernel/inc/MantidKernel/IValidator.h#L146). The best solution I've found so far is to treat `nullptr_t` the same as we treat other 
pointer types. 

I also updated IValidator.h to use type_traits from the standard library instead of boost.

**To test:**

Build `APITest` with GCC 6.1

This is part of #17146.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Kernel/inc/MantidKernel/IValidator.h
+++ b/Framework/Kernel/inc/MantidKernel/IValidator.h
@@ -12,8 +12,6 @@
 #include <boost/any.hpp>
 #include <boost/shared_ptr.hpp>
 #include <boost/make_shared.hpp>
-#include <boost/type_traits/is_convertible.hpp>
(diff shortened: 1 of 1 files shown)


rosswhitfield:
`APITest` compiles now on Arch with gcc `6.1`. The three APITests still fail.

FedeMPouzols:
Smart trick. Hopefully they fix this in GCC soonish.

PR #17159, opened 2016-08-08
Add options to ColorBarWidget to change shown checkboxes and their label/tooltip text

### LINKED ISSUE #16154: Add options to the ColorBarWidget, labels and tooltips should not always be SliceViewer-centric
The `ColorBarWidget` is a `MantidWidget` and is convenient in interfaces that are not necessarily used in the same context as the `SliceViewer`. The widget includes two "autoscale" tick boxes with labels, behavior and tooltips that are too specific to the `SliceViewer`. The widget is now used in the imaging interface which will need a bit of polishing. See #16131 for example.

- An option to change the tooltip texts, and maybe the label texts, will probably be good.
- An option to hide some or both tick boxes, simplifying the widget, could be considered.
- An option to modify the behavior of the tick boxes could be remotely considered but I think it won't be needed for imaging.

dtasev:
## Description of work.

The ColorBarWidget.cpp now has methods for changing both of the [checkboxes' label, tooltip text](https://github.com/mantidproject/mantid/commit/713a181a3bc51eb0b79dac859b0849e15402c8a3#diff-4aa05715847f2ffc407c1b4bd6c6ca59R381) and [visibility](https://github.com/mantidproject/mantid/commit/480f71ba1d4c22dbbb2931fef1201ea2e96ca792#diff-4aa05715847f2ffc407c1b4bd6c6ca59R108).
## To test:
- Go to `Interfaces` > `Diffraction` > `Tomographic Reconstruction`, go to tab `ROI etc.`
- Check that only **1** checkbox is shown, should be `Autoscale on load`, and that the tooltip is correct `"This flag signals that the color scale range will be set automatically to the current slice range when an image is loaded. Note that auto scaling will be applied when an image is loaded for the very first time."`
- Look for any spelling mistakes in the tooltip, I might've missed them
- Check that the Color Bar Widget behaves normally in the other places it's used (such as the Slice Viewer on a Workspace), and any other usages that I am not aware of.

Fixes #16154 .

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/ColorBarWidget.h
+++ b/MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/ColorBarWidget.h
@@ -48,6 +48,12 @@ public:
   ColorBarWidget(QWidget *parent = 0);
   ~ColorBarWidget() override;
 
+  enum CheckboxStrategy {
+	  ADD_AUTOSCALE_CURRENT_SLICE = 0,
+	  ADD_AUTOSCALE_ON_LOAD = 1,
+	  ADD_AUTOSCALE_BOTH = 2
+  };
+
   void updateColorMap();
 
   void setViewRange(double min, double max);
@@ -55,6 +61,7 @@ public:
   void setMinimum(double min);
   void setMaximum(double max);
   void setRenderMode(bool rendering);
+  void setCheckBoxMode(CheckboxStrategy strategy);
 
   double getMinimum() const;
   double getMaximum() const;
@@ -73,6 +80,8 @@ public:
   bool getAutoScale() const;
 
   bool getAutoColorScaleforCurrentSlice() const;
+
+
 
 public slots:
   void changedMinimum();
--- a/MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/ColorBarWidget.h
+++ b/MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/ColorBarWidget.h
@@ -47,9 +47,9 @@ public:
   ~ColorBarWidget() override;
 
   enum CheckboxStrategy {
-	  ADD_AUTOSCALE_CURRENT_SLICE = 0,
-	  ADD_AUTOSCALE_ON_LOAD = 1,
-	  ADD_AUTOSCALE_BOTH = 2
+    ADD_AUTOSCALE_CURRENT_SLICE = 0,
+    ADD_AUTOSCALE_ON_LOAD = 1,
+    ADD_AUTOSCALE_BOTH = 2
   };
 
   void updateColorMap();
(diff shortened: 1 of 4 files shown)


MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/ColorBarWidget.h
```diff
@@ -56,11 +60,15 @@ class EXPORT_OPT_MANTIDQT_MANTIDWIDGETS ColorBarWidget : public QWidget {
   void setMaximum(double max);
   void setRenderMode(bool rendering);
 
+  /// Set which checkboxes are shown in the window
+  void setCheckBoxMode(CheckboxStrategy strategy);
```
FedeMPouzols: Nice. don't hesitate to add bits of documentation like this when you see they're missing.

MantidQt/MantidWidgets/src/ColorBarWidget.cpp
```diff
@@ -1,13 +1,10 @@
 #include "MantidQtAPI/MantidColorMap.h"
 #include "MantidQtMantidWidgets/ColorBarWidget.h"
 #include "MantidQtAPI/QScienceSpinBox.h"
-#include "qwt_scale_div.h"
 #include "MantidQtAPI/PowerScaleEngine.h"
-#include <iosfwd>
 #include <qwt_scale_map.h>
 #include <qwt_scale_widget.h>
 #include <QKeyEvent>
-#include <qwt_plot.h>
```
FedeMPouzols: Good catch with these unused headers. Did you do it manually, or does ReSharper help?

DimitarTasev: ReSharper with VS2015 grays out all unused headers, with a pretty good accuracy, at least for now.

FedeMPouzols:
This looks good. It adds:
- Methods to modify the label text and the tooltip text of the two checkboxes that the `ColorBarWidget` has at the bottom (auto-scale and auto-scale for current slice).
- A method to change the availabitily/visibility of the auto-scale checkboxes.

The default behavior is the classic behavior of the SliceViewer, but these modifications are convenient to customize the appearance and behavior of the color bar widget for different interfaces. The color bar widget in the SliceViewer looks and works as usual. 

sv_screenshot


In the tomographic reconstruction / ROI tab the color bar widget is now more simple and the text(s) make better sense. The  option "autoscale on load" works well:

tomo_screen1

tomo_screen2


:shipit:

MantidQt/MantidWidgets/src/ColorBarWidget.cpp
FedeMPouzols: Interesting to know. This is probably interesting to other developers using MSVC.

PR #17160, opened 2016-08-08
ReflectometryReductionOne flood correction property

### LINKED ISSUE #17085: Add additional input property to ReflectometryReductionOne to apply flood corrections.
This is an issue split off from task 5 of issue #16455. 

We need an additional input property to apply flood corrections. This property was suggested to be called `DetectorEfficiencyCorrection`. This will be a workspace which can contain:
- A single data point, in this case this value should be used as a normalisation constant shared across the different spectra in the input ws.
- A list of spectra with a single value: in this case each spectrum will have its own normalisation constant.

PranavBahuguna:
This update adds a new property `DetectorEfficiencyCorrection` - an input workspace that applies flood corrections. ~~This property replaces the `RegionOfDirectBeam` input property and replaces the `Multidetector?` and `Perform Direct Beam Normalization` stages in the workflow diagram for ReflectometryReductionOne.~~ It was later instead decided that this property as well as those stages would remain after all. In addition, the input workspace is now converted to units of wavelength before being split into detectors and monitor workspaces.

**To test:**

~~The old tests involving `RegionOfDirectBeam` have been removed~~. Two new tests was added to to account for the new `DetectorEfficiencyCorrection` property. They each test that:
- The flood correction workspace has the same number of spectra as the input workspace. 
- The flood correction workspace does not have more than one value per spectra.

Simply run the tests to ensure that they pass.

Fixes #17085.
#### Reviewer

##### Code Review
##### Functional Tests


Framework/Algorithms/src/ReflectometryReductionOne.cpp
```diff
@@ -521,6 +519,26 @@ void ReflectometryReductionOne::exec() {
...
+    }
+    for (size_t i = 0; i < decTemp->getNumberHistograms(); i++) {
+      for (size_t j = 0; j < decTemp->blocksize(); j++) {
+        if (decTemp->y(i)[j] == 0) {
+          throw std::invalid_argument(
+              "The DetectorEfficiencyCorrection  "
+              "workspace cannot contain any zero values.");
+        }
+      }
+    }
```
raquelalvarezbanos: The right place to check that input properties are what we expect is the method `validateInputs()`. As an example see [here](https://github.com/mantidproject/mantid/blob/master/Framework/Algorithms/src/MaxEnt.cpp#L168-L186)

Framework/Algorithms/src/ReflectometryReductionOne.cpp
```diff
@@ -578,29 +590,30 @@ void ReflectometryReductionOne::exec() {
...
-        buffer << db.front() << "-" << db.back();
-        MatrixWorkspace_sptr regionOfDirectBeamWS = this->toLamDetector(
-            buffer.str(), runWS, wavelengthInterval, wavelengthStep);
+    if (detectorEfficiencyCorrection.is_initialized()) {
+      MatrixWorkspace_sptr decWS = *detectorEfficiencyCorrection;
+      if (decWS->blocksize() > 1) {
+        // If using separate normalization constants for each spectra, rebin
+        // to the detector workspace
+        if (decWS->blocksize() < detectorWS->blocksize()) {
+          throw std::invalid_argument("The number of y-values in the detector "
```
raquelalvarezbanos: There's something here I don't understand. According to the issue, there are two possibilities for the `DetectorEfficiencyCorrection` workspace:

1. It is a workspace with a single spectrum and a single value in this spectrum
2. It is a workspace with `N` spectra and a single value per spectrum.

@PranavBahuguna Am I correct about this? If so, there's no point in checking `blocksize()` as it'll always be 1. In fact, the condition:

```
decWS->blocksize() < detectorWS->blocksize()
```

should/must always be true. What we should be checking is that `blocksize() == 1` and that if the number of spectra in `decWS` (if more than one) matches the number of spectra in `detectorWS`.

PranavBahuguna: The issue at heart is that in the original code, both the `RegionOfDirectBeam` WS (now `decWS`) and the `detector` WS were both provided as 1-spectra many values workspaces. The blocksize checks were to ensure the `decWS` wasn't smaller than the `detectorWS` it was being applied to. However, `decWS` is now provided externally and is in many-spectra 1-value format. I will deal with this by attempting to convert it into the original form - 1-spectra many-values.

raquelalvarezbanos: To summarize what we want:

- Case 1: Input workspace has M spectra (the number of bins is irrelevant) and flood workspace has M spectra with one bin in each spectrum. A script exemplifying what we want:

```
M=10

# Input workspace: 10 Spectra - 6 bins
x1 = [0,1,2,3,4,5,6] * M
y1 = [1,1,1,1,1,1,1] * M
ws1 = CreateWorkspace(x1, y1, NSpec=M)

# Flood workspace: 10 Spectra - 1 bin
x2 = [0] * M
y2 = [2] * M
ws2 = CreateWorkspace(x2, y2, NSpec=M)

ws3 = Divide(ws1, ws2)
```

- Case 2: Input workspace has M spectra (number of bins is irrelevant) and flood workspace has one spectrum with a single value. Example:

```
M=10

# Input workspace: 10 Spectra - 6 bins
x1 = [0,1,2,3,4,5,6] * M
y1 = [1,1,1,1,1,1,1] * M
ws1 = CreateWorkspace(x1, y1, NSpec=M)

# Flood workspace: 1 spectrum - 1 bin
x2 = [0]
y2 = [2]
ws2 = CreateWorkspace(x2, y2)

ws3 = Divide(ws1, ws2)
```

In both cases we can directly run `Divide` without the need to convert into the original form of 1-spectrum-many-values.

Framework/Algorithms/src/ReflectometryReductionOne.cpp
```diff
@@ -278,6 +278,50 @@ void ReflectometryReductionOne::init() {
...
+
+  if (decWS->getAxis(0)->unit()->unitID() != "Wavelength") {
+    result["DetectorEfficiencyCorrection"] = "Detector Efficiency Correction "
+      "workspace x-units must be in wavelength";
+  }
+  else if (decWS->getNumberHistograms() > 1 &&
+      decWS->getNumberHistograms() < runWS->getNumberHistograms()) {
+    result["DetectorEfficiencyCorrection"] = "The number of spectra in the "
+      "Detector Efficiency Correction workspace must either be 1 or equal to "
+      "the number of spectra in the input workspace";
```
raquelalvarezbanos: If the number of spectra must be either 1 or equal to the number of spectra in the input workspace, shouldn't we be checking:

```
else if (decWS->getNumberHistograms() > 1 
&&
decWS->getNumberHistograms() != runWS->getNumberHistograms())
```

Framework/Algorithms/src/ReflectometryReductionOne.cpp
```diff
@@ -278,6 +278,50 @@ void ReflectometryReductionOne::init() {
...
+    double firstVal = decWS->y(0)[0];
+    if (firstVal == 0) {
+      result["DetectorEfficiencyCorrection"] = "Detector Efficiency Correction "
+        "workspace cannot contain any zero values.";
+    } else {
+      for (size_t i = 1; i < decWS->getNumberHistograms(); i++) {
+        if (decWS->y(i)[0] != firstVal) {
+          result["DetectorEfficiencyCorrection"] = "All values in the Detector "
+            "Efficiciency Correction Workspace must be equal";
+          break;
```
raquelalvarezbanos: Values in the `DetectorEfficiencyCorrection` workspace don't need to be the same. In fact, in most cases they will be different, as the idea is that this workspace will be used to normalize each spectrum by a different normalization constant, taking into account possible differences in efficiencies. Besides, if all numbers were the same, why would we need to have more than one spectrum? One would be enough if that was the case. The above `else` is not what we want. 

Aso, we don't need to check that values are non-zero. If a user inputs a detector efficiency workspace full of zeroes, they will get `NaN`s in the output workspace. That should be enough to warn them that they are inputing something incorrect. The `if (firstVal == 0)` can be removed as well from the checks.

PranavBahuguna: After meeting with instrument scientists, we decided on the following:

- Flood workspace should be checked if it is in units of wavelength and if not, convert it to units of wavelength.
- ConvertUnits to be carried out on input workspace before splitting it into detectors and monitors.
- After both conversions, divide input workspace by flood workspace.

Framework/Algorithms/src/ReflectometryReductionOne.cpp
```diff
@@ -280,6 +278,37 @@ void ReflectometryReductionOne::init() {
...
+  MatrixWorkspace_sptr decWS = getProperty("DetectorEfficiencyCorrection");
+  MatrixWorkspace_sptr runWS = getProperty("InputWorkspace");
+
+  if (!decWS) {
+    return result;
+  }
+
+  if (decWS->getNumberHistograms() > 1 &&
+      decWS->getNumberHistograms() != runWS->getNumberHistograms()) {
+    result["DetectorEfficiencyCorrection"] =
```
raquelalvarezbanos: :+1:

Framework/Algorithms/src/ReflectometryReductionOne.cpp
```diff
@@ -280,6 +278,37 @@ void ReflectometryReductionOne::init() {
...
+        "The number of spectra in the "
+        "Detector Efficiency Correction workspace must either be 1 or equal to "
+        "the number of spectra in the input workspace";
+  } else if (decWS->blocksize() > 1) {
+    result["DetectorEfficiencyCorrection"] =
```
raquelalvarezbanos: :+1:

PranavBahuguna: In addition, we also decided to keep the `RegionOfDirectBeam` input property after all. This also means that the `Multidetector?` and `Perform Direct Beam Normalization` stages in the workflow diagram are preserved.

Framework/Algorithms/inc/MantidAlgorithms/ReflectometryWorkflowBase.h
```diff
@@ -52,18 +52,18 @@ class DLLExport ReflectometryWorkflowBase : public API::DataProcessorAlgorithm {
...
 
-  /// Convert the input workspace to wavelength, splitting according to the
-  /// properties provided.
+  /// Splits the input workspace into detector and monitor workpsaces according
+  /// to the properties provided.
   DetectorMonitorWorkspacePair
   toLam(Mantid::API::MatrixWorkspace_sptr toConvert,
         const std::string &processingCommands,
         const OptionalInteger monitorIndex, const MinMax &wavelengthMinMax,
         const OptionalMinMax &backgroundMinMax, const double &wavelengthStep);
```
raquelalvarezbanos: Could we rename this method? It no longer deals with wavelength conversion (that is now done outside) and can cause confusion in the future for new developers working in this area. The same happens with: 

`ReflectometryWorkflowBase::toLamMonitor()` 
and 
`ReflectometryWorkflowBase::toLamDetector()`

`toLam()` could be renamed to something like `splitDetectorsMonitors()` (as you have indicated in the description above). I am not sure about the names of the other two, perhaps something like:

`ReflectometryWorkflowBase::createDetectorWS()`
and
`ReflectometryWorkflowBase::createMonitorWS()`

would work.

raquelalvarezbanos: And perhaps indicate in the descriptions that workspaces are expected to be in wavelength.

raquelalvarezbanos:
I'm closing this PR as we first need to clarify some issues related to the algorithm and write a new version to which we'll add new properties, once it is properly tested (see https://github.com/mantidproject/mantid/issues/17429)

PR #17167, opened 2016-08-09
Fix segfault when using MD workspaces

### LINKED ISSUE #17165: segfault when algorithm dialog opens when MD workspace is present
The problem doesn't exist at d1be85cb95a75b9ad7b3b0ae4255e861e773499b but does exist at 1e56170f874b7d5b1ff5d58e66d89e77c8d6ed0f therefore was introduced with #16967 

To reproduce:
* Load a MD workspace `mdws = LoadMD('MAPS_MDEW.nxs')`
* Open an algorithm dialog which **doesn't** take a MD workspace as input. _e.g._ `Plus`

Result is:
```
MantidApplication-[Fatal] Unexpected exception: basic_string::_M_create
MantidApplication-[Fatal] Continue working.
tcmalloc: large alloc 7021238724389208064 bytes == (nil) @ 
MantidApplication-[Fatal] Unexpected exception: std::bad_alloc
MantidApplication-[Fatal] Continue working.
bin/launch_mantidplot.sh: line 49: 23857 Segmentation fault      (core dumped) LD_PRELOAD=${LOCAL_PRELOAD} TCMALLOC_RELEASE_RATE=${TCM_RELEASE} TCMALLOC_LARGE_ALLOC_REPORT_THRESHOLD=${TCM_REPORT} LD_LIBRARY_PATH=${LOCAL_LDPATH} QT_API=pyqt $VGLRUN$INSTALLDIR/MantidPlot $*
```

MD algorithms work fine _e.g._ `PlusMD`

It looks like MD workspaces are being included in the list of possible InputWorkspaces when they shouldn't be.

@AndreiSavici and @VickieLynch can confirm the problem.

cc: @raquelalvarezbanos @FedeMPouzols @martyngigg as you were also involved in the pull request

DavidFair:
**Description of work.**
As part of the changes following #16967 the returned type was changed from an unordered map to a vector. 
In `WorkspaceProperty.h` the loop erases various elements which invalidates the vector iterator which is subsequently dereferenced in the for loop crashing the program.

I've checked all the changed files from the previous PR for any other similar cases however this seems to be the only place where an iterator is invalidated in a loop

All test cases documented in #17165 have been fixed however if there are any more test cases I am happy to try them locally.

**To test:**
Open a MD workspace for example `MAPS_MDEW.nxs`.
Attempt to open the dialog box for an algorithm which does not take MD workspaces such as `Plus`
If the dialog box opens without crashing the test was successful.

Fixes #17165 .

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/API/inc/MantidAPI/WorkspaceProperty.h
+++ b/Framework/API/inc/MantidAPI/WorkspaceProperty.h
@@ -331,15 +331,14 @@ public:
       // Copy-construct a temporary workspace property to test the validity of
       // each workspace
       WorkspaceProperty<TYPE> tester(*this);
-      for (auto it = vals.begin(); it != vals.end();) {
-        // Remove any workspace that's not valid for this algorithm
-        if (!tester.setValue(*it).empty()) {
-          // Erase invalidates iterator so assign iterator returned from erase
-          it = vals.erase(
-              it++); // Post-fix so that it erase the previous when returned
-        } else
-          ++it;
-      }
+
+      // Remove any workspace that's not valid for this algorithm
+      auto eraseIter = remove_if(vals.begin(), vals.end(),
+                                 [&tester](const std::string &wsName) {
+                                   return !tester.setValue(wsName).empty();
+                                 });
+      // Erase everything past returned iterator afterwards for readability
+      vals.erase(eraseIter, vals.end());
       return vals;
     } else {
       // For output workspaces, just return an empty set
--- a/Framework/API/inc/MantidAPI/WorkspaceProperty.h
+++ b/Framework/API/inc/MantidAPI/WorkspaceProperty.h
@@ -334,7 +334,8 @@ public:
       for (auto it = vals.begin(); it != vals.end();) {
         // Remove any workspace that's not valid for this algorithm
         if (!tester.setValue(*it).empty()) {
-          vals.erase(
+          // Erase invalidates iterator so assign iterator returned from erase
+          it = vals.erase(
               it++); // Post-fix so that it erase the previous when returned
         } else
           ++it;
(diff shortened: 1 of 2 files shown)


martyngigg:
Should there be a unit test for this? It looks like it should be testable.

DavidFair:
I've added a unit test which tests the two concrete types available at `API` level. However I think its worth adding system tests to test  trying to use an unexpected workspace type on a `WorkspaceProperty` so I have opened #17175

peterfpeterson:
This fixed the problem I saw and commented in the [original issue](https://github.com/mantidproject/mantid/issues/17165#issuecomment-238560639). I'm not sure how to make that an automated test though.

AndreiSavici:
My issue seems to be fixed as well.

martyngigg:
@DavidFair I think your test here is fine and you don't need extra tests around the concrete types in `DataObjects`. As `WorkspaceProperty` is templated and there are no specializations then it will operate in the same manner regardless of the type. The test types in the API layer simply provide something that can be instantiated to decouple the interface from the implementation.

martyngigg:
@peterfpeterson: @DavidFair has added a unit test around `WorkspaceProperty` method that caused the issue, which I think should be sufficient to cover this.

rosswhitfield:
Fixed for me as well. 👍

FedeMPouzols:
This definitely fixes the issue. If everyone is happy, please some of the testers put a "squirrel stamp" on it.
I'm puzzled that none of the doc or system tests caught this. Are they too tidy and hence we are never stressing the system enough?

martyngigg:
:shipit:

PR #17171, opened 2016-08-09
Fix Python Exception for Algorithms Propagation under Python 3

martyngigg:
Fixes exception message propagation from Python algorithms in Python 3. The fixes were avoid using `PyBytes_AsString` when converting strings from Python to C++  [here](https://github.com/mantidproject/mantid/pull/17171/files#diff-97aa0ca20fef3541c75155f3f53915f0R32) & [here](https://github.com/mantidproject/mantid/pull/17171/files#diff-97aa0ca20fef3541c75155f3f53915f0R57). In Python 3 the Python strings are unicode objects not byte objects. 

I have also taken the opportunity to tidy up some of this code by introducing a single variadic `callMethod` wrapper that replaces the existing special cases.

**To test:**
- All Python 2 unit tests should still pass
- The `GetEiT0atSNSTest` unit test should now pass under Python 3

---
#### Reviewer

##### Code Review
##### Functional Tests

Allows for more perfect wrapping of boost::python::call_method without
having to define additional wrappers for more arguments.

### DIFF
--- a/Framework/PythonInterface/mantid/kernel/src/Environment/ErrorHandling.cpp
+++ b/Framework/PythonInterface/mantid/kernel/src/Environment/ErrorHandling.cpp
@@ -4,10 +4,15 @@
 #include "MantidPythonInterface/kernel/Environment/ErrorHandling.h"
 #include "MantidPythonInterface/kernel/Environment/GlobalInterpreterLock.h"
 
+#include <boost/python/extract.hpp>
+#include <boost/python/object.hpp>
 #include <frameobject.h> //Python
 
 #include <sstream>
 #include <stdexcept>
+
+using boost::python::extract;
+using boost::python::object;
 
 namespace Mantid {
 namespace PythonInterface {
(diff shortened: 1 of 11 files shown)


martyngigg:
@FedeMPouzols @rosswhitfield: This fixes the problem with the blank exception messages in Python 3 as noted in #16993

Framework/PythonInterface/mantid/kernel/src/Environment/ErrorHandling.cpp
```diff
@@ -64,8 +70,7 @@ void throwRuntimeError(const bool withTrace) {
   Py_XDECREF(traceback);
   Py_XDECREF(exception);
   Py_XDECREF(value);
-  // Raise this error as a C++ error
-  throw std::runtime_error(msg.str());
+  m_msg = builder.str();
```
FedeMPouzols: Just out of curiosity, is this constructor marked as `noexcept` just for performance reasons, or was it required for some other reason? I get the feeling that exceptions could be raised from here, producing an abrupt termination.

FedeMPouzols:
This is working well for me with Python 3.5 on debian. `DakotaChiSquaredTest` now passes, and I can see exception messages in the Python console. I'm now letting the other tests run but they should be happy as well (the ones that work with Python 3). MSVC seems to be picky about the `noexcept` in PythonException: http://builds.mantidproject.org/job/pull_requests-win7/9101/warnings43Result/new/source.1914648663137147177/#37

Framework/PythonInterface/mantid/kernel/src/Environment/ErrorHandling.cpp
martyngigg: Thanks for pointing this out. I needed `noexcept` for `what()` but i didn't think enough for the constructor. The string allocation could throw so I will remove the `noexcept` marker

rosswhitfield:
Can confirm `GetEiT0atSNSTest` and `DakotaChiSquaredTest` now pass. 👍

FedeMPouzols:
This definitely fixes the issue (see #16993) and in addition makes the code more concise and tidy.
I'd say we can :shipit:

PR #17174, opened 2016-08-09
Algorithms H-I updated with HistogramData

### LINKED ISSUE #17036: HistogramData Rollout Framework/Algorithm H-I
HistogramData rollout based on new Histrogram interface introduced to mantid. This issue deals with files in Framework/Algorithm starting with the letters H-I and their associated unit tests. See here.

Files to be checked (and their associated unit tests):
He3TubeEfficiency
HRPDSlabCanAbsorption
IdentifyNoisyDetectors
ImggTomographicReconstruction
IntegrateByComponent
Integration
InterpolatingRebin
InvertMask
IQTransform

dtasev:
## Description of work.

Refactored to use HistogramData:
- He3TubeEfficiency
- HRPDSlabCanAbsorption
- IdentifyNoisyDetectors
- ImggTomographicReconstruction
- IntegrateByComponent
- Integration
- InterpolatingRebin
- InvertMask
- IQTransform
## To test.

Run `AlgorithmsTest` to ensure that the tests pass.

Fixes #17036.

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/Algorithms/src/InterpolatingRebin.cpp
```diff
@@ -200,12 +201,9 @@ void InterpolatingRebin::outputYandEValues(
...
 **/
-void InterpolatingRebin::cubicInterpolation(const HistogramData::BinEdges &xOld,
-                                            const MantidVec &yOld,
-                                            const MantidVec &eOld,
-                                            const HistogramData::BinEdges &xNew,
-                                            MantidVec &yNew,
-                                            MantidVec &eNew) const {
+void InterpolatingRebin::cubicInterpolation(
+    const BinEdges &xOld, const HistogramY &yOld, const HistogramE &eOld,
+    const BinEdges &xNew, HistogramY &yNew, HistogramE &eNew) const {
```
SimonHeybrock: I would consider passing `Histogram` as arguments, instead of all the individual X, Y, and E vectors. That should also make things cleaner at the call site. Maybe `Histogram cubicInterpolation(const Histogram &histogram, const BinEdges &xNew)`.

Note that you can then also remove some of the lengths checks below, since `Histogram` guarantees this (if `xMode` is `BinEdges`, but you can make sure it is at the workspace level).

Framework/Algorithms/src/InterpolatingRebin.cpp
```diff
@@ -356,10 +354,10 @@ void InterpolatingRebin::cubicInterpolation(const HistogramData::BinEdges &xOld,
 */
 void InterpolatingRebin::noInterpolation(const HistogramData::BinEdges &xOld,
                                          const double yOld,
-                                         const MantidVec &eOld,
+                                         const HistogramE &eOld,
                                          const HistogramData::BinEdges &xNew,
-                                         MantidVec &yNew,
-                                         MantidVec &eNew) const {
+                                         HistogramY &yNew,
+                                         HistogramE &eNew) const {
```
SimonHeybrock: Change function signature, similar to my comment above?

Framework/Algorithms/test/IntegrationTest.h
```diff
@@ -447,11 +451,11 @@ class IntegrationTest : public CxxTest::TestSuite {
...
     for (int j = 0; j < 5; ++j) {
-      for (int k = 0; k < 5; ++k) {
-        space2D->dataX(j)[k] = 0.9 * k;
-        space2D->dataY(j)[k] = 2 * k + double(j);
-        space2D->dataE(j)[k] = 1.0;
-      }
+      auto &e = space2D->mutableE(j);
+      space2D->setPoints(j, Points(5, LinearGenerator(0, 0.9))); // assign X
+      space2D->setCounts(j,
+                         Counts(5, LinearGenerator(double(j), 2))); // assign Y
```
SimonHeybrock: You should not need the explicit constructor calls here, e.g., `space2D->setPoints(j, 5, LinearGenerator(0, 0.9));` should compile and do the same thing.

Framework/Algorithms/inc/MantidAlgorithms/He3TubeEfficiency.h
```diff
@@ -109,20 +109,21 @@ class DLLExport He3TubeEfficiency : public API::Algorithm {
...
-  const Geometry::ParameterMap *paraMap;
+  const Geometry::ParameterMap *m_paraMap;
   /// A lookup of previously seen shape objects used to save calculation time as
   /// most detectors have the same shape
-  std::map<const Geometry::Object *, std::pair<double, Kernel::V3D>> shapeCache;
+  std::map<const Geometry::Object *, std::pair<double, Kernel::V3D>>
+      m_shapeCache;
   /// Sample position
-  Kernel::V3D samplePos;
+  Kernel::V3D m_samplePos;
```
LamarMoore: Nice work bringing this in line with the standard.

Framework/Algorithms/src/HRPDSlabCanAbsorption.cpp
```diff
@@ -97,10 +97,10 @@ void HRPDSlabCanAbsorption::exec() {
 
   const size_t numHists = workspace->getNumberHistograms();
   const size_t specSize = workspace->blocksize();
-  //
+
   Progress progress(this, 0.91, 1.0, numHists);
   for (size_t i = 0; i < numHists; ++i) {
-    MantidVec &Y = workspace->dataY(i);
+    auto &Y = workspace->mutableY(i);
```
LamarMoore: Can this be moved below the `try{}catch{}` block? Closer to where it is used.

Framework/Algorithms/src/He3TubeEfficiency.cpp
```diff
@@ -89,54 +88,53 @@ void He3TubeEfficiency::init() {
...
     try {
       this->correctForEfficiency(i);
     } catch (std::out_of_range &) {
       // if we don't have all the data there will be spectra we can't correct,
       // avoid leaving the workspace part corrected
-      Mantid::MantidVec &dud = this->outputWS->dataY(i);
-      std::transform(dud.begin(), dud.end(), dud.begin(),
-                     std::bind2nd(std::multiplies<double>(), 0));
+      auto &dud = m_outputWS->mutableY(i);
+      dud.assign(dud.size(), 0);
```
LamarMoore: With the changes @SimonHeybrock made in [this](https://github.com/mantidproject/mantid/pull/17147) PR, you will be able to do things like `m_outputWS->mutableY(i) = 0;`. Look out for when that gets merged.

Framework/Algorithms/src/He3TubeEfficiency.cpp
```diff
@@ -147,60 +145,63 @@ void He3TubeEfficiency::exec() {
...
+
+  auto &yOut = m_outputWS->mutableY(spectraIndex);
+  auto &eOut = m_outputWS->mutableE(spectraIndex);
+  auto yOutItr = yOut.begin();
+  auto eOutItr = eOut.begin();
+
+  const auto wavelength = m_inputWS->points(spectraIndex);
+  auto wavelenItr = wavelength.cbegin();
+
+  for (; yOutItr != yOut.end(); ++yOutItr, ++eOutItr) {
```
LamarMoore: maybe use <algorithm> for these. Would have the added benefit of making the code more readable. Would be interested to see if there is an unacceptable performance hit when this is split out into to "loops" for Y and E.

Framework/Algorithms/src/IQTransform.cpp
```diff
@@ -215,9 +217,9 @@ void IQTransform::guinierRods(API::MatrixWorkspace_sptr ws) {
  * number
  */
 void IQTransform::guinierSheets(API::MatrixWorkspace_sptr ws) {
-  MantidVec &X = ws->dataX(0);
-  MantidVec &Y = ws->dataY(0);
-  MantidVec &E = ws->dataE(0);
+  auto &X = ws->mutableX(0);
+  auto &Y = ws->mutableY(0);
+  auto &E = ws->mutableE(0);
```
LamarMoore: This will also become easier with Simon's PR.

Framework/Algorithms/src/He3TubeEfficiency.cpp
SimonHeybrock: @LamarMoore I am wondering if it would make sense to add `operator*(Histogram &histogram, std::vector<double>)`. We have it for `double` in the PR that you mentioned above, maybe it also makes sense for a vector? Are you aware of any other cases where something like this is used?

In this case we could then use something like

```cpp
std::vector<double> effs(length);
std::generate(effs.begin(), effs.end(), detectorEfficiency);
outputWS->setHistogram(i, inputWS->histogram() * effs);
```

Framework/Algorithms/src/IQTransform.cpp
SimonHeybrock: @LamarMoore Right now we do not support cross-type operations in that PR, I think. That is, right now you cannot do

```cpp
ws->mutable(0) /= ws->mutableY(0);
```
To me it is not obvious whether this is something we want, or whether the added dangers outweigh the benefits, losing some type safety. What do you think?

LamarMoore: You're right. Maybe better to leave it as is.

Framework/Algorithms/inc/MantidAlgorithms/InterpolatingRebin.h
```diff
@@ -93,15 +96,20 @@ class DLLExport InterpolatingRebin : public Rebin {
...
   void noInterpolation(const HistogramData::BinEdges &xOld, const double yOld,
-                       const MantidVec &eOld,
-                       const HistogramData::BinEdges &xNew, MantidVec &yNew,
-                       MantidVec &eNew) const;
-  double estimateError(const MantidVec &xsOld, const MantidVec &esOld,
+                       const HistogramData::HistogramE &eOld,
+                       const HistogramData::BinEdges &xNew,
+                       HistogramData::HistogramY &yNew,
+                       HistogramData::HistogramE &eNew) const;
+  double estimateError(const HistogramData::HistogramX &xsOld,
```
DimitarTasev: @SimonHeybrock Is there a replacement function for this in HistogramData?

SimonHeybrock: @DimitarTasev No, for now `HistogramData` only contains basics. I think Lamar wanted to start a list of things we might consider adding, maybe append this to that list (but in any case that would not be implemented as part of this pull request).

Framework/Algorithms/src/InterpolatingRebin.cpp
```diff
@@ -200,27 +201,18 @@ void InterpolatingRebin::outputYandEValues(
...
 *x-values
 **/
-void InterpolatingRebin::cubicInterpolation(const HistogramData::BinEdges &xOld,
-                                            const MantidVec &yOld,
-                                            const MantidVec &eOld,
-                                            const HistogramData::BinEdges &xNew,
-                                            MantidVec &yNew,
-                                            MantidVec &eNew) const {
-  // Make sure y and e vectors are of correct sizes
+void InterpolatingRebin::cubicInterpolation(const Histogram &oldHistogram,
```
DimitarTasev: @SimonHeybrock I tried doing it the way you recommended in the previous comment `Histogram cubicInterpolation(const Histogram &histogram, const BinEdges &xNew)`, but I wan't sure how to create Y and E after that, because they base their size on `outputW->mutableY` and `outputW->mutableE`

Framework/Algorithms/src/He3TubeEfficiency.cpp
```diff
@@ -168,30 +167,29 @@ void He3TubeEfficiency::correctForEfficiency(std::size_t spectraIndex) {
...
-    ++wavelenItr;
-  }
+  // reset wavelength iterator
+  std::transform(eValues.cbegin(), eValues.cend(), wavelength.cbegin(),
+                 eOut.begin(), [&](const double e, const double wavelen) {
+                   double effcorr =
+                       this->detectorEfficiency(exp_constant * wavelen, scale);
+
+                   return e * effcorr;
+                 });
```
SimonHeybrock: You do not want to compute `detectorEfficiency` twice. This is probably a significant cost overhead and will deteriorate the performance.

Framework/Algorithms/src/InterpolatingRebin.cpp
```diff
@@ -201,24 +201,18 @@ void InterpolatingRebin::outputYandEValues(
...
+                                            const BinEdges &xNew,
+                                            HistogramY &yNew,
+                                            HistogramE &eNew) const {
+  const auto &yOld = oldHistogram.y();
+
   const size_t size_old = yOld.size();
-  if (size_old == 0)
-    throw std::runtime_error("Empty spectrum found aborting");
```
SimonHeybrock: You probably want to keep the zero-length check.

Framework/Algorithms/src/InterpolatingRebin.cpp
SimonHeybrock: @DimitarTasev  For example:

```cpp
Histogram InterpolatingRebin::cubicInterpolation(const Histogram &oldHistogram, const BinEdges &xNew) const {
// ...
   if (!canInterpol) {
     if (VectorHelper::isConstantValue(yOld.rawData())) {
       // this copies the single y-value into the output array, errors are still
       // calculated from the nearest input data points
       // this is as much as we need to do in this (trival) case
      return noInterpolation(oldHistogram, xNew);
     } else { // some points are two close to the edge of the data
// ...
}
```

The size of `outputW->mutableY` and `outputW->mutableE` is simply the size of `xNew` minus 1.

Framework/Algorithms/src/InterpolatingRebin.cpp
```diff
@@ -352,16 +349,16 @@ void InterpolatingRebin::cubicInterpolation(
...
 * input data points
 */
-void InterpolatingRebin::noInterpolation(const HistogramData::BinEdges &xOld,
-                                         const double yOld,
-                                         const HistogramE &eOld,
-                                         const HistogramData::BinEdges &xNew,
-                                         HistogramY &yNew,
+void InterpolatingRebin::noInterpolation(const Histogram &oldHistogram,
+                                         const BinEdges &xNew, HistogramY &yNew,
                                          HistogramE &eNew) const {
```
SimonHeybrock: Remove `yNew` and `eNew` arguments, return `Histogram` instread.

Framework/Algorithms/src/InterpolatingRebin.cpp
```diff
@@ -352,16 +349,16 @@ void InterpolatingRebin::cubicInterpolation(
...
-  yNew.assign(yNew.size(), yOld);
-  for (MantidVec::size_type i = 0; i < eNew.size(); ++i) {
-    eNew[i] = estimateError(xOld.rawData(), eOld, xNew[i]);
-  }
+  // same size as in oldHistogram
```
SimonHeybrock: ??

Framework/Algorithms/test/IntegrationTest.h
```diff
@@ -452,9 +452,8 @@ class IntegrationTest : public CxxTest::TestSuite {
 
     for (int j = 0; j < 5; ++j) {
       auto &e = space2D->mutableE(j);
-      space2D->setPoints(j, Points(5, LinearGenerator(0, 0.9))); // assign X
-      space2D->setCounts(j,
-                         Counts(5, LinearGenerator(double(j), 2))); // assign Y
+      space2D->setPoints(j, 5, LinearGenerator(0, 0.9));       // assign X
+      space2D->setCounts(j, 5, LinearGenerator(double(j), 2)); // assign Y
```
SimonHeybrock: A while ago we decided to remove all C-style casts. Always use `static_cast` instead.

Framework/Algorithms/test/InterpolatingRebinTest.h
```diff
@@ -276,12 +276,16 @@ class InterpolatingRebinTest : public CxxTest::TestSuite {
...
     for (; i < nBins; i++, j += 1.5) {
-      retVal->mutableX(0)[i] = j * 0.5;
-      retVal->mutableY(0)[i] = j;
-      retVal->mutableE(0)[i] = retVal->y(0)[i] / 8;
+      x[i] = j * 0.5;
+      y[i] = j;
+      e[i] = yConst[i] / 8;
     }
-    retVal->mutableX(0)[i] = j * 0.5;
+    x[i] = j * 0.5;
```
SimonHeybrock: `retVal->setHistogram(0, BinEdges(nBins + 1, LinearGenerator(0.5, 0.75), Counts(nBins, LinearGenerator(1.0, 1.0), CountStandardDeviations(nBins, LinearGenerator(1.0/8.0, 1.0/8.0))`?

Framework/Algorithms/test/InterpolatingRebinTest.h
```diff
@@ -293,14 +297,14 @@ class InterpolatingRebinTest : public CxxTest::TestSuite {
 
     // the first histogram has all zeros
     retVal->setBinEdges(0, nBins + 1, LinearGenerator(0.0, 1.0));
-    retVal->dataY(0).assign(retVal->dataY(0).size(), 0.0);
-    retVal->dataE(0).assign(retVal->dataE(0).size(), 0.0);
+    retVal->mutableY(0).assign(retVal->mutableY(0).size(), 0.0);
+    retVal->mutableE(0).assign(retVal->mutableE(0).size(), 0.0);
```
SimonHeybrock: `retVal->mutableY(0) = 0.0;`

Framework/Algorithms/test/InterpolatingRebinTest.h
```diff
@@ -293,14 +297,14 @@ class InterpolatingRebinTest : public CxxTest::TestSuite {
...
 
     // the second has NAN values
     retVal->setBinEdges(1, nBins + 1, LinearGenerator(0.0, 1.0));
-    retVal->dataY(1).assign(retVal->dataY(1).size(),
+    retVal->mutableY(1).assign(retVal->mutableY(1).size(),
```
SimonHeybrock: Here too use `operator=`

Framework/Algorithms/src/InterpolatingRebin.cpp
DimitarTasev: @SimonHeybrock I have a problem, which I figured out where it's happening, but I have no idea why. In the state it's currently on the branch, the `WorkspaceHelpers::makeDistribution(outputW, false);`, [found here](https://github.com/mantidproject/mantid/blob/815866e9d46bb390e78960e6d2677f78408dfa44/Framework/Algorithms/src/InterpolatingRebin.cpp#L108) changes the `outputW` values to the ones seen on the test, however after refactoring([refactored code here](http://pastebin.com/115Wc3VA)) that line doesn't seems to change anything in the `outputW` and thus the tests fail. Any ideas why that's happening?

SimonHeybrock: If it does not do anything, it implies that the workspace is already in the expected state, i.e., the histograms have `XMode::Counts` (instead of `XMode::Frequencies`). See if you used `setCounts()` somewhere where you should not have.

martyngigg:
I've finally managed to track the segfault down to something to do with the changes in `InterpolatingRebin`. If I revert only those and keep the rest then the segfaults do not occur. I now need to find out why this is happening.

Framework/Algorithms/src/InterpolatingRebin.cpp
```diff
@@ -341,29 +334,32 @@ void InterpolatingRebin::cubicInterpolation(const HistogramData::BinEdges &xOld,
...
+                                              const BinEdges &xNew) const {
+  Histogram newHistogram{xNew, Frequencies(xNew.size() - 1)};
+  auto &yNew = newHistogram.mutableY();
+  auto &eNew = newHistogram.mutableE();
+
+  yNew.assign(yNew.size(), oldHistogram.y().front());
+
+  const auto &xOldData = oldHistogram.x().rawData();
+  const auto &eOld = oldHistogram.e();
+  std::transform(xNew.cbegin(), xNew.cend(), eNew.begin(),
```
SimonHeybrock: @martyngigg Probably it is this line. `xNew` in longer than `eNew`.

Framework/Algorithms/src/InterpolatingRebin.cpp
```diff
@@ -187,40 +180,30 @@ void InterpolatingRebin::outputYandEValues(
...
-    throw std::runtime_error(
-        "y and error vectors must be of same size & 1 shorter than x");
+    throw std::runtime_error("Empty spectrum found, aborting!");
+
+  const size_t size_new = xNew.size() - 1; // -1 because BinEdges
 
   // get the bin centres of the input data
   std::vector<double> xCensOld(size_new);
-  VectorHelper::convertToBinCentre(xOld.rawData(), xCensOld);
+  VectorHelper::convertToBinCentre(oldHistogram.x().rawData(), xCensOld);
```
SimonHeybrock: `auto xCensOld = oldHistogram.points()`

Framework/Algorithms/src/InterpolatingRebin.cpp
```diff
@@ -187,40 +180,30 @@ void InterpolatingRebin::outputYandEValues(
...
   // the centres of the output data
   std::vector<double> xCensNew(size_new);
   VectorHelper::convertToBinCentre(xNew.rawData(), xCensNew);
```
SimonHeybrock: `Points xCensNew(xNew);`

martyngigg:
Thanks, @SimonHeybrock. @DimitarTasev Can you take a look at fixing these issues?

Framework/Algorithms/src/He3TubeEfficiency.cpp
```diff
@@ -147,60 +146,59 @@ void He3TubeEfficiency::exec() {
...
+      yValues.cbegin(), yValues.cend(), effCorrection.cbegin(), yOut.begin(),
+      [&](const double y, const double effcorr) { return y * effcorr; });
+
+  const auto &eValues = m_inputWS->e(spectraIndex);
+  auto &eOut = m_outputWS->mutableE(spectraIndex);
+
+  // reset wavelength iterator
+  std::transform(
+      eValues.cbegin(), eValues.cend(), effCorrection.cbegin(), eOut.begin(),
+      [&](const double e, const double effcorr) { return e * effcorr; });
```
SimonHeybrock: @LamarMoore and me just discussed this. We concluded that we want to provide `operator*` for multiplying `Histogram` by `std::vector<double>`, then most of this reduces to a single line. Keep in touch with @LamarMoore to see when this is implemented. Probably it will be a follow up to https://github.com/mantidproject/mantid/pull/17193.

Framework/Algorithms/src/He3TubeEfficiency.cpp
```diff
@@ -473,16 +463,33 @@ void He3TubeEfficiency::execEvent() {
...
+* @param events :: the list of events to correct
+* @param expval :: the value of the exponent for the detector efficiency
+*/
 template <class T>
 void He3TubeEfficiency::eventHelper(std::vector<T> &events, double expval) {
   const double scale = this->getProperty("ScaleFactor");
-  typename std::vector<T>::iterator it;
-  for (it = events.begin(); it != events.end(); ++it) {
+
+  for (auto it = events.begin(); it != events.end(); ++it) {
```
SimonHeybrock: `for(auto &event : events) {` would be even simpler.

Framework/Algorithms/src/IQTransform.cpp
```diff
@@ -118,8 +117,10 @@ void IQTransform::exec() {
   outputWS->setYUnit("");
   // Copy the data over. Assume single spectrum input (output will be).
   outputWS->setPoints(0, tmpWS->points(0));
-  MantidVec &Y = outputWS->dataY(0) = tmpWS->dataY(0);
-  outputWS->dataE(0) = tmpWS->dataE(0);
+  outputWS->setSharedY(0, tmpWS->sharedY(0));
+  outputWS->setSharedE(0, tmpWS->sharedE(0));
+
+  auto &Y = outputWS->mutableY(0);
```
SimonHeybrock: You are triggering a copy even in the cases where the background is not subtracted. Maybe simply use

```cpp
if (backround > 0.0)
  outputWS->mutableY(0) -= background;
```

below.

(Actually this is really minor here, since we are simply dealing with a single spectrum, but I looks slightly more readable, maybe?)

martyngigg:
I would suggest moving this to release `3.10`. There's no benefit it doing these refactorings this late in the release cycle.

martyngigg:
@DTasev @LamarMoore I think this PR got overlooked from last release. I suggest we wait until the release branch has been created and then we can fix any remaining problems if that's okay?

NickDraper:
@DanNixon Can you pick this up as the developer

Framework/Algorithms/src/He3TubeEfficiency.cpp
peterfpeterson: That PR was merged in September 2016. This can be changed to `eOut = eValues * effCorrection;`

Framework/Algorithms/src/He3TubeEfficiency.cpp
```diff
@@ -267,10 +265,10 @@ void He3TubeEfficiency::getDetectorGeometry(const Geometry::IDetector &det,
         "The algorithm works for instruments with one-to-one "
         "spectra-to-detector maps only!");
   }
-  std::map<const Geometry::IObject *,
-           std::pair<double, Kernel::V3D>>::const_iterator it =
-      this->shapeCache.find(shape_sptr.get());
-  if (it == this->shapeCache.end()) {
+  // std::map<const Geometry::Object *, std::pair<double,
```
peterfpeterson: Remove commented out code

peterfpeterson (approved):
The improvements are appreciated. There is still a small improvement suggested, but it can be done at another time.

PR #17202, opened 2016-08-11
Check container size to prevent calling memcpy on a nullptr.

quantumsteve:
Description of work.

the [undefined behavior sanitizer](http://builds.mantidproject.org/view/Static%20Analysis/job/undefined_behavior_sanitizer_6.1/2/warnings24Result/new/category.67232232/) is reporting the following error

```
../Framework/PythonInterface/mantid/kernel/src/Converters/CloneToNumpy.cpp:68:3: runtime error: null pointer passed as argument 2, which is declared to never be null
```

std::memcpy is being called when the length of the vector is 0, leading to undefined behavior. I think the best option is the check if the vector is empty and call a another function to return an empty container. 

**To test:**

Review changes and, if possible, run tests with  `-DWITH_UBSAN=ON`.

No GitHub issue.

Should this be in the release notes? 

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/PythonInterface/mantid/kernel/src/Converters/CloneToNumpy.cpp
```diff
@@ -65,6 +81,7 @@ PyObject *cloneND(const ElementType *carray, const int ndims,
   }
   void *arrayData = PyArray_DATA(nparray);
```
martyngigg: Is there a reason we can't just check for the zero-length here and return rather than requiring a separate function?

martyngigg:
I ran this with `UBSAN=ON` under gcc 5.3 and the errors resulting from calling `std::memcpy` are now gone as expected.

There's just the question above about the separate function to address but then this can be marked for merging.

Framework/PythonInterface/mantid/kernel/src/Converters/CloneToNumpy.cpp
quantumsteve: In hindsight, adding another member function looks like premature optimization.

quantumsteve:
@martyngigg Is this branch ready to :squirrel:?

martyngigg:
Yes - it looks good now.

PR #17213, opened 2016-08-12
IndirectCorrections - post plot and save

### LINKED ISSUE #17021: Indirect Corrections - Post execution plotting and saving
Indirect Interfaces should be able to plot and save after the algorithm is run by clicking a plot and save button rather than having to specify plot and save options before the interface is executed.

This issue should cover the post plotting and saving implementation for the `Indirect > Corrections` interface

* [x] Container Subtraction
* [x] Calculate Paalman Pings
* [x] Apply Paalman Pings
* [ ] Absorption blocked by #17227

louisemccann:
The plotting and saving of the results workspace has been changed to after the execution of the algorithm for all IndirectCorrections tabs.

**To test:**
Files to test with: \olympic\babylon5\Public\Louise McCann\Corrections
For each tab load relevant files and run algorithm. Check each plot option plots a suitable graph and that the appropriate workspaces have been saved in your save directory.
- [ ] Container Subtraction
- [ ] Calculate Paalman Pings
- [ ] Apply Paalman Pings
- [ ] Absorption

Fixes #17021 

---
#### Reviewer

##### Code Review
##### Functional Tests


raquelalvarezbanos:
@louisemccann Is there a reason why you have assigned yourself this PR? With PR's we normally leave them unassigned, so other developers can assign themselves for testing. I have assigned myself for testing, let me know if you are happy with that.

louisemccann:
@raquelalvarezbanos  sorry, I made the mistake of assigning myself when creating the pull request

raquelalvarezbanos:
It's OK, just wanted to make sure that it was OK for me to test it

MantidQt/CustomInterfaces/src/Indirect/AbsorptionCorrections.cpp
```diff
@@ -301,29 +298,44 @@ void AbsorptionCorrections::algorithmComplete(bool error) {
...
+ * Handle saving of workspace
+ */
+void AbsorptionCorrections::saveClicked() {
+
+  checkADSForPlotSaveWorkspace(m_pythonExportWsName, false);
+  addSaveWorkspace(QString::fromStdString(m_pythonExportWsName));
+  bool keepCorrectionFactors = m_uiForm.ckKeepFactors->isChecked();
+  checkADSForPlotSaveWorkspace(m_outputFactorsWsName.toStdString(), false);
+  if (keepCorrectionFactors)
+    addSaveWorkspace(m_outputFactorsWsName);
```
raquelalvarezbanos: Just a very minor comment: if `keepCorrectionFactors == False` I assume we don't even need to check if the corresponding workspace is in the ADS, so we could avoid this check by moving line 313 inside the `if` statment.

MantidQt/CustomInterfaces/src/Indirect/AbsorptionCorrections.cpp
```diff
@@ -301,29 +298,44 @@ void AbsorptionCorrections::algorithmComplete(bool error) {
...
+    if (m_uiForm.ckUseCanCorrections->isChecked()) {
+      plotCorr.push_back(QString::fromStdString(outputFactorsWsName) + "_acc");
+      QString shiftedWs = QString::fromStdString(
+          m_absCorAlgo->getPropertyValue("CanWorkspace"));
+      plotData.push_back(shiftedWs);
     }
-    IndirectTab::plotSpectrum(plotData, 0);
+    plotSpectrum(plotCorr, 0);
   }
+  IndirectTab::plotSpectrum(plotData, 0);
```
raquelalvarezbanos: Is there a reason why sometimes we use `IndirectTab::method()` and sometimes just `method()`? For instance in this case, as far as I understand `plotSpectrum()` is defined and implemented in `IndirectTab` so there is no need to specify the base class here. It is OK if you want to explicitly say that `plotSpectrum` is implemented there, but if so I would do that everywhere for consistency. I've seen that with `checkADSForPlotSaveWorkspace()` too.

raquelalvarezbanos:
I've been able to run `Container Subtraction`, `Calculate Paalman Pings` and `Absorption` with the files provided. However, when I try to run `Apply Paalman Pings` I get an unexpected exception:

```
Invalid value for property Target (string) "" : select a value
```

I may be doing something wrong (although we should properly deal with this), these are the options I am using:

- Input: "irs26176_graphite002_red.nxs"
- Container: "irs26174_graphite002_red.nxs"
- Geometry: "Flat Plate"
- The rest of the options disabled

raquelalvarezbanos:
In debug mode, I get an ASSERT failure when running `Container Subtraction`. My guess is that the problem is related to the preview plot (I don't think is related to changes in this PR), therefore I think this should be looked into in a separate issue . I've created https://github.com/mantidproject/mantid/issues/17326

MantidQt/CustomInterfaces/src/Indirect/ContainerSubtraction.cpp
```diff
@@ -453,12 +441,41 @@ void ContainerSubtraction::absCorComplete(bool error) {
...
 
+/**
+ * Handles saving of workspace
+ */
+void ContainerSubtraction::saveClicked() {
+
+  // Check workspace exists
+  IndirectTab::checkADSForPlotSaveWorkspace(m_pythonExportWsName, false);
+  addSaveWorkspaceToQueue(QString::fromStdString(m_pythonExportWsName));
+}
```
raquelalvarezbanos: I have a question about the intended behaviour of `Save Results`: from the issue description I understand that when a user has run the reduction, they should be able to save the result just by clicking on `Save Result`. However, the result is only saved when `Run` is clicked again. The reason is that in the `saveClicked()` slot above, we are just adding the instruction to save the workspace to the list of algorithms to run, but those will not be run until the program reaches the first `m_batchAlgoRunner->executeBatchAsync()` which happens in line L176 in `ContainerSubtraction.cpp`.

MantidQt/CustomInterfaces/src/Indirect/CalculatePaalmanPings.cpp
```diff
@@ -425,5 +421,30 @@ void CalculatePaalmanPings::addShapeSpecificCanOptions(IAlgorithm_sptr alg,
...
+/**
+ * Handles saving of workspace
+ */
+void CalculatePaalmanPings::saveClicked() {
+
+  checkADSForPlotSaveWorkspace(m_pythonExportWsName, false);
+  const auto outputWs = QString::fromStdString(m_pythonExportWsName);
+  addSaveWorkspaceToQueue(outputWs);
+  m_batchAlgoRunner->executeBatchAsync();
+}
```
raquelalvarezbanos: OK, I see here that the intended behaviour is what I thought. You just forgot to add the last line in `void ContainerSubtraction::saveClicked()`.

MantidQt/CustomInterfaces/src/Indirect/AbsorptionCorrections.cpp
```diff
@@ -114,21 +117,15 @@ void AbsorptionCorrections::run() {
 
   // Set the correction workspace to keep the factors if desired
   bool keepCorrectionFactors = m_uiForm.ckKeepFactors->isChecked();
-  QString outputFactorsWsName = outputBaseName + "_" + sampleShape + "_Factors";
+  QString m_outputFactorsWsName =
```
raquelalvarezbanos: Do we want to define a new `m_outputFactorWsName` local variable here or should we just use the member variable already defined in the `.h`?

raquelalvarezbanos:
A summary of what I've tested:

- `Container Subtraction`: `Plot` works but `Save Result` doesn't.
- `Calculate Paalman Pings`: `Plot` works (it produces an empty plot but I guess that's because I am not running the reduction with the right parameters) and `Save Result` works too.
- `Apply Paalman Pings`: I get an unexpected exception when running the reduction (I may be using the wrong parameters) I haven't looked into the problem.
- `Absorption`: `Plot` works but `Save Result` does not work properly (see one of my comments above).

raquelalvarezbanos:
Should these changes be mentioned in the release notes?

raquelalvarezbanos:
All tabs work as expected now. @louisemccann : should this be mentioned in the release notes?

raquelalvarezbanos:
Release notes will be updated once the task is complete for all interfaces and tabs.

:shipit:

MantidQt/CustomInterfaces/src/Indirect/ContainerSubtraction.cpp
```diff
@@ -453,12 +441,43 @@ void ContainerSubtraction::absCorComplete(bool error) {
...
+
+/**
+ * Handles Mantid plotting of workspace
+ */
+void ContainerSubtraction::plotClicked() {
+  QString plotType = m_uiForm.cbPlotOutput->currentText();
+
+  if (checkADSForPlotSaveWorkspace(m_pythonExportWsName, true)) {
+
+    if (plotType == "Spectra" || plotType == "Both")
```
FedeMPouzols: If you see these string comparisons cropping up in several places, probably it's worth defining an enum type for the plot type. I think there's also at least two very similar "plotClicked()" methods, maybe you can share most of its code rather than duplicating? It's a small difference, but as this stuff is hard to test and is usually tested manually, having these may help prevent bugs in the plotting that can easily remain unnoticed.

PR #17246, opened 2016-08-16
Invisible workspaces option bug fix 

### LINKED ISSUE #6798: Workspace list shows invisible workspaces when next visible is created
This issue was originally [TRAC 5952](http://trac.mantidproject.org/mantid/ticket/5952)

To reproduce:
* Go to View->Preferences->Mantid->Options and uncheck "show invisible workspaces"
* Click Load;
* Pick a file and prefix the suggested workspace name with ```__```
* The algorithm will run but there should be nothing in the list
* Click load again and run the same thing but without the prefix. The invisible workspace will appear alongside the visible after the algorithm completes

PranavBahuguna:
Description of work.

As stated in the original issue, creating hidden workspaces then checking/unchecking the "show hidden workspaces" option has no effect on those workspaces:
- If "show hidden workspaces" was already checked, unchecking it doesn't cause hidden workspaces to disappear.
- If "show hidden workspaces" was already unchecked, checking it doesn't cause hidden workspaces to appear.
- The work done has fixed this so clicking on this option produces the intended effect.

Fixes #6798.

**Testing Instructions**:
- Click 'Load' in the workspaces dock.
- Pick a file and make it hidden by prefixing the output workspace name with __
- There will either be a hidden workspace in the dock or no workspace depending on whether "show hidden workspaces" was already checked or not. Either way, go to View->Preferences->Mantid->Options and check/uncheck "show invisible workspaces". Then click 'Apply' or 'OK'.
- The state of the hidden workspaces should now be altered.
#### Reviewer

##### Code Review
##### Functional Tests


Framework/Kernel/inc/MantidKernel/GitHubApiHelper.h
```diff
@@ -0,0 +1,62 @@
...
+  void addAuthenticationToken() {
+    addHeader("Authorization",
+              "token 8ec7afc857540ee60af78cba1cf7779a6ed0b6b9");
+  }
+};
+
+} // namespace Kernel
+} // namespace Mantid
+
+#endif /* MANTID_KERNEL_GitHubApiHelper_H_ */
```
raquelalvarezbanos: Was this file added by mistake?

PranavBahuguna: I don't know actually. It must have been missing somehow before and got included now. What should I do with it, as its not part of my actual work?

raquelalvarezbanos: I think it should be removed. I've seen that file appearing in my local repository a couple of times but I removed it before commiting

MantidPlot/src/Mantid/MantidDock.cpp
```diff
@@ -404,8 +404,8 @@ void MantidDockWidget::setItemIcon(QTreeWidgetItem *item,
 */
 void MantidDockWidget::updateTree() {
   // do not update until the counter is zero
-  if (m_updateCount.deref())
-    return;
+  // if (m_updateCount.deref())
+  //   return;
```
raquelalvarezbanos: I am not sure that this is the best fix (although I am not familiar enough with `MantidDock` to know what the best fix would be):

As the comment above says, the `if` above avoids updating the tree until the counter is zero. I have been debugging the code with this `if` block enabled (and keeping you other changes) and I think the problem appears when `m_updateCount ==0`. In that case, according to Qt documentation, `m_updateCount.deref()` will:

```
Atomically decrement the value of this QAtomicInt. Return true if the new value is non-zero, false otherwise
```

but as the new value is `-1` it will return `true`. Instead of removing the `if` block, I've tried (keeping your other changes):

```
  // do not update until the counter is zero
	m_updateCount.deref();
	if (int(m_updateCount) > 0) {
		return;
	}
```

and that seems to work too. However I can't be sure that it's the right think to do, therefore I am leaving this PR unassigned so that other developer can test it.

martyngigg: I agree with Raquel here, we need to keep the updateCount check. This is in place to stop updates happening too frequently during heavy processing.

I guess the issue here is the update is called without the ADS having called `MantidDockWidget::incrementUpdateCount()` so the counts en up out of sync.

I wonder if a simpler way to avoid this would to be separate out the update count part from the actual updating. If we had a new `updateTreeOnADSUpdate()` slot that did:
```c++
void MantidDockWidget::updateTreeOnADSUpdate() {
  // do not update until the counter is zero
  if (m_updateCount.deref())
    return;
updateTree();
}
```

we could remove the update count check from `updateTree` and still just call this from the `ConfigDialog`. We would then just need to update the signal/slot connection on line 139 to use the new method and it should avoid the scenario of the counter going below zero?

martyngigg:
I'm going to move this back to `3.9` as the bug was submitted by me and not a user.

martyngigg:
I've retested this and I can confirm it works as intended. The workspaces prefixed with `__` are hidden/shown depending on the values of the config option. This is also correctly persisted across sessions.

martyngigg (approved):
:shipit:

PR #17257, opened 2016-08-17
HistogramRebin RebinCounts/Frequencies implementation

### LINKED ISSUE #17225: HistogramData Rebin Free Function Implementation
Currently, there are many helper functions within the Kernel::VectorHelper module which operate on std::vectors. Some of this functionality is suitable for addition to the HistogramData module. This ticket deals with adding the Rebin functions to HistogramData.

LamarMoore:
Description of work.

**To test:**
- Code Review
- Ensure unit tests are sensible
- Compare performance tests with those of VectorHelper

Fixes #17225.

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/HistogramData/test/HistogramRebinTest.h
```diff
@@ -0,0 +1,324 @@
...
+
+  void testExecRebinFrequency() {
+    TS_ASSERT_THROWS_NOTHING(rebinFrequencies(
+        getFrequencyHistogram(), BinEdges(10, LinearGenerator(0, 0.5))));
+  }
+
+  void testRebinFrequenciesFailModes() {
+    BinEdges edges(5, LinearGenerator(0, 2));
+    // Incorrect YMode
+    TS_ASSERT_THROWS(rebinFrequencies(getCountsHistogram(), edges),
```
SimonHeybrock: Don't need this, just provide a single `rebin()` that branches internally to call either `rebinCounts()` or `rebinFrequencies()`.

Framework/HistogramData/test/HistogramRebinTest.h
```diff
@@ -0,0 +1,324 @@
...
+    TS_ASSERT_THROWS(rebinCounts(getCountsHistogram(), edges),
+                     std::runtime_error);
+    TS_ASSERT_THROWS(rebinFrequencies(getFrequencyHistogram(), edges),
+                     std::runtime_error);
+  }
+
+  void testRebinFailsInputBinEdgesInvalid() {
+    std::vector<double> binEdges{1, 2, 3, 3, 5, 7};
+    Histogram hist(BinEdges(std::move(binEdges)), Counts(5, 10));
+    BinEdges edges{1, 2, 3, 4, 5, 6};
```
SimonHeybrock: Since `VectorHelper::rebin()` failed differently depending on whether the bad bin edges coincided in input and output or not, maybe we should test both cases here and in the test above this one.

Framework/HistogramData/test/HistogramRebinTest.h
```diff
@@ -0,0 +1,324 @@
...
+    Histogram hist(BinEdges{0, 1, 2}, Counts{10, 10});
+    Histogram histFreq(BinEdges{0, 1, 2}, Frequencies{12, 12});
+    BinEdges edges{0, 0.5, 1, 1.5, 2};
+
+    auto outCounts = rebinCounts(hist, edges);
+    auto outFreq = rebinFrequencies(histFreq, edges);
+
+    for (size_t i = 0; i < outCounts.y().size(); i++) {
+      TS_ASSERT_EQUALS(outCounts.y()[i], 5.0);
+      TS_ASSERT_EQUALS(outFreq.y()[i], 12.0);
```
SimonHeybrock: All these tests should check E as well.

Framework/HistogramData/test/HistogramRebinTest.h
```diff
@@ -0,0 +1,324 @@
...
+    auto outFreq = rebinFrequencies(histFreq, edges);
+
+    TS_ASSERT_EQUALS(outCounts.y()[0], hist.y()[0] + (hist.y()[1] / 4));
+    TS_ASSERT_EQUALS(outCounts.y()[1], hist.y()[1] / 2);
+    TS_ASSERT_EQUALS(outCounts.y()[2], (hist.y()[1] / 4) + hist.y()[2]);
+
+    TS_ASSERT_EQUALS(outFreq.y()[0], (histFreq.y()[0] + histFreq.y()[1]) / 2);
+    TS_ASSERT_EQUALS(outFreq.y()[1], histFreq.y()[1]);
+    TS_ASSERT_EQUALS(outFreq.y()[2], (histFreq.y()[1] + histFreq.y()[2]) / 2);
+  }
```
SimonHeybrock: Maybe we should also add tests for start and end (the tests above cover the internal range):

```
|   |   |   |
  |   |   |
```

```
  |   |   |
|   |   |   |
```


```
|   |   |   |   |   |
        |   |
```
etc.?

Framework/HistogramData/src/HistogramRebin.cpp
```diff
@@ -132,30 +132,30 @@ Histogram rebinFrequencies(const Histogram &input, const BinEdges &binEdges) {
       } else {
         factor = 1 / nwidth;
         ynew[inew] *= factor;
-        enew[inew] = sqrt(enew[inew]) * factor;
+        enew[inew] = sqrt(enew[inew] * factor);
```
LamarMoore: @SimonHeybrock  I made this change and seem to be getting the correct values for frequencyStandardDeviations.

SimonHeybrock: The original looks correct to me. Could you check this again please?

Framework/HistogramData/src/HistogramRebin.cpp
```diff
@@ -0,0 +1,157 @@
...
+using Mantid::HistogramData::FrequencyStandardDeviations;
+using Mantid::HistogramData::FrequencyVariances;
+
+namespace {
+Histogram rebinCounts(const Histogram &input, const BinEdges &binEdges) {
+  if (input.yMode() != Histogram::YMode::Counts ||
+      input.xMode() != Histogram::XMode::BinEdges) {
+    throw std::runtime_error(
+        "Histogram  XMode should be BinEdges and YMode should be Counts.");
+  }
```
SimonHeybrock: I don't think this needs to be checked here again, since this function is only visible within this file...

Framework/HistogramData/inc/MantidHistogramData/HistogramRebin.h
```diff
@@ -0,0 +1,39 @@
+#ifndef MANTID_HISTOGRAMDATA_HISTOGRAMREBIN_H_
+#define MANTID_HISTOGRAMDATA_HISTOGRAMREBIN_H_
+
+#include "MantidHistogramData/DllConfig.h"
+namespace Mantid {
+namespace HistogramData {
+class Histogram;
+class BinEdges;
+
+/** HistogramRebin : TODO: DESCRIPTION
```
SimonHeybrock: Remove this line and add doxygen in the cpp file as usual (since there is no class here...)?

Framework/HistogramData/src/HistogramRebin.cpp
```diff
@@ -0,0 +1,157 @@
...
+  return Histogram(binEdges, newCounts,
+                   CountStandardDeviations(std::move(newCountVariances)));
+}
+
+Histogram rebinFrequencies(const Histogram &input, const BinEdges &binEdges) {
+  if (input.yMode() != Histogram::YMode::Frequencies ||
+      input.xMode() != Histogram::XMode::BinEdges) {
+    throw std::runtime_error(
+        "Histogram  XMode should be BinEdges and YMode should be Frequencies.");
+  }
```
SimonHeybrock: Remove, as above?

Framework/HistogramData/src/HistogramRebin.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+  auto &xnew = binEdges.rawData();
+  Counts newCounts(xnew.size() - 1);
+  CountVariances newCountVariances(xnew.size() - 1);
+  auto &ynew = newCounts.mutableData();
+  auto &enew = newCountVariances.mutableData();
+
+  auto size_yold = yold.size();
+  auto size_ynew = ynew.size();
+  size_t iold = 0, inew = 0;
```
SimonHeybrock: Please declare variables on separate lines :)

Framework/HistogramData/src/HistogramRebin.cpp
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  return Histogram(binEdges, newFrequencies, newFrequencyStdDev);
+}
+} // anonymous namespace
+
+namespace Mantid {
+namespace HistogramData {
+
+Histogram rebin(const Histogram &input, const BinEdges &binEdges) {
```
SimonHeybrock: Doxygen.

Framework/HistogramData/test/HistogramRebinTest.h
```diff
@@ -0,0 +1,439 @@
...
+
+  void testExecRebinFrequency() {
+    TS_ASSERT_THROWS_NOTHING(
+        rebin(getFrequencyHistogram(), BinEdges(10, LinearGenerator(0, 0.5))));
+  }
+
+  void testRebinNoYModeDefined() {
+    BinEdges edges(5, LinearGenerator(0, 2));
+    // X Mode Counts
```
SimonHeybrock: Points.

Framework/HistogramData/test/HistogramRebinTest.h
```diff
@@ -0,0 +1,439 @@
...
+    TS_ASSERT_THROWS(
+        rebin(Histogram(BinEdges(10, LinearGenerator(0, 0.5))), edges),
+        std::runtime_error);
+  }
+
+  void testRebinFailsBinEdgesInvalid() {
+    std::vector<double> binEdges{1, 2, 3, 3, 5, 7};
+    BinEdges edges(binEdges);
+
+    TS_ASSERT_THROWS(rebin(getCountsHistogram(), edges), std::runtime_error);
```
SimonHeybrock: Maybe just add a few more cases, e.g., with the bad values at start and end?

We should also test with the new bad bin edges coinciding with the old once *and* not coinciding. Remember that this was an issue with the old implementation?

Framework/HistogramData/test/HistogramRebinTest.h
```diff
@@ -0,0 +1,439 @@
...
+    TS_ASSERT_EQUALS(outFreq.y().rawData(), histFreq.y().rawData());
+    TS_ASSERT_EQUALS(outFreq.e().rawData(), histFreq.e().rawData());
+  }
+
+  void testBinEdgesOutsideInputBins() {
+    auto histCounts = getCountsHistogram();
+    auto histFreq = getFrequencyHistogram();
+
+    auto outCounts = rebin(histCounts, BinEdges(10, LinearGenerator(30, 1)));
+    auto outFreq = rebin(histFreq, BinEdges(5, LinearGenerator(100, 2)));
```
SimonHeybrock: Why different values?

Framework/HistogramData/src/HistogramRebin.cpp
```diff
@@ -0,0 +1,157 @@
+#include "MantidHistogramData/HistogramRebin.h"
```
SimonHeybrock: I would rename this (and the other related files) to `Rebin.h`. Since it is in the `HistogramData` module it will be obvious, people can just use:

```cpp
#include "MantidHistogramData/Rebin.h"
```

Framework/Kernel/test/RebinHistogramTest.h
```diff
@@ -1,9 +1,11 @@
 #ifndef REBINHISTOGRAM_TEST_H_
 #define REBINHISTOGRAM_TEST_H_
```
SimonHeybrock: There is RebinHistogram and HistogramRebin??

Framework/Kernel/test/VectorHelperTest.h
```diff
@@ -383,4 +385,107 @@ class VectorHelperTest : public CxxTest::TestSuite {
...
+
+    for (size_t i = 0; i < counts.size(); i++)
+      frequencies[i] = counts[i] / (binEdges[i + 1] - binEdges[i]);
+
+    std::transform(counts.cbegin(), counts.cend(), errors.begin(),
+                   [](const double count) { return sqrt(count); });
+
+    std::transform(frequencies.cbegin(), frequencies.cend(),
+                   frequencyErrors.begin(),
+                   [](const double freq) { return sqrt(freq); });
```
SimonHeybrock: Why all this init effort for performance tests? Shouldn't all zeros be just as good (apart from the bin edges)?

SimonHeybrock (commented):
@LamarMoore Looks mostly ok, I did not review the math though, will do that another time.

Framework/Kernel/test/RebinHistogramTest.h
LamarMoore: This is the unit test for the VectorHelper::RebinHistogram function.

LamarMoore:
@SimonHeybrock, after renaming HistogramRebin to Rebin, I am encountering a conflict on the build server with Algorithms/Rebin

Framework/HistogramData/src/Rebin.cpp
```diff
@@ -0,0 +1,154 @@
...
+} // anonymous namespace
+
+namespace Mantid {
+namespace HistogramData {
+
+/** Rebins data according to a new set of bin edges.
+* @param input :: input histogram data to be rebinned.
+* @param binEdges :: input will be rebinned according to this set of bin edges.
+* @returns The rebinned histogram.
+* @throws std::runtime_error if the input histogram xmode is not BinEdges and
```
SimonHeybrock: `or` not `and`

Framework/HistogramData/src/Rebin.cpp
```diff
@@ -0,0 +1,154 @@
...
+* the input yMode is undefined.
```
SimonHeybrock: or if the bin edges in either input or output are yield non-positive bin widths.

Framework/HistogramData/src/Rebin.cpp
```diff
@@ -0,0 +1,154 @@
...
+    else if (xo_high <= xn_low)
+      iold++; /* old and new bins do not overlap */
+    else {
+      // delta is the overlap of the bins on the x axis
+      auto delta = xo_high < xn_high ? xo_high : xn_high;
+      delta -= xo_low > xn_low ? xo_low : xn_low;
+
+      if (delta <= 0.0) {
+        throw std::runtime_error("Negative or zero overlaps not allowed.");
+      }
```
SimonHeybrock: Is this possible given your check of the widths above?

Framework/HistogramData/src/Rebin.cpp
```diff
@@ -0,0 +1,154 @@
...
+    else if (xo_high <= xn_low)
+      iold++; /* old and new bins do not overlap */
+    else {
+      //        delta is the overlap of the bins on the x axis
+      auto delta = xo_high < xn_high ? xo_high : xn_high;
+      delta -= xo_low > xn_low ? xo_low : xn_low;
+
+      if (delta <= 0.0) {
+        throw std::runtime_error("Negative or zero bin overlaps not allowed.");
+      }
```
SimonHeybrock: Needed?

Framework/HistogramData/test/RebinTest.h
```diff
@@ -0,0 +1,446 @@
...
+        rebin(getFrequencyHistogram(), BinEdges(10, LinearGenerator(0, 0.5))));
+  }
+
+  void testRebinNoYModeDefined() {
+    BinEdges edges(5, LinearGenerator(0, 2));
+    // X Mode Points
+    TS_ASSERT_THROWS(
+        rebin(Histogram(Histogram::XMode::Points, Histogram::YMode::Counts),
+              edges),
+        std::runtime_error);
```
SimonHeybrock: Strictly speaking you don't know if this throws because of the XMode (without looking into the code). Could be anything, e.g., the `nullptr` you have. Can you pass a valid `Histogram` instead?

Framework/HistogramData/test/RebinTest.h
```diff
@@ -0,0 +1,446 @@
...
+    // No YMode set
+    TS_ASSERT_THROWS(
+        rebin(Histogram(BinEdges(10, LinearGenerator(0, 0.5))), edges),
+        std::runtime_error);
+  }
+
+  void testRebinFailsCentralBinEdgesInvalid() {
+    std::vector<double> binEdges{1, 2, 3, 3, 5, 7};
+    BinEdges edges(binEdges);
```
SimonHeybrock: `BinEdges` has a constructor from `initializer_list`, no need to have the vector first (same comment for many tests below).

Framework/HistogramData/test/RebinTest.h
```diff
@@ -0,0 +1,446 @@
...
+    auto outCounts = rebin(histCounts, histCounts.binEdges());
+    auto outFreq = rebin(histFreq, histFreq.binEdges());
+
+    TS_ASSERT_EQUALS(outCounts.x().rawData(), histCounts.x().rawData());
+    TS_ASSERT_EQUALS(outCounts.y().rawData(), histCounts.y().rawData());
+    TS_ASSERT_EQUALS(outCounts.e().rawData(), histCounts.e().rawData());
+
+    TS_ASSERT_EQUALS(outFreq.x().rawData(), histFreq.x().rawData());
+    TS_ASSERT_EQUALS(outFreq.y().rawData(), histFreq.y().rawData());
+    TS_ASSERT_EQUALS(outFreq.e().rawData(), histFreq.e().rawData());
```
SimonHeybrock: Didn't you once write that test helper that gives `operator==` so you don't have to write `rawData` every time?

Framework/HistogramData/test/RebinTest.h
```diff
@@ -0,0 +1,446 @@
...
+
+  void testBinEdgesOutsideInputBins() {
+    auto histCounts = getCountsHistogram();
+    auto histFreq = getFrequencyHistogram();
+
+    auto outCounts = rebin(histCounts, BinEdges(10, LinearGenerator(30, 1)));
+    auto outFreq = rebin(histFreq, BinEdges(10, LinearGenerator(30, 2)));
+
+    TS_ASSERT(std::all_of(outCounts.y().cbegin(), outCounts.y().cend(),
+                          [](const double i) { return i == 0; }));
```
SimonHeybrock: `e()`?

Framework/HistogramData/test/RebinTest.h
```diff
@@ -0,0 +1,446 @@
...
+    Histogram hist(BinEdges{0, 1, 2}, Counts{10, 10});
+    Histogram histFreq(BinEdges{0, 1, 2}, Frequencies{12, 12});
+    BinEdges edges{0, 0.5, 1, 1.5, 2};
+
+    auto outCounts = rebin(hist, edges);
+    auto outFreq = rebin(histFreq, edges);
+
+    for (size_t i = 0; i < outCounts.y().size(); i++) {
+      TS_ASSERT_EQUALS(outCounts.y()[i], 5.0);
+      TS_ASSERT_DELTA(outCounts.e()[i], std::sqrt(5), 1e-5);
```
SimonHeybrock: Here and below: Why `1e-5`? Have you tried `1e-14` or something similar?

Framework/Kernel/test/RebinHistogramTest.h
```diff
@@ -52,4 +54,77 @@ class RebinHistogramTest : public CxxTest::TestSuite {
...
+    counts.resize(binSize - 1);
+    errors.resize(binSize - 1);
+
+    std::iota(binEdges.begin(), binEdges.end(), 0);
+    std::generate(counts.begin(), counts.end(),
+                  []() { return static_cast<double>(rand() % 1000); });
+
+    std::transform(counts.cbegin(), counts.cend(), errors.begin(),
+                   [](const double count) { return sqrt(count); });
+  }
```
SimonHeybrock: Why all the effort to make this tests data? Should not matter for a performance test.

SimonHeybrock (commented):
Hi Lamar! I looked at this in detail once more. Some small comments:

Framework/HistogramData/test/RebinTest.h
LamarMoore: Doesn't work if the bins are invalid. In order to use edges with invalid bins I have to use a vector.

Framework/HistogramData/test/RebinTest.h
LamarMoore: It doesn't make a difference. I can change it to a lower tolerance.

SimonHeybrock: It makes a big difference. Someone might break `rebin` without breaking tests if the tolerances are so large. So always make bounds as tight as feasible.

Framework/HistogramData/test/RebinTest.h
```diff
@@ -123,11 +126,11 @@ class RebinTest : public CxxTest::TestSuite {
 
     for (size_t i = 0; i < outCounts.y().size(); i++) {
       TS_ASSERT_EQUALS(outCounts.y()[i], 5.0);
-      TS_ASSERT_DELTA(outCounts.e()[i], std::sqrt(5), 1e-5);
+      TS_ASSERT_DELTA(outCounts.e()[i], std::sqrt(5), 1e-14);
       TS_ASSERT_EQUALS(outFreq.y()[i], 12.0);
       TS_ASSERT_DELTA(outFreq.e()[i],
                       std::sqrt(outFreq.y()[i] / (edges[i + 1] - edges[i])),
-                      1e-5);
+                      1e-14);
```
SimonHeybrock: `1e-5` is also used many more times below.

SimonHeybrock:
Looks good now, thanks for the work!
:shipit: --- @second-reviewer: This is a core addition and will be used to replace `VectorHelper::rebin`, so please have a close look at this before merging.

Framework/HistogramData/src/Rebin.cpp
```diff
@@ -0,0 +1,147 @@
...
+      throw std::runtime_error("Negative or zero bin widths not allowed.");
+
+    if (xn_high <= xo_low)
+      inew++; /* old and new bins do not overlap */
+    else if (xo_high <= xn_low)
+      iold++; /* old and new bins do not overlap */
+    else {
+      // delta is the overlap of the bins on the x axis
+      auto delta = xo_high < xn_high ? xo_high : xn_high;
+      delta -= xo_low > xn_low ? xo_low : xn_low;
```
martyngigg: The [`VectorHelper`](https://github.com/mantidproject/mantid/blob/master/Framework/Kernel/src/VectorHelper.cpp#L160) code has a check for `delta < 0` to indicate no overlap. This seems to be missing here.

Is that the case or have I misunderstood something about checking elsewhere?

Framework/HistogramData/src/Rebin.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+      if (xn_high > xo_high) {
+        iold++;
+      } else {
+        inew++;
+      }
+    }
+  }
+
+  return Histogram(binEdges, newCounts,
```
martyngigg: I have compared this with the old `VectorHelper::rebin` routine and I'm confident that they are doing the same thing mathematically with the addition of better checking for zero widths etc.

martyngigg (approved):
This looks like a great improvement to the rebinning routines. It's nice to finally extract them from `VectorHelper` and have them defined along with domain objects rather than mixed in with other operations that happen to operate on the same underlying types.

PR #17259, opened 2016-08-17
Multi-dataset interface: fix a bug in plotting the function guess.

### LINKED ISSUE #17256: Multi-dataset fit interface gives error with some uninitialised functions.
### Expected behavior

No errors.

### Actual behavior

If a function is added that cannot be used with the default parameters/attributes the interface gives an unexpected error message and may crash Mantid.

1. Open the Multi-dataset interface.
2. Add Resolution function (category General).

Unexpected exception message.

### Platforms affected

All.

mantid-roman:
Catch any std exception from the guess function call and log the message.

**To test:**

- Add a Resolution function to Multi-dataset interface.
- No "Unexpected exception" dialog should show.
- A warning about using uninitialised TabulatedFunction should be seen in the log window.
- Set some data to the function by setting either Filename or Workspace attribute.
- The warning shouldn't show any more.

Fixes #17256.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidQt/CustomInterfaces/src/MultiDatasetFit/MultiDatasetFit.cpp
+++ b/MantidQt/CustomInterfaces/src/MultiDatasetFit/MultiDatasetFit.cpp
@@ -717,7 +717,7 @@ void MultiDatasetFit::updateGuessFunction(const QString &, const QString &) {
 /// Log a warning
 /// @param msg :: A warning message to log.
 void MultiDatasetFit::logWarning(const std::string &msg) {
-  g_log.warning() << msg << std::endl;
+  g_log.warning(msg);
 }
 
 } // CustomInterfaces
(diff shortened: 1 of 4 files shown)


AntonPiccardoSelg:
* Code looks good.
* Following the described steps above does not crash Mantid but rather produces a log warning as mentioned in the description
* Once a workspace is set for the Resolution function, the warning disappears (again  as described)

Happy to :shipit:

MantidQt/CustomInterfaces/src/MultiDatasetFit/MultiDatasetFit.cpp
```diff
@@ -713,5 +714,11 @@ void MultiDatasetFit::updateGuessFunction(const QString &, const QString &) {
   m_plotController->updateGuessFunction(*m_functionBrowser->getFunction());
 }
 
+/// Log a warning
+/// @param msg :: A warning message to log.
+void MultiDatasetFit::logWarning(const std::string &msg) {
+  g_log.warning() << msg << std::endl;
```
FedeMPouzols: Not that it makes any real difference, bit forma consistency `\n` would be preferred. Also, the message can be passed as a parameter rather than using the stream operators.

PR #17264, opened 2016-08-17
Fitting histogram data

### LINKED ISSUE #16796: Fitting histogram data, not points
From Mantid forum, http://forum.mantidproject.org/t/fitting-histogram-data-not-points/130

*For a histogram data set, at present, fitting evaluates the function at the centre of each bin. If the bins are large and the function has any significant curvature, the point value can differ from the average over the bin, which is what the data probably represents. If there are fast "wiggles" in the fit function these can really confuse the fit if the maximum or minimum happens to line up with the bin centre and has the right sign to appear to improve the fit.*

*Often it's easy to analytically integrate a fit function over a range of X values, so the correct average and its derivatives could be calculated with little extra effort.*

*Could a fit function declare that it knows about this, and is then passed an X array of bin edges to work with, and a flag to say so?*

*Alternatively a function could assume it is always working with a histogram and reconstruct the bin edges for itself (as ConvertToHistogram does) and then do the averaging. That assumes equal sized bins which is usually the case.*

*The attached script demonstrates the problem. It fits a Lorentzian to a generated Lorentzian plus a small amount of noise, with various degrees of binning. I've defined the bin-size-aware fit function for comparison with the built in one.*

[HistoAwareFits.py.txt](https://github.com/mantidproject/mantid/files/347712/HistoAwareFits.py.txt)

*The point-data function slightly over-estimates the amplitude if the bins are too big, significantly over-estimates the FWHM, and there can be scatter in the peak centre value depending on how the peak lines up with the bin boundaries.*

---

@mantid-roman Do you know if this would be possible to implement?

mantid-roman:
Made necessary changes to fit functions, function domains and Fit algorithm to be able to fit 1D functions to histograms in a "histogram aware" way. It means that in the least squares the calculated values are obtained by integrated the function within each bin. This way produces more accurate results for large bins than when using values at the bin centres.

**To test:**
1. Simple test
   - Run this script
   
   ```
   x = [-1, -0.8,-0.6,-0.4,-0.2,0,0.2,0.4,0.6,0.8,1]
   y = [0.0302241,0.04333437713434279,0.06408122585358178,0.09365777093187931,0.1211189415908434,0.1211189415908435,0.0936577709318793,0.06408122585358172,0.04333437713434274,0.03022406683891948]
   e = [0.01] * len(y)
   ws = CreateWorkspace(x, y, e)
   fun = 'name=Lorentzian,FWHM=0.9'
   Fit(fun, ws, Normalise=True, Output='normal')
   Fit(fun, ws, EvaluationType="Histogram", Output='histo')
   ```
   - Check that parameters of the "histo" fit are closer to the correct values which are `Amplitude=1` and `FWHM=1`.
2. More complex test.
   - Run the script:
   
   ```
   import math
   import numpy
   
   # time range for fit
   xdata=WorkspaceFactory.create("Workspace2D",XLength=401,YLength=400,NVectors=1)
   xdata.dataX(0)[:]=numpy.linspace(-10.0,10.0,401)
   xdata.dataY(0)[:]=numpy.random.normal(loc=0.0,scale=0.01,size=400)
   xdata.dataE(0)[:]=0.1
   AnalysisDataService.addOrReplace("xdata",xdata)
   
   BuiltIn="name=Lorentzian"
   
   # evaluate the function and add noise
   ytheory=EvaluateFunction(Function=BuiltIn+",Amplitude=20.0,PeakCentre=0.0,FWHM=1.0",InputWorkspace=xdata,OutputWorkspace="ytheory",EvaluationType='Histogram')
   yexper=Plus(LHSWorkspace=xdata,RHSWorkspace=ytheory,OutputWorkspace="DynamicRange")
   Guesses=(20,0.1,1.1)
   # adjust the X axis and repeat the fits. Print the parameters adjusted back to a line centred at 0.
   Results=WorkspaceFactory.createTable()
   Results.addColumn("int","Bunching",plottype=1)
   Results.addColumn("float","stdAmp",plottype=2)
   Results.addColumn("float","stdAmpErr",plottype=5)
   Results.addColumn("float","stdCen",plottype=2)
   Results.addColumn("float","stdCenErr",plottype=5)
   Results.addColumn("float","stdFWHM",plottype=2)
   Results.addColumn("float","stdFWHMErr",plottype=5)
   Results.addColumn("float","stdCost",plottype=2)
   Results.addColumn("float","awareAmp",plottype=2)
   Results.addColumn("float","awareAmpErr",plottype=5)
   Results.addColumn("float","awareCen",plottype=2)
   Results.addColumn("float","awareCenErr",plottype=5)
   Results.addColumn("float","awareFWHM",plottype=2)
   Results.addColumn("float","awareFWHMErr",plottype=5)
   Results.addColumn("float","awareCost",plottype=2)
   for bun in range(1,31):
   row=[0.0]*15
   row[0]=bun
   ws=Rebunch(InputWorkspace=yexper,NBunch=bun)
   for i,histo in enumerate(('CentrePoint','Histogram')):
       kw = {'Normalise': True} if i == 0 else {}
       ffr=Fit(Function="name=Lorentzian,Amplitude="+str(Guesses[0])+",PeakCentre="+str(Guesses[1])+",FWHM="+str(Guesses[2]),
               InputWorkspace=ws,WorkspaceIndex=1,EvaluationType=histo,Output="HistoFit", **kw)
       params=ffr[3].column(1)
       errors=ffr[3].column(2)
       row[i*7+1:i*7+8]=[params[0],errors[0],params[1],errors[1],params[2],errors[2],params[3]]
       print bun,params[0],errors[0],params[1],errors[1],params[2],errors[2],params[3]
   Results.addRow(row)
   AnalysisDataService.addOrReplace("HistoAwareResults",Results)
   ```
   - This script generates a Lorentzian peak with some random noise then rebunches it in a loop gradually increasing the bin size. For each bin size the data gets fitted twice: first the usual way, then the "histogram aware" way.
   - The results of the fits are in the "HistoAwareResults" table. Parameters from the normal fits are prefixed with "std" and those obtained the new way are prefixed with "aware".
   - Plot parameter values from this table and check that the "aware" kind has smaller variations.

Fixes #16796.

Release notes added to `docs/source/release/v3.8.0/framework.rst`.

---
#### Reviewer

##### Code Review
##### Functional Tests


docs/source/algorithms/Fit-v1.rst
```diff
@@ -178,6 +178,37 @@ Fitting to data in a MatrixWorkspace
...
+
+If the input MatrixWorkspace contains histogram data then there are two
+way to evalue the fit function on it and it can be set via the `EvaluationType`
+property. If `EvaluationType` is "CentrePoint" (which is the default) the
+function is evaluated at the centre of the bin. If `EvaluationType` is set to
+"Histogram" then the function is integrated within each bin and these integrals
+are used to evaluate the cost function.
+
+The "Histogram" evaluation type can give more accurate results when fitting
+histograms with very large bins. It also less sensitive to a particular binning.
```
SimonHeybrock: Nice figure! 👍 Do you maybe want to point out in more detail that in particular the estimate for FWHM has improved a lot?

(I am mostly traveling this week and have some other appointments, if no one picks this up I could probably review next week).

docs/source/algorithms/Fit-v1.rst
```diff
@@ -178,6 +178,37 @@ Fitting to data in a MatrixWorkspace
 The error values in the input workspace are used to weight the data in
 the fit. Zero error values are not allowed and are replaced with ones.
 
+Fitting to histogram data
+#########################
+
+If the input MatrixWorkspace contains histogram data then there are two
+way to evalue the fit function on it and it can be set via the `EvaluationType`
+property. If `EvaluationType` is "CentrePoint" (which is the default) the
```
SimonHeybrock: Looking at the result, I would make `Histogram` the default. Did you decide to keep the current behavior since it might break some things?

Framework/API/src/FunctionDomain1D.cpp
```diff
@@ -127,5 +127,37 @@ FunctionDomain1DSpectrum::FunctionDomain1DSpectrum(
     std::vector<double>::const_iterator to)
     : FunctionDomain1DVector(from, to), m_workspaceIndex(wi) {}
 
+/// Constructor.
+/// @param bins :: A vector with bin boundaries.
+FunctionDomain1DHistogram::FunctionDomain1DHistogram(
+    const std::vector<double> &bins)
```
SimonHeybrock: This looks identical to the other constructor, I think you can just delegate to the other constructor:

```cpp
  : FunctionDomain1DHistogram(bins.begin(), bins.end())) {}
```

Framework/API/inc/MantidAPI/IFunction1D.h
```diff
@@ -80,6 +80,14 @@ class MANTID_API_DLL IFunction1D : public virtual IFunction {
                                const size_t nData);
 
 protected:
+  /// Calculate histogram data for the given bin boundaries.
+  virtual void histogram1D(double *out, double left, const double *right,
+                           const size_t nBins) const;
+  /// Devivatives of the histogram.
```
SimonHeybrock: It took me a while to understand that this is the derivative of the integral by the different parameters of the fit function. Maybe explain this more clearly here in the doxygen?

Framework/CurveFitting/src/IMWDomainCreator.cpp
```diff
@@ -0,0 +1,494 @@
...
+ * @param workspacePropertyName :: Name of the workspace property.
+ * @param domainType :: Type of the domain: Simple, Sequential, or Parallel.
+ */
+IMWDomainCreator::IMWDomainCreator(Kernel::IPropertyManager *fit,
+                                   const std::string &workspacePropertyName,
+                                   IMWDomainCreator::DomainType domainType)
+    : API::IDomainCreator(
+          fit, std::vector<std::string>(1, workspacePropertyName), domainType),
+      m_workspaceIndex(-1), m_startX(EMPTY_DBL()), m_endX(EMPTY_DBL()),
+      /*m_maxSize(0), m_normalise(false),*/ m_startIndex(0) {
```
SimonHeybrock: Remove commented member inits?

Framework/CurveFitting/src/IMWDomainCreator.cpp
```diff
@@ -0,0 +1,494 @@
...
+  ws->setTitle("");
+  ws->setYUnitLabel(m_matrixWorkspace->YUnitLabel());
+  ws->setYUnit(m_matrixWorkspace->YUnit());
+  ws->getAxis(0)->unit() = m_matrixWorkspace->getAxis(0)->unit();
+  auto tAxis = new API::TextAxis(nhistograms);
+  ws->replaceAxis(1, tAxis);
+
+  const MantidVec &inputX = m_matrixWorkspace->readX(m_workspaceIndex);
+  const MantidVec &inputY = m_matrixWorkspace->readY(m_workspaceIndex);
+  const MantidVec &inputE = m_matrixWorkspace->readE(m_workspaceIndex);
```
SimonHeybrock: You can use the new `HistogramData` interface, e.g., `const auto &inputX = m_matrixWorkspace->x(index)`. Similar for Y and E. If you need a non-const reference (as below) use `mutableX(index)` etc.

Framework/CurveFitting/test/Algorithms/PlotPeakByLogValueTest.h
```diff
@@ -552,18 +552,88 @@ class PlotPeakByLogValueTest : public CxxTest::TestSuite {
...
+
+  void test_histogram_fit() {
+    size_t nbins = 10;
+    auto ws =
+        WorkspaceFactory::Instance().create("Workspace2D", 3, nbins + 1, nbins);
+    double x0 = -10.0;
+    double x1 = 10.0;
+    double dx = (x1 - x0) / static_cast<double>(nbins);
+    for (size_t i = 0; i < nbins; ++i) {
+      auto x = x0 + double(i) * dx;
```
SimonHeybrock: There is a new and slightly nicer way to create such a binning:

```cpp
ws->setBinEdges(0, nbins + 1, HistogramData::LinearGenerator(x0, x1));
ws->setSharedX(1, ws->sharedX(0));
ws->setSharedX(2, ws->sharedX(0));
```

Framework/CurveFitting/test/HistogramDomainCreatorTest.h
```diff
@@ -0,0 +1,517 @@
...
+    FunctionValues_sptr values;
+    TS_ASSERT_THROWS_NOTHING(creator.createDomain(domain, values));
+    TS_ASSERT_EQUALS(domain->size(), 10);
+    TS_ASSERT_EQUALS(values->size(), 10);
+    TS_ASSERT_EQUALS(ws->blocksize(), 10);
+    auto &h = dynamic_cast<FunctionDomain1DHistogram &>(*domain);
+    auto &x = ws->readX(0);
+    for (size_t j = 0; j < 10; ++j) {
+      TS_ASSERT_EQUALS(h[j], x[j + 1]);
+    }
```
SimonHeybrock: Is it possible to test also `x[0]`?

Framework/CurveFitting/src/Algorithms/PlotPeakByLogValue.cpp
```diff
@@ -268,10 +276,16 @@ void PlotPeakByLogValue::exec() {
...
+        bool histogramFit = getPropertyValue("EvaluationType") == "Histogram";
+
         // Fit the function
         API::IAlgorithm_sptr fit =
             AlgorithmManager::Instance().createUnmanaged("Fit");
         fit->initialize();
+        if (histogramFit) {
+          fit->setPropertyValue("EvaluationType",
+                                getPropertyValue("EvaluationType"));
+        }
```
SimonHeybrock: Why not always forward that property value? Would create fewer problems in case the default is changed?

SimonHeybrock:
This looks all very nice. Tests function as described, and the result are much better with `EvaluationType="Histogram"`. Existing code has been refactored and moved into a new base class (amounting for a significant fraction of the changes in this PR). 👍 

@mantid-roman Out of interest: In the second python example you give the peak height for the histogram-based fit are systematically slightly smaller than the true value (i.e., all values are 19.x, never 20.x). Is is to be expected from the method?

As commented above, I would recommend making `"Histogram"` the default mode. To me it looks like the old method is basically wrong, so we have a strong argument for justifying a behavior change. Maybe we can get a second opinion, if you don't feel comfortable making this decision?

SimonHeybrock:
Another thought regarding the default mode: Have you tried setting the default to `"Histogram"` and run the system tests?

docs/source/algorithms/Fit-v1.rst
mantid-roman: Histogram fit only works for the Lorentzian and Gaussian, other peaks would require some sort of numerical integration. This could be an overkill for histograms with small bins and in cases when a high accuracy isn't required.

mantid-roman:
@SimonHeybrock The shift in the amplitude in the second example is due to the noise. If you run it several times you get different shifts (if the noise is random).

docs/source/algorithms/Fit-v1.rst
SimonHeybrock: Yes, that is a good point. Would it make sense to have different defaults, depending on the fit function (probably cumbersome to implement though?)?

mantid-roman: That could be very confusing for the users I guess. Anyway, a change of an important default shouldn't be decided without a wider discussion.

SimonHeybrock:
@mantid-roman Latest changes look good. As discussed above the default fit mode cannot be changed easily, do you instead want to make it more clear in the documentation and UI tooltips that `Histogram` is the recommended type, if available?

Apart from that I think this is ready once Jenkins gives his ok.

mantid-roman:
The tooltips is a good point but I am not sure about making recommendations to scientists who are supposed to know what they are doing (they are scientists after all).

SimonHeybrock:
@mantid-roman I guess we just have to find an adequate formulation such as "Histogram mode typically yields significantly better results for narrow peaks"?

mantid-roman:
Fit's description has this line: "The "Histogram" evaluation type can give more accurate results when fitting
histograms with very large bins. It also less sensitive to a particular binning."

SimonHeybrock:
If you replace "can" by "will typically" I would be happy :) (if you think that statement is correct)

mantid-roman:
If you think there is something incorrect in the description let's discuss it. I am OK with changing "can" to "will typically" although "typically" implies that we have extensive statistics on the performance of this type of fit which we don't. Again we have only two function types implemented so for a user who uses IkedaCarpenter or Convolution most of the time a histogram fit will typically result in a failure.

SimonHeybrock:
With the latest changes the `Evaluate Function As` option is not displayed if the input workspace contains point data. Nice :+1:.

I'd say :shipit:!

@mantid-roman I was wondering why some of the fit settings (including the new one) are only displayed in the "Fit function" widget, but not in the dialog of the "Fit" algorithm. Am I doing something wrong?

mantid-roman:
@SimonHeybrock The work on the dialog has never been finished because of the complexity of the input but also because users (apparently) never use it and never requested it to be fixed. But it is there because the framework requires its existence.

PR #17265, opened 2016-08-17
HistogramData Rollout: XDataConverter

### LINKED ISSUE #16878: HistogramData Rollout: FrameWork/Algorithms/XDataConverter
This issue relates to the broader HistogramData rollout see [here](http://docs.mantidproject.org/nightly/concepts/HistogramData.html#rollout-status) and [here](https://github.com/mantidproject/mantid/issues/16872). We are currently refactoring code in order to make use of the new interface by replacing references to the old interface and making a few improvements where possible. In this issue you will be modifying XDataConverter.cpp.

### Examples

Examples of changes include:
 - `MantidVec &X = inputWS->dataX(0);` becomes `auto &X = inputWS->mutableX(0);`
 - `outputWS->dataX(i) = inputWS->readX(i);` becomes `outputWS->setSharedX(i, inputWS->sharedX(i));`
 - 
```c 
outputWS->dataY(i) = inputWS->readY(i);
outputWS->dataE(i) = intputWS->readE(i);
```
becomes
```
outputWS->setHistogram(i, inputWS->histogram(i));
```
### Tasks
There are a few tasks which should be completed when working on these files:
1.Add performance tests where they may be missing. Read [this](http://www.mantidproject.org/Writing_Performance_Tests) document carefully for information on how to create and run performance tests.
2. Run the SystemTest/PerformanceTest for the algorithm before making any changes and make a note of the execution time. How to run unit tests can be found [here](http://www.mantidproject.org/Running_the_unit_tests).
3. Modify XDataConverter.cpp as well as its unit test (XDataConverter.cpp/XDataConverterTest.cpp) to conform to the new Histogram standard.
3. After making changes run the SystemTest/Performance test and make sure the execution time has not increased significantly.

PranavBahuguna:
As part of the HistogramData Rollout, the existing code for `XDataConverter` and its unit tests has been refactored to make use of the new methods in HistogramData. In addition, I have also written performance tests to verify that this refactor has not resulted in an increase in execution time.

Fixes #16878 .

**To test:**

Run performance tests for `ConvertToHistogramTest` and `ConvertToPointDataTest` - the execution time for both tests should be reduced from around 1.1s to 0.3s.

#### Reviewer

##### Code Review
##### Functional Tests


Framework/Algorithms/src/XDataConverter.cpp
```diff
@@ -103,7 +103,7 @@ void XDataConverter::setXData(API::MatrixWorkspace_sptr outputWS,
         }
       }
     }
-    outputWS->setX(index, m_cachedX);
+    outputWS->setSharedX(index, m_cachedX);
   } else {
     const MantidVec &xBoundaries = inputWS->readX(index);
     MantidVec &xPoints = outputWS->dataX(index);
```
SimonHeybrock: There are a few uses of the old `readX` and `dataX` that need to be replaced, here and above.

SimonHeybrock:
@PranavBahuguna Greate performance improvement! Could you add this to the release notes, there is a section that contains a short list of algorithms that have seen performance improvement due to `HistogramData` changes. Just a a short note there, with a rough estimate of the speedup (i.e. 3x to 4x speedup).

Framework/Algorithms/src/ConvertToHistogram.cpp
```diff
@@ -42,7 +42,7 @@ bool ConvertToHistogram::isWorkspaceLogical(
     const MatrixWorkspace_sptr inputWS) const {
```
SimonHeybrock: I think you can actually remove this method from all involved classes. The check in this method is not necessary anymore, since `Histogram` guarantees that this is always true.

Framework/Algorithms/src/XDataConverter.cpp
PranavBahuguna: I have actually asked about those before. The problem is that some of the functions involved (such as `Kernel::VectorHelper::convertToBinCentre` called from `ConvertToPointData::calculateXPoints` requires parameters in MantidVec form. I was not able to use the new functions since they were not being given in that form. Changing `Kernel::VectorHelper::convertToBinCentre` parameters is impractical as it is used by many other algorithms.

SimonHeybrock: You can change `ConvertToPointData::calculateXPoints` to take `HistogramX` as arguments. Then either use the `rawData()` accessor to get a vector, or change `convertToBinCentre` (there are just a few algorithms that use it).

SimonHeybrock: Also: The function body of `ConvertToHistogram::calculateXPoints` can be replaced by a call to `VectorHelper::convertToBinBoundary`.

SimonHeybrock: @PranavBahuguna Actually, you can do even better:

Change the signature of the conversion function to:

```cpp
Kernel::cow_ptr<HistogramX> XDataConverter::calculateXPoints(Kernel::cow_ptr<HistogramX> inputX) const;
```

You can then directly use the constructors of `Points` and `BinEdges` to to the work, instead of using the functions in `VectorHelper` (the latter will be removed soon), for example:

```cpp
Kernel::cow_ptr<HistogramX> ConvertToPointData::calculateXPoints(Kernel::cow_ptr<HistogramX> inputX) const {
  return Points(BinEdges(std::move(inputX))).cowData();
}
```

At the call site use something like

```cpp
outputWS->setSharedX(i, calculateXPoints(inputWS->sharedX()));
```

I think this should not change the performance results, since there will be no additional copy.

If you have trouble understanding this, don't hesitate to contact me on Slack, or get in touch with @LamarMoore, if you want to discuss this with someone fact to face.

SimonHeybrock:
Changes look good now, code is shorter than before, performance test added, improved sharing drastically increases performance and reduces memory consumption. :shipit:.

(Note: We may want to clean up the release notes at some point, we will probably be adding more similar items about algorithms that have improved performance. Add new section?)

PR #17266, opened 2016-08-17
ConvertUnits now works with Point Workspaces

### LINKED ISSUE #13280: CovertUnit only accepts Histogram data, but should probably also work for Point data
A point data set cannot be used with the ConverUnits algorihtm, the validator blocks the user to do so. The point data set needs to be converted into a histo data set before using the ConvertUnits algorithm

dtasev:
## Description of work
- `ConvertUnits` now accepts workspaces with Point data.
- `ConvertUnits` has a new property that toggles that functionality on and off
- 2 new Unit Tests have been added to test the new functionality with Point data
- A new Performance Test has been added to test the new functionality with Point data
- The InputWorkspace is converted to Bins, and, at the end, back to Points
## To test:

I used the workspace [here](https://drive.google.com/file/d/0B2Mj8e56Z8xTZjNSNFNScHhBT2c/view?usp=sharing) <- can also be found from the TrainingCourseData, with the helper Python Script found [here](http://pastebin.com/mSwMzx2L)

If you want to try without the fix:
- Load the `164199` workspace into Mantid
- Run the helper Python Script
- It should fail on `ConvertUnits(points_ws, Target='Wavelength')` as `ConvertUnits` doesn't accept Point Data

Testing after applying the fix:
- Load the `164199` workspace into Mantid
- Run the helper Python Script
- Everything should be completed successfully
- You can now test the `ConvertUnits` manually from the `Algorithms` window
- The `ConvertUnits` dialogue has a new option: `ConvertFromPointData`, set by default to `True`
- Check the tooltip text of the checkbox for spelling mistakes and clarity
- Try converting a Points workspace with the option unchecked
- This should fail and an exception should be shown in the log, check the message for spelling mistakes and clarity
- Try again with the option checked, it should now run successfully

Additional:
- Run the `ConvertUnits` unit test. There are 2 new tests. All tests must pass
- Run the `ConvertUnits` performance test. There is 1 new test. All tests must pass. (they take a long time, they might need to be reworked, as they currently load files in)
- Check the documentation for spelling mistakes/clarity. The documentation has been changed both for the release notes (link) and for the algorithm itself.
- Build the documentation and check for bad formatting, any spelling mistakes or anything unclear.
- Check the code quality/readability

Fixes #13280.

Added in release notes [here](https://github.com/mantidproject/mantid/pull/17266/files#diff-fef8ac613fa6d94a43015ee70c230735R69)

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Algorithms/src/ConvertUnits.cpp
+++ b/Framework/Algorithms/src/ConvertUnits.cpp
@@ -75,10 +75,11 @@ void ConvertUnits::init() {
                   "recommended (see "
                   "http://www.mantidproject.org/ConvertUnits).");
 
-  declareProperty("ConvertFromPointData", true,
-                  "When checked, if the Input Workspace contains Points\n"
-                  "the algorithm ConvertToHistogram will be run to convert\n"
-                  "the Points to Bins. The Output Workspace will contains Bins."
+  declareProperty(
+      "ConvertFromPointData", true,
+      "When checked, if the Input Workspace contains Points\n"
+      "the algorithm ConvertToHistogram will be run to convert\n"
+      "the Points to Bins. The Output Workspace will contains Bins.")
 }
 
 /** Executes the algorithm
--- a/Framework/Algorithms/src/ConvertUnits.cpp
+++ b/Framework/Algorithms/src/ConvertUnits.cpp
@@ -41,45 +41,45 @@ ConvertUnits::ConvertUnits()
 
 /// Initialisation method
 void ConvertUnits::init() {
-  auto wsValidator = boost::make_shared<CompositeValidator>();
-  wsValidator->add<WorkspaceUnitValidator>();
-  declareProperty(make_unique<WorkspaceProperty<API::MatrixWorkspace>>(
-                      "InputWorkspace", "", Direction::Input, wsValidator),
-                  "Name of the input workspace");
-  declareProperty(make_unique<WorkspaceProperty<API::MatrixWorkspace>>(
-                      "OutputWorkspace", "", Direction::Output),
-                  "Name of the output workspace, can be the same as the input");
+	auto wsValidator = boost::make_shared<CompositeValidator>();
+	wsValidator->add<WorkspaceUnitValidator>();
+	declareProperty(make_unique<WorkspaceProperty<API::MatrixWorkspace>>(
+		"InputWorkspace", "", Direction::Input, wsValidator),
+		"Name of the input workspace");
+	declareProperty(make_unique<WorkspaceProperty<API::MatrixWorkspace>>(
+		"OutputWorkspace", "", Direction::Output),
+		"Name of the output workspace, can be the same as the input");
 
-  // Extract the current contents of the UnitFactory to be the allowed values of
-  // the Target property
-  declareProperty("Target", "", boost::make_shared<StringListValidator>(
-                                    UnitFactory::Instance().getKeys()),
-                  "The name of the units to convert to (must be one of those "
-                  "registered in\n"
-                  "the Unit Factory)");
-  std::vector<std::string> propOptions{"Elastic", "Direct", "Indirect"};
-  declareProperty("EMode", "Elastic",
-                  boost::make_shared<StringListValidator>(propOptions),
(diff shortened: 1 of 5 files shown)


Framework/Algorithms/src/ConvertUnits.cpp
```diff
@@ -80,19 +74,58 @@ void ConvertUnits::init() {
...
 
 /** Executes the algorithm
- *  @throw std::runtime_error If the input workspace has not had its unit set
- *  @throw NotImplementedError If the input workspace contains point (not
- * histogram) data
- *  @throw InstrumentDefinitionError If unable to calculate source-sample
- * distance
- */
+*  @throw std::runtime_error If the input workspace has not had its unit set
+*  @throw NotImplementedError If the input workspace contains point (not
```
DavidFair: Does this still need to throw `NotImplementedError` as we now accept point data

Framework/Algorithms/src/ConvertUnits.cpp
```diff
@@ -80,19 +74,58 @@ void ConvertUnits::init() {
...
+* histogram) data
+*  @throw InstrumentDefinitionError If unable to calculate source-sample
+* distance
+*/
 void ConvertUnits::exec() {
   // Get the workspaces
+  bool acceptPointData = getProperty("ConvertFromPointData");
```
DavidFair: `const`

Framework/Algorithms/src/ConvertUnits.cpp
```diff
@@ -80,19 +74,58 @@ void ConvertUnits::init() {
...
+
   MatrixWorkspace_sptr inputWS = getProperty("InputWorkspace");
```
DavidFair: Could this be `const` as well if the input WS is not changed

Framework/Algorithms/src/ConvertUnits.cpp
```diff
@@ -80,19 +74,58 @@ void ConvertUnits::init() {
...
+      MatrixWorkspace_sptr temp = convToHist->getProperty("OutputWorkspace");
+      correctWS = boost::dynamic_pointer_cast<MatrixWorkspace>(temp);
+
+      if (!correctWS->isHistogramData()) {
+        throw std::runtime_error(
+            "Failed to convert workspace from Points to Bins");
+      }
+    } else {
+      throw std::runtime_error("Workspace contains points, you can either run "
+                               "ConvertToHistogram on it, or check "
```
DavidFair: "or check" implies to me the user needs to actually check something (i.e. verify something). Could we say "or set ConvertFromPointData to enabled"

Framework/Algorithms/src/ConvertUnits.cpp
```diff
@@ -122,15 +155,15 @@ void ConvertUnits::exec() {
     }
   }
 
-  if (inputWS->x(0).size() < 2) {
+  if (correctWS->x(0).size() < 2) {
```
DavidFair: Could we put in a comment why this value is 2?

DimitarTasev:
@DavidFair I'm thinking I should change the `ConvertFromPointData` name as well, as I believe this implies that the workspace will be converted to bins, maybe use something like `HandlePointData`?

Framework/Algorithms/src/ConvertUnits.cpp
DimitarTasev: It fails to compile with `MatrixWorkspace_const_sptr`

Framework/Algorithms/src/ConvertUnits.cpp
DimitarTasev: I think the error message below is enough? It's because when the workspace is holding `BinEdges`, having 1 value doesn't makes sense, because the first would be the start of the bin and the second the end. If there's a single value it would be holding `Points`.

DimitarTasev: Maybe it's not as clear as I thought, I'll put a little comment in basically saying what I said above.

DavidFair:
Tested locally and working as expected. When it finishes the final PR tests I'm happy to :shipit:

PR #17277, opened 2016-08-18
Engg:: Use current instrument prefix if numerical only filename

### LINKED ISSUE #17140: Engg:: If default instrument is not set calibration fails to find files
If the default instrument is not set to EnginX on the start-up screen the interface will fail to select and instrument and subsequently fail to find the necessary files.

@FedeMPouzols We could either select EnginX by default or give an error informing the user they have not selected an instrument. Which behaviour would be best?

DavidFair:
Description of work.

Previously unless the user set their favourite instrument to `ENGIN-X` the load algorithm would fail to find numerical only runs e.g. `236516` as an input. 
This PR checks if the entry only contains numeric digits and appends the selected instrument under Engineering Diffraction as a prefix for the `Load` algorithm.

**To test:**
All test nxs files are located under system test data.

Ensure your favourite instrument is set to something other than `ENGIN-X`.
Open Engineering Diffraction under Interfaces
Enter the following run numbers (or any substitute):
Van: 236516
Calibration: 241391
Select calibrate
If Mantid manages to find all files and run the calibration regardless of the success of that calibration this means this test is successful.

Next select the Focus tab
Enter file: 194547
Start the focus process
If Mantid manages to find the file to focus and runs the algorithm regardless of the success of the focussing means this test is successful .

Next 

Fixes #17140 .

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidQt/CustomInterfaces/src/EnggDiffraction/EnggDiffractionPresenter.cpp
+++ b/MantidQt/CustomInterfaces/src/EnggDiffraction/EnggDiffractionPresenter.cpp
@@ -10,6 +10,7 @@
 #include "MantidQtCustomInterfaces/EnggDiffraction/EnggDiffractionPresenter.h"
 #include "MantidQtCustomInterfaces/EnggDiffraction/IEnggDiffractionView.h"
 
+#include <algorithm>
 #include <fstream>
 
 #include <boost/lexical_cast.hpp>
@@ -36,8 +37,7 @@ const std::string EnggDiffractionPresenter::g_shortMsgRBNumberRequired =
 const std::string EnggDiffractionPresenter::g_msgRBNumberRequired =
     std::string("An experiment reference number (or so called \"RB "
                 "number\" at ISIS) is "
-                "required to effectively use this interface, "
-                "this can be entered at the top of the screen. \n") +
+                "required to effectively use this interface. \n") +
     "The output calibration, focusing and fitting results will be "
     "saved in directories named using the RB number entered.";
 
@@ -598,7 +598,6 @@ void EnggDiffractionPresenter::processRBNumberChange() {
   const std::string rbn = m_view->getRBNumber();
   auto valid = validateRBNumber(rbn);
   m_view->enableTabs(valid);
-  m_view->highlightRbNumber(valid);
   m_view->splashMessage(!valid, g_shortMsgRBNumberRequired,
                         g_msgRBNumberRequired);
   if (!valid) {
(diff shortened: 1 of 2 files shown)


PranavBahuguna:
All things appear to be working as expected. :shipit:

DavidFair:
Before this is merged could @FedeMPouzols check if over and make sure this is the behaviour he described in #17140

MantidQt/CustomInterfaces/src/EnggDiffraction/EnggDiffractionPresenter.cpp
```diff
@@ -1214,6 +1218,51 @@ void EnggDiffractionPresenter::doCalib(const EnggDiffCalibSettings &cs,
...
+    const std::string vanNo, const std::string cerNo, std::string &outVanName,
+    std::string &outCerName) const {
+  // If the file is numerical only we need to append
+  // it in case the favorite instrument isn't set to ENGINX
+  const std::string currentInst = m_view->currentInstrument();
+  // Vanadium file
+  if (std::all_of(vanNo.begin(), vanNo.end(), ::isdigit)) {
+    // This only has digits - append prefix
+    outVanName = currentInst + vanNo;
+  }
```
FedeMPouzols: Looks good. You could have also implemented this 4 parameters overload of `appendCalibInstPrefix` by calling the 2 parameters overload of the same method which would be essentially just the `if(std::all_of(...)) { ... }`

MantidQt/CustomInterfaces/src/EnggDiffraction/EnggDiffractionPresenter.cpp
```diff
@@ -1665,7 +1714,14 @@ void EnggDiffractionPresenter::doFocusing(const EnggDiffCalibSettings &cs,
   MatrixWorkspace_sptr inWS;
 
   const std::string vanNo = m_view->currentVanadiumNo();
-  loadOrCalcVanadiumWorkspaces(vanNo, cs.m_inputDirCalib, vanIntegWS,
+
+  // Append instrument name if numerical only entry
+  std::string vanFileName;
```
FedeMPouzols: Not the point of this branch, but because of the way this interface has evolved we have some confusion between "No" and "Filename". `No` is the number. This vanFilename would be something like ENGINX22345 which is closer to the file name but still missing the 0s and the extension. Not totally sure, but maybe we could introduce here another name, "fileHint"? The smart file finder behind `Load`  will turn that into a file name as it is found on disk (and with a full path).

FedeMPouzols:
This looks good, and implements the best alternative considered in #17140 in a sound way.
I tested it with some ENGINX files and it behaved well. The calibration process runs correctly, and with log level `information` and `debug` I can see that it finds the correct files regardless of whether I set ENGIN-X as default instrument or not. Error handling (wrong numbers, empty values, etc.) is good too.
When we add other instruments (IMAT) this should work out of the box without issues.

So I definitely agree: :shipit:

PR #17290, opened 2016-08-19
Create python algorithm SelectNexusFilesByMetadata

### LINKED ISSUE #17272: Create algorithm for selecting nexus files based on metadata criteria
At ILL, a new algorithm is needed to select the nexus files from the given file list, which satisfy given metadata criteria, without loading the data itself.

gvardany:
This PR adds a new python algorithm for filtering of multiple nexus files by metadata without loading.

**To test:**

1. Collect some set of files.
2. Pick some nexus metadata entries, and build some criteria from them.
3. Run the algorithm following the example in docs.
4. Check in the returned files, that the criterion is indeed satisfactory.
5. Try more edge cases you can think of.

Fixes #17272 .

---
#### Reviewer

##### Code Review
##### Functional Tests


AndreiSavici:
The algorithm works nicely with SNS event nexus files.
You disabled `consider-using-enumerate`. You should use enumerate instead of the range of len.

```
for i in range(len(splitted)):
    if i%2==1:
        print splitted[i]
```

could be replaced by:

```
for i,s in enumerate(splitted):
    if i%2==1:
        print s
```

In the help file you should mention that if your Nexus entry is an array, it will use the first value (some people would probably want to use the average) 

**The main problem with this algorithm is the name.** `FindFiles` should be replaced with something like `SelectHDFFilesByCriteria` or something like that. You don't need to find files, you select them. And not all files, only hdf ones, that meet a certain condition.  Your current name is too general, and does not explain what the algorithm does

gvardany:
@AndreiSavici Thanks for the review, I will change to enumerate in the loop and add a note about the first value in array. Regarding the name, I was first thinking about `FilterNexusFilesByMetadata` what do you think?

AndreiSavici:
@gvardany `Filter` might mean that you do some event filtering. I would use `SelectNexusFilesByMetadata`

gvardany:
PR was closed due to renaming of branch. See #17300.

PR #17324, opened 2016-08-24
Update class_maker scripts

martyngigg:
Fixes an issue with `class_maker` where if a `CMakeLists.txt` file contained more than 1 reference to a variable such as `SRC_FILES` then the tool would try to modify all instances and the file would be corrupted. The script now only modifies the first instance of `set ( SRC_FILES` etc found. In addition there is also a new `--no-cmake` argument to prevent any modifications to cmake files.

I have also taken the opportunity to modify the scripts to be compatible with Python 2 & 3.

**To test:**

Use `class_maker` to generate a file in the `Geometry` library. The files should be created and the modifications to the `CMakeLists.txt` should only touch the expected lines at the top and **not** modify lines referencing opencascade later in the file.

Try out the `--no-cmake` option to confirm no cmake files are touched.

No issue. Internal change.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/buildconfig/class_maker.py
+++ b/buildconfig/class_maker.py
@@ -167,7 +167,7 @@ namespace {subproject} {{
 #======================================================================
 def write_test(subproject, classname, filename, args):
     """Write a class test file"""
-    print("Writing test file to",filename)
+    print("Writing test file to", filename)
     f = open(filename, 'w')
 
     guard = "MANTID_{}_{}TEST_H_".format(subproject.upper(), classname.upper())
--- a/buildconfig/class_maker.py
+++ b/buildconfig/class_maker.py
@@ -1,30 +1,30 @@
 #!/usr/bin/env python
 """ Utility for generating a class file, header, and test file """
+from __future__ import (absolute_import, division, print_function, unicode_literals)
 
 import sys
 import os
 try:
     import argparse
     useArgparse = True
-except ImportError, e:
+except ImportError as e:
     import optparse # deprecated in v2.7
     useArgparse = False
 import datetime
 import re
 import cmakelists_utils
 from cmakelists_utils import *
-import commands
 
 VERSION = "1.0"
 
 #======================================================================
 def write_header(subproject, classname, filename, args):
     """Write a class header file"""
-    print "Writing header file to %s" % filename
+    print("Writing header file to", filename)
     f = open(filename, 'w')
 
     subproject_upper = subproject.upper()
-    guard = "MANTID_%s_%s_H_" % (subproject_upper, classname.upper())
+    guard = "MANTID_{}_{}_H_".format(subproject_upper, classname.upper())
 
     # Create an Algorithm header; will not use it if not an algo
     algorithm_header = """
@@ -47,17 +47,18 @@ private:
         alg_include = ""
 
     # The full text
-    s = """#ifndef %s
-#define %s
+    s = r"""#ifndef {}
+#define {}
 
-%s
-#include "Mantid%s/DllConfig.h"
-namespace Mantid {
-namespace %s {
+#include "Mantid{}/DllConfig.h"
+{}
 
-/** %s : TODO: DESCRIPTION
+namespace Mantid {{
+namespace {} {{
 
-  Copyright &copy; %s ISIS Rutherford Appleton Laboratory, NScD Oak Ridge
+/** {} : TODO: DESCRIPTION
+
+  Copyright &copy; {} ISIS Rutherford Appleton Laboratory, NScD Oak Ridge
   National Laboratory & European Spallation Source
 
   This file is part of Mantid.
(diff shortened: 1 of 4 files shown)


buildconfig/class_maker.py
```diff
@@ -78,13 +79,13 @@ def write_header(subproject, classname, filename, args):
...
+class MANTID_{}_DLL {}{} {{
+public:{}}};
 
-} // namespace %s
-} // namespace Mantid
+}} // namespace {}
+}} // namespace Mantid
 
-#endif /* %s */""" % (guard, guard, subproject,
+#endif /* {} */""".format(guard, guard, subproject,
```
soininen: Here one could save a bit of repetition with regards to the format parameters if named arguments were used:
```python
"""#ifndef {guard}
#define {guard}
#include "Mantid{subproject}/Dll.Config.h"
# [...]
#endif /* {guard} */""".format(guard = guard, subproject = subproject)
```
 Might also make the template strings easier to comprehend.

buildconfig/class_maker.py
```diff
@@ -141,29 +142,29 @@ def write_source(subproject, classname, filename, args):
...
-} // namespace Mantid
-""" % (
+    s = """#include "Mantid{}/{}{}.h"
+
+namespace Mantid {{
+namespace {} {{
+{}{}
+}} // namespace {}
+}} // namespace Mantid
+""".format(
```
soininen: Also here formatting field names might be nice.

buildconfig/class_maker.py
```diff
@@ -174,24 +175,24 @@ def write_source(subproject, classname, filename, args):
 #======================================================================
 def write_test(subproject, classname, filename, args):
     """Write a class test file"""
-    print "Writing test file to %s" % filename
+    print("Writing test file to",filename)
```
soininen: Missing space before `filename`?

buildconfig/class_maker.py
```diff
@@ -207,39 +208,39 @@ def write_test(subproject, classname, filename, args):
...
-  }
+  }}
 
 
-};
+}};
 
 
-#endif /* %s */""" % (
+#endif /* {} */""".format(
```
soininen: Formatting field names?

buildconfig/cmakelists_utils.py
```diff
@@ -28,27 +30,29 @@ def find_basedir(project, subproject):
...
     read all the lines to get all the files,
     add your new line,
     sort them,
-    rewrite. Yay!"""
-
+    rewrite. Only touches first section found to avoid messing up any other set
+    sections in the rest of the file
+    """
     search_for1 = "set ( %s" % cmake_tag
     search_for2 = "set (%s" % cmake_tag
```
soininen: Shouldn't we use the new-style string formatting here as well?

buildconfig/delete_class.py
```diff
@@ -37,8 +37,8 @@ def delete_all(subproject, classname, args):
     # Insert into the cmake list
     remove_from_cmake(subproject, classname, args, args.source_subfolder)
 
-    print "   Files were removed to Framework/%s/CMakeLists.txt !" % subproject
-    print
+    print("   Files were removed to Framework/%s/CMakeLists.txt !" % subproject)
+    print()
```
soininen: New style string formatting?

buildconfig/move_class.py
```diff
@@ -44,7 +44,7 @@ def move_one(subproject, classname, newproject, newclassname, oldfilename, newfi
         f = open(newfilename, 'w')
         f.write(text)
     except RuntimeError as err:
-        print err
+        print(err)
```
soininen: Should we write error to stderr?

buildconfig/move_class.py
```diff
@@ -64,16 +64,16 @@ def move_all(subproject, classname, newproject, newclassname, args):
...
+        print("\nError! Header file %s already exists. Use --force to overwrite.\n" % newheaderfile)
         return
     if args.cpp and not overwrite and os.path.exists(newsourcefile):
-        print "\nError! Source file %s already exists. Use --force to overwrite.\n" % newsourcefile
+        print("\nError! Source file %s already exists. Use --force to overwrite.\n" % newsourcefile)
         return
     if args.test and not overwrite and os.path.exists(newtestfile):
-        print "\nError! Test file %s already exists. Use --force to overwrite.\n" % newtestfile
+        print("\nError! Test file %s already exists. Use --force to overwrite.\n" % newtestfile)
         return
```
soininen: New style string formatting?

soininen:
I tried to test this by running commands like

```
python2 buildconfig/class_maker.py Geometry ClassMakerTestGeometry
```

and

```
python2 buildconfig/class_maker.py --subfolder Rendering Geometry OCRenderingStuff1
```

on the master branch, but didn't see any problems with the `CMakeLists.txt` files. Could you perhaps elaborate how to reproduce the error cases? The changes to `redo_cmake_section()` seems to be doing what is expected.

Anyway, after the changes, the script runs on Python 3.4.3 and both python 2.7.6 and 3.4.3 generate the same files and changes. The small enhancements to the generated files compared to previous behavior are welcome, I think.

martyngigg:
Thanks for checking this over. It's a good point about the formatting specifiers with keywords, I was being lazy by not updating them. I'll take a look at that.

It looks like there are no files left on `master` that exhibit the original error. The problem with the `Geometry` file was resolved by https://github.com/mantidproject/mantid/commit/b8c570c0003bfbc1c33397272b5a96403000a136. I have a branch where I have a similar issue in a different file so I was prompted to solve it for that. If you revert back to the way the `CMakeLists.txt` was in that diff then you should see that the new script still works.

soininen:
I run the script before and after the changes in the `Geometry`'s `CMakeLists.txt` and indeed, there was some unexpected modifications to the `CMakelists.txt` file before the fixes of this PR. This also brings us closer to Python 3 compatibility. Great job! :shipit:

PR #17332, opened 2016-08-24
Work around gcc 5 internal compiler error

peterfpeterson:
With gcc 5.3.1 there was an error associated with `DataObjects::Peak`:

```
Framework/DataObjects/src/Peak.cpp:239:7: error: function 'Mantid::DataObjects::Peak& 
Mantid::DataObjects::Peak::operator=(Mantid::DataObjects::Peak&&)' defaulted on its 
redeclaration with an exception-specification that differs from the implicit exception-specification ''
Peak &Peak::operator=(Peak &&) noexcept = default;
```

It appears to be exercising an [internal compiler error](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=58265). Moving the `default` into the header fixes gcc 5, but breaks gcc 4. gcc 6 is happy with the existing code as well. This PR is a work-around specific to gcc 5.

**To test:**

If the builds pass, it is good.

_No associated issue_

_Does not need to be in the release notes_ since it isn't user facing

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/DataObjects/inc/MantidDataObjects/Peak.h
+++ b/Framework/DataObjects/inc/MantidDataObjects/Peak.h
@@ -51,6 +51,9 @@ public:
 #if defined(_MSC_VER) && _MSC_VER <= 1900
   Peak(Peak &&);
   Peak &operator=(Peak &&);
+#elif defined(__GNUC__) && (__GNUC__ == 5)
+  Peak(Peak &&) noexcept = default;
+  Peak &operator=(Peak &&) noexcept = default;
 #else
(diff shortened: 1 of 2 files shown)


peterfpeterson:
Killed the RHEL7 build as another change is about to be pushed.

AndreiSavici:
Works on Ubuntu 16.04. :shipit:

FedeMPouzols:
All the platforms have built succesfully. This also works for me with GCC 5.3.1. So I think it is safe to merge this.

PR #17337, opened 2016-08-25
SelectNexusFilesByMetadata to require h5py

### LINKED ISSUE #17335: SelectNexusFilesByMetadata to require h5py
`SelectNexusFilesByMetadata` requires `h5py`, so it should register itself only if it is installed.

gvardany:
This PR makes aforementioned algorithm to run only if `h5py` package is present. 
The algorithm will be subscribed (and documented) regardless, but it will not run (neither be tested) if the required package is missing.

Fixes #17335.

#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/PythonInterface/plugins/algorithms/SelectNexusFilesByMetadata.py
+++ b/Framework/PythonInterface/plugins/algorithms/SelectNexusFilesByMetadata.py
@@ -1,7 +1,6 @@
 #pylint: disable=eval-used
 from __future__ import (absolute_import, division, print_function)
 
-import h5py
 from mantid.simpleapi import *
 from mantid.kernel import *
 from mantid.api import *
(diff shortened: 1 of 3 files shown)


Framework/PythonInterface/test/python/plugins/algorithms/SelectNexusFilesByMetadataTest.py
```diff
@@ -56,4 +57,10 @@ def test_string_criteria(self):
         self.assertTrue(outfiles[0].endswith('ILLD33_001030.nxs'),'Should be the file name')
 
 if __name__=="__main__":
-    unittest.main()
+    # run the test if only if the required package is present
+    try:
+        import h5py
+    except ImportError:
+        pass
+    else:
```
MichaelWedel: Probably you could get away without the `else`-clause:
```
try:
    import h5py
    unittest.main()
except ImportError:
    pass```

File: Framework/PythonInterface/plugins/algorithms/SelectNexusFilesByMetadata.py [hunk 2]
```diff
@@ -52,6 +51,13 @@ def PyInit(self):
                              doc='Comma separated list of the fully resolved file names satisfying the given criteria.')
 
     def PyExec(self):
+
+        # run only if h5py is present
+        try:
+            import h5py
+        except ImportError:
+            raise RuntimeError('This algorithm requires h5py package')
```
MichaelWedel: In other cases there's a link to the package on PyPI, which gives users a hint where to get the package. If h5py can be installed via pip, I think it would be a good idea to include such a link here (and maybe also in the documentation).

MichaelWedel:
Tested locally without h5py package installed, the tests pass and the documentation is generated. :shipit:

PR #17343, opened 2016-08-25
Update documentation for FindSXPeaks property

### LINKED ISSUE #17253: Update resolution parameter documentation for FindSXPeaks
The resolution parameter is a little vague in what it actually means. We should add more information about what it is actually defined as. Unit is described as 'pixels'. It is used to compare if two peaks are the same:

```cpp
  bool compare(const SXPeak &rhs, double tolerance) const {
    if (std::abs(_t / npixels - rhs._t / rhs.npixels) >
        tolerance * _t / npixels)
      return false;
    if (std::abs(_phi / npixels - rhs._phi / rhs.npixels) >
        tolerance * _phi / npixels)
      return false;
    if (std::abs(_th2 / npixels - rhs._th2 / rhs.npixels) >
        tolerance * _th2 / npixels)
      return false;
    return true;
  }
```

samueljackson92:
Updated the documentation to attempt to better explain what the property actually does.

**To test:**
- Make sure the doctests are passing.
- Read the changes. Do they make sense?

Fixes #17253 .

---
#### Reviewer

##### Code Review
##### Functional Tests

This adds a longer description of the resolution parameter and corrects
some of the wording of the docs.

### DIFF
--- a/Framework/Crystal/inc/MantidCrystal/FindSXPeaks.h
+++ b/Framework/Crystal/inc/MantidCrystal/FindSXPeaks.h
@@ -197,9 +197,9 @@ public:
   const std::string name() const override { return "FindSXPeaks"; }
   /// Summary of algorithms purpose
   const std::string summary() const override {
-    return "Takes a 2D workspace as input and find the FindSXPeaksimum in each "
-           "1D spectrum. This is used in particular for single crystal as a "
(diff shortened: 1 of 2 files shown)


AndreiSavici:
Do you want to add it to the release notes?
Otherwise things look OK.
:shipit:, depending on the reply to the above question

PR #17347, opened 2016-08-26
Rewrite of Chop with new GUI, including MultiRep for LET.

### LINKED ISSUE #10147: Inelastic planning tools - Porting Chop functionality
This issue was originally [TRAC 9304](http://trac.mantidproject.org/mantid/ticket/9304)

This ticket is blocked by : 
 - [X] [TRAC10611](http://trac.mantidproject.org/mantid/ticket/10611)
 - [X] [TRAC10940](http://trac.mantidproject.org/mantid/ticket/10940)

Includes porting Horace Planner, Chop, etc to Mantid.  

 - - - - 
Keywords: SSC,2014,All

mducle:
This is a rewrite of PyChop with a new GUI and CLI. The PyChop code previously in Mantid (by JWT) was based on an incomplete Matlab port of the original Fortran code (by TGP), whilst this version includes a more faithful translation (by RAE) of the Fortran. In addition it includes code (by RIB and DJV) to calculate the multiple reps allowed by Merlin and LET, and the resolution of LET which CHOP does not support.

**To test:**

Run through the attached IPython notebook: 
[PyChop2_ipynb.zip](https://github.com/mantidproject/mantid/files/438300/PyChop2_ipynb.zip)

Run MantidPlot - there should be a new interface under `Direct` called `PyChop`. 

Or, in a MantidPython shell do:

```
from PyChop import PyChop2, PyChopGui
merlin = PyChop2('merlin','s',250)
merlin.setEi(25)
resolution = merlin.getResolution()
print resolution
resolution, flux = PyChop2.calculate(inst='maps', chtyp='a', freq=500, ei=600, etrans=range(0,550,50))
print resolution
print flux
PyChop2.showGUI()
```

Expected output:

```
[ 0.53847416  0.54998197  0.56540704  0.5843418   0.6065874   0.63201212
  0.66050558  0.691962    0.72627472  0.76333545  0.80303543  0.84526708
  0.88992562  0.93691037  0.98612563  1.03748123  1.09089277  1.14628162
  1.20357476]
[ 17.12380445  16.00789732  14.95785911  13.97730028  13.07002193
  12.23992828  11.49090381  10.82667552  10.25072764   9.76644397
   9.37794616]
2379.08526288
```

Fixes #10147 

---
#### Reviewer

##### Code Review
##### Functional Tests


AndreiSavici:
First impression when I looked at the GUI: no help button
I am not sure the flux calculations are correct. I doubt that the flux on MAPS increases as steeply as in the figure below
flux

AndreiSavici:
Something is wrong in the rst file for help. It does not show the second reference correctly
screenshot from 2016-09-16 17-49-22

In the first equation is it `\Sigma v` or is it squared, as in the text above?

mducle:
> No help button

Ok. I'll put it in by Monday.

> Flux-Ei plot does not hold plot (I want to overplot different choppers)

Yes. I thought that it might get too confusing to have too many lines on there. But I can see why overplotting would be useful. Also, would it be better to have a side-by-side plot (e.g. like in Flux-Freq) or keep the shared x-axes plot?

> Something is wrong in the rst file.

Yes. I don't know why it does that. Probably me not understanding how rst works - I want to reference the same reference twice and it seems to do this instead of giving one index.

> In the first equation is it squared

Oops! Yes you're right. It should be squared.

scripts/PyChop/ISISFermi.py
```diff
@@ -0,0 +1,506 @@
...
+        Output:
+            flux - the flux in n / cm**2 . uA.s
+        """
+        Ei = self.Ei if Ei_in is None else Ei_in
+        if Ei is None:
+            raise ValueError('Incident energy has not been specified')
+        if frequency:
+            oldfreq = self.freq
+            self.setFrequency(frequency)
+        moderator_flux = self.getMeasuredFlux(Ei)
```
mducle: > I am not sure the flux calculations are correct.

The flux is taken here from white beam measurements using a calibrated monitor by Rob Bewley. The measurements just keeps going up at high energies / low wavelengths and go out to about 4eV (0.15 Angstrom). I'll ask Rob about the measurements.

If this line is modified to use `getAnalyticFlux` instead, then you see the expected fall off at high energies.

scripts/PyChop/ISISDisk.py
```diff
@@ -0,0 +1,384 @@
...
+        res_list = []
+        if not np.shape(Et):
+            Et = [Et]
+        for ie in ie_list:
+            t_mod_chop = 252.82 * lastChopDist * np.sqrt(81.81/Eis[ie])
+            res = []
+            for en in Et:
+                Ef = Eis[ie] - en
+                fac = (Ef/Eis[ie])**1.5
+                chopRes = (2*chop_width[ie]/t_mod_chop) * ((self.samp_det+self.chop_samp+lastChopDist)/self.samp_det)*fac
```
AndreiSavici: There is a missing term in this formula. It should read 
`chopRes = (2*chop_width[ie]/t_mod_chop) *(1+ ((self.samp_det+self.chop_samp+lastChopDist)/self.samp_det)*fac)` Same error in the documentation. I will attach the derivation

docs/source/interfaces/PyChop.rst
```diff
@@ -0,0 +1,191 @@
...
+
+The time variances above are defined at the moderator and chopper positions respectively.
+As the neutron bunches travel towards the sample and detector they also spread out,
+and the final time (energy) widths are determined by the geometry (distances) of the
+instrument. Specifically, the relative energy width is given by the sum in quadrature
+of each of the contributing time widths, which we will restrict here to the two major
+terms, :math:`\tau_{\mathrm{mod}}` and :math:`\tau_{\mathrm{chop}}` `[4]`_:
+
+.. math:: \left( \frac{\Delta E}{E_i}\right )^2 = 
+        \left[ 2\frac{\tau_{\mathrm{chop}}}{t_{\mathrm{chop}}} \frac{l_0+l_1}{l_2}
```
AndreiSavici: There should be `1+\frac{l_0+l_1}{l_2} \left(\frac{E_f}{E_i}\right)^\frac{3}{2}`

AndreiSavici:
Looks good :shipit:

scripts/PyChop/ISISDisk.py
```diff
@@ -0,0 +1,384 @@
...
+        Returns the instrument variant name
+        """
+        return self.variant
+
+    def setFrequency(self, frequency, **kwargs):
+        """
+        Sets the chopper frequencies, in Hz.
+        If scalar, sets the resolution chopper freq to this and the pulse remover to freq/2
+        """
+        if 'LET' in self.instname:
```
FedeMPouzols: This has quite a few branches where bugs could remain hidden, and I'm not sure how well this is tested/testable. Maybe it could be reorganized, use dictionaries, etc.
(I understand this code has a long history behind, just noting in case there's some chance for further refactoring.)

mducle: You're right that it should be reorganised. I don't really like it that at the moment there are almost entirely separate branches for LET, and for MERLIN/MAPS/MARI. But, LET was enough different that it would require rewritting a lot of the legacy code for the other instruments to accomodate it (the code for the other instruments is essentially a port of a FORTRAN program - so should probably also be rewritten). 

We have future plans to discuss with Doug and other SNS instrument scientists about a unified framework, which can also handle CNCS/ARCS/SEQUOIA (probably not HYSPEC, but possibly also including it). 

Also, the code for MERLIN/MAPS/MARI can't calculate multiple-Ei, which will be an option on all three instruments in future.

So, I'll like change this tool in the next release, but in the mean time I would like to push something which is underneath the same as the tools currently used, but with a better UI.

AndreiSavici:
Here is the link to the derivation of the resolution function. I've included a term tor detector timing as well. https://github.com/mantidproject/documents/blob/master/Design/PyChop-resolution/energy_resolution.pdf

PR #17363, opened 2016-08-30
SaveDiffCal optional workspaces

peterfpeterson:
The `MaskWorkspace` and `GroupingWorkspace` are not actually optional. If they are not supplied there is a null pointer de-reference that crashes mantid.

**To test:**

Since there are new unit tests that exercise the bug, just review the code and mark it :squirrel:

_There is no associate issue_

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/DataHandling/src/SaveDiffCal.cpp
+++ b/Framework/DataHandling/src/SaveDiffCal.cpp
@@ -123,12 +123,11 @@ void SaveDiffCal::writeIntFieldFromSVWS(
   std::vector<int32_t> values(m_numValues, 1);
 
   if (bool(ws)) {
-    int32_t value;
     for (size_t i = 0; i < m_numValues; ++i) {
       auto &ids = ws->getSpectrum(i).getDetectorIDs();
       auto found = m_detidToIndex.find(*(ids.begin()));
       if (found != m_detidToIndex.end()) {
-        value = static_cast<int32_t>(ws->getValue(found->first));
+        int32_t value = static_cast<int32_t>(ws->getValue(found->first));
         // in maskworkspace 0=use, 1=dontuse - backwards from the file
         if (isMask) {
           if (value == 0)
--- a/Framework/DataHandling/src/SaveDiffCal.cpp
+++ b/Framework/DataHandling/src/SaveDiffCal.cpp
@@ -122,8 +122,8 @@ void SaveDiffCal::writeIntFieldFromSVWS(
   // output array defaults to all one (one group, use the pixel)
   std::vector<int32_t> values(m_numValues, 1);
 
-  int32_t value;
   if (bool(ws)) {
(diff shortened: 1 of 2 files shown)


rosswhitfield:
It works. :shipit:

Framework/DataHandling/src/SaveDiffCal.cpp
```diff
@@ -117,26 +117,27 @@ void SaveDiffCal::writeIntFieldFromSVWS(
...
+      auto &ids = ws->getSpectrum(i).getDetectorIDs();
+      auto found = m_detidToIndex.find(*(ids.begin()));
+      if (found != m_detidToIndex.end()) {
+        value = static_cast<int32_t>(ws->getValue(found->first));
+        // in maskworkspace 0=use, 1=dontuse - backwards from the file
+        if (isMask) {
+          if (value == 0)
+            value = 1;
+          else
+            value = 0;
```
quantumsteve: Are these last four lines the same as `value = (value != 0)`?

Framework/DataHandling/src/SaveDiffCal.cpp
```diff
@@ -117,26 +117,27 @@ void SaveDiffCal::writeIntFieldFromSVWS(
...
-      // in maskworkspace 0=use, 1=dontuse
-      if (isMask) {
-        if (value == 0)
-          value = 1;
-        else
-          value = 0;
+  std::vector<int32_t> values(m_numValues, 1);
+
+  if (bool(ws)) {
+    int32_t value;
```
quantumsteve: Is the scoping of `value` correct?

Framework/DataHandling/src/SaveDiffCal.cpp
peterfpeterson: This "normalizes" the values to be `0` or `1`. Depending on how the mask is made the values can be `0` or any other integer.

Framework/DataHandling/src/SaveDiffCal.cpp
peterfpeterson: Outside of the loop to reduce the amount of create/destroy that happens with the variable. Probably could put it inside the loop and the compiler work out memory.

quantumsteve: I don't think the performance argument holds for primitive types on the stack. Unless you have a benchmark showing otherwise, I'd put it inside the loop.

Framework/DataHandling/src/SaveDiffCal.cpp
quantumsteve: Oops, I had the logic backwards ( should it actually be `value = (value == 0)`? ). I think it's only a style difference and will merge whichever one you prefer.

PR #17378, opened 2016-08-31
Make SingletonHolder<T>::Instance() thread-safe

### LINKED ISSUE #17364: Should SingletonHolder<T>::Instance()  use double-checked locking?
### Expected behavior

singleton can be called from multiple threads

### Actual behavior

I noticed that [SingletonHolder<T>::Instance()](https://github.com/mantidproject/mantid/blob/adabe9c92d3b45d8125c7d1bb5514552dd1e68b5/Framework/Kernel/inc/MantidKernel/SingletonHolder.h#L79) just has a single check on `pinstance`. Since it is called concurrently from multiple threads, should this be replaced with  [double-checked locking](https://en.wikipedia.org/wiki/Double-checked_locking), `std::call_once`, or a static initializer?

### Platforms affected

All

quantumsteve:
Description of work.

This changes `SingletonHolder<T>::Instance` to use thread-safe function local static initialization. The current implementation with a single check is not necessarily thread-safe.

**To test:**

Thorough code review. C++11 magic statics are not widely used in the Mantid codebase. 
Which tests access singletons from multiple threads?

Fixes #17364.

---
#### Reviewer

##### Code Review
##### Functional Tests


quantumsteve:
Getting the visibility correct on all three platforms will take a lot of work. I think it is best to reduce the size of the changes.

PR #17380, opened 2016-08-31
Fixes multiple minor issues with HB3A reduction

### LINKED ISSUE #16930: Correct header of Fullprof peak file exported from HB3A
For neutron only diffraction (i.e., no K vector), the header should be 
3i4, ...
The magnetic data should start with
4i4, ...
### Expected behavior

For applying k-shift
1. it should be integrated with 'Index Peak'.
2. calculate [(HKL)  - (K)] and round it to nearest integer. Then calculate the difference between the original value and its rounded value.  Average the difference among all the peaks and add back to the original K.

wdzhou:
Description of work.

Resolve the issues:
1. Exp 524, Scan 180: half peak is scanned
2. Exp 524, Scan 188 and 206
3. Exp 533, Scans 331-333

**To test:**

All the cases presented above should be passed.

Fixes #16930.

---
#### Reviewer

##### Code Review
##### Functional Tests


scripts/HFIR_4Circle_Reduction/absorption.py
```diff
@@ -0,0 +1,363 @@
+import numpy
+import numpy.linalg
+import math
+from mantid.api import AnalysisDataService
+
+# Do absorption correction
+
+
+class Lattice(object):
```
rosswhitfield: Why do you implement your own Lattice class, why not use Mantids? `from mantid.geometry import UnitCell` It appears to have everything in it you require, see [here](http://docs.mantidproject.org/nightly/concepts/Lattice.html#unit-cells).

wdzhou: Just tried to run the same script as the scientist.  It will be deleted later in my code.

VickieLynch:
Works well with exp 474. :shipit: Another ticket should give warning if scan is not available to download for merge.  Also a ticket should work on save/load project improvements.

Issue #17271, opened 2016-08-18
Engg: Calibrate causes Mantid to crash whilst fitting peaks


DavidFair:
Currently whilst running `EnggCalibrate` through the Engineering interface an exception is thrown by Python most likely whilst fitting peaks.

This is caused because the python script throws `PythonExcept` whereas the catch block catches `std::runtime_error` at https://github.com/mantidproject/mantid/blob/master/MantidQt/CustomInterfaces/src/EnggDiffraction/EnggDiffractionPresenter.cpp#L1147

This has been introduced since 3.7.2 where it runs correctly.

DavidFair:
Working with @martyngigg we have traced this back to how exceptions thrown from python are handled by `Algorithm`. When the PR is merged fixing the issue we need to test calibrate runs correctly and handles `EnggFitPeaks` throwing a `runtime_error`. 
We then need to analyse why the following calibration inputs causes `EnggFitPeaks` to not find any peaks which makes it throw said exception with the message `Could not fit any peak. Failed to fit peak type`

Vanadium WS: 256355
Cerium Oxide WS: 252415

DavidFair:
Was fixed in #17391

PR #17391, opened 2016-09-01
Engg:: Stop calibration if peaks cannot be found

### LINKED ISSUE #17282: Engg:: Calibrate should not continue if EnggFitPeaks fails
### Expected behavior
`EnggCalibrate` should exit with an error if `EnggFitPeaks` which it calls fails for any reason
### Actual behavior
`EnggCalibrate` currently continues regardless of the success of `EnggFitPeaks` this can lead to a bad calibration state and subsequent errors in the data processing. 
Open the Engineering Diffraction Interface
Enter the following runs
Van: 256355
Cerium: 252415
Run calibrate - `EnggFitPeaks` should fail to find any peaks and throw a `runtime_error` which calibrate catches and continues regardless
### Platforms affected
All

Additionally we attempt to catch any exceptions thrown by `EnggFitPeaks` this will not work without setting the algorithm to re-throw these exceptions. 
The better approach is to execute the algorithm then check that is has executed successfully before continuing whilst letting the `Algorithms` class handle all exceptions on our behalf. 

Could be completed with #17283 as they would work well together

DavidFair:
Description of work.
This pull request makes the Engineering Diffraction interface stop calibrating if `EnggFitPeaks` fails to find any peaks. As the algorithm was not set to re-throw the try catch block was not doing anything so instead we check if it successfully executed. 

**To test:**
Open the `Engineering Diffraction` interface
The input files are located at `\\olympic\Babylon5\Public\David Fairbrother\z_test`

Enter Vanadium run: `256355`
Enter Cerium run: `255415`
Run calibrate. This should fail and not produce an output. 

---

There are two additional minor fixes which were completed with this issue.
The `Engineering Diffraction Testing Guide` should now appear under the interfaces category in documentation.

On opening and closing the `Engineering Interface` in Mantid the temporary workspace `engggui_calibration_banks_parameters` should now be removed.

Fixes #17282 and #17271 .

---
#### Reviewer

##### Code Review
##### Functional Tests


FedeMPouzols:
Looks good. This makes sure that the calibration process is stopped early and gracefully when the calibration algorithm doesn't run correctly (for example the fit results is too bad because of poor or mismatched input calibration data). It also fixes two other minor issues in the interface clean up code and its documentation.

:shipit:

PR #17394, opened 2016-09-02
Add fake SpectrumInfo

### LINKED ISSUE #17322: Add class SpectrumInfo as a wrapper for basic instrument access
`SpectrumInfo` will be part of Instrument-2.0. To make refactoring easier at a later stage, we add this now to encapsulate direct access to detector IDs and the instrument.

SimonHeybrock:
This PR adds `API::SpectrumInfo`. `SpectrumInfo` is intended to be part of Instrument-2.0. The class added here is an intermediate step:
- The interface is identical or very similar to what we plan for Instrument-2.0.
- Functionality is provided by wrapping interaction with the existing workspace and instrument APIs.

The introduction of `SpectrumInfo` leads to more concise code in many algorithms. Furthermore it will make the rollout of Instrument-2.0 easier, and also helps to slightly reduce the number of direct use of detector IDs. The latter helps for the rollout of indexing changes, which in turn are a prerequisite for Instrument-2.0 and better MPI support.

Currently there are a few workarounds that are necessary as long as `SpectrumInfo` is only wrapping existing functionality. This will be cleaned up once we put the actual implementation in place:
- `SpectrumInfo` is reallocated with every access, since otherwise we cannot tell if the lower layers have changed (in particular the `ParameterMap`).
- Reallocation implies issues with threading. The getter `MatrixWorkspace::spectrumInfo()` is thus not thread safe. For OpenMP we are enforcing this by checking the thread ID. We have no safety net for other threading libraries (similar to `EventWorkspace`).
- Currently `MatrixWorkspace` or rather `ISpectrum` can contain detector IDs that are not part of the instrument. Many algorithms dealt with this by catching (and ignoring) exceptions related to this. Instrument-2.0 will not accept invalid detector IDs. However, this is a really major change/fix. Thus, the current `API::SpectrumInfo` silently ignores such invalid IDs, making the behavior basically in line with the current behavior.

Todos (for later):
- Figure out if and how we can deal with Algorithms that use the parameter map, e.g., for obtaining `EFixed`. Should this be done via `SpectrumInfo`?
- Provide accessors to detectors, e.g., `Detector_sptr SpectrumInfo(const size_t) const` (return type is legacy, for now).

Both of these are not clear yet, basically this is a design decision for Instrument-2.0 that we have not made yet.

**To test:**

Code review.

Fixes #17322.

---
#### Reviewer

##### Code Review
##### Functional Tests


SimonHeybrock:
Note that I simply based `SpectrumInfo` on the existing class `GeometryInfo`. As a next step I would remove `GeoemtryInfo` (which precedes our Instrument-2.0 design)  and put the functionality directly into `SpectrumInfo`.

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -179,6 +181,20 @@ const std::string MatrixWorkspace::getTitle() const {
...
 }
 
+/** Return a reference to the SpectrumInfo object. NOT THREAD-SAFE!
+ *
+ * Currently this method is not thread-safe, and any modifications of the
+ * instrument or instrument parameters such as mask flags will not be visible in
+ * existing SpectrumInfo references. After modification, obtain a new reference
+ * by calling this method again.
+ */
+const SpectrumInfo &MatrixWorkspace::spectrumInfo() const {
```
SimonHeybrock: I now added a safety check at this position (not pushed, since there are other local changes that would make a first review more difficult):

```cpp
  if(PARALLEL_THREAD_NUMBER != 0)
    throw std::runtime_error("MatrixWorkspace::spectrumInfo(): Thread ID is "
                             "not 0. This indicates that the method is called "
                             "in a threaded context. This is not allowed.");
```

Framework/API/src/SpectrumInfo.cpp
```diff
@@ -0,0 +1,42 @@
...
+  return getInfo(index).isMasked();
+}
+
+const GeometryInfo &SpectrumInfo::getInfo(const size_t index) const {
+  updateCachedInfo(index);
+  size_t thread = static_cast<size_t>(PARALLEL_THREAD_NUMBER);
+  return *m_info[thread];
+}
+
+void SpectrumInfo::updateCachedInfo(const size_t index) const {
```
OwenArnold: Looks like it is absolutely necessary to make this call in many situations?

Framework/API/src/SpectrumInfo.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include "MantidAPI/GeometryInfo.h"
+#include "MantidAPI/SpectrumInfo.h"
+#include "MantidKernel/MultiThreaded.h"
+#include "MantidKernel/make_unique.h"
+
+namespace Mantid {
+namespace API {
+
+SpectrumInfo::SpectrumInfo(const MatrixWorkspace &workspace)
+    : m_factory{Kernel::make_unique<GeometryInfoFactory>(workspace)},
```
OwenArnold: I think what's been doing here is good, but taking a look at `GeometryInfo` and how that interplays with `GeometryInfoFactory` I think things are a little confused. `GeometryInfo` is created with a back-reference to the `GeometryInfoFactory` and `GeometryInfoFactory` does a lot more than just create `GeometryInfo`. I think a better design would be for `GeometryInfo` not to keep the back-reference and for `GeometryInfoFactory` to pass all the required parameters (source pos, sample pos etc) as constructor arguments. Or if performance is important some  kind of `InstrumentGeometryCore` containing that information that can be passed by shared_ptr to the `GeometryInfos`

Framework/API/src/SpectrumInfo.cpp
SimonHeybrock: I am using the OpenMP thread ID as an index, just like `EventWorkspaceMRU`. The difference is that I keep only one entry per thread, unlike the MRU which keeps 50 and uses some really complicated boost map for that purpose.

Framework/API/src/SpectrumInfo.cpp
SimonHeybrock: Basing `SpectrumInfo` on `GeometryInfo` was just an intermediate step, such that we could quickly see if the interface and the interaction with `MatrixWorkspace` was ok. I will push some more changes that clean this up in a minute.

OwenArnold:
@stuartcampbell This is the PR I just mentioned on slack. [Here's](https://github.com/DMSC-Instrument-Data/instrument-prototype/blob/master/cow_instrument/SpectrumInfo.h) the analogue in `Instrument 2.0`

SimonHeybrock:
I now added performance tests equivalent to those @OwenArnold added in https://github.com/mantidproject/mantid/pull/17400.

Results are as follows:

```
test_rotate_bank_and_read_positions_x10 time="0.934603"
test_move_bank_and_read_positions_x10" time="0.891934"
test_rotate_bank_and_read_positions_SpectrumInfo_x10" time="0.668242"
test_move_bank_and_read_positions_SpectrumInfo_x10" time="0.661005"
```

That is, a 30% speedup by using `SpectrumInfo` for reads, even for these tests which include writes.

There are also existing tests for the `SpectrumInfo` performance. Via the current `MatrixWorkspace` interface typical access (L2 and two-theta) takes 0.045 s, via `SpectrumInfo` 0.017 s.

stuartcampbell:
The code changes look good and all the tests pass, so I say :shipit:

FedeMPouzols:
This branch has developed a merge conflict. Maybe it's just the release notes?

PR #17400, opened 2016-09-02
Write rotate/translate performance tests

OwenArnold:
These are just tests. I think these should provide a reasonable comparison against [Instrument 2.0](https://github.com/DMSC-Instrument-Data/instrument-prototype/pull/32)

### DIFF
--- a/Framework/API/test/MatrixWorkspaceTest.h
+++ b/Framework/API/test/MatrixWorkspaceTest.h
@@ -7,6 +7,7 @@
 #include "MantidAPI/SpectraAxis.h"
 #include "MantidAPI/SpectrumDetectorMapping.h"
 #include "MantidAPI/WorkspaceFactory.h"
+#include "MantidGeometry/Instrument/ComponentHelper.h"
 #include "MantidGeometry/Instrument.h"
 #include "MantidGeometry/Instrument/Detector.h"
 #include "MantidKernel/make_cow.h"
@@ -15,6 +16,7 @@
 #include "MantidTestHelpers/FakeGmockObjects.h"
 #include "MantidTestHelpers/FakeObjects.h"
 #include "MantidTestHelpers/InstrumentCreationHelper.h"
+#include "MantidTestHelpers/ComponentCreationHelper.h"
 #include "MantidTestHelpers/NexusTestHelper.h"
 #include "PropertyManagerHelper.h"
 
@@ -1450,6 +1452,7 @@ private:
 };
 
 class MatrixWorkspaceTestPerformance : public CxxTest::TestSuite {
+
 public:
   static MatrixWorkspaceTestPerformance *createSuite() {
     return new MatrixWorkspaceTestPerformance();
@@ -1459,6 +1462,8 @@ public:
   }
 
   MatrixWorkspaceTestPerformance() : m_workspace(nullptr) {
+    using namespace Mantid::Geometry;
+
     size_t numberOfHistograms = 10000;
     size_t numberOfBins = 1;
     m_workspace.init(numberOfHistograms, numberOfBins, numberOfBins - 1);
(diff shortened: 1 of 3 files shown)


Framework/Geometry/test/ComponentHelperTest.h
```diff
@@ -117,4 +117,221 @@ class ComponentHelperTest : public CxxTest::TestSuite {
...
+    using namespace Mantid::Kernel;
+
+    ComponentHelper::rotateComponent(
+        *m_sansBank, *m_paramMap, m_zRotation,
+        Mantid::Geometry::ComponentHelper::Relative);
+
+    V3D pos;
+    for (int i = 1;
+         i <= static_cast<int>(m_sansInstrument->getNumberDetectors()); ++i) {
+      pos += m_sansInstrument->getDetector(i)->getPos();
```
SimonHeybrock: Accessing positions this way is cheaper than the access via `MatrixWorkspace`, I think. Is this intentional?

SimonHeybrock:
How about having benchmarks for different instrument sizes to make this more realistic? For example, 10k pixels, 100k pixels, and 1M pixels?

Framework/Geometry/test/ComponentHelperTest.h
OwenArnold: Yes. But it's an important point. I'll make sure that there are equivalents at the `MatixWorkspace` vs Instrument 2.0 `SpectrumInfo` level. I guess you'll be adding performance tests for the Instrument 1.0 `SpectrumInfo` too.

OwenArnold: In fact I'll add a comment about that.

Framework/API/test/MatrixWorkspaceTest.h
```diff
@@ -1486,8 +1514,51 @@ class MatrixWorkspaceTestPerformance : public CxxTest::TestSuite {
...
+    // move the bank
+    ComponentHelper::moveComponent(*m_sansBank, *m_paramMap, m_pos,
+                                   Mantid::Geometry::ComponentHelper::Relative);
+
+    V3D pos;
+    for (size_t i = 1; i < m_workspaceSans.getNumberHistograms(); ++i) {
+      pos += m_workspaceSans.getDetector(i)->getPos();
+    }
+  }
+
```
SimonHeybrock: Great! I can later simply copy these tests and use `SpectrumInfo`, which should give a read speed-up, i.e., we then have three cases to compare.

Framework/Geometry/test/ComponentHelperTest.h
```diff
@@ -117,4 +117,221 @@ class ComponentHelperTest : public CxxTest::TestSuite {
...
+
+    ComponentHelper::moveComponent(*m_sansBank, *m_paramMap, m_pos,
+                                   Mantid::Geometry::ComponentHelper::Relative);
+
+    V3D pos;
+    for (int i = 1;
+         i <= static_cast<int>(m_sansInstrument->getNumberDetectors()); ++i) {
+      pos += m_sansInstrument->getDetector(i)->getPos();
+    }
+  }
```
SimonHeybrock: @OwenArnold Regarding the comparison with Google benchmark. My point was that in tests like this one we also should execute this at least a few times in a loop. To compare results we might need to do some extra math, but the tests should be more realistic then (taking account cache effects etc.).

OwenArnold: @SimonHeybrock. Fair point. Shall I attempt to execute these each say 10x? I'll adjust the test names to indicate as much.

SimonHeybrock: Yes, 10x sounds reasonable :+1:

SimonHeybrock:
:shipit:

OwenArnold:
I notice that the builds didn't get run. Probably down to Jenkins?

PR #17406, opened 2016-09-05
Refs #17405. Refactoring of CompareSampleLogs.

### LINKED ISSUE #17405: Refactor CompareSampleLogs
### Expected behavior
Experience shows, that sometimes numeric sample logs are represented as strings. However, users expect them to be compared as numbers (with certain tolerance).

The aim of the refactoring is to implement the behaviour expected by users, e.g. compare temperatures as numbers, even if they have type 'String'.

### Platforms affected
All

Please, assign me.

mganeva:
Refactoring of ComateSampleLogs algorithm to fix the mentioned issue and some minor code improvements. 

2 new unit tests.

**To test:**
Check that nothing has got broken. The simple test for new behavoiur:

``` python
ws1 = CreateSampleWorkspace()
ws2 = CreateSampleWorkspace()
AddSampleLog(ws1, 'chopper_speed', '13995', 'String')
AddSampleLog(ws2, 'chopper_speed', '14004', 'Number')
result = CompareSampleLogs([ws1, ws2], 'chopper_speed', 10.0)
print result
```

the result must be an empty string.

Fixes #17405. 

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/PythonInterface/plugins/algorithms/CompareSampleLogs.py
+++ b/Framework/PythonInterface/plugins/algorithms/CompareSampleLogs.py
@@ -15,6 +15,8 @@ class CompareSampleLogs(PythonAlgorithm):
         Init
         """
         PythonAlgorithm.__init__(self)
+        self.tolerance = 0.0
+        self.wslist = []
 
     def category(self):
         """
@@ -39,8 +41,8 @@ class CompareSampleLogs(PythonAlgorithm):
 
         self.declareProperty(StringArrayProperty(name="SampleLogs", direction=Direction.Input, validator=validator),
                              doc="Comma separated list of sample logs to compare.")
-        self.declareProperty("Tolerance", 1e-3, validator=FloatBoundedValidator(lower=1e-7, upper=1.0),
-                             doc="Tolerance for comparison of double values.")
+        self.declareProperty("Tolerance", 1e-3, validator=FloatBoundedValidator(lower=1e-7, upper=100.0),
+                             doc="Tolerance for comparison of numeric values.")
         self.declareProperty("Result", "A string that will be empty if all the logs match, "
                              "otherwise will contain a comma separated list of  not matching logs", Direction.Output)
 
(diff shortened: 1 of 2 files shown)


DavidFair:
LGTM :shipit:

Framework/PythonInterface/plugins/algorithms/CompareSampleLogs.py
```diff
@@ -39,8 +41,8 @@ def PyInit(self):
 
         self.declareProperty(StringArrayProperty(name="SampleLogs", direction=Direction.Input, validator=validator),
                              doc="Comma separated list of sample logs to compare.")
-        self.declareProperty("Tolerance", 1e-3, validator=FloatBoundedValidator(lower=1e-7, upper=1.0),
-                             doc="Tolerance for comparison of double values.")
+        self.declareProperty("Tolerance", 1e-3, validator=FloatBoundedValidator(lower=1e-7, upper=100.0),
```
AndreiSavici: Why 100?

mganeva: TOFTOF chopper speeds can differ around 10. 100 is just to give some freedom. May be better would be to remove the upper limit, but I don't know any property which could have reasonable tolerance more than 100.

AndreiSavici: Load the CNCS_7860 from doc tests. Run `FilterBadPulses` to eliminate al the  values when the accelerator was down. The `proton_charge` log is in picoCoulombs, with an average of 1.81E7, and a standard deviation of about 82000. It all depends on the units. I would just leave a lower bound

AndreiSavici:
What would happen if one of the properties is NaN?
```
import numpy as np
ws1.mutableRun().addProperty('chopper_speed',np.nan,True)
```

mganeva:
@AndreiSavici Good question. I will check it on Monday. An remove the upper bound too.

mganeva:
OSX build has failed due to the "SliceViewerMantidPlotTest_SliceViewerPythonInterfaceTest" and the windows build has failed due to the "AlgorithmsTest_PDCalibrationTest". May I ask someone to restart these bulds?

PR #17407, opened 2016-09-05
MultipleFileProperty to support OptionalLoad action

### LINKED ISSUE #17358: MultipleFileProperty to support optional action
`MultipleFileProperty` needs to support `OptionalLoad` action similar to `FileProperty`.

gvardany:
`MultipleFileProperty` does not currently support any actions of `FileProperty`, just defaulting to mandatory `Load` behaviour. This PR implements support for `OptionalLoad` action coherent to `FileProperty`.

**To test:**
1. Pick your favourite algorithm that has `MultipleFileProperty` (e.g. `Load`).
2. Make it optional by adding `action=FileAction.OptionalLoad` (python) or `action=FileProperty::FileAction::OptionalLoad` (C++)\* as a second argument to the `MultipleFileProperty` constructor in property declaration.
3. Make sure that it became optional (e.g. the empty input will now be accepted, while before it would not be)
4. Try what happens with unsupported actions, e.g. `Save`.

*Make sure simpleapi or `FileProperty.h` are imported/included.

Fixes #17358 .

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/API/inc/MantidAPI/MultipleFileProperty.h
+++ b/Framework/API/inc/MantidAPI/MultipleFileProperty.h
@@ -139,6 +139,8 @@ public:
   std::string setValue(const std::string &propValue) override;
   std::string value() const override;
   std::string getDefault() const override;
+  /// Checks if this property is optional
+  bool isOptional() const;
 
   /// @return the vector of suggested extensions. For use in GUIs showing files.
   std::vector<std::string> getExts() const { return m_exts; }
@@ -151,6 +153,8 @@ public:
   operator=;
 
 private:
+  /// Returns a string depending on whether an empty value is valid
+  std::string isEmptyValueValid() const;
   std::string setValueAsSingleFile(const std::string &propValue);
   std::string setValueAsMultipleFiles(const std::string &propValue);
   /// Whether or not the user has turned on multifile loading.
@@ -163,6 +167,8 @@ private:
   /// The default file extension associated with the type of file this property
   /// will handle
   std::string m_defaultExt;
+  /// The action type of this property, Load (dafault) or OptionalLoad are supported
+  unsigned int m_action;
 };
 
 } // namespace API
--- a/Framework/API/inc/MantidAPI/MultipleFileProperty.h
+++ b/Framework/API/inc/MantidAPI/MultipleFileProperty.h
@@ -119,7 +119,12 @@ namespace API {
 class DLLExport MultipleFileProperty
     : public Kernel::PropertyWithValue<std::vector<std::vector<std::string>>> {
 public:
-  /// Constructor
+  /// Default constructor
+  MultipleFileProperty(
+      const std::string &name, unsigned int action,
+      const std::vector<std::string> &exts = std::vector<std::string>());
+
+  /// Alternative constructor with default action
   MultipleFileProperty(
       const std::string &name,
       const std::vector<std::string> &exts = std::vector<std::string>());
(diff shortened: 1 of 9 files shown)


samueljackson92:
Tested this with ```Load``` in C++ and ```IndirectILLReduction``` in Python. Changes work as expected except that ```Load``` still showed the invalid asterisk, despite being an optional property. The Python version worked as expected.

gvardany:
I had a look into this, seems `Load` uses a dedicated dialog, while for other algorithms the dialog is automatically generated, and in this case they behave correctly. Can you try with some other C++ algorithm, like `LoadFITS` ?

samueljackson92:
@gvardany sorry for the slow response on this. I've been away the last few days. I shall try and take a look today.

samueljackson92:
@gvardany This works fine with LoadFITS. :shipit:

Framework/API/inc/MantidAPI/MultipleFileProperty.h
```diff
@@ -119,7 +119,12 @@ namespace API {
 class DLLExport MultipleFileProperty
     : public Kernel::PropertyWithValue<std::vector<std::vector<std::string>>> {
 public:
-  /// Constructor
+  /// Default constructor
```
SimonHeybrock: This is not a default constructor? Also, Doxygen for member methods should be in the source, not in the header (actually you have them in the source, so just remove those from the header).

The same goes for other methods you added to the header file below.

Framework/API/src/MultipleFileProperty.cpp
```diff
@@ -37,17 +37,29 @@ bool doesNotContainWildCard(const std::string &ext) {
...
                                            const std::vector<std::string> &exts)
     : PropertyWithValue<std::vector<std::vector<std::string>>>(
           name, std::vector<std::vector<std::string>>(),
-          boost::make_shared<MultiFileValidator>(exts), Direction::Input),
-      m_multiFileLoadingEnabled(), m_exts(), m_parser(), m_defaultExt("") {
+          boost::make_shared<MultiFileValidator>(
+              exts, (action == FileProperty::Load)),
+          Direction::Input),
+      m_multiFileLoadingEnabled(), m_exts(), m_parser(), m_defaultExt(""),
+      m_action() {
```
SimonHeybrock: No need to explicitly call default constructors (or empty string constructor), can remove those.

Framework/API/test/MultipleFilePropertyTest.h
```diff
@@ -616,6 +618,12 @@ class MultipleFilePropertyTest : public CxxTest::TestSuite {
     TS_ASSERT_EQUALS(fileNames[0].size(), 1);
   }
 
+  void test_multiFileOptionalLoad() {
+    MultipleFileProperty p("Filename", FileProperty::OptionalLoad);
+    p.setValue("myJunkFile.nxs");
+    TS_ASSERT(p.isValid().empty());
+  }
```
SimonHeybrock: Does it make sense to add a test for the case of no file?

Framework/API/inc/MantidAPI/MultipleFileProperty.h
gvardany: Done.

Framework/API/src/MultipleFileProperty.cpp
gvardany: Done.

Framework/API/test/MultipleFilePropertyTest.h
gvardany: Added.

peterfpeterson:
@samueljackson92 is this still ready to be merged into master?

samueljackson92:
@peterfpeterson Looks as though @SimonHeybrock's comments have been addressed. This needs the merge conflict to be fixed, but otherwise I'd say :shipit:

PR #17415, opened 2016-09-06
Change psi to 0 if found to be NaN in LoadNXSPE 

### LINKED ISSUE #17409: Loading Powder diffraction sample from NXPSE with psi =nan will result in invalid geniometer
If in an NXPSE file `psi=nan`
this will result in The `goniometer` rotation matrix to become a 3 x 3 matrix of `NaN` which cause algorithms (such as `ConvertToMD`) to fail

EltayebAhmed:
Powder diffraction stored in .nxspe files may have an entry of `psi` with valuen `NaN`. This would cause running `ConvertToMD` with the property `QDimensions` set to `|Q|` on the workspaces to fail. I have added a check to LoadNXSPE to force `psi` to `0` (the defualt value) if it is found to be equal to `NaN` in the file

**To test:**
1. Load a powder diffraction sample such as that at `\\olympic\Babylon5\Public\EltayebAhmed\MAR21335_Ei60.00meV.nxspe`
2. Convert to an MD workspace in |Q| vs `E` by invoking
`md_workspace = ConvertToMD(InputWorkspace="MAR21335_Ei60.00meV", QDimensions="|Q|")`
3. Verify that a new workspace called `md_workspace` appeared
4. Verify that workspace is not empty by executing in the Script interpreter 
`md_workspace.getNPoints()` and verifying it is not `0`
Fixes #17409 

Release notes updated
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/DataHandling/src/LoadNXSPE.cpp
+++ b/Framework/DataHandling/src/LoadNXSPE.cpp
@@ -134,7 +134,7 @@ void LoadNXSPE::exec() {
     file.getData(temporary);
     psi = temporary.at(0);
     file.closeData();
-    if (!isfinite(psi)) {
+    if (!std::isfinite(psi)) {
       psi = 0;
       g_log.warning() << "Value of psi found in file was NaN, changing to 0"
                       << "\n";
--- a/Framework/DataHandling/src/LoadNXSPE.cpp
+++ b/Framework/DataHandling/src/LoadNXSPE.cpp
@@ -134,7 +134,7 @@ void LoadNXSPE::exec() {
     file.getData(temporary);
     psi = temporary.at(0);
     file.closeData();
-    if (!std::isfinite(psi)) {
+    if (!boost::math::isfinite(psi)) {
       psi = 0;
       g_log.warning() << "Value of psi found in file was NaN, changing to 0"
                       << "\n";
--- a/Framework/DataHandling/src/LoadNXSPE.cpp
+++ b/Framework/DataHandling/src/LoadNXSPE.cpp
@@ -17,7 +17,6 @@
 #include "MantidGeometry/Surfaces/Sphere.h"
 
 #include <boost/regex.hpp>
-#include <boost/math/special_functions/fpclassify.hpp>
 
 #include <map>
 #include <sstream>
--- a/Framework/DataHandling/src/LoadNXSPE.cpp
+++ b/Framework/DataHandling/src/LoadNXSPE.cpp
@@ -135,6 +135,10 @@ void LoadNXSPE::exec() {
     file.getData(temporary);
     psi = temporary.at(0);
     file.closeData();
+	if (!isfinite(psi)) {
+		psi = 0;
+		g_log.warning() << "Value of psi found in file was NaN, changing to 0" << "\n";
+	}
   }
 
   int kikfscaling = 0;
--- a/Framework/DataHandling/src/LoadNXSPE.cpp
+++ b/Framework/DataHandling/src/LoadNXSPE.cpp
@@ -135,10 +135,11 @@ void LoadNXSPE::exec() {
     file.getData(temporary);
     psi = temporary.at(0);
     file.closeData();
-	if (!isfinite(psi)) {
-		psi = 0;
-		g_log.warning() << "Value of psi found in file was NaN, changing to 0" << "\n";
-	}
+    if (!isfinite(psi)) {
+      psi = 0;
+      g_log.warning() << "Value of psi found in file was NaN, changing to 0"
+                      << "\n";
+    }
   }
 
   int kikfscaling = 0;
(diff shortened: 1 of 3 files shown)


Framework/DataHandling/src/LoadNXSPE.cpp
```diff
@@ -135,6 +135,11 @@ void LoadNXSPE::exec() {
     file.getData(temporary);
     psi = temporary.at(0);
     file.closeData();
+    if (!isfinite(psi)) {
```
AndreiSavici: You should either use `std::isfinite` or `boost::math::isfinite` like elsewhere in the file. I would change both to `std::isfinite`

EltayebAhmed:
The changes have now passed the unit tests

EltayebAhmed:
@abuts  This is pull request i mentioned

EltayebAhmed:
@abuts I am not sure if that is possible. The only constructor on the `Run` attribute is the default constructor. That constructor will call the default constructor on `Goniometer` which will initialize it to an identity matrix.

abuts:
There is run method 
run->getGoniometer()->isDefined() which should return false if psi was NaN

abuts:
@mducle -- is interested in this too

Framework/Geometry/src/Instrument/Goniometer.cpp
```diff
@@ -75,7 +75,13 @@ const Kernel::DblMatrix &Goniometer::getR() const { return R; }
...
 
 /// Function reports if the goniometer is defined
-bool Goniometer::isDefined() const { return initFromR || (!motors.empty()); }
+bool Goniometer::isDefined() const {
+	bool r_contains_nan = false;
+	std::vector<double> vec = R.getVector();
+	for (std::vector<double>::iterator it = vec.begin(); it != vec.end(); ++it)
+		r_contains_nan = r_contains_nan || (!std::isfinite(*it));
+	return (!r_contains_nan) && ( initFromR || (!motors.empty())); 
+}
```
EltayebAhmed: @mducle @abuts  @AndreiSavici 
Doing this fixed the problem. Can you see this having any undesirable side effects?

Framework/Geometry/src/Instrument/Goniometer.cpp
```diff
@@ -75,7 +75,13 @@ const Kernel::DblMatrix &Goniometer::getR() const { return R; }
...
 
 /// Function reports if the goniometer is defined
-bool Goniometer::isDefined() const { return initFromR || (!motors.empty()); }
+bool Goniometer::isDefined() const {
+  bool r_contains_nan = false;
+  std::vector<double> vec = R.getVector();
+  for (std::vector<double>::iterator it = vec.begin(); it != vec.end(); ++it)
+    r_contains_nan = r_contains_nan || (!std::isfinite(*it));
+  return (!r_contains_nan) && (initFromR || (!motors.empty()));
+}
```
EltayebAhmed: @mducle @abuts @AndreiSavici
Doing this fixed the problem. Can you see this having any undesirable side effects?

EltayebAhmed:
Although there is the issue of if the goniometer contains a `NaN` it will be "isDefined" will get stuck to False

EltayebAhmed:
So pushing a `NaN` to a goniometer via `pushAxis` would make the goniometer remain in an undefined state and it would still remain undefined even if you pushed more motor axis

mducle:
I think the change to `isDefined` is good.

abuts:
This solution looks fine to me. The code which sets 0 at loadNXSPE should be removed.

Framework/Geometry/src/Instrument/Goniometer.cpp
```diff
@@ -128,6 +124,15 @@ void Goniometer::pushAxis(std::string name, double axisx, double axisy,
     throw std::runtime_error(
         "Initialized from a rotation matrix, so no axes can be pushed.");
   } else {
+    if (!boost::math::isfinite(axisx) || !boost::math::isfinite(axisy) ||
+        !boost::math::isfinite(axisz) || !boost::math::isfinite(angle)) {
```
mducle: To get this to compile on Linux, it seems you either have to add `#include <boost/math/special_functions/fpclassify.hpp>` or change `boost::math::isfinite` to `std:isfinite` - I think the later is preferable, since we don't really need boost for this because the variables here are just `double`.

Framework/Geometry/src/Instrument/Goniometer.cpp
```diff
@@ -17,6 +17,8 @@ using Kernel::DblMatrix;
 using Kernel::V3D;
 using Kernel::Quat;
 
+Mantid::Kernel::Logger g_log("Goniometer");
```
mducle: To get it to compile on Linux, you need to add `#include "MantidKernel/Logger.h"` at the start of the file.

AndreiSavici:
Works as described. Loaded an nxspe file, with psi=Nan (checked in the sample logs). ConvertToMD gives sensible results for powder and warns about lack of goniometer for Q3D.
:shipit:

martyngigg (approved):

PR #17428, opened 2016-09-08
MergeRuns to handle merging sample logs

### LINKED ISSUE #16928: Extend MergeRuns to allow merging of sample logs
Currently MergeRuns just keeps the metadata from the first file in any merge. This is a problem at the ILL where some metadata should be averaged, summed, taken as min/max etc.

This issue should address this by extending MergeRuns to treat the metadata as expected. For any particular sample log the operation when merging needs to be defined. The instrument parameters file can be extended to include how to treat required sample logs.

If any sample logs that are given in the instrument parameters file are missing they should be skipped.

This is very similar behaviour to that in [TOFTOFMergeRuns](http://docs.mantidproject.org/nightly/algorithms/TOFTOFMergeRuns-v1.html). This ticket will provide an alternative to TOFTOFMergeRuns that should produce the same results.

TOFTOFMergeRuns also checks some parameters match in the files it is merging. These should also be configurable in the instrument parameter file.

ianbush:
This PR adds options to MergeRuns to deal with the sample logs. Behaviour can be to merge as a time series (numeric values), merge as a list, warn if values are different and fail if values are different.

How to deal with each value is defined in either the instrument parameters file or as arguments to MergeRuns.

**To test:**

Functional:

Either pick your favourite NeXus files or grab some ILL IN6 files from here -  https://www.ill.eu/fileadmin/users_files/img/instruments_and_support/support_facilities/computing_for_science/Computing_for_Science/Data_analysis/IN6_Script_Data.zip

Pick some parameters and try adding them to the instrument parameters file (see updated MergeRuns-v1.rst for example XML for IN6 files).

Do the same with setting the options via the MergeRuns algorithm.

Check documentation updates make sense.

Code:

Check unit test pass. Any obvious unit tests missing for the new behaviour?  Would be useful to review the code too, but there is quite a lot there.

Fixes #16928.

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/Algorithms/src/MergeRuns.cpp
```diff
@@ -120,7 +165,16 @@ void MergeRuns::exec() {
...
       // Add the current workspace to the total
-      outWS = outWS + addee;
+      // Update the sample logs
+      try {
+        sampleLogsBehaviour.mergeSampleLogs(*it, outWS);
+        outWS = outWS + addee;
+        sampleLogsBehaviour.setUpdatedSampleLogs(outWS);
+      } catch (std::invalid_argument &e) {
+        g_log.error() << "Could not merge run: " << it->get()->name()
+                      << ". Reason: \"" << e.what() << "\".";
```
SimonHeybrock: Maybe extend this message to mention that it is simply skipped, i.e., the algorithm continues and does not just abort?

Framework/Algorithms/inc/MantidAlgorithms/MergeRuns/SampleLogsBehaviour.h
```diff
@@ -0,0 +1,114 @@
...
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+class MANTID_ALGORITHMS_DLL SampleLogsBehaviour {
+public:
+  enum MergeLogType { time_series, list, warn, fail };
```
SimonHeybrock: I think more standard formatting should be used, i.e., `TimeSeries`, and `List`.
Also consider using a strongly-typed enum (`enum class MergeLogType`).

Framework/Algorithms/src/MergeRuns/SampleLogsBehaviour.cpp
```diff
@@ -0,0 +1,463 @@
...
+      try {
+        // See if property exists already - merging an output of MergeRuns
+        prop =
+            std::shared_ptr<Property>(ws->getLog(item + LIST_SUFFIX)->clone());
+      } catch (std::invalid_argument &) {
+        ws->mutableRun().addProperty(item + LIST_SUFFIX, prop->value());
+        prop =
+            std::shared_ptr<Property>(ws->getLog(item + LIST_SUFFIX)->clone());
+      }
+    }
```
SimonHeybrock: No else?

Framework/Algorithms/src/MergeRuns/SampleLogsBehaviour.cpp
```diff
@@ -0,0 +1,463 @@
...
+
+    map[item] = {mergeType, prop, tolerance, isNumeric};
+  }
```
SimonHeybrock: This `for` loop is really massive and quite hard to understand. Is there a way to restructure this a bit and move some code to helper functions?

Framework/Algorithms/src/MergeRuns/SampleLogsBehaviour.cpp
```diff
@@ -0,0 +1,463 @@
...
+ * tolerances are set to that value.
+ *
+ * @param numberNames the number of sample log names
+ * @param tolerances a vector containing strings with the tolerances
+ * @param paramsTolerances a string containing nothing, or tolerances to use
+ * @return a vector of doubles of size numberNames
+ */
+std::vector<double> SampleLogsBehaviour::createTolerancesVector(
+    size_t numberNames, const std::vector<std::string> &tolerances,
+    const std::string paramsTolerances) {
```
SimonHeybrock: This is only used to check if it is empty? Is this correct? It seems weird.

Framework/Algorithms/src/MergeRuns/SampleLogsBehaviour.cpp
```diff
@@ -0,0 +1,463 @@
...
+          name, addeeWS->name(), addeeWSProperty->value(),
+          behaviour.property->value());
+    }
+  } else {
+    if (behaviour.property->value().compare(addeeWSProperty->value()) != 0) {
+      m_logger.warning() << generateDifferenceMessage(
+          name, addeeWS->name(), addeeWSProperty->value(),
+          behaviour.property->value());
+    }
+  }
```
SimonHeybrock: Maybe slightly reorder things such that you do not have to write the call to the logging things twice?

Framework/Algorithms/src/MergeRuns/SampleLogsBehaviour.cpp
```diff
@@ -0,0 +1,463 @@
...
+ *
+ * @param addeeWS the workspace being merged
+ * @param addeeWSProperty the property value of the workspace being merged
+ * @param behaviour the merge behaviour struct, containing information about the
+ *merge
+ * @param addeeWSNumber a double for the addeeWS value (0 if it is not numeric)
+ * @param outWSNumber a double for the outWS value (0 if it is not numeric)
+ * @param name the name of the sample log to check
+ */
+void SampleLogsBehaviour::checkErrorProperty(
```
SimonHeybrock: Similar to `checkWarnProperty`, could probably be written slightly nicer. Maybe the tolerance check (which is the same in both cases) can be moved to a separate method `checkTolerance` or something like that?

Framework/Algorithms/src/MergeRuns/SampleLogsBehaviour.cpp
```diff
@@ -0,0 +1,463 @@
...
+
+    if (item.second.type == time_series) {
+      propertyToReset = item.first + TIME_SERIES_SUFFIX;
+    } else if (item.second.type == list) {
+      propertyToReset = item.first + LIST_SUFFIX;
+    } else {
+      return;
+    }
+
+    Property *outWSProperty = ws->mutableRun().getProperty(propertyToReset);
```
SimonHeybrock: `const`?

Framework/Algorithms/src/MergeRuns/SampleLogsBehaviour.cpp
```diff
@@ -0,0 +1,463 @@
...
+    const API::MatrixWorkspace_sptr &ws) {
+  for (auto item : m_logMap) {
+    std::string propertyToReset = item.first;
+
+    if (item.second.type == time_series) {
+      propertyToReset = item.first + TIME_SERIES_SUFFIX;
+    } else if (item.second.type == list) {
+      propertyToReset = item.first + LIST_SUFFIX;
+    } else {
+      return;
```
SimonHeybrock: Really `return`? Not `continue`?

Framework/Algorithms/src/MergeRuns/SampleLogsBehaviour.cpp
```diff
@@ -0,0 +1,463 @@
...
+      auto property =
+          std::unique_ptr<Kernel::Property>(item.second.property->clone());
+      ws->mutableRun().addProperty(std::move(property), true);
+    } else if (item.second.type == list) {
+      propertyToReset = item.first + LIST_SUFFIX;
+      ws->mutableRun()
+          .getProperty(propertyToReset)
+          ->setValue(item.second.property->value());
+    } else {
+      return;
```
SimonHeybrock: Really `return`?

Framework/Algorithms/src/MergeRuns/SampleLogsBehaviour.cpp
```diff
@@ -0,0 +1,463 @@
...
+}
+
+/**
+ * Set the values in the map to be the same as those in the output workspace.
+ *
+ * @param ws the merged workspace
+ */
+void SampleLogsBehaviour::setUpdatedSampleLogs(
+    const API::MatrixWorkspace_sptr &ws) {
+  for (auto item : m_logMap) {
```
SimonHeybrock: Should this be a reference? `item` is modified below, but the value is a `shared_ptr`, so it is ok... but it is confusing and inefficient.

Framework/Algorithms/src/MergeRuns/SampleLogsBehaviour.cpp
```diff
@@ -0,0 +1,463 @@
...
+  }
+}
+
+/**
+ * Resets the sample logs in the workspace to the values in the map.
+ *
+ * @param ws the merged workspace to reset the sample logs for
+ */
+void SampleLogsBehaviour::resetSampleLogs(const API::MatrixWorkspace_sptr &ws) {
+  for (auto item : m_logMap) {
```
SimonHeybrock: `const &`?

Framework/Algorithms/test/MergeRunsTest.h
```diff
@@ -132,6 +140,90 @@ class MergeRunsTest : public CxxTest::TestSuite {
     return group;
   }
 
+  template <typename T>
+  WorkspaceGroup_sptr create_group_workspace_with_sample_logs(
+      std::string merge_type, std::string merge_list, T value_1, T value_2,
+      T value_3, T value_4, std::string tolerances = "") {
```
SimonHeybrock: Pass all parameters by `const &`?

Framework/Algorithms/test/MergeRunsTest.h
```diff
@@ -132,6 +140,90 @@ class MergeRunsTest : public CxxTest::TestSuite {
...
+    AnalysisDataService::Instance().addOrReplace("b1", b);
+    AnalysisDataService::Instance().addOrReplace("group1", group);
+    return group;
+  }
+
+  template <typename T>
+  MatrixWorkspace_sptr
+  create_workspace_with_sample_logs(std::string merge_type,
+                                    std::string merge_list, T value_1,
+                                    T value_2, std::string tolerances = "") {
```
SimonHeybrock: `const &`?

Framework/Algorithms/test/MergeRunsTest.h
```diff
@@ -847,6 +939,388 @@ class MergeRunsTest : public CxxTest::TestSuite {
     TS_ASSERT(alg.useCustomInputPropertyName());
   }
 
+  void do_test_mergeSampleLogs(WorkspaceGroup_sptr input,
+                               std::string propertyName, std::string mergeType,
```
SimonHeybrock: `const &`?

Framework/Algorithms/test/MergeRunsTest.h
```diff
@@ -847,6 +939,388 @@ class MergeRunsTest : public CxxTest::TestSuite {
...
+    alg.initialize();
+
+    do_test_mergeSampleLogs_modified_alg(alg, input, propertyName, mergeType,
+                                         result, filesMerged, noOutput);
+  }
+
+  void do_test_mergeSampleLogs_modified_alg(MergeRuns &alg,
+                                            WorkspaceGroup_sptr input,
+                                            std::string propertyName,
+                                            std::string mergeType,
```
SimonHeybrock: `const &`?

docs/source/algorithms/MergeRuns-v1.rst
```diff
@@ -60,13 +60,47 @@ If group workspaces are provided that are not multi-period, this
 algorithm will merge across all nested workspaces, to give a singe
 output matrix workspace.
 
+Merging Sample Logs
+###################
+
+Sample logs are optionally merged when running this algorithm. The behaviour
+when merging is defined in the instrument parameter file, but can be added to
+or overridden or via this algorithm. Definitions in the XML file are given as
```
SimonHeybrock: One `or` too many?

docs/source/release/v3.8.0/framework.rst
```diff
@@ -98,6 +98,10 @@ Improved
 
 - :ref:`LoadEmptyInstrument <algm-LoadEmptyInstrument>` now also accepts a workspace name as input, as an alternative to an instrument definition xml file.
 
+- :ref:`LoadEmptyInstrument <algm-LoadEmptyInstrument>` now also accepts a workspace name as input, as an alternative to an instrument definition xml file.
```
SimonHeybrock: This line was accidentally copied, I think.

Framework/Algorithms/inc/MantidAlgorithms/MergeRuns/SampleLogsBehaviour.h
```diff
@@ -0,0 +1,114 @@
...
+  static const std::string LIST_SUFFIX;
+
+  typedef struct {
+    MergeLogType type;
+    std::shared_ptr<Kernel::Property> property;
+    double tolerance;
+    bool isNumeric;
+  } SampleLogBehaviour;
+
+  SampleLogsBehaviour(const API::MatrixWorkspace_sptr &ws,
```
SimonHeybrock: As far as I have seen below, there is no need for having a `shared_ptr` here or in any of the methods of this class. `const MatrixWorkspace &` and `MatrixWorkspace &` should do the job and (marginally) reduce the overhead and code noise.

SimonHeybrock (commented):
I have reviewed the code, and all in all it looks really good (apart from minor comments). Well structured and many tests cases :+1:.

Will now try to run a manual test...

SimonHeybrock:
(Why did github put my "final" comment on the review in the middle of all the other line-comments??)

SimonHeybrock:
@ianbush I just read your comments regarding `TOFTOFMergeRuns` in the issue description. Would it now be possible to use `MergeRuns` internally in `TOFTOFMergeRuns`?

SimonHeybrock:
@ianbush I just tried setting the "Warn" and "Fail" option for the *same* parameter. I get "Error when making list of merge items, sample log "sample.regulation_temperature" defined more than once!". That is maybe to e expected, but wouldn't it make sense to be able to set a warning threshold *and* a *different* fail threshold?

SimonHeybrock:
@ianbush Regarding documentation: Another thing I notice is that the tolerances are with respect to the first workspace that is *selected* in the GUI. Would it make sense to add a comment somewhere, that it is not the order in the workspace list, or something, but really the order of selection?

SimonHeybrock:
@ianbush When using the "list" or "time_series" option: The result is a new log entry, e.g., "myvalue_list", but "myvalue" is also still present, with the value of the first workspace that was selected.

Is this the intended behavior? Would it make sense to remove it? Or average? If not, maybe just make that explicit in the documentation, i.e., that the values of the first workspace will show up?

Framework/Algorithms/src/MergeRuns/SampleLogsBehaviour.cpp
```diff
@@ -396,19 +458,46 @@ void SampleLogsBehaviour::checkErrorProperty(
     const MatrixWorkspace_sptr &addeeWS, Property *addeeWSProperty,
     const SampleLogBehaviour &behaviour, const double addeeWSNumber,
     const double outWSNumber, const std::string name) {
+
+  if (!isWithinTolerance(behaviour, addeeWSNumber, outWSNumber) &&
+      !stringPropertiesMatch(behaviour, addeeWSProperty)) {
```
SimonHeybrock: This is now slightly different from before? Previously the string property check was not taken into account in the "numeric" code branch?

ianbush: Yes, it is actually a bit different, although it should not change anything. It might be a little safer as if `isWithingTolerance` returns false, but the strings match, the warning or error will not be generated, though that should never happen.

ianbush:
Thanks for the review and comments, they are very useful. I am just going through and making some of the code changes at the moment.

The algorithm should be close to being able to replace TOFTOFMergeRuns. I think the two things that might be different are:
* Errors cause TOFTOFMergeRuns to abort, whereas here the algorithm continues
* If bins are different MergeRuns rebins, whereas TOFTOFMergeRuns aborts. Perhaps this should be added as a checkbox option to MergeRuns?

I agree that it makes sense to allow the sample logs to be used in more than one type of check, I will have a look at changing that. Originally there were separate options that overwrote the original parameters, and it made more sense to keep it unique, but now the sample logs for time series and list have different names there is no longer a need for that restriction.

The behaviour to keep the original sample log was intentional, to avoid altering existing behaviour, for example where someone wants to get a run number from the first workspace entry.

SimonHeybrock:
@ianbush Updates look good so far, let me know when this is done (more doc-updates to come?). I will be offline for a few hours, but will have a look tonight so we can get it into the release.

Framework/Algorithms/src/MergeRuns/SampleLogsBehaviour.cpp
ianbush: No, warn and fail do not require adding a new sample log, I have added a comment to this effect.

Framework/Algorithms/src/MergeRuns/SampleLogsBehaviour.cpp
ianbush: Yes, it was a bit of a monster! I have pulled out some of the methods to helper methods and added some comments, so it is hopefully clearer now.

Framework/Algorithms/src/MergeRuns/SampleLogsBehaviour.cpp
ianbush: Yes, that is weird, especially when checking the vector is empty gives the same result! Fixed.

ianbush:
This should be ready for another look now. I have put a thumbs up next to everything that was fixed. I added a comment about the selection in the rst docs - I did not know that was how it worked!

The one outstanding thing, which I will create a separate issue for, is to allow skipping files when bins are different.

SimonHeybrock:
@ianbush Great! Looks good, just Doxygen is failing with the latest changes...

SimonHeybrock:
:shipit: once Jenkins gives his ok.

Framework/API/src/Run.cpp
```diff
@@ -42,6 +42,16 @@ Kernel::Logger g_log("Run");
 // Public member functions
 //----------------------------------------------------------------------
 
+boost::shared_ptr<Run> Run::clone() {
```
FedeMPouzols: Could this method be `const`? Just in case it is not intentional/constrained by something else.

PR #17434, opened 2016-09-08
Tiled plot option for "Plot spectrum"

### LINKED ISSUE #9239: Make it easier to create tiled plots
Original Reporter: Arturs Bekasovs

This looks like a long-requested feature, but today Muon scientists in particular mentioned it once again.

It is currently possible to create plots with multiple layers, however this is nowhere near to trivial. You need to create a graph, then add layer to it and add curve to the new layer. Tiling 4 curves like that requires a decent amount of mouse clicks.

On the other hand, it's quite easy to plot multiple curves on the same layer - you just plot a list of spectra. Although it seems like a reasonable default, it makes sense to add an option to create new layers instead of adding curves to the same one.

Possible solution is to add another option to plotSpectrum Python func and a tick-box to "Plot spectra..." menu, where you select which spectra you want to plot (appearing when you drag&drop as well). Additionally, it might be a good idea to provide an alternative in a few other places, e.g. when you select certain spectra/bins in data view, you might have "Plot spectra/bin tiled" in addition to existing "Plot spectra/bin".

There should be some pre-cautions in terms of the number of tiles we can display at the same time.

tom-perkins:
Enable the easy creation of a MultiLayer plot from right-clicking on a workspace.

At present, it is possible to create such a plot, but you have to plot each spectrum individually and then drag them onto a tiled plot, or write a Python script. This PR provides a way to do it from right-clicking a workspace in the GUI.

When used on a WorkspaceGroup, each workspace gets its own subplot, with the specified spectra in each. When used on a single workspace, each of the specified spectra is plotted in its own subplot.

**Example use case:**
After fitting several workspaces at once, the results are contained in a WorkspaceGroup. There is currently no easy way to visualise the results - you have to plot each fit individually in its own window, or have them all on the same axes (messy). With the new option, you can create a multiple plot of {data, fit, difference} for each workspace in its own subplot.

**To test:**
- Create a GroupWorkspace containing several multi-spectra MatrixWorkspaces.
- Right-click on the GroupWorkspace and choose "Plot Spectrum..."
- In the dialog, check the "Tiled Plot" option. Select a range of spectra to plot (or choose to plot all).
- (If there are a large number of workspaces in the group, it should warn you about this).
- Should get a plot of each workspace in its own subplot:
  - Axis titles should only be present on the leftmost column and bottom row (or second-to-last row if there is no plot underneath).
  - Axis scales should be the same for all subplots, and the scale should be large enough to show all the data in each plot. 
  - Font size of axis titles, labels and legends should be the same for all subplots.
  - There should not be excessive white space between the subplots.
  - Each subplot should have a legend to show which workspace it came from.
- Now right-click on one of the MatrixWorkspaces from the group, and do a tiled plot of this.
  - Each spectrum (from those selected) should be shown in its own subplot.

Fixes #9239.

[Release notes](https://github.com/mantidproject/mantid/blob/39583c9aa21c316b36e05de9d2ffbaffa8eac995/docs/source/release/v3.8.0/ui.rst#plotting-improvements) have been updated.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidPlot/src/Graph.cpp
+++ b/MantidPlot/src/Graph.cpp
@@ -2834,9 +2834,9 @@ PlotCurve *Graph::insertCurve(Table *w, const QString &xColName,
 }
 
 PlotCurve *Graph::insertCurve(QString workspaceName, int index, bool err,
-                              Graph::CurveType style) {
+                              Graph::CurveType style, bool distribution) {
   return (new MantidMatrixCurve(workspaceName, this, index,
-                                MantidMatrixCurve::Spectrum, err, false,
+                                MantidMatrixCurve::Spectrum, err, distribution,
                                 style));
 }
 
--- a/MantidPlot/src/Graph.h
+++ b/MantidPlot/src/Graph.h
@@ -290,7 +290,8 @@ public slots:
                          const QString &yColName, int style, int startRow = 0,
                          int endRow = -1);
   PlotCurve *insertCurve(QString workspaceName, int index, bool err = false,
-                         Graph::CurveType style = Graph::Unspecified);
+                         Graph::CurveType style = Graph::Unspecified,
+                         bool distribution = false);
   PlotCurve *insertCurve(PlotCurve *c, int lineWidth = -1,
                          int curveType = User);
   void insertPlotItem(QwtPlotItem *i, int type);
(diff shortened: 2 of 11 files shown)


raquelalvarezbanos:
Tested on a `WorkspaceGroup` of `OFFSPEC` data:

tiledplotgroup

and on a single `MatrixWorkspace` of `ARGUS` data:

tiledworkspace

Works exactly as described in the testing instructions. The existing plot options work as before. I only noticed that if both `Waterfall Plot` and `Tiled Plot` are selected at the same time, the latter will be considered for plotting. I guess this is fine, as one should either use `Waterfall` or `Tiled`.

raquelalvarezbanos:
:shipit:

PR #17448, opened 2016-09-09
Engg:: Write out calibration file twice to avoid race condition

### LINKED ISSUE #17296: Engg GUI: fails to copy output calibration files to "General" directory when the files are not written fast enough
### Expected behavior

Output calibration files should be produced in the `User/RB<number>/Calibration` output directory and also copied into the general `Calibration` directory.

### Actual behavior

Sometimes the files are not copied into the general directory. This was observed the other day with Saurabh. I could replicate this on debian. I get messages like:
```
Cannot copy the file /home/fedemp/EnginX_Mantid/User/RB0003fff/Calibration/ENGINX_236516_241391_bank_North.prm to the general/all users directories because it cannot be read.
```
But when I go and check the file is there!
You'll get this only the first time, if you repeat the process, because the files are already there, they will be copied without issues. 

The problem is I think that the files are produced in a Python script which runs in a separate thread, immediately after starting that script, we try to copy the file. We'll need to be more patient. This happens in `EnggDiffractionPresenter::doCalib` when they call `copyToGeneral()`.

Run a calibration with the GUI, for example using the data / run numbers given in #17277. Note this bug may be elusive, you need to be lucky to catch it.

### Platforms affected

Probably all.

DavidFair:
Description of work.
The engineering interface currently runs the calibration and saves out the files to both 
`User/<RB Number>/Calibration` and `Calibration` in the EnginX folder. 
Previously the implementation saved the file in `User` then copied that file to `Calibration` in the top level directory, however as the saving method is a Python script and the copy step was C++ there was no easy way to verify the Python script executed before the C++. This meant that especially on slower machines the file would not exist when we came to copy it so the file would only exist in one location.

To avoid this we now save the file twice in both locations - whilst this incurs some additional overhead the calibration procedure can take some time so a minor time increase will not be noticeable.

**To test:**
- Go to the EnginX folder - this will be located at either `C:\EnginX_Mantid` or `~/EnginX_Mantid` and delete and existing calibrations from the top level `Calibration` folder and `User/<RB Number>/Calibration`
- Copy the files from `\\olympic\Babylon5\Public\David Fairbrother\z_test17296` for testing
- Open the Engineering Diffraction interface
- Open the calibration tab
- Take note of the RB number you have entered at the top of the screen
- Enter the following:
  Vanadium run : `236516`
  Calibration Sample: `241391`
- Run the calibration
- On completion check that 3 `.prm` files are created in the `Calibration` and `User/<RB Number>/Calibration` folders.

Fixes #17296 .

---
#### Reviewer

##### Code Review
##### Functional Tests


FedeMPouzols:
Looks like a good solution to me. Eventually it would be nice if the GSAS-PRM file saving routine is integrated into the algorithm SaveGSS, but that requires a bit of harmonization.

PranavBahuguna:
@DavidFair On reviewing this PR, I have found the following:

Following the provided testing instructions does indeed produce 3 .prm files in the top-level `Calibration` and `User/<RB Number>/Calibration` folders.

I have found however that if the top-level `Calibration` folder is missing, it will not be created and no .prm files will be saved there. The reverse is not true for the `User/<RB Number>/Calibration` folder, in which that folder will be created and files stored in if it is missing. 

As I believe the algorithm is supposed to create a top-level `Calibration` folder if missing, this is an issue that should be rectified as well.

PranavBahuguna:
All looks good now :shipit:

FedeMPouzols:
This solves the I/O race issue by replacing a save+copy with two saves. It works well. I also tried starting from empty "user" and "general" directories, and empty "root / Enginx_Mantid" directory, and it managed to create the tree of subdirectories.
I'd also say :shipit:

PR #17456, opened 2016-09-09
Fix LR autoreduce logic

mdoucet:
This PR takes care of two minor issues. 

1- When the meta-data related to the measurement sequence is bad, there was an instance where the algorithm was trying to access a non-existent item in a list. That should be protected against.

2- When normalizing to 1 under the critical edge, there is the possibility of dividing by zero if there is no data under the cut-off we asked for. In this case, just skip the normalization.

**To test:**
- Review the code
- Make sure tests pass
- Run reduction for run REFL_144854, which should bring up both problems. You should get a warning about the sequence information, and the reduced file should have a meta-data entry stating that the scaling factor is 1.

There is no GitHub Issue for this PR. All the info is here.

---
#### Reviewer

##### Code Review
##### Functional Tests


ricleal:
Minimal changes. Works as expected. :shipit:

Issue #15703, opened 2016-03-18
Crystal field: add form factor correction


mantid-roman:
Some of the crystal field fitting functions should be able to perform the form factor correction.

This is a part of #15684

mducle:
@mantid-roman As discussed in the meeting, this issue will be better handled by an algorithm separate from the crystal field fitting. I've hacked together an algorithm which uses the tables coded into `MagneticIon`. I'll get together some tests for it, and then open a PR.

(There is also an associated issue, #17457 to extend `MagneticIon` to include 5d ions and the dipole formula for rare earths).

mantid-roman:
@mducle Could you also add a docs rst page at `docs/source/algorithms`?

mducle:
Yes. Will do... just trying to get the example to work... The dataset I want to use is for PrAl3 but the form factor of Pr3+ is not included in the tables (it's not in the international tables for crystallography for some reason - but it's in McPhase and also the tables on the ILL webpage https://www.ill.eu/sites/ccsl/ffacts/ffactnode7.html and I'm trying to hunt down where those numbers actually come from).

PR #17458, opened 2016-09-11
Extending MagneticIon to 5d ions and including the dipole formula for magnetic form factors

### LINKED ISSUE #17457: Add 5d ions to MagneticIon and update form factor formula for 4f/5f ions
The current `analyticalFormFactor` method of `MagneticIon` returns only the spin-only magnetic form factor, which includes only the zeroth order integrals \<j0(Q)\>. This is good enough for transition metal ions, where the orbital angular momentum is quenched (effectively zero), but for rare earth ions (4f or 5f outer shell electrons), they also have to be taken into account even in the usual dipole approximation. In this case, the 2nd order integrals \<j2(Q)\> have to be included, and the form factor is given by: _FF(Q)=\<j0\>+(1-2/g)\<j2\>_ where _g_ is the Lande g-factor (=2 for spin-only systems). The analytical coefficients of the \<j2\> (and indeed the \<j4\> and \<j6\>) integrals are already included in `MagneticIon` so it is only a matter for also including the Lande g-factors and modifying `analyticalFormFactor` to use this equation.

In addition, the current tables of coefficients (which are an analytical fit to ab initio calculations of the relevant integrals) do not include the 6th period transition metal ions (with 5d outer shell electrons). This is because these ions are not included in the Internation Tables for Crystallography, from which the table here are copied. However, recent work has been published tabulating the coefficients for the 5d elements, and there is some demand for their inclusion.

mducle:
This PR adds the tables of coefficients for 5d ions (Hf-Au) for the analytical approximation to the Hartree-Fock calculated radial integrals going into the magnetic form factor.

In addition, it also amends the method `analyticalFormFactor` to output the form factor F(Q) in the dipole approximation rather than just the spin-only forms. This will only affects rare earth ions (Ce-Yb, U-Am). Transition metal ions will be the same a before.

**To test:**

The changes are not exposed to Python. The easiest way to test from Python is to use the `MagFormFactorCorrection` algorithm from #15703 and specify a name for the `FormFactorWorkspace` property. The algorithm will then produce a point data workspace with the calculated form factor for the Q-range of the input workspace. You can then check that at Q=0, the form factor is 1 and that it falls off to zero by around 20 Angstrom^-1. The analytical approximation is not perfect and the analytical form factor may become negative at some points. 

Fixes #17457 .

Documentation to follow.

---
#### Reviewer

##### Code Review
##### Functional Tests


AndreiSavici:
Changes look reasonable. :shipit:

PR #17436, opened 2016-09-08
Removed Parallel statements from SANSSolidAngleCorrection

ricleal:
The `HFIRTestsAPIv2.HFIRTestsAPIv2` system test is failing after:
https://github.com/mantidproject/mantid/pull/17346

I have removed the Parallel statements as quick fix. For HFIR instruments, this is now 2 seconds slower. More time is needed to find the problem.

All tests should pass.

---
#### Reviewer

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):
##### Code Review
- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)? Is it well structured with small focussed classes/methods/functions?
- [ ] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?
##### Functional Tests
- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant documentation been added/updated?
- [ ] Is user-facing documentation written in a user-friendly manner?
- [ ] Has developer documentation been updated if required?
- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

PR #17346, opened 2016-08-25
Faster ParameterMap

quantumsteve:
Description of work.

This improves the performance of SCDCalibratePanels and other algorithms which access the ParameterMap from multiple threads. The `tbb::concurrent_unordered_multimap` supports concurrent insertion and traversal, while critical sections enclose all calls to `unsafe_erase(...)`. FindTBB.cmake was copied from VTK.

`boost::iequal`, which uses [locales and contains a mutex](http://stackoverflow.com/questions/1771386/explain-the-need-for-mutexes-in-locales-please) was replaced with `strcasecmp` on Linux and `_stricmp` on Windows.

The critical section in SCDCalibratePanels was eliminated by moving `push_back` outside of the parallel loop and accessing the vector by its index.

Fixed a bug in `ParameterMap::clearParametersByName` where not all names associated with a particular component were being checked/cleared. 

scdcalibratepanels

SHA=5135bc0
1 3m10.239s
2 2m50.114s
4 2m48.759s
8 2m48.095s
16 2m50.160s
32 2m58.167s
64 3m2.704s
(unspecified) 3m5.546s

mantidpythonnightly August 30, 2016
1 1m46.937s 1m46.341s
2 1m25.669s 1m23.106s
4 1m44.806s 1m43.665s
8 1m44.524s 1m42.377s
16 1m52.687s 1m48.501s
32 1m58.103s 1m56.550s
64 2m0.437s 2m0.435s
(unspecified) 1m58.707s 1m58.828s

this branch
1 1m16.127s 1m16.631s
2  59.282s 59.166s
4  57.371s 57.686s
8   50.838s 50.196s
16 48.046s 48.653s 
32 47.882s 46.000s
64 43.811s 43.856s
(unspecified) 44.636s 45.408s

**To test:**

Fixes #17308.

---
#### Reviewer

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):
##### Code Review
- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)? Is it well structured with small focussed classes/methods/functions?
- [ ] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?
##### Functional Tests
- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
[...]

PR #17460, opened 2016-09-12
Fix concurrency issues in Cache and ParameterMap

quantumsteve:
Description of work.

This attempts to fix the concurrency issues in `Cache` and `ParameterMap` classes which were causing the `HFIRTestsAPIv2.HFIRTestsAPIv2` system test to fail.  I used `boost::atomic_load` to protect copies in In `ParameterMap`. I used a mutex in `Cache` because `BoundingBox` contains a `std::vector` object and therefore is not trivially copyable.

**To test:**

Run `./systemtest -R HFIRTestsAPIv2.HFIRTestsAPIv2` and try getting it to fail.

This (hopefully) fixes a concurrency issue from #17346 that was temporarily addressed in #17436.

---
#### Reviewer

##### Code Review
##### Functional Tests


ricleal:
I launched the `./systemtest -R HFIRTestsAPIv2.HFIRTestsAPIv2` quite a few times and it never failed.
I guess it's working as expected. :shipit:

Framework/Kernel/inc/MantidKernel/Cache.h
```diff
@@ -74,7 +99,26 @@ template <class KEYTYPE, class VALUETYPE> class DLLExport Cache {
    * @param value The new value for the key
    */
   void setCache(const KEYTYPE &key, const VALUETYPE &value) {
-    m_cacheMap[key] = value;
+    //
+    auto where = m_cacheMap.find(key);
```
SimonHeybrock: Now thread A can run `find` while thread B modifies `m_cacheMap`. Are you sure that is possible with `tbb::concurrent_unordered_map`? It looks wrong to me.

Framework/Kernel/inc/MantidKernel/Cache.h
```diff
@@ -123,6 +167,7 @@ template <class KEYTYPE, class VALUETYPE> class DLLExport Cache {
     bool isValid = it_found != m_cacheMap.end();
 
     if (isValid) {
+      std::lock_guard<std::mutex> lock(m_mutex);
```
SimonHeybrock: What is this supposed to do? It does not seem to protect against modification of the map after the `find`?

Framework/Kernel/inc/MantidKernel/Cache.h
quantumsteve: The [documentation](https://software.intel.com/en-us/node/506171) claims concurrent access to `find` and `insert` is allowed, but not `unsafe_erase`.

Framework/Kernel/inc/MantidKernel/Cache.h
quantumsteve: My goal was to protect updates to the keys, but like you say above, this mutex alone is most likely not sufficient. 

Here's one [solution using atomics](http://stackoverflow.com/questions/15185856/tbb-concurrent-hash-maps-how-to-compare-and-set). Unfortunately, `BoundingBox` isn't trivially copyable, make this this solution more difficult to implement.

SimonHeybrock: After looking at the documentation you linked I think this might be fine even without the mutex? Insertion will not invalidate iterators (if I understood correctly), so this should be fine?

Framework/Kernel/inc/MantidKernel/Cache.h
SimonHeybrock: Interesting link, thanks! So apparently even two threads can insert at the same time, and one of the insertions would be dropped afterwards? What I am not sure about is whether or not *modification* of values is safe.

SimonHeybrock:
@quantumsteve I see you reverted the `tbb` changes, did you want to avoid potential issues with the release which is imminent, or was that just for testing?

quantumsteve:
@SimonHeybrock Yes I want to avoid issues with the imminent release. Making Cache thread-safe without (visible) mutexes is more complex than I originally thought.

PR #17461, opened 2016-09-12
Remove final occurence of HAS_UNORDERED_MAP_H

martyngigg:
`HAS_UNORDERED_MAP_H` is a hangover from the days where `unordered_map` was in `tr1`. This is now standard across all compilers so the check is no longer necessary and has been removed.

**To test:**

Code review and check all tests pass.

_No issue number_

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/API/inc/MantidAPI/SpectraDetectorTypes.h
+++ b/Framework/API/inc/MantidAPI/SpectraDetectorTypes.h
@@ -4,29 +4,16 @@
 // Includes
 //------------------------------------------------------------------------------
 #include "MantidGeometry/IDTypes.h"
-
-#ifndef HAS_UNORDERED_MAP_H
-#include <map>
-#else
-#include <tr1/unordered_map>
-#endif
+#include <unordered_map>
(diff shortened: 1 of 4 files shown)


SimonHeybrock:
I am not quite sure I understand: Previously, it seems `HAS_UNORDERED_MAP_H` was not defined (at least I cannot find it anywhere in our build or source), so all the typedefs defaulted to `std::map`. Now they will all be `std::unordered_map`. Correct?

martyngigg:
Correct. Long ago we had some compilers that did not have `unordered_map` in `tr1`  so we had that define in some places over the code. I thought we had removed all of these once we knew we could use `unordered_map` everywhere but it looks like we missed this one.

SimonHeybrock:
Ok, looks good then. This could potentially also provide some speedup 👍. Changing from `map` to `unordered_map` should not cause any trouble (apart from really obscure uses), unit tests pass, so :shipit:.

Issue #17423, opened 2016-09-07
Engg:: Closing fit interface can cause exception and crash


DavidFair:
### Expected behavior
Closing the interface should complete successfully
### Actual behavior
If the user closes the engineering interface or Mantid whilst on the fitting tab an exception will be thrown for a memory access violation.
### Steps to reproduce the behavior
It is unknown if this is related but a bad input was present (`ENGINX00193749.nxs`) at the time. 
Open the engineering interface
Select the fitting tab
Close the interface or Mantid
### Platforms affected
All

DavidFair:
Could be related to #17096

DavidFair:
This should be fixed by PR #17469 but if we are still getting random crashes we can open another issue

PR #17469, opened 2016-09-12
Engg: Fix crash on RB number change and closing interface

### LINKED ISSUE #17096: Engg:: Eng diff interface crashes on close after changing RB number
Changing the RB number within the interface then closing the window using the close button in the control box causes Mantid to crash.

Open the engineering diffraction interface. Change the value held in RB number and then close the window using the close button in the control box

### Platforms affected
Windows 
Linux 
(Possibly all)

DavidFair:
Description of work.
This PR fixes an issue where QT would attempt to deference a dangling pointer to the view (MVP pattern) if a notification was fired after processing a shutdown signal. 
The very first thing we now do when the view is being shutdown is set a flag in the presenter telling it that we are in the process of shutting down the view. If QT attempts to fire another notification we just return without performing any actions as we can't guarantee the state after. 

**To test:**
Open the Engineering Diffraction Interface.
Change the RB number at the top of the interface
Immediately close the interface - if it closes without crashing this PR works.

Fixes #17096 and possibly the random crashes when closing the interface described in #17423 .

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/EnggDiffraction/EnggDiffFittingPresenter.h
+++ b/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/EnggDiffraction/EnggDiffFittingPresenter.h
@@ -222,6 +222,9 @@ private:
 
   /// Associated view for this presenter (MVP pattern)
   IEnggDiffFittingView *const m_view;
+
+  /// Holds if the view is in the process of being closed
+  bool m_viewHasClosed;
 };
 
 } // namespace CustomInterfaces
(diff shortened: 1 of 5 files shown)


PranavBahuguna:
On testing this PR, the code performs as expected in the tests - there is no longer any crashes when RB number is changed and interface is closed. All seems good to me :shipit:

FedeMPouzols:
Good fix!

I wonder if other MVP interfaces could have similar issues (like the DataProcessingUI and reflectometry ones? @raquelalvarezbanos ?).

Somehow I find the new flag redundant. Would it be possible/better to just set `m_view` as `nullptr` on shutdown, and check `m_view` at the beginning of  the `notify()` methods instead of using a new flag?

DavidFair:
I had a look - the problem is its the view who held the member whose nullptr that was dereferenced. 
We could do some refactoring and make it so that the view is the first thing we call the destructor on then we are guaranteed to not receive any notifications from QT however we need to be careful as a lot of the code assumes m_view is always accessible so we could potentially introduce the same issue into the presenter. 
I'll open a low priority maintenance issue and look if there is a way we can safely close everything down and ensure QT doesn't attempt to fire more notifications without having to do any additional checks like this.

raquelalvarezbanos:
Thanks @FedeMPouzols I haven't seen this happening in the Reflectometry interface, but I'll bear it in mind.

PR #17473, opened 2016-09-12
17454 build without ph

### LINKED ISSUE #17454: Mantid is not building on Windows without precompiled headers
If one tries to build recent Mantid without pre-compiled headers, build fails. 

### Platforms affected
Discovered on Windows 10

abuts:
This should enable building Mantid under windows without precompiled headers enabled. 

**To test:**
Test by code review. Changes are trivial. 

You probably need windows, VS2015 SP3 to test it properly, then when building with cmake you should disable precompiled headers. Current mantid build fails before changes but you can trust me -- it works after the changes. 

Fixes #17454.

Internal, no release notes are necessary.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/DataHandling/inc/MantidDataHandling/LoadEventNexus.h
+++ b/Framework/DataHandling/inc/MantidDataHandling/LoadEventNexus.h
@@ -14,7 +14,11 @@
 #include "MantidKernel/TimeSeriesProperty.h"
 #include "MantidDataHandling/EventWorkspaceCollection.h"
 
+#ifdef _WIN32 // fixing windows issue causing conflict between
+// winnt char and nexus char
 #undef CHAR
+#endif
+
 #include <nexus/NeXusFile.hpp>
 #include <nexus/NeXusException.hpp>
 
--- a/Framework/DataHandling/inc/MantidDataHandling/LoadEventNexus.h
+++ b/Framework/DataHandling/inc/MantidDataHandling/LoadEventNexus.h
@@ -7,14 +7,17 @@
 #include "MantidAPI/IFileLoader.h"
 #include "MantidDataHandling/BankPulseTimes.h"
 #include "MantidDataObjects/EventWorkspace.h"
-#include <nexus/NeXusFile.hpp>
-#include <nexus/NeXusException.hpp>
 #include "MantidDataObjects/Events.h"
 #include "MantidAPI/WorkspaceGroup.h"
 #include "MantidGeometry/Instrument.h"
 #include "MantidGeometry/Instrument/ParameterMap.h"
 #include "MantidKernel/TimeSeriesProperty.h"
 #include "MantidDataHandling/EventWorkspaceCollection.h"
+
+#undef CHAR
+#include <nexus/NeXusFile.hpp>
+#include <nexus/NeXusException.hpp>
+
 #include <memory>
 #include <mutex>
 #include <boost/lexical_cast.hpp>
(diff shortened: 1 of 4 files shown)


MantidPlot/src/WindowFactory.h
```diff
@@ -14,6 +14,7 @@
 #include <cstring>
 #include <map>
 #include <vector>
+#include <iterator>
```
peterfpeterson: This doesn't look like it belongs in this PR

abuts: It does belongs to it -- it need <iterator> to compile without PH

Framework/DataHandling/inc/MantidDataHandling/LoadEventNexus.h
```diff
@@ -7,14 +7,21 @@
...
 #include "MantidDataObjects/Events.h"
 #include "MantidAPI/WorkspaceGroup.h"
 #include "MantidGeometry/Instrument.h"
 #include "MantidGeometry/Instrument/ParameterMap.h"
 #include "MantidKernel/TimeSeriesProperty.h"
 #include "MantidDataHandling/EventWorkspaceCollection.h"
+
+#ifdef _WIN32 // fixing windows issue causing conflict between
+// winnt char and nexus char
+#undef CHAR
```
DavidFair: Where is `CHAR` defined - I've looked and it's not a standard define from the [MSVC compiler](https://msdn.microsoft.com/en-us/library/b0084kay.aspx) and grepping for `#define CHAR` isn't revealing anything? 
Although it's extremely unlikely if anyone relies on the previous definition of `CHAR` and Nexus `CHAR` redefines it by doing this it could lead to issues.

abuts: char conflicting with nexus is defined in winnt.h and its probably the windows compiler bug as nexus char is defined in nexus namespace.  
The compiler bug is probable as it works fine if precompiled headers are enabled but fails if they are not. 
Despite that, you still need to be clever to be able to compile this piece of code without precompiled headers.

DavidFair:
Builds on Windows 10 with Visual Studio 2015 Update 3 so it resolves the issue. :shipit:

Framework/DataHandling/inc/MantidDataHandling/LoadEventNexus.h
martyngigg: I think it might be better to remove the `nexus` includes from this header and just forward declare the `::NeXus::File` class

abuts: I tried this -- does not work or have not find a way to make it working. 

The conflict occurs when ::Nexus templates are instantiated within ProcessBankData.c where winnt. and nexus.hpp definitions are clashing. 

Naturally I can not predefine only ::Nexus::file as more nexus objects are defined in this header, requering full definition.

Framework/DataHandling/src/ProcessBankData.cpp
```diff
@@ -1,6 +1,6 @@
 #include "MantidDataHandling/ProcessBankData.h"
-#include "MantidDataHandling/LoadEventNexus.h"
-#include "MantidDataObjects/Events.h"
+
+//#include "MantidDataObjects/Events.h"
```
quantumsteve: Why are you keeping commented out code?

abuts: that's a leftover -- Initially I was not sure It would compile without and then forget to remove

MantidPlot/src/WindowFactory.h
quantumsteve: `<iterator>` defines `std::back_inserter(...)`

PR #17487, opened 2016-09-13
Disable file logging for development builds

martyngigg:
Disables logging to a file for developer builds to avoid race condition issues with running unit tests in parallel. It should solve the recurring problems, particularly on Windows, with randomly failing tests. It should also speed up the tests.

I have also take the chance to simplify the path handling of the log file and moved the default location for windows to `AppDataDir` to match Linux. 

**To test:**

Code review. Please make sure you agree with the changes. All tests should pass.

_No issue number_

---
#### Reviewer

##### Code Review
##### Functional Tests


peterfpeterson:
The new style of [building mpi](http://builds.mantidproject.org/view/All/job/master_clean_rhel6_mpi_no_tests/) uses the root level `CMakeLists.txt`. I think you may be changing dead code in `Framework/Kernel/CMakeLists.txt`. The main part of the build script is
```shell
# Use the new Framework build technique (see pull request #17127)
scl enable devtoolset-2 "cmake -G 'Unix Makefiles' -DMPI_BUILD=ON -DENABLE_MANTIDPLOT=OFF -DCMAKE_SKIP_INSTALL_RPATH=ON -DCMAKE_INSTALL_PREFIX=/sw/fermi/mantid-mpi -DINSTALL_ROOT=/sw/fermi/mantid-mpi $WORKSPACE"
scl enable devtoolset-2 "cmake --build . --target package -- -j$BUILD_THREADS"
```

martyngigg:
Isn't the properties file still generated there though? `MPI_BUILD` is still defined so the code in `CMakeLists.txt` is still being used isn't it?

peterfpeterson:
I guess I didn't expand the diff enough to see that bit of context.

peterfpeterson:
Would it be better to add a `cmake` variable to enable file logging and turn it off on the build servers? That way, file logging is still stressed by developers, but skipped by the build servers.

martyngigg:
I agree, having it as an option is better and leaving developers stressing the functionality is better.

buildconfig/Jenkins/buildscript
```diff
@@ -185,7 +185,7 @@ rm -f *.dmg *.rpm *.deb *.tar.gz
 ###############################################################################
 # CMake configuration
 ###############################################################################
-$SCL_ON_RHEL6 "cmake ${CMAKE_GENERATOR} -DCMAKE_BUILD_TYPE=${BUILD_CONFIG} -DENABLE_CPACK=ON -DMAKE_VATES=ON -DParaView_DIR=${PARAVIEW_DIR} -DMANTID_DATA_STORE=${MANTID_DATA_STORE} -DDOCS_HTML=ON -DENABLE_CONDA=ON ${DIST_FLAGS} ${PACKAGINGVARS} .."
+$SCL_ON_RHEL6 "cmake ${CMAKE_GENERATOR} -DCMAKE_BUILD_TYPE=${BUILD_CONFIG} -DENABLE_CPACK=ON -DMAKE_VATES=ON -DParaView_DIR=${PARAVIEW_DIR} -DMANTID_DATA_STORE=${MANTID_DATA_STORE} -DDOCS_HTML=ON -DENABLE_CONDA=ON -DENABLE_FILE_LOGGING=OFF ${DIST_FLAGS} ${PACKAGINGVARS} .."
```
peterfpeterson: Implementing it this way means release builds (and nightly, ...) won't have file logging enabled unless the user explicitly puts it in their properties file. It should probably be put together with the `if [[ "$BUILDPKG" == true ]];` logic on line 134 and combined into the `PACKAGINGVARS` variable. The `else` that will need to be added is adding the extra flag. Something similar will need to be done in the windows build script.

Framework/Kernel/CMakeLists.txt
```diff
@@ -593,11 +593,15 @@ if ( MPI_BUILD )
   set ( PV_PLUGINS "" )
   set ( IGNORE_PARAVIEW "1" )
 else ()
+  option ( ENABLE_FILE_LOGGING "Enable logging to file for development builds. It is always enabled for packages" ON )
```
peterfpeterson: Does this belong outside of the `MPI_BUILD` check? `FILELOGGER_CHANNEL` is not a defined variable if a `MPI_BUILD` is used with this version of the code.

buildconfig/Jenkins/buildscript
martyngigg: Actually the properties file for the packages always has file logging enabled so users won't have to do anything. There is a separate file created for installers, see [here](https://github.com/mantidproject/mantid/pull/17487/files#diff-1b97644da97e0ada3471c80aa1b4eb1dR633). 

Should I make the option more explicit? `ENABLE_DEVBUILD_FILE_LOGGING` perhaps?

peterfpeterson:
The current version generates the "correct" properties file with `MPI_BUILD`.

martyngigg:
Is this ready to go? I'm hoping it will bring back stability to the builds.

peterfpeterson:
:shipit:

PR #17514, opened 2016-09-16
Reduce memory consumption of Histogram performance test.

SimonHeybrock:
See https://github.com/mantidproject/mantid/pull/17121#issuecomment-247519671.
`master_performancetests` has been failing since that PR has been merged, this should fix the issue (too much memory).

**To test:**

Code review, no issue, no release notes.

---
#### Reviewer

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):
##### Code Review
- [x] Is the code of an acceptable quality?
- [x] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)? Is it well structured with small focussed classes/methods/functions?
- [x] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?
##### Functional Tests
- [x] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant documentation been added/updated?
- [ ] Is user-facing documentation written in a user-friendly manner?
- [ ] Has developer documentation been updated if required?
- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

Probably this also fixes an issue with one of the tests: test_share_X
would previously test self-assignment, which was probably faster than
actually setting a new X pointer.

PR #17515, opened 2016-09-16
Reduce memory consumption in Histogram performance test by another 2x

SimonHeybrock:
See https://github.com/mantidproject/mantid/pull/17514. The performance test build server has 3.6 GB RAM. Previously we got very close to this and the test ran extremely slow. Reducing this by 2x should bring the memory consumption below 2 GB.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/HistogramData/test/HistogramTest.h
+++ b/Framework/HistogramData/test/HistogramTest.h
@@ -988,7 +988,7 @@ public:
   }
 
 private:
-  const size_t nHists = 100000;
+  const size_t nHists = 50000;
   const size_t histSize = 4000;
   std::vector<Histogram> hists;
   HistogramX xData;


LamarMoore:
:shipit:

FedeMPouzols:
Looks good again. Let's see how fast/slow is the `performancetests` machine with ~2GB memory consumption. After this we could add a note on memory consumption on the [wiki page](http://www.mantidproject.org/Writing_Performance_Tests) (although there will be the usual issues with outdated figures over time).

SimonHeybrock:
@FedeMPouzols I am also getting into contact with or hardware group, to see if we can get a better build server. Not sure though, since they use VMs for everything, and that might not work so well for performance tests...

SimonHeybrock:
@FedeMPouzols Went down from 108 seconds to 2.3 seconds :)

PR #17531, opened 2016-09-16
Disable log file flushing by default

### LINKED ISSUE #17513: Logging makes execution 3x slower on Windows.
### Expected behavior

Logging should (at least by default) have a minimal effect on the execution time. 

### Actual behavior

System Test: HFIRTestsAPIv2.HFIRTestsAPIv2
time taken (defaults)
282.54 second
284.69 seconds
289.60 seconds

time taken (disable logging by removing code in ThreadSafeLogStreamBuf::writeToDevice)
92.00 seconds
91.25 seconds
96.05 seconds

I would expect other system tests to show similar results.

From the build directory run `systemtest.bat -C Release -R HFIRTestsAPIv2.HFIRTestsAPIv2` 

Comment out [ThreadSafeLogStream.cpp L31-L39](
(https://github.com/mantidproject/mantid/blob/e84e49c512eca65625f4add4db1beeb0b0127c0e/Framework/Kernel/src/ThreadSafeLogStream.cpp#L31)

rebuild and rerun the system test HFIRTestsAPIv2.HFIRTestsAPIv2 .

### Platforms affected

I can only reproduce this issue on Windows

I tried doing the same timings on OS X and the results were all within 3 seconds.

martyngigg:
The Poco `FileChannel` requests that `LogFile` buffers are flushed on receipt of each newline character. These changes set the `.flush` property to `false` to leave this decision up to the OS.

**To test:**

See instructions in the issue to reproduce the issue.

Fixes #17513

[Release notes updated](https://github.com/mantidproject/mantid/commit/6d1bc6fe16453aea8c02708600acc2a64f9b14c2)

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/docs/source/release/v3.8.0/framework.rst
+++ b/docs/source/release/v3.8.0/framework.rst
@@ -15,7 +15,7 @@ Framework Changes
 
 - A race condition when accessing a singleton from multiple threads was fixed.
 
-- Log file buffers are no longer flushed by default for each newline received, increasing the speed of some system tests on Windows by 3x. 
+- Log file buffers are no longer flushed by default for each newline received, increasing the speed of some system tests on Windows by 4.5x.
 
 HistogramData
 -------------
--- a/docs/source/release/v3.8.0/framework.rst
+++ b/docs/source/release/v3.8.0/framework.rst
@@ -13,7 +13,9 @@ Framework Changes
   is almost identical, but a small number of cases (such as adding the workspaces ``Z`` and ``z``) will work
   in a more predictable manner.
 
-- A race condition when accessing a singleton from multiple threads was fixed. 
+- A race condition when accessing a singleton from multiple threads was fixed.
+
+- Log file buffers are no longer flushed by default for each newline received, increasing the speed of some system tests on Windows by 3x. 
 
 HistogramData
 -------------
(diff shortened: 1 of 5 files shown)


quantumsteve:
Do we also want to add this option to the [default user properties file](https://github.com/mantidproject/mantid/blob/f41b7c25708dfbc594e900df2451355d4f1da261/Framework/Kernel/src/ConfigService.cpp#L644)? The [Poco documentation](https://pocoproject.org/docs/Poco.FileChannel.html) suggests flushing at the end end of each line _may_ make a difference when investigating a system crash.

docs/source/release/v3.8.0/framework.rst
```diff
@@ -13,7 +13,9 @@ Framework Changes
   is almost identical, but a small number of cases (such as adding the workspaces ``Z`` and ``z``) will work
   in a more predictable manner.
 
-- A race condition when accessing a singleton from multiple threads was fixed. 
+- A race condition when accessing a singleton from multiple threads was fixed.
+
+- Log file buffers are no longer flushed by default for each newline received, increasing the speed of some system tests on Windows by 3x.
```
quantumsteve: Turning of flushing to disk makes `HFIRTestsAPIv2.HFIRTestsAPIv2` run even faster than with just commenting out `ThreadSafeLogStreamBuf::writeToDevice(...)`.  In that case, the speedup between master and this pull request is closer to 4.5x

martyngigg:
Good point about the default user properties file. It's worth the option being there.

docs/source/release/v3.8.0/framework.rst
martyngigg: Even better!

quantumsteve:
I confirmed that these changes dramatically improve logging performance on Windows. The same tests show no change on OS X. :shipit:

PR #17558, opened 2016-09-21
Move matplotlib import to plot function in ISISDisk

### LINKED ISSUE #17557: Change matplotlib import in PyChop
Unit tests for PyChop fails on machines without matplotlib. The tests themselves act only on the CLI part of PyChop, but there is an `import` at the start of the `ISISDisk` module because a method of the `ISISDisk` class is to plot the distance-time diagram. This is not needed on for the unit tests, so the `import` should be moved into the plotting function.

mducle:
This is just a quick PR to move the `import` of matplotlib into the function which needs it, so that unit tests which doesn't need matplotlib can run on machines without mpl.

Fixes #17557 

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/scripts/PyChop/ISISDisk.py
+++ b/scripts/PyChop/ISISDisk.py
@@ -9,7 +9,6 @@ spectrometer (LET) - using the functions in MulpyRep and additional tables of in
 import numpy as np
 from . import MulpyRep
 from .ISISFermi import ISISFermi
-from matplotlib import pyplot
 
 class ISISDisk:
     """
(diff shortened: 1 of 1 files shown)


scripts/PyChop/ISISDisk.py
```diff
@@ -338,6 +337,7 @@ def plotFrame(self, h_plt=None, Ei_in=None, frequency=None):
         Plots the time-distance diagram into a given Matplotlib axes, i
         for a give focused incident energy (in meV) and chopper frequencies (in Hz).
         """
+        from matplotlib import pyplot
```
AndreiSavici: You need to import this only if `h_plt==None`. And I would use  `try...except ImportError`

AndreiSavici:
:shipit:

martyngigg (approved):

PR #17563, opened 2016-09-21
Use const string reference to remove cpp check warning

### LINKED ISSUE #17447: Investigate and fix CppCheck violation in Projection.cpp - setUnits
Currently in `Projection.cpp` under `PythonInterface` the function `setUnits` takes a string parameter as a non const value type.
It then proceeds to use this in a single `if` statement - for this reason it probably could be turned into a `const &` type which would also remove the violation and remove the creation of rvalues. 

We need to write a unit test to check this doesn't unintentionally change the behaviour - especially as this is part of the interface for Python before making the change.

DavidFair:
Description of work.
This changes a small function which only passes a string for a comparison to be a const reference avoiding the overhead of copying the string into the function. 

**To test:**
The CI test should pass - specifically `ProjectionTest`

Fixes #17447 .

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/PythonInterface/mantid/api/src/Exports/Projection.cpp
+++ b/Framework/PythonInterface/mantid/api/src/Exports/Projection.cpp
@@ -23,7 +23,7 @@ std::string getUnit(Projection &p, size_t nd) {
   return (p.getUnit(nd) == RLU ? "r" : "a");
 }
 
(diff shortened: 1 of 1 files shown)


quantumsteve:
LGTM :shipit:

PR #17571, opened 2016-09-22
Re #17557 - added check for mpl in PyChop/__init__.py

### LINKED ISSUE #17557: Change matplotlib import in PyChop

mducle:
Adds an additional test to not load Matplotlib on systems that don't have it - e.g. the build servers for running unit tests which doesn't test the GUI.

**To test:**

Check that unit tests run cleanly on systems without mpl.

Fixes #17557 .

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/scripts/PyChop/PyChopGui.py
+++ b/scripts/PyChop/PyChopGui.py
@@ -2,6 +2,7 @@
 # pylint: disable=line-too-long, invalid-name, unused-argument, unused-import, multiple-statements
 # pylint: disable=attribute-defined-outside-init, protected-access, super-on-old-class, redefined-outer-name
 # pylint: disable=too-many-statements, too-many-instance-attributes, too-many-locals, too-many-branches
+# pylint: disable=too-many-public-methods
 
 """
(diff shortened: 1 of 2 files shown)


martyngigg (changes requested):
This looks fine.

Is it worth defining a `showGUI` in the except case that just prints a warning when called? This avoids a user having to check if `showGUI` is defined?

martyngigg:
@mducle The pylint build has failed

mducle:
ooops. Sorry... added more ignores. I should probably fix it rather tell pylint to ignore it, but a bit late now.

martyngigg (approved):
Thanks. This looks good now.

:shipit:

PR #17578, opened 2016-09-22
Instrument counting rate

### LINKED ISSUE #16834: Algorithm to calculate instrument counting rate
In normal circumstances an instrument in event mode counts neutrons with constant steady rate. Sometimes hardware issues cause it to count much faster. This appears as spurious signals on the final neutronic images and users want to filter this signal.

The suggested algorithm should be the basis of the filtering. It should calculate neutrons counting rate as the function of the experiment time and add appropriate logs to the event workspace for further event filtering on the basis of these logs.

abuts:
Fixes #16834 Algorithm to calculate instrument counting rate in event mode. 
See ticket for the description on why one wants to do it. Also look at the algorithm documentation -- test if explanation is clear.  

**To test:**

Try to break it. 
I have this algorithm's working prototype, written in Python but it is extremely slow and clumsy, e.g. the prototype spends ~12 min to process typical 4hrs Merlin run, when this algorithm finishes calculations in less than 10 sec. This is why the main purpose of this algorithm is speed and robustness vrt. various input conditions one may find in a real experiment.
#### Reviewer

##### Code Review
##### Functional Tests


Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+
+  // define X-axis in target units
+  double dX = (Xmax - m_XRangeMin) / numXBins;
+  std::vector<double> xx(numXBins);
+  for (int i = 0; i < numXBins; ++i) {
+    xx[i] = m_XRangeMin + (0.5 + static_cast<double>(i)) * dX;
+  }
+  auto ax0 = new API::NumericAxis(xx);
+  ax0->setUnit(RangeUnits);
+  m_visWs->replaceAxis(0, ax0);
```
soininen: To make the ownership change crystal clear, you could use
```c++
auto ax0 = make_unique<API::NumericAxis>(xx);
ax0->setUnit(RangeUnits);
m_visWs->replaceAxis(0, ax0.release());
```

You can also reuse `ax0` below instead of introducing `ax1`.

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+
+/** Helper function, mainly for testing
+* @return  true if count rate should be normalized and false
+* otherwise */
+bool CalcCountRate::notmalizeCountRate() const { return m_normalizeResult; }
+/** Helper function, mainly for testing
+* @return  true if log derivative is used instead of log itself */
+bool CalcCountRate::useLogDerivative() const { return m_useLogDerivative; }
+
+/** method to prepare normalization vector for the visuzliation workspace using
```
soininen: visuzliation -> visualization

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+* data from normalization log with, usually, different number of time steps
+* Here we assume that the numner of time points in the visualization workspace
```
soininen: numner -> number

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+* is smaller or equal to the number of points in the normalization log.
+*/
+void CalcCountRate::buildVisWSNormalization(
+    std::vector<double> &normalization) {
```
soininen: Why do you need to pass the `normalization` parameter? It seems to be unused here.

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+  // visualization workspace should be present and initialized at this stage:
+  auto ax = dynamic_cast<API::NumericAxis *>(m_visWs->getAxis(1));
+  if (!ax)
+    throw std::runtime_error(
+        "Can not retrieve Y-axis from visualization workspace");
+
+  normalization.assign(ax->length(), 0.);
+  // For more accurate logging (in a future, if necessary:)
+  // auto t_bins = ax->createBinBoundaries();
+  // double dt = t_bins[2] - t_bins[1]; // equal bins, first bin may be werid.
```
soininen: werid -> weird

Framework/Algorithms/test/CalcCountRateTest.h
```diff
@@ -0,0 +1,478 @@
...
+    CalcCountRateTester alg;
+
+    alg.initialize();
+    alg.setProperty("Workspace", sws);
+    alg.setProperty("RangeUnits", "dSpacing");
+
+    alg.setSearchRanges(sws); // explicit ranges:
+    //-- real workspace ranges returned
+
+    auto ranget = alg.getXRanges();
```
soininen: ranget -> ranges? Ranget might be the correct plural in Finnish, but I guess we need to stick to English here.

Framework/Algorithms/test/CalcCountRateTest.h
```diff
@@ -0,0 +1,478 @@
...
+    alg.initialize();
+
+    alg.setProperty("NumTimeSteps", 120);
+    alg.setProperty("NormalizeTheRate", true);
+    alg.setProperty("UseLogDerivative", true);
+    alg.setProperty("UseNormLogGranularity", true);
+
+    alg.setProperty("Workspace", sws);
+
+    TS_ASSERT_THROWS_NOTHING(alg.execute());
```
soininen: To my understanding, calling `execute()` almost never throws (even if the algorithm actually does!) unless you call `alg.setRethrows(true)` first. So I recommend doing that in all the affected tests.

Framework/Algorithms/test/CalcCountRateTest.h
```diff
@@ -0,0 +1,478 @@
...
+    const MantidVec &X = testVisWS->readX(0);
+    const MantidVec &Y = testVisWS->readY(0);
+    TS_ASSERT_EQUALS(X.size(), 201);
+    TS_ASSERT_EQUALS(Y.size(), 200);
+    auto Yax = dynamic_cast<API::NumericAxis *>(testVisWS->getAxis(1));
+    TS_ASSERT(Yax);
+    if (!Yax)
+      return;
+
+    // auto &YaxVal = Yax->getValues();
```
soininen: Commented-out code?

Framework/Algorithms/test/CalcCountRateTest.h
```diff
@@ -0,0 +1,478 @@
...
+
+    auto newLog = dynamic_cast<Kernel::TimeSeriesProperty<double> *>(
+        sws->run().getLogData("block_count_rate"));
+    TS_ASSERT(newLog);
+    if (!newLog)
+      return;
+    // these times are shifted by start time so firts time is 0
+    // auto times = newLog->timesAsVectorSeconds();
```
soininen: firts -> first
Commented-out code?

Framework/Algorithms/test/CalcCountRateTest.h
```diff
@@ -0,0 +1,478 @@
...
+
+    TS_ASSERT_THROWS_NOTHING(alg.execute());
+
+    API::MatrixWorkspace_sptr testVisWS = alg.getProperty("VisualizationWs");
+    TS_ASSERT(testVisWS);
+    if (!testVisWS)
+      return;
+    TS_ASSERT_EQUALS(testVisWS->getNumberHistograms(), 50);
+    const MantidVec &X = testVisWS->readX(0);
+    const MantidVec &Y = testVisWS->readY(0);
```
soininen: The `readX()` and `readY()` methods are deprecated. Please, use `x()` and `y()` instead.

Framework/Algorithms/test/CalcCountRateTest.h
```diff
@@ -0,0 +1,478 @@
...
+    TS_ASSERT_EQUALS(X.size(), 201);
+    TS_ASSERT_EQUALS(Y.size(), 200);
+    auto Yax = dynamic_cast<API::NumericAxis *>(testVisWS->getAxis(1));
+    TS_ASSERT(Yax);
+    if (!Yax)
+      return;
+
+    // auto &YaxVal = Yax->getValues();
```
soininen: Commented-out code?

Framework/Algorithms/test/CalcCountRateTest.h
```diff
@@ -0,0 +1,478 @@
...
+
+    auto newLog = dynamic_cast<Kernel::TimeSeriesProperty<double> *>(
+        sws->run().getLogData("block_count_rate"));
+    TS_ASSERT(newLog);
+    if (!newLog)
+      return;
+    // these times are shifted by start time so firts time is 0
+    // auto times = newLog->timesAsVectorSeconds();
```
soininen: Copy-paste typo (firts)? Commented out code?

Framework/Algorithms/test/CalcCountRateTest.h
```diff
@@ -0,0 +1,478 @@
...
+    MantidVec counts = newLog->valuesAsVector();
+
+    for (size_t i = 0; i < testVisWS->getNumberHistograms(); ++i) {
+      const MantidVec &Y = testVisWS->readY(i);
```
soininen: `readY()` deprecated.

Framework/Algorithms/test/CalcCountRateTest.h
```diff
@@ -0,0 +1,478 @@
...
+      double sum = std::accumulate(Y.begin(), Y.end(), 0.);
+      TSM_ASSERT_DELTA("Incorrect counts at index: " +
+                           boost::lexical_cast<std::string>(i),
```
soininen: Should we move to the `std::to_string()` functions at some point?

Framework/DataObjects/src/EventList.cpp
```diff
@@ -2220,6 +2220,47 @@ void EventList::generateCountsHistogramPulseTime(const MantidVec &X,
     }
   } // end if (there are any events to histogram)
 }
+/** With respect to PulseTime Fill a histogram given homoheneopus histogram
```
soininen: Fill -> fill, homoheneopus -> homogeneous?

Framework/DataObjects/src/EventList.cpp
```diff
@@ -2220,6 +2220,47 @@ void EventList::generateCountsHistogramPulseTime(const MantidVec &X,
...
+*                in binning.
+* @param xMax :: Maximal Pulse time value to constrain binning by (include the
+*                times smaller than right boundary, excluding equal)
+* @param Y :: The generated counts histogram
+* @param TOF_min -- min TOF used to constrain histogram ranges
+* @param TOF_max -- max TOF used to constrain histogram ranges
+*/
+void EventList::generateCountsHistogramPulseTime(const double &xMin,
+                                                 const double &xMax,
+                                                 MantidVec &Y,
```
soininen: Why would you pass `xMin` and `xMax` as const references?

Framework/DataObjects/src/EventList.cpp
```diff
@@ -3106,6 +3147,56 @@ DateAndTime EventList::getPulseTimeMax() const {
...
+  // no events is a soft error
+  if (this->empty())
+    return;
+
+  // when events are ordered by pulse time just need the first/last values
+  if (this->order == PULSETIME_SORT) {
+    switch (eventType) {
+    case TOF:
+      tMin = this->events.begin()->pulseTime();
+      tMax = this->events.rbegin()->pulseTime();
```
soininen: `events.front()` and `events.back()`?

Framework/DataObjects/src/EventList.cpp
```diff
@@ -3106,6 +3147,56 @@ DateAndTime EventList::getPulseTimeMax() const {
...
+      tMin = this->weightedEventsNoTime.begin()->pulseTime();
+      tMax = this->weightedEventsNoTime.rbegin()->pulseTime();
+      return;
+    }
+  }
+
+  // now we are stuck with a linear search
+  size_t numEvents = this->getNumberEvents();
+  DateAndTime temp = tMax; // start with the smallest possible value
+  for (size_t i = 0; i < numEvents; i++) {
```
soininen: Isn't there something like `std::minmax` in `<algorithm>`? Couldn't that be used here?

Framework/Kernel/src/TimeSeriesProperty.cpp
```diff
@@ -2139,6 +2139,52 @@ void TimeSeriesProperty<TYPE>::saveProperty(::NeXus::File *file) {
   saveTimeVector(file);
   file->closeGroup();
 }
+/** Calculate constant step hystogram of the time series data.
```
soininen: hystogram->histogram

docs/source/release/v3.9.0/direct_inelastic.rst
```diff
@@ -5,5 +5,12 @@ Direct Inelastic Changes
 .. contents:: Table of Contents
    :local:
 
+Improvements
+------------
+
+- New algorithm :ref:`CalcCountRate <algm-CalcCountRate>` allows to calculate instrument counting rate as function of the experiment 
+  time to be able to filter spurions, which may sometime appear on ISIS instruments.
```
soininen: sometime -> sometimes

docs/source/algorithms/CalcCountRate-v1.rst
```diff
@@ -0,0 +1,83 @@
...
+.. summary::
+
+.. alias::
+
+.. properties::
+
+Description
+-----------
+
+The algorithm works with event workspace produced by instrument operating in event mode and calculates the rate at which an instrument counts
```
soininen: workspace -> workspaces, instrument -> instruments

soininen (changes requested):

Framework/Algorithms/inc/MantidAlgorithms/CalcCountRate.h
```diff
@@ -0,0 +1,114 @@
...
+*/
+class DLLExport CalcCountRate : public API::Algorithm {
+public:
+  const std::string name() const override;
+  int version() const override;
+  const std::string category() const override;
+  const std::string summary() const override;
+  /// Helper function: true if count rate should be normalized and false
+  /// otherwise
+  bool notmalizeCountRate() const;
```
abuts: OK

Framework/Algorithms/inc/MantidAlgorithms/CalcCountRate.h
```diff
@@ -0,0 +1,114 @@
...
+  /// present on input workspace.
+  bool m_normalizeResult{false};
+  Kernel::TimeSeriesProperty<double> const *m_pNormalizationLog{nullptr};
+  /// default number of points in the target log
+  int m_numLogSteps{200};
+  /// specify if rate is calculated in selected frame interval (range defined)
+  /// or all frame should be used
+  bool m_rangeExplicit{false};
+  bool m_useLogDerivative{false};
+  /// spurion search ranges (TOF or unigs requested)
```
abuts: ok

Framework/Algorithms/inc/MantidAlgorithms/CalcCountRate.h
```diff
@@ -0,0 +1,114 @@
...
+
+  void
+  setOutLogParameters(const DataObjects::EventWorkspace_sptr &InputWorkspace);
+
+  void calcRateLog(DataObjects::EventWorkspace_sptr &InputWorkspace,
+                   Kernel::TimeSeriesProperty<double> *const targLog);
+
+  void checkAndInitVisWorkspace();
+
+  void hisogramEvents(const DataObjects::EventList &el,
```
abuts: ok

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+namespace Mantid {
+namespace Algorithms {
+
+// Register the algorithm into the AlgorithmFactory
+DECLARE_ALGORITHM(CalcCountRate)
+
+//----------------------------------------------------------------------------------------------
+
+/// Algorithms name for identification. @see Algorithm::name
+const std::string CalcCountRate::name() const { return "CalcCountRate"; }
```
abuts: That's the tricky one. As you may notice, initial name was Calculate but changed to Calc as second thought because its indeed too many Calculate algorithms. 
And according to Mantid conventions http://www.mantidproject.org/Mantid_Standards#Algorithms, the algorithms which do similar things should be named similarly, so the second suggestion is not comply with standards. 

I have no tough opinion on that, but the change to Calc was the second thought. I am glad to accept this change if you indeed think this is incorrect after the second thought.

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+  declareProperty(
+      Kernel::make_unique<Kernel::PropertyWithValue<std::string>>(
+          "RangeUnits", "Energy",
+          boost::make_shared<Kernel::StringListValidator>(
+              Kernel::UnitFactory::Instance().getKeys()),
+          Kernel::Direction::Input),
+      "The units from Mantid Unit factory for calculating the "
+      "counting rate and XMin-XMax ranges are in. If the "
+      "X-axis of the input workspace is not expressed"
+      "in this units, unit conversion will be performed, so the "
```
abuts: ok

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+      "workspace should contain all necessary information for this "
+      "conversion. E.g. if *RangeUnits* is *EnergyTransfer*, Ei "
+      "log containing incident energy value should be attached to the "
+      "input workspace.");
+  std::vector<std::string> propOptions{"Elastic", "Direct", "Indirect"};
+  declareProperty("EMode", "Elastic",
+                  boost::make_shared<Kernel::StringListValidator>(propOptions),
+                  "The range units above conversion mode (default: elastic)");
```
abuts: yes.

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+  setPropertyGroup("NormalizationLogName", used_logs_mode);
+  setPropertyGroup("UseNormLogGranularity", used_logs_mode);
+
+  declareProperty(
+      "CountRateLogName", "block_count_rate",
+      boost::make_shared<Kernel::MandatoryValidator<std::string>>(),
+      "The name of the processed time series log with count rate to be added"
+      " to the source workspace");
+  // visualisation group
+  std::string spur_vis_mode("Spurion visualisation");
```
abuts: ok

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+
+  DataObjects::EventWorkspace_sptr sourceWS = getProperty("Workspace");
+  API::EventType et = sourceWS->getEventType();
+  if (et == API::EventType::WEIGHTED_NOTIME) {
+    throw std::runtime_error("Event workspace " + sourceWS->name() +
+                             " contains events without necessary frame "
+                             "information. Can not process counting rate");
+  }
+
+  // Identity correct way to treat input logs and general properties of output
```
abuts: ok

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+
+#pragma omp parallel
+  {
+#pragma omp single
+    {
+      // initialize thread's histogram buffer
+      nThreads = PARALLEL_NUMBER_OF_THREADS;
+      Buff.resize(nThreads);
+      for (int i = 0; i < nThreads; i++) {
+        Buff[i].assign(m_numLogSteps, 0);
```
abuts: resizing before parallel section will not work, as real number of threads will be known in parallel section only, and before the parallel section its usually one. 

Common memory initialization has to be done on a single thread, otherwise each thread will reallocate memory and the first one will certainly already use memory reallocated by the last one. 

Allocating thread-specific memory indeed can be done by single thread. That's just instinct I follow on allocating memory -- making this change, though as usual with parallel, it makes me think a lot.

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+      }
+    }
+    if (!countNormalization.empty() && this->buildVisWS()) {
+#pragma omp for
+      for (int64_t j = 0; j < int64_t(m_visNorm.size()); j++) {
+        this->normalizeVisWs(j);
+      }
+    }
+  }
+  // recover histogram timing, but start assuming run starts at
```
abuts: old comment from prototype algorithm. Modified

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+      continue;
+    if (tof < m_XRangeMin || tof >= m_XRangeMax)
+      continue;
+
+    size_t n_spec = static_cast<size_t>((pulsetime - m_visT0) / m_visDT);
+    size_t n_bin = static_cast<size_t>((tof - m_XRangeMin) / m_visDX);
+    (spectraLocks + n_spec)->lock();
+    auto &Y = m_visWs->mutableY(n_spec);
+    Y[n_bin]++;
+    (spectraLocks + n_spec)->unlock();
```
abuts: its usually a bit slower as in addition to everything else causes memory allocation on heap. As its very unlikely the operators 295-296 will throw as all code for underlying memory is enclosed within the program and the locks are within the double loop, I think that  reliability vs speed compromise in this case can be safely resolved to speed.

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+  }
+}
+
+/** Normalize single spectra of the normalization workspace using prepared
```
abuts: ok

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+ * normzlization log
+ @param wsIndex -- appropriate visualization workspace index to normalize
+ spectra
+ */
+void CalcCountRate::normalizeVisWs(int64_t wsIndex) {
+
+  auto Y = m_visWs->dataY(wsIndex);
```
abuts: yes, noticed it.

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+}
+/*Analyse input log parameters and logs, attached to the workspace and identify
+ * the parameters of the target log, including experiment time.
+ *
+ @param InputWorkspace -- input workspace to analyse logs
+ */
+void CalcCountRate::setOutLogParameters(
+    const DataObjects::EventWorkspace_sptr &InputWorkspace) {
+
+  std::string NormLogName = getProperty("NormalizationLogName");
```
abuts: its normal mode of operation to overwrite countingRate log. But indeed, if one tries to overwrite proton_charge and use it for normalization, the result will be bad. Blocking this.

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+    m_pNormalizationLog = m_tmpLogHolder.get();
+    m_useLogDerivative = true;
+  }
+
+  ///
+  if (m_normalizeResult) {
+    if (!useLogAccuracy) {
+      g_log.warning() << "Change of the counting log accuracy while "
+                         "normalizing by log values is not implemented. Will "
+                         "use log accuracy.";
```
abuts: ok

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+  }
+
+  m_TRangeMin = runTMin;
+  // histogramming excludes rightmost events. Modify max limit to keep them
+  double t_epsilon = double(runTMax.totalNanoseconds()) *
+                     (1 + std::numeric_limits<double>::epsilon());
+  int64_t eps_increment =
+      static_cast<int64_t>(t_epsilon - double(runTMax.totalNanoseconds()));
+  m_TRangeMax =
+      runTMax + eps_increment; // *(1+std::numeric_limits<double>::epsilon());
```
abuts: it how it should look like but DateTime does not have multiplication operator.

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+    }
+
+    conv->setProperty("InputWorkspace", InputWorkspace);
+    conv->setPropertyValue("OutputWorkspace", wsName);
+    std::string Emode = getProperty("Emode");
+    conv->setProperty("Emode", Emode);
+    conv->setProperty("Target", RangeUnits);
+
+    conv->execute();
+    wst = conv->getProperty("OutputWorkspace");
```
abuts: I expect it to throw otherwise , though always unsure on when it does throw and when it does not. will be no harm to add the check.

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+    wst = InputWorkspace;
+  }
+  m_workingWS = boost::dynamic_pointer_cast<DataObjects::EventWorkspace>(wst);
+  if (!m_workingWS) {
+    throw std::runtime_error("SetWSDataRanges:Can not retrieve EventWorkspace "
+                             "after converting units");
+  }
+  // data ranges
+  m_XRangeMin = getProperty("XMin");
+  m_XRangeMax = getProperty("XMax");
```
abuts: Added

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+  }
+
+  if (m_XRangeMin == EMPTY_DBL()) {
+    m_XRangeMin = realMin;
+  }
+  if (m_XRangeMax == EMPTY_DBL()) {
+    // include rightmost limit into the histogramming
+    m_XRangeMax = realMax * (1. + std::numeric_limits<double>::epsilon());
+  }
+  if (m_XRangeMin < realMin) {
```
abuts: ok, may sometimes happens when units are converted

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+
+  m_visWs = boost::dynamic_pointer_cast<DataObjects::Workspace2D>(
+      API::WorkspaceFactory::Instance().create("Workspace2D", numTBins,
+                                               numXBins + 1, numXBins));
+  m_visWs->setTitle(visWSName);
+
+  double Xmax = m_XRangeMax;
+  // a bit dodgy code. It can be generalized
+  if (std::isinf(Xmax)) {
+    auto Xbin = m_workingWS->readX(0);
```
abuts: reasonable

Framework/Algorithms/src/CalcCountRate.cpp
abuts: that's mean in this case. Ownership is crystal clear in these 3 rows.

Framework/Algorithms/src/CalcCountRate.cpp
abuts: ok

Framework/Algorithms/src/CalcCountRate.cpp
abuts: ok

Framework/Algorithms/src/CalcCountRate.cpp
abuts: used, at the changed code its 661, 661. In original code it was also very complex algorithm including both interpolation and extrapolation but in final version I've decided that debugging it would be overkill for visualization purposes.

Framework/Algorithms/src/CalcCountRate.cpp
abuts: ok

Framework/Algorithms/test/CalcCountRateTest.h
abuts: ok

Framework/Algorithms/test/CalcCountRateTest.h
abuts: probably ideas for further testing. As not implemented, I am removing them

Framework/Algorithms/test/CalcCountRateTest.h
abuts: see above.

Framework/Algorithms/test/CalcCountRateTest.h
abuts: interesting. Suggested method returns not a vector but something else. Not important in this case, but the code itself is pretty convoluted. Wander, how it exposed to Python, bearing in mind there are tons of external python code using readX,readY

Framework/Algorithms/test/CalcCountRateTest.h
abuts: see above.

Framework/Algorithms/test/CalcCountRateTest.h
abuts: ibid.

Framework/Algorithms/test/CalcCountRateTest.h
abuts: ibid.

Framework/Algorithms/test/CalcCountRateTest.h
abuts: boost claims its the fastest, though it not important in this case. I also remember some problems with this on some OS (not Windows). But let's try.

Framework/DataObjects/src/EventList.cpp
abuts: just because was just passing DateAndTime this way. I think, compiler well knows what to do in this case.

Framework/DataObjects/src/EventList.cpp
abuts: I've just copied the expression used in getPulseTimeMin getPulseTimeMax. Following "minimal perturbations to framework" principle, let's live it as it is so would agree it may be better to change it everywhere in the EventList.

Framework/DataObjects/src/EventList.cpp
abuts: again, the code just copied from the existing one using minimal changes... Minmax may be ok, but comparison operator would be complex and the code becomes heterogeneous. "Minimal perturbations"...

Framework/Kernel/src/TimeSeriesProperty.cpp
abuts: ok

docs/source/release/v3.9.0/direct_inelastic.rst
abuts: ok

Framework/Algorithms/src/CalcCountRate.cpp
```diff
@@ -0,0 +1,653 @@
...
+
+/// Algorithms name for identification. @see Algorithm::name
+const std::string CalcCountRate::name() const { return "CalcCountRate"; }
+
+/// Algorithm's version for identification. @see Algorithm::version
+int CalcCountRate::version() const { return 1; }
+
+/// Algorithm's category for identification. @see Algorithm::category
+const std::string CalcCountRate::category() const {
+  return "Inelastic\\Utility";
```
abuts: Diagnostics good, events..., may be filtering, though not exactly. Ok, add it to filtering -- there somebody looking for spurion filtering may want to look and something like RebinBySampleTime is already sitting.

Framework/Algorithms/src/CalcCountRate.cpp
soininen: Indeed, now when I think about it, `CountRateCalculation` doesn't fit the naming scheme of Mantid. However, I wouldn't shorten Calculate to Calc, even if there are too many algorithms starting with that prefix. It would just create another 'subcategory' of algorithms, similarly to the ComputeSomething algorithms: I do remember there is a vanadium calibration algorithm, but I don't remember whether it was CalculateCalibrationCoefVan or ComputeCalibrationCoefVan. Adding Calc as a possible prefix would just complicate things. So I guess renaming this algorithm back to `CalculateCountRate` would be the best choice.

Framework/Algorithms/src/CalcCountRate.cpp
soininen: I'm just worrying about `mutableY()` throwing and the lock never being freed but maybe I'm being overly paranoid.

Framework/Algorithms/test/CalcCountRateTest.h
soininen: To my understanding, these new histogram methods are not yet exposed to Python, so it remains to be seen how they will work there.

Framework/Algorithms/src/CalcCountRate.cpp
abuts: its not in this case but allocating thread memory on a thread is always better. I red the docs and made the changes. Should be ok.

Framework/Algorithms/src/CalcCountRate.cpp
abuts: I share you concern but after careful consideration think it will never happens. Workspace is allocated within the program and should be well verified.

Framework/Algorithms/src/CalcCountRate.cpp
abuts: May be InstrimetCountingRate then? All algorithms calculate something so Calculate is in a way superficial

Framework/Algorithms/src/CalcCountRate.cpp
soininen: I wonder why `replaceAxis` doesn't take `unique_ptr` as parameter instead of raw pointer if the ownership is transferred... Anyway, to my understanding, `ax0->setUnit(RangeUnits)` may throw a `NotFoundError` and then ax0 will leak.

abuts: To be fair, setUnit will never throw 'notFound' as RangeUnits are selected from the list of the existing units by units property. 

I replaced the code by unique pointers as its not important here and I have no time for empty talks, though feel pretty annoyed by recent trend to use modern more expensive constructs where the cheaper older one work very well.

Framework/Algorithms/src/CalculateCountRate.cpp
```diff
@@ -0,0 +1,719 @@
...
+  }
+  if (m_XRangeMax > realMax) {
+    g_log.debug() << "Workspace constrain max range changed from: "
+                  << m_XRangeMax << " To: " << realMax << std::endl;
+    m_XRangeMax = realMax * (1. + std::numeric_limits<double>::epsilon());
+  }
+  // check final ranges valid
+  if (m_XRangeMin > m_XRangeMax) {
+    throw std::invalid_argument(
+        " Minimal spurion search data range is bigger than the maximal range.");
```
soininen: It's good that these tests are now here, but when I give only the XMax value, and it happens to be too small, the error I see is this 'Minimal spurion search data range is bigger than the maximal range.' which may be a bit confusing. I would expect the error message to be more like 'XMax too small'. Would it make sense to improve the error reporting in this case?

Framework/Algorithms/src/CalculateCountRate.cpp
```diff
@@ -0,0 +1,719 @@
...
+      Kernel::make_unique<Kernel::PropertyWithValue<int>>(
+          "NumTimeSteps", 200, mustBeReasonable, Kernel::Direction::Input),
+      "Number of time steps (time accuracy) the visualization workspace has. "
+      "Also number of steps in 'CountRateLogName' log if "
+      "'UseNormLogGranularity' is set to false. Should be bigger than 3");
+  declareProperty(
+      Kernel::make_unique<Kernel::PropertyWithValue<int>>(
+          "XResolution", 100, mustBeReasonable, Kernel::Direction::Input),
+      "Number of steps (accuracy) of the visualization workspace has along "
+      "X-axis. ");
```
soininen: Mantid dies, if *XResolution* is given a value-too-large. Would there be a sensible upper limit this property? Or maybe these kinds of deaths should be prevented at some lower lever of the software.

soininen (changes requested):
Looks good, except a few small typos remain here a there. Also, some property validation details need some consideration.

Framework/Algorithms/src/CalculateCountRate.cpp
```diff
@@ -0,0 +1,719 @@
...
+  m_normalizeResult = getProperty("NormalizeTheRate");
+  bool useLogDerivative = getProperty("UseLogDerivative");
+  bool useLogAccuracy = getProperty("UseNormLogGranularity");
+
+  bool logPresent = InputWorkspace->run().hasProperty(NormLogName);
+  if (!logPresent) {
+    if (m_normalizeResult) {
+      this->disableNormalization(
+          "Normalization log " + NormLogName +
+          "' values requested but the log is not attached to the "
```
abuts: ok

Framework/Algorithms/src/CalculateCountRate.cpp
abuts: I think its not dying but went into swap (Otherwise its bug which should be fixed elsewhere) on creation of visualization workspace when the workspace is too big for memory.  Its difficult to judge on achievable accuracy here (somebody may want to process single-spectra summed workspace with high accuracy vs big LET workspace from long run with bad accuracy) so the issue should be addressed within workspace factory where requested vs available memory should be estimated. I think it was there some time ago when we had file-based (managed) workspaces. Anyway, if this is an issue, it should clearly be addressed globally within the workspace factory.

Framework/Algorithms/src/CalculateCountRate.cpp
abuts: I think swapping the checks will resolve this issue.

Framework/Algorithms/src/CalculateCountRate.cpp
```diff
@@ -0,0 +1,719 @@
...
+    m_XRangeMax = realMax * (1. + std::numeric_limits<double>::epsilon());
+  }
+  // check final ranges valid
+  if (m_XRangeMin > m_XRangeMax) {
+    throw std::invalid_argument(
+        " Minimal spurion search data range is bigger than the maximal range.");
+  }
+  if (m_XRangeMax < realMin || m_XRangeMin > realMax) {
+    throw std::invalid_argument(
+        " Spurion data search range lies outsize of the workspace data range.");
```
abuts: ok

Framework/Algorithms/src/CalculateCountRate.cpp
```diff
@@ -0,0 +1,719 @@
...
+ */
+void CalculateCountRate::setOutLogParameters(
+    const DataObjects::EventWorkspace_sptr &InputWorkspace) {
+
+  std::string NormLogName = getProperty("NormalizationLogName");
+  std::string TargetLog = getProperty("CountRateLogName");
+  if (NormLogName == TargetLog) {
+    throw std::invalid_argument("Target log name: " + TargetLog +
+                                " and normalization log name: " + NormLogName +
+                                " Can not be the same");
```
abuts: ok

soininen (approved):

soininen:
The code looks good, and documentation & release notes have been written. Except for the out-of-memory situation (high *XResolution* value, discussed in the comments), I couldn't break the algorithm by manual runs. :shipit:

PR #17579, opened 2016-09-22
Python 3 Conversion - Group 7

DavidFair:
Description of work.
As part of #16739 this converts all the python algorithms in group 7. 

**To test:**
All system tests should pass especially those for the python interface.

_Does not need to be in the release notes._

---
#### Reviewer

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):
##### Code Review
- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)? Is it well structured with small focussed classes/methods/functions?
- [ ] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?
##### Functional Tests
- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant documentation been added/updated?
- [ ] Is user-facing documentation written in a user-friendly manner?
- [ ] Has developer documentation been updated if required?
- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

PR #17595, opened 2016-09-23
Python 3 Conversion - Group 8

DavidFair:
Description of work.

As part of #16739 this converts all the python algorithms in group 8. 

**To test:**
Build Mantid against Python 3.x following the instructions [here](http://www.mantidproject.org/Python_3)
Test each algorithm works correctly - also I would merge in master first as that contains an updated `find_data.py` algorithm which is needed for some algorithms 

---
#### Reviewer

##### Code Review
##### Functional Tests


DavidFair:
`find_data.py` is converted in #17579

DavidFair:
For the tester here is a list of algorithms changes:
```
OSIRISDiffractionReduction.py
PoldiDataAnalysis.py
Quest.py
ReactorSANSResolution.py
REFLReprocess.py
ResNorm2.py
SANSAbsoluteScale.py
SANSAzimuthalAverage1D.py
SANSBeamSpreaderTransmission.py
SANSDarkRunBackgroundCorrection.py
SANSDirectBeamTransmission.py
SANSFitShiftScale.py
SANSMask.py
SANSReduction.py
SANSStitch.py
SavePlot1D.py
SimulatedDensityOfStates.py
SofQWMoments.py
SwapWidths.py
TimeSlice.py
TOSCABankCorrection.py
```

Additionally I would merge branch `16739_python_3_conversion_group_7` whilst testing as it will fix `find_data.py`

rosswhitfield:
@DimitarTasev Are you testing this? If not can you remove your assignment and I will test it. Thanks!

DimitarTasev:
@rosswhitfield No problems, go ahead!

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/SimulatedDensityOfStates.py
```diff
@@ -488,7 +491,7 @@ def _compute_partial_ion_workflow(self, partial_ions, frequencies, eigenvectors,
         total_workspace = None
 
         # Output each contribution to it's own workspace
-        for ion_name, ions in partial_ions.items():
+        for ion_name, ions in list(partial_ions.items()):
```
rosswhitfield: You don't need to transform to list when you iterate over it. There are a fewer other places in this pull request where 2to3 did this also.

rosswhitfield:
`SimulatedDensityOfStatesTest` sometimes fails.
```
571: Traceback (most recent call last):
571:   File "/home/rwp/mantid/Framework/PythonInterface/test/python/plugins/algorithms/SimulatedDensityOfStatesTest.py", line 253, in test_isotopes_are_parsed_correctly
571:     self.assertEqual('wks_grp_C(13)', wks_grp.getItem(0).getName())
571: AssertionError: 'wks_grp_C(13)' != 'wks_grp_H(2)'
571: - wks_grp_C(13)
571: ?         ^ ^^
571: + wks_grp_H(2)
571: ?  
```

I believe this is because of `SimulatedDensityOfStates` storing the ions in a `dict` and in python 3 the `dict` doesn't have a guaranteed order. I don't know the best way to solve this.

DavidFair:
@rosswhitfield We could tackle this in one of two ways.
- If order is important we could use an ordered dictionary in the algorithm, but we need to ensure that `load_phonon` also returns an ordered dict. 
- If its only the unit test which is reliant on the order we could create a small function in the unit test that searches all the elements in the group for the test condition and return true or false directly into an assert instead.

I am not familiar enough with the algorithm to know which would be the best route to take?

rosswhitfield:
I'm also not familiar with the algorithm. @jburle and @ElliotAOram were the last two people to work on it so hopefully one of them can answer.

DavidFair:
I've talked to @louisemccann who is looking after this script now. We tested it against Python 3 vs 2 and the unordered nature does not cause any issues as the results generated as the same against both. 
So we are going to wrap the unit tests in a method which checks all elements in the dictionary instead of assuming the position in the interim.

In the longer term she wants to move to using an ordered dictionary so that the order the chemical formulas are passed in are consistent between runs. I have opened a low priority issue for this change at a later date and assigned it to her: #17750

rosswhitfield:
LGTM :shipit:

PR #17607, opened 2016-09-27
A 'moving window average' mode for `CalculateFlatBackground`

### LINKED ISSUE #17584: A moving window average mode for the `CalculateFlatBackground` algorithm
The [`CalculateFlatBackground`](http://docs.mantidproject.org/nightly/algorithms/CalculateFlatBackground-v1.html) algorithm currently supports two modes of operation, *Linear fit* and *Mean*. However, both modes are unfeasible for use with the TOF instruments at ILL as a fitting range needs to be specified for them. A mode which calculates the flat background as a moving window average would be handier.

soininen:
This PR implements a _moving window average_ mode for the [`CalculateFlatBackground`](http://docs.mantidproject.org/nightly/algorithms/CalculateFlatBackground-v1.html) algorithm. The new mode calculates a window average for each point in the spectra and takes the minimum as the flat background value. This is handy for TOF spectra acquired at ILL when there is no well-defined background plateau.

Resolves #17584 .

**To test:**

Try the algorithm on some real data and check that it returns sane results. With _AveragingWindowWidth_ set to 1 it should give just the minimum value. Note that the algorithm uses cyclic boundary conditions when calculating the moving average. The two original modes of operation (_Linear Fit_ and _Mean_) should work exactly as they used to except that _XMin_ and _XMax_ are not validated until the algorithm is run.

_Release notes have been updated accordingly._

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/Algorithms/src/CalculateFlatBackground.cpp
```diff
@@ -157,11 +206,18 @@ void CalculateFlatBackground::exec() {
...
+      case Modes::MEAN:
+        background = Mean(outputWS, currentSpec, startX, endX, variance);
+        break;
+      case Modes::MOVING_AVERAGE:
+        background = movingAverage(outputWS, currentSpec, windowWidth);
+        break;
+      }
       if (background < 0) {
         g_log.warning() << "Problem with calculating the background number of "
                            "counts spectrum with index " << currentSpec
```
mantid-roman: Could this warning be more specific? When I saw it in the log window I couldn't understand what was going on.

Framework/Algorithms/src/CalculateFlatBackground.cpp
```diff
@@ -443,5 +499,40 @@ double CalculateFlatBackground::LinearFit(API::MatrixWorkspace_sptr WS,
...
+  // while adding a new point. Saves us summing all the points in the window.
+  for (size_t i = 0; i < ys.size() - 1; ++i) {
+    currentMin = std::min(currentMin, sum / static_cast<double>(windowWidth));
+    sum -= ys[i];
+    size_t j = i + windowWidth;
+    // Cyclic boundary conditions.
+    if (j >= ys.size()) {
+      j -= ys.size();
+    }
+    sum += ys[j];
```
mantid-roman: A more stable way to update `sum` would be `sum += ys[j] - ys[i]`. This could prevent warning messages in some edge cases.

soininen: Indeed.

mantid-roman:
:shipit:

Framework/Algorithms/src/CalculateFlatBackground.cpp
```diff
@@ -38,22 +41,35 @@ void CalculateFlatBackground::init() {
   declareProperty(make_unique<WorkspaceProperty<>>("OutputWorkspace", "",
                                                    Direction::Output),
                   "Name to use for the output workspace.");
-  auto mustHaveValue = boost::make_shared<MandatoryValidator<double>>();
-
-  declareProperty("StartX", Mantid::EMPTY_DBL(), mustHaveValue,
-                  "The X value at which to start the background fit");
-  declareProperty("EndX", Mantid::EMPTY_DBL(), mustHaveValue,
-                  "The X value at which to end the background fit");
+  std::vector<std::string> modeOptions{"Linear Fit", "Mean", "Moving Average"};
```
martyngigg: We need to be careful when dealing with the property declarations as if we shift the order then it affects the Python generated `simpleapi` function.

On master this

```
python -c "import mantid.simpleapi as sapi; help(sapi.CalculateFlatBackground)"
```

gives

```
CalculateFlatBackground(InputWorkspace,StartX,EndX,WorkspaceIndexList=None,Mode=None,OutputMode=None,SkipMonitors=None,NullifyNegativeValues=None, Version=1)
```

as the function signature.

With these changes we now get

```
CalculateFlatBackground(InputWorkspace,Mode=None,StartX=None,EndX=None,AveragingWindowWidth=None,WorkspaceIndexList=None,OutputMode=None,SkipMonitors=None,NullifyNegativeValues=None, Version=1)
```

If we could guarantee that all users just use keyword arguments then this would be fine but at the moment a call such as `CalculateFlatBackground(ws, 1.0, 10.0)` in a user script would fail with these changes because the order of the arguments has changed. It's okay that `StartX` & `EndX` are now optional but we need to maintain the order of the arguments so that positional argument calls still work.

martyngigg (changes requested):

Framework/Algorithms/src/CalculateFlatBackground.cpp
soininen: Nice catch! Will fix.

martyngigg (approved):

PR #17609, opened 2016-09-27
Kernel::MersenneTwister: replace boost implementation with C++11 version

### LINKED ISSUE #14944: Kernel::MersenneTwister: replace the underlying boost implementation with the C++11 one
In Framework/Kernel/MersenneTwister.

Now that we have C++11 support on all platforms it should be possible to just replace `<boost/random/mersenne_twister.hpp>` and related headers, using the `std:mt19937` implementation available from  `<random>`. 

Remove the dependency on `boost/random/` headers. The `std` implementation resembles the `boost` one very closely so it should be possible to do this as a one-to-one replacement.

PranavBahuguna:
This PR focuses on removing the dependency on using `boost/random` headers and associated methods, instead using the `std::mt19937` implementation from `<random>` as C++11 support is now available on all platforms. As the sequence of numbers given by the new implementation for a specific seed are different, it was also necessary to modify several algorithm unit tests reliant on this RNG.

**To test:**

Simply ensure that all the tests that have been altered now pass.

Fixes #14944.

#### Reviewer

##### Code Review
##### Functional Tests


martyngigg:
@PranavBahuguna I think the milestone is set wrong here. Should this be in `3.9`?

Framework/Kernel/test/MersenneTwisterTest.h
```diff
@@ -139,10 +139,10 @@ class MersenneTwisterTest : public CxxTest::TestSuite {
 
 private:
   void assertSequenceCorrectForSeed_39857239(MersenneTwister &randGen) {
-    double expectedValues[10] = {
-        0.203374452656, 0.597970068222, 0.120683325687,  0.92372657801,
-        0.734524340136, 0.467380537419, 0.0712658402044, 0.204503614921,
-        0.487210249063, 0.885743656661};
+    double expectedValues[10] = {0.597970068269, 0.923726578038, 0.46738053759,
```
martyngigg: Do we know why the values change for the same seed given that it is supposed to be the same random number generator type?

PranavBahuguna: While I'm not certain of this, I believe that C++11 implementation of Mersenne Twister algorithm is different to that of the boost version (as Mersenne Twister is a class of RNG's rather than a specific type, so can produce different numbers for the same seed).

martyngigg: A comment in https://github.com/TRIQS/triqs/issues/226#issuecomment-126280169 suggests that it is actually `uniform_real_distribution` that produces different results and the underlying engine produces the same numbers. This is probably something we will have to live with.

Framework/Algorithms/test/AnnularRingAbsorptionTest.h
```diff
@@ -43,11 +43,11 @@ class AnnularRingAbsorptionTest : public CxxTest::TestSuite {
     MatrixWorkspace_sptr outWS = alg->getProperty("OutputWorkspace");
     TS_ASSERT(outWS);
 
-    const double delta(1e-08);
+    const double delta(1e-04);
```
DavidFair: Is 1e-04 enough precision - that seems like a lot of variation between platforms?

PranavBahuguna: It is a lot of variation. I had to change it as a direct result of the algorithm receiving different random numbers from the new Mersenne Twister algorithm, and the ones it now receives happen to produce a lot more variation than before. I'm not honestly sure what to do about this as the AnnularRingAbsorption algorithm itself is producing such extreme variation.

DavidFair:
So I looked into the reduced precision with `AnnularRingAbsorption`. The kernel still correctly returns the same values in the same order on both Linux and Windows from the PRNG so the issue isn't with the changes completed. 

As `AnnularRingAbsorption` relies on `MonteCarloAbsorption` to do the heavy work and the precision of that has not changed there must be something in the algorithm of both causing an issue.

Running the same algorithm on both platforms then subtracting the two resulting workspaces is resulting in between 1e-05 and 1e-07 difference randomly across the workspace - I don't know if this is too significant a difference so we need to clarify what impact this PR might have with that level of uncertainty. 

I've unassigned myself so that someone more familiar with `AnnularRingAbsorption` such as @rosswhitfield can look into this.

Framework/CurveFitting/test/Algorithms/ConvertToYSpaceTest.h
```diff
@@ -63,9 +63,9 @@ class ConvertToYSpaceTest : public CxxTest::TestSuite {
     TS_ASSERT_DELTA(-1.670937938, ySpOutX[npts / 2], 1e-08);
     TS_ASSERT_DELTA(17.99449408, ySpOutX.back(), 1e-08);
     // Y
-    TS_ASSERT_DELTA(-0.01152733, ySpOutY.front(), 1e-08);
-    TS_ASSERT_DELTA(5.56667697, ySpOutY[npts / 2], 1e-08);
-    TS_ASSERT_DELTA(-0.35141703, ySpOutY.back(), 1e-08);
+    TS_ASSERT_DELTA(0.48319776, ySpOutY.front(), 1e-08);
```
martyngigg: I'm a little confused as to why these values change. `ConvertToYSpace` doesn't use random numbers does it?

PranavBahuguna: The algorithm itself doesn't. However, the test makes use of `ComptonProfileTestHelpers::createTestWorkspace` which does make use of MersenneTwister, and thus it changes the contents of the workspace it uses for the algorithm. This in turn requires the test values to be changed.

martyngigg:
@PranavBahuguna Firstly thanks for putting the effort in to do this. 

I've had a chat with some other TSC members and given the numerical changes and now possible differences across platforms we are a little uncomfortable at pushing this through at the moment. We 
already unfortunately have a variations coming from other parts of the code and at the moment we'd be keen to keep this area more common.

I'm sure we will come back to this at some point and having seen the required changes we can make better decisions in the future so I don't see it has wasted effort.

PR #17613, opened 2016-09-27
Fix rename in a workspace group can cause duplicates

### LINKED ISSUE #17551: Renaming workspaces in a group allows two workspaces to share a single name
### Expected behavior
On renaming a workspace the `Override Existing` option should mean if you rename a workspace to an existing name in a group that is overridden
### Actual behavior
Two workspaces share a single name and Mantid enters an undefined state.
Run this script 
```python 
CreateSampleWorkspace(OutputWorkspace='bob', Function='User Defined', UserDefinedFunction='name=LinearBackground,A0=3.22542,A1=-0.0633193')
CreateSampleWorkspace(OutputWorkspace='bob1', Function='User Defined', UserDefinedFunction='name=LinearBackground,A0=3.22542,A1=-0.0633193')
CreateSampleWorkspace(OutputWorkspace='bob2', Function='User Defined', UserDefinedFunction='name=LinearBackground,A0=3.22542,A1=-0.0633193')
CreateSampleWorkspace(OutputWorkspace='bob3', Function='User Defined', UserDefinedFunction='name=LinearBackground,A0=3.22542,A1=-0.0633193')
GroupWorkspaces(InputWorkspaces='bob,bob1,bob2,bob3', OutputWorkspace='NewGroup')
RenameWorkspace(InputWorkspace='bob1', OutputWorkspace='bob')
```
If two workspaces called `Bob` exist in the group this bug is present
### Platforms affected
All

DavidFair:
Description of work.
This PR fixes an issue where two workspaces can have the same name if the target name already belongs to a group workspace. 
I've updated the notification to a before replace and moved it to before we update the ADS. However functionally the call still performs the same work with the additional step or removing any and all duplicates. 

Additionally it corrects and renames a few functions and unit tests to clarify their actual behaviour. 
**To test:**
There is a unit test added which exhibits this behaviour - WorkspaceGroupTest_testRenameWorkspaceDuplicates

Additionally running this script will confirm the problem is fixed if only 3 workspaces - "bob", "bob1" and "bob2" are printed in the final print statement. Additionally the renamed workspace should hold no data after renaming (thus ensuring it came from `bob1`

``` python
CreateSampleWorkspace(OutputWorkspace='bob', Function='User Defined', UserDefinedFunction='name=LinearBackground,A0=3.22542,A1=-0.0633193')
CreateSampleWorkspace(OutputWorkspace='bob1')
CreateSampleWorkspace(OutputWorkspace='bob2', Function='User Defined', UserDefinedFunction='name=LinearBackground,A0=3.22542,A1=-0.0633193')
CreateSampleWorkspace(OutputWorkspace='bob3', Function='User Defined', UserDefinedFunction='name=LinearBackground,A0=3.22542,A1=-0.0633193')
group = GroupWorkspaces(InputWorkspaces='bob,bob1,bob2,bob3', OutputWorkspace='NewGroup')

nameList = group.getNames()
print nameList

RenameWorkspace(InputWorkspace='bob1', OutputWorkspace='bob')

# Update the handle to the group
group = mtd["NewGroup"]
nameList = group.getNames()
print nameList
```

Fixes #17551 .

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/Kernel/inc/MantidKernel/DataService.h
```diff
@@ -310,28 +310,42 @@ template <typename T> class DLLExport DataService {
...
+      notificationCenter.postNotification(new BeforeReplaceNotification(
+          newName, newNameObject, oldNameWsObject));
+    }
+
+    datamap.erase(oldNameWsIter);
+
+    // Set the oldNameWsIter to end to ensure nobody invokes undefined behavior
+    oldNameWsIter = datamap.end();
+    // Call find again as iterators have been invalidated
+    newNameWsIter = datamap.find(newName);
```
martyngigg: I don't think these 2 lines are necessary. I think dereferencing an end() iterator is also undefined behaviour.

The second find should also not be necessary as http://en.cppreference.com/w/cpp/container/map/erase states that erase only invalidates iterators to the erased element and others remain intact.

Framework/Kernel/inc/MantidKernel/DataService.h
```diff
@@ -310,28 +310,42 @@ template <typename T> class DLLExport DataService {
...
+
+    if (newNameWsIter != datamap.end()) {
+      newNameWsIter->second = std::move(oldNameWsObject);
```
martyngigg: I think we should post the `AfterReplaceNotficiation` here to keep consistency. It should be possible to hook into before/after whether the delete happens here or in `::remove()`.

Framework/API/src/WorkspaceGroup.cpp
```diff
@@ -278,21 +279,36 @@ void WorkspaceGroup::workspaceDeleteHandle(
...
+ * Replaces a member if it was replaced in the ADS and checks
+ * for duplicate members within the group
+ * @param notice :: A pointer to a workspace before-replace notification object
  */
-void WorkspaceGroup::workspaceReplaceHandle(
+void WorkspaceGroup::workspaceBeforeReplaceHandle(
     Mantid::API::WorkspaceBeforeReplaceNotification_ptr notice) {
   std::lock_guard<std::recursive_mutex> _lock(m_mutex);
 
   const std::string replacedName = notice->objectName();
```
martyngigg: Can we improve this method a bit more by removing its dependency on names completely? A replace is just saying that essentially the "old" object is about the be deleted and replaced by this new one. I don't see a reason why we should query the names at all.

If we just check for matching pointers on the `oldObject` and the contents and remove any matching pointer then wouldn't that take care of everything?

martyngigg (changes requested):

Framework/API/src/WorkspaceGroup.cpp
```diff
@@ -278,21 +279,33 @@ void WorkspaceGroup::workspaceDeleteHandle(
...
+
   for (auto &workspace : m_workspaces) {
-    if ((*workspace).name() == replacedName) {
+    if (workspace == oldObject) {
       workspace = notice->newObject();
       break;
     }
   }
+
+  // Remove any duplicate pointers
```
martyngigg: I think we're nearly there with this now but I think there is a bug here.

The docs for [std::unique](http://en.cppreference.com/w/cpp/algorithm/unique) states that it only removes *consecutive* duplicates so if I updated your unit test as below to include a 3rd member

```
 void testRenameWorkspaceDuplicates() {
    // Tests that setting a workspace to the name of an existing
    // workspace in a group is correctly handled by the workspace group

    constexpr int nHist = 20, nBins = 10;
    // Register the workspace in the data service
    MatrixWorkspace_sptr work_in1 =
        WorkspaceCreationHelper::Create2DWorkspace123(nHist, nBins);
    MatrixWorkspace_sptr work_in2 =
        WorkspaceCreationHelper::Create2DWorkspace154(nHist, nBins);
    MatrixWorkspace_sptr work_in3 =
        WorkspaceCreationHelper::Create2DWorkspace154(nHist, nBins);

    const std::string ws1Name = "test_ws1";
    const std::string ws2Name = "test_ws2";
    const std::string ws3Name = "test_ws3";

    WorkspaceGroup_sptr wsSptr = WorkspaceGroup_sptr(new WorkspaceGroup);
    if (wsSptr) {
      AnalysisDataService::Instance().add("test_group", wsSptr);
      AnalysisDataService::Instance().add(ws1Name, work_in1);
      wsSptr->add(ws1Name);
      AnalysisDataService::Instance().add(ws2Name, work_in2);
      wsSptr->add(ws2Name);
      AnalysisDataService::Instance().add(ws3Name, work_in3);
      wsSptr->add(ws3Name);
    }

    TS_ASSERT_EQUALS(wsSptr->getNumberOfEntries(), 3);

    // Rename workspace 3 to 1
    AnalysisDataService::Instance().rename(ws3Name, ws1Name);

    TS_ASSERT_EQUALS(wsSptr->getNumberOfEntries(), 2);
    TS_ASSERT_EQUALS(wsSptr->contains(work_in3), true);
    TS_ASSERT_EQUALS(wsSptr->contains(work_in2), true);
    TS_ASSERT_EQUALS(wsSptr->contains(work_in1), false);

  }
```
then this test fails. I would suggest updating the unit test to include a 3rd member like this. 

It looks like you'd need to sort the vector for `std::unique` to work as we would want here. I thinking sorting shouldn't be a problem here.

I think it is also more idiomatic to use `std::unique` coupled with `std::erase` rather than `resize` as in the link above.

Framework/Algorithms/test/WorkspaceGroupTest.h
```diff
@@ -296,6 +296,39 @@ class WorkspaceGroupTest : public CxxTest::TestSuite {
     AnalysisDataService::Instance().remove("InputWS");
   }
 
+  void testRenameWorkspaceDuplicates() {
```
martyngigg: This is a great test to add but can we also add a separate test for renaming of a workspace that is part of a group to a name that doesn't already exist in the group? That just covers the other use case here then.

martyngigg (changes requested):

Framework/API/src/WorkspaceGroup.cpp
```diff
@@ -278,21 +279,32 @@ void WorkspaceGroup::workspaceDeleteHandle(
...
+  std::sort(m_workspaces.begin(), m_workspaces.end());
+  auto endIter =
+      std::unique(m_workspaces.begin(), m_workspaces.end(),
```
martyngigg: Excellent. This is looking good now.

I have one final comment. Do we need the lambda here? I thought as `shared_ptr` defines `operator==` then the default comparison of `unique` would suffice?

If that is the case I think we could reduce this to:
```c++
m_workspaces.erase(std::unique(m_workspaces.begin(), m_workspaces.end()), 
                                   m_workspaces.end);
```

DavidFair: Good point - I was originally getting the raw pointers and forgot to remove that lambda when I realised it wasn't needed.
I am just figure out what is causing the unit tests below to fail and then this should be pretty much there unless there is any other cases either of us can think of?

martyngigg:
@DavidFair I'm guessing the sorting has caused problems with some tests expecting a certain order

DavidFair:
Yes quite a few tests do
`wsGroupPtr->getAt(1)` and this messes up their order.

I'm thinking could we use something like an `unordered_map` and store the shared pointers as keys. If we hit a key that already exists we remove it from the vector. This preserves ordering and means we can consider non-consecutive keys

martyngigg:
What would you store as the value in the map?

DavidFair:
Just something generic like a bool as it takes the minimum amount of space.
So it would look like this in pseudo code:

```
for (ptr : groupMembers){
   if (duplicateMap.find(ptr) != duplicateMap.end()){
       // Remove duplicate
   } else {
       duplicateMap.insert(std::make_pair<...>(ptr, false));
   }
}
```
This means gives an amortized insert and lookup or `O(1)` and worst case `O(n)` so shouldn't add significant overhead.

martyngigg:
What about `unordered_set`?

Framework/API/src/WorkspaceGroup.cpp
```diff
@@ -278,21 +279,35 @@ void WorkspaceGroup::workspaceDeleteHandle(
...
  */
-void WorkspaceGroup::workspaceReplaceHandle(
+void WorkspaceGroup::workspaceBeforeReplaceHandle(
     Mantid::API::WorkspaceBeforeReplaceNotification_ptr notice) {
   std::lock_guard<std::recursive_mutex> _lock(m_mutex);
 
-  const std::string replacedName = notice->objectName();
+  const auto oldObject = notice->oldObject();
+
   for (auto &workspace : m_workspaces) {
```
martyngigg: Can we do this whole replace and remove unique in a single loop? It seems a shame to have to loop over the thing twice.

martyngigg: If we had an additional `else if(workspace == newObject)` and then stored an iterator to that position we could erase that single position after the loop? We'd need to take the `break` out but it would at least avoid more allocations by having to loop over and create a set each time.

DavidFair:
WRGT using the STL instead of boost:
As we use `boost:shared_ptr` the STL will not hash them and has a static assert at compile time to detect this so unfortunately we need to use the boost implementation of an unordered set.

martyngigg:
Ah I see. Out of curiosity is it possible to add some function definition that makes `boost::shared_ptr` hashable? I know you can do that in Qt with `QHashMap`.

What do you think of the other suggestion for doing everything in the loop once?

DavidFair:
I've changed the loop to the following to do everything at one:
```
  for (auto it = m_workspaces.begin(); it != m_workspaces.end();) {
    if (*it == oldObject) {
      // Found the object to replace
      (*it) = notice->newObject();
      seenWorkspaces.insert(*it);
      ++it;
    } else if (seenWorkspaces.find(*it) != seenWorkspaces.end()) {
      // Duplicate
      it = m_workspaces.erase(it);
    } else {
      seenWorkspaces.insert(*it);
      ++it;
    }
  }
```
We can override the hash method like [so](http://stackoverflow.com/a/18079061) and this should make it work for the STL.

Framework/API/src/WorkspaceGroup.cpp
```diff
@@ -278,21 +279,43 @@ void WorkspaceGroup::workspaceDeleteHandle(
...
-  for (auto &workspace : m_workspaces) {
-    if ((*workspace).name() == replacedName) {
-      workspace = notice->newObject();
+  const auto oldObject = notice->oldObject();
+  const auto newObject = notice->newObject();
+
+  bool foundOld(false);
+  bool foundDuplicate(false);
+
+  auto it = m_workspaces.begin();
```
martyngigg: I have 1 final picky thing here. The scope of `it` can be reduced as it's not needed outside of the `for` loop:

`for (auto it = m_workspaces.begin(); it != m_workspaces.end(); ++it)`

This means that `it` cannot accidentally used outside of the loop.

martyngigg (approved):

martyngigg:
This looks good now in my opinion.

Assuming OSX passes, :shipit:

PR #17621, opened 2016-09-28
Add special options for tiled plots of fit results

### LINKED ISSUE #17602: Beta test feedback: Muon Analysis
*Feedback from beta testers for Mantid 3.7.20160922.1557, Windows*

**Higher priority** (aim to fix these for release)
- [x] *Fit to raw data* option doesn't work (most important point to fix) - #17605
- [x] Forgets selected groups on changing tab - #17619
- [x] Multilayer plot (tiled plot) - recognise muon fits and apply 'special' options - #17621
  - Data as points with errors, fit as line without points
- [x] Default label to run number - #17619

**Normal priority** (push back to 3.9)
- [x] "Sequential fit" checks the simultaneous fit label and warns if it's already used. But this label won't be used anyway, so there's no need to check it - #17619
- [ ] Automatically plot the tiled plot in the plot window after a simultaneous fit (may have to wait for #17511)
- [ ] Multiple results table: think of a better header than *f0*, *f1* for datasets
- [ ] Setting tie *f0.Phi=f1.Phi* caused a crash once (not second time) - try to replicate?
- [ ] Error message when entering a backward run range in sequential fit dialog e.g. 12345-43 (this is probably right, maybe the message should just be more descriptive)
- [ ] Check where start time is set

tom-perkins:
_Small change requested by beta testers - they would like this to make the release_

New in Mantid 3.8 is the ability to create a multi layer plot from right-clicking a workspace group.
For the case when this is fit results, scientists would like the plot to automatically have the data as points and the fit curve as a line (no points).

**To test:**

_Data files: `\\olympic\babylon5\Scratch\TomPerkins\test_workspaces`_
- Load the file `multi_plot_test_workspace_Fit.nxs`, you should get a workspace group
- Right-click on the group, choose "Plot Spectrum"
- In the dialog, tick "Tiled Plot" and select "Plot All" 
- Should get a plot where the data curve is points (with a line) and the fit curve is just a line. 
  - The size of the points depends on your default plot settings (_View/Preferences/2D Plots/Curves_)
  - The "difference" curve is left as default, so will also depend on your default settings
  - To get errors, can choose "Plot Spectrum with Errors"
- Now load the second file, `multi_plot_test_workspace_notFit.nxs`. This is the same data, but the axes have been renamed so it doesn't look like fit results.
- Make the same plot. This time, all curves should be in the default style - no special options.

Fixes #17602. (The rest of the points in that issue will be pushed back to 3.9.)

No release notes - this feature is new to Mantid 3.8 and is already in the release notes.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidPlot/src/Mantid/MantidUI.cpp
+++ b/MantidPlot/src/Mantid/MantidUI.cpp
@@ -130,6 +130,42 @@ bool drawXAxisLabel(const int row, const int col, const int nRows,
     return false;
   }
 }
+
+/// Spectra names for a fit results workspace
+const std::vector<std::string> FIT_RESULTS_SPECTRA_NAMES{"Data", "Calc",
+                                                         "Diff"};
+
+/// Decide whether the named workspace is the results from a fit
+/// (will have 3 spectra called "Data", "Calc" and "Diff")
+bool workspaceIsFitResult(const QString &wsName) {
+  bool isFit = false;
+  const auto &ws = AnalysisDataService::Instance().retrieveWS<MatrixWorkspace>(
+      wsName.toStdString());
+  if (ws) {
+    if (FIT_RESULTS_SPECTRA_NAMES.size() == ws->getNumberHistograms()) {
+      std::vector<std::string> spectraNames;
+      const auto specAxis = ws->getAxis(1); // y
+      for (size_t iSpec = 0; iSpec < FIT_RESULTS_SPECTRA_NAMES.size();
+           ++iSpec) {
+        spectraNames.push_back(specAxis->label(iSpec));
+      }
+      isFit = spectraNames == FIT_RESULTS_SPECTRA_NAMES;
+    }
+  }
+  return isFit;
+}
+
+/// Return curve type for spectrum of a set of fit results
+Graph::CurveType getCurveTypeForFitResult(const size_t spectrum) {
+  switch (spectrum) {
+  case 0:
+    return Graph::CurveType::LineSymbols;
+  case 1:
+    return Graph::CurveType::Line;
+  default:
+    return Graph::CurveType::Unspecified;
+  }
+}
 }
 
 MantidUI::MantidUI(ApplicationWindow *aw)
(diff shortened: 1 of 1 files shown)


KarlPalmen:
I see that the difference between the two sets of plots 2x2 tiling is that `Fit` has the black curve named `data`, the red curve named `calc` and the green curve named `diff`. In the `notFit` they of `0`, `1`, `2` respectively indicating workspace indices. This seems to be the intended result of the test. Is it?

KarlPalmen:
I don't see a unit test from these new spectra names.

tom-perkins:
I renamed the curves in the `notFit` workspace so it would not be recognised as a fit (the criteria for the workspace being a fit result are that it should be a matrix workspace with 3 spectra called "Data", "Calc" and "Diff").

The difference should be:
*Fit workspace*
Black curve is line with points, red curve is line **without** points, green is default
*Non-fit workspace*
All three curves are plotted according to default settings.

There are no tests for `MantidUI` unfortunately, I don't think one can be added here.

KarlPalmen:
I see that the red `Calc` curve has no dots on it and the other 5 curves over both plots have dots.

tom-perkins:
That sounds correct, assuming that your default plot setting (*View/Preferences/2D Plots/Curves/Default curve style*) is "Line + Symbol"

KarlPalmen:
It seems I'll need to repeat the test with different settings of *View/Preferences/2D Plots/Curves*.

KarlPalmen:
Plotting after setting my default curves to be smooth without any dots (points) makes only `Data` has dots, which is in accordance to your description. 

Given that its not feasible to add unit test now :shipit: .

martyngigg:
I'm not too comfortable with this going into the release, at least in this form. It doesn't seem like a bug fix. It alters the behaviour of something that wider groups of users would have seen.

I think the tiled plot option is a great extension but having it operate slightly different in some circumstances with no way to change it seems like it could confuse people. I'd be more comfortable if we could at least have a way of controlling it maybe?

@NickDraper Any thoughts?

NickDraper:
Given that this was a request out of the muon group, and it is additional functionality (not a change to something that was present before) I would be tempted to include it, we can look to standardize the special nature of it in a future release, or add the ways to control it then.

martyngigg (approved):

martyngigg:
True, it is new in this release. I'll put it in.

PR #17626, opened 2016-09-28
Fix calling `GetEiMonDet` from Python when using workspace indices

### LINKED ISSUE #17623: Fix the type of  *Monitor* input property in `GetEiMonDet`
### Expected behavior
One should be able to execute the `GetEiMonDet` algorithm from Python using workspace indices as *IndexType*.

### Actual behavior
C++ expects to get the property value as `status_t` type, which is **unsigned** and 64 bits long on a 64 bit system --- this does not play well with Python which by default deals with **signed** integers.

```python
import numpy

x = numpy.tile(numpy.arange(-10.0, 10.1, 0.5), 2)
y = numpy.tile(numpy.zeros(len(x) - 1), 2)
y[len(y) / 2 - 2: len(y) /  2 + 2] = 1 # Fake elastic peak
ws = CreateWorkspace(x, y, NSpec=2, UnitX='TOF')
EditInstrumentGeometry(Workspace=ws,
                       PrimaryFlightPath=0.01,
                       L2=[0.01, 0.1],
                       Polar=[180, 0],
                       InstrumentName='Omnium')

eppTable = FindEPP(InputWorkspace=ws)
GetEiMonDet(DetectorWorkspace=ws,
            DetectorEPPTable=eppTable,
            IndexType='WorkspaceIndex',
            Detectors='1',
            Monitor='0',
            PulseInterval=22.0,
            NominalIncidentEnergy=200.0)
```
The above script will throw an error:
```
RuntimeError: Attempt to assign property Monitor to incorrect type. Expected type uint64_t
  at line 20 in '/home/cs/soininen/Sources/scripts/GetEiMonDet2_index_failure.py'
  caused by line 866 in '/users/soininen/mantid-build/build-Debug/bin/mantid/simpleapi.py'
```

soininen:
When using 'WorkspaceIndex' as _IndexType_ in `GetEiMonDet`, the algorithm casts the _Monitor_ property directly to a `size_t`. This does not work when the algorithm is called from Python since `size_t` is unsigned and Python can by default only pass signed integers. This PR now fixes the problem and also introduces a new sanity check (_Monitor_ cannot be negative if it is a workspace index).

**To test:**

Running

``` python
import numpy

x = numpy.tile(numpy.arange(-10.0, 10.1, 0.5), 2)
spectrum = numpy.zeros(len(x) - 1)
spectrum[len(spectrum) / 2 - 2: len(spectrum) /  2 + 2] = 1 # Fake elastic peak
y = numpy.tile(spectrum, 2)
ws = CreateWorkspace(x, y, NSpec=2, UnitX='TOF')
EditInstrumentGeometry(Workspace=ws,
                       PrimaryFlightPath=0.01,
                       L2=[0.01, 0.1],
                       Polar=[180, 0],
                       InstrumentName='Omnium')

eppTable = FindEPP(InputWorkspace=ws)
GetEiMonDet(DetectorWorkspace=ws,
            DetectorEPPTable=eppTable,
            IndexType='WorkspaceIndex',
            Detectors='1',
            Monitor='0',
            PulseInterval=22.0,
            NominalIncidentEnergy=200.0)
```

before applying these changes will produce an error. After applying the changes, the above script should run without problems.

Fixes #17623 

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/Algorithms/src/GetEiMonDet2.cpp
```diff
@@ -425,7 +425,11 @@ void GetEiMonDet2::parseIndices(std::vector<size_t> &detectorIndices,
         spectraListParser.parse(getProperty(PropertyNames::DETECTORS)));
     auto back = std::back_inserter(detectorIndices);
     std::copy(detectors.begin(), detectors.end(), back);
-    monitorIndex = getProperty(PropertyNames::MONITOR);
+    int monitor = getProperty(PropertyNames::MONITOR);
+    if (monitor < 0) {
```
mantid-roman: You should set a BoundedValidator on the property.

soininen: Yes, of course. Somehow I was thinking that somebody might be mad enough to use negative detector id or spectrum number here. I guess that will never happen and validating the *Monitor* property makes the check here superfluous.

mantid-roman: Sorry, I didn't realise (by looking only at the code changes in the PR) that the check was being applied to the case of a workspace index only. What I also didn't see was code duplication: properties DETECTORS and MONITOR get read 3 times.

To be honest I don't know whether a det id can be negative or not. May I ask you to refactor the method and get the properties only once before the `if` and process them afterwards. If you are also unsure about the sign of the IDs please revert your last change and do the check for workspace indices only.

What I also noticed is that the DETECTORS property is a string. Do you know if there is a reason for that? I think `ArrayProperty<int>` suites better to it.

mantid-roman:
There is function template `mapIndices<...>` in the `.cpp` file. It should be placed into an anonymous namespace.

Framework/Algorithms/src/GetEiMonDet2.cpp
soininen: The code duplication was partially due to some type magick --- I think I was trying to be too smart here. Moving the `getProperty()` calls out of the `if`s does indeed simplify the code a bit.

The reason for the string *Detectors* property is that since the string is parsed by `UserStringParser`, users can specify complex ranges like `"1, 10-15, 78, 88, 100-110"`.

mantid-roman: `ArrayProperty` can do exactly the same. Please have a look in `ArrayPropertyTest.h` for examples.

mantid-roman: Whichever way you prefer. If you decide to open a new issue please refer its number in this PR.

soininen: Ok, I'll handle the switch as a separate issue #17702

mantid-roman:
The check-list just reminded me that there needs to be a unit test that checks the response on a negative `Monitor`

mantid-roman:
I think we are done :shipit:

Issue #17641, opened 2016-09-30
Umbrella: HistogramData rollout during maintenance period


SimonHeybrock:
Here is a link on tips for refactoring the legacy Histogram interface:
https://github.com/mantidproject/documents/blob/e83deb6201eac7b0ff1471ea3f2520b054f8f212/Help/HistogramData/HistogramData-refactoring.md

Please pick a chunk of 5 to 10 algorithm at a time:
1. Pick algorithms from the **top**.
2. Write a comment on this issue, listing the algorithms you picked.
3. Make sure no one picked the same algorithm while writing the comment (refresh page).
4. Create an issue, start refactoring, and create a PR. See below for a template for copy and paste to make creating the issue quicker.
5. Do not hesitate to create the PR early for an intermediate review, or contact Simon Heybrock or Lamar Moore on Slack when in doubt or when anything is unclear.

Issue template:

```
HistogramData rollout: algorithms <X to Y>

This issue is part of the HistogramData rollout, see https://github.com/mantidproject/mantid/issues/17641.
Algorithms covered in this issue are:
```

PR template (edit as needed):

```
HistogramData rollout: algorithms <X to Y>

## Description of work.
This PR is part of refactoring effort to use the new interface of the HistogramData module.
See https://github.com/mantidproject/mantid/issues/17641 for details.

The following algorithms were refactored to use HistogramData:

Performance tests were added for:

The following algorithms have improved performance:

## To test.

Code review.

<Add any manual testing not covered by unit tests, especially GUIs>

Fixes <your issue>.
Internal change, no release notes (potential performance improvements will be added to release notes as part of the umbrella issue).
```

Algorithms that use `(data|read)(X|Y|E)`:

I am listing the `cpp` files here for convenience, but you are also responsible for the corresponding unit tests (and header files):

*All algorithms complete*

Already assigned / done:

```
Framework/Examples/ModifyData.cpp
Framework/Examples/WorkspaceAlgorithm.cpp
Framework/Geometry/inc/MantidGeometry/Instrument/Component.h
Framework/Geometry/src/ComponentParser.cpp
Framework/Geometry/src/Instrument/Component.cpp
Framework/Algorithms/src/PDCalibration.cpp
Framework/Algorithms/src/PDFFourierTransform.cpp
Framework/Algorithms/src/PhaseQuadMuon.cpp
Framework/Algorithms/src/PlotAsymmetryByLogValue.cpp
[...]

PR #17674, opened 2016-10-03
Algorithms 11 to 15 updated with Histogram Data

### LINKED ISSUE #17661: HistogramData rollout: algorithms 11 to 15
HistogramData rollout: algorithms

This issue is part of the HistogramData rollout, see #17641.

```
Framework/Algorithms/src/RadiusSum.cpp
Framework/Algorithms/src/RayTracerTester.cpp
Framework/Algorithms/src/ReadGroupsFromFile.cpp
Framework/Algorithms/src/RealFFT.cpp
Framework/Algorithms/src/Rebin2D.cpp
```

dtasev:
HistogramData rollout: algorithms

This issue is part of the HistogramData rollout, see #17641.

```
Framework/Algorithms/src/RadiusSum.cpp
Framework/Algorithms/src/RayTracerTester.cpp
Framework/Algorithms/src/ReadGroupsFromFile.cpp
Framework/Algorithms/src/RealFFT.cpp
Framework/Algorithms/src/Rebin2D.cpp
```

Fixes #17661


Framework/Algorithms/src/RadiusSum.cpp
```diff
@@ -148,12 +149,12 @@ std::vector<double> RadiusSum::processNumericImageRadiusSum() {
   // or the center of the bin, if it is a histogram.
 
   g_log.debug() << "Define the X positions of the pixels\n";
-  auto &refX = inputWS->getSpectrum(0).readX();
-  auto &refY = inputWS->getSpectrum(0).readY();
+  auto &refX = inputWS->getSpectrum(0).x();
+  auto &refY = inputWS->getSpectrum(0).y();
   std::vector<double> x_pos(refY.size());
```
SimonHeybrock: I think you can simply use `auto x_pos = inputWS->points(0);` here and get rid of the `if/else` block below.

Framework/Algorithms/src/RadiusSum.cpp
```diff
@@ -619,14 +620,14 @@ void RadiusSum::setUpOutputWorkspace(std::vector<double> &values) {
       inputWS, 1, values.size() + 1, values.size());
 
   g_log.debug() << "Set the data\n";
-  MantidVec &refY = outputWS->dataY(0);
-  std::copy(values.begin(), values.end(), refY.begin());
+  auto &refY = outputWS->mutableY(0);
+  std::copy(values.cbegin(), values.cend(), refY.begin());
```
SimonHeybrock: `operator=` should work here and is shorter.

Framework/Algorithms/src/RealFFT.cpp
```diff
@@ -106,29 +106,34 @@ void RealFFT::exec() {
...
+      y3[i] = dx * sqrt(re * re + im * im); // modulus
     }
     if (inWS->isHistogramData()) {
-      outWS->dataX(0)[yOutSize] = outWS->dataX(0)[yOutSize - 1] + df;
+      outWS->mutableX(0)[yOutSize] = outWS->mutableX(0)[yOutSize - 1] + df;
     }
-    outWS->dataX(1) = outWS->dataX(0);
-    outWS->dataX(2) = outWS->dataX(0);
+    outWS->mutableX(1) = outWS->mutableX(0);
+    outWS->mutableX(2) = outWS->mutableX(0);
```
SimonHeybrock: Share?

Framework/Algorithms/src/RealFFT.cpp
```diff
@@ -173,13 +181,14 @@ void RealFFT::exec() {
...
       double x = df * i;
-      outWS->dataX(0)[i] = x;
-      outWS->dataY(0)[i] = data[i] / df;
+      xData[i] = x; // xData used from above
+      y[i] = data[i] / df;
     }
     if (outWS->isHistogramData())
-      outWS->dataX(0)[yOutSize] = outWS->dataX(0)[yOutSize - 1] + df;
+      outWS->mutableX(0)[yOutSize] = outWS->mutableX(0)[yOutSize - 1] + df;
     // outWS->getAxis(1)->spectraNo(0)=inWS->getAxis(1)->spectraNo(spec);
```
SimonHeybrock: I think these 10 lines could be written in 2 using `std::generate` with `HistogramData::LinearGenerator` for X, and `operator/=(double)` for scaling Y.

Framework/Algorithms/src/Rebin2D.cpp
```diff
@@ -126,22 +128,23 @@ void Rebin2D::exec() {
   {
     PARALLEL_START_INTERUPT_REGION
 
+    m_progress->report("Computing polygon intersections");
```
SimonHeybrock: You moved this out of the inner loop (for performance?), but if you do so shouldn't you also adapt the `nreports` variable that gives the length of the progress bar?

DimitarTasev: @SimonHeybrock Yes, I moved it out because the algorithm was spending most of its time writing to std out, the execution speed went down from 160ms to 40ms. 

Is it better to:
- Add a `m_progress->report()` in the inner loop, so that the progress bar scales at the same rate, I tested it and the performance is not significantly affected if there is no string in report('...')
or
- Change size of `nreports` to be the same as `numYbins' (`inputWS->getNumberHistograms()`

SimonHeybrock: Great improvement 👍 .
I guess the bar still updates at a decent rate with you latter suggestion? Then I'd go for that.

Framework/Algorithms/src/RadiusSum.cpp
```diff
@@ -619,14 +620,14 @@ void RadiusSum::setUpOutputWorkspace(std::vector<double> &values) {
...
 
   g_log.debug() << "Set the bins limits\n";
-  MantidVec &refX = outputWS->dataX(0);
-  double bin_size = (max_radius - min_radius) / num_bins;
+  auto &refX = outputWS->mutableX(0);
 
-  for (int i = 0; i < (static_cast<int>(refX.size())) - 1; i++)
+  double bin_size = (max_radius - min_radius) / num_bins;
+  for (int i = 0; i < (static_cast<int>(refX.size())) - 1; ++i)
```
DimitarTasev: @SimonHeybrock Is it possible to use `LinearGenerator` here?

SimonHeybrock: Yes.

Framework/Algorithms/src/RealFFT.cpp
DimitarTasev: @SimonHeybrock Is there a way to do the `y[i] = data[i] / df;` in a single line, what I have currently is:
```
y = *(newData.get()); // because dereferencing shared_array
y /= df;
```

SimonHeybrock: @DimitarTasev Why not `outWS->mutableY(0) /= df;`?

SimonHeybrock: Wait, that is not what you need...

SimonHeybrock: @DimitarTasev Why does this code need `data` to be a `boost::shared_array<double>`? Can't you just directly let the code above modify the data via `outWS->mutableY` and then just divide by `df` in the end?

DimitarTasev: @SimonHeybrock I think it was done because the gsl functions take a `double data[]`, I'll try if it can work with vector instead

DimitarTasev: @SimonHeybrock The problem now is that the `gsl` functions need a non-const `double data[]`. I can't think of another way of doing it rather than constructing a new vector, maybe I can use the counts instead?

SimonHeybrock: @DimitarTasev `&(ws->mutableY(i)[0])`?

Framework/Algorithms/src/RadiusSum.cpp
```diff
@@ -619,16 +613,16 @@ void RadiusSum::setUpOutputWorkspace(std::vector<double> &values) {
...
+
+  auto xSize = outputWS->mutableX(0).size();
   double bin_size = (max_radius - min_radius) / num_bins;
+  outputWS->setBinEdges(0, xSize,
+                        HistogramData::LinearGenerator(min_radius, bin_size));
 
-  for (int i = 0; i < (static_cast<int>(refX.size())) - 1; i++)
-    refX[i] = min_radius + i * bin_size;
-  refX.back() = max_radius;
+  outputWS->mutableX(0).back() = max_radius;
```
DimitarTasev: @SimonHeybrock Do you think this is not needed anymore because of `LinearGenerator`, I ran the tests with both and it passes with both, but decided to leave it just in case, as it seems that was the intention before?

SimonHeybrock: Well, I think the reason it is there is to avoid floating point inaccuracies. With exact arithmetic it would make no difference. In finite precision there will be a very small difference. In my opinion difference is negligible (as confirmed by your tests) --- if this finite precision error would make a difference anywhere it would indicate a much larger issue somewhere.

So I'd remove it.

SimonHeybrock:
:shipit:

PR #17691, opened 2016-10-04
Fix "next/previous run" co-adding run instead of replacing existing

### LINKED ISSUE #17688: "Next file" button co-adds instead of replacing
*Mantid 3.7.20161003.2026, noticed during testing* 

#### Expected behaviour
"Next/Previous run" buttons load the next/previous run, replacing the currently loaded run

#### Actual behaviour
The next/previous run is loaded and co-added to the currently loaded run
N.B. This only happens when the run was selected with the "Browse" button. It works correctly when you type the run number in and then use the next/previous buttons.

- Open *Interfaces/Muon/Muon Analysis*
- Use the Browse button to load a file (NOT by typing the run number in)
- Use the Next or Previous button to load the next run
- Run is co-added rather than replacing the existing one

tom-perkins:
Fixed a bug where the "load next/previous" would co-add the new run to the currently loaded run instead of replacing it in certain cases (where the file path contained the `-` character).

This is a small fix and it would be good to fix it for the release. However, the bug was present in previous releases, is not major and has not been seen by users, so it can be pushed back if necessary.

No test - the fix is in the GUI (untested) part of the MuonAnalysis interface.

**To test:**

Download the folder `\\olympic\babylon5\Scratch\TomPerkins\MUSR-test` and add it to Mantid's data path. These files are in the unit test data, but the important point is that the folder's name contains the character `-`.
- Open _Interfaces/Muon/Muon Analysis_
- Use the Browse button to load one of these files (do not load it by typing the run number in as this doesn't cause the bug to appear)
- Click 'next' or 'previous'
- Next or previous run should replace the currently loaded run, rather than being co-added to it.
  load
- To test that co-adding runs still works when entering a range: type `15189-91` in the box and hit enter as below. Runs should be loaded and added together. If you click "next" it will extend the range to `00015189-00015192`.
  load2

Fixes #17688 

[Release notes](https://github.com/mantidproject/mantid/blob/425ca33ca649b481c2da4455edd5730e76c07f6b/docs/source/release/v3.9.0/muon.rst) updated. If pushed back to 3.9, the note needs to be moved to the new set of release notes.

---
#### Reviewer

##### Code Review
##### Functional Tests


KarlPalmen:
The test does not include any tests for co-adding when it is required.

tom-perkins:
:+1: Tests updated above

KarlPalmen:
:shipit:

martyngigg:
@tom-perkins It's great that we spotted and fixed this. As the Muon scientists have already done their beta testing and the release date is very close I think I would like to air on the side of caution and push this back to `3.9` since there are no unit tests around this area.

Could you make the necessary updates to this PR to put it back to `3.9`?

tom-perkins:
Merged in master, moved release notes and changed base branch. There are a couple of unrelated commits included in this PR as a result of this process.

KarlPalmen:
I'm running the test again in case the merging in of master has changed anything.

KarlPalmen:
It still works well :shipit:

KarlPalmen:
Testing again

KarlPalmen:
I works well :shipit:

PR #17711, opened 2016-10-05
Fix display of data in Splatterplot View (VSI)

### LINKED ISSUE #17640: No data shown in scatter plot mode
Currently nothing is shown when loading an event workspace and opening VSI. Swapping between views also does not work.

### Expected behavior
Data should be shown.

### Actual behavior
Data is not shown.

Observed using ```MDEvent_ellipPeaks_new``` which @AntonPiccardoSelg has a copy of.

### Platforms affected
All.

AntonPiccardoSelg:
Fixes #17640

The opacity is set to 0.5 by default in the SplatterPlot mode. Apparently there is some issue with the depth peeling in PV5.1.0 when more than view is present. The same faulty behaviour we are seeing in the splatterplot view can be re-created in PV when a data set is opened in two views. 

Setting the opacity to 1.0 avoids any depth peeling and displays the data correctly in the splatterplot view. For now this is probably the easiest solution. 

I still need to verify this, but I think that PV5.1.2 has the fix for this issue. See [here](https://blog.kitware.com/paraview-5-1-2-release-notes/). If this is the case we should switch to this version and reset the old opacity default value.

**To test:**
1. Open the an MDEventWorkspace in the VSI and switch to the splatterplot
   - Confirm that the data is displayed
2. Switch to another view and back
   - Confirm that the data is still being displayed
3. Set the opacity control to anything other than 1.0
   - Confirm that the rendered image disappears
4. Set it back to 1.0
   - Confirm that the rendered image reappears

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Vates/VatesSimpleGui/ViewWidgets/src/SplatterPlotView.cpp
+++ b/Vates/VatesSimpleGui/ViewWidgets/src/SplatterPlotView.cpp
@@ -193,7 +193,7 @@ void SplatterPlotView::render() {
       builder->createDataRepresentation(src->getOutputPort(0), this->m_view);
   vtkSMPropertyHelper(drep->getProxy(), "Representation").Set(renderType);
   if (!isPeaksWorkspace) {
-    vtkSMPropertyHelper(drep->getProxy(), "Opacity").Set(0.5);
+    vtkSMPropertyHelper(drep->getProxy(), "Opacity").Set(1.0);
     vtkSMPropertyHelper(drep->getProxy(), "GaussianRadius").Set(0.005);
   } else {
     vtkSMPropertyHelper(drep->getProxy(), "LineWidth").Set(2);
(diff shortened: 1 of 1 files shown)


martyngigg:
@quantumsteve Is it worth upgrading to PV 5.1.2 now or is that a bad idea close to the release?

quantumsteve:
@martyngigg v5.1.2 is a minor change and shouldn't require any changes in Mantid. I'd prefer upgrading ParaView over this workaround.

martyngigg:
I'm just building `5.1.2` now to verify it fixes the issue.

martyngigg:
I've built `5.1.2` and rebuilt mantid against it. I now see the same image as I did with `3.7.2` on my test data set.

martyngigg:
Shall we get 5.1.2 building on all of the machines?

martyngigg:
#17718 updates Mantid to build against ParaView 5.1.2 so the fix here should no longer be needed.

PR #17725, opened 2016-10-06
IndirectDataAnalysis - Save Ascii files

### LINKED ISSUE #16419: Indirect Data Reduction Energy Transfer - Load Ascii files
### Expected behavior
* Tosca data Files saved to Ascii with the functionality in the Indirect Data Reduction Energy Transfer can be loaded back into Mantid using LoadAscii

### Actual behavior
* Once a file has been saved to Ascii Mantid is unable to load it back in 

* Save TOSCA data as ascii using the Energy Transfer interface 

### Platforms affected
* All

**This is caused by TOSCA producing 7 column data and LoadAscii can only deal with 1 and 3**

louisemccann:
The SaveAscii algorithm used in Indirect > Data Reduction> Energy Transfer has been updated to version 2 to enable easier re-loading of saved ascii files back into mantid.

**To test:**

Indirect > Data Reduction > Energy Transfer
Ensure default save directory is set
Ensure Instrument is set to `TOSCA`
Run Files: `TSC15352.raw`
Rebin String: `-2.5,0.015,3,-0.005,1000`
Press "Run"
Ensure `ASCII` checkbox is ticked
Press "Save"
Close interface, delete `tosca15352-TO96_2` workspace

Press `Ctrl`+`Shift`+`F`
Browse to default save directory
Select `tosca15352-TO96_2.dat`
Ensure file loads properly

Fixes #16419.

[Release Notes](https://github.com/mantidproject/mantid/blob/51225396354376c770de652e2cfdaa639e165210/docs/source/release/v3.9.0/indirect_inelastic.rst)

---
#### Reviewer

##### Code Review
##### Functional Tests


reimundILL:
It is not part of the issue, but you may have a look into the following example (using python format)

```Python

from __future__ import (absolute_import, division, print_function)

# old
logger.information('Workspace %s need data chop: %s' % (ws_name, str(chopped_data)))
# new and RHEL6 compatible
logger.information('Workspace {0} need data chop: {1}'.format(ws_name, str(chopped_data)))
```

Also only as information, there is in line 15 FIle instead of File (since you found many other mistakes, which I like).

In my opinion, all Version 1 related comments should be deleted (unuseful)

``` Python
# Version 1 of SaveAscii produces output that works better with excel/origin
# For some reason this has to be done with an algorithm object, using the function
# wrapper with Version did not change the version that was run
```

Finally, I think you are not using Indirect > Data Analysis but Indirect > Data Reduction?
However, also release notes say it is Indirect > Data Analysis.

Only in Indirect > Data Reduction > ISIS Energy Transfer I can select save > ASCII.

I'll start testing then.

reimundILL:
I run tests with unitttest data TSC10076.raw:
when I reload the saved ASCII file and chose Unit Energy_inWavenumber
I run algorithm CompareWorkspace and get the message: Histogram/point-like mismatch.
I see that X-values differ, and one workspace has 1532 bins, the other 1531 (this was the reloaded one). So, one time there are Histogram data, one time a distribution.
I do not know if this is correct behavior, I would expect the same workspaces.

louisemccann:
@reimundILL Thank you for the suggestions. I think the Histogram/point-like mismatch is correct behaviour for SaveAscii - the midpoint of the bin is stored rather than the bin edges.

reimundILL:
@louisemccann Everything seems to look good. Smart solution found. Maybe the ASCII saving can be described in the documentation (user expectation - may not know about SaveAscii properties). It's true that users may prefer a matrix when they use other software for postprocessing or visualisation (pgfplots, MatLab, Origin, vtk, Excel?). I think the python 3 thing may be nice for your future code, since you only changed code in one place and not the many other places (another issue). :shipit:

PR #17732, opened 2016-10-06
Fix Mantid crash when deleting many workspaces in MantidDock

### LINKED ISSUE #17729: Fix Mantid crash when closing many workspaces via MantidDock
Rob mentioned that closing about 200 workspaces at once via MantidDock causes a crash. 

**Solution**
Avoid deleting the workspaces via the `DeleteWorkspace` algorithm and clear them instead directly from the ADS.

AntonPiccardoSelg:
Fixes #17729

We delete the workspaces from the MantidDock directly in the ADS without the indirection of the algorithm.

**To test:**

*\* Test that deleting many workspaces does not crash**
1. Create many wokspaces with:

``` python
for i in range(3000):
    name = "WS_" + str(i)
    CreateSampleWorkspace(OutputWorkspace=name)
```
1. Select all in the ADS
2. Press the delte button
   - Confirm that Mantid does not crash and that the time it takes to clear the 3000 workspaces is  acceptable.

*\* Test that deleting few very large data sets does not disturb the user experience (too much)**
1. Find a sample data set here: smb://olympic/babylon5/Scratch/Anton/SampleWS . The relevant file is MDEvent_ellipPeaks_new.nxs
1. If you have a 16GB computer run the follow script, else you have to modify the times you can load the file.

``` python
for i in range(10):
    name = "WS_" + str(i)
    Load(Filename='PATH_TO/MDEvent_ellipPeaks_new.nxs', OutputWorkspace=name)
```
1. Select all workspace and delete them.
   - Confirm that the time taken is acceptable and that the screen does not freeze. It will be unusable tough.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidPlot/src/Mantid/MantidUI.cpp
+++ b/MantidPlot/src/Mantid/MantidUI.cpp
@@ -460,7 +460,7 @@ void MantidUI::saveNexusWorkspace() { executeSaveNexus(); }
 */
 void MantidUI::deleteWorkspace(const QString &workspaceName) {
   auto workspaceNameStdString = workspaceName.toStdString();
-  auto& ads = Mantid::API::AnalysisDataService::Instance();
+  auto &ads = Mantid::API::AnalysisDataService::Instance();
   if (ads.doesExist(workspaceNameStdString)) {
     ads.remove(workspaceNameStdString);
   }
--- a/MantidPlot/src/Mantid/MantidUI.cpp
+++ b/MantidPlot/src/Mantid/MantidUI.cpp
@@ -459,10 +459,11 @@ void MantidUI::saveNexusWorkspace() { executeSaveNexus(); }
 @param workspaceName :: Name of the workspace to delete
 */
 void MantidUI::deleteWorkspace(const QString &workspaceName) {
-  auto alg = createAlgorithm("DeleteWorkspace");
-  alg->setLogging(false);
-  alg->setPropertyValue("Workspace", workspaceName.toStdString());
-  executeAlgorithmAsync(alg);
+  auto workspaceNameStdString = workspaceName.toStdString();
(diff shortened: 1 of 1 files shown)


MantidPlot/src/Mantid/MantidUI.cpp
```diff
@@ -459,10 +459,11 @@ void MantidUI::saveNexusWorkspace() { executeSaveNexus(); }
 @param workspaceName :: Name of the workspace to delete
 */
 void MantidUI::deleteWorkspace(const QString &workspaceName) {
-  auto alg = createAlgorithm("DeleteWorkspace");
-  alg->setLogging(false);
-  alg->setPropertyValue("Workspace", workspaceName.toStdString());
-  executeAlgorithmAsync(alg);
+  auto workspaceNameStdString = workspaceName.toStdString();
```
martyngigg: I agree with the swap to calling the ADS directly. Using an algorithm to do this seems like overkill.

MantidPlot/src/Mantid/MantidUI.cpp
```diff
@@ -459,10 +459,11 @@ void MantidUI::saveNexusWorkspace() { executeSaveNexus(); }
...
+  auto &ads = Mantid::API::AnalysisDataService::Instance();
+  if (ads.doesExist(workspaceNameStdString)) {
```
martyngigg: Is it necessary to check the existence first? `DataService::remove` does nothing if the name does not exist so this results in a double look up most of the time.

martyngigg (approved):
I've tested the provided scripts before and after the changes. Before the change I get a crash on Ubuntu if I increase the loop terminator to 5000.

After the changes the deletion happens successfully.

martyngigg:
:shipit:

PR #17749, opened 2016-10-10
Bring back old delete behaviour for Mantid Dock

### LINKED ISSUE #17748: Revert workspace deletion fix for MantidDock
Deleting many workspaces through the Manitd Dock can cause Mantid to crash. This was fixed [here](https://github.com/mantidproject/mantid/pull/17731). 

Unfortunately this does not play well with GroupWorkspaces. Only the parent will be deleted and the child workspaces become available in the MantidDock. 

For now it will be better to revert the changes and return to the old behaviour since this has been only reported by a single user. A future fix might be to batch the deletion of the workspaces in the MantidDock.

AntonPiccardoSelg:
Reverts the recent changes to the workspace deletion via the MantidDock.
## To test

Make sure that the changes have reverted to the old code.

Fixes #17748

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidPlot/src/Mantid/MantidUI.cpp
+++ b/MantidPlot/src/Mantid/MantidUI.cpp
@@ -459,8 +459,11 @@ void MantidUI::saveNexusWorkspace() { executeSaveNexus(); }
 @param workspaceName :: Name of the workspace to delete
 */
 void MantidUI::deleteWorkspace(const QString &workspaceName) {
+  auto workspaceNameStdString = workspaceName.toStdString();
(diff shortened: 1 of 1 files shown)


martyngigg (approved):
The code is now back at the state it was in v3.7.1

Deletion of a workspace group now occurs as expected in MantidPlot.

The issue of possible crash when deleting hundreds of workspaces at once will be addressed again for `3.9`

martyngigg:
:shipit:

PR #17755, opened 2016-10-10
Remove nearest neighbours code from matrix workspace

### LINKED ISSUE #17744: MatrixWorkspace::getNeighbours: ignoreMaskedDetectors flag ignored in all but first call
If we look at one example:

``` cpp
std::map<specnum_t, V3D>
MatrixWorkspace::getNeighbours(specnum_t spec, const double radius,
                               bool ignoreMaskedDetectors) const {
  if (!m_nearestNeighbours) {
    buildNearestNeighbours(ignoreMaskedDetectors);
  }
  std::map<specnum_t, V3D> neighbours =
      m_nearestNeighbours->neighboursInRadius(spec, radius);
  return neighbours;
}
```

We see that `ignoreMaskedDetectors` is ignored if the nearest-neighbor map has already been initialized.

If the initialization of the map has been triggered by a call to `getNeighbours` with the **same** value of `ignoreMaskedDetectors` everything is fine, otherwise the result will be wrong if the workspace has masked detectors.

### LINKED ISSUE #17745: Remove NearestNeighbours from MatrixWorkspace
The nearest-neighbor mechanism in MatrixWorkspace looks like it can easily be moved out of that class. It is used in very few places, so the performance implications should be minor. On the plus side, `MatrixWorkspace` would get simpler, and we would implicitly fix https://github.com/mantidproject/mantid/issues/17744.

### LINKED ISSUE #17751: MatrixWorkspace::getNeighboursExact: nNeighbours flag ignored in all but first call
``` cpp
std::map<specnum_t, V3D>
MatrixWorkspace::getNeighboursExact(specnum_t spec, const int nNeighbours,
                                    bool ignoreMaskedDetectors) const {
  if (!m_nearestNeighbours) {
    SpectrumDetectorMapping spectraMap(this);
    m_nearestNeighbours = boost::make_shared<NearestNeighbours>(
        nNeighbours, this->getInstrument(), spectraMap.getMapping(),
        ignoreMaskedDetectors);
  }
  std::map<specnum_t, V3D> neighbours = m_nearestNeighbours->neighbours(spec);
  return neighbours;
}
```

### LINKED ISSUE #17754: NearestNeighbours has inverted intepretation of ignoreMaskedDetectors.
``` cpp
std::map<specnum_t, IDetector_const_sptr>
NearestNeighbours::getSpectraDetectors(
    boost::shared_ptr<const Instrument> instrument,
    const ISpectrumDetectorMapping &spectraMap) {
  std::map<specnum_t, IDetector_const_sptr> spectra;
  if (spectraMap.empty())
    return spectra;
  auto cend = spectraMap.cend();
  for (auto citr = spectraMap.cbegin(); citr != cend; ++citr) {
    const std::vector<detid_t> detIDs(citr->second.begin(), citr->second.end());
    IDetector_const_sptr det = instrument->getDetectorG(detIDs);
    // Always ignore monitors and ignore masked detectors if requested.
    bool heedMasking = !m_bIgnoreMaskedDetectors && det->isMasked();
    if (!det->isMonitor() && !heedMasking) {
      spectra.emplace(citr->first, det);
    }
  }
  return spectra;
}
```

According to Doxygen for the class `*@param ignoreMaskedDetectors :: flag indicating that masked detectors should be ignored.` detectors that are masked should not be added to `spectra` in the loop above if `ignoreMaskedDetectors = true`. Obviously this is not the case.

### LINKED ISSUE #17766: Undefined behavior due to shared_ptr::release call from different threads?
`MatrixWorkspace::maskWorkspaceIndex` clears the `NearestNeighbours` object. The former is a `const` method and called from OpenMP loops, the latter is the same object for all threads. There is no locking mechanism, so probably this could lead to double-free or similar issues.

SimonHeybrock:
This PR removes the access to nearest-neighbour information from `MatrixWorkspace`. Instead a small class is provided for access.

By doing the allocation of the underlying `NearestNeighbours` object only on demand in the affected algorithms we avoid several issues with invalid state of the `NearestNeighbours` object in `MatrixWorkspace` that had previously required manually triggered updates in several places and had some related bugs.

**To test:**

Thorough code review.

Fixes #17745.

Several bugs were found and fixed as well:
Fixes #17744.
Fixes #17751.
Fixes #17754.
Fixes #17766.

Release notes were updated.

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/API/src/NearestNeighbourInfo.cpp
```diff
@@ -0,0 +1,70 @@
...
+*/
+std::map<specnum_t, Kernel::V3D>
+NearestNeighbourInfo::getNeighbours(specnum_t spec, const double radius) const {
+  return m_nearestNeighbours.neighboursInRadius(spec, radius);
+}
+
+/** Queries the NearestNeighbours object for the selected spectrum number.
+*
+* @param spec :: spectrum number of the detector you are looking at
+*be ignored. True to ignore detectors.
```
KarlPalmen: Sentence does not make sense to me.

KarlPalmen:
:shipit:

PR #17758, opened 2016-10-10
Fix for Larmor batch mode bug

### LINKED ISSUE #17757: Batch Reduction Bug for LARMOR
Rob wrote: 

> The attached zip file contains the contents of
> C:\users\larmor_mgr\Appdata\Roaming\mantidproject
> From ndllarmor2 but I’ve seen the bug on another machine (from which I deleted and didn’t save the files)
> If you replace your version of this folder with the zipped up one and run a batch mode reduction (example attached) and then compare it to the same data reduced in the same session by a single reduction they differ by a factor of 2 in this case.
> 
> Debugging was nasty…. Having checked on my office machine and found normal behaviour, I cleared the registry and uninstalled, delete c:\mantid and reinstalled but the error kept on coming back. I had to pick through the debug log to eventually spot that the reduction on the NDLLARMOR1 and 2 machines was using appdata/roaming.
> I’ve compared the larmor definition and the parameter files in the folder and they are exactly the same as the ones in c:\mantidinstall\instruments. It seems that the error is doing something really odd as it produces errors and also changes monitor settings. Below is the log window from the GUI for batch and single mode. As you can see the Single and Batch mode produce slightly different errors although I have no idea why.
> 
> I have no idea what the root cause is but I’m more concerned by the fact that I have no idea when the error crept in as it is not the sort of thing that I would necessarily think to check for. It also shows just how difficult it is to keep track of all the different places that mantid is storing stuff. We’ve had a couple of problems before with the instrument definitions stored in appdata. Is there a way to make it much more obvious if this is being used.
> e.g. it should never be used unless there is explicit change of setting made in mantid.user.properties.

AntonPiccardoSelg:
Fixes #17757 

The problem that Rob reported stemmed from a check that we perform when loading data. When the ISIS SANS Gui loads the first time, it generates an empty instrument and extracts information from it. The empty instrument is based on `Instrument_Definition.xml`.

This means that older IDFs cannot be used (this has lead to the SANS2DTubes option I believe). When loading data we check if the IDF that was used in the initial SANS setup is the same as the one which is associated with the data set. If not we swap the one of the SANS setup for the one fo the data file (which is the one to use). It also means that we need to read in the user file again. In batch mode this rereading missed the monitor spectrum. This is now being set manually. Also we only reload the IDF if the two files are not identical in content and we don't care about file location.
# To Test

Test this on Windows
Please find the test material here: \olympic\Babylon5\Scratch\Anton\Testing\17757\LARMOR_BUG
1. Open the ISIS SANS Gui
2. Set the instrument to LARMOR
3. Load the user file
4. Select the batch mode
5. Load the batch mode file
6. Run a 1D reduction
7. Switch to single mode (the data entries should be prepopulated from the batch run)
8. Run a 1D reduction 
9. Compare the output workspaces for both reductions (Workspaces `A 1 mM LAS  Ca` and `12931rear_1D_0.9_12.5`). 
   - Confirm that they are identical (plot the two spectra on top of each other for example).
10. Go to your AppData\Roaming\mantidproject folder. Copy the contain `mantid` folder into a secure location. Place the `mantid` folder from the test data into the location
11. Follow steps 3 to 9 and confirm that the reduced workspaces are still the same.
12. Replace the `mantid` folder with your old `mantid` folder.

For release notes see [here](https://github.com/mantidproject/mantid/pull/17758/commits/2f793ad14519859aa794794bd9050f6afe5bf90b)

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/docs/source/release/v3.9.0/sans.rst
+++ b/docs/source/release/v3.9.0/sans.rst
@@ -9,6 +9,7 @@ Bug Fixes
 ---------
 
 |
+- IDF bug when using Larmor in batch mode was resolved.
 
 X uncertainties (delta-Q)
 -------------------------
--- a/scripts/SANS/isis_reducer.py
+++ b/scripts/SANS/isis_reducer.py
@@ -773,7 +773,7 @@ class ISISReducer(Reducer):
 
         # Now check both underlying files. If they are equal, then don't do anything
         # else switch the underlying instrument
-        if su.are_two_files_identical(idf_path_reducer, idf_path_reducer):
+        if su.are_two_files_identical(idf_path_workspace, idf_path_reducer):
             return
         else:
             logger.notice("Updating the IDF of the Reducer. Switching from " +
--- a/scripts/SANS/isis_reducer.py
+++ b/scripts/SANS/isis_reducer.py
@@ -18,8 +18,11 @@ import SANSUtility as su
 import os
 import copy
 import sys
+from collections import namedtuple
 
 logger = Logger("ISISReducer")
+
+temp_reduction_framework_settings = namedtuple('temp_reduction_framework_settings', 'incident_monitor')
 
 ################################################################################
 # Avoid a bug with deepcopy in python 2.6, details and workaround here:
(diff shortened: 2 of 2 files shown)


tom-perkins:
Testing on Win7

martyngigg:
Confirmed that the output of the reduction in single/batch mode for LARMOR are now identical.

:shipit:

PR #17759, opened 2016-10-10
HistogramData rollout: algorithms <RebinByPulseTimes to RemoveBins>

### LINKED ISSUE #17665: HistogramData rollout: algorithms <15 to 25>
This issue is part of the HistogramData rollout, see https://github.com/mantidproject/mantid/issues/17641.

Framework/Algorithms/src/RebinByPulseTimes.cpp
Framework/Algorithms/src/RebinByTimeAtSample.cpp
Framework/Algorithms/src/Rebin.cpp
Framework/Algorithms/src/RebinToWorkspace.cpp
Framework/Algorithms/src/Rebunch.cpp
Framework/Algorithms/src/ReflectometryReductionOne.cpp
Framework/Algorithms/src/ReflectometryWorkflowBase.cpp
Framework/Algorithms/src/Regroup.cpp
Framework/Algorithms/src/RemoveBackground.cpp
Framework/Algorithms/src/RemoveBins.cpp

LamarMoore:
## Description of work.

This PR is part of refactoring effort to use the new interface of the HistogramData module.
See https://github.com/mantidproject/mantid/issues/17641 for details.

The following algorithms were refactored to use HistogramData:

Performance tests were added for:
RemoveBins
The following algorithms have improved performance:
RemoveBins improved performance by 35% for testRemoveFromMiddle.
## To test.

Code review.

Fixes #17665.


Framework/Algorithms/src/RebinByPulseTimes.cpp
```diff
@@ -60,11 +60,11 @@ void RebinByPulseTimes::doHistogramming(IEventWorkspace_sptr inWS,
     el.generateHistogramPulseTime(*XValues_new, y_data, e_data);
```
SimonHeybrock: Can we change the signature of this? Such that we can do

```cpp
outputWS->setHistogram(i, el.generateHistogramPulseTime(binEdges));
```

Framework/Algorithms/src/RebinByTimeAtSample.cpp
```diff
@@ -76,11 +78,11 @@ void RebinByTimeAtSample::doHistogramming(IEventWorkspace_sptr inWS,
                                      tofOffset);
```
SimonHeybrock: Have you considered changing the signature (see above)?

Framework/Algorithms/src/Rebunch.cpp
```diff
@@ -82,24 +83,20 @@ void Rebunch::exec() {
...
-    MantidVec &YValues_new = outputW->dataY(hist);
-    MantidVec &YErrors_new = outputW->dataE(hist);
-
     // output data arrays are implicitly filled by function
     if (point) {
-      rebunch_point(XValues, YValues, YErrors, XValues_new, YValues_new,
-                    YErrors_new, n_bunch);
+      rebunch_point(inputW->x(hist), inputW->y(hist), inputW->e(hist),
+                    outputW->mutableX(hist), outputW->mutableY(hist),
+                    outputW->mutableE(hist), n_bunch);
```
SimonHeybrock: Why not the nicer function signature based on `Histogram` that we discussed and tested a few days ago?

SimonHeybrock: Since a `Histogram` knows if it is `point` or `dist` there also is no need for checking this here, that can be done internally in the `rebunch` method.

SimonHeybrock: This also does not seem to cover the 4th case (point *and* frequencies)? Is that a bug in `Rebunch`? Maybe it should simply abort in that case?

Framework/Algorithms/src/Regroup.cpp
```diff
@@ -129,12 +132,9 @@ void Regroup::exec() {
...
  **/
-void Regroup::rebin(const std::vector<double> &xold,
-                    const std::vector<double> &yold,
-                    const std::vector<double> &eold,
-                    const std::vector<int> &xoldIndex,
-                    std::vector<double> &ynew, std::vector<double> &enew,
-                    bool distribution) {
+void Regroup::rebin(const HistogramX &xold, const HistogramY &yold,
+                    const HistogramE &eold, std::vector<int> &xoldIndex,
+                    HistogramY &ynew, HistogramE &enew, bool distribution) {
```
SimonHeybrock: Have you tried modernizing the signature?

Framework/Algorithms/src/RemoveBackground.cpp
```diff
@@ -123,11 +127,11 @@ void RemoveBackground::exec() {
...
     PARALLEL_START_INTERUPT_REGION
     // get references to output Workspace X-arrays.
-    MantidVec &XValues = outputWS->dataX(hist);
+    auto &XValues = outputWS->mutableX(hist);
     // get references to output workspace data and error. If it is new
     // workspace, data will be copied there, if old, modified in-place
-    MantidVec &YValues = outputWS->dataY(hist);
-    MantidVec &YErrors = outputWS->dataE(hist);
+    auto &YValues = outputWS->mutableY(hist);
+    auto &YErrors = outputWS->mutableE(hist);
```
SimonHeybrock: Can we pass a `Histogram` to `removeBackground`? It looks inefficient at first sight, because it would not be an in-place modification (`outputWS->setHistogram(hist, removeBackground(outputWS->histogram(hist))`), but `outputWS` is just created above and should have all shared data, so there would be basically no overhead. Except on second sight, if it is an in-place modification of the input workspace...

What do you think?

SimonHeybrock: (having had a look at `BackgroundHelper` below it may not be worth it, because there is a mess there anyway that is too much effort to do in a nicer way as part of these changes.)

Framework/Algorithms/test/RebinByTimeAtSampleTest.h
```diff
@@ -248,13 +248,13 @@ class RebinByTimeAtSampleTest : public CxxTest::TestSuite, public Super {
     TSM_ASSERT_EQUALS("Should not loose spectrum", 3,
                       result->getNumberHistograms());
 
-    auto y1 = result->readY(0);
+    auto y1 = result->y(0);
```
SimonHeybrock: `&`?

Framework/Algorithms/test/RebinByTimeAtSampleTest.h
```diff
@@ -248,13 +248,13 @@ class RebinByTimeAtSampleTest : public CxxTest::TestSuite, public Super {
...
     auto y1Sum = std::accumulate(y1.begin(), y1.end(), 0.0);
 
-    auto y2 = result->readY(1);
+    auto y2 = result->y(1);
```
SimonHeybrock: &?

Framework/Algorithms/test/RebinByTimeAtSampleTest.h
```diff
@@ -248,13 +248,13 @@ class RebinByTimeAtSampleTest : public CxxTest::TestSuite, public Super {
...
     auto y2Sum = std::accumulate(y2.begin(), y2.end(), 0.0);
 
-    auto y3 = result->readY(2);
+    auto y3 = result->y(2);
```
SimonHeybrock: &?

Framework/Algorithms/test/RebunchTest.h
```diff
@@ -154,46 +154,42 @@ class RebunchTest : public CxxTest::TestSuite {
...
+    BinEdges x(static_cast<size_t>(size), LinearGenerator(0.5, 0.75));
+    Counts y(size - 1);
+    CountVariances e(size - 1);
+
+    std::iota(y.begin(), y.end(), 1.0);
+    std::iota(e.begin(), e.end(), 1.0);
+
+    retVal->setBinEdges(0, x);
+    retVal->setCounts(0, y);
+    retVal->setCountVariances(0, e);
```
SimonHeybrock: `retVal->setHistogram(0, x, y);`
Note that you can omit `e`, since it is just the `sqrt`, it will be computed automatically.

SimonHeybrock: And why not use `LinearGenerator` also for `y`, so `std::iota` can be avoided?

Framework/Algorithms/test/RebunchTest.h
```diff
@@ -154,46 +154,42 @@ class RebunchTest : public CxxTest::TestSuite {
...
+    Points x(size, LinearGenerator(0.5, 0.5));
+    Counts y(size);
+    CountVariances e(size);
+
+    std::iota(y.begin(), y.end(), 1.0);
+    std::iota(e.begin(), e.end(), 1.0);
+
+    retVal->setPoints(0, x);
+    retVal->setCounts(0, y);
+    retVal->setCountVariances(0, e);
```
SimonHeybrock: See above.

Framework/Algorithms/test/RebunchTest.h
```diff
@@ -154,46 +154,42 @@ class RebunchTest : public CxxTest::TestSuite {
...
     retVal->initialize(ylen, xlen, xlen - 1);
-    double j = 1.0;
-
-    for (auto &x : x1) {
-      x = j * 0.5;
-      j += 1.5;
-    }
 
     Counts y1(xlen - 1);
     CountVariances e1(xlen - 1);
```
SimonHeybrock: See above.

Framework/Algorithms/test/RebunchTest.h
```diff
@@ -210,16 +206,10 @@ class RebunchTest : public CxxTest::TestSuite {
...
     retVal->initialize(ylen, xlen, xlen);
-    double j = 1.0;
-
-    for (auto &x : x1) {
-      x = j * 0.5;
-      j += 1.5;
-    }
 
     Counts y1(xlen);
     std::iota(y1.begin(), y1.end(), 0.0);
```
SimonHeybrock: Can omit `e1`, I think (if you use `setHistogram`, that is).

Framework/Algorithms/test/RegroupTest.h
```diff
@@ -59,8 +59,9 @@ class RegroupTest : public CxxTest::TestSuite {
   Workspace2D_sptr Create1DWorkspace(int size) {
     auto retVal = createWorkspace<Workspace2D>(1, size, size - 1);
     double j = 1.0;
+	auto &X = retVal->mutableX(0);
     for (int i = 0; i < size; i++) {
-      retVal->dataX(0)[i] = j * 0.5;
+      X[i] = j * 0.5;
       j += 1.5;
     }
     retVal->setCounts(0, size - 1, 3.0);
```
SimonHeybrock: `LinearGenerator`, use `setHistogram` and omit variances?

SimonHeybrock (commented):
Looks good, just some comments on how things could be made nicer.

Framework/Algorithms/src/RebinByTimeAtSample.cpp
LamarMoore: I did but I was worried the extra overhead in allocation and copying might result in a performance hit. ???

Framework/Algorithms/src/Rebunch.cpp
LamarMoore: Same reason as above. Took a performance hit so reverted to the messy signature. Do you think this is related to the problem you are currently investigating @SimonHeybrock ?

SimonHeybrock: As you know I tested this a few days ago with gcc (branch `rebunch_Histogram_refactor_test`) and did not see a difference. What were your results?

LamarMoore: When I initially tested the use of Histogram I noted almost a 70% degradation in performance which improved when replacing the std::vectors for Histogram types, still resulted in a degradation (close to 50%). It was at this point I contacted you.

Framework/Algorithms/src/Rebin.cpp
```diff
@@ -251,9 +251,9 @@ void Rebin::exec() {
...
       // get const references to input Workspace arrays (no copying)
-      const MantidVec &XValues = inputWS->readX(hist);
-      const MantidVec &YValues = inputWS->readY(hist);
-      const MantidVec &YErrors = inputWS->readE(hist);
+      auto &XValues = inputWS->x(hist).rawData();
+      auto &YValues = inputWS->y(hist).rawData();
+      auto &YErrors = inputWS->e(hist).rawData();
 
       // get references to output workspace data (no copying)
       MantidVec &YValues_new = outputWS->dataY(hist);
```
SimonHeybrock: Leftover from old interface.

Framework/Algorithms/src/Rebunch.cpp
SimonHeybrock: So did you test the changes I made in that branch?

Framework/Algorithms/src/Rebin.cpp
LamarMoore: Yes I will change this once the HistogramData/Rebin stuff gets merged

Framework/Algorithms/src/RebinByPulseTimes.cpp
LamarMoore: @SimonHeybrock  this would require a lot of changes to EventList. I think it should be handled in another issue.

Framework/Algorithms/src/Rebin.cpp
```diff
@@ -250,27 +251,15 @@ void Rebin::exec() {
...
       try {
-        VectorHelper::rebin(XValues, YValues, YErrors, XValues_new.rawData(),
-                            YValues_new, YErrors_new, dist);
+        outputWS->setHistogram(
+            hist, std::move(HistogramData::rebin(inputWS->histogram(hist),
+                                                 XValues_new)));
       } catch (std::exception &ex) {
         g_log.error() << "Error in rebin function: " << ex.what() << '\n';
         throw;
       }
```
SimonHeybrock: Is there a point in catching and rethrowing? I think `Algorithm` will just do the same, so I see little benefit.

Framework/Algorithms/src/Rebin.cpp
```diff
@@ -347,12 +336,12 @@ void Rebin::propagateMasks(API::MatrixWorkspace_const_sptr inputWS,
...
   // Create a vector to hold the redistributed weights
-  const MantidVec &XValues_new = outputWS->readX(hist);
+  auto &XValues_new = outputWS->x(hist);
   MantidVec newWeights(XValues_new.size() - 1), zeroes2(XValues_new.size() - 1);
   // Use rebin function to redistribute the weights. Note that distribution flag
   // is set
-  VectorHelper::rebin(masked_bins, weights, zeroes, XValues_new, newWeights,
-                      zeroes2, true);
+  VectorHelper::rebin(masked_bins, weights, zeroes, XValues_new.rawData(),
+                      newWeights, zeroes2, true);
```
SimonHeybrock: Am I overlooking something or could you also use your new rebin method here if you put the weights into a histogram as frequencies?

SimonHeybrock (commented):
Looks good, consider if these 2 further cleanup suggestions make sense:

Framework/HistogramData/src/Rebin.cpp
```diff
@@ -109,14 +109,18 @@ Histogram rebinFrequencies(const Histogram &input, const BinEdges &binEdges) {
...
     }
   }
 
+  for (size_t i = 0; i < size_ynew; ++i) {
+    auto width = xnew[i + 1] - xnew[i];
+    auto factor = 1 / width;
+    ynew[i] *= factor;
+    enew[i] = sqrt(enew[i]) * factor;
+  }
+
```
SimonHeybrock: Why?

SimonHeybrock: Could you write some tests for this?

LamarMoore:
@raquelalvarezbanos fixed the ReflectometryReductionOne unit tests to conform to the new HistogramData Rebin. Previously failing due to x data with identical values producing zero bin widths.

LamarMoore:
@raquelalvarezbanos fixed the GenericDataProcessorPresenter unit tests to conform to the new HistogramData Rebin. This used ReflectometryReductionOne and was failing due to x data with identical values producing zero bin widths.

Framework/HistogramData/test/RebinTest.h
```diff
@@ -321,7 +321,7 @@ class RebinTest : public CxxTest::TestSuite {
 
     TS_ASSERT_DELTA(outFreq.e()[0], std::sqrt(outFreq.y()[0]), 1e-14);
     TS_ASSERT_DELTA(outFreq.e()[1], std::sqrt(outFreq.y()[1]), 1e-14);
-    TS_ASSERT_DELTA(outFreq.e()[2], outFreq.y()[2], 1e-14);
+    TS_ASSERT_DELTA(outFreq.e()[2], std::sqrt(outFreq.y()[2]), 1e-14);
```
SimonHeybrock: Are you saying that the old unit test was wrong so it did not catch the issue?

Framework/HistogramData/src/Rebin.cpp
```diff
@@ -109,14 +109,18 @@ Histogram rebinFrequencies(const Histogram &input, const BinEdges &binEdges) {
...
       } else {
-        auto factor = 1 / nwidth;
-        ynew[inew] *= factor;
-        enew[inew] = sqrt(enew[inew]) * factor;
         inew++;
       }
     }
   }
 
+  for (size_t i = 0; i < size_ynew; ++i) {
```
SimonHeybrock: What are the performance implications of putting this outside the loop?

SimonHeybrock (commented):
Regarding your `rebin` changes:

Framework/HistogramData/test/RebinTest.h
LamarMoore: Yes

Framework/Algorithms/src/Rebin.cpp
```diff
@@ -333,15 +331,15 @@ void Rebin::propagateMasks(API::MatrixWorkspace_const_sptr inputWS,
...
+  auto errSize = weights.size();
+  Histogram oldHist(BinEdges(std::move(masked_bins)),
+                    Frequencies(std::move(weights)),
+                    FrequencyStandardDeviations(errSize, 0));
   // Use rebin function to redistribute the weights. Note that distribution flag
   // is set
-  VectorHelper::rebin(masked_bins, weights, zeroes, XValues_new.rawData(),
-                      newWeights, zeroes2, true);
+  auto newHist = HistogramData::rebin(oldHist, outputWS->binEdges(hist));
+  auto newWeights = newHist.y();
```
SimonHeybrock: reference?

Framework/Algorithms/src/Rebin.cpp
```diff
@@ -333,15 +331,15 @@ void Rebin::propagateMasks(API::MatrixWorkspace_const_sptr inputWS,
...
+  //// Create a zero vector for the errors because we don't care about them here
+  auto errSize = weights.size();
+  Histogram oldHist(BinEdges(std::move(masked_bins)),
+                    Frequencies(std::move(weights)),
+                    FrequencyStandardDeviations(errSize, 0));
   // Use rebin function to redistribute the weights. Note that distribution flag
   // is set
-  VectorHelper::rebin(masked_bins, weights, zeroes, XValues_new.rawData(),
-                      newWeights, zeroes2, true);
+  auto newHist = HistogramData::rebin(oldHist, outputWS->binEdges(hist));
```
SimonHeybrock: Aren't there some more changes to do? Am I overlooking something or are you not using the result?

SimonHeybrock: Sorry, not I think it is fine.

Framework/Algorithms/src/Rebin.cpp
```diff
@@ -338,13 +353,20 @@ void Rebin::propagateMasks(API::MatrixWorkspace_const_sptr inputWS,
...
-  auto newWeights = newHist.y();
+  bool ignoreErrors = getProperty("IgnoreBinErrors");
 
-  // Now process the output vector and fill the new masking list
-  for (size_t index = 0; index < newWeights.size(); ++index) {
-    if (newWeights[index] > 0.0)
-      outputWS->flagMasked(hist, index, newWeights[index]);
+  try {
+    auto newHist = HistogramData::rebin(oldHist, outputWS->binEdges(hist));
+    auto newWeights = newHist.y();
```
SimonHeybrock: Reference?

Framework/Algorithms/src/Rebin.cpp
```diff
@@ -112,6 +114,10 @@ void Rebin::init() {
   declareProperty(
       "FullBinsOnly", false,
       "Omit the final bin if it's width is smaller than the step size");
+
+  declareProperty("IgnoreBinErrors", false, "Ignore errors related to "
+                                            "zero/negative bin widths in "
+                                            "input/output workspaces.");
```
SimonHeybrock: These changes to Rebin look good and reasonable to me. 👍  Consider extending this comment to state what happens to spectra which such errors (as far as I can see Y and E are set to 0?).

Framework/HistogramData/inc/MantidHistogramData/Exception.h
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Mantid {
+namespace HistogramData {
+namespace Exception {
+
+/// Thrown whenever invalid bin edges are encountered by HistogramData Rebin
+class MANTID_HISTOGRAMDATA_DLL InvalidBinEdgesError
+    : public std::runtime_error {
+public:
+  InvalidBinEdgesError(const char *what);
```
SimonHeybrock: In C++11 there is a nice way to avoid this trivial implementation:

```cpp
  using std::runtime_error::runtime_error;
```
i.e., you can just bring the base class constructors into scope.

SimonHeybrock (commented):
A few small comments:

Framework/Algorithms/src/Rebin.cpp
```diff
@@ -115,9 +115,11 @@ void Rebin::init() {
       "FullBinsOnly", false,
       "Omit the final bin if it's width is smaller than the step size");
 
-  declareProperty("IgnoreBinErrors", false, "Ignore errors related to "
-                                            "zero/negative bin widths in "
-                                            "input/output workspaces.");
+  declareProperty("IgnoreBinErrors", false,
+                  "Ignore errors related to "
+                  "zero/negative bin widths in "
+                  "input/output workspaces. Wehn ignored, the signal and "
```
SimonHeybrock: Typo.

LamarMoore:
Interesting point about floating point representation for this PR. The `MERLINReduction` system test was failing due to a mismatch in the reference and output data. After a lot of digging I discovered the source of the issue was this block of code in the new `HistogramData::Rebin`:
```cpp
auto factor = 1 / owidth;
ynew[inew] += yold[iold] * delta * factor;
enew[inew] += eold[iold] * eold[iold] * delta * factor;
```
In the old `VectorHelper:Rebin`, this operation was performed using the following code:
```cpp
ynew[inew] += yold[iold] * delta / width;
enew[inew] += eold[iold] * eold[iold] * delta / width;
```
Introducing the calculation of `factor` produced a result (in a few cases) which differed by the least significant bit. So the number `6.63347e-06 ` becomes: 
`0.0000066334727689309362153643329085284108 ` in  `VectorHelper::Rebin` and 
`0.0000066334727689309353683313856542280718` in `HistogramData::Rebin`. 

`MERLINReduction` uses the `DirectEnergyConversion` script for performing the data reduction. The arguments passed to the script for MERLINReduction include multiplication factors on the order of 10^12 in total which resulted in physically insignificant portions of the data exploding and affecting the final output data. Just something to keep in mind when designing tests. Ensure they aren't so sensitive that they can be tripped up by limitations in floating point representation.

SimonHeybrock:
@LamarMoore Sorry for the wait, I forgot about this.

Latest changes look ok :shipit:.

@LamarMoore Could you please add a link to an issue for the `MERLIN` test, so we make sure this information is not forgotten (after all you had to spend some time tracking this down) next time some changes in the last bit break it? In principle just going to another compiler version or CPU could cause that, I think, so we are just "lucky" that the small change you did above is sufficient for trick the finite precision arithmetic to give the same result..

martyngigg:
@LamarMoore Nice work on tracking that change down. As Simon says, is there an issue for the `MERLIN` test?

martyngigg:
I'm going to merge this as there are a lot of changes and I don't want it to get stale.

LamarMoore:
@martyngigg @SimonHeybrock I contacted the maintainer of the MERLINReduction code @abuts who has said that he is satisfied with the fix and will not likely make any changes to the system test in the foreseeable future.

SimonHeybrock:
@LamarMoore Then he must have misunderstood. The issue was not fixed, this is just a workaround and it can potentially break any time, e.g., when you switch to a new Visual Studio version.

PR #17767, opened 2016-10-11
HistogramData rollout: algorithms 76 to 81

### LINKED ISSUE #17710: HistogramData rollout: algorithms 76 to 81
HistogramData rollout: algorithms 76 to 81

This issue is part of the HistogramData rollout, see https://github.com/mantidproject/mantid/issues/17641.

```
Framework/CurveFitting/src/Algorithms/FitPowderDiffPeaks.cpp
Framework/CurveFitting/src/Algorithms/LeBailFit.cpp
Framework/CurveFitting/src/Algorithms/NormaliseByPeakArea.cpp
Framework/CurveFitting/src/Algorithms/PawleyFit.cpp
Framework/CurveFitting/src/Algorithms/RefinePowderInstrumentParameters3.cpp
```

~~Framework/CurveFitting/src/Algorithms/LeBailFunction.cpp~~

raquelalvarezbanos:
Description of work.

**To test:**

Fixes #17710

---
#### Reviewer

##### Code Review
##### Functional Tests


SimonHeybrock:
@raquelalvarezbanos Would you like me to have a look at your changes or are you still going to make substantial modifications?

raquelalvarezbanos:
Thanks @SimonHeybrock. I don't plan to make substantial modifications. I only need to figure out what is wrong with `RefinePowderInstrumentParameters`'s performance test. The error is:

/Users/builder/Jenkins/workspace/pull_requests-osx/Framework/CurveFitting/test/Algorithms/RefinePowderInstrumentParameters3Test.h:485: createSuite() already declared

I remember seeing that locally but in the end it worked. This is the only thing I wanted to fix, but it is OK if you want to start reviewing.

Also, I am seeing strange numbers in terms of performance: none of the algorithms for which I wrote performance tests showed significant improvement (which I guess can be normal). However, for `NormaliseByPeakArea` I am seeing a decrease in performance: `NormaliseByPeakAreaTestPerformance::test_sum_false()` took ~1.31s to run on my machine (Windows, release mode) before the changes, and ~3.86s after the changes. I guess I must be doing something wrong. `NormaliseByPeakAreaTestPerformance::test_sum_true()` also showed a decrease in performance, but only slightly.

Framework/CurveFitting/test/Algorithms/RefinePowderInstrumentParameters3Test.h
```diff
@@ -204,289 +470,75 @@ class RefinePowderInstrumentParameters3Test : public CxxTest::TestSuite {
...
-
-    if (bankid == 1) {
-      string filename(
-          "/home/wzz/Mantid/Code/debug/MyTestData/bank1profile.txt");
-      importTableTextFile(filename, strparams, colnames, 6);
-    } else {
-      throw runtime_error("generateInstrumentProfile supports bank 1 only.");
-    }
+class RefinePowderInstrumentParameters3TestPerformance
+    : public CxxTest::TestSuite {
```
SimonHeybrock: @raquelalvarezbanos Someone else had the same problem recently: cxxtest beaks if the class declaration spans 2 lines (but clang-format forces it to two lines). In the end we used this trick that saves a few characters, see if that fits:

```cpp
using CxxTest::TestSuite;
class RefinePowderInstrumentParameters3TestPerformance : public TestSuite {
```

raquelalvarezbanos: Yes, it fits, thanks!

Framework/CurveFitting/inc/MantidCurveFitting/Algorithms/LeBailFit.h
```diff
@@ -267,9 +272,9 @@ class DLLExport LeBailFit : public API::Algorithm {
   //--------------  Background function Suite
   //----------------------------------------
   /// Re-fit background according to the new values
-  void fitBackground(size_t wsindex, API::FunctionDomain1DVector domain,
-                     API::FunctionValues values,
-                     std::vector<double> &background);
+  // std::vector<double> fitBackground(size_t wsindex,
+  //                                  API::FunctionDomain1DVector domain,
+  //                                  API::FunctionValues values);
```
SimonHeybrock: Remove?

raquelalvarezbanos: I was tempted to remove it, but I didn't in the end because of [this](https://github.com/mantidproject/mantid/blob/17710_HistogramData_Rollout_Algorithms_76-81_new/Framework/CurveFitting/src/Algorithms/LeBailFit.cpp#L2413) comment. I don't know if anyone is planning to do work on this... Methods `LeBailFit::smoothBackgroundExponential` and `LeBailFit::smoothBackgroundAnalytical` also contain dead code which I didn't remove.

Framework/CurveFitting/src/Algorithms/FitPowderDiffPeaks.cpp
```diff
@@ -3339,7 +3315,7 @@ bool observePeakParameters(Workspace2D_sptr dataws, size_t wsindex,
  * @param Y :: vector to get maximum value from
  * @return index of the maximum value
  */
-size_t findMaxValue(const MantidVec Y) {
+size_t findMaxValue(const std::vector<double> &Y) {
```
SimonHeybrock: Is this used for anything but `HistogramY`? If not you could change the argument type and avoid the `rawData()` call above.

SimonHeybrock: On second thought, could `std::max_element` be used here (with `std::distance`)?

Framework/CurveFitting/src/Algorithms/LeBailFit.cpp
```diff
@@ -446,15 +447,15 @@ void LeBailFit::processInputBackground() {
...
+*  0: data;  1: calculated pattern; 3: difference
+*  4: input pattern w/o background
+*  5~5+(N-1): optional individual peak
+*/
 void LeBailFit::execPatternCalculation() {
   // Generate domain and values vectors
-  const MantidVec &vecX = m_dataWS->readX(m_wsIndex);
-  MantidVec &vecY = m_outputWS->dataY(CALDATAINDEX);
+  const auto &vecX = m_dataWS->x(m_wsIndex).rawData();
+  std::vector<double> vecY(m_outputWS->y(CALDATAINDEX).size(), 0);
```
SimonHeybrock: Why is this being zero-initialized now, instead of copied? Is it just overwritten below?

SimonHeybrock: Having read a bit more below: You had to get into some mess to get the new data types into `calculateDiffractionPattern`. Have you considered changing the parameter types of the method to `HistogramX` and `HistogramY` instead of `std::vector<double>`?  I think that would make things much easier at the call sites.

Framework/CurveFitting/src/Algorithms/LeBailFit.cpp
```diff
@@ -471,31 +472,34 @@ void LeBailFit::execPatternCalculation() {
...
-      emptyvec, vecY, rfactor);
+      m_dataWS->x(m_wsIndex).rawData(), m_dataWS->y(m_wsIndex).rawData(), true,
+      true, emptyvec, vecY, rfactor);
+  m_outputWS->mutableY(CALDATAINDEX) = vecY;
 
   size_t numpts = vecY.size();
   // Calculate background
-  MantidVec &vec_bkgd = m_outputWS->dataY(INPUTBKGDINDEX);
+  std::vector<double> vec_bkgd(m_outputWS->y(INPUTBKGDINDEX).size(), 0);
   m_lebailFunction->function(vec_bkgd, vecX, false, true);
```
SimonHeybrock: Consider adapting the function signature of `LeBailFunction::function`. It does not seem to be used outside this algorithm so it would be a local change and would make your life easier here, I think.

Framework/CurveFitting/src/Algorithms/LeBailFit.cpp
```diff
@@ -471,31 +472,34 @@ void LeBailFit::execPatternCalculation() {
...
+  m_outputWS->mutableY(INPUTBKGDINDEX) = vec_bkgd;
+
   for (size_t i = 0; i < numpts; ++i)
-    m_outputWS->dataY(INPUTPUREPEAKINDEX)[i] =
-        m_outputWS->readY(OBSDATAINDEX)[i] -
-        m_outputWS->readY(INPUTBKGDINDEX)[i];
+    m_outputWS->mutableY(INPUTPUREPEAKINDEX)[i] =
+        m_outputWS->y(OBSDATAINDEX)[i] - m_outputWS->y(INPUTBKGDINDEX)[i];
```
SimonHeybrock: I think you can just write `m_outputWS->mutableY(INPUTPUREPEAKINDEX) = m_outputWS->y(OBSDATAINDEX) - m_outputWS->y(INPUTBKGDINDEX)`.

Framework/CurveFitting/src/Algorithms/LeBailFit.cpp
```diff
@@ -471,31 +472,34 @@ void LeBailFit::execPatternCalculation() {
...
 
   // Set up output workspaces
   for (size_t i = 0; i < numpts; ++i) {
-    m_outputWS->dataY(DATADIFFINDEX)[i] =
-        m_outputWS->readY(OBSDATAINDEX)[i] - m_outputWS->readY(CALDATAINDEX)[i];
+    m_outputWS->mutableY(DATADIFFINDEX)[i] =
+        m_outputWS->y(OBSDATAINDEX)[i] - m_outputWS->y(CALDATAINDEX)[i];
   }
```
SimonHeybrock: Same here.

Framework/CurveFitting/src/Algorithms/LeBailFit.cpp
```diff
@@ -471,31 +472,34 @@ void LeBailFit::execPatternCalculation() {
...
 
   // Calcualte individual peaks
   bool ploteachpeak = this->getProperty("PlotIndividualPeaks");
   g_log.information() << "Output individual peaks  = " << ploteachpeak << ".\n";
   if (ploteachpeak) {
     for (size_t ipk = 0; ipk < m_lebailFunction->getNumberOfPeaks(); ++ipk) {
-      MantidVec &vecTemp = m_outputWS->dataY(9 + ipk);
+      std::vector<double> vecTemp(m_outputWS->y(9 + ipk).size(), 0);
       m_lebailFunction->calPeak(ipk, vecTemp, vecX);
```
SimonHeybrock: Adapt the function signature?

Framework/CurveFitting/src/Algorithms/LeBailFit.cpp
```diff
@@ -551,15 +555,15 @@ void LeBailFit::execRefineBackground() {
   vector<double> backgroundvalues(numpts);
```
SimonHeybrock:
```cpp
vector<double> backgroundvalues(values);
```
?

Framework/CurveFitting/src/Algorithms/LeBailFit.cpp
```diff
@@ -579,14 +583,14 @@ void LeBailFit::execRefineBackground() {
     m_backgroundFunction->function(domain, values);
     for (size_t i = 0; i < numpts; ++i) {
       backgroundvalues[i] = values[i];
```
SimonHeybrock: Why not `backgroundvalues = values;` outside the loop?

Framework/CurveFitting/src/Algorithms/LeBailFit.cpp
```diff
@@ -632,29 +636,29 @@ void LeBailFit::execRefineBackground() {
   m_backgroundFunction->function(domain, values);
   for (size_t i = 0; i < numpts; ++i) {
     backgroundvalues[i] = values[i];
```
SimonHeybrock: Assign vectors?

Framework/CurveFitting/src/Algorithms/LeBailFit.cpp
```diff
@@ -632,29 +636,29 @@ void LeBailFit::execRefineBackground() {
...
 
   g_log.notice() << "[RefineBackground] Best Rwp = " << bestR.Rwp
                  << ",  vs. recovered best Rwp = " << outputR.Rwp << ".\n";
 
   // 4. Add data (0: experimental, 1: calcualted, 2: difference)
   for (size_t i = 0; i < numpts; ++i) {
-    m_outputWS->dataY(1)[i] = valueVec[i] + backgroundvalues[i];
-    m_outputWS->dataY(2)[i] = vecY[i] - (valueVec[i] + backgroundvalues[i]);
+    m_outputWS->mutableY(1)[i] = valueVec[i] + backgroundvalues[i];
+    m_outputWS->mutableY(2)[i] = vecY[i] - (valueVec[i] + backgroundvalues[i]);
```
SimonHeybrock: In general, calling `dataY` or `mutableY` in a loop over bins should be avoided, since it has non-negligible cost due to the copy-on-write mechanism. Here it is probably fine since it is done only for 2 histograms, but if you have a big workspace with many histograms you can see the difference.

Framework/CurveFitting/src/Algorithms/LeBailFit.cpp
```diff
@@ -632,29 +636,29 @@ void LeBailFit::execRefineBackground() {
...
   }
 
   //   (3: peak without background, 4: input background)
   // m_backgroundFunction->function(domain, values);
   for (size_t i = 0; i < values.size(); ++i) {
-    m_outputWS->dataY(CALBKGDINDEX)[i] = backgroundvalues[i];
-    m_outputWS->dataY(CALPUREPEAKINDEX)[i] = valueVec[i];
+    m_outputWS->mutableY(CALBKGDINDEX)[i] = backgroundvalues[i];
+    m_outputWS->mutableY(CALPUREPEAKINDEX)[i] = valueVec[i];
   }
```
SimonHeybrock: Wouldn't `=` without loop work here?

Framework/CurveFitting/src/Algorithms/LeBailFit.cpp
```diff
@@ -1478,14 +1479,16 @@ void LeBailFit::execRandomWalkMinimizer(size_t maxcycles,
   Rfactor startR(-DBL_MAX, -DBL_MAX);
 
   // Process background to make a pure peak spectrum in output workspace
-  MantidVec &vecBkgd = m_outputWS->dataY(INPUTBKGDINDEX);
+  std::vector<double> vecBkgd(m_outputWS->y(INPUTBKGDINDEX).size(), 0);
   m_lebailFunction->function(vecBkgd, vecX, false, true);
```
SimonHeybrock: See above, can avoid some copies if you change the function signature.

Framework/CurveFitting/src/Algorithms/LeBailFit.cpp
```diff
@@ -1508,10 +1511,10 @@ void LeBailFit::execRandomWalkMinimizer(size_t maxcycles,
...
-  MantidVec &vecDiff = m_outputWS->dataY(DATADIFFINDEX);
-  MantidVec &vecCalPurePeak = m_outputWS->dataY(CALPUREPEAKINDEX);
-  MantidVec &vecCalBkgd = m_outputWS->dataY(CALBKGDINDEX);
+  auto &vecCalY = m_outputWS->mutableY(CALDATAINDEX);
+  auto &vecDiff = m_outputWS->mutableY(DATADIFFINDEX);
+  auto &vecCalPurePeak = m_outputWS->mutableY(CALPUREPEAKINDEX);
+  auto &vecCalBkgd = m_outputWS->mutableY(CALBKGDINDEX);
   for (size_t i = 0; i < numpts; ++i) {
     // Calculated (refined) data
     vecCalY[i] = vecCalPurePeaks[i] + vecBkgd[i];
```
SimonHeybrock: There are `operator=` and `operator+` etc. for the types in `HistogramData`, so you can avoid the loop.

Framework/CurveFitting/src/Algorithms/LeBailFit.cpp
```diff
@@ -1961,29 +1964,29 @@ void LeBailFit::addParameterToMCMinimize(vector<string> &parnamesforMC,
...
+*values
+* @param vecBkgd:: vector of background values (input)
+* @param values :: (output) function values, i.e., summation of all peaks and
+*background in option
+* @param rfactor:  R-factor (Rwp and Rp) as output
+*
+* @return :: boolean value.  whether all the peaks' parameters are physical.
+*/
+bool LeBailFit::calculateDiffractionPattern(const std::vector<double> &vecX,
+                                            const std::vector<double> &vecY,
```
SimonHeybrock: Have you considered changing these types to `HistgramX`, etc.?

Framework/CurveFitting/src/Algorithms/NormaliseByPeakArea.cpp
```diff
@@ -283,29 +275,31 @@ double NormaliseByPeakArea::fitToMassPeak(const MatrixWorkspace_sptr &yspace,
...
+*/
 void NormaliseByPeakArea::normaliseTOFData(const double area,
                                            const size_t index) {
-  const auto &inY = m_inputWS->readY(index);
-  auto &outY = m_normalisedWS->dataY(index);
+  const auto &inY = m_inputWS->y(index);
+  std::vector<double> outY(m_normalisedWS->y(index).size());
   std::transform(inY.begin(), inY.end(), outY.begin(),
                  std::bind2nd(std::divides<double>(), area));
+  m_normalisedWS->mutableY(index) = outY;
```
SimonHeybrock: Why are you creating a temporary vector?

You can either do the transform on `HistogramY`, or use `operator/=`:

```cpp
m_normalisedWS->mutableY(index) /= area;
```

The latter might be slightly slower than doing the transform directly.

Framework/CurveFitting/src/Algorithms/NormaliseByPeakArea.cpp
```diff
@@ -283,29 +275,31 @@ double NormaliseByPeakArea::fitToMassPeak(const MatrixWorkspace_sptr &yspace,
...
 
-  const auto &inE = m_inputWS->readE(index);
-  auto &outE = m_normalisedWS->dataE(index);
+  const auto &inE = m_inputWS->e(index);
+  std::vector<double> outE(m_normalisedWS->e(index).size());
   std::transform(inE.begin(), inE.end(), outE.begin(),
                  std::bind2nd(std::divides<double>(), area));
+  m_normalisedWS->mutableE(index) = outE;
```
SimonHeybrock: Same here. But actually you could just use `Histogram &operator/=(Histogram &histogram, const double factor)` (include `MantidHistogramData/HistogramMath.h`) and get rid of the entire function, I think?

Framework/CurveFitting/src/Algorithms/NormaliseByPeakArea.cpp
```diff
@@ -283,29 +275,31 @@ double NormaliseByPeakArea::fitToMassPeak(const MatrixWorkspace_sptr &yspace,
...
- * @param index Index of the workspace. Only used when not summing.
- */
+* @param accumWS Workspace used to accumulate the final data
+* @param yValues Input signal values for y-space
+* @param eValues Input errors values for y-space
+* @param index Index of the workspace. Only used when not summing.
+*/
 void NormaliseByPeakArea::saveToOutput(const API::MatrixWorkspace_sptr &accumWS,
                                        const std::vector<double> &yValues,
                                        const std::vector<double> &eValues,
```
SimonHeybrock: Consider changing the function parameter types, that will make things easier at the call sites of this function.

Framework/CurveFitting/src/Algorithms/NormaliseByPeakArea.cpp
```diff
@@ -314,10 +308,10 @@ void NormaliseByPeakArea::saveToOutput(const API::MatrixWorkspace_sptr &accumWS,
 
   if (m_sumResults) {
     const size_t npts(accumWS->blocksize());
-    auto &accumY = accumWS->dataY(0);
-    auto &accumE = accumWS->dataE(0);
-    const auto accumYCopy = accumWS->readY(0);
-    const auto accumECopy = accumWS->readE(0);
```
SimonHeybrock: Why is this making a copy of the vectors? Do you see a reason that I am overlooking?

Framework/CurveFitting/src/Algorithms/NormaliseByPeakArea.cpp
```diff
@@ -354,11 +348,11 @@ void NormaliseByPeakArea::symmetriseYSpace() {
       static_cast<int64_t>(m_symmetrisedWS->getNumberHistograms());
 
   for (int64_t i = 0; i < nhist; ++i) {
-    const auto &xsym = m_symmetrisedWS->readX(i);
-    auto &ySymOut = m_symmetrisedWS->dataY(i);
-    auto &eSymOut = m_symmetrisedWS->dataE(i);
-    const auto yIn = m_yspaceWS->readY(i); // copy
-    const auto eIn = m_yspaceWS->readE(i); // copy
```
SimonHeybrock: Again, why?

SimonHeybrock (commented):
You certainly got one of the chunks with more required  changes...

I finished looking through your changes to the algorithm sources (I'll do the tests later). Looks mostly good, some comments below (one of them may explain the slowdown you are seeing):

Framework/CurveFitting/test/Algorithms/FitPowderDiffPeaksTest.h
```diff
@@ -107,8 +436,18 @@ class FitPowderDiffPeaksTest : public CxxTest::TestSuite {
     if (!peakdataws) {
       return;
     }
-
     TS_ASSERT_EQUALS(peakdataws->getNumberHistograms(), 5);
+    TS_ASSERT_EQUALS(peakdataws->blocksize(), 2284);
+    TS_ASSERT_EQUALS(peakdataws->x(0).rawData(), peakdataws->x(1).rawData());
+    TS_ASSERT_EQUALS(peakdataws->x(0).rawData(), peakdataws->x(2).rawData());
+    TS_ASSERT_EQUALS(peakdataws->x(0).rawData(), peakdataws->x(3).rawData());
+    TS_ASSERT_EQUALS(peakdataws->x(0).rawData(), peakdataws->x(4).rawData());
```
SimonHeybrock: There is an `operator==` for these types available in `#include "MantidTestHelpers/HistogramDataTestHelper.h"` (it is not included by default since we do not want to do exact vector comparison in general code). But using `rawData` is also fine :+1:.

Framework/CurveFitting/test/Algorithms/FitPowderDiffPeaksTest.h
```diff
@@ -117,355 +456,90 @@ class FitPowderDiffPeaksTest : public CxxTest::TestSuite {
...
-      errss << "createReflectionWorkspace does not support option " << option
-            << '\n';
-      errss << "Supported options include 1 (LaB6 for PG3 bank 1). ";
-      throw invalid_argument(errss.str());
-    }
-
-    // 3. Output information
-    std::cout << "Created Table Workspace with " << hklws->rowCount()
-              << " entries of peaks.\n";
+  void setUp() {
```
SimonHeybrock: Use the `override` keyword for `setUp` and `tearDown` please. We will move the Mantid builds to Ubuntu 16.04 soon and the newer gcc creates warnings otherwise, which makes the builds fail.

Framework/CurveFitting/test/Algorithms/FitPowderDiffPeaksTest.h
```diff
@@ -117,355 +456,90 @@ class FitPowderDiffPeaksTest : public CxxTest::TestSuite {
...
-    newrow19 << "Tcross" << 0.356;
-    TableRow newrow20 = tablews->appendRow();
-    newrow20 << "Width" << 5.00256;
-    TableRow newrow21 = tablews->appendRow();
-    newrow21 << "Zero" << 0.0;
-    TableRow newrow22 = tablews->appendRow();
-    newrow22 << "Zerot" << 499.99;
-
-    return;
+  void tearDown() {
```
SimonHeybrock: `override`

Framework/CurveFitting/test/Algorithms/LeBailFitTest.h
```diff
@@ -30,6 +21,642 @@ using namespace std;
...
+  const size_t size = vecX.size();
+  auto dataws = boost::dynamic_pointer_cast<API::MatrixWorkspace>(
+      API::WorkspaceFactory::Instance().create("Workspace2D", 1, size, size));
+
+  // c) Input data
+  for (size_t i = 0; i < size; ++i) {
+    dataws->dataX(0)[i] = vecX[i];
+    dataws->dataY(0)[i] = vecY[i];
+    dataws->dataE(0)[i] = vecE[i];
+  }
```
SimonHeybrock: You can do this without temporary arrays:

```cpp
dataws->setHistogram(0, BinEdges{ /* initializer_list */ }, Counts {...}, CountStandardDeviations{...});
```

Framework/CurveFitting/test/Algorithms/LeBailFitTest.h
```diff
@@ -30,6 +21,642 @@ using namespace std;
...
+  size_t size = vecx.size();
+  auto dataws = boost::dynamic_pointer_cast<API::MatrixWorkspace>(
+      API::WorkspaceFactory::Instance().create("Workspace2D", 1, size, size));
+
+  // c) Input data
+  for (size_t i = 0; i < size; ++i) {
+    dataws->dataX(0)[i] = vecx[i];
+    dataws->dataY(0)[i] = vecy[i];
+    dataws->dataE(0)[i] = vece[i];
+  }
```
SimonHeybrock: Could use `setHistogram`.

Framework/CurveFitting/test/Algorithms/LeBailFitTest.h
```diff
@@ -30,6 +21,642 @@ using namespace std;
...
+  // Set error
+  const MantidVec &vecY = ws->readY(0);
+  MantidVec &vecE = ws->dataE(0);
+  size_t numpts = vecY.size();
+  for (size_t i = 0; i < numpts; ++i) {
+    double y = vecY[i];
+    double e = 1.0;
+    if (y > 1.0)
+      e = sqrt(y);
+    vecE[i] = e;
```
SimonHeybrock: `vecE[i] = std::max(1.0, sqrt(vecY[i]));`?

Framework/CurveFitting/test/Algorithms/RefinePowderInstrumentParameters3Test.h
```diff
@@ -21,6 +21,289 @@ using namespace Mantid::API;
...
+      std::stringstream ss;
+      ss.str(line);
+      ss >> x >> y >> e;
+      vecX.push_back(x);
+      vecY.push_back(y);
+      if (e < 0.00001) {
+        if (y > 1.0)
+          e = std::sqrt(y);
+        else
+          e = 1.0;
```
SimonHeybrock: `std::max`?

raquelalvarezbanos:
Thanks @SimonHeybrock I'll start adding your suggested changes.

SimonHeybrock (commented):
Only some minor (mostly cosmetic) comments on the tests:

Framework/CurveFitting/inc/MantidCurveFitting/Algorithms/LeBailFit.h
SimonHeybrock: Hmm, that was added at least two years ago, I'd just remove these things.

raquelalvarezbanos: OK, I'll remove dead code and code that is never called.

Framework/CurveFitting/src/Algorithms/LeBailFit.cpp
raquelalvarezbanos: Yes, that was exactly the problem. As you say, I guess updating the method to take `HistogramX` (and `Y`, etc) is the best solution. I'll change that.

Framework/CurveFitting/src/Algorithms/NormaliseByPeakArea.cpp
raquelalvarezbanos: No idea, I'll remove that.

Framework/CurveFitting/src/Algorithms/LeBailFit.cpp
raquelalvarezbanos: Yes, my mistake, sorry.

Framework/CurveFitting/test/Algorithms/FitPowderDiffPeaksTest.h
raquelalvarezbanos: Thanks, I didn't know, but if it exists, I'll use it :)

Framework/CurveFitting/test/Algorithms/FitPowderDiffPeaksTest.h
raquelalvarezbanos: :+1:

Framework/CurveFitting/test/Algorithms/LeBailFitTest.h
raquelalvarezbanos: I am trying to do the following:

```
BinEdges vecX = { 70931.750000, 70943.609000, ...};
Counts vecY = {0.000000,    0.000000,    0.695623, ...};
CountStandardDeviations vecE = {1.000000,  1.000000,  1.000000, ...};

auto dataws = boost::dynamic_pointer_cast<API::MatrixWorkspace>(
      API::WorkspaceFactory::Instance().create("Workspace2D", 1, 127, 127));

dataws->setHistogram(0, vecX, vecY, vecE);
```

but I get an exception in `Histogram::checkSize()` [here](https://github.com/mantidproject/mantid/blob/master/Framework/HistogramData/inc/MantidHistogramData/Histogram.h#L399-L400). At that point `target` is 126 but all my vectors `vecX`, `vecY`, `vecE`, were 127, I guess it is because as `vecX` is `BinEdges` it is expected to have an extra element?

SimonHeybrock: @raquelalvarezbanos Sorry, my bad, you need `Points` instead of `BinEdges` (as I should have seen from the workspace creation which uses `size` for both X and Y length).

raquelalvarezbanos: :+1:

raquelalvarezbanos:
@SimonHeybrock : thanks for the thorough testing, I think have addressed all your suggestions, but please let me know if I missed anything. I'll measure again performance.

Framework/CurveFitting/src/Algorithms/LeBailFit.cpp
```diff
@@ -1478,14 +1464,16 @@ void LeBailFit::execRandomWalkMinimizer(size_t maxcycles,
   Rfactor startR(-DBL_MAX, -DBL_MAX);
 
   // Process background to make a pure peak spectrum in output workspace
-  MantidVec &vecBkgd = m_outputWS->dataY(INPUTBKGDINDEX);
-  m_lebailFunction->function(vecBkgd, vecX, false, true);
-  MantidVec &dataPurePeak = m_outputWS->dataY(INPUTPUREPEAKINDEX);
+  std::vector<double> vecBkgd =
+      m_lebailFunction->function(vecX.rawData(), false, true);
```
SimonHeybrock: Remove `rawData`?

Framework/CurveFitting/src/Algorithms/LeBailFit.cpp
```diff
@@ -1478,14 +1464,16 @@ void LeBailFit::execRandomWalkMinimizer(size_t maxcycles,
...
+  m_outputWS->mutableY(INPUTBKGDINDEX) = vecBkgd;
+  std::vector<double> dataPurePeak(m_outputWS->y(INPUTPUREPEAKINDEX).size(), 0);
   transform(vecInY.begin(), vecInY.end(), vecBkgd.begin(), dataPurePeak.begin(),
             ::minus<double>());
```
SimonHeybrock: Why not store the result of `m_lebailFunction->function` in a `HistogramY` (or directly return one?)? THen you can simply use `operator-` or `-=` here.

Framework/CurveFitting/src/Algorithms/LeBailFunction.cpp
```diff
@@ -78,33 +79,32 @@ LeBailFunction::LeBailFunction(std::string peaktype) {
...
-                              bool calbkgd) const {
-  if (out.size() != xvalues.size())
-    throw runtime_error("xvalues and out have different sizes.");
+* @param xvalues :: input vector
+* @param calpeaks :: if true, calculate peaks
+* @param calbkgd :: if true, then calculate background and add to output.
+* otherwise, assume zero background
+* @return :: output vector
+*/
+std::vector<double>
```
SimonHeybrock: As mentioned above, consider returning `HistogramY`, unless that does not work somewhere?

Framework/CurveFitting/src/Algorithms/LeBailFunction.cpp
```diff
@@ -135,12 +135,16 @@ void LeBailFunction::function(std::vector<double> &out,
...
+
+  return out;
 }
 
 /**  Calculate a single peak's value
-  */
-void LeBailFunction::calPeak(size_t ipk, std::vector<double> &out,
-                             const std::vector<double> &xvalues) const {
+*/
+std::vector<double> LeBailFunction::calPeak(size_t ipk,
```
SimonHeybrock: Have you considered returning `HistogramY`?

Framework/CurveFitting/src/Algorithms/NormaliseByPeakArea.cpp
```diff
@@ -283,45 +275,38 @@ double NormaliseByPeakArea::fitToMassPeak(const MatrixWorkspace_sptr &yspace,
...
+* @param accumWS Workspace used to accumulate the final data
+* @param yValues Input signal values for y-space
+* @param eValues Input errors values for y-space
+* @param index Index of the workspace. Only used when not summing.
+*/
 void NormaliseByPeakArea::saveToOutput(const API::MatrixWorkspace_sptr &accumWS,
-                                       const std::vector<double> &yValues,
-                                       const std::vector<double> &eValues,
+                                       const HistogramData::HistogramY &yValues,
+                                       const HistogramData::HistogramE &eValues,
```
SimonHeybrock: Look at how this function is called, and how Y and E are used below --- at least sometimes data could be shared, so consider passing shared pointers here, i.e., the result of `sharedY` etc, and then use `setSharedY` below.

Framework/CurveFitting/test/Algorithms/FitPowderDiffPeaksTest.h
SimonHeybrock: Still missing an `override` here (unless the github diff is messed up).

Framework/CurveFitting/test/Algorithms/FitPowderDiffPeaksTest.h
SimonHeybrock: `override`

SimonHeybrock (commented):
@raquelalvarezbanos Good. I noticed only a few very minor things:

Framework/CurveFitting/src/Algorithms/NormaliseByPeakArea.cpp
```diff
@@ -305,8 +307,8 @@ void NormaliseByPeakArea::saveToOutput(const API::MatrixWorkspace_sptr &accumWS,
     for (size_t j = 0; j < npts; ++j) {
       double accumYj = accumWS->y(0)[j];
       double accumEj = accumWS->e(0)[j];
-      double rhsYj = yValues[j];
-      double rhsEj = eValues[j];
+      double rhsYj = yValues->rawData()[j];
+      double rhsEj = eValues->rawData()[j];
```
SimonHeybrock: Consider moving the call to `rawData` outside the loop. Just like the old `dataY()` methods (and as done for `accumWS->mutableY()` above), it is better to buffer the reference since the copy-on-write mechanism makes write access more expensive.

SimonHeybrock (commented):
Looks all good now, just one small comment:

Framework/CurveFitting/src/Algorithms/NormaliseByPeakArea.cpp
```diff
@@ -303,12 +303,14 @@ void NormaliseByPeakArea::saveToOutput(
     const size_t npts(accumWS->blocksize());
     auto &accumY = accumWS->mutableY(0);
     auto &accumE = accumWS->mutableE(0);
+    auto yValuesRaw = yValues->rawData();
+    auto eValuesRaw = eValues->rawData();
```
SimonHeybrock: Reference?

SimonHeybrock:
👍 
:shipit: once Jenkins gives his ok.

raquelalvarezbanos:
Thanks again @SimonHeybrock for testing!

SimonHeybrock:
@raquelalvarezbanos No problem, you're welcome. Thanks for the work!

martyngigg:
@raquelalvarezbanos Sorry that this has been lingering around. It looks like the Ubuntu build has some warnings to take care of. Once those are taken care of I can merge it.

raquelalvarezbanos:
Thanks @martyngigg !

PR #17775, opened 2016-10-11
Check if data loaded before updating presenter

### LINKED ISSUE #17769: Exception on muon fitting tab when no data loaded
#### Expected behaviour
If no data is loaded, clicking the controls on the Muon Analysis fitting tab should not do anything.

#### Actual behaviour
Exception - see Mantid's "unhandled exception" dialog. Clicking "continue" is enough to make the error go away and you can continue as before.

Open Muon Analysis interface, but do not load any data. Instead, go straight to the "Data Analysis" tab. Then change the radio button under "Runs" (or some other control) - see exception.

#### Note
This is an unusual use case. The way scientists use this interface is to load data on the Home tab and only then go to the fitting tab. Nevertheless, we should handle this case better and program defensively.

tom-perkins:
Fix an exception seen when clicking controls on the fitting tab of Muon Analysis before any data had been loaded. Only update the presenter if some data has been loaded.

**To test:**
- Set facility to ISIS or SMuS, open _Interfaces/Muon/Muon Analysis_
- _Do not_ load any data. Instead go straight to the "Data Analysis" tab
- Click on various controls here. You should not see any unhandled exception dialog.

(No unit tests - the fit data/function presenters are tested, but the exception here is thrown before we get to the testable part of this interface)

Fixes #17769 

Release notes not necessary - this is an unlikely use case

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidQt/CustomInterfaces/src/Muon/MuonAnalysis.cpp
+++ b/MantidQt/CustomInterfaces/src/Muon/MuonAnalysis.cpp
@@ -2903,7 +2903,7 @@ std::string MuonAnalysis::getSubtractedPeriods() const {
  * Pass this information to the fit helper
  */
 void MuonAnalysis::dataToFitChanged() {
-  if (m_fitDataPresenter) {
+  if (m_fitDataPresenter && m_loaded) { // Only act if some data is loaded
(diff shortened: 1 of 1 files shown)


raquelalvarezbanos:
I didn't get any exceptions.

:shipit:

PR #17795, opened 2016-10-12
FindEPP : Fitting range

### LINKED ISSUE #17761: FindEPP : improved initial guess for sigma and fitting range
`FindEPP` deduces the initial guesses for peak centre and fitting range based on some logic.
Currently, it takes the subarray of `y_values`, wherever the value is above half maximum, and takes first and last element indices from this subarray to guess the full-width at half-maximum (FWHM). This is later used to set the initial sigma and fitting range. While working fine for smooth peaks, this is not the best algorithm for real life data that can have large noise at tails. Alternative way is to find the first elements around the peak, which drop below `0.5* height` and use those to identify the FWHM, and eventually initial sigma, and fitting range.

gvardany:
`FindEPP` deduces the initial guesses for peak centre and fitting range based on some logic.
Currently, it takes the subarray of `y_values`, wherever the value is above half maximum, and takes first and last element indices from this subarray to guess the full-width at half-maximum (FWHM). This is later used to set the initial sigma and fitting range. While working fine for smooth peaks, this is not the best algorithm for real life data that can have large noise at tails. Alternative way is to find the first elements around the maximum, which drop below `0.5 * height`, and use those to identify the FWHM, and eventually the initial sigma, and the fitting range. This PR implements the latter, which makes the algorithm more robust and reusable for real data. This improvement is essential for IN16B reduction at ILL, which makes use of `FindEPP` but not quite successfully with the current implementation.

**To test:**
1. Load some real-data containing one peak.
2. Run `FindEPP` and check if the fit output is reasonable.
3. Make sure the current tests pass.

Fixes #17761 .

---
#### Reviewer

##### Code Review
##### Functional Tests


gvardany:
@mganeva Could you please have a look if this change does not cause problems for your data?

mganeva:
I've checked it on some test TOFTOF datasets and did not notice any difference in the result.

louisemccann:
Tested with raw IRIS data, looks good to me :shipit:

PR #17800, opened 2016-10-13
Peak position diagnostic tool

### LINKED ISSUE #17753: Diagnostic peak comparison tool on instrument pick tab.
This is a sub issue of #10144 .

Create an additional tool on the instrument viewer pick tab that allows the user to select two already created peaks. This will then compute the d-spacing of each and the angles between them. The computed information should then be listed in the info window on the pick tab.

samueljackson92:
Added a new tool to the instrument view that allows the user to select two peaks and examine there properties and the angles between them.

**To test:**
- Load something that you can look at peaks with. I used WISH00035988.raw.
- Open the instrument view and select the pick tab
- Use the peak selection tool to choose some peaks (at least two)
- Choose two peaks on the interface: click one peak then click another.
- Observe the information about them displayed in the box on the left is correct

Fixes #17753 .

---
#### Reviewer

##### Code Review
##### Functional Tests


samueljackson92:
@OwenArnold might want to take a look at this as well.

raquelalvarezbanos:
@samueljackson92 You'll have to show me how use the new tool, I don't seem to be able to pick peaks with it...

samueljackson92:
@raquelalvarezbanos This should be ready for testing now unless the build servers throw up any issues.

raquelalvarezbanos:
Looks good:

You can select as many peaks as you want and then use the new tool to select pairs of peaks. Some information about the selected pair is shown on the left panel.

newtool

:shipit: when all tests pass

MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/InputController.h
```diff
@@ -245,29 +245,59 @@ class EXPORT_OPT_MANTIDQT_MANTIDWIDGETS InputControllerDraw
...
-  void erase(const QRect &);
-
 private:
   void drawCursor(QPixmap *cursor) override;
-  void signalLeftClick() override;
   void setPosition(const QPoint &pos) override;
   void resize() override;
 
-  QRect m_rect;
+protected:
```
martyngigg: It's best practice to avoid protected data members where possible - can we keep these private? Protected data increases the number of places that could possibly modify the state and cause trouble: http://softwareengineering.stackexchange.com/questions/162643/why-is-clean-code-suggesting-avoiding-protected-variables

For `m_image`, can be add a constructor parameter passing in the QPixmap for the icon? As far as I can see it is not used anywhere else?

For `m_rect` I think we need to be careful too as there is a variable of the same name in a derived class and it's not then clear which variable is being used.

MantidQt/MantidWidgets/src/InstrumentView/InstrumentWidgetPickTab.cpp
```diff
@@ -15,14 +15,19 @@
...
-#include "MantidAPI/ITableWorkspace.h"
+#include "MantidAPI/FrameworkManager.h"
 #include "MantidAPI/IPeaksWorkspace.h"
+#include "MantidAPI/ITableWorkspace.h"
 #include "MantidAPI/MatrixWorkspace.h"
 #include "MantidAPI/Sample.h"
 #include "MantidAPI/WorkspaceFactory.h"
 #include "MantidAPI/TableRow.h"
-#include "MantidAPI/FrameworkManager.h"
+#include "MantidAPI/WorkspaceFactory.h"
```
martyngigg: Do we need all of these extra includes? I can't see where things like `ConfigService`, `WorkspaceFactory`, `ParameterMap.h` are needed?

MantidQt/MantidWidgets/src/InstrumentView/ProjectionSurface.cpp
```diff
@@ -191,34 +197,20 @@ void ProjectionSurface::draw(MantidGLWidget *widget, bool picking) const {
...
     if (!picking) {
       QPainter painter(widget);
-      RectF windowRect = getSurfaceBounds();
-      m_maskShapes.setWindow(windowRect, painter.viewport());
-      m_maskShapes.draw(painter);
-      for (int i = 0; i < m_peakShapes.size(); ++i) {
-        m_peakShapes[i]->setWindow(windowRect, painter.viewport());
-        m_peakShapes[i]->draw(painter);
-      }
+      drawMaskShapes(painter);
```
martyngigg: A nice bit of refactoring here. This looks much cleaner and and it's easier to follow what is happening.

martyngigg (changes requested):

MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/InputController.h
samueljackson92: 1) Yes that is possible to add a constructor for that, in that case it might actually be easier to just have one class instead of base + 2 derived classes. They do the same thing but use different signals/images.

2) Dependant on the answer to 1 anyway, but also I don't see where the duplicate member name is?

MantidQt/MantidWidgets/src/InstrumentView/InstrumentWidgetPickTab.cpp
samueljackson92: Agreed, I think this was just clang-format moving things around

MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/InputController.h
martyngigg: https://github.com/mantidproject/mantid/blob/51daa1c0424fd6d8dc79f1ed7b350943e70f1123/MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/InputController.h#L328

Could we have the base class have a virtual method that the derived classes implement here. The method could take the `QRect` as a parameter and this way it could be `private` in the base class?

martyngigg: You're right, my mistake. I misread the name.

martyngigg:
@samueljackson92 Looks like some clang formatting is in order...

martyngigg:
@raquelalvarezbanos Would you mind retesting the functionality when you have some time?

raquelalvarezbanos:
Yes, I will re-test the functionality. Thanks for reviewing this @martyngigg

raquelalvarezbanos:
Re-tested functionality, works as before.

martyngigg (approved):

PR #17810, opened 2016-10-13
Files without an extension can not be seen in load/save dialogues

### LINKED ISSUE #17807: Files with no extension can not be selected from load dialogue
### Expected behavior

Selecting all files shows all files.

### Actual behavior

Files with no extension are not shown. Files with an extension, including hidden files, show up fine.

Make a file with no extension. Try and load it from Mantid's Load algorithm.

### Platforms affected

On a brief investigation Linux does not show files without an extension, Mac shows the file but does not allow it to be selected and Windows allows the files to be selected.

ianbush:
This PR replaces most file filter instances of `*.*` with `*`. This should mean files with no extension are shown in load and save dialogues.

**To test:**

Take a look at the dialogues changed. For a few of them try browsing to a file with no extension, and check it can now be seen.

This needs to be tested on a selection of platforms, at least one for Windows, one for Mac and one for Linux. Please add a comment if you have tested on one of the platforms.

A review of the changes should also be made to make sure there are no potential problems with this, as it changes a large number of dialogues.

Fixes #17807 .

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/MantidPlot/src/FitDialog.cpp
+++ b/MantidPlot/src/FitDialog.cpp
@@ -659,7 +659,7 @@ void FitDialog::saveUserFunction() {
           "Parent of FitDialog is not ApplicationWindow as expected.");
     }
     QString filter = tr("MantidPlot fit model") + " (*.fit);;";
-    filter += tr("All files") + " (*.*)";
+    filter += tr("All files") + " (*)";
     QString fn = MantidQt::API::FileDialogHandler::getSaveFileName(
         app, tr("MantidPlot") + " - " + tr("Save Fit Model As"),
         app->fitModelsPath + "/" + name, filter);
@@ -1420,7 +1420,7 @@ void FitDialog::saveInitialGuesses() {
           "Parent of FitDialog is not ApplicationWindow as expected.");
     }
     QString filter = tr("MantidPlot fit model") + " (*.fit);;";
-    filter += tr("All files") + " (*.*)";
+    filter += tr("All files") + " (*)";
     QString fn = MantidQt::API::FileDialogHandler::getSaveFileName(
         app, tr("MantidPlot") + " - " + tr("Save Fit Model As"),
         app->fitModelsPath + "/" + d_current_fit->objectName(), filter);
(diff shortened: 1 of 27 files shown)


KarlPalmen:
I'm testing for Windows 7.

KarlPalmen:
I wonder would it be better to define `"All files (*)"` in one place and get it wherever it is needed.

KarlPalmen:
ISIS SANS -> Add Runs -> Browse  does not have an All Files option. I take it this is OK.

KarlPalmen:
I've tried numerous file dialogs and whenever an All Files option was available and used, I got files without file extension in the list of files shown. *One nut for the squirrel*.

ianbush:
@KarlPalmen Yes, I would agree these should be defined in one place. I think that is better as a separate maintenance ticket though, as it would be good to do the same for things like NeXus files `.nxs, .nx5, .hdf` and text `.csv, .txt, .dat`.

This ticket should not add an All Files option for any interfaces, such as ISIS SANS, that do not have it already. In that case it is safest to assume the original author of the interface added the right options.

KarlPalmen:
Yes I agree. We could even have a few standard file dialog boxes for Mantid rather then having code dispersed.

tom-perkins:
Testing on Ubuntu 14.04

tom-perkins:
Tested all the changed dialogs that I could find - "All files" successfully found a file with no extension in each case. :+1:  for Ubuntu 14.04

ianbush:
Ticket for further tidying this up - #17829.

gvardany:
Tested on Mac OS X 10.11, was able to see and load a file with no extension.

KarlPalmen:
Red Hat still needs testing. Are there any other operating systems that need testing?

ianbush:
If everything looks good on Ubuntu 14.04, Max OS X 10.11 and Windows 7 I am confident everything is correct, so I do not think it also needs testing on RedHat (unless there are normally major differences on it for the GUI).

martyngigg:
I think both Ubuntu & RedHat would be GTK based so extra testing on RHEL is probably not required.

KarlPalmen:
No one has stated that RHEL testing is needed. :shipit:

PR #17824, opened 2016-10-14
Fixed assignment issue in ConvertUnits.

### LINKED ISSUE #17821: Fix issue with assignment in ConvertUnits::removeUnphysicalBins for Indirect
```cpp
    // Next, loop again copying in the correct range for each spectrum
    for (int64_t j = 0; j < int64_t(numSpec); ++j) {
      auto edges = workspace->binEdges(j);
      auto k = lastBins[j];

      result->mutableX(j).assign(edges.cbegin(), edges.cbegin() + k);

      // If the entire X range is not covered, generate fake values.
      if (k < maxBins) {
        std::iota(result->mutableX(j).begin() + k, result->mutableX(j).end(),
                  workspace->x(j)[k] + 1);
      }

      result->mutableY(j)
          .assign(workspace->y(j).cbegin(), workspace->y(j).cbegin() + (k - 1));
      result->mutableE(j)
          .assign(workspace->e(j).cbegin(), workspace->e(j).cbegin() + (k - 1));
    }
```

These should be copy calls, not assignments (`std::vector::assign` does not assign elements, but the whole range, i.e., changes the length which is unintentional here.

This throws an error (size mismatch) right now, so at least no bad data a created, but it should be fixed quickly.

This bug was introduced in a recent `Histogram` refactoring. The old code looked like this:

```cpp
    // Next, loop again copying in the correct range for each spectrum
    for (int64_t j = 0; j < int64_t(numSpec); ++j) {
      const MantidVec &X = workspace->readX(j);
      const MantidVec &Y = workspace->readY(j);
      const MantidVec &E = workspace->readE(j);
      MantidVec &Xnew = result->dataX(j);
      MantidVec &Ynew = result->dataY(j);
      MantidVec &Enew = result->dataE(j);
      int k;
      for (k = 0; k < lastBins[j] - 1; ++k) {
        Xnew[k] = X[k];
        Ynew[k] = Y[k];
        Enew[k] = E[k];
      }
      Xnew[k] = X[k];
      ++k;
      // If necessary, add on some fake values to the end of the X array (Y&E
      // will be zero)
      if (k < maxBins) {
        for (int l = k; l < maxBins; ++l) {
          Xnew[l] = X[k] + 1 + l - k;
        }
      }
    }
```

SimonHeybrock:
See issue for description of the problem. I tried adding a regression test but could not figure out a way to trigger the code section that causes the issue (ideas welcome).

**To test:**

One way to test this is as follows:

``` python
Load(Filename='/home/simon/data/TrainingCourseData/MAR11015.raw', OutputWorkspace='MAR11015')
converted = ConvertUnits(InputWorkspace='MAR11015',Target='DeltaE',EMode='Indirect',EFixed=10)
```

Before the fix this fails with a size mismatch error from FixedLengthVector. After the fix this should work.
Of course MARI is not an indirect instrument so this actually should not work, i.e., I would not want to add this as a system test.

Fixes #17821.

This issue was released as part of v3.8 and might need to be in a patch release, so I am not adding release notes for now.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/Algorithms/src/ConvertUnits.cpp
+++ b/Framework/Algorithms/src/ConvertUnits.cpp
@@ -813,18 +813,18 @@ API::MatrixWorkspace_sptr ConvertUnits::removeUnphysicalBins(
       auto edges = workspace->binEdges(j);
       auto k = lastBins[j];
 
-      result->mutableX(j).assign(edges.cbegin(), edges.cbegin() + k);
+      auto &X = result->mutableX(j);
+      std::copy(edges.cbegin(), edges.cbegin() + k, X.begin());
 
       // If the entire X range is not covered, generate fake values.
       if (k < maxBins) {
-        std::iota(result->mutableX(j).begin() + k, result->mutableX(j).end(),
-                  workspace->x(j)[k] + 1);
+        std::iota(X.begin() + k, X.end(), workspace->x(j)[k] + 1);
       }
 
-      result->mutableY(j)
-          .assign(workspace->y(j).cbegin(), workspace->y(j).cbegin() + (k - 1));
-      result->mutableE(j)
-          .assign(workspace->e(j).cbegin(), workspace->e(j).cbegin() + (k - 1));
+      std::copy(workspace->y(j).cbegin(), workspace->y(j).cbegin() + (k - 1),
+                result->mutableY(j).begin());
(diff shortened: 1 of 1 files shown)


KarlPalmen:
Test worked, but with a copy of the RAW file got from the data archive rather than the pathname given in the script.

KarlPalmen:
:shipit:

martyngigg (approved):

Issue #17622, opened 2016-09-28
Muon Analysis: points from beta testers (for next release)


tom-perkins:
_Feedback from beta testers for Mantid 3.7.20160922.1557, Windows_

These are the lower priority points that were pushed back to release 3.9 (see #17602)
- [ ] Automatically plot the tiled plot in the plot window after a simultaneous fit (#17971)
- [ ] Multiple results table: think of a better header than f0, f1 for datasets (#17976)
- [x] ~~Setting tie f0.Phi=f1.Phi caused a crash once (not second time) - try to replicate?~~
- [ ] Error message when entering a backward run range in sequential fit dialog e.g. 12345-43 (this is probably right, maybe the message should just be more descriptive) (#17978)
- [ ] Check where start time is set

tom-perkins:
(Can no longer reproduce the crash when setting ties)

tom-perkins:
Separated out into separate issues, so there is nothing left in this issue. Closing as complete.

PR #17830, opened 2016-10-14
Expose plotSubplots to Python

### LINKED ISSUE #17511: Expose multi-plot option to Python and use in interfaces
PR #17434 adds a new multilayer plot option on a workspace group (each workspace in one tile) or single workspace (each spectrum in one tile).

Expose this functionality to Python so that it can be used from scripts and from custom interfaces.
In particular, if available from Python, this can be used for plotting multi-fit results in Muon Analysis (see #15518).

(The existing command `newTiledWindow` creates a tiled window rather than the multilayer plot introduced in #17434).

tom-perkins:
Expose the `plotSubplots` command to Python. This functionality was previously available in C++, but being able to use it from Python is useful (for example it is needed for plotting multi-fit results in the muon interface).

Added the ability to plot in an existing window rather than creating a new graph. This is for consistency with other plot commands.

**To test:**

Use the following script to try out the `plotSubplots` command. You can get the data from `\\olympic\babylon5\Scratch\TomPerkins\test_subplot_data`, or use some other workspaces if you prefer.

``` python
# Load some files
names = []
runs = range(24580, 24584)
base = 'MUSR000'
ext = '.nxs'
for run in runs:
    ws = str(run)
    file = base + ws + ext
    names.append(ws)
    Load(file, OutputWorkspace=ws)

# Plot spectra from one workspace
#4 tiles, one spectrum in each
g1 = plotSubplots(names[0], range(4))

# Plot spectra from 4 workspaces
#4 tiles, 4 spectra in each
g2 = plotSubplots(names, range(4))

# Try a WorkspaceGroup
group = GroupWorkspaces(names)
g3 = plotSubplots(group, 0, error_bars=True)

# Plot something else in an existing window 
plotSubplots(group, 1, window=g3)
```

Fixes #17511.
Re #17622 (related, but doesn't close that issue)

[Release notes](https://github.com/mantidproject/mantid/blob/111beb7f94c186bcf8ac805151711430c02efc84/docs/source/release/v3.9.0/ui.rst#plotting-improvements) have been updated.

---
#### Reviewer

##### Code Review
##### Functional Tests


gvardany:
Changes look and work fine. Except for some alignment issues, which probably should go to a separate ticket. Can you try to reproduce similar to these?

I think this is a very handy tool from scripting window, so it may worth an illustration in release notes.
Also it was not so clear to me, what are the actual values for `distribution` argument for a user to type in. 

- Canvas alignment for one workspace many spectra:
one-ws

- Legend alignment for many workspaces
many-ws

- Legends are completely off, when plotting one workspace many spectra over an existing window holding many workspaces. Though when exporting the window to `.jpeg`, there they look correct.
one-after-many

tom-perkins:
Thanks, I've added a screenshot to the release notes and opened an issue #17874 for the alignment issues you described. I can reproduce these problems too, I think it may be due to when the legend is added versus when the graph is scaled (but will investigate!)

The `distribution` argument takes a `MantidQt::DistributionFlag` enum, just like `plotSpectrum` and other existing commands. The values are 0 for default, 1 to force plotting as a distribution and 2 to disable plotting as a distribution.

gvardany:
Thanks for the update, the figure looks nice, happy to :shipit:

PR #17834, opened 2016-10-15
Replace PARALLEL_FOR1 and PARALLEL_FOR2 with PARALLEL_FOR_IF.

quantumsteve:
Description of work.

This replaces `PARALLEL_FOR1` and `PARALLEL_FOR2` with `PARALLEL_FOR_IF` and the variadic function ThreadSafe(...) to determine whether one or more workspaces thread-safe.

**To test:**

This follows up on #17145.

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/DataHandling/src/LoadEventNexus.cpp
```diff
@@ -1614,7 +1614,7 @@ void LoadEventNexus::loadEvents(API::Progress *const prog,
       if (mT0 != 0.0) {
         int64_t numHistograms =
             static_cast<int64_t>(m_ws->getNumberHistograms());
-        PARALLEL_FOR1(m_ws)
+        PARALLEL_FOR_IF(m_ws->threadSafe())
```
quantumsteve: `m_ws` is an `EventWorkspaceCollection` which does NOT inherit from `DataItem` and `m_ws->threadsafe() will always return true. Is there any reason to keep this check?

quantumsteve:
@martyngigg Would you mind looking at this next week (Before the next review meeting)?

Framework/DataHandling/src/LoadEventNexus.cpp
martyngigg: Is it worth keeping for consistency with the the general style in many other algorithms?

martyngigg:
Apart from the comment raised around the `EventWorkspaceCollection` in LoadEventNexus this looks good.

Framework/DataHandling/src/LoadEventNexus.cpp
quantumsteve: Is it worth maintaining the static_assert that the workspace inherits from DataItem?

martyngigg: Maybe it's better to remove that and just require a pointer type with a `threadSafe()` method?

martyngigg:
@quantumsteve The branch has developer a conflict

quantumsteve:
@martyngigg I removed the `static_assert` that was incompatible with `EventWorkspaceCollection`.

martyngigg:
This looks good now. It provides a much cleaner macro for the OMP loops.

:shipit:

PR #17843, opened 2016-10-17
Remove assigned but never used variables

rosswhitfield:
This should fix most cases of variables that are assigned but then never used. As reported by flake8, F841

This may disagree with Pylint because flake8 doesn't count unused function arguments while Pylint does.

**To test:**
- Code review
- Test should still pass

**Does not need to be in the release notes.**

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/scripts/PyChop/ISISFermi.py
+++ b/scripts/PyChop/ISISFermi.py
@@ -184,10 +184,8 @@ class ISISFermi:
             raise ValueError('Incident energy has not be specified')
         chop_par = self.__chopperParameters[self.instname][self.choppername]['par']
         pslit = chop_par[0] / 1000.00
-        dslat = (chop_par[0] + chop_par[1]) / 1000.00
         radius = chop_par[2] / 1000.00
         rho = chop_par[3] / 1000.00
-        tjit = chop_par[5] * 1.0e-6
         return Chop.tchop(self.freq, Ei, pslit, radius, rho)
 
     def getModWidth(self, Ei_in=None):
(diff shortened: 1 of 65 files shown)


scripts/PyChop/ISISFermi.py
```diff
@@ -184,10 +184,8 @@ def getChopWidth(self, Ei_in=None):
             raise ValueError('Incident energy has not be specified')
         chop_par = self.__chopperParameters[self.instname][self.choppername]['par']
         pslit = chop_par[0] / 1000.00
-        dslat = (chop_par[0] + chop_par[1]) / 1000.00
```
AndreiSavici: I would just comment out changes in this file, since they might need them later

scripts/SANS/ISISCommandInterface.py
```diff
@@ -44,6 +44,7 @@ def appwidgets():
 
 def SetVerboseMode(state):
     # TODO: this needs to be on the reducer
+    global _VERBOSE_
```
AndreiSavici: Global???? WTF? I would leave it to fail without global

scripts/SANS/isis_reducer.py
```diff
@@ -847,15 +847,15 @@ def is_based_on_event(self):
         if self.is_can():
             sample = self.get_can()
             try:
-                dummy_ws = mtd[can.loader.wksp_name + "_monitors"]
+                mtd[can.loader.wksp_name + "_monitors"]
```
AndreiSavici: add a note that this should be done using something like `if can.loader.wksp_name + "_monitors" in mtd.getObjectNames():`

AndreiSavici (changes requested):
Just add TODOs in several places, and comment out instead of deleting the lines that might be useful later. It might increase the number of pylint warnings, so one should keep an eye on that

scripts/SANS/ISISCommandInterface.py
rosswhitfield: But now the function actually works. Since this previously did nothing but is called by parts of Mantid I might as well replace it with `pass`, to make it obvious this function does nothing?

scripts/PyChop/ISISFermi.py
rosswhitfield: That's what the git history is for.

AndreiSavici (approved):

AndreiSavici:
:shipit: when tests pass

Issue #17844, opened 2016-10-17
ISIS reduction: Custom filename should overwrite custom function for file name. 


abuts:
When users deploy ISIS reduction script, they can provide custom function, which calculates file names to save reduction results from incident energy, instrument run number and other properties of interest to user or just custom file name to save these results. 

If both are provided, custom function currently overrides custom file name.  

I got custom request to change this behaviour to opposite e.g. custom file name overwrites custom function.

AndreiSavici:
Branch merged into master a while ago, but the PR had no `Fixes ...`

PR #17846, opened 2016-10-17
Re #17844 Should fix the issue.

abuts:
This should be very easy ticket to test. See #17844

To satisfy user requests I've just moved around two getters and, if both file name and custom function are defined, instead of calculating custom function first, I am returning the file name. 

No need in release notes -- 12 min on fixing the issue and request is just trivial, though it took an hour to understand what they are actually want because the real problem behind their requests is unreliable reload of a Python package.  This solution solves particular user request, but I have no idea how reload package issue can be addressed. 

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/scripts/Inelastic/Direct/PropertiesDescriptors.py
+++ b/scripts/Inelastic/Direct/PropertiesDescriptors.py
@@ -472,33 +472,39 @@ class SaveFileName(PropDescriptor):
         self._custom_print = None
 
     def __get__(self,instance,owner=None):
-
+        # getter functional interface.
         if instance is None:
             return self
+
+        # if custom file name provided, use it
+        if self._file_name:
+            return self._file_name
+
+        # if custom function to generate file name is proivede, use this function
         if self._custom_print is not None:
             return self._custom_print()
 
-        if self._file_name:
-            return self._file_name
+        # user provided nothing.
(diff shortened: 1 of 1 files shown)


gvardany:
Looks fine code wise, is there a way/data to try out how it functions?

abuts:
The Mantid part of the code is covered by unit tests in Mantid/scripts/test/ + system tests check general functionality available for users (e.g.  https://github.com/mantidproject/mantid/blob/master/Testing/SystemTests/tests/analysis/ISIS_MariReduction.py) would run as in real reduction if you machine is set up to run system tests. 
MARI users wrote custom extension, which causes reload problem (see http://shadow.nd.rl.ac.uk/wiki/idr/index.php/ISIS_direct_reduction_script_description.) so it should also run if you set it up for the files present in system tests directory. 

But unless you want to learn how ISIS reduction works, I would not bother much with checking functionality because as it is a Python and user wanted it yesterday, I have deployed the changes on reduction machine and it works to user's satisfaction.

gvardany:
System tests run, `/Users/vardanyan/mantid/scripts/test/MariReduction.py` runs with no problem.
If users are satisfied :shipit:

PR #17860, opened 2016-10-19
Move Property implementation out of headers.

### LINKED ISSUE #17859: Move Property template implementations from headers into sources
This could be another small step in improving compile times.

SimonHeybrock:
This is a small step that helps to reduce the number of includes. The templates `PropertyWithValue` and `WorkspaceProperty` are now not defined in their header files anymore. Instead we are using explicit instantiations.

These changes reduce the average number of distinct recursively included files for `cpp` files in `Framework/` from 110 to 107.

**To test:**

In principle passing builds and tests should indicate whether this is ok or not, but I'd prefer if someone could do a code review.

Fixes #17859.

---
#### Reviewer

##### Code Review
##### Functional Tests


SimonHeybrock:
@OwenArnold I resolved some merge conflicts, if builds pass I think this is ready.

Framework/Algorithms/src/NormaliseByCurrent.cpp
```diff
@@ -42,15 +44,14 @@ double NormaliseByCurrent::extractCharge(
   int nPeriods = 0;
   try {
     Property *nPeriodsProperty = run.getLogData("nperiods");
-    Kernel::toValue<int>(nPeriodsProperty->value(), nPeriods);
+    nPeriods = boost::lexical_cast<int>(nPeriodsProperty->value());
```
OwenArnold: with c++11 you can also now do this with [tostring](http://en.cppreference.com/w/cpp/string/basic_string/to_string) but I don'r recommend changing anything.

OwenArnold:
These changes make sense to me. In the old days, when the framework was perceived to allow for user-provided workspaces and property types, the header implementations made sense. However, this is not the reality for the c++ side of the framework, so this is a better tradeoff.

OwenArnold:
:shipit:

rosswhitfield:
This has broken building with clang on Linux, you can't assume `__clang__` means OSX. I have a fix here #18213

PR #17866, opened 2016-10-20
Fix 'do not assign a lambda expression, use a def'

rosswhitfield:
This should fix all flake8 E731 warnings.

**To test:**
- Code review

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/scripts/SANS/isis_reduction_steps.py
+++ b/scripts/SANS/isis_reduction_steps.py
@@ -1754,7 +1754,7 @@ class DarkRunSubtraction(object):
 
         # Get the indices with settings which correspond to the individual settings
         def get_indices(time_flag, mon_flag):
-            return [i for i, val in enumerate(use_time) if use_time[i] == time_flag and use_mon[i] == mon_flag]
+            return [i for i, use_time_i in enumerate(use_time) if use_time_i == time_flag and use_mon[i] == mon_flag]
 
         indices_time_detector = get_indices(True, False)
         indices_time_monitor = get_indices(True, True)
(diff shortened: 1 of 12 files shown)


scripts/SANS/isis_reduction_steps.py
```diff
@@ -1843,7 +1846,9 @@ def _get_final_setting_monitors(self, run_number, use_mean, use_time, mon_number
                     monitor_mon_numbers.append(mon_numbers[index])
 
         # Check if the mean value is identical for all entries
-        are_all_same = lambda a_list : all([a_list[0] == a_list[i] for i in range(0, len(a_list))])
+        def are_all_same(a_list):
```
AndreiSavici: It is used only once. Try instead `len(set(a_list))==1`

scripts/SANS/isis_reduction_steps.py
```diff
@@ -1753,15 +1753,18 @@ def _evaluate_settings(self):
             run_number.append(setting.run_number)
 
         # Get the indices with settings which correspond to the individual settings
-        get_indices = lambda time_flag, mon_flag : [i for i, val in enumerate(use_time)
-                                                    if use_time[i] == time_flag and use_mon[i] == mon_flag]
+        def get_indices(time_flag, mon_flag):
+            return [i for i, val in enumerate(use_time) if use_time[i] == time_flag and use_mon[i] == mon_flag]
```
AndreiSavici: Are you using val anywhere?

rosswhitfield: No, doesn't look like it.

rosswhitfield: Although `use_time[i]` can be replaced with `val`

AndreiSavici:
LGTM :shipit:

Issue #16019, opened 2016-04-28
Indirect Elwin - Obtain unit from sample logs


ElliotAOram:
The Elwin interface/algorithm should automatically obtain the correct unit for the input workspace by checking the sample logs for the workspace

SpencerHowells:
At the moment the code assumes that the x-title is 'temperature'. If we want this to be general then the x-title should be the SE name.

PR #17867, opened 2016-10-20
Indirect - QuickRun Algorithm

### LINKED ISSUE #14616: IndirectQuickRun algorithm
The MSG have asked for an algorithm that does the following:
- [ ] Loads in `.raw` or `.nxs` files
  - [ ] if files are `.raw` reduce them using data reduction
- [ ] Runs the files through the `Elwin` algorithm 
- ~~Optional run through `Iqt`~~

The purpose of this is to get an overview of data quickly and in an automated way.

_Iqt option removed as it's input can not be produced automatically_

louisemccann:
MSG have asked for an EnergyWindowScan algorithm that reduces .raw files and then runs them through ElasticWindowMultiple and then an option for MSDFit. The aim of this is to speed up the workflow of Indirect algorithms. Also in this algorithm the Elastic Incoherent Structure Factor is calculated from the elastic and inelastic ranges.
EnergyWindowScan is a general algorithm that requires all the input parameters for EnergyTransfer, IndirectEnergyWindowScan is a bit more customised, this is the basis for implementation of an interface.

**To test:**

Algorithms > EnergyWindowScan 
InputFiles `irs26176`, Instrument=`IRIS`, Analyser=`graphite`, Reflection=`002`, SpectraRange=`3, 50`, ElasticRange=`-0.5, 0`, InelasticRange=`0, 0.5`, GroupingMethod=`All`

Repeat for IndirectEnergyWindowScan with same parameters, output should be the same.

Repeat for both algorithms with GroupingMethod=`Individual` and MSDFit checked.

Change instrument to `OSIRIS` and InputFiles `osi100318,osi100319,osi100320,osi100321,osi100322` replacing SpectraRange with `963, 1004`
Use RunNumbers `100318-100322` for IndirectEnergyWindowScan

Fixes #14616.

[Release Notes](https://github.com/mantidproject/mantid/blob/14616_IndirectQuickRun_algorithm/docs/source/release/v3.9.0/indirect_inelastic.rst)

---
#### Reviewer

##### Code Review
##### Functional Tests


reimundILL:
@louisemccann  Hi Louise, I noticed that both WorkflowAlgorithms have the same summary and in my opinion, IndirectEnergyWindowScan only saves and plots as an addition. My understanding of WorkflowAlgorithm was that there must be an OutputWorkspace set - but I did not find a documentation mentioning it.

louisemccann:
@reimundILL The IndirectEnergyWindowScan is a sort of 'placeholder' for an interface so that EnergyWindowScan doesn't have to be refactored too much when one is eventually made

reimundILL:
In `EnergyWindowScan` is a function `_rename_reduction` defined but I can't see where it is used.

reimundILL:
The code
``` Python

        # Disable sum files if there is only one file
        if len(self._data_files) == 1:
            if self._sum_files:
                logger.warning('SumFiles disabled when only one input file is provided.')
self._sum_files = False

```
may be irrelevant since `self._sum_files` is `False` anyways.

reimundILL:
You are calling other algorithms like `ISISIndirectEnergyTransfer`, which already validate the inputs. Maybe it is not needed to repeat validation in `EnergyWindowScan`. I noticed that the background will not be treated, is this desired?

reimundILL:
About the EISF, elastic incoherent structure factor, I would wish that you cite some document. It is an important factor and I found a different way of computing it. I see in `EnergyWindowScan` a simple Division. I read somewhere else something like ... experimental definition ... I elastic / (I elastic + I qens). There is as well an ... intermediate estimate ... EISF = I(Q,t=inf)/I(Q,t=0s) found in How to use Mantid for low energy inelastic neutron scattering data analysis on indirect geometry instruments written by Sanghamitra Mukhopadhyay. I guess that's what you are using.

louisemccann:
I think it is better to validate the inputs as they are put into the algorithm rather than coming across them half way through running it. The background is left out as this algorithm is meant to reduce the amount of things the users do - if they need background treatment this is something I can easily put in later.

reimundILL:
I started testing `EnergyWindowscan` with file `... mantid/build/ExternalData/Testing/Data/UnitTest/IRS26176.RAW`.

It is much easier for me to load files when using something like
``` Python
        self.declareProperty(MultipleFileProperty('Run', extensions=['nxs']),
                             doc='File path of run (s).')
```
instead of something like
```Python
        self.declareProperty(StringArrayProperty(name='InputFiles'),
                             doc='Comma separated list of input files')
```

There is a warning `Log file for run irs26176 not found
No temperature found for run: irs26176`.
May be a data related issue?

Basically, workspaces get created that contain only one entry, x-value is 0. Can it be usefull to put this information combined in a `TableWorkspace`?

Then, when I run the algorithm `IndirectEnergyWindowScan` with the same parameters I get errors:
``` Python
Error in execution of algorithm IndirectEnergyWindowScan:
invalid literal for int() with base 10: '... mantid/build/ExternalData/Testing/Data/UnitTest/IRS26176.RAW'
  at line 114 in ' ... IndirectEnergyWindowScan.py'
  caused by line 202 in ' ... IndirectEnergyWindowScan.py'
  caused by line 297 in ' ... IndirectEnergyWindowScan.py'
```

Further test for `GroupingMethod` and MSDfit works and gives same error for `IndirectEnergyWindowScan`

louisemccann:
`irs26176` is a test run and doesn't have associated temperature/logs. The OSIRIS runs have temperatures and when running them there should be more than one x value.

reimundILL:
I do not find compatible OSIRIS files for unit tests. All files available return errors. But I've seen that the output workspaces filled with nan's or zeros indeed have in general larger sizes.

louisemccann:
@raquelalvarezbanos could you check this with the OSIRIS runs?

SpencerHowells:
At the beginning of an experiment the use may perform a series of short runs over a temperature range. The purpose is to see if there is a change in the scattering and in particular of there is any quasielastic scattering. If there is then proper long runs can then be performed at appropriate temperatures chosen from the temperature variation.
The user selects the appropriate SE name for the temperature. If the Sample Logs does not contain the temperature the run number is used instead.
The EISF can be determined from the S(q,w) or I(q,t). Transforming to I(Q,t) requires care so is not appropriate for these scans. The correct way of doing it with S(Q,w) is to separate the elastic & quasielastic components using a convolution fitting procedure - another procedure not appropriate for these scans. the EISF is then the ratio of elastic to total scattering.
A suitable reference is probably the book Quasielastic Neutron Scattering by Bee.
Instead it integrates over the energy range of then elastic peak and over the whole range for the total scattering. The elastic integral will of course contain some quasielastic scattering, but is not important for the aim of the scan.
A choice of energy range for the total scattering is variable because of a request from Users in that that it may help for some samples.

reimundILL:
Thank you for this helpful comment. It makes me understand experiment and its dedicated code better. In particular, I was not aware that run numbers are used instead of temperature. For the ILL code there was the wish to either use temperature, time, or some other physical quantity like pressure.

SpencerHowells:
Issue #16019 is an update so that the Elwin interface/algorithm would automatically obtain the correct unit for the input workspace by checking the sample logs for the workspace. This was thought of when doing an experiment with pressure variation!

reimundILL:
The variable `_workspace_names` in `energyWindowScans` maybe can be deleted (it is and remains an empty list).

Unit tests pass and I could successful test with a single OSIRIS file `osi89757`.
Doc test pass as well.

reimundILL:
From my point of view, I stopped the code review and tests. I do not know about your plans, if you wish to do some last changes or not.

The algorithms are in general of good quality and will help your users. I was happy to learn about your data analysis.

louisemccann:
Sanghamitra has asked that the name of the algorithm is changed back to IndirectQuickRun. I will also add in contingencies for combinations of conditions that are incompatible. These should be the last changes

raquelalvarezbanos:
@louisemccann : I've tried the `OSIRIS` files again but I still get the same error when running `EnergyWindowScan`. I may be doing something wrong: this is how I am running it:

energy-window-scan

and the error is:

```
Error in execution of algorithm EnergyWindowScan:
Index out of range
  at line 265 in '/home/raquel/mantid/mantid/Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/ISISIndirectEnergyTransfer.py'
  caused by line 603 in '/home/raquel/mantid/mantid/scripts/Inelastic/IndirectReductionCommon.py'
  at line 152 in '/home/raquel/mantid/mantid/Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/EnergyWindowScan.py'
```

`IndirectQuickRun` works both for `GroupingMethod=All` and `GroupingMethod=Individual` with `MSDFit` enabled.

louisemccann:
@raquelalvarezbanos The instrument should be OSIRIS

raquelalvarezbanos:
There must be something wrong with my build as I cannot make it work (Ubuntu). But Louise has tested it on a VM and it works for her.

reimundILL:
Can the file `IndirectEnergyWindowScan-v1.rst` be deleted? It got replaced by `IndirectQuickRun-v1.rst`?

louisemccann:
@reimundILL Yes, the name of the algorithm has been changed to IndirectQuickRun

reimundILL:
I see that a file `docs/source/algorithms/IndirectEnergyWindowScan-v1.rst` exists. **I think it should be deleted**. The algorithm `IndirectEnergyWindowScan` does not exist anymore. There are already two docs .rst files (EnergyWindowScan and IndirectQuickRun)

louisemccann:
Ah, sorry, thank you for clarifying

reimundILL:
No problem at all.
Everything looks good, in case all tests pass :shipit:

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/IndirectQuickRun.py
```diff
@@ -0,0 +1,284 @@
...
+            SaveNexusProcessed(InputWorkspace=self._scan_ws + '_msd',
+                               Filename=msd_path)
+            msd_fit_path = os.path.join(workdir, self._scan_ws + '_msd_fit.nxs')
+            logger.information('Creating file : %s' % msd_fit_path)
+            SaveNexusProcessed(InputWorkspace=self._scan_ws + '_msd_fit',
+                               Filename=msd_fit_path)
+
+    def _plot_result(self):
+        import mantidplot as mp
+        nhist = mtd[self._scan_ws + '_el_eq1'].getNumberHistograms()  # noqa
```
rosswhitfield: Why all the `# noqa`, these all look easily fixable

PR #17875, opened 2016-10-24
Add class DetectorInfo

### LINKED ISSUE #17791: Add wrapper class DetectorInfo
This is a prerequisite for direct masking support and an intermediate step to Instrument-2.0, which will eventually replace the internals of this class while maintaining the same interface.

SimonHeybrock:
This is part of Instrument-2.0 and provides a layer underneath `SpectrumInfo`. Currently `DetectorInfo` is merely a wrapper class around the existing instrument code.

`SpectrumInfo` not computes its data based on `DetectorInfo` in some cases.

Some hints for explanation and guidance as to what to watch out for during code review:
- I managed to remove the reinstantiation of `SpectrumInfo` for every call to `MatrixWorkspace::spectrumInfo()`. Instead there is now a virtual method `ExperimentInfo::invalidateInstrumentReferences()` that should catch all bad cases, in particular a change of the `Instrument` or `ParameterMap` pointer (both are cached in `SpectrumInfo`, as a parameterized instrument pointer). Please have a really close look here if this mechanism has any holes.
- Another problem is access of the `ParameterMap` for modification: It has a copy-on-write mechanism, which would be fine in principle, but the problem is that `ExperimentInfo::getInstrument()` returns a parameterized instrument, i.e., it increases the reference count to the `ParameterMap`. Therefore, if you have a look at the algorithms that access the map, you sometimes see comments that refer to calling `instrumentParameters()` _before_ `getInstrument()` to avoid a copy. Due to this "mechanism" it is quite tricky to get a sane reference to the map in `SpectrumInfo`, see the long code comments.
- To avoid repeated calls to `MatrixWorkspace::getDetector()` (or `Instrument::getDetector()`), there currently is a call to `DetectorInfo::setCachedDetector()` in several methods of `SpectrumInfo`. This is a workaround and will be removed once we cleaned up more parts of the instrument.
- There currently is a translation mechanism for `detid_t` as part of `DetectorInfo`, that will also be removed eventually.
- Both `SpectrumInfo` and `DetectorInfo` currently need to be part of `API`, but once they do not depend on `MatrixWorkspace` and `Instrument` anymore they will be moved to their own module.

**To test:**
Thorough code review.
Fixes #17791.

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -180,35 +180,76 @@ const std::string MatrixWorkspace::getTitle() const {
...
   return *m_spectrumInfo;
 }
 
+/** Return a const reference to the DetectorInfo object.
+ */
+const DetectorInfo &MatrixWorkspace::detectorInfo() const {
+  return spectrumInfo().detectorInfo();
+}
+
+/** Return a non-const reference to the SpectrumInfo object. Not thread safe.
```
ianbush: Comment should say DetectorInfo object

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -180,35 +180,76 @@ const std::string MatrixWorkspace::getTitle() const {
...
+  // responsible for are released. Creating SpectrumInfo with a non-const
+  // reference to a MatrixWorkspace will cache to result of
+  // ExperimentInfo::instrumentParameters() which will later be used by
+  // modifications. This will trigger a copy if required. SpectrumInfo will
+  // obtain the parameterized instrument after obtaining the ParameterMap such
+  // that both refer to the same instance. This is the same method existing
+  // algorithms use to avoid an extra copy (call
+  // MatrixWorkspace::instrumentParameters() then
+  // MatrixWorkspace::getInstrument()).
+  m_spectrumInfo = nullptr;
```
ianbush: Is this required in case `make_unique<SpectrumInfo>(*this)` fails?

SimonHeybrock: No:

`SpectrumInfo` caches a parameterized copy of the instrument (obtained via `MatrixWorkspace::getInstrument()`). The reference count to the `ParameterMap` is thus at least 2 (1 from the workspace, 1 from `SpectrumInfo`).

Below we want to create a `SpectrumInfo` from a non-const workspace. That will get the non-const parameter map from the workspace. If the reference count is larger than 1 this will make a copy of the parameter map (copy-on-write).

To avoid unnecessary copies we are thus first releasing the old `SpectrumInfo`, dropping the reference count to the parameter map by 1. If the workspace is the sole remaining owner `make_unique<SpectrumInfo>(*this)` will then NOT trigger a copy.

ianbush: Right, I understand now.

SimonHeybrock: Would it maybe be possible to make `SpectrumInfo` the only owner of the `ParameterMap`, i.e., remove it from `ExperimentInfo`?

ianbush: I am not sure, but I can not see any reason why not on a quick look. That would be a step in the right direction too.

Framework/API/src/DetectorInfo.cpp
```diff
@@ -0,0 +1,233 @@
...
+  size_t thread = static_cast<size_t>(PARALLEL_THREAD_NUMBER);
+  if (m_lastIndex[thread] != index) {
+    m_lastIndex[thread] = index;
+    m_detectors[thread] = m_instrument.getDetector(m_detectorIDs[index]);
+  }
+
+  return *m_detectors[thread];
+}
+
+/// Sets the cached detector. This is an optimiyation used by SpectrumInfo.
```
ianbush: Typo in optimisation

Framework/API/src/DetectorInfo.cpp
```diff
@@ -0,0 +1,233 @@
...
+/// Returns a sorted vector of all detector IDs.
+const std::vector<detid_t> DetectorInfo::detectorIDs() const {
+  return m_detectorIDs;
+}
+
+const Geometry::IDetector &DetectorInfo::getDetector(const size_t index) const {
+  size_t thread = static_cast<size_t>(PARALLEL_THREAD_NUMBER);
+  if (m_lastIndex[thread] != index) {
+    m_lastIndex[thread] = index;
+    m_detectors[thread] = m_instrument.getDetector(m_detectorIDs[index]);
```
ianbush: I would rename `m_detectors`  to `m_lastDetectors` - this makes it clearer this is for caching.

ianbush:
> I managed to remove the reinstantiation of SpectrumInfo for every call to MatrixWorkspace::spectrumInfo(). Instead there is now a virtual method ExperimentInfo::invalidateInstrumentReferences() that should catch all bad cases, in particular a change of the Instrument or ParameterMap pointer (both are cached in SpectrumInfo, as a parameterized instrument pointer). Please have a really close look here if this mechanism has any holes.

I could not see any problems with this. `invalidateInstrumentReferences()` seems to be called everywhere that it is required.

> Another problem is access of the ParameterMap for modification: It has a copy-on-write mechanism, which would be fine in principle, but the problem is that ExperimentInfo::getInstrument() returns a parameterized instrument, i.e., it increases the reference count to the ParameterMap. Therefore, if you have a look at the algorithms that access the map, you sometimes see comments that refer to calling instrumentParameters() before getInstrument() to avoid a copy. Due to this "mechanism" it is quite tricky to get a sane reference to the map in SpectrumInfo, see the long code comments.

This looks like it should work to achieve this. It might be worth adding more comments as per what you explained above, as it takes a bit of thinking about.

> To avoid repeated calls to MatrixWorkspace::getDetector() (or Instrument::getDetector()), there currently is a call to DetectorInfo::setCachedDetector() in several methods of SpectrumInfo. This is a workaround and will be removed once we cleaned up more parts of the instrument.

This looks good.

> There currently is a translation mechanism for detid_t as part of DetectorInfo, that will also be removed eventually.

This looks good too as a current workaround.

Framework/API/src/MatrixWorkspace.cpp
SimonHeybrock: Having had another look: This is not quite as simple as I thought. We would still have parameterized instruments which implicitly increase the reference count. I don't see a simple way of working around this --- we would still need to juggle with resetting all references to the parameter map by such instrument copies. It is probably easier and more efficient to postpone such changes until the point where we can fix it properly (after a lot of other fundamental changes).

Additionally I would also have a concern regarding MDWorkspaces: `MultipleExperimentInfos` is used in that case, but currently I don't have any settled ideas on how to deal with `SpectrumInfo` in the context of MDWorkspaces, so we might get ourselves into a mess.

SimonHeybrock:
@ianbush I just had a discussion with @OwenArnold: As I mentioned earlier, MD-Workspaces do also access detectors, [including position information](https://github.com/mantidproject/mantid/blob/master/Framework/MDAlgorithms/src/ConvertCWPDMDToSpectra.cpp#L376). Therefore, since we intend to move position information out if the parameter map into `DetectorInfo`, we need to make sure that MD-Workspaces also have a `DetectorInfo`.

The most natural way seems to be to move `DetectorInfo` into `ExperimentInfo`. MD-Workspaces can then access it via their base class `MultipleExperimentInfos`. `MatrixWorkspaces` will add the `SpectrumInfo` object on top of `DetectorInfo`. For MD-Workspaces there is no equivalent for `SpectrumInfo` (at least for now).

Do you see anything wrong with that?

We could make this change right away, but maybe doing so in a separate PR would be better?

ianbush:
These changes all look good :shipit:

Framework/API/src/SpectrumInfo.cpp
```diff
@@ -43,10 +40,14 @@ bool SpectrumInfo::isMasked(const size_t index) const {
...
-  else
-    return getDetector(index).getDistance(getSource()) - l1();
+  double l2{0.0};
+  const auto &dets = getDetectorVector(index);
+  for (const auto &det : dets) {
+    const auto &detIndex = m_detectorInfo.indexOf(det->getID());
+    m_detectorInfo.setCachedDetector(detIndex, det);
+    l2 += m_detectorInfo.l2(detIndex);
+  }
+  return l2 / static_cast<double>(dets.size());
```
OwenArnold: Good. This is what we thought/had in the prototype.

Framework/API/test/DetectorInfoTest.h
```diff
@@ -0,0 +1,318 @@
...
+      inst->add(det);
+      inst->markAsDetector(det);
+      ws->getSpectrum(i).addDetectorID(static_cast<detid_t>(i));
+      if (i % 2 == 0)
+        pmap.addBool(det->getComponentID(), "masked", true);
+    }
+    return std::move(ws);
+  }
+};
+
```
OwenArnold: Would this be a good time to add a performance test suite? At least to help spot regressions.

Framework/API/src/SpectrumInfo.cpp
```diff
@@ -82,36 +79,50 @@ double SpectrumInfo::signedTwoTheta(const size_t index) const {
...
-  return getDetector(index).getPos();
+  // The body of this function is almost a 1:1 duplicate of
+  // DetectorGroup::getPos. The former should be cleaned up once subsequent
+  // changes allow for this, and the latter should be removed.
+  Kernel::V3D newPos;
+  const auto &dets = getDetectorVector(index);
+  for (const auto &det : dets) {
+    const auto &detIndex = m_detectorInfo.indexOf(det->getID());
+    m_detectorInfo.setCachedDetector(detIndex, det);
+    newPos += m_detectorInfo.position(detIndex);
```
OwenArnold: So I think this bit of code answers one of my questions - `DetectorInfo` can be accessed in a mutable fashion. How does the `SpectrumInfo` stay synchronized. We always compute from `DetectorInfo`.

Framework/API/inc/MantidAPI/DetectorInfo.h
```diff
@@ -0,0 +1,136 @@
...
+  void cacheSource() const;
+  void cacheSample() const;
+
+  // These cache init functions are not thread-safe! Use only in combination
+  // with std::call_once!
+  void doCacheSource() const;
+  void doCacheSample() const;
+  void cacheL1() const;
+
+  Geometry::ParameterMap *m_pmap;
```
SimonHeybrock: As long as we have parametrized instruments and parametrized detectors there will almost always be multiple owners of the `ParameterMap`. Once we managed to removed those we can think about who the actual owner should be. Right now I do not think that it should be `DetectorInfo`, since the parameter will actually mostly contain information about components that are *not* detectors.

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -892,7 +881,41 @@ ExperimentInfo::getInstrumentFilename(const std::string &instrumentName,
...
 
-//--------------------------------------------------------------------------------------------
+/** Return a const reference to the DetectorInfo object.
+ *
+ * Any modifications of the instrument or instrument parameters will invalidate
+ * this reference.
+ */
+const DetectorInfo &ExperimentInfo::detectorInfo() const {
+  if (!m_detectorInfo) {
+    std::lock_guard<std::mutex> lock{m_detectorInfoMutex};
```
SimonHeybrock: The method is `const` so I wanted to make it thread-safe. Admittedly, `MatrixWorkspace::detectorInfo()` should typically be called outside threaded contexts, but there are exceptions.

Framework/API/src/SpectrumInfo.cpp
```diff
@@ -82,36 +79,50 @@ double SpectrumInfo::signedTwoTheta(const size_t index) const {
...
+  }
+
+  // We can have very small values (< Tolerance) of each component that should
+  // be zero
+  if (std::abs(newPos[0]) < Mantid::Kernel::Tolerance)
```
SimonHeybrock: See the code comment a few lines above. I just copied this from `DetectorGroup`. I would also prefer removing it.

Framework/API/src/SpectrumInfo.cpp
OwenArnold: 👍

PR #17879, opened 2016-10-24
Hfir dlambda

### LINKED ISSUE #17762: HFIR SANS Resolution
**Email from users**

We just found out that the delta q (4th column) is a few times higher than the actual value. We need to correct that asap.
I suspect it is due to the change made earlier on the wavelength spread: from meta data, <wavelength_spread type="FLOAT32">0.132353</wavelength_spread> now is used correctly in giving out the correct wavelength spread from ~5.60 A to 6.40 A by doing 6A*0.13~=0.8A, but when the change was made for that, the delta q calculation (instrument resolution) should be using 0.13, the ratio, not the 0.8A.

**Mathieu email**

There seems to be an inconsistency between `LoadSpice2D` and `ReactorSANSResolution`.

Files:

https://github.com/mantidproject/mantid/blob/master/Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/ReactorSANSResolution.py

The wavelength-spread is stored as Angstrom but the resolution calculator expects it as the ratio d(lambda)/lambda.

``` python
addRunProperty<double>("wavelength-spread", m_dwavelength, "Angstrom");
```

I suggest we change it in the loader and store as a ratio. I want to make sure we don’t break anything else though, so let’s talk when you are back.

**Notes**:

It might be better to add to `LoadSpice2D`:

``` python
addRunProperty<double>("wavelength-spread-ratio", m_dwavelength);
```

And check if this property exists in `ReactorSANSResolution`. If so, use it rather than `wavelength-spread`.

Needs further investigation.

ricleal:
See: #17762

New property added "wavelength-spread-ratio" and there was a "/4.0" missing in the formula.

**To test:**

Tests should pass.
Users have tested and are already happy:

> Great! Push that to the nightly build. It is correct now.
> I used wrong numbers in my calculation. Sorry about the confusion!

Fixes #17762

No release notes.

---
#### Reviewer

##### Code Review
##### Functional Tests

### DIFF
--- a/Framework/DataHandling/src/LoadSpice2D.cpp
+++ b/Framework/DataHandling/src/LoadSpice2D.cpp
@@ -517,7 +517,8 @@ void LoadSpice2D::setMetadataAsRunProperties(
 
   addRunProperty<double>("wavelength", m_wavelength, "Angstrom");
   addRunProperty<double>("wavelength-spread", m_dwavelength, "Angstrom");
-  addRunProperty<double>("wavelength-spread-ratio", m_dwavelength/m_wavelength);
+  addRunProperty<double>("wavelength-spread-ratio",
+                         m_dwavelength/m_wavelength);
 
   addRunProperty<double>(metadata, "Counters/monitor", "monitor", "");
   addRunProperty<double>(metadata, "Counters/time", "timer", "sec");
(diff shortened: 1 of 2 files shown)


mdoucet:
Small and clear change. :shipit:

PR #17882, opened 2016-10-25
CorrectFlightPaths fix and rename

### LINKED ISSUE #17873: Fix CorrectFlightPaths for IN5
After the new histogram roll out CorrectFlightPaths has been left broken. The problem is in this commit, with the solution noted - https://github.com/mantidproject/mantid/commit/117bcd4c018ada5237b5898b24688874d17ed3ac#diff-0c3df4bb2b11e7886758f205589f7476.

This ticket should provide a fix and add some basic unit tests to stop it breaking in the future.

This algorithm is specific to IN5 at the ILL and not yet used, so does not need to go in a patch release.

ianbush:
CorrectFlightPaths had a bug after the recent histogram changes. This meant that the spectra in the output workspace were empty. This PR fixes that, and adds a unit test.

The algorithm is also renamed to ConvertToConstantL2 and the documentation updated to state that this is not intended as part of any reduction process for IN5.

**To test:**

Check that the file `UnitTest/ILL/IN5/104007.nxs` can be loaded and run via ConvertToConstantL2. Afterwards looking at the detector table the detectors should all have the same distance to the sample.

Running via CorrectFlightPaths should give the same results but the deprecated warning.

Documentation should show CorrectFlightPaths is deprecated, and documentation should be updated for ConvertToConstantL2.

Fixes #17873.

---
#### Reviewer

##### Code Review
##### Functional Tests


docs/source/algorithms/ConvertToConstantL2-v1.rst
```diff
@@ -0,0 +1,32 @@
...
+
+.. properties::
+
+Description
+-----------
+
+.. warning::
+
+   This algorithm will be deprecated in the next version of Mantid. Please, use :ref:`algm-ConverToConstantL2` instead, which
+   is the new name for this algorithm.
```
raquelalvarezbanos: Should this sentence be in `CorrectFlightPaths-v1` instead of here? Also, I noticed the documentation does not include a user example, would it make sense to add one?

ianbush: Yes, it should! It also seems I had not set the deprecated algorithm quite correctly, the documentation should show up correctly now.

I have also added the usage example, but in doing so I am a bit suspicious a monitor is being moved that should not be, so I will mark this as in progress for now and take a look again next week.

raquelalvarezbanos: OK, thanks Ian. Just let me know when it's ready to be tested again.

ianbush:
#17954 (work in progress) needs to be fixed before changes in this PR can continue.

ianbush:
@raquelalvarezbanos This should be ready to check again, once the system tests pass.

The deprecated warnings should now all be in the correct place and the usage example has been added. There is also a unit test added to check the monitors do not move.

raquelalvarezbanos:
OK. It seems to work as before, except for the last modifications. For instance, I can see now in the detector table that monitor has not been moved (`R=0.5`), the new unit test will ensure this.

Looks good to me.

:shipit:

PR #17885, opened 2016-10-26
Crystal field instrument resolution

### LINKED ISSUE #17517: Implement fixing of peak widths using instrument resolution for Crystal Field fitting
This issue is part of #15684 

It was identified in an implementation meeting that the current fits do not constrain the peak widths enough resulting in poor fits to measured data. In principle, the measured peaks are a convolution of an intrinsic peak (a delta function for an isolated ion, but with finite lifetimes in a crystal or further broadening due to dispersion) and an instrumental peak. The intrinsic widths will tend to be small and may sometimes be ignored, whilst the instrumental peak width can be calculated, e.g. using `PyChop`. 

It would make the fits much more stable if all the peak parameters can be fixed, so that only the model (crystal field) parameters are varied. The crystal field parameters determine the peak area and centre positions, so at present the peak widths are varied (simulataneously with the crystal field parameters, hence leading to some instability) to fit the data. This issue is to implement a way to fix the peak widths from a measured or calculated instrument resolution function.

mantid-roman:
Implemented dynamic boundary constraints on peak widths in fitting to a crystal field spectrum. The boundaries depend on the peak centre via a user provided function.

**To test:**

Run this script

```
from CrystalField import CrystalField, CrystalFieldFit
origin = CrystalField('Ce', 'C2v', B20=0.37, B22=3.97, B40=-0.0317, B42=-0.116, B44=-0.12,
              Temperature=44.0, FWHM=1.0)
origin.peaks.param[0]['FWHM'] = 1.01
origin.peaks.param[1]['FWHM'] = 1.1
origin.peaks.param[2]['FWHM'] = 1.8

x, y = origin.getSpectrum()
ws = makeWorkspace(x, y)

cf = CrystalField('Ce', 'C2v', B20=0.37, B22=3.97, B40=-0.0317, B42=-0.116, B44=-0.12,
                  Temperature=44.0, FWHM=1.0, ResolutionModel=([0, 70], [1, 2]), FWHMVariation=0.1)

fit = CrystalFieldFit(Model=cf, InputWorkspace=ws)
fit.fit()

res_width = lambda x: x / 70.0 + 1.0
pars = mtd['fit_Parameters']

c1 = pars.cell(38, 1)
w1 = pars.cell(39, 1)
print c1, w1, res_width(c1)

c2 = pars.cell(41, 1)
w2 = pars.cell(42, 1)
print c2, w2, res_width(c2)
```

Check that `w1` and `w2` are within `FWHMVariation` of values calculated with `res_width`.

Fixes #17517.

Release notes updated in direct_inelastic.rst

---
#### Reviewer

##### Code Review
##### Functional Tests


scripts/Inelastic/CrystalField/function.py
```diff
@@ -489,3 +489,105 @@ def update(self, func1, func2=None):
...
+                    model.
+        :param xstart:
+        :param xend:
+        :param accuracy: (Optional) If given and model argument contains functions it's used
+                    to tabulate the functions such that linear interpolation between the
+                    tabulated points has this accuracy. If not given a default value is used.
+        """
+        from inspect import isfunction
+        self.multi = False
+        if isfunction(model):
```
mducle: Could we change this check (and also in line 527) to `isfunction(model) or ismethod(model)`? Because, I would also like to pass the bound method `getResolution` of the `PyChop2` object to `ResolutionModel`, without using a `lambda` - at the moment you have to do:

```
from PyChop import PyChop2
from CrystalField import CrystalField, ResolutionModel
chop_res = PyChop2('MARI', 'S', 250)
chop_res.setEi(40)
res_mod = ResolutionModel(lambda x: chop_res.getResolution(x), xstart=-10, xend=35, accuracy=0.1)
cf = CrystalField('Ce', 'C4v', B20=0.451, B40=0.025, B44=0.317, Temperature=4, 
                  ResolutionModel=res_mod, WidthVariation=0.1)
```

whereas it would be nicer for the user if:

```
res_mod = ResolutionModel(chop_res.getResolution, xstart=-10, xend=35, accuracy=0.1)
```

worked.

mantid-roman: OK

mducle:
I'm getting some errors: `Cannot calculate peak width: peak at 27.699299 is outside given range (-10.000000 : 24.000000)` - this is because I'm using a dataset with Ei=25meV which means that an energy transfer of 27.7 meV is not physically measureable. Hence, the resolution model cannot give the width at 27.7 meV. 

I'm not sure what to do about this. I think there are two options: 

1. Extrapolate from previous width vs energy.
2. Ignore peaks with energy outside the defined range of the resolution model (e.g. set peak width to be infinity / peak intensity to zero). 

I think that option 2 makes more sense, because if you've defined a resolution model, the presumably it's based on some physical instruments with physical restrictions; but I'll ask Adroja and Tatiana and see what they think.

mducle (changes requested):
So, after discussion with Tatiana, I think we should ignore peaks which are outside the resolution model range (set their intensity to zero). We considered extrapolation as too inaccurate.

For the direct geometry instruments, we think we will almost always use `PyChop` to generate the resolution function, so the only energy range which cannot be calculated is when the energy transfer is larger than the incident energy, which not measureable - in which case the correct behaviour would be to ignore those peaks.

So, could you change `calculateWidth` to not throw an error, but to return say `NAN` and then for `buildSpectrumFunction` / `updateSpectrumFunction` to detect this and then ignore this peak, like it does for peaks with too small intensity.

mducle (approved):

mducle:
Changes works well for me. Assuming the system tests pass then :shipit:

PR #17889, opened 2016-10-26
New algorithms `BinWidthAtX` and `MedianBinWidth`

### LINKED ISSUE #17802: `RebinToSmallestBin` wrapper algorithm for automatic rebinning
In the use cases foreseen at the ILL, automatic rebinning to the minimum bin width should provide readily acceptable binning to e.g. the rugged workspaces resulting from calling the [`SofQW`](http://docs.mantidproject.org/nightly/algorithms/SofQW-v1.html) algorithm.

soininen:
These utility algorithms give the bin width at selected bin (`BinWidthAtX`) or the median over all spectrum's bins (`MedianBinWidth`). The information could be used, for example, as a good initial guess for the [`Rebin`](http://docs.mantidproject.org/nightly/algorithms/Rebin-v1.html) algorithm and the like. The algorithms also provide optional "rounding" for the bin widths.

Originally this development branch was to implement what is described in #17802 but turns out the users had a bit different needs and this PR tries to answer for those. Thus, this closes #17802 as a side effect.

**To test:**

A throughout stress testing is desirable. Maybe there are some corner cases that are not taken into account by the algorithms. 

**Release notes were updated accordingly.**

---
#### Reviewer

##### Code Review
##### Functional Tests


Framework/PythonInterface/plugins/algorithms/BinWidthAtX.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def PyExec(self):
+        '''
+        Averages the bin widths at X.
+        '''
+        inputWs = self.getProperty(self._PROP_INPUT_WS).value
+        x = self.getProperty(self._PROP_X_VALUE).value
+        roundingMode = self.getProperty(
+            roundinghelper.PROP_NAME_ROUNDING_MODE).value
+        inputIsPointData = inputWs.isDistribution()
```
AndreiSavici: The output of `isDistribution()` does not mean that the workspace has point data. Use `isHistogram()`

Framework/PythonInterface/plugins/algorithms/BinWidthAtX.py
```diff
@@ -0,0 +1,87 @@
...
+        Declares algorithm's properties.
+        '''
+        self.declareProperty(
+            MatrixWorkspaceProperty(name=self._PROP_INPUT_WS,
+                                    defaultValue='',
+                                    direction=Direction.Input),
+            doc='The workspace containing the input data')
+        self.declareProperty(
+            name=self._PROP_X_VALUE, defaultValue=0.0,
+            direction=Direction.Input, doc='')
```
AndreiSavici: You should add a docstring

Framework/PythonInterface/plugins/algorithms/MedianBinWidth.py
```diff
@@ -0,0 +1,78 @@
...
+                             doc='The averaged median bin width')
+
+    def PyExec(self):
+        '''
+        Averages the median bin widths of the input workspace.
+        '''
+        inputWs = self.getProperty(self._PROP_INPUT_WS).value
+        roundingMode = self.getProperty(
+            roundinghelper.PROP_NAME_ROUNDING_MODE).value
+        inputIsPointData = inputWs.isDistribution()
```
AndreiSavici: Same comment: `isDistribution()==True` does not mean that the workspace contains point data

AndreiSavici:
Changing from point data to histogram and back is OK only if you have equally spaced bins
```
import numpy as np
a=range(5)*10
a=np.array(a)
a*=a
w=CreateWorkspace(a,a,Nspec=10)
w1=ConvertToHistogram(w)
w2=ConvertToPointData(w1)
print CompareWorkspaces(w,w2)
```

Framework/PythonInterface/plugins/algorithms/BinWidthAtX.py
soininen: Will do.

soininen:
These algorithms were developed [`Rebin`](http://docs.mantidproject.org/nightly/algorithms/Rebin-v1.html) in mind, and that does the point data -> histogram conversion. However, maybe it is better to keep these algorithms simple and leave the responsibility of the conversion to somebody else. So, I have now removed the conversion and the algorithms accept histogram input only.

Framework/PythonInterface/plugins/algorithms/MedianBinWidth.py
```diff
@@ -0,0 +1,73 @@
...
+    def PyExec(self):
+        '''
+        Averages the median bin widths of the input workspace.
+        '''
+        inputWs = self.getProperty(self._PROP_INPUT_WS).value
+        roundingMode = self.getProperty(
+            roundinghelper.PROP_NAME_ROUNDING_MODE).value
+        n = inputWs.getNumberHistograms()
+        medians = numpy.empty(n)
+        for wsIndex in range(n):
```
AndreiSavici: You can do without the for loop, only with numpy functions:
```
x2D=inputWs.extractX()
dx2D=numpy.diff(x2D)
medians=numpy.median(dx2D,axis=1)
```

Framework/PythonInterface/plugins/algorithms/roundinghelper.py
```diff
@@ -0,0 +1,37 @@
...
+                      defaultValue=ROUNDING_NONE, validator=rounding,
+                      direction=Direction.Input, doc='Bin width rounding')
+
+
+def round(x, mode):
+    '''
+    Rounds x depending on the rounding mode selected.
+    '''
+    if mode == ROUNDING_TEN_TO_INT:
+        return 10.0**numpy.floor(numpy.log10(x))
```
AndreiSavici: I did not see a check for x<=0

AndreiSavici:
@soininen please also merge the master into your branch

Framework/PythonInterface/plugins/algorithms/MedianBinWidth.py
soininen: Interesting approach. However, this could consume quite a lot of memory as all *X* values are copied to `x2D`, no? Is there a horrible performance impact in using the for loop here? If not, I would leave it as is.

In any case, I'll change the `dxs = xs[1:] - xs[:-1]` to `dxs = numpy.diff(xs)`.

Framework/PythonInterface/plugins/algorithms/roundinghelper.py
soininen: Indeed. This brought me to another point, that these algorithms should be agnostic to the ordering of the bin boundaries, i.e. give meaningful results regardless of whether the x axis is ascending or descending. Unfortunately, turns out that `binIndexOf()` in `MatrixWorkspace` doesn't work with descending binning neither.

I opened an issue #18020 for this. I think I need to fix that issue before I can finish this one.

Framework/PythonInterface/plugins/algorithms/MedianBinWidth.py
AndreiSavici:
```
import numpy as np

x=Load(Filename='/SNS/HYS/IPTS-16818/shared/autoreduce/4pixel/HYS_149202_4pixel.nxspe')
def loopway(ws):
    n=ws.getNumberHistograms()
    medians=np.empty(n)
    for wsIndex in range(n):
        xs=ws.readX(wsIndex)
        dxs=xs[1:]-xs[:-1]
        medians[wsIndex]=np.median(dxs)
    return medians

def npway(ws):
    x2D=ws.extractX()
    dx2D=np.diff(x2D)
    medians=np.median(dx2D,axis=1)
    return medians
```
I did some timing tests. The NXSPE file is 5120 spectra, 442 bins
``` 
%timeit loopway(x)
1 loop, best of 3: 317 ms per loop
%timeit npway(x)
10 loops, best of 3: 35.9 ms per loop
```
Almost a factor of 10 increase in speed

soininen: Real testing to actually get some solid data? Okay, I tried this with data from the IN5 instrument which has 98304 spectra and 1024 bins. Mantid's `getMemorySize()` returns 2415953776 for this; Python's `meliae.scanner.get_recursive_size()` reports that `x2D` consumes 806101000 bytes. 

For timing, I get the 10 times faster speeds of `npway` for small workspaces. For the IN5 data, the speedup is less, 3.7 vs. 0.9 seconds, so 4x faster.

As a side note, using `numpy.diff()` is slightly slower than `x[1:] - x[:-1]` in the for loop.

By looking at the numbers, I think I'll implement this in the npway, as you suggested. If we run into memory problems with IN5, perhaps we need to rewrite `MedianBinWidthAtX` in C++.

AndreiSavici: If `numpy.diff` is slow, you can do `x2D[:,1:]-x2D[:,:-1]`. The idea is to avoid for loops (slow)

soininen: I timeited the npway without `numpy.diff` and it is as fast as with `numpy.diff`. For some reason, there is a performance impact when `numpy.diff` is run in the for loop.

soininen:
@AndreiSavici Changed the median calculation as you requested and merged master.

AndreiSavici:
Things look OK. :shipit:

Issue #17831, opened 2016-10-14
Umbrella: Powder diffraction instruments migration to PEARL scripts


DavidFair:
This umbrella issue holds the high level steps needed for the migration of powder diffraction instruments at ISIS to the PEARL powder diffraction script.

**Note this ticket will update and change as items are added to the roadmap**

### Completed:
 Prep work:
~~#17839~~
### Refactoring Work:
~~#18108~~

---
## All Instruments
#19113 - Unit tests
#19105 - Switch to `AlignAndFocusPowder`
- [x] Update release notes when we are happy with the scripts. 
- [x] Ensure system tests tidy up after themselves
- [x] Abstract instrument should have a default behaviour which instruments then override - e.g. `NormaliseByCurrent` this should help reduce the amount of boilerplate written in each instrument class.
- [x]  Create an example calibration folder for each instrument that shows the file structure
- [x] Include the mandatory parameters for each method in the documentation 
- [x] Ensure everything in `isis_powder` is Python 3 ready - see #18550 

# POLARIS and PEARL

- [x] Add option to specify ext (single file only) for loading intermediate files

### PEARL:
- [x] Add method for creating .cal file at a later date
- [x] We should move the remaining hardcoded values into the adv. config file

### POLARIS:
~~#17949~~
- [x] As part of above at a later date focused runs should also be grouped possible 

# IMAT & GEM
#18645 - IMAT and ENGIN-X
#18643 - GEM
## IMAT
- [ ] Refactor the Engineering Diffraction GUI to include IMAT

# Low priority
- [ ] Give the option to use a flat directory for the calibration directory - POLARIS
- [x] Some scientists require GEM to have its banks further subdivided according to a grouping file. This should be trivial to add to GEM once everything else is in place
- [x] Various workspaces leak out of the script - once all work is completed we should identify where we are not deleting workspaces (usually because the label in Python gets reassigned to a new workspace)
- [x] Paths should expand path macros when encountered
- [ ] Possibly look into detecting when a run has been completed and processing that. Data Reduction Service?

DavidFair:
As this is an ongoing process moving the ticket to 3.10

Anders-Markvardsen:
From discussion with @josephframsay and @DanNixon this issue is out-of-date and requirements and priorities have changed since this issue was last worked on

PR #17899, opened 2016-10-27
Pearl Powder Diffraction Script - Prep work

### LINKED ISSUE #17839: PEARL Scripts - Prep work
Copy from umbrella ticket:
---
- [x] Move duplicated code (e.g. debug steps) into separate functions
- [x] Verify files are there before executing
- [x] Remove use of globals/hardcoded values where possible
- [x] Migrate functions so there is a hierarchy of calls with most abstract on top and the calls to Mantid API's are wrapped where possible
---

DavidFair:
Description of work.
As part of #17831 - Migrating Powder Diffraction instruments to use a common script.
This PR is to merge in a working implementation with old API support before starting the migration of additional instruments into the script. 

I am open to any suggestions or changes - however as part of the ongoing process I'll add them into the next PR instead of modifying this PR (unless they affect the data being generated) as this script is in a tested working state. 

This PR lays down the prep work for the next step, migration of POLARIS. It does the following:
- Removed all globals (except for a global debug flag and a small work around for the ADS overwriting similar names)
- Made there a single entry/exit point for various public calls
- Removed the requirement that certain APIs be called in an (undocumented) order to setup the directories
- Separated public and private API using notation 
- Implemented an Abstract instrument and PEARL specific overrides and hooks which helps contain instrument specific workarounds or parameters
- Completely moved and refactored the existing code out of `pearl_routines`
- Migrated everything to Python 2/3 (I may have missed one please let me know if I have)
- Rewrote `pearl_routines` as a wrapper around the new implementation and verified compatibility with existing tests
- Verified new data matched old data, then rewrote system tests to correctly use reference workspaces with the old API
- Implemented new unit tests for functions which are unit testable in both `AbstractInstrument` and `PEARL` instrument

---

There are still some steps still to do as part of future issues:
- Generate absorption corrections does not work as the existing implementation is incorrect and there are no unit tests to verify the behaviour. This will be fixed by working through the history of existing correction files.
- A user prefs file will be created and read for each instrument. This means that users can edit some of the advanced settings or add common settings (i.e. for their machine) in a separate script. This discourages modification to the current scripts. 
- Creating a silicon calibration has not data to verify the new API against so will need additional testing. 
- Using system tests directly calling the new API - there should be no difference in the algorithms only the entry point hence this isn't implemented yet. 
- Documentation - this is because the new API is still in flux especially when new instruments need to be added and the entry points might change

**To test:**
System tests should pass against the reference data

Run the following:

``` python
import PearlPowder_Pearl

calibration_dir =  *to_set*
raw_data_dir =  *to_set*
output_dir = *to_set*
tt_mode = "TT70"

pearl_obj = PearlPowder_Pearl.Pearl(user_name="Test", calibration_dir=calibration_dir,
                                    raw_data_dir=raw_data_dir, output_dir=output_dir, tt_mode=tt_mode)
output = pearl_obj.focus(run_number=92476, focus_mode="mods")
```

I recommend setting `calibration_dir` and `raw_data_dir` to 
`*SystemTestData*/PEARL/Focus_Test/Calibration/` and 
`*SystemTestData*/PEARL/Focus_Test/RawFiles/` respectively 
Also don't forget to set the output directory to your choice.

The paths **must** end with a `/` or `\\` character (remembering that `\` escapes so Windows paths need to be separated by `\\`

Fixes #17839 .

---
#### Reviewer

##### Code Review
##### Functional Tests


DavidFair:
Additionally the baseline warnings of Flake8 will need to be reduced by 6 warnings following these changes

rosswhitfield:
Please don't add `# noqa` just to hide flake8 warnings that could otherwise be fixed. Leave the warning there so we know it's something that should be looked at.

martyngigg:
I'm not entirely sure about the naming scheme here. Things like `PearlPowder_Pearl` seem a little repetitive at first glance.

Are these names and file structure open for negotiation? The ISIS reflectometry scripts live under the `Reflectometry` directory and in an `isis_reflectometry` package. Could we have a similar approach for diffraction? `Diffraction/isis_powder/...`. Then an import would be `from isis_powder import PEARL` or something?

DavidFair:
Thats a good idea - I was uneasy with the naming scheme but couldn't think of an alternative at the time.
How would we implement this and have separate `.py` files so that when the user does from `isis_powder import <instrument>` it loads `instrument.py`?

martyngigg:
The first thing that is required will be an `__init__.py` file in the `isis_powder` directory: https://docs.python.org/2/tutorial/modules.html#packages. If you look at the [reflectometry](https://github.com/mantidproject/mantid/tree/master/scripts/Reflectometry/isis_reflectometry) then you can see how that is organized.

They don't have different files per instrument but a common `quick` procedure in `quick.py`. Do we need separate `instrument.py` files?

DavidFair:
I wanted to keep a structure where we have a common script that performs all the steps and multiple classes for the instruments which override the behaviour of a hook when they need to do something specific or different. 
This means we need a class per instrument which implements the instrument specific logic so I thought about keeping them in separate files?

martyngigg:
Your example now looks good I think.

martyngigg:
Fine as initial prep work for improving the user scripts that we have captured.

There is ongoing work with the ISIS powder group as to exactly how this set of scripts that only supported PEARL will support all of them.

:shipit:

martyngigg:
@AndreiSavici Could you look at merging this?

AndreiSavici:
I just have a question: why must the path end with `\`? You should use `os.path.join` function to merge the path pieces together, with or without path separators in the input

DavidFair:
I realised after I submitted the PR that OS.path could do that for you. 
I've put that change into the next load of changes as it only affects people using the new style API which is still changing. *Pearl_routines* should hide this limitation in the mean time if that makes sense.

rosswhitfield:
`PythonScriptsTest_ISIS_Powder_AbstractInstTest` fails with python 3. @DavidFair can you look into that http://builds.mantidproject.org/job/python3/159/

PR #17901, opened 2016-10-27
Change PVPython tests to only check the major version number

rosswhitfield:
The check on paraview version is stricter than it needs to be, this makes it hard to work with newer versions of paraview. These tests are only really testing if `paraview.simple` is able to be imported, so checking just the major version number is more than enough.

**To test:**
- The test should now pass with any `5.*` paraview version

---
#### Reviewer

##### Code Review
##### Functional Tests


quantumsteve:
LGTM. :shipit:

PR #17910, opened 2016-10-31
Fix duplicate names for group workspace

### LINKED ISSUE #17903: Fix duplicate names when loading group workspace via LoadNexusProcessed
Loading a saved group workspace via a the algorihtm `LoadNexusProcssed` does not respect the workspace name which is saved in the file. This can lead to odd scenarios where duplicate names can appear in the ADS. It is not quite obvious why this feature exists in the first place ( Owen mentioned that it could be related to loading of the IDF).

The cause of this is related to the way we select the workspace names. We don't respect the saved name of the workspace name in the file.

Run this script

``` python
ws_1 = CreateSampleWorkspace()
ws_new_1 = CreateSampleWorkspace()
group_ws = GroupWorkspaces(InputWorkspaces=[ws_1, ws_new_1])
YOUR_FILENAME = "C:/Sandbox/test_file.nxs"
SaveNexus(Filename=YOUR_FILENAME, Inputworkspace=group_ws)
DeleteWorkspace(group_ws)
reloaded_ws = LoadNexusProcessed(Filename=YOUR_FILENAME)
```

Check the children of reloaded_ws on the AnalysisDataService and make sure that they have the same name
### Platforms affected

All

AntonPiccardoSelg:
Fixes #17903

Loading a saved group workspace via a the algorihtm `LoadNexusProcssed` did not respect the workspace name which is saved in the file. If a workspace name is found it is used now. If there are several workspaces which would load with the same workspace name then the underlying names are not used. 

The `LoadNexusProcessed` code makes use of `openLocal` on the Nexus file which is an optimization. During this bugfix this has caused a lot of troubles since it relies on a particular NXEntry to be open when a function is called. This means that code changes far away from where `openLocal` is called can influence the outcome of `openLocal`. We should maybe revisit this and check if the performance gain is really worth this risk.

### Test that similarly ending workspace names are reloaded just as is

Run this script

```python
ws_1 = CreateSampleWorkspace()
ws_new_1 = CreateSampleWorkspace()
group_ws = GroupWorkspaces(InputWorkspaces=[ws_1, ws_new_1])
YOUR_FILENAME = "C:/Sandbox/test_file.nxs"
SaveNexus(Filename=YOUR_FILENAME, Inputworkspace=group_ws)
DeleteWorkspace(group_ws)
reloaded_ws = LoadNexusProcessed(Filename=YOUR_FILENAME)
```

Confirm that the loaded child workspaces are ws_1 and ws_new_1. You can contrast this with the result you obtain from Mantid3.8 

### Ensure that entries in the file with the same name are not reloaded with the same name.

Run this script

```python
file_path = "C:/Sandbox/test_multi.nxs"

for element in range(2):
    ws = CreateSampleWorkspace()
    if element == 0:
        SaveNexusProcessed(InputWorkspace=ws, Filename=file_path,Append=False)
    else:
        SaveNexusProcessed(InputWorkspace=ws, Filename=file_path,Append=True)
DeleteWorkspace(ws)
reloaded = Load(Filename=file_path)
```
This should load ws_1 and ws_2 as child workspaces. Again you can contrast this with the results that one obtains from Mantid3.8

## Release Notes

See [here](https://github.com/mantidproject/mantid/pull/17910/commits/c38a6fa9a9f7f2a797b2a5785761466fef9a326d) for release notes.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


KarlPalmen:
It functions as specified and better than before. I'll leave it to @martyngigg to do the code review.

Framework/DataHandling/src/LoadNexusProcessed.cpp
```diff
@@ -421,9 +424,10 @@ void LoadNexusProcessed::exec() {
     std::string base_name = getPropertyValue("OutputWorkspace");
     // First member of group should be the group itself, for some reason!
 
-    // load names of each of the workspaces and check for a common stem
+    // load names of each of the workspaces. Note that if we have duplicate
+    // names then we don't select them
     std::vector<std::string> names(nWorkspaceEntries + 1);
-    bool commonStem = bIsMultiPeriod || checkForCommonNameStem(root, names);
+    extractWorkspaceNames(root, names);
```
martyngigg: I know this is picky but it would be good to gradually cleanup functions/methods that use this in/out style and prefer something that just returns the new vector.

With move semantics and return value optimization having `names = extractWorkspaceNames(root);` is much cleaner.

Framework/DataHandling/src/LoadNexusProcessed.cpp
```diff
@@ -577,33 +572,29 @@ void LoadNexusProcessed::correctForWorkspaceNameClash(std::string &wsName) {
...
 */
-bool LoadNexusProcessed::checkForCommonNameStem(
+void LoadNexusProcessed::extractWorkspaceNames(
     NXRoot &root, std::vector<std::string> &names) {
-  bool success(true);
   size_t nWorkspaceEntries = root.groups().size();
+  std::ostringstream os;
   for (size_t p = 1; p <= nWorkspaceEntries; ++p) {
-    std::ostringstream os;
+    os.str(std::string());
```
martyngigg: I think `boost::lexical_cast` might make more sense here and might be a bit cleaner than using a stream. I also imagine there are optimizations for single integer values like this.

Framework/DataHandling/src/LoadNexusProcessed.cpp
```diff
@@ -577,33 +572,29 @@ void LoadNexusProcessed::correctForWorkspaceNameClash(std::string &wsName) {
...
-    // check if the workspace name has an index on the end
-    if (!boost::regex_match(names[p], results, exp)) {
-      success = false;
-    }
   }
 
-  return success;
+  // Check that there are no duplicates in the workspace name
+  // This can cause severe problems
+  auto it = std::unique(names.begin(), names.end());
```
martyngigg: I don't quite understand what is happening here? Doesn't the vector just get left full of empty strings?

martyngigg (changes requested):

martyngigg:
@KarlPalmen Thank you for testing out the functionality.

Framework/DataHandling/src/LoadNexusProcessed.cpp
AntonPiccardoSelg: The overall intention of the changes are to respect workspace name settings that are stored in a file. However, if you store two `mantid_workspace_XXX` entries with the same name (which you can easily do -- see the second test scenario), then selecting those names will cause issues with the ADS. Hence if we encounter a workspace name twice, then we ignore the stored workspace names and use the default means of workspace name selection. The default is triggered by not finding any stored names, ie a collection of empty strings -- see [here](https://github.com/mantidproject/mantid/blob/8f4873bfe2a60536c4ad993ada63f3347d7481f3/Framework/DataHandling/src/LoadNexusProcessed.cpp#L531:L541)

I will go ahead with the other changes.

martyngigg: I see - that makes sense. Thanks for providing the context.

martyngigg (approved):

martyngigg:
Assuming the builds pass then :shipit:

AntonPiccardoSelg:
@martyngigg + @KarlPalmen  Thanks for testing this PR!

AndreiSavici:
please merge master into your branch

PR #234, opened 2015-02-18
Added ISIS-specific properties to StartLiveData.

mantid-roman:
Trac ticket [#11059](http://trac.mantidproject.org/mantid/ticket/11059)

To test run Example 2 from the documentation page.

PR #17926, opened 2016-11-01
Make listener properties in StartLiveData dynamic

### LINKED ISSUE #17912: StartLiveData should dynamically update Listener Properties based on selected LiveListener
### Expected behavior

When an instrument is selected in the StartLiveData, the Listener Properties  group should be updated to correctly reflect the properties provided by the specific LiveListener being used.

### Actual behavior

The Listener Properties group always contains the SpectraList and PeriodList properties of HistaDataListener, regardless of what specific LiveListener is being used, or what its properties are.

1. Run Mantid with ISIS selected as the default facility (or adjust setting in About > First Time Setup)
1. Execute the StartLiveData Algorithm in the sidebar
1. Select Instrument `ENGIN-X_EVENT`
1. There should be no Listener Properties (But `SpectraList` and `PeriodList` will be displayed)
1. Select Instrument `ENGIN-X`
1. Listener Properties should show and allow setting `SpectraList` and `PeriodList` properties

### Platforms affected
Should affect all. Test only on Ubuntu 16.04.

MikeHart85:
As required by [More Flexible LiveListener](https://github.com/mantidproject/documents/blob/master/Design/MoreFlexibleILiveListener.md) design document.

This effectively reverts #234, plus modifications and additions to still achieve its purpose.

SpectraList and PeriodList properties are no longer hard-coded as part of StartLiveData.

Instead, StartLiveData will dynamically add the properties of the specific LiveListener it uses once the Instrument is set.

The Python simpleapi has been updated with a specialized version of StartLiveData to allow evaluating other arguments based on the value of the Instrument argument.

**To test:**

See associated issue for testing GUI behaviour.

For Python interface:

1. Run a `FakeISISHistoDAE` algorithm with `NPeriods=5`, `NSpectra=10` and `NBins=100`.
1. Run the following code in the console:
```python
ConfigService.setFacility('TEST_LIVE')
StartLiveData(Instrument='ISIS_Histogram', OutputWorkspace='OutWS', AccumulationWorkspace='AccWS', UpdateEvery=1, AccumulationMethod='Replace', PeriodList=[1,3], SpectraList=[2,4,6])
```
This should work both before and after this change.

Fixes #17912.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/LiveData/src/StartLiveData.cpp
+++ b/Framework/LiveData/src/StartLiveData.cpp
@@ -91,7 +91,8 @@ void StartLiveData::afterPropertySet(const std::string &propName) {
         boost::dynamic_pointer_cast<IPropertyManager>(listener);
 
     if (propertyManagerListener) {
-      const std::vector<Property *> listenerProps = propertyManagerListener->getProperties();
+      const std::vector<Property *> listenerProps =
+          propertyManagerListener->getProperties();
       for (auto listenerProp : listenerProps) {
         auto prop = std::unique_ptr<Property>(listenerProp->clone());
         prop->setGroup(listenerPropertyGroup);
--- a/Framework/LiveData/src/StartLiveData.cpp
+++ b/Framework/LiveData/src/StartLiveData.cpp
@@ -76,28 +76,43 @@ void StartLiveData::init() {
  */
 void StartLiveData::afterPropertySet(const std::string &propName) {
   if (propName == "Instrument") {
-    // Remove old listener's properties
-    const std::vector<Property *> existingProps = getProperties();
-    for (auto existingProp : existingProps) {
-      if (existingProp->getGroup() == listenerPropertyGroup) {
-        removeProperty(existingProp->name());
-      }
-    }
+    // Properties of old listener, if any, need to be removed
+    removeListenerProperties();
 
-    // Add new listener's properties
+    // Get of instance of listener for this instrument
     auto listener = LiveListenerFactory::Instance().create(
         getPropertyValue(propName), false);
-    auto propertyManagerListener =
-        boost::dynamic_pointer_cast<IPropertyManager>(listener);
 
-    if (propertyManagerListener) {
-      const std::vector<Property *> listenerProps =
-          propertyManagerListener->getProperties();
-      for (auto listenerProp : listenerProps) {
-        auto prop = std::unique_ptr<Property>(listenerProp->clone());
-        prop->setGroup(listenerPropertyGroup);
-        declareProperty(std::move(prop));
-      }
+    // Copy over properties of new listener to this algorithm
+    copyListenerProperties(listener);
+  }
+}
+
+/**
+ * Copies properties from an ILiveListener to this algorithm. This makes them
+ * appear in the "listener properties" group on the StartLiveData custom dialog.
+ *
+ * @param listener ILiveListener from which to copy properties
+ */
+void StartLiveData::copyListenerProperties(const boost::shared_ptr<ILiveListener> &listener)
+{
+  // Add clones of listener's properties to this algorithm
+  for (auto listenerProp : listener->getProperties()) {
+    auto prop = std::unique_ptr<Property>(listenerProp->clone());
+    prop->setGroup(listenerPropertyGroup);
+    declareProperty(std::move(prop));
+  }
+}
+
+/**
+ * Removes previously copied ILiveListener properties.
+ */
+void StartLiveData::removeListenerProperties()
+{
+  // Remove all properties tagged with the listener property group
+  for (auto prop : getProperties()) {
+    if (prop->getGroup() == listenerPropertyGroup) {
+      removeProperty(prop->name());
     }
   }
 }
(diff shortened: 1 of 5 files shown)


Framework/PythonInterface/mantid/simpleapi.py
```diff
@@ -253,8 +253,14 @@ def StartLiveData(*args, **kwargs):
     except KeyError:
         pass
 
-    # Handle unpacking / workspace names from lhs
+    # LHS Handling currently unsupported for StartLiveData
     lhs = _kernel.funcinspect.lhs_info()
+    if lhs[0] > 0:  # Number of terms on the lhs
+        raise RuntimeError("Assigning the output of StartLiveData is currently "
```
OwenArnold: OK is this the fix until we figure out how to deal with optional output workspaces?

Framework/LiveData/src/StartLiveData.cpp
```diff
@@ -85,6 +70,38 @@ void StartLiveData::init() {
                   "completes.");
 }
 
+/**
+ * After Instrument property is set copy any properties that the instrument's
```
OwenArnold: If i'm being picky: I'd say that this comment will not necessarily apply to future modifications since the listener will not be tied to the instrument.

Framework/LiveData/src/StartLiveData.cpp
```diff
@@ -85,6 +70,38 @@ void StartLiveData::init() {
...
+    // Remove old listener's properties
+    const std::vector<Property *> existingProps = getProperties();
+    for (auto existingProp : existingProps) {
+      if (existingProp->getGroup() == listenerPropertyGroup) {
+        removeProperty(existingProp->name());
+      }
+    }
+
+    // Add new listener's properties
+    auto listener = LiveListenerFactory::Instance().create(
```
OwenArnold: Might be useful to have a comment about this (following) code relating to the  transfer of ownership. Or to put this in a separate named method.

docs/source/release/v3.9.0/framework.rst
```diff
@@ -40,6 +41,7 @@ Python Algorithms
 
 - :ref:`MatchPeaks <algm-MatchPeaks>` performs circular shift operation (numpy roll) along the x-axis to align the peaks in the spectra.
 - :ref:`FindEPP <algm-FindEPP>` is improved to better determine the initial parameters and range for the fitting.
+- :ref:`StartLiveData <algm-StartLiveData>` can now accept LiveListener properties as parameters, based on the Instrument parameter.
```
OwenArnold: I would be careful about "Instrument parameter". I initially misread this as "Instrument parameters" which is a completely different thing.

OwenArnold: I see that there is no concepts page for the LiveListener. We should definitely add that as part of a separate PR. However, I note that the algorithm rst for StartLiveData has not been updated as part of this PR. I think that it would be important to have some information in there about the fact that the algorithm is dynamically going to be picking up properties from it's listener such as `SpectrumList` and `PeriodList` otherwise that information will just have vanished, and I believe that these changes were originally made (the ones you just reverted), because users complained about not seeing documentation for them.

Framework/PythonInterface/mantid/simpleapi.py
MikeHart85: After talking to @martyngigg, we decided that for now this will have to be considered a limitation of the simpleapi. There are several issues with capturing the output:

1. Unwrapping is not working properly because of the optional "AccumulationWorkspace" output, and there's no easy fix for this.
1. Even capturing in a single variable (so capturing output as a tuple without unwrapping) is setting the AccumulationWorkspace argument to the name of the variable on the left.
1. There if the name of the first variable on the left matches the value of the "OutputWorkspace" argument, there is very inconsistent / buggy behaviour because now both OutputWorkspace and AccumulationWorkspace have the same name.
1. Since OutputWorkspace is being periodically replaced in the ADS, any reference to it you get in Python is invalidated shortly after you get it.

Framework/LiveData/src/StartLiveData.cpp
MikeHart85: Addressed by refactoring with two helper functions and simplifying code a bit.

docs/source/release/v3.9.0/framework.rst
MikeHart85: I've tried clarifying the wording a bit, let me know if it works better this way.

MikeHart85: Addressed by adding a section on listener properties to StartLiveData rst.

docs/source/algorithms/StartLiveData-v1.rst
```diff
@@ -23,6 +23,29 @@ simply calls :ref:`algm-LoadLiveData` at a fixed interval.
...
+
+.. code-block:: python
+
+    StartLiveData(Instrument='ISIS_Histogram', OutputWorkspace='wsOut', UpdateEvery=1,
+                  AccumulationMethod='Replace', PeriodList=[1,3], SpectraList=[2,4,6])
+
+PeriodList and SpectraList are properties of the ISISHistoDataListener. They
+are available as arguments in this call because Instrument is set to
+'ISIS_Histogram', which uses that listener.
+
```
OwenArnold: Good documentation!

MikeHart85: Thank you! :)

OwenArnold:
:shipit:

PR #17928, opened 2016-11-02
Add nx can sas to isis sans gui

### LINKED ISSUE #17583: Add NxCanSAS save options to ISIS SANS Gui
The new NXcanSAS format should be made available as an export format in the ISIS SANS Gui.

It should be available for the two save functions and the batch mode.

AntonPiccardoSelg:
Fixes #17583

The new file format NXcanSAS will be the new standard for ISIS SANS reduction results. We want to make available with this PR.

**To test:**

The relevant data can be found here: \\olympic\babylon5\Scratch\Anton\Testing\17583_add_nx_can_sas_option

Note that the folder contains a Windows installer

1. Open the ISIS SANS GUI
2. Set the instrument to SANS2D
3. Load the user file: USER_SANS2D_154E_2p4_4m_M3_Xpress_8mm_SampleChanger.txt
4. Set the workspaces to:
sample_scatter: 34484
sample_trans: 34502
sample_direct: 34461
can_scatter: 34481
can_trans: 34505
can_direct: 34461

5. Press load
6. Press 1DReduce
7. Once the reduction has completed, in the save section, only select `NxCanSAS` and press `SaveResult`
  * Confirm that the workspace is saved out
8. Load the file into mantid using `LoadNXcanSAS`
  * Confirm that it can be loaded
9. Delete the saved file
10. Press save other
11. Select only `NXcanSAS` and set the workspace to `34484rear_1D_1.75_16.5` which is the reduced workspace
12. Press save
  * Confirm that the workspace was saved out
13. Load the file into mantid using `LoadNXcanSAS`
  * Confirm that it can be loaded
  * Load the 

# Release notes

For release notes, see [here](https://github.com/mantidproject/mantid/pull/17928/commits/bd11db67a49e01360ea969036d87a666f4bcd387)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


lottiegreenwood:
works as described :shipit:

AntonPiccardoSelg:
@martyngigg Your suggested changes were applied

Framework/DataHandling/src/LoadNXcanSAS.cpp
```diff
@@ -568,5 +549,53 @@ void LoadNXcanSAS::exec() {
...
+  // Set the workspace on the output
+
+  std::string propertyName;
+  if (name == "sample")
+    propertyName = "TransmissionWorkspace";
+  else
+    propertyName = "TransmissionCanWorkspace";
+  const std::string doc = "The transmission workspace";
+
+  declareProperty(
```
martyngigg: This is definitely better. Would it be clearer to have these marked as optional outputs in `init` rather than declared at run time?

AntonPiccardoSelg: Does that mean they would appear in the Algorithm Dialog?

martyngigg: You're right, they would.

AntonPiccardoSelg: In which case there is a discrepancy between [`LoadCanSAS1D`](http://docs.mantidproject.org/nightly/algorithms/LoadCanSAS1D-v2.html) and [`LoadNXcanSAS`](http://docs.mantidproject.org/v3.8.0/algorithms/LoadNXcanSAS-v1.html). If we expose the properties in the algorthim dialog, then they would only provide a result if they are queried, right? Unfortunately the users expect them to appear without having to provide a name for them (the name is constructed from some file information). Is there a way around this?

martyngigg: I see - this is the best way to deal with that then. My main concern was the use of the ADS in the algorithm. I'm trying to limit this from occurring when I see it.

I will just give it a quick functional test with these changes and merge it if it's good then.

martyngigg:
All of the steps above still work as expected.

I did notice a warning from HDF5 when loading the saved file:
```
HDF5-DIAG: Error detected in HDF5 (1.8.11) thread 140072137504512:
  #000: ../../../src/H5D.c line 334 in H5Dopen2(): not found
    major: Dataset
    minor: Object not found
  #001: ../../../src/H5Gloc.c line 430 in H5G_loc_find(): can't find object
    major: Symbol table
    minor: Object not found
  #002: ../../../src/H5Gtraverse.c line 861 in H5G_traverse(): internal path traversal failed
    major: Symbol table
    minor: Object not found
  #003: ../../../src/H5Gtraverse.c line 641 in H5G_traverse_real(): traversal operator failed
    major: Symbol table
    minor: Callback failed
  #004: ../../../src/H5Gloc.c line 385 in H5G_loc_find_cb(): object 'Qdev' doesn't exist
    major: Symbol table
    minor: Object not found
```

Is this something we should try to suppress (probably in another issue)?

AntonPiccardoSelg:
@martyngigg In principle yes, but I am a bit surprised that there is any output. I think that we would have to set H5CPP to be less verbose. The execption is thrown [here](https://github.com/mantidproject/mantid/blob/master/Framework/DataHandling/src/LoadNXcanSAS.cpp#L201:L211) and all potential execption types are being handled (see [here](http://hdf4.org/HDF5/doc1.8/cpplus_RM/class_h5_1_1_common_f_g.html#a6a2557f173c32577db9feeb1b77d4e40)). 

I can check if there is some setting to make H5CPP quiet, but as you say that should probably be in another issue #17982.

martyngigg:
@AntonPiccardoSelg You are right. The exceptions are handled correctly it's just the default HDF5 behaviour when encountering an API error - https://support.hdfgroup.org/HDF5/doc/H5.user/Debugging.html.

Happy for this to be addressed in another issue.

PR #17929, opened 2016-11-02
Prompt peak correction for LARMOR

### LINKED ISSUE #17599: Provide Prompt Peak correction for Larmor
Larmor needs a prompt peak correction in the ISIS reduction workflow. This is very similar to the way LOQ handles the prompt peak.
Steps:
- Check what LOQ is doing
- Design a solution which handles Larmor and LOQ

AntonPiccardoSelg:
Fixes #17599

Adds the capability of the prompt peak correction and monitor unwrapping for LARMOR. We want to make sure that the feature does not break the reduction chain. Once this PR is merged Rob will test it from the Nightly Builds. Once he is fully ok with it we will add a system test and release notes. The features are off by default and need to be explicitly enabled via a SANS commands, hence these changes will not affect other users.

# To test

Please find the required data here:  \\olympic\Babylon5\Scratch\Anton\LARMOR_PROMPT_PEAK_WITH_MONITOR_UNWRAP

### Run standard reduction

1. Open the ISIS SANS GUI
2. Select LARMOR as your instrument
3. Load the user file "use_file_standard.txt
4. Add LARMOR00009848_corrected for the sample scatter, sample transmission and the sample direct
5. Press "Load data"
6. Press "1D Reduce"
  * Confirm that there were no errors

### Run reduction with prompt peak correction for LARMOR enabled

1. Open the ISIS SANS GUI
2. Select LARMOR as your instrument
3. Load the user file "user_file_with_prompt_peak_removealtxt
4. Add LARMOR00009848_corrected for the sample scatter, sample transmission and the sample direct
5. Press "Load data"
6. Press "1D Reduce"
  * Confirm that there were no errors

### Run reduction with prompt peak correction and monitor unwrapping for LARMOR enabled

1. Open the ISIS SANS GUI
2. Select LARMOR as your instrument
3. Load the user file "user_file_with_prompt_peak_removeal_and_monitor_unwrapping.txt
4. Add LARMOR00009848_corrected for the sample scatter, sample transmission and the sample direct
5. Press "Load data"
6. Press "1D Reduce"
  * Confirm that there were no errors

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


AntonPiccardoSelg:
@lottiegreenwood  The PR is ready for testing. I have added the instructions.

lottiegreenwood:
tested, works as described. Code review looks good :shipit:

Framework/Algorithms/src/UnwrapMonitorsInTOF.cpp
```diff
@@ -331,15 +331,27 @@ void UnwrapMonitorsInTOF::exec() {
   const auto workspaceIndices =
       getWorkspaceIndicesForMonitors(outputWorkspace.get());
 
+  bool spectrumIsHistogramData(true);
```
martyngigg: Is there a reason this is defined outside of the loop? It would seem best to define it locally in the loop.

scripts/SANS/isis_reduction_steps.py
```diff
@@ -4217,6 +4282,7 @@ def execute(self, reducer, workspace):
             Reads the geometry information stored in the workspace
             but doesn't replace values that have been previously set
         """
+        _reducer = reducer  # noqa
```
martyngigg: Is this used anywhere?

scripts/SANS/isis_reduction_steps.py
```diff
@@ -4282,7 +4282,7 @@ def execute(self, reducer, workspace):
             Reads the geometry information stored in the workspace
             but doesn't replace values that have been previously set
         """
-        _reducer = reducer  # noqa
+        _ = reducer  # noqa
```
martyngigg: I think I'm missing something. What's the reason that this line needs to exist at all?

AntonPiccardoSelg: I cannot remember why I added it, but I suspect that it was because of either a pylint or flake8 warning (or both), because we have an unused argument.

Since the class is a reduction step it has an `execute(self, reducer, workspace)` method. where the second argument is just not used, but it is part of the contract.

martyngigg: Ah I see. Sorry, I had missed the function arguments. Could you just have the signature as `def execute(self, _, workspace)` to indicate it is not used?

AntonPiccardoSelg: The class is derived from `ReductionStep` and overrides the `execute` method. See [here](https://github.com/mantidproject/mantid/blob/master/scripts/reducer_singleton.py#L31:L39). Changing the signature would of course be possible, but if the method is used via a kwarg, then it would break if we changed the argument signature here. Essentially we would be breaking the contract here.

On the other hand, I had a quick look and I can only find cases where the argument is used via its position, hence I would expect that the suggested change should not break anything in the current reduction setup. 

Do you think we should go ahead with this change?

martyngigg: Good point about the keywords. Let's keep it as it is for safety.

PR #17944, opened 2016-11-02
Fix run status log for aborted runs

### LINKED ISSUE #17924: Set run status correctly for aborted runs
*Reported by James Lord, Mantid 3.8*

#### Expected behaviour
If a run is aborted and then restarted, the `running` log should be set to false at all times during the aborted run.

#### Actual behaviour
`running` is false between the abort and the second start, but is left as true before the abort - despite the data from this aborted run not being in the file.

Look at `EMU00066122`: (section of event log below)

| Time (sec)       | Time (actual)           | Event |
| :-------------: |:-------------:|:-----:|
| -57      | 10:01:47 | BEGIN |
| -9      | 10:02:35 | ABORT      |
| 0 | 10:02:44 | BEGIN      |

The `running` sample log, which is generated by `LogParser`, is true from time -57 to time -9, false from time -9 to time 0 and then true from time 0 onwards. It should be false at all times before time 0, i.e. when the ABORT is seen at time -9 this makes anything before it count as "not running".

This has relevance when filtering log data by the running status, e.g. in the Sample Logs dialog. See #17877.

tom-perkins:
Correct the generation of the `running` log on loading a file: if the run was aborted and then restarted, the `running` log should only show that it was running after the restart. At all times before the restart, `running` should be false. (This is important for filtering other logs by `running`, and was requested by scientists).

Add tests for old and new style logs, and also did some maintenance on the other tests in `LogParserTest`.

**To test:**

Use any datafile where the run was aborted, then restarted. For example, try `EMU00066122.nxs`, which can be found at `\\olympic\babylon5\Scratch\TomPerkins\test_tsp_filtering`:

```python
tuple = Load('EMU00066122')
ws = tuple[0]
run = ws.run()
log = run.getProperty("running")
for i in range(log.size()):
    print log.nthTime(i), log.nthValue(i)
```
Before the fix, `running` would be `True` from the start time (10:01:47) until the run was aborted at 10:02:35.
After the fix, at all times before the abort, `running` should be `False`.

Fixes #17924.

[Release notes](https://github.com/mantidproject/mantid/blob/6de56f59dafe1c8ba1aea8bcc3b2e31af2d72924/docs/source/release/v3.9.0/framework.rst#bug-fixes) have been updated.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Kernel/inc/MantidKernel/LogParser.h
+++ b/Framework/Kernel/inc/MantidKernel/LogParser.h
@@ -100,7 +100,7 @@ private:
       Kernel::TimeSeriesProperty<bool> *status);
 
   /// Available commands.
-  enum commands { NONE = 0, BEGIN, END, CHANGE_PERIOD };
+  enum class commands { NONE = 0, BEGIN, END, CHANGE_PERIOD, ABORT };
 
   /// Typedef for a map of string commands to an enum of strongly typed
   /// commands.
--- a/Framework/Kernel/test/LogParserTest.h
+++ b/Framework/Kernel/test/LogParserTest.h
@@ -654,6 +654,38 @@ public:
     TS_ASSERT_EQUALS(expectedRunning, runningMap);
   }
 
+  /// If a run is aborted and then restarted, the "running" log should be set to
+  /// false at all times during the aborted run.
+  void test_abort_runningLogAlwaysFalseBeforeRestart_oldStyleCommands() {
+    auto log = make_unique<TimeSeriesProperty<std::string>>("MyICPevent");
+
+    // (This is a cut-down version of EMU66122, changed to "old style" commands)
+    const DateAndTime timeZero{"2016-10-01T10:01:44"};
+    const std::vector<DateAndTime> times{
+        timeZero,        timeZero + 3.0,  timeZero + 3.0,  timeZero + 51.0,
+        timeZero + 57.0, timeZero + 60.0, timeZero + 60.0, timeZero + 111.0};
+    const std::vector<std::string> values{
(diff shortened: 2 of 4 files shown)


PranavBahuguna:
The code itself seems fine, no issues there.

On running the tests as described, I can confirm that all times before 10:02:35 (where abort took place) now have a value of false, as expected.

Everything seems fine to me :shipit:

Framework/Kernel/inc/MantidKernel/LogParser.h
```diff
@@ -100,7 +100,7 @@ class MANTID_KERNEL_DLL LogParser {
       Kernel::TimeSeriesProperty<bool> *status);
 
   /// Available commands.
-  enum commands { NONE = 0, BEGIN, END, CHANGE_PERIOD };
+  enum class commands { NONE = 0, BEGIN, END, CHANGE_PERIOD, ABORT };
```
martyngigg: Nice introduction of enum class.

Framework/Kernel/test/LogParserTest.h
```diff
@@ -45,41 +46,37 @@ class LogParserTest : public CxxTest::TestSuite {
   void testGood() {
     mkICP();
     mkGood();
-    Property *icp_log =
-        LogParser::createLogProperty(icp_file.path(), "icpevent");
-    LogParser lp(icp_log);
-    Property *p1 = lp.createLogProperty(log_num_good.path(), "good");
+    const auto &icp_log = std::unique_ptr<Property>(
```
martyngigg: :+1: for `unique_ptr`

PR #17948, opened 2016-11-02
EnggVanadiumCorrections fixed using the wrong bank index 

### LINKED ISSUE #17934: EnggVanadiumCorrections gets the wrong bank indices
On line [L360](https://github.com/mantidproject/mantid/blob/master/Framework/PythonInterface/plugins/algorithms/EnggVanadiumCorrections.py#L360) in function `_divideByCurves` the for loop goes through bank indices 0 and 1, and 0 is not a valid value for the `idxs = EnggUtils.getWsIndicesForBank(ws, b)` [utils method](https://github.com/mantidproject/mantid/blob/master/scripts/Engineering/EnggUtils.py#L194) as that checks which indices are in bank 0 => none, so it adds the ones from the 3rd bank (indices > 2400) which isn't used anyway, or error when run on the imaging ones.

dtasev:
Description of work.

The function `_divideByCurves` no longer has the problem with the wrong bank id, because the proper id is being set while the `curves` dictionary is being created.

**To test:**
Adding a helpful `print(idxs)` in `_divideByCurves` can help make sure the indices are correct.

- Run vanadium calibrations with both banks selected
- the `print(idxs)` in `_divideByCurves` will now show the proper indices being iterated
  - 0-1200 on first loop, 1200-2400 on second
- Compare ids to those without the fix the `print(idxs)` 
    - that should show 2413-2512 on first loop and 0-1200 on second

Fixes #17934 .

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/PythonInterface/plugins/algorithms/EnggVanadiumCorrections.py
+++ b/Framework/PythonInterface/plugins/algorithms/EnggVanadiumCorrections.py
@@ -155,7 +155,7 @@ class EnggVanadiumCorrections(PythonAlgorithm):
         divides by a curve fitted to the sum of the set of spectra of the corresponding bank.
 
         @param ws :: workspace to work on / correct
-        @param curvesWS :: a workspace with the per-bank curves for Vanadium data, 
+        @param curvesWS :: a workspace with the per-bank curves for Vanadium data,
                 this will contain 3 histograms per instrument bank
         """
         curvesDict = self._precalcWStoDict(curvesWS)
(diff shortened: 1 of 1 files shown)


DavidFair:
This correctly loads the banks - I'm not sure why we weren't previously seeing this it only became apparent when trying to port IMAT into the Engg scripts.

:shipit:

DavidFair (approved):

PR #17961, opened 2016-11-03
Move interpolation from VectorHelper to HistogramData

### LINKED ISSUE #17913: Add interpolation methods to HistogramData
Currently `VectorHelper` contains a single linear interpolation method but #17801 requires a spline and `SplineInterpolation` is not designed in a way that is easy to use for a single spectrum.

`HistogramData` needs an interpolation operation similar to how `rebin` has been implemented.

martyngigg:
Moves the interpolation function from `VectorHelper` to the `HistogramData` module and adds cubic spline interpolation. The routines rely on understanding the type of data they are dealing with and therefore belong with the HistogramData module.

The `VectorHelper` function has been removed.

**To test:**

Code review. Current tests check that the behaviour has not changed for the existing absorption
correction algorithms.

Fixes #17913 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/src/MonteCarloAbsorption.cpp
+++ b/Framework/Algorithms/src/MonteCarloAbsorption.cpp
@@ -163,7 +163,7 @@ MonteCarloAbsorption::doSimulation(const MatrixWorkspace &inputWS,
     PARALLEL_START_INTERUPT_REGION
 
     auto &outE = outputWS->mutableE(i);
-    //    // The input was cloned so clear the errors out
+    // The input was cloned so clear the errors out
     outE = 0.0;
     // Final detector position
     IDetector_const_sptr detector;
--- a/Framework/Algorithms/src/MonteCarloAbsorption.cpp
+++ b/Framework/Algorithms/src/MonteCarloAbsorption.cpp
@@ -22,11 +22,13 @@
 #include "MantidKernel/VectorHelper.h"
 
 #include "MantidHistogramData/HistogramX.h"
+#include "MantidHistogramData/Interpolate.h"
 
 using namespace Mantid::API;
 using namespace Mantid::Geometry;
 using namespace Mantid::Kernel;
 using Mantid::HistogramData::HistogramX;
+using Mantid::HistogramData::interpolateLinearInplace;
 namespace PhysicalConstants = Mantid::PhysicalConstants;
 
 /// @cond
@@ -160,13 +162,9 @@ MonteCarloAbsorption::doSimulation(const MatrixWorkspace &inputWS,
   for (int64_t i = 0; i < nhists; ++i) {
     PARALLEL_START_INTERUPT_REGION
 
-    auto xvalues = outputWS->points(i);
-    auto &signal = outputWS->mutableY(i);
-    auto &errors = outputWS->mutableE(i);
-    // The input was cloned so clear the errors out
-    // Y values are all overwritten later
-    errors = 0.0;
-
+    auto &outE = outputWS->mutableE(i);
+    //    // The input was cloned so clear the errors out
+    outE = 0.0;
     // Final detector position
     IDetector_const_sptr detector;
     try {
@@ -179,10 +177,12 @@ MonteCarloAbsorption::doSimulation(const MatrixWorkspace &inputWS,
     const double lambdaFixed = toWavelength(efixed.value(detector));
     MersenneTwister rng(seed);
 
+    auto &outY = outputWS->mutableY(i);
+    const auto lambdas = outputWS->points(i);
     // Simulation for each requested wavelength point
     for (int j = 0; j < nbins; j += lambdaStepSize) {
       prog.report(reportMsg);
-      const double lambdaStep = xvalues[j];
+      const double lambdaStep = lambdas[j];
       double lambdaIn(lambdaStep), lambdaOut(lambdaStep);
       if (efixed.emode() == DeltaEMode::Direct) {
         lambdaIn = lambdaFixed;
@@ -191,7 +191,7 @@ MonteCarloAbsorption::doSimulation(const MatrixWorkspace &inputWS,
       } else {
         // elastic case already initialized
       }
-      std::tie(signal[j], std::ignore) =
+      std::tie(outY[j], std::ignore) =
           strategy.calculate(rng, detPos, lambdaIn, lambdaOut);
 
       // Ensure we have the last point for the interpolation
@@ -202,8 +202,9 @@ MonteCarloAbsorption::doSimulation(const MatrixWorkspace &inputWS,
 
     // Interpolate through points not simulated
     if (lambdaStepSize > 1) {
-      Kernel::VectorHelper::linearlyInterpolateY(
-          xvalues.rawData(), outputWS->dataY(i), lambdaStepSize);
+      auto histnew = outputWS->histogram(i);
+      interpolateLinearInplace(histnew, lambdaStepSize);
+      outputWS->setHistogram(i, histnew);
     }
 
     PARALLEL_END_INTERUPT_REGION
--- a/Framework/HistogramData/src/Interpolate.cpp
+++ b/Framework/HistogramData/src/Interpolate.cpp
@@ -113,7 +113,7 @@ void interpolateYCSplineInplace(const Histogram &input, const size_t stepSize,
     yc[i] = yold[step];
   }
   // Ensure we have the last value
-  xc.back() = xold[nypts - 1];
+  xc.back() = xold.back();
   yc.back() = yold.back();
 
   gsl_interp_accel *acc = gsl_interp_accel_alloc();
--- a/Framework/HistogramData/src/Interpolate.cpp
+++ b/Framework/HistogramData/src/Interpolate.cpp
@@ -69,16 +69,16 @@ void sanityCheck(const Histogram &input, const size_t stepSize,
  */
 void interpolateYLinearInplace(const Histogram &input, const size_t stepSize,
                                HistogramY &ynew) {
-  auto xold = input.points();
-  auto &yold = input.y();
-  auto nypts = yold.size();
+  const auto xold = input.points();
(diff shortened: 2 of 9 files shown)


Framework/HistogramData/src/Interpolate.cpp
```diff
@@ -0,0 +1,209 @@
...
+      ynew[i] = yold[i];
+      continue;
+    }
+    // Linear interpolation
+    ynew[i] = (xp - x1) * y2 + (x2 - xp) * y1;
+    ynew[i] *= overgap;
+    step++;
+  }
+}
+/**
```
soininen: Missing newline.

Framework/HistogramData/src/Interpolate.cpp
```diff
@@ -0,0 +1,209 @@
...
+}
+
+/**
+ * Perform linear interpolation assuming the input data set to be a Histogram
+ * with XMode=Points. It is assumed all sanity cheks have been performed
+ * by the interpolateLinear entry point.
+ * @param input See interpolateLinear
+ * @param stepSize See interpolateLinear
+ * @param ynew A reference to the output Y values
+ * @return See interpolateLinear
```
soininen: This, and some other functions in this file don't actually return anything.

Framework/HistogramData/src/Interpolate.cpp
```diff
@@ -0,0 +1,209 @@
...
+ * @param input See interpolateCSpline
+ * @param stepSize See interpolateCSpline
+ * @param ynew A reference to the output Y values
+ * @return See interpolateCSpline
+ */
+void interpolateYCSplineInplace(const Histogram &input, const size_t stepSize,
+                                HistogramY &ynew) {
+  auto xold = input.points();
+  auto &yold = input.y();
+  auto nypts = yold.size();
```
soininen: These variables could all be `const`, no? Except `xold`, if the next comment is valid...

Framework/HistogramData/src/Interpolate.cpp
```diff
@@ -0,0 +1,209 @@
...
+  for (size_t step = 0, i = 0; step < nypts; step += stepSize, i += 1) {
+    xc[i] = xold[step];
+    yc[i] = yold[step];
+  }
+  // Ensure we have the last value
+  xc.back() = xold.back();
+  yc.back() = yold.back();
+
+  gsl_interp_accel *acc = gsl_interp_accel_alloc();
+  gsl_spline *spline = gsl_spline_alloc(gsl_interp_cspline, ncalc);
```
soininen: `std::unique_ptr` could be used here instead of manual resource handling, but I'm unsure if it is needed for such a short function.

Framework/HistogramData/src/Interpolate.cpp
```diff
@@ -0,0 +1,209 @@
...
+void interpolateYCSplineInplace(const Histogram &input, const size_t stepSize,
+                                HistogramY &ynew) {
+  auto xold = input.points();
+  auto &yold = input.y();
+  auto nypts = yold.size();
+
+  const auto ncalc = numberCalculated(nypts, stepSize);
+  std::vector<double> xc(ncalc), yc(ncalc);
+  for (size_t step = 0, i = 0; step < nypts; step += stepSize, i += 1) {
+    xc[i] = xold[step];
```
soininen: Is copying `xold` to `xc` necessary here? To my understanding, `xold` is a copy-on-write thingie, so you could write to it directly without altering the original data. I might be wrong here, thought.

Framework/HistogramData/src/Interpolate.cpp
```diff
@@ -0,0 +1,209 @@
...
+/**
+ * Linearly interpolate through the y values of a histogram assuming that the
+ * calculated "nodes" are stepSize apart. Currently errors are not treated.
+ * @param input Input histogram defining x values and containing calculated
+ * Y values at stepSize intervals. It is assumed that the first/last points
+ * are always calculated points.
+ * @param stepSize The space, in indices, between the calculated points
+ * @return A new Histogram with the y-values from the result of a linear
+ * interpolation. If the XMode of the output will match that of the input
+ */
```
soininen: The doxygen is kind of missing something cut short. Also, no `@params` etc.

Framework/HistogramData/src/Interpolate.cpp
```diff
@@ -0,0 +1,209 @@
...
+ * Interpolate through the y values of a histogram using a cubic spline,
+ * assuming that the calculated "nodes" are stepSize apart.
+ * Currently errors are ignored.
+ * @param input Input histogram defining x values and containing calculated
+ * Y values at stepSize intervals. It is assumed that the first/last points
+ * are always calculated points.
+ * @param stepSize The space, in indices, between the calculated points
+ * @return A new Histogram with the y-values from the result of a linear
+ * interpolation. If the XMode of the output will match that of the input
+ */
```
soininen: The `@return` doxygen seems to be missing some text.

soininen (changes requested):

Framework/Algorithms/src/MonteCarloAbsorption.cpp
```diff
@@ -160,13 +162,9 @@ MonteCarloAbsorption::doSimulation(const MatrixWorkspace &inputWS,
...
 
-    auto xvalues = outputWS->points(i);
-    auto &signal = outputWS->mutableY(i);
-    auto &errors = outputWS->mutableE(i);
-    // The input was cloned so clear the errors out
-    // Y values are all overwritten later
-    errors = 0.0;
-
+    auto &outE = outputWS->mutableE(i);
+    //    // The input was cloned so clear the errors out
```
martyngigg: Yes, nice catch.

Framework/HistogramData/src/Interpolate.cpp
```diff
@@ -0,0 +1,209 @@
...
+constexpr const char *LINEAR_NAME = "Linear";
+constexpr const char *CSPLINE_NAME = "CSpline";
+
+/**
+ * Compute the number of pre-calculated points given the ysize and step size
+ * @param ysize The total number of points in the data set
+ * @param stepSize The step size between each calculated point
+ * @return The number of calculated nodes
+ */
+inline size_t numberCalculated(const size_t ysize, const size_t stepSize) {
```
martyngigg: It probably could. The current situations don't know enough at compile time but there is no harm in it being there.

Framework/HistogramData/src/Interpolate.cpp
```diff
@@ -0,0 +1,209 @@
...
+ * by the interpolateLinear entry point.
+ * @param input See interpolateLinear
+ * @param stepSize See interpolateLinear
+ * @param ynew A reference to the output Y values
+ * @return See interpolateLinear
+ */
+void interpolateYLinearInplace(const Histogram &input, const size_t stepSize,
+                               HistogramY &ynew) {
+  auto xold = input.points();
+  auto &yold = input.y();
```
martyngigg: `input` itself is const so `auto &yold` will remain `const` but `xold` should be `const`

Framework/HistogramData/src/Interpolate.cpp
```diff
@@ -0,0 +1,209 @@
...
+  auto nypts = yold.size();
+  size_t step(stepSize), index2(0);
+  double x1(0.), x2(0.), y1(0.), y2(0.), overgap(0.);
+  // Copy over end value skipped by loop
+  ynew.back() = yold.back();
+  for (size_t i = 0; i < nypts - 1; ++i) // Last point has been calculated
+  {
+    double xp = xold[i];
```
martyngigg: Yes I think it can.

Framework/HistogramData/src/Interpolate.cpp
martyngigg: :+1:

Framework/HistogramData/src/Interpolate.cpp
martyngigg: True, I could use `unique_ptr` with a custom deleter rather than using `delete`. I like it, I will try that.

Framework/HistogramData/src/Interpolate.cpp
martyngigg: In this case I want `xc` to just contain the pre-calculated points and so it's smaller than `xold`.I don't know if it might be confusing to try and orchestrate that with the cow pointers. I'll have a look.

Framework/HistogramData/src/Interpolate.cpp
```diff
@@ -0,0 +1,220 @@
...
+    yc[i] = yold[step];
+  }
+  // Ensure we have the last value
+  xc.back() = xold.back();
+  yc.back() = yold.back();
+
+  // Manage gsl memory with unique_ptrs
+  using gsl_spline_uptr = std::unique_ptr<gsl_spline, void (*)(gsl_spline *)>;
+  auto spline = gsl_spline_uptr(gsl_spline_alloc(gsl_interp_cspline, ncalc),
+                                [](gsl_spline *sp) { gsl_spline_free(sp); });
```
soininen: I don't think you need the lambda function here. It should be possible to pass `gsl_spline_free` directly to the pointer's constructor as the deleter.

On another note, would it be possible to use `auto` and `make_unique` here? Might improve readability.

soininen (approved):

Framework/HistogramData/src/Interpolate.cpp
martyngigg: Good point, I'll remove the lambda.

I don't think it's possible to use `make_unique` here as there's no way for it to know what are the arguments to forward to the type and what could be a custom deleter?

soininen: Yes, you are right about `make_unique`. If I read my references right, it cannot be used with custom deleters.

soininen:
The code looks nice and clean. The tests pass, so :shipit:

Issue #18014, opened 2016-11-08
HFIR SANS New file format


ricleal:
New XML version: `sans_spice_xml_format_version = 1.03`

There have been several changers to the experiment metadata (i.e. data stored in the xml files) that are mostly for supporting table scans and auto-reduction. The changes include:

- [x] 1) Added Scan_Type to header with values "Scattering", "Transmission", "Both", or "UNDEFINED".

- [x] 2) Removed Header Item Transmission_for_Scan.

- [x] 3) source_distance is now given in meters instead of mm.

- [x] 4) Added Start_Time and End_Time to Header.

- [x] 5) Changed sans_spice_xml_format_version in header from 1.02 to 1.03.

- [x] 6) Sample_Thickness now given in mm instead of cm.

- [x] 7) Sample_Type in header now set to “Open Beam”, “Empty Cell”, “Background”, “Sample”, or “UNDEFINED”.

- [x] 8) Beam Trap Diameter in header now correctly set.

- [x] 9) Sample Aperture in header now correctly set.

- [x] 10) New Sample_ID element in header.

- [x] 11) New Sample Changer Position element in header.

- [x] 12) Scan Title is automatically generated as a combination of sample name, instrument config and setup, and parameter values.

- [x] 13) Removed old reference scan elements from header (Beam_Center_For_Config_Scan, Beam_Center_For_Flood_Scan, Dark_Current_For_Config_Scan, Dark_Current_For_Flood_Scan, and Flood_Pattern_Scan).

- [x] 14) Added new reference scan elements to header (Ref_Scan_Main_Bar, Ref_Scan_Main_Beam_Center,  Ref_Scan_Main_Dark_Current, Ref_Scan_Main_Flood, Ref_Scan_West_Wing_Bar, Ref_Scan_West_Wing_Dark_Current, Ref_Scan_West_Wing_Flood).

- [x] 15) Added new sections for sample columns, background columns, and sample parameters. These are under the new <sample_info> section.

ricleal:
For now I just converted units in the loader to the old mantid units. I.e.:
- source_distance from m to mm
- Sample_Thickness from mm to cm
We may  need to change the default units (m and mm) for reducer and parse the units as they are.

PR #18024, opened 2016-11-09
Re #18014 Changed loader for XML version 1.03

ricleal:
See ticket #18014 

No release notes.

**To test:**

1) Tests should pass. They can also be launched manually:
`ctest -VV --output-on-failure -R DataHandlingTest_LoadSpice`

2) The file attached can also be loaded.

[BioSANS_exp337_scan0050_0001.xml.zip](https://github.com/mantidproject/mantid/files/581555/BioSANS_exp337_scan0050_0001.xml.zip)


mdoucet:
Small change. Looks good to me. :shipit:

Issue #17743, opened 2016-10-07
Rollout SpectrumInfo and DetectorInfo


SimonHeybrock:
## SpectrumInfo and DetectorInfo Rollout

This is an umbrella issue for the `SpectrumInfo` and `DetectorInfo` rollout, following a similar strategy to the `HistogramData` rollout.

**Please read [InstrumentAccessLayers](http://docs.mantidproject.org/nightly/concepts/InstrumentAccessLayers.html) first for information and examples for the rollout.**

Please pick a group at a time and place your name next to what you picked.

### Issue Template

```
SpectrumInfo and DetectorInfo rollout: algorithms <X to Y>

This issue is part of the SpectrumInfo and DetectorInfo rollout, see https://github.com/mantidproject/mantid/issues/17743.
Algorithms covered in this issue are:

```

### PR Template

```
SpectrumInfo rollout: algorithms <X to Y>

## Description of work.
This PR is part of refactoring effort to use the new interface of the SpectrumInfo and DetectorInfo modules.
See https://github.com/mantidproject/mantid/issues/17743 for details.

The following algorithms were refactored to use SpectrumInfo/DetectorInfo:

Performance tests were added for:

The followinNg algorithms have improved performance:

## To test.

Code review.

<Add any manual testing not covered by unit tests, especially GUIs>

Fixes <your issue>.
Internal change, no release notes (potential performance improvements will be added to release notes as part of the umbrella issue).
```

### File list

Below is a list of files using `MatrixWorkspace::getDetector()` or `Instrument::getDetector()` that still need updating. Header (`.h`), code (`.cpp`) and test (`Test.cpp`) files all might updating - please check them all.

Place your name next to the group you picked.

### Files that can not be updated now

```
Framework/MDAlgorithms/src/MDNormDirectSC.cpp (uses cached detector grouping)
Framework/MDAlgorithms/src/MDNormSCD.cpp (uses cached detector grouping)
Framework/DataHandling/src/MoveInstrumentComponent.cpp (uses PeaksWorkspace which doesn't supply SpectrumInfo)
Framework/DataHandling/src/RotateInstrumentComponent.cpp (uses PeaksWorkspace which doesn't supply SpectrumInfo)
Framework/Algorithms/src/SpecularReflectionAlgorithm.cpp (requires non-trivial refactoring to avoid cache problems)
```

ianbush:
The following algorithms are done in #18031:

* Framework/Algorithms/src/BinaryOperation.cpp
[...]

PR #18034, opened 2016-11-10
SpectrumInfo rollout 1 to 5

### LINKED ISSUE #18031: SpectrumInfo rollout: algorithms 1 to 5
This issue is part of the SpectrumInfo rollout, see https://github.com/mantidproject/mantid/issues/17743.

* Framework/Algorithms/src/BinaryOperation.cpp
* Framework/Algorithms/src/CalculateTransmission.cpp
* Framework/Algorithms/src/CalculateTransmissionBeamSpreader.cpp
* Framework/Algorithms/src/ConvertSpectrumAxis.cpp
* Framework/Algorithms/src/ConvertSpectrumAxis2.cpp

ianbush:
## Description of work.
This PR is part of refactoring effort to use the new interface of the SpectrumInfo module.

The following algorithms were refactored to use SpectrumInfo:

```
Framework/Algorithms/src/BinaryOperation.cpp
Framework/Algorithms/src/CalculateTransmission.cpp
Framework/Algorithms/src/CalculateTransmissionBeamSpreader.cpp
Framework/Algorithms/src/ConvertSpectrumAxis.cpp
Framework/Algorithms/src/ConvertSpectrumAxis2.cpp
```

Performance tests were added for: `ConvertSpectrumAxis2`

## To test.

Code review.

Fixes #18031 .

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


ianbush:
The following algorithms are done in #18031:

* Framework/Algorithms/src/BinaryOperation.cpp
* Framework/Algorithms/src/CalculateTransmission.cpp
* Framework/Algorithms/src/CalculateTransmissionBeamSpreader.cpp
* Framework/Algorithms/src/ConvertSpectrumAxis.cpp
* Framework/Algorithms/src/ConvertSpectrumAxis2.cpp

Framework/API/src/SpectrumInfo.cpp
```diff
@@ -176,8 +176,15 @@ const Geometry::IDetector &SpectrumInfo::getDetector(const size_t index) const {
...
-      det_ptrs.push_back(m_detectorInfo.getDetectorPtr(detIndex));
+      try {
+        const auto detIndex = m_detectorInfo.indexOf(id);
+        det_ptrs.push_back(m_detectorInfo.getDetectorPtr(detIndex));
+      } catch (std::out_of_range) {
+        // Workspaces can contain invalid detector IDs. Those IDs will be
+        // silently ignored here until this is fixed. Some valid IDs will exist
+        // if hasDetectors or hasUniqueDetectors has returned true, but there
+        // could still be invalid IDs.
+      }
```
ianbush: For the reviewer - is this a good way to deal with spectra that have a mix of valid and invalid IDs? This only seems to happen in rare cases.

OwenArnold: Interesting question. Since the method is called `getValidDetectorIDs` I would indeed only expect it to return valid detector ids. However the try-catch is going to adversely affect performance. Do you have any idea of how big the regression will be?

Framework/API/src/SpectrumInfo.cpp
```diff
@@ -176,8 +176,15 @@ const Geometry::IDetector &SpectrumInfo::getDetector(const size_t index) const {
     // Else need to construct a DetectorGroup and use that
     std::vector<boost::shared_ptr<const Geometry::IDetector>> det_ptrs;
     for (const auto &id : dets) {
-      const auto detIndex = m_detectorInfo.indexOf(id);
-      det_ptrs.push_back(m_detectorInfo.getDetectorPtr(detIndex));
+      try {
+        const auto detIndex = m_detectorInfo.indexOf(id);
+        det_ptrs.push_back(m_detectorInfo.getDetectorPtr(detIndex));
+      } catch (std::out_of_range) {
```
OwenArnold: Shouldn't you be catching via reference?

Framework/API/src/SpectrumInfo.cpp
OwenArnold: Could you also explain what the exact circumstances are that lead to this throwing at the moment?

Framework/Algorithms/inc/MantidAlgorithms/BinaryOperation.h
```diff
@@ -111,8 +111,8 @@ class DLLExport BinaryOperation : public API::Algorithm {
   /// masked. If so then the output spectra has zeroed data
   /// and is also masked. The function returns true if further processing is not
   /// required on the spectra.
-  virtual bool propagateSpectraMask(const API::MatrixWorkspace_const_sptr lhs,
-                                    const API::MatrixWorkspace_const_sptr rhs,
+  virtual bool propagateSpectraMask(const API::SpectrumInfo &lhsSpectrumInfo,
```
OwenArnold: Seems sensible. I expect that this function was doing nothing with the shared pointer, and the copy of the function arguments was clearly going to be expensive, as well as being ambiguous (regarding the shared_ptr aspect that is).

Framework/Algorithms/inc/MantidAlgorithms/BinaryOperation.h
```diff
@@ -111,8 +111,8 @@ class DLLExport BinaryOperation : public API::Algorithm {
   /// masked. If so then the output spectra has zeroed data
```
OwenArnold: I think you are relying on `SpectrumInfo` being forward declared or included in one of the included headers. Not sure that is wise.

OwenArnold: As a side note. I notice that this header is passing shared_pointers by value where it probably shouldn't be. Not sure it's worth you fixing that here though.

Framework/Algorithms/src/BinaryOperation.cpp
```diff
@@ -432,29 +432,23 @@ std::string BinaryOperation::checkSizeCompatibility(
...
  * the binary operation should be performed.
  */
-bool BinaryOperation::propagateSpectraMask(
-    const API::MatrixWorkspace_const_sptr lhs,
-    const API::MatrixWorkspace_const_sptr rhs, const int64_t index,
-    API::MatrixWorkspace_sptr out) {
+bool BinaryOperation::propagateSpectraMask(const SpectrumInfo &lhsSpectrumInfo,
+                                           const SpectrumInfo &rhsSpectrumInfo,
+                                           const int64_t index,
+                                           API::MatrixWorkspace_sptr out) {
```
OwenArnold: since you've changed this signature. How about updating the type of that out parameter? Probably don't need that shared_ptr on the out parameter.

Framework/Algorithms/src/BinaryOperation.cpp
```diff
@@ -432,29 +432,23 @@ std::string BinaryOperation::checkSizeCompatibility(
...
-    det_rhs = rhs->getDetector(index);
-  } catch (std::runtime_error &) {
-  } catch (std::domain_error &) {
-    // try statement will throw a domain_error when the axis is not a spectra
-    // axis.
-    return continueOp;
-  }
-  if ((det_lhs && det_lhs->isMasked()) || (det_rhs && det_rhs->isMasked())) {
+
+  if ((lhsSpectrumInfo.hasDetectors(index) &&
```
OwenArnold: Might be useful to add a method that combines the `hasDetectors(size_t)` and `isMasked(size_t)` on `spectrumInfo` as a convenience. Not a major issue though. Would improve readability and may have some minor performance benefits.

Framework/API/src/SpectrumInfo.cpp
ianbush: You cans see the system tests that failed here - http://builds.mantidproject.org/job/pull_requests-rhel7/10305/.

In the case of the `LoadTest` failures it is caused by running something like `Load("TSC15352+15353.raw)`. There is something going wrong in the loader, so one spectra has a spectrum number of `-1` and is associated with mostly invalid detector IDs, but some valid ones are still in there.

I did not look much into the SNSConvertToMD*, but something similar seems to happen, though the spectrum numbers are at least valid.

I do not know how expensive the try catch would be compared with other approaches. I did also try iterating over the detectors in `getDetector()` to check if they were valid, but this seemed messy and even more expensive.

The better option may be to get the bad loaders fixed now, but it may take some time to work through fixing them.

Framework/Algorithms/src/ConvertSpectrumAxis.cpp
```diff
@@ -172,14 +174,15 @@ void ConvertSpectrumAxis::exec() {
   setProperty("OutputWorkspace", outputWS);
 }
 
-double ConvertSpectrumAxis::getEfixed(IDetector_const_sptr detector,
-                                      MatrixWorkspace_const_sptr inputWS,
-                                      int emode) const {
+double
+ConvertSpectrumAxis::getEfixed(const Mantid::Geometry::IDetector &detector,
```
OwenArnold: I take it that it is necessary to get `EFixed` at the detector level?

Framework/Algorithms/src/ConvertSpectrumAxis.cpp
```diff
@@ -192,29 +195,29 @@ double ConvertSpectrumAxis::getEfixed(IDetector_const_sptr detector,
...
         }
       } else {
         efixed = 0.0;
         g_log.warning() << "Efixed could not be found for detector "
-                        << detector->getID() << ", set to 0.0\n";
+                        << detector.getID() << ", set to 0.0\n";
       }
     } else if (emode == 2) {
-      std::vector<double> efixedVec = detector->getNumberParameter("Efixed");
+      std::vector<double> efixedVec = detector.getNumberParameter("Efixed");
```
OwenArnold: Ok. I think i've answered my own question here.

Framework/Algorithms/src/ConvertSpectrumAxis.cpp
```diff
@@ -95,15 +96,16 @@ void ConvertSpectrumAxis::exec() {
...
-      if (!detector->isMonitor()) {
-        twoTheta = inputWS->detectorTwoTheta(*detector);
-        l2 = detector->getDistance(*sample);
+      if (!spectrumInfo.isMonitor(i)) {
+        twoTheta = spectrumInfo.twoTheta(i);
+        l2 = spectrumInfo.l2(i);
         l1val = l1;
-        efixed = getEfixed(detector, inputWS, emode); // get efixed
+        efixed =
+            getEfixed(spectrumInfo.detector(i), inputWS, emode); // get efixed
```
OwenArnold: I'm wondering if getting e-fixed should be a function of the `SpectrumInfo`. Seems like a general thing to want to get.

Framework/Algorithms/test/BinaryOperationTest.h
```diff
@@ -174,18 +175,14 @@ class BinaryOperationTest : public CxxTest::TestSuite {
...
-      IDetector_const_sptr det;
-      try {
-        det = output->getDetector(i);
-      } catch (Mantid::Kernel::Exception::NotFoundError &) {
-      }
+      auto spectrumInfo = output->spectrumInfo();
 
-      TSM_ASSERT("Detector was found", det);
-      if (det) {
+      TSM_ASSERT("Detector was found", spectrumInfo.hasDetectors(i));
```
OwenArnold: I appreciate that this was like this already, but when the assertion failes the user will see the message "Detector was found" implying that it should not have been. You need to introduce a negative into the test message.

Framework/Algorithms/test/ConvertSpectrumAxis2Test.h
```diff
@@ -391,4 +391,34 @@ class ConvertSpectrumAxis2Test : public CxxTest::TestSuite {
...
+
+  ConvertSpectrumAxis2TestPerformance() {}
+
+  void test_conversion_to_signed_theta_with_many_entries() {
+    std::string inputWSName("inputWS");
+    std::string outputWSName("outputWS");
+
+    auto testWS = WorkspaceCreationHelper::create2DWorkspaceWithFullInstrument(
+        20000, 20000);
+    AnalysisDataService::Instance().addOrReplace(inputWSName, testWS);
```
OwenArnold: Why are you adding things to the ADS in the test?

Framework/Algorithms/test/ConvertSpectrumAxis2Test.h
```diff
@@ -391,4 +391,34 @@ class ConvertSpectrumAxis2Test : public CxxTest::TestSuite {
...
+
+    Mantid::Algorithms::ConvertSpectrumAxis2 conv;
+
+    conv.initialize();
```
OwenArnold: if you had `conv.setChild(True)`. You probably wouldn't need to do that nonsense with the ADS that everyone seems to like.

Framework/Algorithms/test/ConvertSpectrumAxis2Test.h
```diff
@@ -391,4 +391,34 @@ class ConvertSpectrumAxis2Test : public CxxTest::TestSuite {
...
+    conv.setPropertyValue("InputWorkspace", inputWSName);
+    conv.setPropertyValue("OutputWorkspace", outputWSName);
+    conv.setPropertyValue("Target", "SignedTheta");
+    conv.setPropertyValue("EFixed", "10.0");
+    conv.execute();
+  }
```
OwenArnold: Am i missing something or is this failing to test anything? No expectations/asserts.

Framework/API/src/SpectrumInfo.cpp
OwenArnold: I would say that we have to fix the root cause of the problem rather than taking out an insurance policy (the try catch), where possible. I suspect that the nested try catch will have some large performance issues, but I've never measured it, so can't say for sure.

Framework/Algorithms/src/ConvertSpectrumAxis.cpp
ianbush: I would agree - it is pretty widely used but the code to get e-fixed can be messy. I think if we make this part of `SpectrumInfo` it will make the rollout easier.

Framework/API/src/SpectrumInfo.cpp
ianbush: Yes, the root cause needs fixing. From the comments in `SpectrumInfo::hasDetectors()` and `SpectrumInfo::hasUniqueDetectors()` I was expecting some of the Instrument 2.0 changes will prevent workspaces having invalid detector IDs, so the for loops in those methods will then be removed, as would the try catch.

So should we fix the loaders now, and omit the try/catch, or can we do something more fundamental to prevent the possibility of having invalid detector IDs in the workspace now?

OwenArnold: I think that returning a `Detector` is not something that may even be possible in `Instrument 2.0`. We would therefore not, for example, need the concept of a `DetectorGroup`. However since there can be a 1:n mapping between spectra and detectors, a loop would still be required in a lot of `SpectrumInfo` methods. Do you have any idea of how much effort would be involved in fixing the loaders?

Framework/Algorithms/src/ConvertSpectrumAxis.cpp
OwenArnold: One major lesson I think we can learn from things like `HistogramData` is that the benefit of simplifying and making concepts concise in a codebase of this size is reason enough to do it. So yes, sounds like this should be pushed down somewhere into `SpectrumInfo`, `DetectorInfo`

Framework/Algorithms/test/ConvertSpectrumAxis2Test.h
OwenArnold: My bad. I should have spotted that.

ianbush:
For the performance of the try-catch block I tested with the following script. No significant difference was seen with the addition of the try-catch in this case.

```python
import time

GROUPING_PATTERN = '0-9,10-19,20-29,30-39,40-49,50-59,60-69,70-79,80-89,90-99,100-109,110-119,120-129,130-139,140-149,150-159,160-169,170-179,180-189,190-199'

ws = CreateSampleWorkspace()
ws1g = GroupDetectors(InputWorkspace=ws, GroupingPattern=GROUPING_PATTERN)
ws2g = GroupDetectors(InputWorkspace=ws, GroupingPattern=GROUPING_PATTERN)

start_time = time.time()

for i in range(0,10000):
    ws = ws1g + ws2g
    
print(time.time() - start_time)
```

Framework/Algorithms/src/CalculateTransmissionBeamSpreader.cpp
```diff
@@ -244,7 +245,7 @@ API::MatrixWorkspace_sptr
 CalculateTransmissionBeamSpreader::extractSpectrum(API::MatrixWorkspace_sptr WS,
                                                    const size_t index) {
   // Check that given spectra are monitors
-  if (!WS->getDetector(index)->isMonitor()) {
+  if (!WS->spectrumInfo().isMonitor(index)) {
```
SimonHeybrock: How often is this method called? Currently `MatrixWorkspace::spectrumInfo()` is a bit expensive, so doing so for every `index` might need to be avoided.

Framework/Algorithms/src/ConvertSpectrumAxis.cpp
SimonHeybrock: I have looked several times into having a nicer way to get `EFixed` and `EMode`. The problem is that Mantid has two ways of storing/getting it (instrument parameter or in `Run`, I think). So this is not easy to refactor. Also, there is some weird behavior with the way parameters are obtain that might make trouble. For example, if you store a parameter on your detectors, and then group detectors into a spectrum, the parameters will be ignored when getting parameters for the spectrum.

I agree 100% that this should be done in a better way, but it seems to be a can of worms, so we have to think carefully when and how we can do it.

ianbush:
@SimonHeybrock I had a look at implementing something for the eFixed mode. It only really tries to address the case of getting the eFixed for a detector, should it exist. If you have a chance to look at it please let me know what you think.

https://github.com/mantidproject/mantid/pull/18099

Framework/Algorithms/src/CalculateTransmissionBeamSpreader.cpp
ianbush: Surprisingly given the name it is only actually called once for each workspace.

ianbush:
Due to a slightly messed up revert I have opened a new PR https://github.com/mantidproject/mantid/pull/18189. I will close this PR and any new comments will be easier to track on the new PR.

PR #18046, opened 2016-11-14
Fixed issues in the 'Return Background' mode in `CalculateFlatBackground`

### LINKED ISSUE #17717: `CalculateFlatBackground` returns bogus values if background is negative
`CalculateFlatBackground` promises to leave a spectrum unchanged, if a negative background has been calculated for it. However, when run with _OutputMode_ property set to 'Return Background', the values of the returned background for the spectrum in question are not set to zero, as one would expect. In other words,

``` python
# We need a workspace with negative "background"
ws = CreateWorkspace(DataX=[0.0, 0.5, 1.0], DataY=[-0.23, -0.23])
bkg = CalculateFlatBackground(InputWorkspace=ws, OutputMode='Return Background', StartX=0, EndX=1)
# Background for ws is negative -> bkg should be zero. However,
print('{0} != 0'.format(bkg.dataY(0)[0]))
```

soininen:
This PR fixes issues in `CalculateFlatBackground` in the case of returning the background instead of subtracting it. The issues were due to the output workspaces being initially cloned from the input workspace: If any histogram would be skipped by e.g. calculation failure or by not being included in the *WorkspaceIndexList* property, the output would contain the original histograms, not zeros as one would expect. Also, the errors in the returned background workspace were calculated from the input workspace (or directly cloned from there), not the actual errors in the background.

While at it, the 'Moving Average' mode was made more robust with regards to floating point rounding errors and variance calculation was added to it.

**To test:**

Except for some corner cases when the *NullifyNegativeValues* is set to true (these are mentioned in the updated documentation), it should not matter whether one uses the 'Subtract Background' option or manually runs the [`Minus`](http://docs.mantidproject.org/nightly/algorithms/Minus-v1.html) algorithm --- in both cases the end result should be the same. The unit tests already cover some simple cases, but these changes should be tested against more realistic data and usage scenarios.

Fixes #17717 #18053 #18054

**A line noting the fixes was added to the Release notes.**

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/inc/MantidAlgorithms/CalculateFlatBackground.h
+++ b/Framework/Algorithms/inc/MantidAlgorithms/CalculateFlatBackground.h
@@ -7,9 +7,9 @@
 #include "MantidAPI/Algorithm.h"
 
 namespace Mantid {
-  namespace HistogramData {
-    class Histogram;
-  }
+namespace HistogramData {
+class Histogram;
+}
 namespace Algorithms {
 /** Finds a constant background value of each desired spectrum
     and optionally subtracts that value from the entire spectrum.
@@ -56,8 +56,7 @@ namespace Algorithms {
 class DLLExport CalculateFlatBackground : public API::Algorithm {
 public:
   /// (Empty) Constructor
-  CalculateFlatBackground()
-      : API::Algorithm(), m_progress(nullptr) {}
+  CalculateFlatBackground() : API::Algorithm(), m_progress(nullptr) {}
   /// Virtual destructor
   ~CalculateFlatBackground() = default;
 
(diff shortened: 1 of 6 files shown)


Framework/Algorithms/src/CalculateFlatBackground.cpp
```diff
@@ -152,182 +153,127 @@ void CalculateFlatBackground::exec() {
...
-          E[j] = E[j] > background ? E[j] : background;
+          err = es[j] > background ? es[j] : background;
+        }
+        ys[j] = val;
+        es[j] = err;
+      }
+    } else {
+      for (size_t j = 0; j < ys.size(); ++j) {
+        const double originalVal = inputWS->y(i)[j];
+        if (nullifyNegative && background > originalVal) {
```
DavidFair: Could you add brackets to make the condition clearer: `(nullifyNegative && (background > originalVal))`

Framework/Algorithms/inc/MantidAlgorithms/CalculateFlatBackground.h
```diff
@@ -86,24 +83,17 @@ class DLLExport CalculateFlatBackground : public API::Algorithm {
...
-
-  /// variable bin width raw count data must be converted to distributions first
-  /// and then converted back, keep track of this
-  bool m_convertedFromRawCounts;
-  /// the variable which specifies if background should be removed from monitors
-  /// too.
-  bool m_skipMonitors;
-  // if true, negative signals appearing after background removals are set to 0
-  bool m_nullifyNegative;
+  void getWsInds(std::vector<int> &output, const size_t workspaceTotal);
```
DavidFair: Could this just return a `std::vector<int>` instead of passing in a reference

DavidFair (commented):
Just some minor changes suggested

DavidFair:
I'm just running the attached scripts in the issues linked. This resolves the examples in #18053 and #18054 but the script in #17717 still returns a non zero value?

Framework/Algorithms/inc/MantidAlgorithms/CalculateFlatBackground.h
soininen: No. But since this method has become so short with `std::iota`, it might make `exec()` a bit clearer if the code was moved there.

soininen:
Nice catch with #17717 ! Looks like another corner case I did not take into account. I wonder why I didn't make this a unit test in the first place...

soininen:
There was also another bug uncovered by the script in #17717 . This should also be fixed now.

DavidFair (approved):

DavidFair:
These changes fix #17717 as well as #18053 and #18054 . :shipit:

PR #18067, opened 2016-11-15
SetDetScale option for text file input

### LINKED ISSUE #17946: SetDetScale should have text input file
String input is difficult for instruments with many detectors

VickieLynch:
Description of work.
File input was requested since text string is too long for many detectors.
**To test:**
```python
Load(Filename='TOPAZ_3007.peaks', OutputWorkspace='TOPAZ_3007')
SetDetScale(Workspace='TOPAZ_3007', DetScaleFile='setDetScale.txt')

wksp = mtd['TOPAZ_3007']
print wksp.getInstrument().getNumberParameter("detScale49")
```

Fixes #17946 
[setDetScale.txt](https://github.com/mantidproject/mantid/files/592526/setDetScale.txt)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/PythonInterface/plugins/algorithms/SetDetScale.py
```diff
@@ -51,6 +56,14 @@ def PyExec(self):
             comp, value = component.split(":")
```
gvardany: I would expect that if both the list and the file are given, the list should overwrite the scales given in the file. 
So I would move this loop after the loop over the lines in the file.

gvardany (changes requested):
I had one comment (see inline). Also could you please add a release note and update the documentation of the algorithm?

gvardany:
It seems like overwriting the scale does not work, even after loops are swapped. 
Did you manage to test this case?
Maybe one needs to organise the list better, to make sure that the key (i.e. detector ID) is unique in the list?
Anyway, would be nice to have one more unit test for this corner case.

VickieLynch:
I am using a dictionary now and have added test.  See if overwriting works for you.

Vickie E Lynch
S&T Staff
Neutron Data Analysis & Visualization Division

Oak Ridge National Laboratory
PO Box 2008, Oak Ridge, TN 37831-6475
Office: 865-250-4795 | Mobile: 865-250-4795 | Fax: 865-574-6080
lynchve@ornl.gov | Bldg. 8600, Room C-448, MS 6475

gvardany (approved):

gvardany:
Sorry it was my bad, I did not pull your previous commit, but you still caught the email.
Anyway I agree that dictionary here is way more appropriate, and it works nice.
:shipit:

PR #18069, opened 2016-11-15
Fix "plot guess" leaving old guesses on plot

### LINKED ISSUE #18060: "Plot guess" leaves stale curves on graph
#### Expected behaviour
*Plot guess* should only have one guess curve, which corresponds to the function and parameters currently set, and updates itself automatically when parameters are changed.

#### Actual behaviour
Stale plot guesses can get left on the graph.

- Open *Interfaces/Muon/Muon Analysis*. Ensure the multi fitting mode is enabled, i.e. "compatibility mode" is *not* on.
- Load a dataset, go to Data Analysis tab and set up a function
- *Display/Plot guess*: a guess curve is plotted
guess1

- Change one of the fit parameters - nothing happens. 
- *Display/Remove guess*: instead of removing the first guess, a second guess is plotted on top
guess2

- *Display/Remove guess* (again) - the second guess is removed, but there is no way of removing the first one - it's stuck there now.

The same happens if you "plot guess", fit the data and then select "Remove guess" - it plots a second guess instead of removing the first, and it is not possible to remove the first, stale guess.

Note that the problem only appears with the new fitting UI. "Plot guess" works fine in compatibility mode (multi fitting off).

(See also #15872, which was a similar problem.)

#### Reported by
James Lord, Mantid 3.8, Windows

tom-perkins:
Fixed the *Plot guess* functionality in MuonAnalysis. This should work the same as it did with the old fitting UI: there should be only one guess curve, which auto-updates when the function or parameters are changed. "Remove guess" should do what it says and not plot another guess.

When multi fit mode (the new fitting UI) is on, need to remove any plotted guess before updating the function in the model, then plot it again if necessary. This is because the `PropertyHandler` that knows whether a guess is plotted is held by the `CompositeFunction` that is being replaced.

Unit tests added to cover new function.

**To test:**

- Open *Interfaces/Muon/Muon Analysis* and, on the Settings tab, tick "enable multiple fitting".
- Load some data on the Home tab, for example `MUSR00015190.nxs` (unit test data).
- Go to Data Analysis tab and set a fit function in the function browser at the top. Click *Display/Plot guess*.
- Edit one of the parameters of the function and verify the guess curve updates automatically.
- Click *Display/Remove guess* and verify the guess is removed.
- Plot the guess again, then run the fit and remove the guess. This should work OK and not plot a second guess when you try to remove it.

Turn off "enable multiple fitting", then close and reopen MuonAnalysis and follow the same steps. "Plot guess" should work here too (no change from before).

Fixes #18060.

[Release notes](https://github.com/mantidproject/mantid/blob/5fe0bc7b2b4c1d9287a537a18ab3303bab353bd6/docs/source/release/v3.9.0/muon.rst#id1) have been updated.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


KarlPalmen:
When testing, I found after fitting, if I remove the guess and then put it back, it appears in a new colour. This applies regardless of whether multiple fitting is enabled. Is this behaviour OK?

KarlPalmen:
Besides the guess colour issue. It works well :shipit:

tom-perkins:
Regarding the guess colour issue: I tried this in Mantid 3.7.2 and it was the same, so I think this is OK as it matches the old behaviour.

martyngigg:
@tom-perkins Could you merge master in here? The docs have a conflict in them now. Otherwise this looks good.

PR #18075, opened 2016-11-16
More Flexible LiveListeners

### LINKED ISSUE #18007: Implement required changes for MoreFlexibleILiveListener design
The [new design document](https://github.com/mantidproject/documents/blob/master/Design/MoreFlexibleILiveListener.md), added as part of https://github.com/mantidproject/documents/pull/7, aims to extend LiveListener functionality for ESS related requirements.

Part of the requirements have already been implemented in #17912.

The remainder of required changes are closely tied together, so it would be difficult to split them into separate PRs and this will likely be addressed as one big change.

The following features still need to be implemented:

- Multiple Listeners per Instrument should be supported
- StartLiveData should be able to create a custom LiveListener without referencing `facilities.xml`
- Custom interfaces should be updated to support the new features

MikeHart85:
This PR introduces many of the changes planned in the [More Flexible LiveListeners](https://github.com/mantidproject/documents/blob/master/Design/MoreFlexibleILiveListener.md) design document.

Overview of changes:

* New LiveListenerInfo class has been added to Mantid::Kernel
* FacilityInfo and InstrumentInfo classes have been updated to parse new xml syntax
* Facilities.xml has been updated to use new syntax 
* Entries like ENGIN-X and ENGIN-X_EVENT have **NOT** been merged into one instrument yet
* LiveListenerFactory now has two `create` function overloads:
  * Previous version that takes InstrumentName (and now optionally a LiveListenerInfo name)
  * New version that takes a LiveListenerInfo directly (can be used to construct listeners without facilities.xml)
* LiveDataAlgorithm has three new optional properties to override facilities.xml settings: 
  * Connection - Name of `<connection ... />` entry to use
  * Listener - Name of specific LiveListener class to use
  * Address - Address string to use
* StartLiveData dialog has been updated with new controls for above properties
* StartLiveData dialog supports overriding Listener class name and Address string from facilities.xml settings
* Removed connectivity check from MWRunFiles' LiveData Button (preliminary; this control will need rework to support multiple LiveListenerInfo's per InstrumentInfo)

**To test:**

* Ensure StartLiveData dialog changes make sense
* Ensure StartLiveData dialog can still be used for current workflows without issues
* Ensure StartLiveData can still be called from Python without issues

Fixes #18007.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


instrument/Facilities.xml
```diff
@@ -23,140 +24,179 @@
...
+    <livedata>
+      <connection name="histo" address="NDXMERLIN:6789" listener="ISISHistoDataListener" />
+    </livedata>
   </instrument>
 
   <instrument name="MERLIN_EVENT" shortname="MER">
     <technique>Neutron Spectroscopy</technique>
     <technique>TOF Direct Geometry Spectroscopy</technique>
-    <livedata address="NDXMERLIN:10000" listener="ISISLiveEventDataListener" />
+    <livedata>
```
OwenArnold: `MERLIN` and `MERLIN_EVENT` could probably be combined. Is there a compelling reason to do this now?

Framework/Kernel/inc/MantidKernel/InstrumentInfo.h
```diff
@@ -96,10 +104,11 @@ class MANTID_KERNEL_DLL InstrumentInfo {
   std::string m_shortName;        ///< Instrument short name
   ZeroPaddingMap m_zeroPadding;   ///< Run number-dependent zero padding
   std::string m_delimiter; ///< Delimiter between instrument name and run number
-  std::string m_liveListener;    ///< Name of the live listener class
-  std::string m_liveDataAddress; ///< Host & port for live data connection
   std::set<std::string>
       m_technique; ///< List of techniques the instrument can do
+
+  std::vector<LiveListenerInfo> m_listeners; ///< LiveListener connections
```
OwenArnold: If I'm being picky, shouldn't the LiveListeners be const. i.e `std::vector<const LiveListenerInfo>`?

Framework/Kernel/src/InstrumentInfo.cpp
```diff
@@ -121,16 +121,62 @@ std::string InstrumentInfo::filePrefix(unsigned int runNumber) const {
 }
 
 /// Returns the name of the live listener
-const std::string &InstrumentInfo::liveListener() const {
-  return m_liveListener;
+std::string InstrumentInfo::liveListener(const std::string &name) const {
+  if (!hasLiveListenerInfo())
+    return "";
```
OwenArnold: Is there a reason that an empty string is being returned rather than throwing an exception?

Framework/Kernel/src/InstrumentInfo.cpp
```diff
@@ -121,16 +121,62 @@ std::string InstrumentInfo::filePrefix(unsigned int runNumber) const {
...
 
 /** Returns the host & port to connect to for a live data stream
  *  No guarantees are given that the provided string is well-formed and valid
  *    - the caller should check this themselves
  */
-const std::string &InstrumentInfo::liveDataAddress() const {
-  return m_liveDataAddress;
+std::string InstrumentInfo::liveDataAddress(const std::string &name) const {
+  if (!hasLiveListenerInfo())
+    return "";
```
OwenArnold: Same question as above

Framework/Kernel/src/InstrumentInfo.cpp
```diff
@@ -121,16 +121,62 @@ std::string InstrumentInfo::filePrefix(unsigned int runNumber) const {
...
+  if (name.empty())
+    name = m_defaultListener;
+
+  // If no default connection specified, fallback to first connection
+  if (name.empty())
+    return m_listeners.front();
+
+  // Name specified, find requested connection
+  for (auto &listener : m_listeners) {
+    if (listener.name() == name)
```
OwenArnold: Are names stored all lower or upper case. Is this a consideration?

Framework/LiveData/src/LiveDataAlgorithm.cpp
```diff
@@ -178,20 +192,44 @@ bool LiveDataAlgorithm::hasPostProcessing() const {
...
-ILiveListener_sptr LiveDataAlgorithm::getLiveListener() {
+ILiveListener_sptr LiveDataAlgorithm::getLiveListener(bool start) {
   if (m_listener)
     return m_listener;
 
-  // Not stored? Need to create it
-  std::string inst = this->getPropertyValue("Instrument");
-  m_listener = LiveListenerFactory::Instance().create(inst, true, this);
+  // Create a new listener
+  m_listener = createLiveListener(start);
```
OwenArnold: Great! This is what I expected it to look like.

Framework/LiveData/src/LiveDataAlgorithm.cpp
```diff
@@ -178,20 +192,44 @@ bool LiveDataAlgorithm::hasPostProcessing() const {
...
+  std::string inst_name = this->getPropertyValue("Instrument");
+  std::string conn_name = this->getPropertyValue("Connection");
+
+  const auto &inst = ConfigService::Instance().getInstrument(inst_name);
+  const auto &conn = inst.liveListenerInfo(conn_name);
+
+  // See if listener and/or address override has been specified
+  std::string listener = this->getPropertyValue("Listener");
+  if (listener.empty())
+    listener = conn.listener();
```
OwenArnold: I'm wondering if we ought to log this at debug level?

MantidQt/CustomDialogs/src/StartLiveDataDialog.cpp
```diff
@@ -439,5 +466,60 @@ void StartLiveDataDialog::initListenerPropLayout(const QString &inst) {
...
+}
+
+/**
+ * Slot to update connection parameters when connection selected
+ *
+ * @param connection Name of selected live listener connection
+ */
+void StartLiveDataDialog::updateConnectionDetails(const QString &connection) {
+  // Custom connections just enable editting connection parameters
+  if (connection == "[Custom]") {
```
OwenArnold: Good, though I might also consider making ```[Custom]``` some kind of static const member to avoid future mistakes.

OwenArnold:
Code review looks pretty good. I notice that `LiveListenerInfo` did not get introduced with a unit test suite. `InstrumentInfo` does have one. https://github.com/mantidproject/mantid/blob/master/Framework/Kernel/test/InstrumentInfoTest.h

instrument/Facilities.xml
MikeHart85: As discussed, we will probably keep all the _EVENT instrument entries separate in this PR due to complications with the live data button in MWRunFiles.

Framework/Kernel/inc/MantidKernel/InstrumentInfo.h
MikeHart85: I gave it a try, but got a bunch of compiler errors. It seems that std::vector elements must be assignable.

Framework/Kernel/src/InstrumentInfo.cpp
MikeHart85: This was the previous behaviour (the m_liveListener and m_liveDataAddress members would be empty in this scenario). There is a test in InstrumentInfoTest that checks for it, and I remember several places in the code relied on it.

I think now all those places have been changed to use `hasLiveListenerInfo` instead and so only the test remains.

Do you think we should change it or leave as is?

Framework/Kernel/src/InstrumentInfo.cpp
MikeHart85: See above

Framework/Kernel/src/InstrumentInfo.cpp
MikeHart85: As discussed, when using the GUI the names will be whatever is in the Facilities.xml file, and so should always match. But when calling StartLiveData from Python, the user might use different casing. I should change these to case insensitive comparisons.

Framework/LiveData/src/LiveDataAlgorithm.cpp
MikeHart85: Sure, I can add something.

Framework/Kernel/src/LiveListenerInfo.cpp
```diff
@@ -0,0 +1,87 @@
...
+ * @param address Address which listener should use to connect
+ * @param name Name designator for this listener connection info
+ */
+LiveListenerInfo::LiveListenerInfo(const std::string &listener,
+                                   const std::string &address,
+                                   const std::string &name)
+    : m_name(name), m_address(address), m_listener(listener) {}
+
+bool LiveListenerInfo::operator==(const LiveListenerInfo &rhs) const {
+  return (this->address() == rhs.address() &&
```
MikeHart85: Here I would argue it's better to leave it case sensitive.

`address` often contains a hostname, which can be case sensitive in some circumstances. `listener` is a class name and is case sensitive at the latest when you try to create an instance of it using DynamicFactory (it uses `std::map::find()`). If two instances of LiveListenerInfo use different casing, there's a good chance they won't behave the same way.

Framework/LiveData/src/LiveDataAlgorithm.cpp
MikeHart85: Thinking over it some more, it might be confusing if we log here.

`listener` (or `address`) being empty here is nothing unusual. It will happen in both of these legitimate calls, for example:

```
>>> StartLiveData(Instrument='ISIS_Histogram', OutputWorkspace='wsOut', UpdateEvery=1, AccumulationMethod='Replace', PeriodList=[1,3], SpectraList=[2,4,6])
>>> StartLiveData(Instrument='ISIS_Histogram', Connection='histo', OutputWorkspace='wsOut', UpdateEvery=1, AccumulationMethod='Replace', PeriodList=[1,3], SpectraList=[2,4,6])
```

Also, `createLiveListener` can be called multiple times over the course of one `StartLiveData` call like that, or during interactive usage of the custom dialog for it. The values of "Listener" and "Address" may change between calls. Not every LiveListener created is actually used. In many places they are created temporarily just to check whether they support events or not, for example, and then discarded immediately.

instrument/Facilities.xml
OwenArnold: Please create a new Issue for this if you haven't already and assign it to yourself, and of course link the issues together.

Framework/Kernel/inc/MantidKernel/InstrumentInfo.h
OwenArnold: OK. Leave it as it is then.

Framework/Kernel/src/InstrumentInfo.cpp
OwenArnold: I'm not a big fan of this kind of silent error. However, as you point out this is existing behaviour, i think it's beyond the remit of this issue to go and fix that behaviour. You could create a new Issue for this perhaps.

Framework/Kernel/src/LiveListenerInfo.cpp
OwenArnold: I agree for the address. This should remain case sensitive. 

I do not think that the `listener` class name should be allowed to be case sensitive. But we're getting tight on time, and this is probably a good candidate for another issue anyway.

Framework/LiveData/src/LiveDataAlgorithm.cpp
OwenArnold: Ok. No point creating noise in the logs. Leave it as it is.

Framework/Kernel/test/LiveListenerInfoTest.h
```diff
@@ -0,0 +1,177 @@
...
+    info = inst.liveListenerInfo("n2");
+    TS_ASSERT_EQUALS(info.name(), "n2");
+    TS_ASSERT_EQUALS(info.address(), "A");
+    TS_ASSERT_EQUALS(info.listener(), "L");
+
+    TS_ASSERT_THROWS(inst.liveListenerInfo("n3"), std::runtime_error);
+  }
+
+  void test_manual_construction() {
+    TS_ASSERT_THROWS_NOTHING(LiveListenerInfo info());
```
OwenArnold: Switching it to `LiveListenerInfo{}` should do the trick

Framework/Kernel/test/LiveListenerInfoTest.h
```diff
@@ -0,0 +1,177 @@
...
+    TS_ASSERT_EQUALS(info.address(), "a");
+    TS_ASSERT_EQUALS(info.listener(), "l");
+
+    TS_ASSERT_THROWS_NOTHING(info = LiveListenerInfo("l", "a", "n"));
+    TS_ASSERT_EQUALS(info.name(), "n");
+    TS_ASSERT_EQUALS(info.address(), "a");
+    TS_ASSERT_EQUALS(info.listener(), "l");
+  }
+
+  void test_equality() {
```
OwenArnold: Overall, good testing. I like the fact you thought to include some logical equality test.

Framework/Kernel/src/InstrumentInfo.cpp
```diff
@@ -284,7 +285,7 @@ void InstrumentInfo::fillLiveData(const Poco::XML::Element *elem) {
   for (unsigned long i = 0; i < connections->length(); ++i) {
     auto *conn = dynamic_cast<Poco::XML::Element *>(connections->item(i));
     try {
-      m_listeners.emplace_back(LiveListenerInfo(conn));
+      m_listeners.emplace_back(LiveListenerInfo(this, conn));
```
OwenArnold: I think the `emplace_back` call here is not being properly utilised. As in the copy constructor will be called inplace. This should look like:
```cpp
m_listeners.emplace_back(this, conn);
```

OwenArnold:
@mantid-roman @peterfpeterson I have reviewed code and tried out the fake isis event and histo listeners. Would you mind trying this out with a few other listeners you are familiar with? We shouldn't release this without it being well tested.

Framework/Kernel/src/InstrumentInfo.cpp
MikeHart85: As discussed, this calls the move constructor rather than the copy constructor. But you're still right, even the move constructor call can be avoided by just passing in the construction arguments.

peterfpeterson:
The layout of the Connection Parameters should be switched to a grid so more space can be given to the address string. If possible, it would be better to show the start of the string rather than the end.
image

Functionally, everything is fine with ADARA data (tested on NOMAD). :+1:

Framework/PythonInterface/mantid/kernel/src/Exports/InstrumentInfo.cpp
```diff
@@ -46,9 +64,38 @@ void export_InstrumentInfo() {
...
+               args("self", "name"), "Returns a LiveListenerInfo instance for"))
+
+      .def("hasLiveListenerInfo", &InstrumentInfo::hasLiveListenerInfo,
+           arg("self"),
+           "Returns true if this instrument has at least one LiveListenerInfo")
+
+      .def("liveListenerInfoList", &InstrumentInfo::liveListenerInfoList,
+           return_value_policy<copy_const_reference>(), arg("self"),
+           "Returns all available LiveListenerInfo instances as a vector")
+
```
OwenArnold: No updates seem to have been made to the [tests](https://github.com/mantidproject/mantid/blob/master/Framework/PythonInterface/test/python/mantid/kernel)

MikeHart85:
@peterfpeterson 

Switch to QGridLayout as suggested, looks much better this way! :+1: 

Don't think there are any addresses longer than the textbox now but, just in case, I've added a `QLineEdit::home()` call after updating the text to bring the cursor back to the beginning.

peterfpeterson:
The new version looks good unless you want to fix the layout of the rest of the sidebar:
* `Starting Time` frame arrainging the radio buttons poorly
* `Processing` and `Post Processing` frames not being full width
* Some elements are centered (`Preserve Events`, `Accumulation Method`, and `At Run Transition`) when most are full justified

image

These are optional fixes. It can be merged with or without these changes.

VickieLynch:
Works for TOPAZ live data

MikeHart85:
@peterfpeterson 

Might as well do it here while we're at it. I've tried addressing the layout issues you've mentioned, see if the sidebar looks better this way.

AndreiSavici:
Works on HYSPEC and SEQUOIA. When I start the `Step Scan Analysis` interface on HYSPEC or SEQUOIA, the `Use live stream` button is greyed out

OwenArnold:
I'll retest it with these changes. Thanks everyone who's reviewed this thus far.

peterfpeterson:
There is a new warning in `InstrumentInfo.cpp` with clang

MikeHart85:
@AndreiSavici 

Thanks for pointing that out! Should always be enabled now. Could you check that it works as expected now?

@peterfpeterson 

It's the Exports version of InstrumentInfo.cpp... Clang doesn't seem to like those boost macros that are used for overloading at all. I've found a much simpler way of doing it without the macros, hopefully clang will like that better.

AndreiSavici:
I don't think it's related (I see the same in nightly), but I get a bad exception
screenshot from 2016-12-02 16-22-21

It works on Mantid 3.5.2, it fails on any version after that

peterfpeterson:
@AndreiSavici Please add a reference to a new issue for that one.

Everyone: Do we have consensus on merging this one? I'm in favor. @OwenArnold gets to mark it with the final squirrel.

AndreiSavici:
I've created issue #18184 for the step scan analysis

OwenArnold:
:shipit:

martyngigg:
I'm not sure what happened to the RHEL7 build - it seemed to never have been scheduled. I've kicked it off and I'll merge if/when it passes.

PR #18076, opened 2016-11-16
Added separator property to save algorithms using AsciiPointBase

### LINKED ISSUE #18056: Add separator property all save algorithms using AsciiPointBase
Currently for all ascii save algorithms using `AsciiPointBase`, there is no way of changing the separator character used for writing output to file stream (it currently defaults to 'tab' separator). This issue will address this by allowing the separator to be manually set by each algorithm as a property. Separators allowed will currently include comma and space, as well as tabs from before.

The algorithms using `AsciiPointBase` are:
- `SaveReflCustomAscii.cpp`
- `SaveReflThreeColumnAscii.cpp`
- `SaveANSTOAscii.cpp`
- `SaveILLCosmosAscii.cpp`

PranavBahuguna:
This PR adds a 'separator' property to `AsciiPointBase.cpp`. Previously, the separator used defaulted to 'tab' and now this property enables 'comma' and 'space' to also be specified (defaults to tab).

**To test:**

Ensure all unit tests pass.

Fixes #18056 

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


raquelalvarezbanos (commented):
Looks good. I only found two minor things that should be changed.

raquelalvarezbanos:
I've just realized we haven't mentioned this in the release notes. I think we should add a brief description as this change is going to be visible to users. Perhaps the best place is `framework.rst`, in "Improved" algorithms.

raquelalvarezbanos:
Looks good to me. :shipit: when all tests pass.

martyngigg:
@PranavBahuguna @raquelalvarezbanos These changes are nice and it's good that the code has been added in a common place. Unfortunately they *could* have an unintended consequence for user scripts that is in no way obvious from the C++ code. 

The Python functions, e.g. `SaveILLCosmosAscii`, which users would use from a script are automatically generated from the algorithm name & properties. The point that is not obvious from the code is that the declared property order in C++ defines the order of the arguments to the generated Python function. In this case having the new `Separator` property defined as it currently is will mean that any user script will now see `Separator` as the third argument so if they have not called it will keywords  there script will either break or produce unexpected results.

You can see this using the Python `help(SaveILLCosmosAscii)` function:

**after these changes**:

```python
Help on function SaveILLCosmosAscii in module mantid.simpleapi:

SaveILLCosmosAscii(InputWorkspace,Filename,Separator=None,LogList=None,UserContact=None,Title=None, Version=1)
    Saves a 2D workspace to a ascii file.
    
```

**master**:

```python
Help on function SaveILLCosmosAscii in module mantid.simpleapi:

SaveILLCosmosAscii(InputWorkspace,Filename,LogList=None,UserContact=None,Title=None, Version=1)
    Saves a 2D workspace to a ascii file.
```

I think having the separator controllable and keeping the code common is good. Can we slightly modify how we use it? Perhaps the base class could have an `appendSeparatorProperty()` method that is called by the children at the appropriate time?

raquelalvarezbanos:
Hi @martyngigg thanks for reviewing this. I assume the only place where we can safely call the new `appendSeparatorProperty()` method is at the end of each of the `extraProps()` methods defined in the derived classes, otherwise the children would have the same problem? Thanks for pointing this out, I didn't realise this was going to cause problems.

martyngigg:
@raquelalvarezbanos No problem. It's not obvious in any way and we don't have a way of checking for this stuff at the moment which is unfortunate. You are correct the only safe place is at the end of the current properties list.

martyngigg:
Thanks for making those changes, @PranavBahuguna. This is good to go now.

PR #18087, opened 2016-11-21
Crystal field monte carlo parameter estimation

### LINKED ISSUE #15704: Crystal field: Monte Carlo estimation of initial parameters
Implement a function (C++ or python) which estimates initial crystal field parameters the way it is done in FOCUS (Monte Carlo).

This is a part of #15684

mantid-roman:
Added a Monte Carlo algorithm that estimates initial values of a fitting function. Use this algorithm in the Crystal Field python interface.

**To test:**

This script tests `EstimateFitParameters` algorithm:

```
x = [-8.4, -7.4, -6.4, -5.4, -4.4, -3.4, -2.4, -1.4, -0.4, 0.6,  1.6,  2.6,  3.6,  4.6,  5.6,  6.6,  7.6,  8.5]
y = [5.899424685451e-06, 0.0007065291085075, 0.03312893859911,
        0.6260554060356,    4.961949842664,     17.4766663495,
        30.00494980772,     29.34773951093,     20.53693564978,
        12.73990150587,     7.739398423952,     4.694362241084,
        2.847186128946,     1.726807862958,     1.047276595461,
        0.6351380161499,    0.3851804649058,    0.1957211887562]
        
ws = CreateWorkspace(x, y)
fun = 'name=BackToBackExponential,S=1.1,constraints=(50<I<200,0.1<A<300,0.01<B<10,-5<X0<0,0.001<S<4)'

from mantid.api import AlgorithmManager
alg = AlgorithmManager.createUnmanaged('EstimateFitParameters')
alg.initialize()
alg.setProperty('Function', fun)
alg.setProperty('InputWorkspace', ws)
alg.setProperty('CostFunction', 'Unweighted least squares')
alg.setProperty('NSamples', 1000)
# alg.setProperty('FixBadParameters', True)
alg.setProperty('Seed', 1234)
alg.execute()
function = alg.getProperty('Function').value
print(function)
EvaluateFunction(str(function),ws,OutputWorkspace='estimation')
Fit(str(function),ws,Output='fit')
```

The calculated spectrum in `estimation` workspace should be close to the fit data.
The fit should be well converged.

Uncomment the line with `FixBadParameters` property, rerun the script. The fitting function should now 
contain a tie on `A`.

Comment out the `Seed` property. Run the script several times with `FixBadParameters` set to `True` and `False`. Note that `Fit` doesn't always succeed. Also note that with `FixBadParameters` set to `True` the rate of success is higher.

For testing Crystal Field estimation use this script:
```
from CrystalField import CrystalField, CrystalFieldFit

# Create some crystal field data
origin = CrystalField('Ce', 'C2v', B20=0.37737, B22=3.9770, B40=-0.031787, B42=-0.11611, B44=-0.12544,
                      Temperature=44.0, FWHM=1.1)
x, y = origin.getSpectrum()
ws = makeWorkspace(x, y)

# Define a CrystalField object.
cf = CrystalField('Ce', 'C2v', B20=0, B22=0, B40=0, B42=0, B44=0,
                  Temperature=44.0, FWHM=1.0, ResolutionModel=([0, 100], [1, 1]), FWHMVariation=0)

# Set any ties on the field parameters.
cf.ties(B20=0.37737)
# Define the search intervals. Only these parameters will be estimated.
cf.constraints('2<B22<6', '-0.2<B40<0.2', '-0.2<B42<0.2', '-0.2<B44<0.2')
# Create a fit object
fit = CrystalFieldFit(cf, InputWorkspace=ws)
# Find initial values for the field parameters.
fit.monte_carlo(NSamples=1000)
# Run fit
fit.fit()
```

Fixes #15704.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


mantid-roman:
@martyngigg Hi. This PR is mostly about MonteCarloParameters algorithm and very little of crystal field. Crystal field can be considered as a testing case here. Actual crystal field monte carlo will be in a separate issue.

martyngigg (commented):
Generally I think this is great. I just have a couple of comments.

* Should we be slightly more specific in the algorithm name, maybe `MonteCarloFitParameters` or even `EstimateFitParameters`. Is the Monte Carlo part important enough to be in the name?
* If I just run the algorithm from the standard GUI dialog then I can't see the parameters at all as the only access is through the `Function` property. Should we have an optional output to a table like `Fit` too?
* The final thing is whether we should use a separate output property for the new estimated function? The `InOut` style can be confusing, especially when used with the functional-style Python api.

mantid-roman:
I agree with the first point, `EstimateFitParameters` sounds much better.

With the output table I am not so sure. I cannot see a user copying values from it by hand. Instead the algorithm can be used by the fit browser. Actually, none of the fitting algorithms can be used with the standard dialogs. And this is because there is no workspace for a function. I think this issue should be addressed more generally.

The main reason for `InOut` is efficiency, no copying or string conversions. Again here, we don't have a proper python api for functions and that's a problem. It makes all work with functions confusing.

martyngigg:
@mantid-roman Fair point about the dialog. It's certainly not the highest priority and that part can be addressed separately.

docs/source/interfaces/CrystalFieldPythonInterface.rst
```diff
@@ -351,5 +351,39 @@ Then call `fit()` method::
     fit.fit()
     
 After fitting finishes the `CrystalField` object updates automatically and contains new fitted parameter values.
+
+Finding Initial Parameters
+--------------------------
+
+If the initial values of the fitting parameters are not known they can be estimated using `monte_carlo()` method.
+It randomly searches the parameter space in a given region such that the calculated spectra are as close to the
+fit data as possible. The method uses :ref:`MonteCarloParameters <algm-MonteCarloParameters>` internally. See
```
martyngigg: We just need to update this documentation reference with the new algorithm name and then I think this is good to go.

martyngigg (changes requested):

mducle:
The examples works for me, but when I tried to use it with `CrystalFieldSpectrum` and the TmB4 dataset, it gives a `Cost function changed number of parameter` error about 99% of the time. I guess though that this is something we can handle in the Crystal Field Python interface?

mantid-roman:
@mducle We will deal with it in the following issue. We should fix all peak parameters to make that error go away.

martyngigg:
@mantid-roman Just so that I am clear. The issue @mducle raised will be dealt with in a separate issue?

Framework/CurveFitting/src/Algorithms/EstimateFitParameters.cpp
```diff
@@ -0,0 +1,381 @@
...
+        variances[i] += p * p;
+      }
+    }
+    means *= 1.0 / double(nSelection);
+    variances *= 1.0 / double(nSelection);
+    for (size_t i = 0; i < nParams; ++i) {
+      auto mean = means[i];
+      auto sigma = sqrt(variances[i] - mean * mean);
+      if (sigma >= distributionParams[i].second) {
+        badness[i] += double(it + 1);
```
mducle: What is `badness` used for? I guess it's an indication that the parameter distribution is not converging... but it doesn't seem to be used anywhere else in the code.

mantid-roman:
@martyngigg #18133

Framework/CurveFitting/src/Algorithms/EstimateFitParameters.cpp
mantid-roman: You are right It isn't used, I'll delete it.

martyngigg (approved):

martyngigg:
:shipit:

PR #18093, opened 2016-11-22
Fix coverity warning in PDLoadCharacterizations

peterfpeterson:
This adds a constructor so default values for member variables can be set.

**To test:**

Just review the code

Fixes coverity issue 1366006.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/DataHandling/src/PDLoadCharacterizations.cpp
+++ b/Framework/DataHandling/src/PDLoadCharacterizations.cpp
@@ -87,9 +87,6 @@ extra_columns(const std::vector<std::string> &filenames) {
 }
 }
 
-PDLoadCharacterizations::PDLoadCharacterizations()
-    : hasExtras(false), canColumnNames() {}
-
 //----------------------------------------------------------------------------------------------
 /// Algorithm's name for identification. @see Algorithm::name
 const std::string PDLoadCharacterizations::name() const {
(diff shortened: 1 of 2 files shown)


DavidFair:
I think we can completely remove the constructor with some minor changes:
`hasExtras` isn't used within the code at all so this can easily be removed.

`canColumnNames` is a vector of strings which is used in two places - in `exec` and `readCharInfo`. In `readCharInfo` we use the size of the container to determine how much padding to apply so instead we can just hold this size_t as a member variable. If we use C++11 to initialize on declaration to 0 we no longer need the constructor. 

I've added a git patch file with these changes as I didn't want to start pushing to somebody elses branch (has to be a .txt file because of GitHub sigh): [PDLoadChara_changes.txt](https://github.com/mantidproject/mantid/files/608709/PDLoadChara_changes.txt)

Framework/DataHandling/src/PDLoadCharacterizations.cpp
```diff
@@ -354,9 +347,11 @@ void PDLoadCharacterizations::readCharInfo(std::ifstream &file,
     row << boost::lexical_cast<double>(splitted[9]);  // tof_max
     row << boost::lexical_cast<double>(splitted[10]); // wavelength_min
     row << boost::lexical_cast<double>(splitted[11]); // wavelength_max
+
     // pad all extras with empty string
-    for (const auto name : this->canColumnNames)
+    for (size_t i = 14; i < num_of_columns; ++i) {
```
DavidFair: Could we put a comment on where 14 comes from?

DavidFair:
LGTM :shipit: once the tests finish

Framework/DataHandling/src/PDLoadCharacterizations.cpp
```diff
@@ -170,7 +164,7 @@ void PDLoadCharacterizations::exec() {
   wksp->addColumn("double", "tof_max");
   wksp->addColumn("double", "wavelength_min");
   wksp->addColumn("double", "wavelength_max");
-  for (const auto name : this->canColumnNames) {
+  for (const auto name : canColumnNames) {
```
quantumsteve: `const auto &`?

quantumsteve (changes requested):

quantumsteve (approved):

quantumsteve:
@peterfpeterson I'll merge this once the build finish.

PR #18095, opened 2016-11-22
Running external local reconstruction

### LINKED ISSUE #17940: Imaging GUI: Work on running a local reconstruction using TomoPy
Find/fix any problems that arise when trying to run a local reconstruction on a local dataset using TomoPy

Something that might help installing TomoPy on Windows: https://github.com/tomopy/tomopy/issues/110
Dynalist -> [here](https://dynalist.io/d/i4km22FyqWAe9Azdkm62QJGp#z=KMcPmvyoa533jr6H5qHP3wU6)

dtasev:
Description of work.

- Local reconstructions can now be launched from Mantid Imaging interface
  - On Linux only an actual reconstruction can be launched
  - On Windows we can still check if the external process runs, but most likely we'll get an import error
- The system settings weren't updated when the External Interpreter path and the Scripts location was being changed in the `System` tab -> `Execution of reconstruction jobs` -> `Local runs`
- Currently only a **single** reconstruction is possible, as running a second one _while_ the first is running will prompt the user and if accepted will terminate the first process
- Updated the reconstruction scripts to run on both Python 2 and 3 
  - However the tools themselves (TomoPy, Astra, Savu, etc) don't run on Python 3

Prerequisites:
- Python with TomoPy, I used [Anaconda2](https://www.continuum.io/downloads) for both Windows and Linux
- TomoPy v0.1.x is supported
  - This was tested locally with TomoPy v0.1.11 (installed with Anaconda, 2.7)
  - A temporary ["patch"](https://github.com/mantidproject/mantid/pull/18095/files#diff-85d4ffec21b1b51573f92591870ed3ecR619) has been added that allows TomoPy >1.x.x to run as well, however the results are not guaranteed to be accurate
- Test data `RB000888_test_stack_larmor_summed_201510` is available on `babylon5/Public/DimitarTasev` (~76.5 MB)

To test (both Windows and Linux):
-----------
- Build and open interface
- On the `Settings` tab check the external interpreter path, this will probably have just `python`, so make sure it points to the python that has `tomopy` installed
  - The external `python` process initialises it's path from the `PYTHONPATH` variable and it seems to ignore the `PATH` variable, so make sure the `tomopy` package is available in the `PYTHONPATH` variable
- Load a single image (or a stack, it doesn't matter as even a single image will set the paths in the `Setup` tab)
- Check that the paths in the `Setup` tab are correct
  - _Optional_: Go to the `ROI etc` tab and put some settings on (the settings don't matter for this PR)
- On the `Run` tab change resource to local and select `TomoPy` and click reconstruct (setup is optional, there are default values loaded with each tool)
  - This has only been tested with the `gridrec` algorithms (also the default one)
- Check that _any_ output is shown in the `Results Log`, as this should contain any errors as well as standard output, as long as the process started
  - Currently there is no error shown if the process isn't started, this is because Qt doesn't use an exception to signify that it failed to start
    - However there will be nothing added to the `Reconstruction jobs` in the `Run` tab, because no `started()` signal will be emitted from the external process and thus no job will be added 

Fixes #17940.

#### Release Notes:
[`.../release/imaging.rst`](https://github.com/mantidproject/mantid/pull/18095/files#diff-e678fb82fbdc92d52b938c69b0858a2e)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Tomography/TomographyIfaceModel.h
```diff
@@ -250,7 +253,7 @@ class MANTIDQT_CUSTOMINTERFACES_DLL TomographyIfaceModel {
   std::string buildOutReconstructionDir(const std::string &samplesDir,
                                         bool) const;
 
-  bool processIsRunning(int pid);
+  bool processIsRunning(qint64 pid);
```
DavidFair: Does this modify the process or just return a status - it could be `const` to enforce the fact it doesn't accidentally change the process

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Tomography/TomographyProcess.h
```diff
@@ -0,0 +1,87 @@
...
+  Q_OBJECT
+public:
+  // we want a nullptr for parent so we can move it to a thread
+  TomographyProcess() : QProcess(nullptr) {}
+
+  // intentionally copy the vector
+  void setup(const std::string runnable, const std::vector<std::string> &args,
+             const std::string &allOpts) {
+    m_allArgs = allOpts;
+    m_runnable = std::move(QString::fromStdString(runnable));
```
DavidFair: I don't think you need a `std::move` here as the QString is an rvalue.

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Tomography/TomographyProcess.h
```diff
@@ -0,0 +1,87 @@
...
+    m_args = std::move(constructArgumentsFromVector(args));
```
DavidFair: Same with `std::move` here - the compiler can do RVO if it thinks its appropriate

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Tomography/TomographyProcess.h
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  std::string getRunnable() const { return m_runnable.toStdString(); }
+  std::string getArgs() const { return m_allArgs; }
+
+  qint64 getPID() const {
+    auto pid = this->pid();
```
DavidFair: Can this ever fail? Do we check for failure?

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Tomography/TomographyProcess.h
```diff
@@ -0,0 +1,87 @@
...
+
+public slots:
+  /** This method should be used to start the worker as it passes the setup
+   * runnable and args parameters into the base start method
+  */
+  void startWorker() { start(m_runnable, m_args); }
+
+private:
+  QStringList
+  constructArgumentsFromVector(const std::vector<std::string> &args) const {
```
DavidFair: Could we rename this to something like
`convertStringVectorToQStringVector` to reflect what it does (the name I proposed isn't quite right either but is a step in the right direction)

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Tomography/TomographyProcess.h
```diff
@@ -0,0 +1,87 @@
...
+    QStringList list;
+    list.reserve(static_cast<int>(args.size()));
+
+    for (auto &arg : args) {
```
DavidFair: `const`

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Tomography/TomographyThread.h
```diff
@@ -0,0 +1,85 @@
...
+    connect(this, SIGNAL(finished()), this, SLOT(deleteLater()),
+            Qt::DirectConnection);
+
+    worker->moveToThread(this);
+    m_worker = worker;
+  }
+
+public slots:
+  void workerFinished(int exitCode) {
+    UNUSED_ARG(exitCode);
```
DavidFair: Should we not check the exitCode in case something unusual has happened?

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Tomography/TomographyThread.h
```diff
@@ -0,0 +1,85 @@
...
+    // queue up object deletion
+    m_worker->deleteLater();
+    emit workerFinished();
+  }
+
+  void readWorkerStdOut() const {
+    auto *worker = qobject_cast<TomographyProcess *>(sender());
+    QString output(worker->readAllStandardOutput());
+    emit stdOutReady(output);
```
DavidFair: This can be collapsed to
`emit stdOutRead(worker->readAllStandardOutput());`

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Tomography/TomographyThread.h
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  void readWorkerStdErr() const {
+    auto *worker = qobject_cast<TomographyProcess *>(sender());
+    QString output(worker->readAllStandardError());
+    emit stdErrReady(output);
```
DavidFair: Ditto ^

MantidQt/CustomInterfaces/src/Tomography/TomographyIfaceModel.cpp
```diff
@@ -805,8 +752,10 @@ TomographyIfaceModel::loadFITSImage(const std::string &path) {
   }
 }
 
-void TomographyIfaceModel::logMsg(const std::string &msg) {
-  g_log.notice() << msg << '\n';
+void TomographyIfaceModel::logMsg(const std::string &msg) { g_log.notice(msg); }
```
DavidFair: Do we need these wrapped in a separate method - could we not call `g_log` directly instead?

MantidQt/CustomInterfaces/src/Tomography/TomographyIfacePresenter.cpp
```diff
@@ -554,14 +555,97 @@ void TomographyIfacePresenter::processRunRecon() {
...
   }
   processRefreshJobs();
 }
 
+void TomographyIfacePresenter::setupAndRunLocalReconstruction(
+    const std::string &runnable, const std::vector<std::string> &args,
+    const std::string &allOpts) {
+
+  if (m_reconRunning) {
+    auto result = m_view->userConfirmation(
```
DavidFair: `const`

MantidQt/CustomInterfaces/src/Tomography/TomographyIfacePresenter.cpp
```diff
@@ -554,14 +555,97 @@ void TomographyIfacePresenter::processRunRecon() {
...
+          SLOT(workerFinished()));
+
+  connect(worker, SIGNAL(started()), this, SLOT(addProcessToJobList()));
+
+  m_model->doLocalRunReconstructionJob(runnable, args, allOpts, *m_workerThread,
+                                       *worker);
+  m_reconRunning = true;
+}
+
+void TomographyIfacePresenter::workerFinished() { m_reconRunning = false; }
```
DavidFair: The comment for this method seems to have moved down to `reconProcessFailedToStart`

MantidQt/CustomInterfaces/src/Tomography/TomographyIfaceViewQtGUI.cpp
```diff
@@ -1910,20 +1917,34 @@ void TomographyIfaceViewQtGUI::systemSettingsNumericEdited() {
 }
 
 void TomographyIfaceViewQtGUI::resetSystemSettings() {
-  auto reply = QMessageBox::question(
-      this, "Reset Confirmation", "Are you sure you want to <br><strong>RESET "
-                                  "ALL</strong> System settings?<br>This "
-                                  "action cannot be undone!",
-      QMessageBox::Yes | QMessageBox::No);
+  auto title = "Reset Confirmation";
+  auto body = "Are you sure you want to <br><strong>RESET "
```
DavidFair: `const`

scripts/Imaging/IMAT/tomo_reconstruct.py
```diff
@@ -55,6 +56,9 @@
 # So insert in the path the directory that contains this file
 sys.path.insert(0, os.path.split(path.dirname(__file__))[0]) # noqa
 
+# import pydevd
+# pydevd.settrace('localhost', port=61845, stdoutToServer=True, stderrToServer=True)
```
DavidFair: Forgot to take out the lines for your debugger

DavidFair (changes requested):

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Tomography/TomographyProcess.h
DimitarTasev: It can't fail, but if the process isn't started then it returns 0, in which case we're fine

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Tomography/TomographyThread.h
DimitarTasev: All of the error stream is printed in the Result log, so the error can be seen there. This will be an issue when #18102 is implemented, but that will require changing the whole structure of output handling anyway.

MantidQt/CustomInterfaces/src/Tomography/TomographyIfaceModel.cpp
DimitarTasev: The presenter doesn't have access to the `g_log`, all the logging is being done through the model

DimitarTasev:
http://builds.mantidproject.org/job/pull_requests-ubuntu/10801/console http://builds.mantidproject.org/job/pull_requests-ubuntu/10803/console seem to segfault randomly, however I think it's connected to that QThread in here https://github.com/mantidproject/mantid/pull/18095/commits/c67532c634cb4f3c4e79ab3776fb4108e8ae84c9#diff-308900e88f8d01f3019a17a328f3ab0eL788. I'll try removing it after the other builds finish with anything

DavidFair (approved):

DavidFair:
Some review points - these may be separate things I have picked up or related to the above changes:
- If local reconstruction fails (e.g. import error) we currently show the status as a green done instead of failed. Is there an issue open for for this
- Additionally if we cancel a reconstruction the status should show cancelled 
- A way to clear all entries in the Reconstruction Jobs would help avoid it becoming cluttered. Also we should show the most recent first in my opinion else you are always looking at completed jobs. 
- Cancelling the currently running reconstruction doesn't stop the process. This can be seen by using cancelling and restarting multiple times and watching task manager / top
- (As an aside) The experiment reference is not enforced I ran all my constructions with `RB000XXX` Also I would change it to 4 `X`s as I would guess they will rapidly get through 999 experiments in a few years. 
- I'm not sure if its an issue with my own machine. The logging always ended up off screen so I would have to always scroll up to see the status of the reconstruction
- A nice to have is when the user selects the folder for the Reconstruction scripts the Presenter immediately checks if the script exists where it expects and sets that box to red if it cannot find it with an error in the logging window.
- Additionally if we can catch the error `python: can't open file *path/to/script.py` or wrap it with some extra details in the logger so a user knows where the option needs to be set. 

---
- ~~We need to add documentation helping user either set up their `PythonPath` or allow them to browse to their Conda install and append the path for them if they are getting the package from Conda~~
- ~~I also had to install PyFits - this is available on `PIP` so we could distribute it with Mantid if we use it to load FIT images~~ - See below about Conda installation 

Regarding Conda we could automate the setup for the users doing the following:
```
conda create --name tomoEnv 
conda install -n tomoEnv -c dgursoy tomopy
conda install -n tomoEnv -c CEFCA pyfits      (Note: This one does not have conflicting dependencies)
```
The environment will be at a known path relative to the Conda install. If you were in a console Conda can setup the environment by doing `source activate tomoEnv`
I think we should point to the Conda root folder and not the interpreter. This avoids us modifying the Python path too.

DavidFair (changes requested):
See above points ^

DimitarTasev:
```
If local reconstruction fails (e.g. import error) we currently
show the status as a green done instead of failed. 
```
That's because currently we just check if the process is finished, I agree it is a good change, but will be on the backlog for now.
I'll open an issue for this, it will require a bit of changes in TomographyThread, as it would have to store the process ID, and the exit status we have already (but we're just ignoring it), we are going to have to signal the presenter to update the specific ID.

Issue #xxxx

... todo rest

MantidQt/CustomInterfaces/src/Tomography/TomographyIfaceModel.cpp
```diff
@@ -560,6 +561,19 @@ void TomographyIfaceModel::refreshLocalJobsInfo() {
   }
 }
 
+void TomographyIfaceModel::updateProcessInJobList(const qint64 pid,
+                                                  const int exitCode) {
+  // cast to string from qint64 so we can compare
+  const std::string processPID = std::to_string(static_cast<int>(pid));
```
DavidFair: What happens if the PID falls outside of the range of an integer - is there any guarantees that it will always be within the range of an int?

MantidQt/CustomInterfaces/src/Tomography/TomographyIfaceModel.cpp
```diff
@@ -560,6 +561,19 @@ void TomographyIfaceModel::refreshLocalJobsInfo() {
...
+  for (auto &job : m_jobsStatusLocal) {
+    if (job.id == processPID) {
```
DavidFair: `if (job.id == processPID && exitCode == 1){`

MantidQt/CustomInterfaces/src/Tomography/TomographyIfacePresenter.cpp
```diff
@@ -651,17 +668,10 @@ bool TomographyIfacePresenter::isLocalResourceSelected() const {
...
-    m_model->doRefreshJobsInfo("Local");
-    m_view->updateJobsInfoDisplay(m_model->jobsStatus(),
-                                  m_model->jobsStatusLocal());
-    return;
-  }
-
   m_model->doRefreshJobsInfo(m_view->currentComputeResource());
 
   {
+    // BUG :: still crash here if closed during running process
```
DavidFair: Does this bug still exist - I'm not a fan of leaving comments like this as they usually get forgotten about when the bug is fixed. Instead you could link to the line number in the issue.

MantidQt/CustomInterfaces/src/Tomography/TomographyIfacePresenter.cpp
```diff
@@ -671,11 +681,13 @@ void TomographyIfacePresenter::processRefreshJobs() {
 }
 
 void TomographyIfacePresenter::processCancelJobs() {
-  if (m_model->loggedIn().empty())
-    return;
-
   const std::string &resource = m_view->currentComputeResource();
-  if (!isLocalResourceSelected()) {
+  if (isLocalResourceSelected()) {
```
DavidFair: `if (isLocalResourceSelected() && userConfirmationToCancelRecon()){`

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Tomography/TomographyThread.h
```diff
@@ -7,6 +7,9 @@
 #include <QString>
 #include <QThread>
 
+// remove for DEBUG only
```
DavidFair: What does this comment mean? Should this have been removed?

DavidFair (changes requested):
Some minor changes requested

MantidQt/CustomInterfaces/src/Tomography/TomographyIfaceModel.cpp
DimitarTasev: I've removed the static_cast<int>, `to_string` seems to work with a qint64, because that is just a [typedef of an int](http://doc.qt.io/qt-4.8/qtglobal.html#qint64-typedef).

Just for fun:
- Max PID on Linux is 4194303, by default it's 32768 ([source](http://stackoverflow.com/questions/6294133/maximum-pid-in-linux))
- Max PID on Windows is DWORD_MAX, but it must be aligned to 4, so the max PID is 0xFFFFFFFC or 4294967292 ([source](http://stackoverflow.com/questions/17868218/what-is-the-maximum-process-id-on-windows))

So this was a potential very bad and uncaught `static_cast` failure on PID's > 2147483647.

DavidFair (approved):

DavidFair:
Everything works as described or has an associated issue with it. :shipit:

PR #18096, opened 2016-11-22
Initial work migrating Polaris instrument into ISIS_Powder

### LINKED ISSUE #17949: PEARL Scripts - Migration of Polaris
The next steps:

---
- [x] Replicate / Tidy existing workflow from existing scripts and migrate them into new script
- [x] As part of that figure out and isolate instrument specific steps from main steps. For example PEARL uses anvils for their samples and will divide by a pre-calculated file to offset this. This doesn't apply to POLARIS
- [x] Create system tests 
- [x] ~~Unit tests~~ - *Moved for when we move the logic into routines see #17831*
- [x] Ensure OS.path is used throughout when manipulating paths
- [x] Add chopper on/off to the calibration and API options
- [x] Move to using Run number calibration mapping file
- [x] Figure out what options in current pref files do
- [x] Decide on new format to make it a bit more robust to errors in it.
- [x] Move away from dictionaries for calibration and cycle and move them into ~~2~~ a smaller class~~es~~. This helps avoid key errors as they are inspect-able attributes 
- [x] Split `common` into separate files for focusing and calibration....etc.
- [x] Allow ranges for exceptions
- [x] Allow ranges for Vanadium Runs and remove the usage of file names
- [x] Move away from requiring a `raw_data_dir` instead rely on user directories in Mantid - e.g. use `Load`
- [x] Create calibration structure and move to using that
- [x] As part of the above the masking file can move into the TLD of calibration
- [x] Fingerprint SAC pre-calc output file and remove from calibration file
- [x] Enforce chopper on/off selection
- [x] Create instrument focus specialised calls
- [x] ~~Remove solid angle corrections~~ - do a run with/without this enabled to compare
- [x] Move solid angle to be an optional step for instruments who need it
- [x] Implement logic for determining empty runs
- [x] POLARIS Van Absorptions must be set to generate or it fails
- [x] Create diagram for finished workflow for next presentation

DavidFair:
Description of work.

~~**Note this is cannot be merged before PR #18090 as it uses YAML to hold a mapping used throughout the POLARIS script**~~

This PR brings the POLARIS instrument from the original Crystallography scripts onto the ISIS_Powder scripts. A list of things that were changed in this pull requests as part of the migration can be found under #17949 . 

This is still a work in progress. However as the scripts is in a known working state this PR is to merge the current changes in before additional work documented here: #17831 . It also allows us to get feedback from testers using nightly builds.
The next steps include refactoring and tidying, making the scripts more user friendly and incorporating feedback from initial testers and this PR.

Any feedback or suggested changes are very welcome.

**To test:**
As the script is in development there is not much testing which can be performed.
Ensure the system tests pass proving that existing PEARL scripts still work and provide feedback on scripts.

Fixes #17949 .

Does not need to be in the release notes as it is still in progress. 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


scripts/Diffraction/isis_powder/routines/common.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def _load_sum_file_range(run_numbers_list, instrument):
+    read_ws_list = []
+
+    _check_load_range(list_of_runs_to_load=run_numbers_list)
+
+    for run_number in run_numbers_list:
+        file_name = instrument._generate_inst_file_name(run_number=run_number)
+        read_ws = mantid.Load(Filename=file_name, LoadLogFiles="0")
```
soininen: I think you can give `Load` a list of filenames to load which would give you a group workspace. `Load` also checks if more than 100 (?) files are given, so `_check_load_range` would not be needed. However, I don't know if you would be happy with a group workspace. Anyway, might be worth checking that option out.

scripts/Diffraction/isis_powder/routines/common.py
```diff
@@ -0,0 +1,165 @@
...
+    run_boundaries = run_number_string.replace('_', '-')  # Accept either _ or - delimiters
+    run_range_lists = __run_number_generator(processed_string=run_boundaries)
+    out_range = []
+    for range_list in run_range_lists:
+        out_range.extend(range_list)
+
+    return out_range
+
+
+def __run_number_generator(processed_string):
```
soininen: It is a bit hackish, but you could try to use here
```python
numberProperty = mantid.kernel.IntArrayProperty('dummy_property', processed_string)
return numberProperty.value
```
This should do the string parsing for you.

scripts/Diffraction/isis_powder/routines/common.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+    try:
+        if not g_debug:
+            _remove_ws_wrapper(ws=ws_to_remove)
+    except NameError:  # If g_debug has not been set
+        _remove_ws_wrapper(ws=ws_to_remove)
+
+
+def _remove_ws_wrapper(ws):
+    mantid.DeleteWorkspace(ws)
```
soininen: As an extra check, you maybe could do `if mtd.doesExist(ws):` here.

soininen (changes requested):

scripts/Diffraction/isis_powder/pearl.py
```diff
@@ -262,7 +371,18 @@ def _get_monitor_spectrum(self, run_number):
...
 
-    # Support for old API
+    def _generate_raw_data_cycle_dir(self, run_cycle):
+        if self._skip_appending_cycle_to_raw_dir():
+            return self.raw_data_dir
+        str_run_cycle = str(run_cycle)
+
+        # Append current cycle to raw data directory
+        generated_dir = os.path.join(self.raw_data_dir, str_run_cycle)
+        generated_dir += '/'
```
DavidFair: This is support for `pearl_routines` which previously made nasty assumptions such as this. These hacks have been moved into a separate class called `PearlRoutinesWrapper` which can be removed when the aforementioned script is removed. 

Additionally I had to look this up myself apparently Windows accepts these dividers because it silently treats the path as a URI to a file when it encounters them.

scripts/Diffraction/isis_powder/pearl_routines/fmode_output.py
```diff
@@ -0,0 +1,192 @@
...
+
+        append = True
+        index += 1
+    return output_list
+
+
+def _focus_mode_trans(output_file_paths, atten, instrument, calibrated_spectra):
+    summed_ws = mantid.CloneWorkspace(InputWorkspace=calibrated_spectra[0])
+    for i in range(1, 9):  # Add workspaces 2-9 to workspace 1
+        summed_ws = mantid.Plus(LHSWorkspace=summed_ws, RHSWorkspace=calibrated_spectra[i])
```
DavidFair: Good point I didn't realise there was an algorithm. Added as part of the refactoring work [here](https://github.com/mantidproject/mantid/issues/18108)

scripts/Diffraction/isis_powder/polaris.py
```diff
@@ -0,0 +1,373 @@
...
+        for ws in extracted_spectra:
+            common.remove_intermediate_workspace(ws)
+
+        return output
+
+    def _generate_vanadium_absorb_corrections(self, calibration_full_paths, ws_to_match):
+        absorb_ws = mantid.CloneWorkspace(InputWorkspace=ws_to_match)
+
+        # TODO move all of this into defaults
+        cylinder_sample_height = str(4)
```
DavidFair: 👍 I haven't tried passing them as ints/floating points yet. Originally I was matching the original output of the Crystallography scripts and wanted to avoid the chance of floating point errors so opted to pass them by string instead. 
Additionally these were going to be moved out into an advanced configuration file with the future work so are only here as a temporary stopgap so I'll remove the casts in the process too.

scripts/Diffraction/isis_powder/routines/common.py
```diff
@@ -0,0 +1,165 @@
...
+    else:
+        load_monitor_ws = _load_sum_file_range(run_numbers_list=number_list, instrument=instrument)
+        load_monitor_ws = mantid.ExtractSingleSpectrum(InputWorkspace=load_monitor_ws, WorkspaceIndex=monitor_spectra)
+
+    return load_monitor_ws
+
+
+def subtract_sample_empty(ws_to_correct, empty_sample_ws_string, instrument):
+    output = ws_to_correct
+    if empty_sample_ws_string is not None:
```
DavidFair: 👍 This was because the method used to take a string or a list. Nice catch.

scripts/Diffraction/isis_powder/routines/common.py
DavidFair: I'm reluctant to bring group workspaces in as what ends up happening is all future algorithms have to slice the list from the first element onwards to avoid the group workspace if they want to work a workspace at a time such as Pearl grouping 3 workspaces at a time for the output. 

Edit: It also allows us to wind the value right down from the scientists feedback as loading is one of the most expensive operations in this script.

scripts/Diffraction/isis_powder/routines/common.py
DavidFair: Will it handle comma separated ranges e.g. `1-10,12,14-20` loads: `1,2,3,4....8,9,10,12,14,15,16....,18,19,20`? 
I'll add it to the refactoring [here](https://github.com/mantidproject/mantid/issues/18108) to experiment around with it

DavidFair:
Additional suggestions implemented in this [commit](https://github.com/mantidproject/mantid/commit/7ee1199ced24ea78a82f2d081338fec1b37a2073)

scripts/Diffraction/isis_powder/routines/common.py
soininen: Running
```python
mantid.kernel.IntArrayProperty('dummy_property', '1-10,12,14-20').value
```
gives
```
array([ 1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 12, 14, 15, 16, 17, 18, 19,
       20]
```
I think this is what you want here, no?

soininen (approved):

soininen:
Since this is work-in-progress, I only reviewed the code. As issues found here are being addressed in a different branch, I would say :shipit:

PR #18099, opened 2016-11-23
Add eFixed methods to DetectorInfo and SpectrumInfo

### LINKED ISSUE #18081: Add getEiFixed method to SpectrumInfo and DetectorInfo
A common pattern for Indirect instruments is to get `EFixed` from the parameter map for the detector, see [for example](https://github.com/mantidproject/mantid/pull/18034/files/e505cc14185f98769faf1bd44dd8a46c51a69c41#diff-86ae7fca41c2b1046be4eb298a05a517R206).

By adding methods to `SpectrumInfo` and `DetectorInfo` to get these values we can remove a large chunk of duplicated code when refactoring.

ianbush:
`DetectorInfo` and `SpectrumInfo` now have their own `eFixed` methods for getting the value from detectors.

`ExperimentInfo` has been updated to include this. When performing the refactoring for #17743 then either the methods in `ExperimentInfo`, `DetectorInfo` or `SpectrumInfo` should be used for getting the eFixed parameter.

**To test:**

Code review. Are the methods added here adequate for the refactoring, and does it make sense for `SpectrumInfo` to have the eFixed method?

`eFixed` returns 0 when the parameter does not exist from `DetectorInfo`, and averages any detectors with eFixed in `SpectrumInfo`, or returns 0. Does this make sense, or should the eFixed methods return a bool to say if an eFixed exists?

Fixes #18081.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/src/DetectorInfo.cpp
+++ b/Framework/API/src/DetectorInfo.cpp
@@ -95,7 +95,8 @@ Kernel::Quat DetectorInfo::rotation(const size_t index) const {
 
 double DetectorInfo::eFixed(const size_t index) const {
   const auto &detector = getDetector(index);
-  auto eFixedVector = detector.getNumberParameter("Efixed");
+  auto eFixedVector =
+      detector.getNumberParameter(Geometry::ParameterMap::eFixed());
 
   double eFixed = 0.0;
 
--- a/Framework/API/src/DetectorInfo.cpp
+++ b/Framework/API/src/DetectorInfo.cpp
@@ -93,6 +93,19 @@ Kernel::Quat DetectorInfo::rotation(const size_t index) const {
   return getDetector(index).getRotation();
 }
 
+double DetectorInfo::eFixed(const size_t index) const {
+  const auto &detector = getDetector(index);
+  auto eFixedVector = detector.getNumberParameter("Efixed");
+
+  double eFixed = 0.0;
+
+  if (!eFixedVector.empty()) {
+    eFixed = eFixedVector.front();
+  }
+
+  return eFixed;
+}
+
 /// Set the absolute position of the detector with given index.
 void DetectorInfo::setPosition(const size_t index,
                                const Kernel::V3D &position) {
--- a/Framework/API/src/SpectrumInfo.cpp
+++ b/Framework/API/src/SpectrumInfo.cpp
@@ -113,7 +113,6 @@ double SpectrumInfo::eFixed(const size_t index) const {
   return eFixed / static_cast<double>(dets.size());
 }
 
-
 /// Returns true if the spectrum is associated with detectors in the instrument.
 bool SpectrumInfo::hasDetectors(const size_t index) const {
   // Workspaces can contain invalid detector IDs. Those IDs will be silently
--- a/Framework/API/src/SpectrumInfo.cpp
+++ b/Framework/API/src/SpectrumInfo.cpp
@@ -101,6 +101,19 @@ Kernel::V3D SpectrumInfo::position(const size_t index) const {
   return newPos / static_cast<double>(dets.size());
 }
 
+/// Returns the eFixed value of the spectrum with given index.
+double SpectrumInfo::eFixed(const size_t index) const {
+  double eFixed(0);
+  const auto &dets = getDetectorVector(index);
+  for (const auto &det : dets) {
+    const auto &detIndex = m_detectorInfo.indexOf(det->getID());
+    m_detectorInfo.setCachedDetector(detIndex, det);
+    eFixed += m_detectorInfo.eFixed(detIndex);
+  }
+  return eFixed / static_cast<double>(dets.size());
+}
+
+
 /// Returns true if the spectrum is associated with detectors in the instrument.
 bool SpectrumInfo::hasDetectors(const size_t index) const {
   // Workspaces can contain invalid detector IDs. Those IDs will be silently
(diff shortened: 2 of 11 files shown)


ianbush:
@martyngigg Could you comment on whether this looks like a good way to deal with eFixed within `DetectorInfo` and `SpectrumInfo`? Does this seem to match how it might be used?

Framework/API/src/DetectorInfo.cpp
```diff
@@ -93,6 +93,20 @@ Kernel::Quat DetectorInfo::rotation(const size_t index) const {
...
+  const auto &detector = getDetector(index);
+  auto eFixedVector =
+      detector.getNumberParameter(Geometry::ParameterMap::eFixed());
+
+  double eFixed = 0.0;
+
+  if (!eFixedVector.empty()) {
+    eFixed = eFixedVector.front();
+  }
+
```
martyngigg: I think we should throw if "efixed" doesn't exist. I would say that the client should be responsible for what happens if it doesn't exist.

Framework/API/src/SpectrumInfo.cpp
```diff
@@ -101,6 +101,27 @@ Kernel::V3D SpectrumInfo::position(const size_t index) const {
...
+  double eFixed(0);
+  size_t validEFixedSize(0);
+  const auto &dets = getDetectorVector(index);
+
+  for (const auto &det : dets) {
+    const auto &detIndex = m_detectorInfo.indexOf(det->getID());
+    m_detectorInfo.setCachedDetector(detIndex, det);
+    double detectorEFixed = m_detectorInfo.eFixed(detIndex);
+
+    if (detectorEFixed != 0.0) {
```
martyngigg: Can we just use the same averaging logic as above if `m_detectorInfo.eFixed(detIndex);` is going to throw?

I think it would best again to avoid an ambiguous scenario where a detector with an `efixed` has been grouped with one that doesn't have one. It's not clear what to do so I'm tempted to say that the client should deal with that.

Framework/API/src/DetectorInfo.cpp
OwenArnold: We could have the double `efixed` value as an out parameter and use the return to provide an Enum with information allowing the client to decide what to do.

Framework/API/src/SpectrumInfo.cpp
OwenArnold: I agree with @martyngigg  on this.

OwenArnold: Though I'd rather see the use of an enum (see comment above) over a plethora of newly defined exception types.

Framework/API/src/DetectorInfo.cpp
ianbush: I will go down the enum route with this. The possible cases are:
* Single detector with single eFixed
* Single detector with no eFixed value
* Mulitple detectors each with an eFixed values (return average)
* Multiple detectors with no eFixed values
* Multiple detectors with some eFixed values, some missing eFixed values

There is also the case of all/some detectors not being valid. These should be dealt with in the usual way, with a call to `hasDetectors()` followed by the try/catch used [here](https://github.com/mantidproject/mantid/pull/18034/files#diff-207c349b1b7e9189101af6b77d579965L179).

Not wishing for too many options I would suggest `SUCCESS`, `NO_EFIXED` and `EFIXED_NOT_FOUND_FOR_ALL`.

SimonHeybrock: How do we deal with changing behavior? Currently if there are multiple detectors where each has eFixed we are not using the average, I think. Instead it mostly defaults to 0.0?

SimonHeybrock: I would be a bit unhappy about the return type for error reporting. Return values tend to be ignored...

SimonHeybrock:
I feel a bit uneasy about these changes, and I think you saw this manifest itself in the difficulty finding a good function signature for, e.g., `SpectrumInfo::eFixed` (exceptions? out-parameters? return flags?).

The root of this is basically that only a certain type of instrument has this parameter. Furthermore, there are two ways to obtain eFixed (direct vs. indirect), and ultimately we should clean that up, to avoid mess in our algorithms. Apart from that there is also an interface issue in many of our algorithms, where the user can specify the eMode, ignoring completely what the instrument actually is.

Does it make sense to have a method for this optional parameter in the general `SpectrumInfo` class? The same could be asked for other optional parameters that are stored in the parameter map.

I think I would rather keep `SpectrumInfo` generic, and instead have a class that can provide eFixed, say `EFixedInfo`. This class could still use `SpectrumInfo` internally to obtain information from the parameter map, but it can deal in a much nicer way with validation. For example, the energy mode of the instrument can be checked at construction time of `EFixedInfo`, and constructor parameter could define behavior on errors.

Framework/API/src/DetectorInfo.cpp
OwenArnold: @SimonHeybrock. These errors could be ignored. However, as @martyngigg pointed out its up to the client to decide if the absence of one, or more, or these parameters is really an error or not. I would say that given that the client of this code is an Algorithm author, they should be responsible for testing these scenarios anyway at the algorithm level.

OwenArnold:
As @SimonHeybrock pointed out. I think that this somewhat more complex that I initially foresaw. It would be nice to group these `eFixed` somewhere shared away from the specific `Algorithm` but we might have to compromise on that.

ianbush:
For the reasons @SimonHeybrock mentioned I will close this PR for now, as it will take some further work.

Issue #18091, opened 2016-11-22
TOF axis problems with ILL's time-of-flight spectrometers


soininen:
The TOF spectrometers IN4, IN5 and IN6 at ILL produce output files with arbitrary starting point for the time-of-flight axis. At the moment, the TOF axis is adjusted in the `LoadILLTOF` loader to correspond the elastic L1 + L2 flight time. There are some problems with this approach, however. For the calculation, the 'wavelength' sample log is used. This value can vary between datasets which otherwise have the same nominal incident energy and result in incompatible TOF axes, which can give varying blocksizes when workspaces are merged and prevent basic math between them. Further, for IN4 and IN6, the incident energy can be calibrated later using the `GetEiMonDet` algorithm, which leads to elastic peaks not being at the zero-energy transfer when the TOF axis is converted to DeltaE.

PR #18118, opened 2016-11-25
New algorithm CorrectTOFAxis

soininen:
This PR fixes partially the issue #18091 by introducing a new algorithm `CorrectTOFAxis` which can be used to adjust the TOF axis of a workspace so that conversion from 'TOF' to 'DeltaE' will result in the elastic peak being at the zero energy transfer. The complete fix will require the removal of similar adjustments in [`LoadILLTOF`](http://docs.mantidproject.org/nightly/algorithms/LoadILLTOF-v1.html), and the usage of this algorithm together with [`GetEiMonDet`](http://docs.mantidproject.org/nightly/algorithms/GetEiMonDet-v2.html) to get the correct TOF axis for ILL workspaces. `CorrectTOFAxis` in itself is instrument/institution independent, however.

The unit tests contain some functionality common to `GetEiMonDet2Test`. This functionality was moved to WorkspaceCreationHelpers and `GetEiMonDet2Test` was refactored accordingly.

**To test:**

The script below exemplifies the algorithm's three modes of operation. Using all the modes should give the same results: shift the TOF axis slightly and update the 'Ei' and 'wavelength' sample logs. The converted workspaces should have the elastic peaks at the zero energy transfer (although in this case the raw workspace comes pretty close as well).

```python
# Adjust the input path to the Mantid build directory
inputFile = '<mantid-build-directory>/ExternalData/Testing/Data/UnitTest/ILL/IN4/084446.nxs'
raw = Load(inputFile)
# Calibrate incident energy.
EPP = FindEPP(raw)
newEi = GetEiMonDet(raw, EPP, '1-299', 0, PulseInterval=3000)
# Correct the TOF axis directly... (updates the 'Ei' and 'wavelength' sample logs)
correctedWithEPPTable = CorrectTOFAxis(raw, EPPTable=EPP, ReferenceSpectra='1-299', EFixed=newEi)
# ...by using elastic bin index...
elasticChannel = raw.run().getProperty('monitor.elasticpeak').value
correctedWithBinIndex = CorrectTOFAxis(raw, ReferenceSpectra='1-299', ElasticBinIndex=elasticChannel, EFixed=newEi)
# ... or use a reference workspace. (also updates the sample logs)
referencedRaw = CorrectTOFAxis(raw, ReferenceWorkspace=correctedWithBinIndex)
# Convert the units
convertedCorrected = ConvertUnits(correctedWithEPPTable, Target='DeltaE', EMode='Direct')
convertedRaw= ConvertUnits(raw, Target='DeltaE', EMode='Direct')
```

*The release notes were updated accordingly.*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


docs/source/algorithms/TOFAxisCorrection-v1.rst
```diff
@@ -0,0 +1,80 @@
...
+
+If *ReferenceWorkspace* is set, this algorithm copies the X axis as well as the 'Ei' and 'wavelength' sample logs to the *OutputWorkspace*. The rest of the input properties are discarded.
+
+If no *ReferenceWorkspace* is given, the algorithm calculates the average L2 distance and the average elastic peak positions. The L2 values it gets from the *InputWorkspace* while the elastic peak positions are taken from *EPPTable*. *EPPTable* should be in the format returned by the :ref:`algm-FindEPP` algorithm. The spectra over which the averages are taken should be listed in *ReferenceSpectra*. Whether this input property refers to workspace indices, spectrum numbers or detector IDs is specified by *IndexType*. Incident energy can be either specified by *IncidentEnergy* or it is taken from the sample logs of *InputWorkspace*. In case *IncidentEnergy* is specified, the 'Ei' and 'wavelength' sample logs of *OutputWorkspace* are updated accordingly.
+
```
AndreiSavici: Still not clear how you modify TOF. Put a formula!

soininen: Will do. The documentation needs an update, anyway.

martyngigg (commented):
Can I suggest changing the name to `CorrectTOFAxis`?

It puts the doing part first like others that start `Sum`, `Calculate` etc.

soininen:
@martyngigg I don't have any strong arguments against `CorrectTOFAxis`. It was actually in the initial list of possible names. We ended up with the current name to get this algorithm in an easier-to-spot place in the alphabetic list of the html documentation and to prevent it from vanishing in the midsts of the other Correct* algorithms. Anyway, as you say, `CorrectTOFAxis` is more in line with what is already in Mantid so I'll rename it.

martyngigg:
@soininen Thanks. We're definitely aware of the current issue of finding algorithms in both the MantidPlot search and the documentation and I believe there should be some resource to try and tackle both next year.

Framework/Algorithms/src/CorrectTOFAxis.cpp
```diff
@@ -0,0 +1,503 @@
...
+/// Tag for workspace indices
+const static std::string WORKSPACE_INDEX("Workspace Index");
+} // namespace IndexTypes
+
+/** A private namespace listing the properties of CorrectTOFAxis.
+ */
+namespace PropertyNames {
+const static std::string ELASTIC_BIN_INDEX("ElasticBinIndex");
+const static std::string EPP_TABLE("EPPTable");
+const static std::string INCIDENT_ENERGY("IncidentEnergy");
```
martyngigg: Would it make sense to change the name of this property to `EFixed`? We have a few algorithms that can accept an E_i/E_f and we have generally settled on the property name being `EFixed` in these cases.

soininen: The algorithm was meant for direct geometry TOF spectrometers, thus the choice of name here. But if calling Ei *EFixed* is a Mantid convention, then I guess I have to comply. I only wish I had realized this earlier as I have used *IncidentEnergy* in a few algorithms before...

soininen: At least the two indirect instruments I have worked with (the backscattering spectrometers IN16b at ILL and SPHERES at MLZ) don't have any notion of TOF and thus there is no need to deal with the TOF axis at all. I don't know about other indirect instruments, though. The use cases I have planned for this algorithm include only direct geometry instruments but of course I cannot predict how other people might use this.

Anyway, for me this boils down to if I would like to be consistent with some algorithms I have written earlier. In hindsight I think should have used *EFixed* in them as well. I guess this is as good point as any to start using *EFixed*, so I'll make the change.

martyngigg:
Is it expected that the conversion via bin index and via epp table give the same result?

I first ran your script and then converted `correctedWithBinIndex` to `DeltaE`. I then ran `Rebin` with `Params=2` on `convertedRaw`, `convertedCorrected` and `convertedWithBinIndex` and then `SumSpectra` on each of those to get the picture below:

correcttofaxis

It looks as if the bin index correction is 1 bin out from the epp table. Is that expected?

soininen:
I'm not surprised if there is a bin or two difference between the epp method and giving a fixed bin index. In the test script, the epp table is build using the [FindEPP](http://docs.mantidproject.org/nightly/algorithms/FindEPP-v1.html) algorithm, which fits a Gaussian to all histograms. The average peak centre can then differ by a few bins from the 'monitor.elasticpeak' value in the sample logs. The elastic bin index in the sample logs is just something written by the instrument control software. Where this number comes from or how accurate it is, I don't know.

martyngigg:
@soininen Thanks for clarifying that. My main concern was that there might be an off-by-one error somewhere but I hadn't realised the index came from the control software.

Happy to :shipit:

Framework/Algorithms/CMakeLists.txt
```diff
@@ -66,6 +66,7 @@ set ( SRC_FILES
 	src/CorelliCrossCorrelate.cpp
 	src/CorrectFlightPaths.cpp
 	src/CorrectKiKf.cpp
+        src/CorrectTOFAxis.cpp
```
ianbush: These should be tabs instead of spaces to be consistent with the rest of the file.

soininen: I should know to never to edit these files in QtCreator.

Issue #18122, opened 2016-11-25
PulseInterval could be read from the sample logs in GetEiMonDet


soininen:
The ILL loader adds nowadays a 'pulse_interval' sample log for IN4 and IN6 spectrometers. This could be directly used in the [`GetEiMonDet`](http://docs.mantidproject.org/nightly/algorithms/GetEiMonDet-v2.html) algorithm.

PR #18126, opened 2016-11-28
GetEiMonDet interface improvements

### LINKED ISSUE #18121: IndexType options in GetEiMonDet should have a space between the words
The choices in *IndexType* are in CamelCase, but usually these types of choices are written with spaces between the words.

soininen:
This pull request improves the GetEiMonDet algorithm in two ways: the pulse interval can be read from the sample logs, if needed, and the option strings in the *IndexType* property have spaces instead of CamelCase, bringing them closer to other Mantid algorithms.

**To test:**

Code review. The pulse interval sample log part should be covered by unit tests while the new *IndexType* options should work as-is.

Fixes #18121 #18122 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/src/GetEiMonDet2.cpp
+++ b/Framework/Algorithms/src/GetEiMonDet2.cpp
@@ -21,6 +21,7 @@ using namespace Mantid::PhysicalConstants;
 namespace Mantid {
 namespace Algorithms {
 
+namespace {
 /** A private namespace to store string constants dealing with
  *  tables returned by the FindEPP algorithm.
  */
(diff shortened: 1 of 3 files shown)


Framework/Algorithms/src/GetEiMonDet2.cpp
```diff
@@ -38,11 +39,11 @@ const static std::string FIT_STATUS_SUCCESS("success");
  */
 namespace IndexTypes {
 /// Tag for detector ids
-const static std::string DETECTOR_ID("DetectorID");
+const static std::string DETECTOR_ID("Detector ID");
```
martyngigg: Is there a danger of breaking user scripts here that would have used `DetectorID` with no space? From the usage stats I see that this has only ever been used as a child algorithm but I can't find the usage anywhere in the code base.

soininen: This version of GetEiMonDet was added to Mantid quite recently, and the only user at the moment would be me while I'm developing the `DirectILLReduction` algorithm. So, I think it is still OK to make such API breaking changes. The last chance, actually.

martyngigg:
Unit tests look good. I'm happy the slight change to allowed values won't break anything given how new the  algorithm is and the usage reports everything as being internal.

:shipit:

PR #18128, opened 2016-11-28
Fix special values propagating in SumSpectra 

### LINKED ISSUE #18123: Engg: Calibration algorithm fails to fit run
Vanadium Run: 259488
Cerium Run: 264092

It appears that the peaks centres are NaNs indicating something is wrong before the fitting stage.

The algorithm to perform ENGINX calibrations fails with the following trace:
```
Error in execution of algorithm EnggFitPeaks:
Could not fit any peak.  Failed to fit peaks with peak type BackToBackExponential even though FindPeaks found 28 peaks in principle. See the logs for further details.
  at line 89 in 'C:/Users/ieb35538/Documents/Repos/master_mantid/Framework/PythonInterface/plugins/algorithms\EnggFitPeaks.py'
  caused by line 206 in 'C:/Users/ieb35538/Documents/Repos/master_mantid/Framework/PythonInterface/plugins/algorithms\EnggFitPeaks.py'
Error in execution of algorithm EnggCalibrate
Could not fit any peak.  Failed to fit peaks with peak type BackToBackExponential even though FindPeaks found 28 peaks in principle. See the logs for further details.
  at line 89 in 'C:/Users/ieb35538/Documents/Repos/master_mantid/Framework/PythonInterface/plugins/algorithms\EnggFitPeaks.py'
  caused by line 206 in 'C:/Users/ieb35538/Documents/Repos/master_mantid/Framework/PythonInterface/plugins/algorithms\EnggFitPeaks.py'
  at line 146 in 'C:/Users/ieb35538/Documents/Repos/master_mantid/Framework/PythonInterface/plugins/algorithms\EnggCalibrate.py'
  caused by line 176 in 'C:/Users/ieb35538/Documents/Repos/master_mantid/Framework/PythonInterface/plugins/algorithms\EnggCalibrate.py'
Error in calibration. The calibration calculations failed. One of the algorithms did not execute correctly. See log messages for further details. Error: EnggCalibrate failed
```

 Additionally `SplineBackgrounds` uses GSL instead of Mantid functions - see what happens if we use `BSpline` fitting instead

DavidFair:
Description of work.
This PR fixes an issue with SumSpectra where any spectra with "special" values such as NaN would cause the final single spectra to have a `NaN` value(s). This is most apparent when summing masked out detectors whose values have been set to `NaN`. 
We now call replace special values before processing the workspace to replace said values with 0. This was already done within `rebinnedOutput` so the step was moved into its own method and called for 2D workspaces. 

Additionally this PR contains another fix to stop the GUI attempting to parse a path to an existing calibration on startup. Previously a warning would appear when opening the GUI completely unrelated to the GUI start-up.

**To test:**

Download the following .nxs file or create a workspace with random `nan` values in the data:
[18123_Engg_masked_ws.zip](https://github.com/mantidproject/mantid/files/617056/18123_Engg_masked_ws.zip)
Run SumSpectra 
Check the data - any `nan` values should contribute 0 and the final workspace should not contain any special floating point values 

Fixes #18123 .

[Release Notes](https://github.com/mantidproject/mantid/pull/18128/files#diff-f7793b747fb7ac7103dcdcb26c5d7ca5)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


DimitarTasev:
Running a clean rebuild on http://builds.mantidproject.org/job/pull_requests-rhel7/10545/

Framework/Algorithms/src/SumSpectra.cpp
```diff
@@ -222,69 +240,73 @@ SumSpectra::getOutputSpecNo(MatrixWorkspace_const_sptr localworkspace) {
...
-        if (YErrors[k] != 0) {
-          double errsq = YErrors[k] * YErrors[k];
-          YError[k] += errsq;
-          Weight[k] += 1. / errsq;
-          YSum[k] += YValues[k] / errsq;
+
+    for (size_t i = 0; i < YValues.size(); ++i) {
+      if (m_calculateWeightedSum) {
+        if (std::isnormal(YErrors[i])) {
+          double errsq = YErrors[i] * YErrors[i];
```
DimitarTasev: const

Framework/Algorithms/src/SumSpectra.cpp
```diff
@@ -222,69 +240,73 @@ SumSpectra::getOutputSpecNo(MatrixWorkspace_const_sptr localworkspace) {
...
-    const auto &YErrors = localworkspace->e(i);
-    if (m_calculateWeightedSum) {
-      for (int k = 0; k < this->m_yLength; ++k) {
-        if (YErrors[k] != 0) {
-          double errsq = YErrors[k] * YErrors[k];
-          YError[k] += errsq;
-          Weight[k] += 1. / errsq;
-          YSum[k] += YValues[k] / errsq;
+
+    for (size_t i = 0; i < YValues.size(); ++i) {
```
DimitarTasev: Why change from `this->m_yLength`? The `this->m_yLength` is used everywhere else, so I think it might be confusing as to why this single instance is `YValues.size()`, even if they have the same size

docs/source/release/v3.9.0/framework.rst
```diff
@@ -55,6 +55,7 @@ Bug Fixes
 - :ref:`LoadEventNexus <algm-LoadEventNexus>` should no longer leak memory when the execution is cancelled.
 - :ref:`LoadNexusProcessed <algm-LoadNexusProcessed>` will now load the stored workspace names from a processed Nexus file in the case of multiperiod data.
 - If a run is aborted and restarted, the ``running`` log in the workspace will correctly reflect this. (``running`` will be false at all times before the abort.)
+- :ref:`SumSpectra <algm-SumSpectra>` should correctly handle special values in the data such as NaN which previously ended up in the summed spectra.
```
DimitarTasev: Are we sure setting it to 0 is the correct solution? Could it just be a sign that there is a deeper problem with the algorithms using it , that this fix would mask away?

dtasev (changes requested):

DimitarTasev:
Results with PR:
image

The additional option `Remove special values` works correctly. You can see the difference between the workspaces on the right. The top one is without the option to remove special values, and the sum for each bin is now a `nan`. The bottom one is with the option, and all the `nan`'s have been masked away, so there is still data in the output workspace.

The option is disabled by default so it shouldn't cause any compatibility issues with older scripts.

2 unit tests have been added for the class, that test this new functionality.

Changes to Engineering Diffraction interface are working as well.

OSX Build failure (as of 13/12/2016) is unrelated.

:shipit:

dtasev (approved):

PR #18131, opened 2016-11-29
ILL :: Indirect Reduction Framework

### LINKED ISSUE #14179: ILL Energy Transfer option for sum files
Currently ILL Energy Transfer can only handle a single run at a time rather than a range of runs. Additionally it should be possible to select multiple files from the file browser (shift+Leftclick) / (Ctrl+Leftclick)

### LINKED ISSUE #14649: Progress Tracking for ILL Energy Transfer and ILL Calibration
The new (modified) ILL IN16B algorithms `ILLIN16BCalibration.py` and `IndirectILLReduction.py` both need progress tracking.

### LINKED ISSUE #14650: ILL Energy Transfer change to match naming conventions in ISIS Energy Transfer
ILL Energy Transfer needs to create workspaces that match the naming convention in ISIS Energy Transfer to be consistent within Mantid

The current convention for ISIS Energy Transfer is `instrument_analyser_reflection_red`.nxs

### LINKED ISSUE #14653: ILL Energy Transfer and calibration GUI changes
ILL Energy Transfer and ILL Calibration need updates to the GUI. These updates include:
- [ ] Refactoring Save and Plot functionality to interface code from algorithm
- [ ] Remove Mirror Mode option in GUI

### LINKED ISSUE #16589: Resolving problems with ILLIN16b calibration
### Expected behavior

Calibration of workspaces.
### Actual behavior
- Calibration fails completely or produces NAN
- Two different but very similar algorithms are implemented
### Details
- IndirectCalibration.py calls Load() instead of LoadILLIndirect.py, MirrorMode is missing, MergeRuns called
- ILLIN16BCalibration.py MirrorMode default is False required to be True
- Reimplementation of functions in both scripts, e.g. ScaleFactor, ...
- GUI IndirectDataReduction calls IndirectCalibration.py in ILL Energy Transfer
- GUI IndirectDataReduction calls ILLIN16BCalibration.py in ILL Calibration
### Platforms affected

All

### LINKED ISSUE #16753: Indirect ILL Reduction Refactoring
IndirectILLReduction workflow algorithm needs to be upgraded to accommodate new requirements from ILL IN16B.

IN16B can record the data in two modes: with mirror sense, when the acceleration and deceleration phases of the doppler drive are recorded separately, or without. The algorithms should be able to automatically treat the two.

Moreover, 2 different reduction types needed:
- QENS : where the full spectrum is preserved
- FWS : where the spectrum is integrated over energy transfer, and the integrated intensity is scanned as a function of some metadata (typically sample temperature). In turn, FWS can be run at many different doppler energies, including 0, which is referred to as EFWS (elastic) or non-zero (IFWS) (inelastic).

Both of the reductions must implement background subtraction and vanadium calibration in a specific way. 

In case of QENS with mirror mode, the doppler's accelerations phase (left wing) and deceleration phase (right wing) data might be misaligned in terms of energy channel due to DAQ electronics/software.
Hence, the QENS reduction should implement 8 unmirror options (consistent with the existing software) to mitigate the misalignment aposteriori.

All the reductions must be able to run over multiple files (up to 500 for FWS) and pre filter the files according to the reduction type before loading.

The loader should check which slots of single detectors are enabled, and load only those, while moving their positions to the values read from nexus file.

### LINKED ISSUE #17559: New WorkflowAlgorithm IndirectILLFixedWindowScans for IN16B
### Expected behavior

The aim is to implement a workflow algorithm for fixed window scans of the ILL IN16B instrument. Currently, only the workflow algorithm for quasi elastic scans QENS (https://github.com/mantidproject/mantid/issues/16753) is implemented. Inelastic and elastic fixed-window scans have Nexus file entries, where the velocity is 1 and energy is 0 (inelastic) or unequal zero (elastic).

Previous work is done by Spencer Howells for ILLElasticScans.

Make use of https://github.com/mantidproject/documents/blob/master/Project-Management/ILL/WorkspaceAlgorithmRunner.md and SelectNexusFilesByMetadata (https://github.com/mantidproject/mantid/pull/17361)
### Platforms affected

All

gvardany:
This PR implements the reduction framework for IN16B indirect geometry instrument at ILL.
It includes a new version for the loader (which treats single detectors slightly different), 3 new workflow algorithms (deprecating the existing 2) and corresponding upgrade of the GUI. 

The results of the reduction are preliminarily validated against the existing software at ILL (LAMP), and science validation is being run in parallel.

It turned out to be a rather large ticket, but the different parts are in tight relation and have been developed together, so we were not able to break it down to smaller pieces.

**To test:**

1. Use this example data (see the READMEs inside)
[FWS.zip](https://github.com/mantidproject/mantid/files/654308/FWS.zip)
[QENS-1wing.zip](https://github.com/mantidproject/mantid/files/654306/QENS-1wing.zip)
[QENS-2wing.zip](https://github.com/mantidproject/mantid/files/654307/QENS-2wing.zip)
2. Read the documentations.
2. Set ILL/IN16B as facility/instrument.
3. Go to Interfaces -> Indirect -> Data Reduction -> Energy Transfer tab.
4. Run with different options/combinations. Try also to break it.
5. Code review.

Fixes #16753 .
Fixes #17559 .
Fixes #16589 .
Fixes #14179 .
Fixes #14649 .
Fixes #14650 .
Fixes #14653 .

Some of these issues were opened historically, and are no longer relevant.

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


louisemccann:
Initially I have found a couple of issues:
1. In [ILLEnergyTransfer](https://github.com/mantidproject/mantid/blob/81a3680735bebd349bd0204b77e684f689832a10/MantidQt/CustomInterfaces/src/Indirect/ILLEnergyTransfer.cpp) there are some variables named `ok`, `ok1` etc, these could be more descriptive
2. If run numbers are put is as a range, eg `136558-136564` the runs are summed whether the option is ticked or not. This applies to the input, background and calibration fields. This is not the case when the runs are entered in a comma separated list
3. A bit picky but if multiple runs are summed the name of the resulting workspace only contains the first run in the range eg `136558_sample_red`

gvardany:
1. I will try to rename those, they are just temporary flags to check if the casting to double succeeded, their scope is basically a couple of lines, so they seemed straightforward from the first glance.

2. We follow the conventions in http://www.mantidproject.org/MultiFileLoading , so `-` is the sign for the summed range. If summing is not desired one should use `,` or `:` for unsummed range. The checkbox is in the GUI just to facilitate the selection of many files while holding the `Shift` button. In this case they appear as `,` separated list, so the checkbox only serves the purpose to automatically transfer this to `+` separated list behind the scenes. Otherwise, one should be able to sum or not sum without this checkbox just following the above mentioned rules. Regarding the calibration and background files, they always need to be summed if multiple files are specified, no matter the range syntax.

3. When summing many files (order of 100) it is impractical to keep all the run numbers in the workspace name. Here we wanted to be coherent with what the `Load` does with the run number, when multiple files are loaded/summed. I.e. we take the run number from the first run in the list, but we anyway preserve the full list of runs in a sample log entry.

SpencerHowells:
Regarding item 3 above. For  a range of runs we create a name <first_run>to <last run>. This will avoid any confusion in names between a single run & multiple runs.

SpencerHowells:
The last comment should have read - name beginning <first_run>to<last_run>.
Or it could be beginning <first_run>_multiple.

SpencerHowells:
During testing QENS we used the water sample and container background with Unmirror option 6. It crashed in FindEPP. We then looked at the individual sample & container and found that after subtraction the data was quite flat with very little peak. We then used vesicle sample and different Unmirror option. This produced a decent result and ConvFit & BayesQuasi gave good fits.
Checking on the FinDEPP algo I notice that it fits a Gaussian. The data has a Lorentzian shape. Having a peak shape option in FindEPP and using Lorentzian might give better more reliable results.
The above problem also shows the danger of performing a background subtraction at this point- you cannot check what is happening. I do not like it & we do not do it.

gvardany:
Regarding your second point, the problem is actually known, it is in `MatchPeaks` that calls `FindEPP`. There is some race condition when deleting tables generated by `FindEPP` with hard-wired names, so when it is called many times in a parallel loop, it confuses the ADS. This is fixed in PR #18171 , it is just not merged yet. Indeed, the peaks are Lorentzian, but at this point we need just to identify the bin index of the peak, so Gaussian is sufficient. We wanted to use what is already available, `FindEPP`, with minimum intervention there. For the background subtraction, normally they would not do it on-the-fly, but do in `Indirect->Corrections`, but they still wanted the possibility.

SpencerHowells:
I mention the confusion in file names because we have had this experience with users. Your scientists may be familiar with it but visiting users may and they will be running the program at their home university. Do not underestimate the stupid things users will do!
We introduced using <first_run>_multiple to avoid this.
I agree with your comments about FindEPP and background subtraction.
In practice the real background ( nothing in beam) is insignificant & we don't do a subtraction.

gvardany:
Ok, I will ask the users about the adding the `_multiple`.

gvardany:
I have made the requested changes, any chance we can push this for Release 3.9?

louisemccann:
Input provided works as expected - any bad input is picked up and error messages are appropriate
LGTM :shipit:

PR #18137, opened 2016-11-30
Remove global peak radius setting.

### LINKED ISSUE #6391: Make the peak radius an attribute of the function.
This issue was originally [TRAC 5545](http://trac.mantidproject.org/mantid/ticket/5545)

The PeakFunction has a default and quite small 'peak radius' - a cut-off  distance for the peak wings. It is a good thing for algorithms such as FindPeaks but can be a problem for a user who is unfamiliar with it. 

There should not be a global radius. It should be an attribute with different defaults for each peak type.

mantid-roman:
Rather than an attribute the `PeakRadius` is now a property of `Fit` (and some other algorithms) which I think is a much better option.

**To test:**

Fit some peaks (long tailed such as Lorentzians are preferred) with different values of `PeakRadius`. Plot the results, see how the property affects the cut-off region.

Fixes #6391.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


martyngigg (commented):
Should we remove `curvefitting.peakRadius` from https://github.com/mantidproject/mantid/blob/master/Framework/Properties/Mantid.properties.template ?

Framework/API/src/IPeakFunction.cpp
```diff
@@ -78,21 +80,28 @@ class TempJacobian : public Jacobian {
...
-  if (Kernel::ConfigService::Instance().getValue("curvefitting.peakRadius",
-                                                 peakRadius)) {
-    if (peakRadius != s_peakRadius) {
-      setPeakRadius(peakRadius);
-    }
+IPeakFunction::IPeakFunction(int peakRadius) : m_peakRadius(peakRadius) {}
+
+void IPeakFunction::function(const FunctionDomain &domain,
+                             FunctionValues &values) const {
+  auto peakRadius =
```
martyngigg: Why do we construct the `IPeakFunction` with a radius and then override it with the radius from the domain?

mantid-roman: The radius is overridden only when it's smaller than the default.

martyngigg: Could this cause confusion though if a user is trying to set it from `Fit` and then in some circumstances their value is not used?

mantid-roman: I think you are right. I'll make some changes.

Framework/CurveFitting/src/Algorithms/Fit.cpp
```diff
@@ -6,6 +6,7 @@
 
 #include "MantidAPI/CostFunctionFactory.h"
 #include "MantidAPI/FuncMinimizerFactory.h"
+#include "MantidAPI/FunctionDomain1D.h"
```
martyngigg: Is this new include required?

Framework/CurveFitting/src/Algorithms/Fit.cpp
```diff
@@ -134,32 +135,16 @@ void Fit::execConcrete() {
     m_function->addConstraints(contstraints);
   }
 
-  // prepare the function for a fit
-  m_function->setUpForFit();
```
mantid-roman: It was moved to the base class `IFittingAlgorithm`.

Framework/CurveFitting/src/Algorithms/Fit.cpp
mantid-roman: No. I forgot to remove it after I moved the peak radius stuff to the base class.

martyngigg: Okay. If we remove this then I'm happy this can go through.

martyngigg:
I've tried some fitting with a Voigt on some Vesuvio lead data. The peak radius settings seems to respond as expected and the fitting is quicker if I give it a positive, non-zero peak radius.

I think this is probably something not worth doing in this issue but I notice that setting a negative peak radius is allowed. I think users would be surprised at this and we should restrict to `>= 0`.

martyngigg:
I'm happy with this now.

:shipit:

SimonHeybrock:
@mantid-roman This branch has developed a merge conflict.

PR #18138, opened 2016-11-30
SplineInterpolation: generalisation of interpolation and functions LinearBackground, CubicSpline)

### LINKED ISSUE #18097: SplineInterpolation :: do linear interpolation when only two points are given
`SplineInterpolation` rightly throws an error when trying to interpolate 2 points with cubic spline. Generalisation would be relevant to do linear interpolation for the case of 2 points with a warning message instead of error.

vreimund:
Algorithm `SplineInterpolation`
- added option `Linear2Points` for performing a linear interpolation if `WorkspaceToInterpolate` has two points
- added validator for `WorkspaceToInterpolate` and `OutputWorkspaceDeriv`
- new CurveFitting Function `Linear` (similarity with fitting function `LinearBackground`, should be modified in the future), is a non-fitting function but declares fitting parameters like its corresponding CurveFitting Function `CubicSpline`
- modifications in `CubicSpline`, and its unit test (uses smart pointers now)
- fixed bug : MantidPlot manual test: right click on output workspace -> plot spectrum -> spectrum number can now be entered (was constant 1 before)
- I moved the check for ascending x values to interpolation and deleted it from `CubicSpline` since needed also for Linear
- I added a check for the interpolation range to `SplineInterpolation` since it is also needed for Linear. But I did not finally succeed in deleting this `checkXInRange` from `CubicSpline`.

**To test:**

- input validators can be quickly checked manually using doc test example, unit test available
- original `SplineInterpolation` for more than 2 points should work as before
- new linear 2 point interpolation can be tested with:

```Python
# create points for interpolation (input workspace)
dataX1 = np.arange(1, 50, 10) * 0.07
dataY1 = np.sin(dataX1) * 0.1
ws_to_match = CreateWorkspace(dataX1, dataY1)

# create two points to interpolate between (input workspace)
dataX2 = np.array([0.5, 2.5, 0.5, 2.5, 0.5, 2.5])
dataY2 = np.array([0.1, 0.15, 0.1, 0.15, 0.1, 0.15])
ws_to_interpolate = CreateWorkspace(dataX2, dataY2, NSpec=3)

interpolated_ws = SplineInterpolation(WorkspaceToMatch=ws_to_match, WorkspaceToInterpolate=ws_to_interpolate, Linear2Points=True)
```

Results may look like this:
[LinearInterpolation.pdf](https://github.com/mantidproject/mantid/files/636620/LinearInterpolation.pdf)

Fixes #18097.

RELEASE NOTES

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


martyngigg (commented):
While I agree that `Linear` is a far better name for the `Linear function` I don't think we should be introducing a second linear function to confuse users. The list of background functions in `MantidPlot` will now contain both `Linear` & `LinearBackground` and users will, I suspect, be quite confused by this.

We certainly want to move towards the name `Linear` and have `LinearBackground` still work in scripts where it is used. Can this be achieved? What is the reason that the new function has additional attributes?

reimundILL:
I fully agree with your point of view.

Some problems that I am facing are:
- `LinearBackground` is used for fitting, whereas `Linear` (at the moment) is a simple linear function: both functions have `Function1D` as base class and **I would have needed to override the same method** (`function1D`) in a different way (for fitting and for non-fitting case).

To solve the problems (it is more a software design decision):
- As a major step the function **base class(es) may need to be modified**? A new base class for non-fitting functions? Or a more general base class that can handle attributes and fitting parameters at the same time -> that means that there must be a new method in `Function1D` to purely set or evaluate function parameters (attributes) and a different one to evaluate a function with fitting parameters. Only then I see a possibility to be able to combine `Linear` and `LinearBackground`.
- I now changed `Linear`, `CubicSpline` and `SplineInterpolation` to use only attributes for a better understanding

Future topics:
- in my opinion, the `CurveFitting` framework should not be used for interpolation. Interpolation can be a regular algorithm?
- personally, I would generalise the code much more and introduce an algorithm  `Interpolation` instead of `SplineInterpolation` and put the function as option (`CubicSpline`, `Linear`, ... all already defined via GSL maybe no need of `BackgroundFunctions`). This may make the option `Linear2Points` redundant as well since the user will get error messages (e.g.: not enough points) and can choose another function for interpolation instead. Restrictions can be mentioned when calling the algorithm and in the documentation.

martyngigg:
Thanks for the extra information on the background to the changes.

I agree on the point about having `Interpolation` with options - this is much more sensible that separate algorithms. I don't think the `CurveFitting` framework is "used" here it's just that all of the function definitions live in that shared library so algorithms that depend on them were put there. It would really be more better to have a separate functions library I think - but this is a broader long-term design issue. 

The part I'm most struggling with is why there as to be a distinction between a fitting function and a non-fitting function. A function used without `Fit` can just be used as a function to calculate something based on the set parameters and X values. The attribute mechanism is a sightly clunky mechanism to get additional information to the function and keeping the evaluation signature the same. My concern with `Linear` written in this way is that it's confusing for users to work with and its primarily written in this way to make it easier for us to use on the code side.

@mantid-roman Any opinions on this?

mantid-roman:
From my point of view it was too much work for a simple linear interpolation between two points. The changes would have been justified if they introduced a more general interpolation, eg with polynomials of any order or gaussians. The case of arbitrary polynomials (including linear) could be covered by the BSpline. A special interface for an interpolating function would also be useful.

CubicSpline is a fitting function, changing parameters `y0, y1, ...` to attributes makes it unusable for fitting. It shouldn't have been done. I cannot see any good reasons for the changes to CubicSpline.

Function `Linear` defines attributes `n` and `xi, yi` with `i > 1` which it doesn't use and are there to satisfy the "interface" of the interpolating function required by the algorithm. If it's not a fitting function why it inherits IFunction?

reimundILL:
Thanks for the comments

Unfortunately my commit message is not meaningful, I modified the following:
- deleted the function `Linear`
- use parameters again in `CubicSpline`
- use `LinearBackground` also for the interpolation while keeping existing usages unchanged
- added two new, short virtual functions to `IFunction1D` for function and derivative evaluation

Hope it is a better approach now.

I will have a look at the BSpline function.

martyngigg:
@reimundILL Thanks for make some changes there. 

Roman and I are having difficulty seeing the benefits of having an entirely new attribute-based way of performing the calculation. I understand wanting to keep the linear-related information with `LinearBackground` but this feels like an abuse of the attribute system and a pollution of the function interface, which is already pretty heavy. As this is a special case for `SplineInterpolation` might it not be better to just do the gradient calculation locally rather than require all this extra code?

reimundILL:
Ok, I implement(ed) and need to test a solution, which does not introduce `Linear` and does not modify `IFunction1D`. The solution is then a specialisation of `SplineInterpolation`, less general but therefore very light.

martyngigg:
@reimundILL When you say specialisation do you mean a new algorithm? Is that necessary?

reimundILL:
Oh no, sorry, I mean I add some lines to `SplineInterpolation` for the `Lin2Points` option only. I do not introduce a new algorithm, I only add this specialisation for the `Lin2Points` to the existing algorithm.

martyngigg:
No problem, I just wanted to clarify. Thanks.

reimundILL:
@mantid-roman Maybe you can have a look: I am using `BSplines` now according to your suggestion and the interpolation points are correctly set. But then, `calculateSpline` (and `calculateDerivatives`) return zeros. This is due to the different interface of `CubicSplines` and `BSplines`. In `SplineSmoothing`, `BSplines` are used in a similar war than in `SplineInterpolation`. Is it still possible to use `BSplines` for a linear interpolation?

An implementation using the linear gsl interpolation seems to be easier at the moment.

mantid-roman:
I was suggesting to use BSplines for a case of a more general interpolation algorithm, not for this issue though. Sorry if I was misunderstood. The gsl or a hand-written linear interpolation should do here.

reimundILL:
@mantid-roman  Ok I agree, it is much easier to implement then.

reimundILL:
All manual tests pass for me, however, I need to modify the unit test. Some changes made:
- using point data from the beginning
- use `create` for the `OutputWorkspace` (`WorkspaceFactory`) in order to carry over the instrument and so on 
- implemented linear interpolation using gsl with unique pointers

reimundILL:
@martyngigg  The `ApplyPaalmanPingsCorrectionTest` is failing due to the test: initially I see that both input workspaces have **different units** (there can be a validator in `SplineInterpolation`?). Previously the unit was taken from the `WorkspaceToMatch`. Now it is the `WorkspaceToInterpolate`. There are two input workspaces and instrument and x-unit can be copied from the one or the other. I think the initial issue by VickyLynch was that she thinks that the input ws is `WorkspaceToInterpolate` and the test wants the `WorkspaceToMatch` being the parent input workspace. I need to see further.

reimundILL:
@martyngigg tests pass if I use the parent ws `WorkspaceToMatch`. Shall I fix the `ApplyPaalmanPingsCorrectionsTest`? Seems to be sensible to me. And the unit validator for the input workspaces?

reimundILL:
Adding a new input to choose from which input workspace to copy. Then, old behavior is still valid and new flexibility is achieved easily.

martyngigg:
@reimundILL Sorry, I had missed the notifications about this.

The new input sounds like a sensible option.

reimundILL:
@martyngigg I misunderstood the y-values outside of the interpolation range: I repeated the first and last interpolated y-values, but originally the first and last y-values of the WorkspaceToInterpolate should be repeated. For the workspace of derivatives all values out of range are zero. I added this information to the documentation.
And the unit tests did not work for this case. For the new ReferenceWorkspace, I test that x units get copied correctly.

AndreiSavici:
If you want this in the 3.9 release change the branch to merge into from `master` to `release-v3.9`

Framework/CurveFitting/src/Functions/CubicSpline.cpp
```diff
@@ -41,9 +36,9 @@ CubicSpline::CubicSpline()
   declareAttribute("x1", Attribute(1.0));
   declareAttribute("x2", Attribute(2.0));
 
-  declareParameter("y0", 0);
-  declareParameter("y1", 0);
-  declareParameter("y2", 0);
+  declareParameter("Y0", 0);
```
martyngigg: I think this change could break existing code if a script contains a function definition that sets these parameters. Is it necessary?

martyngigg (changes requested):

martyngigg:
@gvardany Is it vital that this goes into `3.9`? Given the beta has gone out and we're a few days from release there are a few more changes here than I'd usually be comfortable with.

martyngigg (approved):

gvardany:
@martyngigg It is not too urgent, but it would be nice.

Framework/CurveFitting/src/Functions/CubicSpline.cpp
```diff
@@ -88,22 +80,17 @@ void CubicSpline::setupInput(boost::scoped_array<double> &x,
 
     // if x[i] is out of order with its neighbours
     if (i > 1 && i < n && (x[i - 1] < x[i - 2] || x[i - 1] > x[i])) {
-      xSortFlag = true;
+      g_log.warning() << "Spline x parameters are not in ascending order. "
+                         "Only x values will be sorted.\n";
+      std::sort(x.get(), x.get() + n);
+      continue;
```
martyngigg: This previous code executed the `y[i] = ` line below but this now continues. Was this intended? It would seem to me that the `y` value would get missed if a `sort` occurred.

martyngigg (changes requested):
A further comment

martyngigg:
@gvardany I think I would prefer that this went into `master` and not `release-v3.9` so that it can have a chance to get some usage in the wild before getting into a stable build - `SplineInterpolation` is used quite a lot so I'd rather not take the risk if that's okay.

gvardany:
I agree, no problem on our side.

martyngigg:
Thanks. Did you see the other comment about the `continue` line `CubicSpline.cpp`?

gvardany:
I did not follow the implementation actually, maybe @reimundILL can comment?

reimundILL:
@martyngigg There is a change in the behavior of `SplineInterpolation`: I intend to sort all values in order to have x increasing (and the corresponding y- and e- values will stay attached to the correct x value) since this is the way it makes sense to me. (I think there is a unit test, but I need to have a look again) I think it is really bad to sort only x values and leave the other values unchanged!

Before, and still there for the `CubicSpline`, only x values will be sorted. I modified the `CubicSpline` together with the not anymore existing `Linear` function, I can leave this file unchanged. There was a mix of responsibilities between interpolation and function evaluation. I think the `CubicSpline` was too complicated for what it is.

Currently, at the beginning `ensureXIncreasing` should guarantee that the requirements for interpolation are met. This way `CubicSpline` will not have the change to sort x values only.

Will have a closer look at the `continue` line.

What is your opinion about the sorting? I do believe that our users understand the sorting like I did and in our reduction, we very likely deal with unsorted data. The reduction workflow would be much smarter with a proper sorting for the interpolation.

reimundILL:
@martyngigg some changes:
- using break instead of continue
- fix sorting for point data
- sorting histogram workspaces should be addressed by algorithm `SortXAxis`
- more detailed logging and documentation

previous behavior (point data):
si0

now:
si1

I may need to log an error in case of using linear interpolation with histogram data and unsorted x values.

reimundILL:
@martyngigg  I like to let you know (sorry for being late) that I will start my maternity leave now - no guarantee that I can respond during the following weeks. My colleagues are not familiar with this development, but can be asked. 

When I summarize the problems with the current `SplineInterpolation`, I find:
- A sorting problem (only x values, not documented)?
- A corrupted OutputWorkspace (simply wrong)
- An undefined (since there are two) InputWorkspace (documentation)

gvardany:
@martyngigg  I attempted to address your outstanding comment, and the outstanding bug noted by @reimundILL about the sorting. Could you please find some time to have another look? Seems we will be using `SplineInterpolation` a lot more in future, and there are set of issues fixed in this PR. It would be nice to get it merged, so we can proceed to further improvements, like error calculation, which is currently not done.

Framework/CurveFitting/src/Functions/CubicSpline.cpp
```diff
@@ -85,25 +79,39 @@ void CubicSpline::setupInput(boost::scoped_array<double> &x,
...
   }
 
   // sort the data points if necessary
   if (xSortFlag) {
     g_log.warning() << "Spline x parameters are not in ascending order. Values "
                        "will be sorted.\n";
-    std::sort(x.get(), x.get() + n);
+
+    using point = std::pair<double, double>;
+    std::vector<point> pairs;
```
martyngigg: The sorting looks good now.

My only remaining comment would be that as we know the size of the final pairs vector it would be good to set that initially, either in the constructor or using `.reserve`, to avoid `push_back` having to resize the vector as it goes.

Framework/CurveFitting/src/Functions/CubicSpline.cpp
```diff
@@ -61,8 +56,7 @@ void CubicSpline::function1D(double *out, const double *xValues,
   boost::scoped_array<double> y(new double[n]);
 
   // setup the reference points and calculate
-  if (m_recalculateSpline)
-    setupInput(x, y, n);
+  setupInput(x, y, n);
```
mantid-roman: Why was `m_recalculateSpline` dropped? `setupInput` may unnecessarily do expensive operations such as getting x values from attributes.

Framework/CurveFitting/src/Functions/CubicSpline.cpp
```diff
@@ -85,25 +79,39 @@ void CubicSpline::setupInput(boost::scoped_array<double> &x,
     std::string yName = "y" + num;
 
     x[i] = getAttribute(xName).asDouble();
+    y[i] = getParameter(yName);
```
mantid-roman: Can the index version of `getParameter` be used here? 
```
y[i] = getParameter(i);
```

gvardany:
I have recovered the parts that she removed for some reason.
With the tests passing, still `SplineInterpolation` behaves odd with linear (see the pic).  So I have to  dig deeper to understand what's going on.

screen shot 2017-07-10 at 14 17 38

gvardany:
Seem it works fine for point data, while for histogram data there is still something fishy. I will have a look and add more tests.

gvardany:
I am finally lost now, and will start from scratch and open separate PRs for cubic spline and interpolation.

PR #18157, opened 2016-12-01
Add SANSState

### LINKED ISSUE #18156: Create SANSState for ISIS SANS
Add a SANSState framework which allows to store SANS settings centrally. The SANSState object should be:
* self-validating
* typed
* serializable
* easy to reason about
* modular

AntonPiccardoSelg:
Fixes #18156

These changes adds the SANSState which is required for the second version of the ISIS reduction framework. The SANSState will allow to have the SANS settings:

* centralized
* self-validating
* typed
* serializable
* easy to reason about
* modular

The changes cannot be directly used by the user, hence a code review is probably the main review tool here.

The release notes will be updated once the entire SANS reducer backend has been roled out.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


AntonPiccardoSelg:
@martyngigg  This is the PR with the core elements of the new SANS reduction. It contains some common definitions and functionality and the state framwork (and tests).

 The failing build on OSX is not related to this PR.

AntonPiccardoSelg:
The updates performed are:
* packages and modules are lower case and trimmed (where applicable)
* the main package is a now in `scripts/SANS`
* a basic Readme file was added (to be extended)
* the `SANSState` prefix was removed in most places
* the old `SANSStateReduction` was renamed to `StateReductionMode`
* the builder and corresponding state was combined
* the `sans_parameters` decorator was renamed to `rename_descriptor_names`

Looking at the `automatic_setter` again, I had also added a parameter which allowed me to exclude the generation of setters for specific attributes. This is desired when an attribute is set by a file value. We don't want to give anyone an easy opportunity to change that value.

@martyngigg  Thanks for the suggestions! If there is anything else you think that should be done, let me know.

martyngigg:
Did you mean to add the `scripts/SANS/__init__.py` file? It will now mean that thing the top-level `SANS` directory will **not** be added to the path as it was before.

AntonPiccardoSelg:
No and also not the one in the `scripts` folder. I removed both.

martyngigg:
The new changes are a definitely improvement on readability. :+1: 

We seem to have a lot of code to deal with serialization/deserialization. Was the [`pickle`](https://docs.python.org/2.7/library/pickle.html#what-can-be-pickled-and-unpickled) module considered?

AntonPiccardoSelg:
I hadn't looked to deeply into `pickle` for this. But I can see some issues with `pickle`. 

The main reason we want to "serialize" and "deserialize" is to be able to pass all our settings into a Mantid algorithm via a `PropertyManagerProperty`. This means that when we "serialize" an object we convert it into a `dict` and when we "deserialize" an object, then we do so from an object of type `PropertyManager`. Hence deserializations does not seem suitable to be used with `pickle`. Regarding serialization, we would run into an issue with the types (in the sans_type module) that we have defined. These are nested classes (since we cannot use `enum`) and it appears that `pickle` does not deal well with that.

AntonPiccardoSelg:
Changes made:
* Removed `SANSConstants` class and capitalised constants
* renamed `sans_type` to `serializable_enum`
* added static methods `to_string` and `from_string` to the typed enums which require it. This feature is now a decorator

martyngigg:
:+1:  The `enums` module is much clearer now so you can see what types are provided by flicking through much more easily.


Things will undoubtedly change as other parts are added but I'm happy with this in its current state. :shipit:

PR #18152, opened 2016-12-01
Stitch1DMany new property 'ScaleFactorFromPeriod'

PranavBahuguna:
When stitching workspaces from multiple groups together, the algorithm currently allows a global scale factor to be specified using properties `UseManualScaleFactor` and `ManualScaleFactor`. This new property `ScaleFactorFromPeriod` allows a set of global scale factors to be determined from workspaces in groups from a specific period and applies them to workspaces in all other periods.

This property is only used when `UseManualScaleFactor` is true and `ManualScaleFactor` is set to its default value. Furthermore, the documentation regarding the `Stitch1DMany` algorithm has been updated and improved.

**To test:**

- Ensure all unit tests pass.
- Ensure that the `ScaleFactorFromPeriod` property accomplishes what it was meant to do.
- Ensure that the documentation is up to date and correct.

Fixes #17819.

*Release notes have been updated.*

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)? Is it well structured with small focussed classes/methods/functions?
- [ ] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant documentation been added/updated?
- [ ] Is user-facing documentation written in a user-friendly manner?
- [ ] Has developer documentation been updated if required?

- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

PR #18162, opened 2016-12-02
Stitch1DMany not stitching when scaling LHS workspace

### LINKED ISSUE #18158: `Stitch1DMany` doesn't stitch properly when scaling left-hand-side workspace
### Expected behavior

`Stitch1DMany` should be able to stitch workspaces correctly when `ScaleRHSWorkspace` is set to false.

### Actual behavior

Runs should be stitched.

The following throws an error:

```
x = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
y = [1, 1, 1, 1, 1, 1, 1, 1, 1]
ws_1 = CreateWorkspace(x, y)

x = [9, 10, 11, 12, 13, 14, 15, 16, 17, 18]
y = [2, 2, 2, 2, 2, 2, 2, 2, 2]
ws_2 = CreateWorkspace(x, y)

stitched, scale_factors = Stitch1DMany(InputWorkspaces='ws_1, ws_2',
                                       Params='1',
                                       ScaleRHSWorkspace=False)
```

However `Stitch1D` works fine:

```
stitched, scale_factors = Stitch1D(ws_1, ws_2, Params='1', ScaleRHSWorkspace=False)
```

### Platforms affected

All

raquelalvarezbanos:
`Stitch1DMany` stitches two or more workspaces by recursively calling `Stitch1D`. By default, it scales the right-hand-side workspace by an appropriate factor, however, there's the option to stitch the left-hand-side workspace (`ScaleRHSWorkspace=False`). This option does not seem to be working unless the workspaces to stitch share the same X values (which is not normally the case). 

When scaling the right-hand-side workspace (default behaviour) the algorithm will take the first two workspaces and stitch them together using `Stitch1D`. The resulting workspace and the next workspace on the list will be then stitched together, and so on.

When scaling the left-hand-side workspace (i.e. when `ScaleRHSWorkspace=False`), the algorithm first reverses the order of the input workspaces (and some other input parameters). Then it applies `Stitch1D` to the first pair, then the resulting workspace is stitched with the next workspace in the list, and so on. The fact that we reverse the order of the input workspaces is making `Stitch1D` fail, as it is not able to calculate properly the integration ranges, overlap regions, etc. The problem is solved if we don't reverse the workspace list. I think we don't need to do this as the order in which pairs of workspaces are stitched should have no effect on the final result (i.e. if we have three workspaces, I think it doesn't matter if we stitch first the first two workspaces and then the result with the third workspace, or if we first stitch the last two workspaces and then the result with the first workspace).

**To test:**

1. Test that the order in which pairs of workspaces are stitched has no effect on the final result. To test this, I've run the following script:

```
x = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
y = [1, 1, 2, 2, 1, 1, 1, 1, 1]
ws_1 = CreateWorkspace(x, y)

x = [9, 10, 11, 12, 13, 14, 15, 16, 17, 18]
y = [3, 3, 3, 3, 3, 3, 3, 3, 3]
ws_2 = CreateWorkspace(x, y)

x = [17, 18, 19, 20, 21, 22, 23, 24, 25, 26]
y = [4, 4, 5, 5, 5, 5, 5, 5, 5]
ws_3 = CreateWorkspace(x, y)

stitched_from_right_to_left, scale_factor_rtl_1 = Stitch1D(ws_2, ws_3, ScaleRHSWorkspace=False)
stitched_from_right_to_left, scale_factor_rtl_2 = Stitch1D(ws_1, stitched_from_right_to_left, ScaleRHSWorkspace=False)

stitched_from_left_to_right, scale_factor_ltr_1 = Stitch1D(ws_1, ws_2, ScaleRHSWorkspace=False)
stitched_from_left_to_right, scale_factor_ltr_2 = Stitch1D(stitched_from_left_to_right, ws_3, ScaleRHSWorkspace=False)

# 'stitched_from_right_to_left' and 'stitched_from_left_to_right' should match
```

2. Test the fix, i.e., test that `Stitch1DMany` is now able to stitch with `ScaleRHSWorkspace = False`: the script below should run successfully with no errors:

```
x = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
y = [1, 1, 1, 1, 1, 1, 1, 1, 1]
ws_1 = CreateWorkspace(x, y)

x = [9, 10, 11, 12, 13, 14, 15, 16, 17, 18]
y = [2, 2, 2, 2, 2, 2, 2, 2, 2]
ws_2 = CreateWorkspace(x, y)

# stitched_1 and stitched_2 should match
stitched_1, scale_factors = Stitch1D(ws_1, ws_2, Params='1', ScaleRHSWorkspace=False)
stitched_2, scale_factors = Stitch1DMany(InputWorkspaces='ws_1, ws_2',
                                       Params='1',
                                       ScaleRHSWorkspace=False)

# stitched_1 and stitched_2 should match
stitched_1, scale_factors = Stitch1D(ws_1, ws_2,
                                       Params='1',
                                       StartOverlap=9,
                                       EndOverlap=10,
                                       ScaleRHSWorkspace=False)
stitched_2, scale_factors = Stitch1DMany(InputWorkspaces='ws_1, ws_2',
                                       Params='1',
                                       StartOverlaps=[9],
                                       EndOverlaps=[10],
                                       ScaleRHSWorkspace=False)
```

Fixes #18158.

No need to mention this in the release notes, as this issue was reported by me.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/src/Stitch1DMany.cpp
+++ b/Framework/Algorithms/src/Stitch1DMany.cpp
@@ -134,13 +134,6 @@ std::map<std::string, std::string> Stitch1DMany::validateInputs() {
   if (m_params.empty())
     errors["Params"] = "At least one parameter must be given.";
 
-  if (!m_scaleRHSWorkspace) {
-    // Flip these around for processing
-    std::reverse(m_inputWorkspaces.begin(), m_inputWorkspaces.end());
-    std::reverse(m_startOverlaps.begin(), m_startOverlaps.end());
-    std::reverse(m_endOverlaps.begin(), m_endOverlaps.end());
-  }
-
   m_scaleFactors.clear();
   m_outputWorkspace.reset();
 
--- a/Framework/Algorithms/test/Stitch1DManyTest.h
+++ b/Framework/Algorithms/test/Stitch1DManyTest.h
@@ -51,12 +51,12 @@ private:
 
     MatrixWorkspace_sptr ws =
         WorkspaceFactory::Instance().create("Workspace2D", 2, nbins + 1, nbins);
-    ws->dataX(0) = xData1;
-    ws->dataX(1) = xData2;
-    ws->dataY(0) = yData1;
-    ws->dataY(1) = yData2;
-    ws->dataE(0) = eData1;
-    ws->dataE(1) = eData2;
+    ws->mutableX(0) = xData1;
+    ws->mutableX(1) = xData2;
+    ws->mutableY(0) = yData1;
+    ws->mutableY(1) = yData2;
+    ws->mutableE(0) = eData1;
+    ws->mutableE(1) = eData2;
     ws->getAxis(0)->unit() = UnitFactory::Instance().create("Wavelength");
 
     return ws;
@@ -187,18 +187,18 @@ public:
     auto stitched = boost::dynamic_pointer_cast<MatrixWorkspace>(outws);
     TS_ASSERT_EQUALS(stitched->getNumberHistograms(), 2);
     TS_ASSERT_EQUALS(stitched->blocksize(), 17);
-    TS_ASSERT_DELTA(stitched->readY(0)[0], 1, 0.00001);
-    TS_ASSERT_DELTA(stitched->readY(0)[9], 1, 0.00001);
-    TS_ASSERT_DELTA(stitched->readY(0)[16], 1, 0.00001);
-    TS_ASSERT_DELTA(stitched->readY(1)[0], 2, 0.00001);
-    TS_ASSERT_DELTA(stitched->readY(1)[9], 2, 0.00001);
-    TS_ASSERT_DELTA(stitched->readY(1)[16], 2, 0.00001);
-    TS_ASSERT_DELTA(stitched->readE(0)[0], 1, 0.00001);
-    TS_ASSERT_DELTA(stitched->readE(0)[9], 0.77919, 0.00001);
-    TS_ASSERT_DELTA(stitched->readE(0)[16], 1.24316, 0.00001);
-    TS_ASSERT_DELTA(stitched->readE(1)[0], 1.41421, 0.00001);
-    TS_ASSERT_DELTA(stitched->readE(1)[9], 1.10982, 0.00001);
-    TS_ASSERT_DELTA(stitched->readE(1)[16], 1.79063, 0.00001);
+    TS_ASSERT_DELTA(stitched->y(0)[0], 1, 0.00001);
+    TS_ASSERT_DELTA(stitched->y(0)[9], 1, 0.00001);
+    TS_ASSERT_DELTA(stitched->y(0)[16], 1, 0.00001);
+    TS_ASSERT_DELTA(stitched->y(1)[0], 2, 0.00001);
+    TS_ASSERT_DELTA(stitched->y(1)[9], 2, 0.00001);
+    TS_ASSERT_DELTA(stitched->y(1)[16], 2, 0.00001);
+    TS_ASSERT_DELTA(stitched->e(0)[0], 1, 0.00001);
+    TS_ASSERT_DELTA(stitched->e(0)[9], 0.77919, 0.00001);
+    TS_ASSERT_DELTA(stitched->e(0)[16], 1.24316, 0.00001);
+    TS_ASSERT_DELTA(stitched->e(1)[0], 1.41421, 0.00001);
+    TS_ASSERT_DELTA(stitched->e(1)[9], 1.10982, 0.00001);
+    TS_ASSERT_DELTA(stitched->e(1)[16], 1.79063, 0.00001);
 
     // Test out sclae factors
     std::vector<double> scales = alg.getProperty("OutScaleFactors");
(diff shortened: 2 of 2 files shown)


gvardany:
I have some general remarks about stitching procedure. 
In my opinion stitching operation should be commutative, e.g. A stitch B = B stitch A, regardless of what is the actual left or right based on the x-axis range. Moreover the algorithm should be generic enough to stitch the workspaces even if RHS x-range completely covers the range in LHS or other way around (full overlap), in which case even the notion of LHS and RHS is undefined. In principle it should also be able to stitch even if there is no overlap at all (currently it fails).

I tried this:
```
x = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
y = [1, 1, 1, 1, 1, 1, 1, 1, 1]
ws_1 = CreateWorkspace(x, y)

x = [9, 10, 11, 12, 13, 14, 15, 16, 17, 18]
y = [2, 2, 2, 2, 2, 2, 2, 2, 2]
ws_2 = CreateWorkspace(x, y)

stitched_1, scale_factors = Stitch1D(ws_1, ws_2, Params='1', ScaleRHSWorkspace=False)
stitched_2, scale_factors = Stitch1D(ws_2, ws_1, Params='1', ScaleRHSWorkspace=True)
```
and expected the same output. Instead the second call to `Stitch1D` fails, since the x-axis range is higher in `ws_2`, so the current logic of overlap finding is screwed. 

In an abstract level, I think LHS and RHS are just a naming convention, but the algorithm should be commutative (with the account of which workspace to scale, of course). In this case the reversing of the workspace list in `Stitch1DMany` looks an elegant way of getting the alternative scaling behaviour.

I would also replace `ScaleRHSWorkspace` boolean property in `Stitch1D` with `ScaleWorkspace`  string array property with allowed values of `['LHS', 'RHS']`  and `ScaleRHSWorkspace` property in `Stitch1DMany` with `StitchDirection` with allowed values of `['LeftToRight', 'RightToLeft']` and just reverse the workspace list accordingly (and that should be just enough).

That's at least how I understand it from the first glance, but maybe I am missing something fundamental here... What do you think?

raquelalvarezbanos:
Thanks for having a look at this. I understand what you say and it makes perfect sense. I agree that in an abstract level, LHS and RHS should be just a naming convention, however, in reflectometry, they do have a meaning: LHS refers to the "low wavelength" run and RHS refers to the "high wavelength" run. This is because a typical run consists of two or three (depending on the experiment) different runs with the same parameters (temperature, magnetic field, etc) but at three different angles. This way they always end up with something like this:

stitch

where the black curve would be the "low wavelength" run (i.e. the run measured at an angle `theta_1`) and the red one would be the "high wavelength" run (measured at an angle `theta_2` > `theta_1`). This means that 1) there's always going to be an overlapping region but not a "full" overlap, 2) the x values are always going to be in ascending order 3) the x-range is going to be higher in the RHS workspace. So as you see this is very specific to reflectometry experiments, and the algorithm makes some assumptions in terms of the input workspaces, overlapping region, etc. I agree with you that in a general case the algorithms should behave as you describe. In fact, I will be happy to help generalize them if other people need to use them, but at this stage I was more concerned about the fact that `Stitch1DMany` does not seem to be working when `ScaleRHSWorkspace` is set to false.

Regarding the names of the properties, and the algorithms in general, I think they all come from old scripts used by scientists, so I would prefer to keep the current property names they've been using.

gvardany:
I see that this algorithm is specifically designed for reflectometry. But the operation does seem a very general transformation of histograms. In fact we were looking for a stitching algorithm in completely different scenario (for QENS backscattering), and we were not able to reuse this algorithm because of the overlap restriction. But even in reflectometry, one might in principle want to inspect the workspaces in different unit, inverse to `q`, let's say `lambda`, in which case for your plot above you will have left and right swapped, and both of them will have descending x-axis.

Anyway if your users are happy, I don't insist on this now. But keep in mind that in future these algorithms will be used for ILL reflectometry instruments as well, and towards this we would like to prepare the tools as flexible as possible, maybe we will consider creating new versions of these algorithms.

For the moment I would at least make notes in the documentations of these algorithms about all the restrictions implied on the inputs (ascending x-axis, LHS, RHS ordering and obligatory partial overlap).

It would be good as well, if these can be checked at some point in the algorithms as early as possible, so one could throw better error messages in case of these requirements don't meet.

raquelalvarezbanos:
I agree that all these details should be better documented. I have created an issue for this which I will fix as soon as I can (definitely for this release): https://github.com/mantidproject/mantid/issues/18200. The reason I am not updating documentation as part of this PR is because there's another PR (https://github.com/mantidproject/mantid/pull/18152) which adds some changes to the documentation and I don't want to create conflicts between branches. I'll wait until that PR is merged in and then start working on the documentation, if you agree. I'll also let you know when that PR is ready, in case you want to test it.

Regarding the example you mention about QENS backscattering, where you couldn't use this algorithm, how did you solve it? Did you implement your own code to do the stitching? Anyway, I am happy if you need to re-write these algorithms or create new versions to make them more general, in fact, I think it would be very good if we had a general version that works for everyone, and I am more than happy to help you with that if you need me. If that's the case, when you start working on it, please let me know.

Let me know if you are OK with improving the documentation/input checking as a separate issue.

gvardany:
Yes it's fine to have separate issue for documentation/input validation. 

Just for the record, I would also expect the `ManualScaleFactor` to be a double list of size N-1, similar to overlaps, instead of being one global factor for all. Of course if only one number is given it can fall back to global scaling behaviour.
 
For our case, we did not do stitching at all at the end (it was not a strict requirement, but just a nice to have one), so that's not critical.

Our reflectometry effort will start actively next year, sure we need to collaborate there.

So I will start reviewing changes in this PR now.

raquelalvarezbanos:
Yes, I also agree that `ManualScaleFactor` should be a vector of N-1 components. I am not sure why the algorithm only takes one value. If you need to change that in the future I'll check with scientists if it's OK to change it, as it would make more sense.

gvardany:
OK :shipit:

PR #17869, opened 2016-10-21
SpectrumInfo: add Detector access

SimonHeybrock:
`SpectrumInfo` now provides access to the underlying detector or detector group. This allows for refactoring algorithms that require more than basic information from the instrument for a spectrum, such as computing the solid angle as seen from the sample, which we are not supporting in `SpectrumInfo` (at least for now).

**To test:**

Thorough code review.

Fixes #17771.
No release notes, internal change (there might be a speedup for certain algorithms).

---
#### Reviewer

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):
##### Code Review
- [x] Is the code of an acceptable quality?
- [x] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)? Is it well structured with small focussed classes/methods/functions?
- [x] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?
##### Functional Tests
- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant documentation been added/updated?
- [ ] Is user-facing documentation written in a user-friendly manner?
- [ ] Has developer documentation been updated if required?
- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

PR #18165, opened 2016-12-02
Base class for fitting functions dependent on momentum transfer (Q)

### LINKED ISSUE #18101: Base class for fitting functions dependent on momentum transfer (Q)
Functions to model the dynamic structure factor `S(Q,E)` need an attribute `Q` become aware of the momentum transfer value for the particular spectrum to be fitted. However, `Q` is not part of the model, but rather a property of the data.

<b>Requirements</b>:
* If the spectrum belongs to a `MatrixWorkspace`, then attempt to learn the `Q` values for each spectrum: Override methods `setWorkspace` and `setMatrixWorkspace` of class `IFunction`. Save the Q-values in a standard vector `Qvalues`.
* Function attribute "Q" can be updated either directly by the user input or by assigning from a value out of `Qvalues`. The particular selected vector index is given by attribute "WorkspaceIndex".
* Rules for updating attribute "Q":
  1. Default value upon initialization is `MAX_DBL`.
  2. If attribute "WorkspaceIndex" is updated and `Qvalues` is not empty, update attribute "Q" with Qvalues[WorkspaceIndex]. Using algorithm `PlotPeakByLogValue` is a user case example.
  3. If domain is of type FunctionDomain1DSpectrum, update attribute "WorkspaceIndex" with the index of the domain, thus prompting the previous rule.
  4. If user inputs a value and `Qvalues` is empty, update attribute "Q" with the input value.

jmborr:
Work done: source, unit test, and documentation.

There's nothing to test manually because this fit function contains pure virtual methods (hence cannot be implemented). Please review the code.

Fixes #18101.

[release notes]() have been updated

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/CurveFitting/CMakeLists.txt
+++ b/Framework/CurveFitting/CMakeLists.txt
@@ -78,6 +78,7 @@ set ( SRC_FILES
 	src/Functions/FlatBackground.cpp
 	src/Functions/FullprofPolynomial.cpp
 	src/Functions/FunctionGenerator.cpp
+	src/Functions/FunctionQDepends.cpp
 	src/Functions/GausDecay.cpp
 	src/Functions/GausOsc.cpp
 	src/Functions/Gaussian.cpp
@@ -231,6 +232,7 @@ set ( INC_FILES
 	inc/MantidCurveFitting/Functions/FlatBackground.h
 	inc/MantidCurveFitting/Functions/FullprofPolynomial.h
 	inc/MantidCurveFitting/Functions/FunctionGenerator.h
+	inc/MantidCurveFitting/Functions/FunctionQDepends.h
 	inc/MantidCurveFitting/Functions/GausDecay.h
 	inc/MantidCurveFitting/Functions/GausOsc.h
 	inc/MantidCurveFitting/Functions/Gaussian.h
@@ -376,6 +378,7 @@ set ( TEST_FILES
 	Functions/ExpDecayTest.h
 	Functions/FlatBackgroundTest.h
 	Functions/FullprofPolynomialTest.h
+	Functions/FunctionQDependsTest.h
 	Functions/GausDecayTest.h
 	Functions/GausOscTest.h
 	Functions/GaussianComptonProfileTest.h
--- a/Framework/CurveFitting/src/Functions/FunctionQDepends.cpp
+++ b/Framework/CurveFitting/src/Functions/FunctionQDepends.cpp
@@ -30,8 +30,10 @@ namespace Functions {
 
 /**
  * @brief declare commonattributes Q and WorkspaceIndex.
- * Subclasses containing additional attributes should override this method by declaring the additional
- * attributes and then calling the parent (this) method to declare Q and WorkspaceIndex.
+ * Subclasses containing additional attributes should override this method by
+ * declaring the additional
+ * attributes and then calling the parent (this) method to declare Q and
+ * WorkspaceIndex.
  */
 void FunctionQDepends::declareAttributes() {
   this->declareAttribute("Q", Attr(EMPTY_DBL()));
@@ -39,8 +41,10 @@ void FunctionQDepends::declareAttributes() {
 }
 
 /**
- * @brief Update attributes WorkspaceIndex and Q according to certain precedence rules.
- * Subclasses featuring additional attributes should override and insert a call to
+ * @brief Update attributes WorkspaceIndex and Q according to certain precedence
+ *rules.
+ * Subclasses featuring additional attributes should override and insert a call
+ *to
  * this method within the overriding setAttribute function.
  * There are two ways to update Q: (i) loading the value from the spectrum, and
  *(ii) manual
@@ -136,8 +140,7 @@ std::vector<double> FunctionQDepends::extractQValues(
         double usignTheta = 0.5 * spectrumInfo.twoTheta(wi);
         double q = Mantid::Kernel::UnitConversion::run(usignTheta, efixed);
         qs.push_back(q);
-      }
-      catch (Kernel::Exception::NotFoundError &) {
+      } catch (Kernel::Exception::NotFoundError &) {
         g_log.debug("Cannot populate Q values from workspace");
         qs.clear();
         break;
--- a/Framework/CurveFitting/src/Functions/FunctionQDepends.cpp
+++ b/Framework/CurveFitting/src/Functions/FunctionQDepends.cpp
@@ -8,6 +8,7 @@
 #include "MantidAPI/MatrixWorkspace.h"
 #include "MantidAPI/NumericAxis.h"
 #include "MantidKernel/UnitConversion.h"
+#include "MantidAPI/SpectrumInfo.h"
 // third party libraries
 // N/A
 // standard library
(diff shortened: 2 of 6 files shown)


Framework/CurveFitting/src/Functions/FunctionQDepends.cpp
```diff
@@ -0,0 +1,152 @@
...
+      dynamic_cast<Mantid::API::NumericAxis *>(workspace->getAxis(1));
+  if (axis_ptr) {
+    const boost::shared_ptr<Kernel::Unit> &unit_ptr = axis_ptr->unit();
+    if (unit_ptr->unitID() == "MomentumTransfer") {
+      qs = axis_ptr->getValues();
+    }
+  }
+  // ...otherwise, compute the momentum transfer for each spectrum, if possible
+  else {
+    size_t numHist = workspace->getNumberHistograms();
```
MichaelWedel: @SimonHeybrock et al are currently rolling out `API::SpectrumInfo`, which makes it much easier to extract `2Theta`. An example is in this PR: https://github.com/mantidproject/mantid/pull/17869/files#diff-da2f4803dedc2187a4d76c3daefdd171 - I'm sure @SimonHeybrock has more examples, also with `2Theta`.

Framework/CurveFitting/test/Functions/FunctionQDependsTest.h
```diff
@@ -0,0 +1,163 @@
...
+  void testConstruction() {
+    TS_ASSERT_THROWS_NOTHING(ImplementsFunctionQDepends f);
+  }
+
+  void testSetWorkspace() {
+    double startX{0.0}, endX{1.0};
+    ImplementsFunctionQDepends f;
+    // test with an non matrix workspace
+    TS_ASSERT_THROWS_NOTHING(
+        f.setMatrixWorkspace(this->unsuitableWS(), 0, startX, endX));
```
MichaelWedel: If the workspace is unsuitable, should this maybe throw an exception? Not sure if this situation can occur in the real world, but if it's completely unsuitable what is the function going to do?

docs/source/fitfunctions/FunctionQDepends.rst
```diff
@@ -0,0 +1,30 @@
...
+- The magnitude of the momemtum transfer, :math:`Q`, as one of their attributes.
+
+Fitting functions for QENS data depending on :math:`Q` should derive from this class.
+
+There are two ways to update attribute :math:`Q` in the fit funtion:
+
+- The user inputs a particular value.
+- The spectrum contains :math:`Q` in addition to the range of energy transfers.
+
+Conflict may arise when both options are available. In that case, priority is given to the :math:`Q`-value contained
```
MichaelWedel: Is there a particular reason why the workspace `Q` takes precedence? This way it's not possible to override it if for some reason `Q` in the workspace is incorrect.

Framework/CurveFitting/src/Functions/FunctionQDepends.cpp
```diff
@@ -0,0 +1,152 @@
...
+namespace CurveFitting {
+namespace Functions {
+
+/* ===========
+   Public
+   ===========*/
+
+FunctionQDepends::FunctionQDepends() {
+  this->declareAttribute("Q", Attr(EMPTY_DBL()));
+  this->declareAttribute("WorkspaceIndex", Attr(EMPTY_INT()));
```
jmborr: To be honest, I didn't know of this method! When I started coding my first function I took as model one function that declared attributes in the constructor (there are few of these functions) and I went on from there. I will move the declarations to this method.

Framework/CurveFitting/src/Functions/FunctionQDepends.cpp
```diff
@@ -0,0 +1,152 @@
...
+ * priority is to accept Q from the spectrum, if a Q value can be derived from
+ *such. In
+ * this case the existing Q value will be overwritten, irrespective of the
+ *mannier in which
+ * the old Q value was set.
+ *
+ * @param attName name of the attribute
+ * @param attValue  value of the attribute
+ */
+void FunctionQDepends::setAttribute(const std::string &attName,
```
jmborr: Yes, you are right. I will insert a comment.

Framework/CurveFitting/src/Functions/FunctionQDepends.cpp
```diff
@@ -0,0 +1,152 @@
...
+                                    const Attr &attValue) {
+  // Q value is tied to WorkspaceIndex if we have a list of Q values
+  if (attName == "WorkspaceIndex") {
+    size_t wi{static_cast<size_t>(
+        attValue.asInt())}; // ah!, the "joys" of C++ strong typing.
+    if (!m_vQ.empty() && wi <= m_vQ.size()) {
```
jmborr: uh uh, you are right. Good catch!

Framework/CurveFitting/src/Functions/FunctionQDepends.cpp
jmborr: Thanks for the tip. I couldn't get rid of the detector variable because of Efixed, but I removed quite a few lines of code using method spectrumInfo::twoTheta

Framework/CurveFitting/test/Functions/FunctionQDependsTest.h
jmborr: There are scenarios when the spectra of the workspace will not contain the necessary information to extract the value of the momentum transfer. In this case, the user will input the Q value either in one of the available graphical interfaces of Mantid, or in a python script statement. That's the reason Q is exposed to the user as an attribute, instead of declared as a private member. For instance, the user may want to load some cheap & dirty synthetic data from a file to a workspace, but the file does not contain the Q-values. The user then has the option to manually insert these values.

docs/source/fitfunctions/FunctionQDepends.rst
jmborr: There are two mutually exclusive user cases and I had to give preference to one of them. The first user case is the one you described, the case when you have corrupt input data. The second user case is when you have correct data but the inexperienced user screws up by manually overriding the Q-value, typically fumbling with the Q-attribute in one of the Mantid GUIs.
In my experience, the latter user case is much more common because most of our users are inexperienced and because reduction will produce spectra with correct Q-values. In the rare event of corrupt data I rather have the user correct their data _outside_ of the fitting process (thus no manual Q override). 
Note there is one exception to this rule: When the data does not contain Q-values (you can think of this data as corrupt), then the user can manually insert them.

Framework/CurveFitting/CMakeLists.txt
```diff
@@ -78,6 +78,7 @@ set ( SRC_FILES
 	src/Functions/FlatBackground.cpp
 	src/Functions/FullprofPolynomial.cpp
 	src/Functions/FunctionGenerator.cpp
+	src/Functions/FunctionQDepends.cpp
```
jmborr: Initially I placed it on API as you suggest, but then I found out I could not write a unit test because DataObjects (Workspaces) are not built before API. The other sensible place down the make hierarchy was CurveFitting, so I placed it there.

jmborr:
@MichaelWedel Hi Michael, can you please take a look to the changes, and if happy give this pull request the pass?

Framework/CurveFitting/src/Functions/FunctionQDepends.cpp
```diff
@@ -0,0 +1,155 @@
...
+  }
+  // ...otherwise, compute the momentum transfer for each spectrum, if possible
+  else {
+    const auto &spectrumInfo = workspace->spectrumInfo();
+    size_t numHist = workspace->getNumberHistograms();
+    for (size_t wi = 0; wi < numHist; wi++) {
+      try {
+        Mantid::Geometry::IDetector_const_sptr detector;
+        detector = workspace->getDetector(wi);
+        double efixed = workspace->getEFixed(detector);
```
MichaelWedel: I seem to remember that `EFixed` was not part of `SpectrumInfo` yet, but it might be worthwhile keeping track of what @SimonHeybrock is doing in that direction, the code might become even shorter then and save some direct workspace access.

MichaelWedel:
I've looked at the code and all questions from the first review have been cleared up. I think it would be good if someone else took a final look at this, preferably @mantid-roman who has most knowledge of the function fitting framework. @jmborr, do you already have a concrete example in mind where this is going to be used? Maybe it's then also easier to see if there's another way to implement this without being too late for an implementation change.

Otherwise I think it's okay to :shipit:

jmborr:
@MichaelWedel After this is merged into master, my next issue was to refactor fit function [TeixeiraWaterSQE](http://docs.mantidproject.org/nightly/fitfunctions/TeixeiraWaterSQE.html) to inherit from FunctionQDepends. Then I'll be able to test the [indirect analysis interface](http://docs.mantidproject.org/nightly/interfaces/Indirect_DataAnalysis.html#i-q-t-fit), the multifit interface, and sequential fitting algorithms like [PlotPeakByLogValue](http://docs.mantidproject.org/nightly/algorithms/PlotPeakByLogValue-v1.html) to see how the automatic update of Q fares.

jmborr:
@MichaelWedel @quantumsteve Thanks!

PR #18171, opened 2016-12-05
MatchPeaks :: minor change in peak finding

### LINKED ISSUE #18168: MatchPeaks :: fix the fitted peak error's check
MatchPeaks is used in IndirectILLReductionQENS in PR #18131.
A small fix is needed here concerning the check of the fitted peak's error, and the logic of falling back to taking the maximum, when the fit result is not reliable.

gvardany:
This PR slightly modifies the peak finding logic in `MatchPeaks` (used by `IndirectILLReductionQENS` in PR #18131 for peak alignment) to cope with some pathological cases if the peak is not so clear.

The logic should be as follows:
If the fitted peak (with it's error taken into account) is within the trusted region (not far from the central bin), it's position should be taken. Otherwise the position of the maximum should be taken if it is also inside the trusted region. If not, then the central bin is taken, meaning that no circular shift will actually be performed (since `to_shift = peak_bin - central_bin`)

**To test:**
Make sure the existing test passes.

Fixes #18168 .

Does not need to be in the release notes, nor in documentation.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/PythonInterface/plugins/algorithms/MatchPeaks.py
+++ b/Framework/PythonInterface/plugins/algorithms/MatchPeaks.py
@@ -236,7 +236,6 @@ class MatchPeaks(PythonAlgorithm):
 
         # Initialisation
         peak_bin = np.ones(mtd[self._input_ws].getNumberHistograms()) * mid_bin
-        peak_plus_error = np.zeros(mtd[self._input_ws].getNumberHistograms())
 
         # Bin range: difference between mid bin and peak bin should be in this range
         tolerance = int(mid_bin / 2)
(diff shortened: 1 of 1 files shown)


Framework/PythonInterface/plugins/algorithms/MatchPeaks.py
```diff
@@ -236,50 +236,71 @@ def _get_peak_position(self, input_ws):
...
         peak_bin = np.ones(mtd[self._input_ws].getNumberHistograms()) * mid_bin
-        peak_plus_error = np.zeros(mtd[self._input_ws].getNumberHistograms())
 
         # Bin range: difference between mid bin and peak bin should be in this range
         tolerance = int(mid_bin / 2)
 
         x_values = mtd[self._input_ws].readX(0)
 
         for i in range(mtd[self._input_ws].getNumberHistograms()):
+
```
martyngigg: Can we remove some of the blank lines? It feels like a bit too much whitespace is taken up here.

martyngigg (changes requested):
Apart from the minor formatting changes I think the changes seem sensible and the tests are still passing without changes.

martyngigg (approved):

martyngigg:
:shipit:

PR #18190, opened 2016-12-06
Calculate intervals for Monte Carlo parameter estimation.

### LINKED ISSUE #18133: Crystal Field: calculate intervals for Monte Carlo parameter estimation.
The input intervals for `EstimateFitParameters` algorithms can be calculated based on the desired maximum energy splitting provided by the user.

Fix all peak parameters during the call to `EstimateFitParameters` to prevent errors due to a change of the number of fitting parameters.

This is a part of #15684

mantid-roman:
Added calculations for the inputs to `EstimateFitParameters` for the field parameters.

**To test:**

Use this script as an example:

```
    from CrystalField import CrystalField, CrystalFieldFit, Background, Function

    # Create some crystal field data
    origin = CrystalField('Ce', 'C2v', B20=0.37737, B22=3.9770, B40=-0.031787, B42=-0.11611, B44=-0.12544,
                          Temperature=44.0, FWHM=1.1)
    x, y = origin.getSpectrum()
    ws = makeWorkspace(x, y)

    # Define a CrystalField object with parameters slightly shifted.
    cf = CrystalField('Ce', 'C2v', B20=0, B22=0, B40=0, B42=0, B44=0,
                      Temperature=44.0, FWHM=1.0, ResolutionModel=([0, 100], [1, 1]), FWHMVariation=0)

    # Set any ties on the field parameters.
    cf.ties(B20=0.37737)
    # Create a fit object
    fit = CrystalFieldFit(cf, InputWorkspace=ws)
    # Find initial values for the field parameters.
    # You need to define the energy splitting and names of parameters to estimate.
    # Optionally additional constraints can be set on tied parameters (eg, peak centres).
    fit.estimate_parameters(EnergySplitting=50,
                            Parameters=['B22', 'B40', 'B42', 'B44'],
                            Constraints='20<f1.PeakCentre<45,20<f2.PeakCentre<45',
                            NSamples=1000)
    print 'Returned', fit.get_number_estimates(), 'sets of parameters.'
    # The first set (the smallest chi squared) is selected by default.
    # Select a different parameter set if required
    fit.select_estimated_parameters(3)
    print cf['B22'], cf['B40'], cf['B42'], cf['B44']
    # Run fit
    fit.fit()
```

Fixes #18133.

The release notes do not need update as they already mention the new functionality.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/CurveFitting/src/Algorithms/EstimateFitParameters.cpp
```diff
@@ -102,6 +106,39 @@ void fixBadParameters(CostFunctions::CostFuncFitting &costFunction,
   }
 }
 
+/// A class for sorting and storing sets of best function parameters.
+class BestParameters {
+  /// Maximum size of the store
+  size_t m_size;
+  /// Actual storage.
+  std::map<double, GSLVector> m_params;
```
martyngigg: Is the order important here? If all we want is a lookup then `unordered_map` should be preferred: http://stackoverflow.com/questions/2196995/is-there-any-advantage-of-using-map-over-unordered-map-in-case-of-trivial-keys

mantid-roman: At  this moment I think it is important. It helps select the `n` best values and find the worst of the best which has to be thrown away after a new best is found. The user may also want to know which is the best, which is the second best and so on.

mducle:
Looks good! Thanks for putting the `IntensityScaling` bugfix in too.

I still have one problem though - if in the example above, I omit the `ResolutionModel` or set `FWHMVariation` to a non-zero value, I get a `Cost function changed number of parameters` error. Could you add something to the `estimate_parameters` method to check if a resolution model has been set, and if not to set it to be a constant (to the value of the `FWHM` parameter), please? After this, we should note in the documentation that using `estimate_parameters` implies a resolution model - which will be set automatically if the user doesn't specify it.

mantid-roman:
@mducle Thanks. I forgot about it. The `ResolutionModel` isn't actually required. It was a way of fixing all peak parameters at the moment. Now I'll use the `FixAllPeaks` property which should do the trick.

mducle:
Looks good and works for me. :shipit:

martyngigg (approved):

PR #18205, opened 2016-12-07
Fix bug with underscore in table column names

### LINKED ISSUE #18196: Project loading with a plot from a  table can cause a blank plot
Reported by Adrian Hillier.

### Expected behavior
When a project is saved and there is a plot open which was created from the columns of a table the plot should reappear when the project is reloaded.

### Actual behavior
The reloaded plot is blank

Create a table with columns with an underscore in (such as `run_number`). Then create a plot that uses this column as either X, Y or E. Save and close mantid, reopen and reload the project. The plot will be blank

### Platforms affected
All.

samueljackson92:
Fixes the issue with blank plots being loaded when created from a table window.

**To test:**

- Follow the instructions in the bug description in the issue report.
- Apply the fix and repeat instructions. Should appear fixed now.

Fixes #18196 .

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


soininen:
I created a table using a script:
```python
table = CreateEmptyTableWorkspace()

table.addColumn('double', 'noundesrcores')
table.setPlotType(0, 1)
table.addColumn('double', 'column_with_underscores')
table.addRow([0.23, -0.42])
table.addRow([2.3, -4.2])
table.addRow([23, -43])
table.addRow([66, -55])
```
Clicked on the 'column_with_underscores' column and chose Plot->Line+Symbol. Mantid plotted only the last two rows, so there is something fishy going on here. Anyway, saving the project, exiting, and reloading will give a completely blank plot before these changes are applied and the changes fix this problem. However, with and without the changes the first two rows were set to zeros, again indicating that something is wrong with numbers in the table workspaces.

samueljackson92:
@soininen I cannot reproduce the issue with the zero rows. For me when I load back in with the changes applied I still get all four points plotted.

screen shot 2016-12-13 at 11 29 45

soininen:
@samueljackson92 I see. Anyway, I've create an issue #18275 for the missing data points in the plots, as this seems to be an unrelated issue.

soininen:
This PR seems to fix what it promises. :shipit:

samueljackson92:
@soininen perhaps. I believe all the values for a table window get dumped into the .mantid project file when written out (which is dumb because we also save the table to a nexus file, so this is probably done for legacy reasons or is a hangover from QtiPlot). Might be worth opening up the project file and seeing if anything looks odd in there with the French locale?

soininen:
@samueljackson92 The plotting fails even without saving/loading the project. I'm just wondering if QtiPlot tries to convert the strings in the table cells to numbers, and this wouldn't work if it expects commas as decimal separators (default in the French locale) while TableWorkspace has dots there.

PR #18211, opened 2016-12-07
Reflectometry interface (Polref): Fix to have 'Get Defaults' wrap polarization constant values in quote marks

### LINKED ISSUE #18208: Reflectometry interface (Polref): 'Get defaults' not wrapping double list parameters in quote marks
### Expected behavior
- Several parameters in the `Settings` tab are given as double lists - a string of comma-separated double values. If consisting of a single value, the quote marks are not necessary.
- When clicking either `Get Defaults` button in the `Settings` tab, it is expected that if it provides a set of values for a double list parameter, it wraps them in quote marks.

### Actual behavior
- It does not wrap the values in quote marks.
- This causes the reduction to fail and throw an error if attempted, unless quote marks are re-inserted.

- The Reflectometry interface can be accessed from Interfaces->Reflectometry->ISIS Reflectometry (Polref).
- On the `Runs` tab, set `Instrument` to POLREF.
- On the `Settings` tab, for the Experiment Settings section, click the `Get Defaults` button.
- Parameters `CRho`, `CAlpha`, `CAp`, `CPp` should be filled with multiple values. Notice that none of them are wrapped in quote marks. If a reduction is attempted without them, it will fail.

PranavBahuguna:
In the `Settings` tab, the polarization constants (`CRho`, `CAlpha`, `CAp`, `CPp`) are given as a string of comma-separated values. The original problem with the 'Get Defaults' button however was that it did not wrap these values with quote marks. Without re-inserting the quote marks, an attempt at data reduction would fail. This PR fixes this by adding the quote marks back in when 'Get Defaults' is clicked.

**To test:**

- Access the Reflectometry interface from from Interfaces->Reflectometry->ISIS Reflectometry (Polref).
- On the Runs tab, set Instrument to POLREF.
- On the Settings tab, for the Experiment Settings section, click the Get Defaults button.
- Parameters CRho, CAlpha, CAp, CPp should be filled with multiple values all enclosed with quote marks.

Fixes #18208.

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/ReflSettingsTabPresenter.h
+++ b/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/ReflSettingsTabPresenter.h
@@ -65,6 +65,7 @@ private:
   void createStitchHints();
   void getExpDefaults();
   void getInstDefaults();
+  void quoteWrap(std::string &str) const;
   IAlgorithm_sptr createReductionAlg();
   Instrument_const_sptr createEmptyInstrument(std::string instName);
   std::string getTransmissionRuns() const;
--- a/MantidQt/CustomInterfaces/src/Reflectometry/ReflSettingsTabPresenter.cpp
+++ b/MantidQt/CustomInterfaces/src/Reflectometry/ReflSettingsTabPresenter.cpp
@@ -1,10 +1,10 @@
 #include "MantidQtCustomInterfaces/Reflectometry/ReflSettingsTabPresenter.h"
-#include "MantidQtCustomInterfaces/Reflectometry/IReflMainWindowPresenter.h"
-#include "MantidQtCustomInterfaces/Reflectometry/IReflSettingsTabView.h"
-#include "MantidQtMantidWidgets/AlgorithmHintStrategy.h"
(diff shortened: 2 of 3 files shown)


raquelalvarezbanos:
Works well when using polarization constants from the parameter file. However I noticed that if users enter just a set of numbers without quotation marks, like `1, 2, 3, 4`, and they click on process the reduction stops but the error message is not very meaningful (something like "Invalid pair"), so it's difficult to know what went wrong. I think we could easily make this work too just by doing something like:

```
  // Add CRho
  auto crho = m_view->getCRho();
  if (!crho.empty()) {
    if (crho.front() != '\"')
      crho = "\"" + crho;
    if (crho.back() != '\"')
      crho = crho + "\"";
    options.push_back("CRho=" + crho);
  }
```

what do you think?

PranavBahuguna:
While I agree that is a good way to go about it, what about giving a warning to the user if they don't wrap their input is incorrectly formatted?

raquelalvarezbanos:
Yes, that's the other option. I think it is the `GenericDataProcessorPresenter` the class that is throwing the error. We probably don't want to check the format there (I think it would be more complicated) but in our `ReflSettingsTabPresenter`. Have a look and feel free to do whatever you prefer (i.e. make the reduction work without quotation marks or change the error message).

PR #18213, opened 2016-12-07
Fix building with clang on Linux

rosswhitfield:
You can't assume `__clang__` only includes OSX, Mantid can also be built on Linux using clang

Currently you get these type of errors when building on Linux with clang:
```
../../Framework/Kernel/src/ArrayBoundedValidator.cpp:143:16: error: duplicate explicit instantiation of 'ArrayBoundedValidator<long>'
template class ArrayBoundedValidator<long>;
               ^
../../Framework/Kernel/src/ArrayBoundedValidator.cpp:141:16: note: previous explicit instantiation is here
template class ArrayBoundedValidator<int64_t>;
               ^
```

This changes the check for the `long` template to be `__clang__` and `__APPLE__`

**To test:**
 * If this doesn't break any of the other builds it should be good
 * If you are really motivated you can build Mantid with clang on Linux or just trust me that it now works

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


peterfpeterson:
LGTM :shipit:

PR #18247, opened 2016-12-12
Remove MatrixWorkspace::maskWorkspaceIndex

### LINKED ISSUE #18201: Remove MatrixWorkspace::maskWorkspaceIndex
`MatrixWorkspace::maskWorkspaceIndex` would need to be refactored to set masking bits via `SpectrumInfo`. However, due to the COW mechanism of `ParameterMap` and the ensuing rebuilds of `SpectrumInfo` this cannot be implemented in a decent way (solutions would be to add locking (slow) or remove the COW mechanism (memory issues for MDWorkspaces)).

This could be done cleanly once Instrument-2.0 has progressed more (when the COW mechanism is no longer required and `SpectrumInfo` will not be rebuilt anymore). Until then, refactoring client code to do this manually appears to be the easiest solution.

SimonHeybrock:
This PR removes `MatrixWorkspace::maskWorkspaceIndex`.

For the Instrument-2.0 rollout, `MatrixWorkspace::maskWorkspaceIndex` would need to be refactored to set masking bits via `SpectrumInfo`. However, due to the COW mechanism of `ParameterMap` and the ensuing rebuilds of `SpectrumInfo` this cannot be implemented in a decent way (solutions would be to add locking (slow) or remove the COW mechanism (memory issues for MDWorkspaces)).

This could be done cleanly once Instrument-2.0 has progressed more (when the COW mechanism is no longer required and `SpectrumInfo` will not be rebuilt anymore). Until then, refactoring client code to do this manually appears to be the easiest solution. 

**To test:**

Code review.

Fixes #18201.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/src/NearestNeighbours.cpp
+++ b/Framework/API/src/NearestNeighbours.cpp
@@ -1,4 +1,4 @@
-#include "MantidGeometry/Instrument/NearestNeighbours.h"
+#include "MantidAPI/NearestNeighbours.h"
 #include "MantidGeometry/Instrument.h"
 #include "MantidGeometry/Instrument/DetectorGroup.h"
 #include "MantidGeometry/Objects/BoundingBox.h"
@@ -8,7 +8,8 @@
 #include "MantidKernel/Timer.h"
 
 namespace Mantid {
-namespace Geometry {
+using namespace Geometry;
+namespace API {
 using Mantid::detid_t;
 using Kernel::V3D;
 
--- a/Framework/API/src/NearestNeighbours.cpp
+++ b/Framework/API/src/NearestNeighbours.cpp
@@ -1,4 +1,5 @@
 #include "MantidAPI/NearestNeighbours.h"
+#include "MantidAPI/SpectrumInfo.h"
 #include "MantidGeometry/Instrument.h"
 #include "MantidGeometry/Instrument/DetectorGroup.h"
 #include "MantidGeometry/Objects/BoundingBox.h"
@@ -22,9 +23,9 @@ using Kernel::V3D;
  * be ignored.
  */
 NearestNeighbours::NearestNeighbours(
-    int nNeighbours, boost::shared_ptr<const Instrument> instrument,
-    const ISpectrumDetectorMapping &spectraMap, bool ignoreMaskedDetectors)
-    : m_instrument(instrument), m_spectraMap(spectraMap),
+    int nNeighbours, const SpectrumInfo &spectrumInfo,
+    const std::vector<specnum_t> spectrumNumbers, bool ignoreMaskedDetectors)
+    : m_spectrumInfo(spectrumInfo), m_spectrumNumbers(spectrumNumbers),
       m_noNeighbours(nNeighbours), m_cutoff(-DBL_MAX), m_radius(0),
       m_bIgnoreMaskedDetectors(ignoreMaskedDetectors) {
   this->build(m_noNeighbours);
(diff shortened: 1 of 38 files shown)


Framework/API/src/NearestNeighbours.cpp
```diff
@@ -37,9 +23,9 @@ NearestNeighbours::NearestNeighbours(
  * be ignored.
  */
 NearestNeighbours::NearestNeighbours(
-    int nNeighbours, boost::shared_ptr<const Instrument> instrument,
-    const ISpectrumDetectorMapping &spectraMap, bool ignoreMaskedDetectors)
-    : m_instrument(instrument), m_spectraMap(spectraMap),
+    int nNeighbours, const SpectrumInfo &spectrumInfo,
+    const std::vector<specnum_t> spectrumNumbers, bool ignoreMaskedDetectors)
+    : m_spectrumInfo(spectrumInfo), m_spectrumNumbers(spectrumNumbers),
```
OwenArnold: @SimonHeybrock I assume thought has been given to lifetimes of `NearestNeighbour` objects as compared to `SpectrumInfos` and whether this can be abused in a way that would result in dead references.

SimonHeybrock: This is only used from within `NearestNeighbourInfo`. Unless another thread modifies the workspace at the same time (which will lead to all kinds of other problems as well), it should be fine.

SimonHeybrock: I am adapting the doxygen to mention that the class should not be used directly.

Framework/API/src/NearestNeighbours.cpp
```diff
@@ -141,18 +126,16 @@ void NearestNeighbours::build(const int noNeighbours) {
   BoundingBox bbox;
   // Base the scaling on the first detector, should be adequate but we can look
   // at this
-  IDetector_const_sptr firstDet = (*spectraDets.begin()).second;
-  firstDet->getBoundingBox(bbox);
+  const auto &firstDet = m_spectrumInfo.detector(indices.front());
```
OwenArnold: This whole first detector stuff would be a lot nicer if it was hidden and collocated somehow . i.e `m_spectrumInfo.getRepresentiviveDetector(indexes)` otherwise the client has to know too much and these rules are too scattered.

SimonHeybrock: I guess that is beyond the scope of this PR.

SimonHeybrock: Also, I do not really like such a concept because it is really dangerous. There is no way of defining such a detector in an arbitrary case (apart from the fact that we actually want to remove as much direct access as possible).

Framework/API/test/NearestNeighboursTest.h
```diff
@@ -225,52 +204,25 @@ class NearestNeighboursTestPerformance : public CxxTest::TestSuite {
...
-    // Parameterized instrument
-    Instrument_sptr m_instrument(new Instrument(instrument, pmap));
 
     // Create the NearestNeighbours object directly.
     for (size_t i = 0; i < 2000; i++) {
-      NearestNeighbours nn(8, m_instrument, spectramap);
+      NearestNeighbours nn(8, ws->spectrumInfo(), getSpectrumNumbers(*ws));
       nn.neighbours(1);
     }
   }
```
OwenArnold: Any idea how the performance tests are performing with these changes before/after?

Framework/API/src/NearestNeighbours.cpp
OwenArnold: Ok. I agree someone would probably have to go out of their way to do dangerous things with this.

Framework/API/test/NearestNeighboursTest.h
SimonHeybrock: No, but this test seems a bit useless anyway. No real code would create the `NearestNeighbours` object and use it to get neighbours only for a single detector.

Framework/API/src/NearestNeighbours.cpp
OwenArnold: I think it actually beyond the scope of this PR, however I've seen a lot of code that uses the first detector to get information on the group. I think that's another discussion to have at a later point.

OwenArnold:
@SimonHeybrock. Code looks pretty good to me. If the tests all pass I think this would be good to go.

Framework/API/test/NearestNeighboursTest.h
```diff
@@ -1,71 +1,78 @@
...
+  auto ws = boost::make_shared<WorkspaceTester>();
+  ws->initialize(end - start + 1, 2, 1);
+  for (specnum_t i = start; i <= end; ++i) {
+    ws->getSpectrum(i - start).setSpectrumNo(i);
+    ws->getSpectrum(i - start).setDetectorID(i);
+  }
+  return ws;
+}
+
+std::vector<specnum_t> getSpectrumNumbers(const MatrixWorkspace &workspace) {
```
SimonHeybrock: 👍

SimonHeybrock:
@OwenArnold All green now.

OwenArnold:
:shipit:

PR #18251, opened 2016-12-12
Allow Angle argument to FlatPlate dictionary for SetSample

### LINKED ISSUE #17194: Allow for rotation in flat plate in SetSample
[SetSample](http://docs.mantidproject.org/nightly/algorithms/SetSample-v1.html#id5) allows for a flat plate geometry but it is currently constrained so that it's surface normal is parallel to the beam.

Add an optional `Angle` parameter to the `FlatPlate` configuration defining it as the angle, in degrees, between the perpendicular and the beam, with the default=0.0 in line with the current behaviour.

martyngigg:
A flat plate sample shape can now be rotated about its own center using `SetSample`

**To test:**

The following script produces a plot of the attenutation factor at a fixed wavelength versus the angle of rotation of the flat plate.

```python
from matplotlib import pyplot as plt
import numpy as np

data = CropWorkspace(data, StartWorkspaceIndex=0, EndWorkspaceIndex=0)
SetSample(data,  Material={'ChemicalFormula': 'V', 'SampleNumberDensity': 0.073})

# geometry in cm
sample_height = 2.0
sample_width = 1.0
sample_thick = 0.1

# simulation options
nlambda_cyl = 20
nlambda_mc = 20
nevents = 20000
interpolation = 'Linear'

# Compute absorption for 
angles = np.arange(91., step=5.)
lambda_index = 50
abs_factors = []
for angle in angles:
    SetSample(data, Geometry={'Shape': 'FlatPlate', 'Height': sample_height, 'Width': sample_width,
                                'Thick': sample_thick, 'Center': [0.,0.,0.], 'Angle': float(angle)})
    abs = MonteCarloAbsorption(data,
                                                  NumberOfWavelengthPoints=nlambda_mc,
                                                  EventsPerPoint=nevents)
    abs_factors.append(abs.readY(0)[lambda_index])
    DeleteWorkspace(abs)
#endfor

# Plot absorption factor for single wavelength against the angle of rotation
plt.plot(angles, abs_factors)
plt.show()
```

As the sample rotates the width and thickness dimensions swap and the attenuation increases due to the extra distance travelled. This manifests itself as a reduction in the fraction of the beam that makes it through:

flate_plate_abs_vs_angle_step_5

Fixes #17194 

[RELEASE NOTES](https://github.com/mantidproject/mantid/commit/a6724b0f17c7d22e17004f701c47db2a10a8dbeb)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/DataHandling/inc/MantidDataHandling/SetSample.h
```diff
@@ -49,13 +50,18 @@ class MANTID_DATAHANDLING_DLL SetSample final : public API::Algorithm {
 
   const Geometry::SampleEnvironment *
   setSampleEnvironment(API::MatrixWorkspace_sptr &workspace,
-                       const Kernel::PropertyManager &args);
+                       const Kernel::PropertyManager_sptr &args);
```
soininen: I'm wondering why pass shared pointers as references? Also, would it be possible to use `PropertyManager_const_sptr` here? Only some getters are used in the method as far as I can see. Same for the `setSampleShape` method and maybe others in this class.

Framework/DataHandling/src/SetSample.cpp
```diff
@@ -258,10 +243,13 @@ void SetSample::setSampleShape(API::MatrixWorkspace_sptr &workspace,
   */
 
   // Try known shapes or CSG first if supplied
-  const auto xml = tryCreateXMLFromArgsOnly(args);
-  if (!xml.empty()) {
-    runSetSampleShape(workspace, xml);
-    return;
+  if (args) {
+    auto refFrame = workspace->getInstrument()->getReferenceFrame();
```
soininen: I think `refFrame` could be delcared `const`.

docs/source/algorithms/SetSample-v1.rst
```diff
@@ -71,14 +71,20 @@ For defining the full shape a key called ``Shape`` specifying the desired shape
 expected along with additional keys specifying the values (all values are assumed to
```
soininen: While at it, you could perhaps fix the typo in the first sentence of the **Environment** subsection as well: 'enviorment' -> 'environment'.

docs/source/algorithms/SetSample-v1.rst
```diff
@@ -132,11 +138,11 @@ The following example uses a test file called ``CRYO-01.xml`` in the
 
    # A fake host workspace, replace this with your real one.
    ws = CreateSampleWorkspace()
-   # Use geometry from environment but set differnet height for sample
+   # Use geometry from environment but set different height for sample
    SetSample(ws, Environment={'Name': 'CRYO-01', 'Container': '8mm'},
              Geometry={'Shape': 'HollowCylinder', 'Height': 4.0,
                        'InnerRadius': 0.8, 'OuterRadius': 1.0,
-                       'Center': [0.,0.,0.], 'Axis':1},
```
soininen: There is a space missing between the colon and '1'.

soininen (changes requested):

soininen:
Are there any plans to extend `SetSample::validateInputs()`? The algorithm seems to happily accept negative sample dimensions.

Framework/DataHandling/inc/MantidDataHandling/SetSample.h
martyngigg: Passing the `shared_ptr`'s by const reference here just avoids increasing the reference count: http://stackoverflow.com/a/37610721. I know I have a reference to keep the object alive at the top level as this is all private. I would usually prefer a plain `const PropertyManager&` but it needs to be a `shared_ptr` to be fed into the child algorithm.  

You're right I can probably use `PropertyManager_const_sptr` here.

Framework/DataHandling/src/SetSample.cpp
martyngigg: :+1:

docs/source/algorithms/SetSample-v1.rst
martyngigg: :+1:

martyngigg:
You're right, I should provide better checking for the basic cases such as negative dimensions. I'll update the code.

martyngigg:
@soininen I've added some changes after your feedback.

Framework/DataHandling/src/SetSample.cpp
```diff
@@ -94,28 +143,61 @@ const std::string SetSample::summary() const {
...
   std::map<std::string, std::string> errors;
 
+  auto existsAndNotEmptyString =
+      [](const PropertyManager &pm, const std::string &name) {
+        if (pm.existsProperty(name)) {
+          const auto value = pm.getPropertyValue(name);
+          if (value.empty())
+            return false;
+          else
+            return true;
```
soininen: Couldn't you just `return !value.empty();`?

Framework/DataHandling/src/SetSample.cpp
```diff
@@ -94,28 +143,61 @@ const std::string SetSample::summary() const {
...
-      std::string name = environArgs->getPropertyValue("Container");
-      if (name.empty()) {
-        errors["Environment"] = "Environment 'Can' flag is an empty string!";
+    if (!existsAndNotEmptyString(*environArgs, SEArgs::CONTAINER)) {
+      errors[PropertyNames::ENVIRONMENT] =
+          "Environment flags require a non-empty 'Container' entry.";
+    }
+  }
+
+  //  // Validate as much of the shape information as possible
```
soininen: A bit nit-picking, but perhaps the comment needs some cleanup.

soininen (changes requested):

Framework/DataHandling/src/SetSample.cpp
martyngigg: Good point :man_facepalming:

soininen (approved):

soininen:
The new angle property seems to work as expected. The test script produces the expected cyclic pattern when the sample is rotated around [-180, 450] degrees. The input properties are validated more rigorously than before, and there are new units tests in place. I think we are done here :shipit:

PR #18276, opened 2016-12-14
Track DetectorInfo in ExperimentInfo::getInstrument and setInstrument

### LINKED ISSUE #18268: Track DetectorInfo in ExperimentInfo::getInstrument() and setInstrument().
Probably solve this by adding a reference/pointer to `DetectorInfo` to `Instrument` that is initialized in `ExperimentInfo::getInstrument()` and used to assign the `DetectorInfo` in `ExperimentInfo::setInstrument()`.

SimonHeybrock:
For the Instrument-2.0 rollout and in particular scanning support for instruments we will soon move data from the `ParameterMap` into the new `DetectorInfo` class. This is not part of the old instrument since the incremental way of adding components and marking them as detectors would cause huge amounts of reallocations and updates of the `DetectorInfo` object. It is therefore (for the time being) owned by experiment info, but still we need to be able to move it around as part of the instrument.

This PR adds this functionality. Currently there is no data in `Beamline::DetectorInfo` apart from the size (number of detectors). A size check was added in `ExperimentInfo::setInstrument` and `ExperimentInfo::getInstrument` to ensure consistency. This required some minor changes to algorithms/tests that set an instrument for a workspace and add detectors *afterwards*. This is no longer possible now, and does actually seem like a very reasonable limitation.

**To test:**

Code review.

Fixes #18268.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/src/ExperimentInfo.cpp
+++ b/Framework/API/src/ExperimentInfo.cpp
@@ -155,10 +155,17 @@ void ExperimentInfo::setInstrument(const Instrument_const_sptr &instr) {
     sptr_instrument = instr;
     m_parmap = boost::make_shared<ParameterMap>();
   }
-  // TODO Add Beamline::DetectorInfo reference to Instrument and assign that
-  // here. For now we just create a new DetectorInfo of the correct size:
-  m_detectorInfo = Kernel::make_unique<Beamline::DetectorInfo>(
-      sptr_instrument->getNumberDetectors());
+  const auto *detInfo = sptr_instrument->detectorInfo();
+  const auto numDets = sptr_instrument->getNumberDetectors();
+  if (detInfo) {
+    if (numDets != detInfo->size())
+      throw std::runtime_error("ExperimentInfo::setInstrument: size mismatch "
+                               "between DetectorInfo and number of detectors "
+                               "in instrument");
+    *m_detectorInfo = *detInfo;
+  } else {
+    m_detectorInfo = Kernel::make_unique<Beamline::DetectorInfo>(numDets);
+  }
 }
 
 /** Get a shared pointer to the parametrized instrument associated with this
@@ -167,8 +174,14 @@ void ExperimentInfo::setInstrument(const Instrument_const_sptr &instr) {
 *  @return The instrument class
 */
 Instrument_const_sptr ExperimentInfo::getInstrument() const {
-  return Geometry::ParComponentFactory::createInstrument(sptr_instrument,
-                                                         m_parmap);
+  if (sptr_instrument->getNumberDetectors() != m_detectorInfo->size())
+    throw std::runtime_error("ExperimentInfo::getInstrument: size mismatch "
+                             "between DetectorInfo and number of detectors in "
+                             "instrument");
+  auto instrument = Geometry::ParComponentFactory::createInstrument(
+      sptr_instrument, m_parmap);
+  instrument->setDetectorInfo(m_detectorInfo.get());
+  return instrument;
 }
 
 /**  Returns a new copy of the instrument parameters
--- a/Framework/API/src/ExperimentInfo.cpp
+++ b/Framework/API/src/ExperimentInfo.cpp
@@ -13,12 +13,15 @@
 #include "MantidGeometry/Instrument/ParComponentFactory.h"
 #include "MantidGeometry/Instrument/XMLInstrumentParameter.h"
 
+#include "MantidBeamline/DetectorInfo.h"
+
 #include "MantidKernel/ConfigService.h"
 #include "MantidKernel/DateAndTime.h"
 #include "MantidKernel/InstrumentInfo.h"
 #include "MantidKernel/Property.h"
 #include "MantidKernel/Strings.h"
 #include "MantidKernel/StringTokenizer.h"
+#include "MantidKernel/make_unique.h"
 
 #include <boost/algorithm/string.hpp>
 #include <boost/make_shared.hpp>
@@ -47,7 +50,8 @@ Kernel::Logger g_log("ExperimentInfo");
 ExperimentInfo::ExperimentInfo()
     : m_moderatorModel(), m_choppers(), m_sample(new Sample()),
       m_run(new Run()), m_parmap(new ParameterMap()),
-      sptr_instrument(new Instrument()) {}
+      sptr_instrument(new Instrument()),
+      m_detectorInfo(Kernel::make_unique<Beamline::DetectorInfo>(0)) {}
 
 /**
  * Constructs the object from a copy if the input. This leaves the new mutex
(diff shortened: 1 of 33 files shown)


Framework/API/src/ExperimentInfo.cpp
```diff
@@ -889,12 +912,13 @@ ExperimentInfo::getInstrumentFilename(const std::string &instrumentName,
...
+  if (!m_detectorInfoWrapper) {
     std::lock_guard<std::mutex> lock{m_detectorInfoMutex};
-    if (!m_detectorInfo)
-      m_detectorInfo = Kernel::make_unique<DetectorInfo>(getInstrument());
+    if (!m_detectorInfoWrapper)
+      m_detectorInfoWrapper =
+          Kernel::make_unique<DetectorInfo>(getInstrument());
   }
-  return *m_detectorInfo;
+  return *m_detectorInfoWrapper;
```
OwenArnold: Method name and what it returns do not match well. Consider renaming method.

Framework/Geometry/src/Instrument.cpp
```diff
@@ -1235,5 +1237,15 @@ Instrument::ContainsState Instrument::containsRectDetectors() const {
...
 } // containsRectDetectors
 
+/// Only for use by ExperimentInfo. Returns the pointer to the DetectorInfo.
+const Beamline::DetectorInfo *Instrument::detectorInfo() const {
+  return m_detectorInfo;
+}
+
+/// Only for use by ExperimentInfo. Sets the pointer to the DetectorInfo.
+void Instrument::setDetectorInfo(const Beamline::DetectorInfo *detectorInfo) {
+  m_detectorInfo = detectorInfo;
```
OwenArnold: This looks like a resource leak waiting to happen. Is this a temporary change? I think this is quite unsafe as it stands.

Framework/Geometry/inc/MantidGeometry/Instrument.h
```diff
@@ -327,6 +327,10 @@ class MANTID_GEOMETRY_DLL Instrument : public CompAssembly {
 
   /// Pointer to the reference frame object.
   boost::shared_ptr<ReferenceFrame> m_referenceFrame;
+
+  /// Pointer to the DetectorInfo object. NULL unless the instrument is
+  /// associated with an ExperimentInfo object.
+  const Beamline::DetectorInfo *m_detectorInfo{nullptr};
```
OwenArnold: I'd be interested to see what else is going to go into this `DetectorInfo` type

Framework/API/src/ExperimentInfo.cpp
SimonHeybrock: This is absolutely intentional. The plan is to step by step move things from `API::DetectorInfo` into `Beamline::DetectorInfo`. The interfaces of the two classes will be kept compatible as far as possible. Once the move is complete this method will be changed to return `Beamline::DetectorInfo`, with little to no changes required in client code.

Framework/Geometry/inc/MantidGeometry/Instrument.h
SimonHeybrock: Eventually everything: `API::DetectorInfo` will be removed. As first steps it will be masking, and then position information.

Framework/API/src/ExperimentInfo.cpp
OwenArnold: Ok. I'll have to read a bit more to establish the interplay with all the existing types.

Framework/Geometry/src/Instrument.cpp
```diff
@@ -1235,5 +1237,15 @@ Instrument::ContainsState Instrument::containsRectDetectors() const {
 
 } // containsRectDetectors
 
+/// Only for use by ExperimentInfo. Returns the pointer to the DetectorInfo.
+const Beamline::DetectorInfo *Instrument::detectorInfo() const {
+  return m_detectorInfo;
```
SimonHeybrock: Yes, for instruments that were created directly, i.e., do not come from `ExperimentInfo::getInstrument()`.

Framework/Geometry/src/Instrument.cpp
SimonHeybrock: Not a resource leak, the `DetectorInfo` is owned by an `ExperimentInfo`. It definitely is a temporary change, but it will take a while until the Instrument-2.0 implementation has progressed far enough to remove it.

You are right that it is unsafe: The pointer could in principle get stale if the workspace owning the `DetectorInfo` is deleted. I believe that this is unlikely given the way Mantid deals with things: (1) We never have a stand-alone instrument, apart from short-lived ones within an algorithm/test. (2) Workspaces are rarely explicitly deleted but just go out of scope when an algorithm/test ends. Do you think the risk is too high?

OwenArnold: Ideally it would be nice to have something with better safety, however, there isn't a good way of dealing with this apart from a copy or a smart_ptr. Both could introduce performance issues. I'm not sure if they would be significant.

Framework/Geometry/src/Instrument.cpp
OwenArnold: Could you add a method to allow the null check? I think `Instrument` had an `isBase` or something like that.

SimonHeybrock: Well, it is a pointer so the client (`ExperimentInfo`) can just check if it is null (https://github.com/mantidproject/mantid/pull/18276/files#diff-f92af670836a7a4178efefc827b27f95R159). The null check would be useful if this would return a reference, I guess?

Framework/Geometry/src/Instrument.cpp
SimonHeybrock: I think making it a `shared_ptr` would not be a big overhead. I'll have a look if that would work.

OwenArnold: Or maybe a `weak_ptr` if "ownership" can't be guaranteed.

OwenArnold: But i think in your case a `share_ptr` does make more sense

OwenArnold:
Looks good. @ianbush Any comments?

OwenArnold:
:shipit:

SimonHeybrock:
@OwenArnold I found a small issue with this. It does not actually make a difference at this point since `Beamline::DetectorInfo` is empty, but goes wrong when we populate it. I pushed a (really small) fix, have a look if this is still ready for shipping.

ianbush:
I have only had a brief look through this but it all looks good to me - and it nicely shows how the `DetectorInfo` class will work for real.

OwenArnold:
:shipit:

ianbush:
Looks good :+1: . Once `Beamline::DetectorInfo` is being used directly we should make sure that the description of `MantidAPI/DetectorInfo.h` is updated appropriately, so anyone not following this closely understands why there will be two `DetectorInfo` classes for some time.

PR #18282, opened 2016-12-14
Indirect Absorption - Update Sample Geometry

### LINKED ISSUE #16843: Indirect Corrections - Use new sample geometry in Absorption algorithms
### Expected behavior

Indirect Absorption algorithms:
- `IndirectCylinderAbsorption`
- `IndirectAnnulusAbsorption`
- `IndirectFlatPlateAbsorption`

should use the new sample geometry. These should be able to use the same tests that are currently being used (with some minor adaptation)
### Actual behavior

Currently still using old sample geometry

N/A
### Platforms affected

N/A

Spencer has suggested that these be implemented as version 2 of each of these algorithms for the time being to ensure that a working version is retained while the new corrections are tested.

### LINKED ISSUE #18214: Indirect Corrections - Absorption always expects a container
### Expected behavior
The Indirect>Corrections>Absorption interface has an check box option to use the container when calculating absorption corrections. 
### Actual behavior
The inputs for the container dimensions and subsequent validations are used whether or not the option to include the container is checked.

louisemccann:
The Indirect Absorption Interface has been updated to calculate the absorption corrections using the Monte Carlo method. The algorithms have been refactored to use this method and implemented as version 2 of each respective absorption algorithm.

**To test:**

`Build\ExternalData\Testing\Data\UnitTest`

Sample: `irs26176_graphite002_red`
Container: `irs26173_graphite002_red`

Interfaces>Indirect>Corrections>Absorption

Use 1cm or 2 cm for sample/container dimensions, given that `container inner < sample inner < sample outer < container outer`

Sample: `H2_O` Mass Density: `1` /Number Density: `0.03`
Container: `V` Mass Density: `6` /Number Density: `0.07`
Test every possible combination of inputs for container options and FlatPlate/Annulus/Cylinder geometry.
If algorithm runs slowly reduce the number of events.

Fixes #16843
Fixes #18214 

[Release Notes](https://github.com/mantidproject/mantid/blob/c647a6efe1d4785475b4620d93d9b6669835ea45/docs/source/release/v3.9.0/indirect_inelastic.rst)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


SpencerHowells:
Sample angle default should be 0.0. This corresponds to sample perpendicular to beam - the fixed value in the FlatPlateAborption algo.

louisemccann:
This PR will also close #18214

gvardany (changes requested):
Could you please update also the GUI tab documentation?
The v2 of the algorithms need their own documentation `-v2.rst` pages.
Could you add a line `Fixes #18214 ` in main PR message, so the second issue will also close itself.

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/IndirectFlatPlateAbsorption2.py
```diff
@@ -0,0 +1,436 @@
...
+class IndirectFlatPlateAbsorption(DataProcessorAlgorithm):
+    # Sample variables
+    _sample_ws_name = None
+    _sample_chemical_formula = None
+    _sample_density_type = None
+    _sample_density = None
+    _sample_height = None
+    _sample_width = None
+    _sample_thickness = None
+    _sample_angle=None
```
gvardany: spaces around `=`

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/IndirectFlatPlateAbsorption2.py
```diff
@@ -0,0 +1,436 @@
...
+        self._abs_ws = self.getPropertyValue('CorrectionsWorkspace')
+        if self._abs_ws == '':
+            self._ass_ws = '__ass'
+            self._acc_ws = '__acc'
+        else:
+            self._ass_ws = self._abs_ws + '_ass'
+            self._acc_ws = self._abs_ws + '_acc'
+
+        # Get beam size defaults
+        # Hard-coded because the ISIS parameters weren't being picked up
```
gvardany: What is the problem here, that the parameters cannot be retrieved? What about instrument other than IRIS/OSIRIS?

martyngigg:
Does it make sense to include this in 3.9? I see the "in progress" label has been added but it seems that @gvardany's comments have been addressed?

It would be good to get the `MonteCarloAbsorption` out in the wild.

gvardany:
Actually I had not reviewed fully yet, but I can do now if this is a priority. Those were just some general remarks.

martyngigg:
@gvardany If you have time that would be great.

gvardany:
I have found few issues so far below, I am still testing with various other options though:

1. In this configuration I still get:
screen shot 2017-01-25 at 15 04 45

> Instrument has no default beam size; will use inputs
> MonteCarloAbsorption: Object::generatePointInObject() - Unable to generate point in object after 500 attempts
> Error in execution of algorithm MonteCarloAbsorption:
> MonteCarloAbsorption: error (see log)
> Error in execution of algorithm IndirectFlatPlateAbsorption
> MonteCarloAbsorption: error (see log)
>   at line 165 in '/Users/vardanyan/mantid/Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/IndirectFlatPlateAbsorption2.py'
>   caused by line 926 in 'mantid/simpleapi.py'
> Got error from algorithm "IndirectFlatPlateAbsorption"
> Stopping batch algorithm because of execution error


2. When I increase the sample width to 0.2, the second problem goes away, but this error remains

> Instrument has no default beam size; will use inputs

3. In this configuration, I get error on can radius, even if use can is unchecked. 
screen shot 2017-01-25 at 15 11 08

>Instrument has no default beam size; will use inputs
>Invalid value for CanRadius: Must be greater than SampleRadius
>Unknown error starting next batch algorithm
>Got error from algorithm "IndirectCylinderAbsorption"
>Stopping batch algorithm because of execution error

4. In any case when `NumberOfWavelengthPoints < 4`, I get e.g.:

> MonteCarloAbsorption: interpolate - CSpline requires 4 points but only 3 were found.

I thinks this is more of an issue in MonteCarloAbsorption, that might need to fall back to linear interpolation if Spline is requested but not enough points are given. Note that eventually we want to use this even for `NumberOfWavelengthPoints = 1`.

5. In `validateInputs` of FlatPlate, the can chemical formula is not checked.

6. In `_test_workspaces` in unit tests I would assert also at least the block size and the number of histograms in output and corrections workspaces.

7. `_getEFixed` seems identical across all the 3, maybe it is worth considering to move it to IndirectCommons ? I am not sure however.

8. Ideally the input validation should also be handled in the GUI, to catch the problems before executing the algorithms.

9. In the documentations I would make explicit that these new versions use MonteCarloAbsorption, with a link to it. And optionally for the older versions make a link to the non-MC algorithms used previously. 

10. Finally, the flowcharts should also be adjusted to accommodate the new change.

SpencerHowells:
4. I've spoken to Martyn and this should be changed in MonteCarloAbsorption.
7. We have been deliberately moving _getEFixed from IndirectCommon to the algo  so that it is properly tested instead of being in a script file.
8. I've checked on this and input validation should always be at the lowest level. Users may also want to use the algo on its own or in a script file.

gvardany:
Assuming you meant 4, 7, 8
4. OK
7. OK
8. I did not mean that they should be removed from the algorithms. They should be both in the algorithms, and in GUI.

louisemccann:
@gvardany I agree with you with regards to documentation - however  it could also be done in a new PR for documentation updates as maintenance.
 I think that updating the unit tests to include blocksize etc, although a valid point, is not vital for this release and would also be better suited for a separate PR so that I can update any other Indirect unit tests that don't have this check.

gvardany:
Yes I agree these are not critical, they can go to a separate ticket.

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/IndirectAnnulusAbsorption2.py
```diff
@@ -0,0 +1,449 @@
...
+            self._ass_ws = '__ass'
+            self._acc_ws = '__acc'
+        else:
+            self._ass_ws = self._abs_ws + '_ass'
+            self._acc_ws = self._abs_ws + '_acc'
+
+        # Get beam size defaults
+        inst = mtd[self._sample_ws_name].getInstrument()
+        has_beam = inst.hasParameter('Workflow.beam-height')
+        default = self.getPropertyValue('DefaultBeamSize')
```
gvardany: getPropertyValue returns string representation, so both true and false will count as true.
You must use getProperty().value to get the right behaviour, the same in other algorithms.

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/IndirectAnnulusAbsorption2.py
```diff
@@ -0,0 +1,449 @@
...
+
+        if default and (has_beam is False):
```
gvardany: `if default and not has_beam`

gvardany:
When I load the `IRS26176.RAW` file, `.getInstrument().hasParameter('Workflow.beam-width')` returns `True`, while for `irs26176_graphite002_red` reduced file it is `False`.  Is there any explanation to this?

SpencerHowells:
The Reduction routine loads the parameters from the IPF. The raw file does not have this info.

SpencerHowells:
I've just re-read the comment & the info is the wrong way round! I don't understand it.

SpencerHowells:
Have you seen Issue 18546 ? Do we have this problem?

gvardany:
Regarding the parameters, it seems like all the parameters starting from `cm-1-convert-choice` in the master parameter file are absent from the existing reduced file `irs26176_graphite002_red`. However when I reduce the raw file anew, the output has them. Can it be that these test files have been generated before these parameters were added or something alike?

SpencerHowells:
I think you've found the answer.

gvardany:
Regarding the issue #18546 ,  I have not encountered it, because on my system the decimal delimiter is `.`. But I am not surprised about it, I think there could be few more places (at least one on my side), e.g. whenever a `,` separated array of doubles are passed from GUI to backend, this may potentially break if the system uses `,` as the decimal point. Maybe it is worth to do a detailed walk-through the interfaces (not only indirect) with both system locale settings.

SpencerHowells:
It was a coincidence because I noticed on your screenshots of the GUI that it is using ,

gvardany:
You are right, I was confused about the issue message 😃 . This issue is about `Interfaces -> General -> Sample Transmission Calculator`. The problem there is when the `Rebin` parameter string is formed from the Qt elements.

SpencerHowells:
That interface was written by one of us! So we may have others with the same problem.

gvardany:
I have one more suggestion to make the view more intuitive. This can also go to a follow-up ticket.

screen shot 2017-01-26 at 09 53 05

gvardany:
Regarding 1: Indeed, when `x-y` profile of the sample seen by the beam is ~O(100) smaller compared to the lateral beam profile, the MonteCarlo generator often fails to generate a valid track, since it has an inherent limit of `500` attempts. Of course this limit value is arguable, but there has to be one. The only way I see is to catch the error in MonteCarloAbsorption, if possible, and just better formulate the error message. One might also try to compare these profiles beforehand and warn that this might happen. I think this is beyond the scope of this PR anyway.

gvardany:
One more note on the GUI: Please make explicit that sample angle is in degrees. Also maybe add a tooltip, that it is the angle from x-axis, so 0 means it is perpendicular to the beam.

gvardany:
I have created the follow up tickets for v3.10 : #18576, #18577, #18579 . This one can be :shipit: once tests pass.

gvardany (approved):

martyngigg:
Failing test is some intermittent problem with the matplotlib test in MantidPlot and unrelated.

PR #18284, opened 2016-12-14
Remove threaded use of (Spectrum|Detector)Info::setMasked.

### LINKED ISSUE #18281: Do not use threading with SpectrumInfo::setMasked and DetectorInfo::setMasked.
Masking information will be stored in `DetectorInfo`. Write access will not be thread-safe (at least in the initial implementation, but it is likely that threading would not actually provide a performance benefit -- it may actually be a drawback).

Go through client code using these methods and refactor where necessary.

SimonHeybrock:
Masking information will be stored in `DetectorInfo`. Write access will not be thread-safe (at least in the initial implementation, but it is likely that threading would not actually provide a performance benefit -- it may actually be a drawback).

Very few cases did actually require changes. I chose the simplest solution (adding a critical region) in most cases. I cannot exclude that this may cause slowdowns, but typically the critical region is in an `if` statement, so it will not be hit a lot by default. If the performance tests show issues (`BinaryOperation` may be a candidate) we can try to find a better solution.

Note that `DetectorInfo` will store a vector of mask flags. As long as the flag is a `bool` there will always be threading issues (writes to different elements of `std::vector<bool>` are not thread safe). Eventually we might consider moving to another flag type to get around this issue, but in any case this is beyond the scope of this PR.

**To test:**

- Code review.
- @reviewers: For this PR it is NOT sufficient to look at my changes. Rather, all uses of `SpectrumInfo::setMasked` and `DetectorInfo::setMasked` should be checked, to make sure that I did not overlook any cases that require changes.

Fixes #18281 .

No release notes, internal change.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/inc/MantidAPI/DetectorInfo.h
+++ b/Framework/API/inc/MantidAPI/DetectorInfo.h
@@ -30,8 +30,10 @@ class SpectrumInfo;
   DetectorInfo provides easy access to commonly used parameters of individual
   detectors, such as mask and monitor flags, L1, L2, and 2-theta.
 
-  This class is thread safe with OpenMP BUT NOT WITH ANY OTHER THREADING LIBRARY
-  such as Poco threads or Intel TBB.
+  This class is thread safe for read operations (const access) with OpenMP BUT
+  NOT WITH ANY OTHER THREADING LIBRARY such as Poco threads or Intel TBB. There
+  are no thread-safety guarantees for write operations (non-const access). Reads
+  concurrent with writes or concurrent writes are not allowed.
 
 
   @author Simon Heybrock
--- a/Framework/API/inc/MantidAPI/SpectrumInfo.h
+++ b/Framework/API/inc/MantidAPI/SpectrumInfo.h
@@ -29,8 +29,10 @@ class MatrixWorkspace;
   spectra (which may correspond to one or more detectors), such as mask and
   monitor flags, L1, L2, and 2-theta.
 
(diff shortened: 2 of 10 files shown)


OwenArnold:
Looking through `DetectorInfo` I find this comment in the class doxygen
```
This class is thread safe with OpenMP BUT NOT WITH ANY OTHER THREADING LIBRARY
  such as Poco threads or Intel TBB.
```

OwenArnold:
@SimonHeybrock There's a comment in `ConvertUnits` about making `convertViaTOF` run in parallel (line 558) but as far as I can tell it's actually not running in parallel.

OwenArnold:
@SimonHeybrock. I manually checked the codebase of this branch - as far as I can tell, you missed nothing. I didn't check all the `..Test.h` files. I assume that no one would be thoughtless enough to mix threading up in unit tests with masking. So everything looks good.

I am a little concerned that someone might try to parallelise some of this at a later date. Should we add comments to all usage about thread safety, or do you think it's enough to see that it's clearly a write operation and developers would be cautious?

SimonHeybrock:
@OwenArnold I agree with all your points, this update should address them.

OwenArnold:
@SimonHeybrock. If we get all green. I think this will be good to merge.

SimonHeybrock:
@OwenArnold All green. :)

OwenArnold:
:shipit:

PR #18288, opened 2016-12-14
Remove unused ThresholdRange class from VatesAPI

quantumsteve:
Description of work.

This removes `ThresholdRange` and derived class from the VatesAPI. I *think* only `NoThresholdRange` and `IgnoreZerosThresholdRange` are being used and the min and max values they calculate are ignored. The `ThresholdRangeWidget` is not part of the VSI. 

This class slows down creation of a `vtkDataSet` by adding a virtual function call in a tight loop and making it considerably more difficult to parallelize that loop. 

**To test:**

1. Verify my logic above for removing `ThresholdRange`. Review changes verifying this doesn't significantly change the user experience.
2. Load a MDHistoWorkspace, open it in the VSI, change views and verify the image is unchanged. 
3. Load a MDEventWorkspace, open it in the VSI, change views and verify the image is unchanged.

There is no GitHub issue 

*Does not need to be in the release notes.* This is am implementation detail and mentioning it in the release notes is more likely to confuse rather than help users.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Vates/VatesAPI/src/vtkMDHexFactory.cpp
```diff
@@ -120,8 +117,7 @@ void vtkMDHexFactory::doCreate(
       API::IMDNode *box = boxes[i];
       Mantid::signal_t signal_normalized = (box->*normFunction)();
 
-      if (std::isfinite(signal_normalized) &&
-          m_thresholdRange->inRange(signal_normalized)) {
```
OwenArnold: @quantumsteve Do you have any idea what the runtime improvement is as a result of removing this?

OwenArnold:
The way the removal has been handled looks ok. I think you are rightly now removing obsolete code. It would be interesting to know that the improvements might be. Anything significant should be recorded against the release notes.

Vates/VatesAPI/src/vtkMDHexFactory.cpp
quantumsteve: I'll time a few workspaces and see what the difference is.

Generating a vtkUnstructuredGrid from a MDEventWorkspace is fast enough that it would take a lot of boxes to make a difference. I'm more optimistic about vtkMDHistoHexFactory, especially if we can later parallelize some of those loops.

quantumsteve: There is a small improvement (4.26s vs 4.82s) generating the cell blanking array in a vtkStructuredGrid. It's a step in the right direction, but it's a fraction of the time spent generating and rendering the `vtkStructuredGrid` and overall not something one would notice without benchmarking.

I tried doing the same with a MDEventWorkspace, but didn't notice any significant improvement in the splatterplot view.

quantumsteve: Once this and PR #18300 are in master, I'll try parallelizing the for loops in `vtkMDHistoHexFactory` and (if successful) will add that to the release notes.

AntonPiccardoSelg:
The changes look good to me. It is good to get rid of that code. I loaded MDHisto and MDEvent data and tested several configurations. I compared it with loading data into the M3.8 release. Subjectively I cannot see any degradation in user experience. 

One thing I have noticed is that the `Threshold` filter in the SplatterPlot mode does not seem to make any difference. Any idea why that is?

quantumsteve:
@AntonPiccardoSelg Is the threshold filter not working or the underlying source not being hidden

AntonPiccardoSelg:
@quantumsteve No idea. The visible workspace does not change. But it appears to be also present in M3.8

quantumsteve:
I had to hide the source before the output from the threshold filter is visible. Hiding the source too early though might cause some users to fear their data disappeared. Is it possible to hide the source at the same time the filter is **applied**?

In either case, I think the issue you encountered, while important, is separate from this pull request.

AntonPiccardoSelg:
Normally when you add a filter to the pipeline that should set the source invisible automatically. Saying that, my experience with the visibility feature in PV was not too nice. Can you hide it after the filter has been applied?

I agree that the issue is not related to the work of this PR.

All looks fine to me. I am happy to :shipit: . 
@quantumsteve  There is a merge conflict

PR #18296, opened 2016-12-15
Fix performance regression in sum spectra

DavidFair:
Description of work.
This PR fixes a performance regression in SumSpectra seen [here](http://builds.mantidproject.org/view/Tests/job/master_performancetests/Master_branch_performance_tests/AlgorithmsTest.SumSpectraTestPerformance.testExec2D.htm) from #18128 pointed out by @LamarMoore (thanks).

The regression was caused by moving an if statement into a loop, this has been moved back out of the loop so it is only evaluated once. There is also an impact from using `std::isNormal` however that is unavoidable. 

Additionally the vector access operator was the hottest call path. This has been resolved by storing the variable and squaring that instead. 

On MSVC - Win 10:
```
Before Changes: 2.75-2.80s
After Changes: 1.39-1.41s
```

**To test:**
Run `SumSpectraTestPerformance` before and after the changes. The time taken (found in the relevant XML file) should reflect the speed up.

No associated issue 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/src/SumSpectra.cpp
+++ b/Framework/Algorithms/src/SumSpectra.cpp
@@ -295,23 +295,24 @@ void SumSpectra::doWorkspace2D(ISpectrum &outSpec, Progress &progress,
 
     const auto &YValues = localworkspace->y(wsIndex);
     const auto &YErrors = localworkspace->e(wsIndex);
-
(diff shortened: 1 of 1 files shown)


LamarMoore:
:shipit:

PR #18300, opened 2016-12-16
Make vtkMDHWSignalArray accessible from multiple threads.

quantumsteve:
Description of work.

This modifies `vtkMDHWSignalArray` to retrieve values from the `MDHistoWorkspace` directly instead of using the the (not threadsafe) `MDHistoWorkspaceIterator`. This, along with PR #18288 will make it possible to perform [this loop](https://github.com/mantidproject/mantid/blob/793647337b1460588c05824536bc8588ebff70b0/Vates/VatesAPI/src/vtkMDHistoHexFactory.cpp#L136) in parallel.

**To test:**

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Vates/VatesAPI/inc/MantidVatesAPI/vtkMDHWSignalArray.h
+++ b/Vates/VatesAPI/inc/MantidVatesAPI/vtkMDHWSignalArray.h
@@ -48,8 +48,7 @@ public:
   void PrintSelf(ostream &os, vtkIndent indent) override;
 
   void InitializeArray(Mantid::DataObjects::MDHistoWorkspace *ws,
-                       VisualNormalization normalization, std::size_t offset,
-                       vtkIdType size);
+                       VisualNormalization normalization, std::size_t offset);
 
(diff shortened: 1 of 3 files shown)


martyngigg (commented):
As we have a raw pointer is the lifetime of the vtkMDHWSignalArray guaranteed to be longer than the workspace?

quantumsteve:
@martyngigg While I assume one wouldn't delete a workspace they are viewing in the VSI, I don't believe there is anything actively preventing that from happening. Would switching to a `boost::shared_ptr` prevent this unlikely situation?

martyngigg:
@quantumsteve Thinking about it a bit more I can see that the original iterator approach would have suffered the same problem too as it would have been invalidated by the deletion. In theory the `shared_ptr` would keep the workspace alive but it look like the VSI gets automatically closed if the workspace is deleted anyway so I think this is actually not an issue.

LGTM :shipit:

PR #18310, opened 2016-12-18
Add fix for Nan in merged data fit

### LINKED ISSUE #18309: ISIS SANS Reduction fails when using Phi-Masking with merged detectors for LOQ
Steve King has a user who has seen isues with phi-masked data with merge-reduction mode:

>Mantid does the sector reductions. The error comes when it tries to merge the main and HAB datasets. And if you specify -90.0 to 90.0 (though I realise that is not very helpful for you) it seems to work ok. So it is a bit of an obscure bug!

Investigation shows that during the use of Q1D we have NaNs in the data which are not filtered out. this causes the `SANSFitShiftScale` algorithm to fail during fitting. We need to remove those NaN elements. It is not immedialtly clear which strategy to apply here, but it appars that interpolation might be the best option.

AntonPiccardoSelg:
Fixes #18309 

Reduced data which contains Nans is a problem for the merge algorithm (during fitting), ie.
```
 [1., 2., 3., nan, 5.]
```
is an issue. It is the result of a too fine Q bin setting of the user. There are several ways to approach this.

1. Replace the nan and apply linear interpolation.
2. Ignore nan values during fitting -> this is ok if only the low angle banks have nan values. We use the high angle banks as the fit model via the `TabulatedFunction`. If it has a nan it will fail
3. Continue to fail if there is a nan present, but provide a better exception message which indicates that the Q bin setting migth be too fine-grained.

We agreed on selecting option 3, since we don't want to alter data and in the case of option 2) it will only work some of the time.

## To Test

Please find the sample data here: \\olympic\Babylon5\Scratch\Anton\LoqData

1. Make sure that the path to the sample data is added to the Mantid path
2. Open the ISIS SANS Reduction GUI
3. Select LOQ as the instrument
4. Load the user file `USER_LOQ_161G_Hedstrom_8mm_Magnet-MERGED.txt`
5. Add the data

|                  | Scatter    |     Transmission    |   Direct     |
| --------------|------------|-----------------------|--------------|
|  Sample      |      LOQ97072    |     LOQ97074                       | LOQ97068         |
|  Sample      |      LOQ97069    |     LOQ97068                               | LOQ97068                 |

6. Press Load data / Wait until the data is loaded
7. Go to the ReductionSettings tab and change the Phi Limit values to -15.0 and 15.0
8. Go to the RunNumbers tab and press 1DReduce
  * Confirm that the reduction fails and that the exception message contains the hint that the Q binning might be too small.

## Release Notes

See [here](https://github.com/mantidproject/mantid/pull/18310/commits/fe98710c8f2dc4e73544e6b339f1de2e47543498)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/SANSFitShiftScale.py
+++ b/Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/SANSFitShiftScale.py
@@ -250,6 +250,10 @@ class SANSFitShiftScale(DataProcessorAlgorithm):
         if all([np.isnan(element) for element in signal]) or all([np.isnan(element) for element in error]):
             return
 
+        # If there are non to fix then don't touch the data
+        if not any([np.isnan(element) for element in signal]) and not any([np.isnan(element) for element in error]):
+            return
+
         # Perform NaN interpolation on a signal of the workspace
         nans, x = nan_checker(signal)
         signal[nans] = np.interp(x(nans), x(~nans), signal[~nans])
--- a/Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/SANSFitShiftScale.py
+++ b/Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/SANSFitShiftScale.py
@@ -143,9 +143,11 @@ class SANSFitShiftScale(DataProcessorAlgorithm):
         front_data_corrected, rear_data_corrected = error_correction.get_error_corrected(rear_data=q_low_angle,
                                                                                          front_data=q_high_angle,
                                                                                          q_min=q_min, q_max=q_max)
+        # If we have Nans in the data then perform a interpolated correction.
+        self.apply_nan_interpolation(front_data_corrected)
+        self.apply_nan_interpolation(rear_data_corrected)
 
         fit = self.createChildAlgorithm('Fit')
-
         # We currently have to put the front_data into the ADS so that the TabulatedFunction has access to it
         front_data_corrected = AnalysisDataService.addOrReplace('front_data_corrected', front_data_corrected)
         front_in_ads = AnalysisDataService.retrieve('front_data_corrected')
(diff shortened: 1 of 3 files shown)


Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/SANSFitShiftScale.py
```diff
@@ -144,8 +144,17 @@ def _determine_factors(self, q_high_angle, q_low_angle, mode, scale, shift):
...
                                                                                          q_min=q_min, q_max=q_max)
 
-        fit = self.createChildAlgorithm('Fit')
+        # The front_data_corrected data set is used as the fit model. Setting the IgnoreInvalidData on the Fit algorithm
+        # will not have any ignore Nans in the model, but only in the data. Hence this will lead to unreadable
+        # error messages of the fit algorithm. We need to catch this before the algorithm starts
+        y_model = front_data_corrected.dataY(0)
+        y_data = rear_data_corrected.dataY(0)
+        if any([np.isnan(element) for element in y_model]) or any([np.isnan(element) for element in y_data]):
+            raise RuntimeError("Trying to merge two merge the two reduced data sets for HAB and LAB failed. "
```
martyngigg: "Trying to merge two merge the two reduced ..." doesn't make sense to me. Do you mean "Trying to merge the two reduced ..."?

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/SANSFitShiftScale.py
```diff
@@ -144,8 +144,17 @@ def _determine_factors(self, q_high_angle, q_low_angle, mode, scale, shift):
...
+                               "You seem to have Nan values in your reduced HAB or LAB data set. This is most likely "
+                               "caused by a too small Q binning. Try to increase the Q bin width.")
 
+        fit = self.createChildAlgorithm('Fit')
```
martyngigg: I can't see where the fit variable is used?

martyngigg:
This looks good now.

The tests give the expected error message for a narrow angle range and small q binning but still function for the more wider ranges.

:shipit:

PR #18311, opened 2016-12-19
Exclude data from fit.

### LINKED ISSUE #18264: Exclude data points from fitting.
Add a property to `Fit` that define regions of the data that need to be excluded from fit.

mantid-roman:
A new property `Exclude` sets ranges to exclude from fit.

**To test:**

Run this script:

```
    x = np.linspace(-10, 10, 100)
    y = np.exp(-4*(x+3)**2) + np.exp(-4*(x-3)**2) + 0.1 - 0.001*x**2
    ws = CreateWorkspace(x, y)
    Fit("name=Polynomial,n=2", ws, Exclude=[-5, -1, 1, 5], Output='out')
```

Only the background should be fitted.

Fixes #18264.

Release notes are updated.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/CurveFitting/src/FitMW.cpp
```diff
@@ -33,6 +34,75 @@ using API::MatrixWorkspace;
...
+  const std::vector<double> &m_exclude;
+
+public:
+  /// Constructor.
+  /// @param exclude :: The value of the "Exclude" property.
+  /// @param startX :: The start of the overall fit interval.
+  /// @param endX :: The end of the overall fit interval.
+  ExcludeRangeFinder(const std::vector<double> &exclude, double startX,
+                     double endX)
+      : m_exclIndex(0), m_startExcludedRange(endX), m_endExcludeRange(endX),
```
gvardany: `m_startExcludedRange(startX)` ?

gvardany (changes requested):
It worked fine, however I have two points to suggest.
Firstly, I find it more intuitive to store the excluded ranges in the helper class as `std::vector<std::pair<double,double>>`. I believe it will make the helper class a lot easier to read and it will save from the overhead of checking odd/even indices.
Second, is the sorting of the ranges. In your test for example, when `exclude=9.0, 1.0, 4.0, -1.0, 6.5, 11.0`, I would expect the union of pair-wise sorted ranges to be excluded, i.e.  `-1.0, 11.0`. While now you sort the whole array, thus excluding`-1.0, 1.0, 4.0, 6.5, 9.0, 11.0`. 
So instead of sorting in `FitMW::setParameters()`, I would have another method in your helper class (invoked in initialisation), which will parse the given vector to a vector of non-overlaping ordered pairs, that represent the union of the ranges to be excluded.

Framework/CurveFitting/src/FitMW.cpp
mantid-roman: ? Will it work better?

mantid-roman:
@gvardany Please suggest an alternative implementation for your first point.

mantid-roman:
@gvardany Agree with your second point that overlapping regions should be combined. But arbitrary vectors should not be allowed: each pair should be ordered.

gvardany:
What is the problem with using pairs? For the sorting, of course in the normal case, pairs are expected to be ordered, but the alternative case should also be treated properly.

mantid-roman:
A simple vector has the correct structure of intervals by default. And some people may consider a vector of pairs of doubles more complex a structure. Are you sure that what you are suggesting will make the code clearer and easier to follow?

gvardany:
I agree that it is more complex structure, but here it seems very natural to use. Anyway, this is not critical, you can leave as is.

mantid-roman:
"Natural" is a very subjective term. If you think that it is natural because intervals and ranges should always be represented by pairs I would disagree. If it is natural because it leads to a cleaner and clearer code please let me know. Otherwise it's not worth the effort.

gvardany:
Well "complex" is as subjective as "natural". If one expects only a single range, it does not of course need to be a pair, but if you have a list of ranges, it is a different story. Anyway, I think it is a matter of taste, so I let the gatekeeper to comment on this.

Framework/CurveFitting/src/FitMW.cpp
```diff
@@ -69,6 +139,13 @@ void FitMW::setParameters() const {
       m_maxSize = static_cast<size_t>(maxSizeInt);
     }
     m_normalise = m_manager->getProperty(m_normalisePropertyName);
+    m_exclude = m_manager->getProperty(m_excludePropertyName);
+    if (m_exclude.size() % 2 != 0) {
+      throw std::runtime_error("Exclude property has an odd number of entries. "
```
gvardany: This throw is no longer needed.

gvardany (commented):
Thanks for the update, the new validator is nice and could be quite handy elsewhere.

martyngigg:
I think I would tend to agree with @gvardany that a range seems more naturally represented by a pair or even a small data struct? The names could then be clearer than `first` & `second`:

```
struct Range {
  double lower;
  double upper;
};
using RangeVector = std::vector<Range>;
```

martyngigg:
@gvardany Unfortunately Roman has been off sick and I don't expect we'll get time to make changes to this before the v3.9 code freeze.

Assuming everything works, would you be happy if we got this merged and cleaned up the code in maintenance?

gvardany (approved):

gvardany:
Yes, that works, :shipit:

PR #18318, opened 2016-12-19
Store masking information in class DetectorInfo

### LINKED ISSUE #18228: Intercept data from ParameterMap for Instrument-2.0 transition
Instrument-2.0 will store certain information such as mask flags and position information in class like `SpectrumInfo` and `DetectorInfo`. Currently this data is stored in `ParameterMap` (only offsets for positions).

For compatibility with legacy files we need to be able to intercept data that is being written to `ParameterMap`, reroute insertion of this data to new destinations, and inject this data when saving a `ParameterMap`.

The following locations have been identified as relevant:

- Loading parameters from an IDF: This is done by several load algorithms. Uses `ExperimentInfo::populateInstrumentParameters()`, which in turn calls `ExperimentInfo::populateWithParameter()`. Redirecting information seems to be straight forward.

- `LoadNexusProcessed`: Uses `ExperimentInfo::readParameterMap()`. Can filter and redirect data (`tokens[1]` is the parameter type, `tokens[2]` is the parameter name). Seems straight forward.

- `LoadMD`: Uses `MDBoxFlatTree::loadExperimentInfos`. This redirects either directly to `ExperimentInfo::readParameterMap()` (see above), or creates `FileBackedExperimentInfo`, which will eventually also calls that method. Seems straight forward.

- `SaveNexusProcessed`: Uses `ExperimentInfo::saveExperimentInfoNexus()`, which calls `Instrument::saveNexus()`. We can solve this one by creating a copy of the instrument, insert data from `DetectorInfo` into the parameter map of the temporary instrument, and call `saveNexus()` on that. Seems straight forward.

- `SaveMD2`: Uses `ExperimentInfo::saveExperimentInfoNexus()`, see above.

@ianbush @OwenArnold: I had a look at the mechanism we discussed, it seems to be surprisingly easy, basically all changes are within `ExperimentInfo`. Can you think of any other cases? Did I overlook anything?

### LINKED ISSUE #18290: Store mask flags in DetectorInfo instead of ParameterMap
Part of Instrument-2.0, masking will be handled by `DetectorInfo`. This needs to be done in combination with https://github.com/mantidproject/mantid/issues/18228.

### LINKED ISSUE #18317: Wrapper for Python getDetector method
To ensure existing Python scripts still work after `DetectorInfo` becomes responsible for storing parameters, such as masking, we need a wrapper for the `Instrument.getDetector()` Python method.

Initially this needs to work for `IDetector::isMasked`, eventually it will wrap other things like positions, rotations etc.

A deprecated warning should be thrown when using `Instrument.getDetector()`, but otherwise scripts should work as they would have previously.

### LINKED ISSUE #18319: Remove masking warning mechanism from DetectorGroup
This calls `IDetector::isMasked` which is no longer supported with masking data stored in `DetectorInfo`.

### LINKED ISSUE #18325: Remove masking awareness from DetectorGroup::solidAngle?
- Calls `IDetector::isMasked` which clashes with masking changes for Instrument-2.0.
- Is not consistent, since no other method in `DetectorGroup` ignores masked detectors?

The solution taken in https://github.com/mantidproject/mantid/pull/18318 is:
- Remove masking awareness from `DetectorGroup::solidAngle`
- Refactor client code to manually deal with masking

The advantages of this approach are:
- It is more visible in client code what is happening.
- `DetectorInfo` is more consistent since now none of its methods is masking aware.
- Easier refactoring for Instrument-2.0 masking rollout.

### LINKED ISSUE #18351: Store pointer to DetectorInfo in parameterized detectors
This is temporary refactoring that will allow for an easier transition to Instrument-2.0.

Prerequisite for:
- Legacy access to `IDetector::isMasked`, see https://github.com/mantidproject/mantid/issues/18317.
- Legacy code in the `Geometry` module that works on the `Component` level and needs access, to e.g., `getPos()`.

### LINKED ISSUE #18354: Remove IDetector::isMasked
Masking information via `SpectrumInfo` or `DetectorInfo` should be used instead.

SimonHeybrock:
This PR moves the storage location of mask flags for detectors from `ParameterMap` into `Beamline::DetectorInfo`. This is part of Instrument-2.0.

Masking data is intercepted when loading and added to `DetectorInfo` instead of `ParameterMap`. Masking data is injected into `ParameterMap` before saving. This ensures **backward and forward compatibility** for files (in particular processed Nexus).

For compatibility with legacy Python scripts a wrapper for `IDetector::isMasked` is provided. As a consequence docstests that are using this method are unchanged for now.

**To test:**

- Thorough code review.
- Manually check backward and forward compatibility of Nexus files: Have two Mantid builds (this branch and maybe a released version such as v3.8), in each of them mask detectors in a workspace, save as Nexus, try to load the file in the respective other build. Check that masking is loaded and looks the same.

Fixes #18290.
Fixes #18228.
Fixes #18319.
Fixes #18325.
Fixes #18351.
Fixes #18317.
Fixes #18354.

With the Python wrappers this is an internal change, so does not need to be in the release notes. We might observe performance improvements in a few cases, if so, those can be added later.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/src/EstimateResolutionDiffraction.cpp
+++ b/Framework/Algorithms/src/EstimateResolutionDiffraction.cpp
@@ -1,4 +1,6 @@
 #include "MantidAlgorithms/EstimateResolutionDiffraction.h"
+#include "MantidAPI/DetectorInfo.h"
+#include "MantidAPI/MatrixWorkspace.h"
 #include "MantidAPI/Run.h"
 #include "MantidAPI/SpectrumInfo.h"
 #include "MantidAPI/WorkspaceProperty.h"
@@ -148,6 +150,7 @@ void EstimateResolutionDiffraction::retrieveInstrumentParameters() {
   */
 void EstimateResolutionDiffraction::estimateDetectorResolution() {
   const auto &spectrumInfo = m_inputWS->spectrumInfo();
+  const auto &detectorInfo = m_inputWS->detectorInfo();
   const auto l1 = spectrumInfo.l1();
   g_log.notice() << "L1 = " << l1 << "\n";
   const V3D samplepos = spectrumInfo.samplePosition();
@@ -187,7 +190,12 @@ void EstimateResolutionDiffraction::estimateDetectorResolution() {
         spectrumInfo.isMonitor(i) ? 0.0 : spectrumInfo.twoTheta(i);
     double theta = 0.5 * twotheta;
 
-    double solidangle = det.solidAngle(samplepos);
+    double solidangle = 0.0;
+    for (const auto detID : m_inputWS->getSpectrum(i).getDetectorIDs()) {
+      const auto index = detectorInfo.indexOf(detID);
+      if (!detectorInfo.isMasked(index))
+        solidangle += detectorInfo.detector(index).solidAngle(samplepos);
+    }
     double deltatheta = sqrt(solidangle);
 
     // Resolution
--- a/Framework/DataHandling/test/SaveNexusProcessedTest.h
+++ b/Framework/DataHandling/test/SaveNexusProcessedTest.h
@@ -7,6 +7,7 @@
 #include "MantidDataObjects/WorkspaceSingleValue.h"
 #include "MantidDataHandling/LoadInstrument.h"
 #include "MantidAPI/Axis.h"
+#include "MantidAPI/DetectorInfo.h"
 #include "MantidAPI/FrameworkManager.h"
 #include "MantidAPI/WorkspaceFactory.h"
 #include "MantidAPI/TableRow.h"
@@ -16,6 +17,7 @@
 #include "MantidDataObjects/TableWorkspace.h"
 #include "MantidAPI/AnalysisDataService.h"
 #include "MantidDataHandling/SaveNexusProcessed.h"
+#include "MantidDataHandling/LoadEmptyInstrument.h"
 #include "MantidDataHandling/LoadMuonNexus.h"
 #include "MantidDataHandling/LoadNexus.h"
 #include "MantidKernel/Strings.h"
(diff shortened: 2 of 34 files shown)


Framework/DataHandling/test/SaveNexusProcessedTest.h
```diff
@@ -814,6 +816,47 @@ class SaveNexusProcessedTest : public CxxTest::TestSuite {
...
+    auto wsReloaded = boost::dynamic_pointer_cast<Workspace2D>(
+        AnalysisDataService::Instance().retrieve("testSpaceReloaded"));
+    TS_ASSERT_EQUALS(wsReloaded->detectorInfo().isMasked(0), false);
+    TS_ASSERT_EQUALS(wsReloaded->detectorInfo().isMasked(1), true);
+    TS_ASSERT_EQUALS(wsReloaded->detectorInfo().isMasked(2), false);
+
+    if (clearfiles)
+      Poco::File(file).remove();
+    AnalysisDataService::Instance().remove("testSpace");
+  }
```
OwenArnold: Useful integration test here. Good confidence that things are working properly.

Framework/Geometry/src/Instrument.cpp
```diff
@@ -1014,7 +1014,18 @@ void Instrument::saveNexus(::NeXus::File *file,
 
   // Now the parameter map, as a NXnote via its saveNexus method
   if (isParametrized()) {
-    const Geometry::ParameterMap &params = *getParameterMap();
+    Geometry::ParameterMap params(*getParameterMap());
+    // Masking is in DetectorInfo. Insert it into ParameterMap so it is saved
+    // alongside other parameters.
+    if (m_detectorInfo) {
```
OwenArnold: Is this functionality being encapsulated at the right level here? Wouldn't an `DetectorInfo::writeMasksToParameterMap`  be a better option?

Framework/Geometry/src/Instrument/ParameterMap.cpp
```diff
@@ -40,6 +40,12 @@ const std::string QUAT_PARAM_NAME = "Quat";
 
 // static logger reference
 Kernel::Logger g_log("ParameterMap");
+
+void checkIsNotMaskingParameter(const std::string &name) {
+  if (name == std::string("masked"))
```
OwenArnold: This is good. Should make our lives much easier in the future.

OwenArnold:
Looks pretty solid to me. And I imagine replacement of other parameters (positions, rotations) etc. would be very similar.

Framework/Geometry/src/Instrument.cpp
SimonHeybrock: That would be an temporary option as long as we have `API::DetectorInfo`, but I think `Beamline::DetectorInfo` should stay independent of the `Geometry` module and anything related to the instrument legacy?

Furthermore, the API of `DetectorInfo` is quite visible on the client side, so adding methods that should never be used is confusing. Of course that could be solved by declaring `Geometry::Instrument` as a `friend`, but in a sense that would add a cyclic inter-module dependency?

OwenArnold: I did wonder if that was the reason that you did it. It would indeed be better if it didn't depend on legacy things. Sounds like you've thought about it in any case.

Framework/Algorithms/src/EstimateResolutionDiffraction.cpp
```diff
@@ -187,7 +190,12 @@ void EstimateResolutionDiffraction::estimateDetectorResolution() {
         spectrumInfo.isMonitor(i) ? 0.0 : spectrumInfo.twoTheta(i);
     double theta = 0.5 * twotheta;
 
-    double solidangle = det.solidAngle(samplepos);
+    double solidangle = 0.0;
+    for (const auto detID : m_inputWS->getSpectrum(i).getDetectorIDs()) {
+      const auto index = detectorInfo.indexOf(detID);
+      if (!detectorInfo.isMasked(index))
+        solidangle += detectorInfo.detector(index).solidAngle(samplepos);
```
OwenArnold: Only unmasked detectors contributing to solid angle. Did you get any feedback on this from the wider development team further to our original discussions?

OwenArnold: I see that you're getting the solid angles off the detectors still, but would it be useful to get it instead off the via the detector info so that this condition could be a policy?

OwenArnold:
The code looks neat and well tested. I have performed functional testing masking/saving/loading between a version of mantid containing these changes and the last release. This worked as expected.

Historically, masking has proved a complex, fragile (and occasionally mysterious) area of the code base. Anything that can be done to improve readability and comments along the way is always greatly appreciated.

Framework/Algorithms/src/EstimateResolutionDiffraction.cpp
SimonHeybrock: No, I did not get feedback. However, in I think all cases I refactored it did actually seem to make sense to ignore masked detectors. This is now done explicitly and manually by clients and thus more obvious to the developer, which I see as an advantage.

Regarding adding this as functionality to `DetectorInfo`, I agree, but I would avoid that at this point for two reasons:
- We are in the middle of some rather complex refactoring.
- I am not sure of this is something that should actually be part of the class interface. Maybe a stand alone method that just uses `DetectorInfo` might make more sense. Since this is not an issue that blocks us anywhere I would rather postpone this and similar design decision until the basics are more settled.

OwenArnold: Ok that sounds like a sensible strategy. Is this the only area where behaviour may be different from Mantid 3.8?

SimonHeybrock: I think so, but `IDetector::solidAngle` does not seem to be used in Python (even though there is an export, it only shows up in an example).

OwenArnold:
In the absence of any other TSC comments :shipit:

PR #18320, opened 2016-12-19
Basic Python exports for SpectrumInfo and DetectorInfo

### LINKED ISSUE #18303: Python exports for SpectrumInfo and DetectorInfo
See title.

SimonHeybrock:
This adds basic Python exports for `SpectrumInfo` and `DetectorInfo`. The goal is to replace access to `IDetector::isMasked` and other methods that will be removed with Instrument-2.0.

See also #18317, which will provide way to keep legacy access possible in a transition period, to avoid/postpone breaking user scripts.

**To test:**

Code review.

Fixes #18303.

I think this should not yet be part of the release notes since the functionality is very incomplete.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/PythonInterface/mantid/api/src/Exports/DetectorInfo.cpp
+++ b/Framework/PythonInterface/mantid/api/src/Exports/DetectorInfo.cpp
@@ -6,6 +6,9 @@ using namespace boost::python;
 
 void export_DetectorInfo() {
   class_<DetectorInfo, boost::noncopyable>("DetectorInfo", no_init)
+      .def("__len__", &DetectorInfo::size, (arg("self")),
+           "Returns the size of the DetectorInfo, i.e., the number of "
+           "detectors in the instrument.")
       .def("size", &DetectorInfo::size, (arg("self")),
            "Returns the size of the DetectorInfo, i.e., the number of "
            "detectors in the instrument.")
--- a/Framework/PythonInterface/mantid/api/src/Exports/DetectorInfo.cpp
+++ b/Framework/PythonInterface/mantid/api/src/Exports/DetectorInfo.cpp
@@ -1,0 +1,14 @@
+#include "MantidAPI/DetectorInfo.h"
+#include <boost/python/class.hpp>
+
+using Mantid::API::DetectorInfo;
+using namespace boost::python;
+
+void export_DetectorInfo() {
+  class_<DetectorInfo, boost::noncopyable>("DetectorInfo", no_init)
+      .def("size", &DetectorInfo::size, (arg("self")),
+           "Returns the size of the DetectorInfo, i.e., the number of "
+           "detectors in the instrument.")
+      .def("isMasked", &DetectorInfo::isMasked, (arg("self"), arg("index")),
+           "Returns True if the detector is masked.");
+}
--- a/Framework/PythonInterface/mantid/api/CMakeLists.txt
+++ b/Framework/PythonInterface/mantid/api/CMakeLists.txt
@@ -13,6 +13,7 @@ set ( EXPORT_FILES
   src/Exports/AlgorithmHistory.cpp
   src/Exports/CatalogManager.cpp
   src/Exports/CatalogSession.cpp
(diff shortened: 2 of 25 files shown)


Framework/PythonInterface/mantid/api/src/Exports/DetectorInfo.cpp
```diff
@@ -0,0 +1,14 @@
+#include "MantidAPI/DetectorInfo.h"
+#include <boost/python/class.hpp>
+
+using Mantid::API::DetectorInfo;
+using namespace boost::python;
+
+void export_DetectorInfo() {
+  class_<DetectorInfo, boost::noncopyable>("DetectorInfo", no_init)
+      .def("size", &DetectorInfo::size, (arg("self")),
```
OwenArnold: I think that `size` is too ambiguous going forward. The prototype has the ability to tell you via `DetectorInfo`: 

* The total number of components  
* The number of which are path components
* The number of which are actually detectors

I'm prefer we were a bit more explicit even though the natural assumption is that we are talking about detector size.

OwenArnold:
Looks like a good start to me. What does @martyngigg  think?

Framework/PythonInterface/mantid/api/src/Exports/DetectorInfo.cpp
martyngigg: If we rename `::size` then I think it should be on both the Python and C++ side to match. I don't see a compelling reason for them to be different.

martyngigg: Sorry, I didn't check the current state of `DetectorInfo` and thought the question was regarding clashing names. I think `size()` is fine here too. Should we add `__len__` too so that `len()` works?

SimonHeybrock: By the way, I do intend to provide iterator support for `SpectrumInfo` and `DetectorInfo`, but I have not yet come up with a solid and nice idea.

martyngigg: @SimonHeybrock Do you mean you're not sure how to implement the iterator behaviour or not yet sure what the iterator should provide?

SimonHeybrock: @martyngigg The latter. I am considering supporting things like this:

```cpp
for (const auto &det : ws.detectorInfo()) {
  if (!det.isMasked())
    pos += det.position();
  // ...
}
```
where the type of `det` is some kind of iterator "proxy" forwarding all methods to `DetectorInfo`. On the other hand not much of our code is using such loops, since often we need an index to access other containers at the same time (or use our OpenMP macros that do not support range-based for), so maybe this is not an urgent addition.

Any thoughts?

martyngigg: @SimonHeybrock You're example would have been my first thought too but I agree that most code does not need to iterate over the detectors at this level. It may be better to wait until we can find a good example of a more real use case for this.

martyngigg:
Should there be some unit tests for the new export in https://github.com/mantidproject/mantid/tree/master/Framework/PythonInterface/test/python/mantid/api?

SimonHeybrock:
@OwenArnold @martyngigg Both `SpectrumInfo` and `DetectorInfo` will eventually move to the new `Beamline` module. Do you think this would pose a problem for Python exports? I guess it would only affect cases where someone uses things like `from mantid.api import SpectrumInfo`, which is rare, I suppose?

martyngigg:
@SimonHeybrock Yes, it will only affect those cases. In this case, especially as it is new, I agree that the number of uses would be sufficiently small that it's not worth worrying too much.

If we feel it necessary in the future we can always provide a shim when the new modules are in place.

OwenArnold:
:shipit: I don't see any more reasons to prevent this from being merged.

PR #18321, opened 2016-12-19
Improve object sampling in Monte Carlo absorption

martyngigg:
A user spotted that the attenuation computed by the `MonteCarloAbsorption` algorithm did not match as closely with the `CylinderAbsorption` as was expected. This was found to be due to the method by which the scattering point within the object was selected.

Before these changes the a flat beam was uniformly sampled and then the depth of the scattering point in the object was randomly selected. This, however, only produces a uniforming sampling of the object if the object is a flat plate. Any other shape would suffer from over sampling in areas where the shape does not conform to the flat plate or had a hole in it. 

This change alters the method of scatter point selection so that a point is randomly selected within the object and only then computing the path that would be taken. A comparison of the results is provided in the test instructions.  

**To test:**

The following script is taken and modified from the usage example:

```python
data = CropWorkspace(data, StartWorkspaceIndex=0, EndWorkspaceIndex=0)
SetSample(data,  Material={'ChemicalFormula': '(Li7)2-C-H4-N-Cl6', 'SampleNumberDensity': 0.07})

# CylinderAbsorption
abscor_cyl = CylinderAbsorption(data, CylinderSampleHeight=3.0, CylinderSampleRadius=1.0,NumberOfSlices=25, NumberOfAnnuli=25,
                                                    NumberofWavelengthPoints=20)

# Monte Carlo
# Default up axis is Y
SetSample(data, Geometry={'Shape': 'Cylinder', 'Height': 3.0, 'Radius': 1.0,
                  'Center': [0.0,0.0,0.0]})
abscor_mc = MonteCarloAbsorption(data, NumberOfWavelengthPoints=20,EventsPerPoint=1000)
```

It should produce the following output in both v3.8 (left) and with these changes (right):

montecarloabs

Play with the parameters and convince yourself the new method is an improvement.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


AntonPiccardoSelg:
The code looks good and understandable. There is testing in place. 

Functionally it is an improvement over M3.8. I played around with various settings and the new functionally seems to be better in all cases. Sometimes it is very noisy, but that is probably because I am not familiar with the response behaviour of this MC for the tested parameter space.

@martyngigg Should we have this in the release notes since a user approached us with this issue?

martyngigg:
@AntonPiccardoSelg Thanks for the reminder, it slipped my mind after I updated the documentation. I'll put  something in there.

Issue #18215, opened 2016-12-07
Add specific missing unit tests


NickDraper:
We are missing unit tests on a few algorithms, this ticket is to deal with these ones:

 1. CalculateSlits
 1. LoadSampleDetailsFromRaw
 1. UnwapMonitor
 1. MaskDetectorsIf
 1. CrossCorrelate
 1. LoadInstrumentFromNexus

You can use github to identify the authors (Blame is often easier than history) and talk to them if the code / documentation is not good enough.

Submit each as a separate pull request

gemmaguest:
Unit tests have now been added for all of these algorithms.

PR #18328, opened 2016-12-20
Add tests for the UnwrapMonitor algorithm

gemmaguest:
This adds missing unit tests for the UnwrapMonitor algorithm.

**To test:**

Run UnwrapMonitorTest.

Refs #18215.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/test/UnwrapMonitorTest.h
+++ b/Framework/Algorithms/test/UnwrapMonitorTest.h
@@ -16,41 +16,6 @@ using namespace Mantid::DataObjects;
 using namespace Mantid::Kernel;
 
 class UnwrapMonitorTest : public CxxTest::TestSuite {
-private:
-  const MatrixWorkspace_sptr makeFakeWorkspace() {
-    Workspace2D_sptr testWS =
-        WorkspaceCreationHelper::create2DWorkspaceWithRectangularInstrument(
-            2, 3, 50);
-    testWS->getAxis(0)->setUnit("TOF");
-    return testWS;
-  }
-
-  void setupAlgorithm(UnwrapMonitor &algo, const MatrixWorkspace_sptr inWS,
-                      const double lref) {
-    if (!algo.isInitialized())
-      algo.initialize();
-    algo.setChild(true);
-    algo.setProperty("InputWorkspace", inWS);
-    algo.setPropertyValue("OutputWorkspace", "outWS");
-    algo.setProperty("LRef", lref);
-  }
-
-  // Run the algorithm and do some basic checks. Returns the output workspace.
-  MatrixWorkspace_const_sptr runAlgorithm(UnwrapMonitor &algo,
-                                          const MatrixWorkspace_sptr inWS) {
-    // run the algorithm
-    TS_ASSERT(algo.execute());
-    TS_ASSERT(algo.isExecuted());
-
-    // verify the output workspace
-    const MatrixWorkspace_const_sptr outWS =
-        algo.getProperty("OutputWorkspace");
-    TS_ASSERT_EQUALS(inWS->getNumberHistograms(),
-                     outWS->getNumberHistograms()); // shouldn't drop histograms
-
-    return outWS;
-  }
-
 public:
   // This pair of boilerplate methods prevent the suite being created statically
   // This means the constructor isn't called when running other tests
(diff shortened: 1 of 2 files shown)


Framework/Algorithms/test/UnwrapMonitorTest.h
```diff
@@ -0,0 +1,212 @@
...
+#include "MantidAlgorithms/UnwrapMonitor.h"
+
+using namespace Mantid::Algorithms;
+using namespace Mantid::API;
+using namespace Mantid::DataObjects;
+using namespace Mantid::Kernel;
+
+class UnwrapMonitorTest : public CxxTest::TestSuite {
+private:
+  void makeFakeWorkspace(std::string wsName) {
```
ianbush: This could return a `const MatrixWorkspace_sptr` and `wsName` can then be dropped.

Framework/Algorithms/test/UnwrapMonitorTest.h
```diff
@@ -0,0 +1,212 @@
...
+    std::string name("UnwrapMonitor");
+    this->makeFakeWorkspace(name);
+    Workspace2D_sptr ws =
+        AnalysisDataService::Instance().retrieveWS<Workspace2D>(name);
+    size_t nHist = ws->getNumberHistograms();
+
+    // run the algorithm
+    UnwrapMonitor algo;
+    if (!algo.isInitialized())
+      algo.initialize();
```
ianbush: If `algo.setChild(true)` is added here...

Framework/Algorithms/test/UnwrapMonitorTest.h
```diff
@@ -0,0 +1,212 @@
...
+    algo.setPropertyValue("InputWorkspace", name);
+    algo.setPropertyValue("OutputWorkspace", name);
+    algo.setProperty("LRef", 11.0);
+    TS_ASSERT(algo.execute());
+    TS_ASSERT(algo.isExecuted());
+
+    // verify the output workspace
+    ws = AnalysisDataService::Instance().retrieveWS<Workspace2D>(name);
```
ianbush: ...then you can use `MatrixWorkspace_const_sptr outWS = algo.getProperty("OutputWorkspace");` here...

Framework/Algorithms/test/UnwrapMonitorTest.h
```diff
@@ -0,0 +1,212 @@
...
+    TS_ASSERT_DELTA(outY[3], 0.0, 1e-6);
+    TS_ASSERT_DELTA(outY[4], 2.991736, 1e-6);
+    TS_ASSERT_DELTA(outY[8], 6.198347, 1e-6);
+    TS_ASSERT_DELTA(outY[11], 3.818182, 1e-6);
+    TS_ASSERT_DELTA(outY[12], 0.0, 1e-6);
+
+    double joinWavelength = algo.getProperty("JoinWavelength");
+    TS_ASSERT_DELTA(joinWavelength, 0.003692, 1e-6);
+
+    AnalysisDataService::Instance().remove(name);
```
ianbush: ...and drop the need to remove the workspace here.

Framework/Algorithms/test/UnwrapMonitorTest.h
```diff
@@ -0,0 +1,212 @@
...
+  static UnwrapMonitorTest *createSuite() { return new UnwrapMonitorTest(); }
+  static void destroySuite(UnwrapMonitorTest *suite) { delete suite; }
+
+  void testLrefLessThanLd() {
+    // setup
+    std::string name("UnwrapMonitor");
+    this->makeFakeWorkspace(name);
+    Workspace2D_sptr ws =
+        AnalysisDataService::Instance().retrieveWS<Workspace2D>(name);
+    size_t nHist = ws->getNumberHistograms();
```
ianbush: I would avoid using `ws` for the input and then the output later. Calling them something like `inWS` and `outWS` makeso make the tests more readable. Both can be kept then `nHist` is not needed - again I think this makes the test easier to read.

ianbush (commented):
The tests added are a nice improvement and cover a good range of use cases - long overdue given when the original algorithm was written!

I realise the test as written does follow a typical approach for Mantid unit tests, but I think the use of the Analysis Data Service can be avoided by setting the algorithm as a child (see comments).

After doing this some of the duplication in the tests should also be moved into a common methods, for example in the setup part I think the only change is the value of `LRef`.

For some tests a number of Y values are checked and others not - was there a particular reason for this? Given the conversion that occurs in the algorithm would it also be good to add some checks for the X axis value?

ianbush:
@gemmaguest Can you add the milestone label too?

gemmaguest:
Thanks for the feedback @ianbush - I'll make those changes. No particular reason for not checking the Y values in each test - I just wasn't sure how much detail it's good to include. I agree it would be good to check some X values too.

ianbush:
Thanks for the updates, this looks good to me :shipit: .

PR #18338, opened 2016-12-21
Fix threshold filter in splatterplot view

### LINKED ISSUE #18330: Threshold filter not working in SplatterPlot view
### Expected behavior

Clicking the "threshold" button, changing the min/max and clicking apply should change the number of points displayed.

### Actual behavior

Regardless of the input, the output of the threshold filter has zero points 

Open a MDEventWorkspace in the VSI, (if necessary) switch to the SplatterPlot view, click on the "threshold filter" and click "apply".

### Platforms affected

all

quantumsteve:
Description of work.

This adds cell data to the vtkUnstructuredGrid generated by the SplatterPlot filter, which is required by the vtkThreshold filter. I also updated the icon to be consistent with the VSI standard view and ParaView. Lastly I removed an incorrect and unnecessary tooltip in the VSI standard view. 

I created a issue #18339 to find a better solution that doesn't require adding cell data to the vtkUnstructuredGrid.

**To test:**

1. Load a MDEventWorkspace and open in in the VSI.
2. (if necessary) Switch to the SplatterPlot view
3. Click the threshold filter button, change the min and max values and click apply.
4. Check if the output changes.

Fixes #18330.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


AntonPiccardoSelg:
The threshold filter does not seem to work for me.

Setting threshold to the very low end produces this:
image

Setting the threshold to the very high end produces this:
image

Maybe I don't quite understand the threshold filter by I would expect the cells with a signal value outside of the threshold range to not be displayed. Here they do seem to be displayed. Am I using it the filter incorrectly?

quantumsteve:
@AntonPiccardoSelg Try hiding `MDScatterPlot1` so you only see the results of the threshold filter.

AntonPiccardoSelg:
The threshold filter works just fine. Hiding the `MDScatterPlot` filter output ensures that we only see the output of the threshold filter.

@quantumsteve Should the `MDScatterPlot` filter not be made automatically invisible once I apply the `Threshold` filter?

quantumsteve:
@AntonPiccardoSelg How would we do that? Is there a signal or callback we can use? I'd like to make that change, but it may be enough additional work to justify a second pull request.

AntonPiccardoSelg:
@quantumsteve In principle it should be PV's job to make the last element in the pipeline invisible when you add a new filter. It must be a bug, since it does it automatically when you apply other filters to a pipeline (see scale or cut filter)

If it is a bug then we can traverse the pipeline to the element which is visible, get its representation and use [`setVisible (bool visible)`](http://www.paraview.org/ParaView/Doc/Nightly/www/cxx-doc/classpqDataRepresentation.html) to make it invisible. I don't really like this, since it is a hack, but I cannot see any other way currently.

quantumsteve:
@AntonPiccardoSelg I can't reproduce the issue inside of ParaView or in the with the threshold filter in the Standard View. Any idea what is different about this situation?

Vates/VatesSimpleGui/ViewWidgets/src/SplatterPlotView.cpp
```diff
@@ -284,6 +288,13 @@ void SplatterPlotView::onThresholdButtonClicked() {
...
   this->m_threshSource = builder->createFilter(
       "filters", MantidQt::API::MdConstants::Threshold, this->m_splatSource);
+  auto filterProxy =
+      builder->createDataRepresentation(this->m_threshSource->getOutputPort(0),
+                                        this->m_view)->getProxy();
+  vtkSMPropertyHelper(filterProxy, "Representation")
+      .Set(g_defaultRepresentation);
+  vtkSMPropertyHelper(filterProxy, "Opacity").Set(g_defaultOpacity);
+  vtkSMPropertyHelper(filterProxy, "GaussianRadius").Set(g_defaultRadius);
   emit this->lockColorControls();
```
quantumsteve: I tried adding `this->setVisibilityListener()` here, but the filter visibility issue still exists.

AntonPiccardoSelg:
Getting this fully right seems to be quite hard because of PV. This should probably be addressed in a spearate PR. Filtering itself works as expected. Therefore, please :shipit:

PR #18343, opened 2016-12-22
Added masking-by-component to MaskDetectors

### LINKED ISSUE #18327: Masking by component in MaskDetectors
Currently [`MaskDetectors`](http://docs.mantidproject.org/nightly/algorithms/MaskDetectors-v1.html) supports only index-based masking. There is, however, need for component name based masking as well.

soininen:
This PR adds a new *ComponentList* input property to [`MaskDetectors`](http://docs.mantidproject.org/nightly/algorithms/MaskDetectors-v1.html). The property takes a list of component names as strings and adds the detectors found recursively in the given components to the list of detector IDs to be masked.

Additionally, `ComponentCreationHelper::createTestInstrumentCylindrical()` was modified slightly to enable proper unit testing and the affected `InstrumentRayTracerTest` was updated.

**To test:**

Load your favorite dataset with attached instrument or just an empty instrument using [`LoadEmptyInstrument`](http://docs.mantidproject.org/nightly/algorithms/LoadEmptyInstrument-v1.html). Try masking different components in the component tree.
- Using a plain component name should mask all components with the name.
- Slash '/' can be used to more accurately specify a component, e.g. 'bank1/tube3'. However, in this case only a single component whose name matches the component string is masked.
- The algorithm should work regardless of the combination of parameters given, e.g. one is able to specify components, spectrum numbers and limit the masked range by *StartWorkspaceIndex* and *EndWorkspaceIndex* simultaneously.
- Non-existing component names don't fail but should print an error to the log.
- Component names with commas ',' will just confuse the algorithm. Consider updating your IDFs.

Fixes #18327 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


docs/source/algorithms/MaskDetectors-v1.rst
```diff
@@ -132,13 +130,14 @@ mask detectors, including
 
 -  Workspace indices
 -  Spectra
--  Detectors
+-  Detects
```
KarlPalmen: Detectors?

KarlPalmen:
I built the documentation and found the following warning
```
15>  C:\Mantid\staging_mantid\docs\source\algorithms\MaskDetectors-v1.rst:446: WARNING: Title underline too short.
15>
15>  Example 7: constraining the masking range
15>  ######################################
```

KarlPalmen:
After reading the documentation, I found out that `Detects` as shown in my comment could be `Detector IDs` as in title of appropriate usage example section.

docs/source/algorithms/MaskDetectors-v1.rst
soininen: Thanks for checking the documentation as well! Indeed, 'Detector IDs' would be a better fit here, also perhaps 'Spectrum numbers' instead of 'Spectra'. I did play with the idea here, thus the unindented typo 'Detects'.

SimonHeybrock:
@soininen Just a small note: I think `MaskDetectors` does not actually mask detectors, but all spectra that contain a detector that is to be masked. If you consider grouped detectors this might not give what you intend.

soininen:
@SimonHeybrock Thanks for the note. You are right, the algorithm's name is misleading. I'll keep it in mind although I don't have to consider detectors groups at the moment. Anyway, the behavior of component-based masking consistent with detector ID-based masking so in that sense there should not be surprises to the user.

KarlPalmen:
Apologies for taking so long. I was off sick with a bad cold after Christmas. Changes are good :shipit: .

Framework/DataHandling/src/MaskDetectors.cpp
```diff
@@ -495,5 +500,65 @@ void MaskDetectors::appendToIndexListFromMaskWS(
...
+        const bool recursively = true;
+        assembly->getChildren(childComponents, recursively);
+        for (const auto &childComp : childComponents) {
+          const auto childDetector =
+              boost::dynamic_pointer_cast<const IDetector>(childComp);
+          if (childDetector) {
+            detectorIDs.emplace(childDetector->getID());
+          }
+        }
+      }
```
SimonHeybrock: Could this be made simpler by using `Instrument::getDetectorsInBank`?

soininen: Thanks for the hint! The code indeed gets simplified, even though `Instrument::getDetectorsInBank` doesn't take into account the case where the given component in itself is a detector. Is this intended behavior? The doxygen documentation says *'Fill a vector with all the detectors contained (at any depth) in a named component.'* I would expect 'at any depth' to mean also the component itself.

docs/source/algorithms/MaskDetectors-v1.rst
```diff
@@ -162,14 +164,14 @@ Operations Involved in Masking
 There are 2 operations to mask a detector and thus spectrum related
 
 1. Set the detector in workspace's instrument's *parameter map* to *masked*.
-2. Clear the data associated with the spectrum with detectors that are masked.
+2. Zero the data associated with the spectrum with detectors that are masked.
```
AndreiSavici: For Event workspaces, it also clears the event list

soininen: Thanks for the clarification! Unfortunately this PR is already merged, but I have opened an issue to fix this: #18407

PR #18350, opened 2017-01-01
Remove duplicate member function Kernel::DataItem::name().

quantumsteve:
Description of work.

This removes the (C++) member function `Kernel::DataItem::name()` which is identical to `API::Workspace::getName()`, but returns a new `std::string` instead of a const reference.  These changes should remove many unnecessary copies.

I also removed the python member function `Workspace::getName()` which duplicates `DataItem::name()`, isn't commonly used, and isn't terribly pythonic. 

The latter would be even better as a read-only data member ( `DataItem::name` ) , but having both at the same time leads to runtime errors. [This change](https://github.com/mantidproject/mantid/files/680805/dataitem_name.txt) would break many scripts and is most likely best done later as part of "Mantid 4.0".  Alternatively, we could move the python interface to only use `DataItem::getName()` and add a read-only data member ( `DataItem::name` ). 

**To test:**

Review changes to `Kernel::DataItem` and `API::Workspace`. Most other changes are renaming that should be caught by the build servers.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


martyngigg:
I think the changes here make sense, there's no reason that the name string needs to be returned by value.

I agree that having attributes as properties would be better. I think an overhaul of the python API documentation has been required for a while as it's very hard for users to decipher how to use methods/properties from the auto-generated documentation.

martyngigg:
@quantumsteve There's a merge conflict now.

martyngigg (approved):

martyngigg:
The OSX error is unrelated to these changes and can be ignored.

:shipit:

PR #18357, opened 2017-01-03
SavePlot1D hide spectra

peterfpeterson:
Add the ability to specify which spectra to show. This will allow NOMAD's autoreduction to use the standard plotting rather than have one hand-written in their script.

**To test:**

Load a dataset and try to export the plot with and without `SpectraList` specified for both images (e.g. `.jpeg`) and plotly.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/SavePlot1D.py
```diff
@@ -104,6 +108,15 @@ def getData(self, ws, wkspIndex, label=''):
 
         return (x, y, label)
 
+    def showSpectrum(self, ws, wkspIndex):
+        spectraNum = ws.getSpectrum(wkspIndex).getSpectrumNo()
+
+        # user not specifying which spectra means show them all
+        if len(self.visibleSpectra) <= 0:
```
AndreiSavici: Can an array length be negative?

peterfpeterson: No, but the return value of `len` is a signed integer so I'd rather be paranoid.

martyngigg:
@peterfpeterson There's a merge conflict

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/SavePlot1D.py
AndreiSavici: http://stackoverflow.com/questions/15650878/returning-from-len-when-64-bits

AndreiSavici:
:shipit:

PR #18361, opened 2017-01-04
Remove uses of IDetector::isMonitor in Algorithms

### LINKED ISSUE #18360: Remove uses of IDetector::isMonitor in Algorithms
This is part of preparations for moving monitor information into `DetectorInfo`.

SimonHeybrock:
Part of Instrument-2.0 rollout. This is just a refactor, behavior is unchanged.

**To test:**

Code review.

Fixes #18360.

No release notes.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/src/ModeratorTzeroLinear.cpp
+++ b/Framework/Algorithms/src/ModeratorTzeroLinear.cpp
@@ -1,6 +1,7 @@
 #include "MantidAlgorithms/ModeratorTzeroLinear.h"
 #include "MantidAPI/Axis.h"
 #include "MantidAPI/Run.h"
+#include "MantidAPI/SpectrumInfo.h"
 #include "MantidAPI/WorkspaceFactory.h"
 #include "MantidAPI/WorkspaceUnitValidator.h"
 #include "MantidDataObjects/EventWorkspace.h"
@@ -130,6 +131,7 @@ void ModeratorTzeroLinear::exec() {
   }
 
   // do the shift in X
+  const auto &spectrumInfo = inputWS->spectrumInfo();
   const size_t numHists = inputWS->getNumberHistograms();
   Progress prog(this, 0.0, 1.0, numHists); // report progress of algorithm
   PARALLEL_FOR_IF(Kernel::threadSafe(*inputWS, *outputWS))
@@ -137,7 +139,7 @@ void ModeratorTzeroLinear::exec() {
     PARALLEL_START_INTERUPT_REGION
     double t_f, L_i;
     size_t wsIndex = static_cast<size_t>(i);
-    calculateTfLi(inputWS, wsIndex, t_f, L_i);
+    calculateTfLi(spectrumInfo, wsIndex, t_f, L_i);
 
     outputWS->setHistogram(i, inputWS->histogram(i));
     // shift the time of flights
@@ -185,6 +187,7 @@ void ModeratorTzeroLinear::execEvent() {
   auto outputWS = boost::dynamic_pointer_cast<EventWorkspace>(matrixOutputWS);
 
   // Loop over the spectra
+  const auto &spectrumInfo = matrixOutputWS->spectrumInfo();
   const size_t numHists = outputWS->getNumberHistograms();
   Progress prog(this, 0.0, 1.0, numHists); // report progress of algorithm
   PARALLEL_FOR_IF(Kernel::threadSafe(*outputWS))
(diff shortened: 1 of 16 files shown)


OwenArnold:
Looks pretty good. Some minor issues noted. Moving forward I expect that the plan is to make the `isMonitor` implementations near identical to those for `isMasked`, but this seems to be a useful first step.

Framework/Algorithms/src/ModeratorTzeroLinear.cpp
```diff
@@ -212,57 +215,35 @@ void ModeratorTzeroLinear::execEvent() {
...
+
+  if (!spectrumInfo.hasDetectors(i)) {
     t_f = TfError;
     return;
   }
 
-  if (det->isMonitor()) {
-    L_i = m_instrument->getSource()->getDistance(*det);
+  if (spectrumInfo.isMonitor(i)) {
+    L_i = spectrumInfo.sourcePosition().distance(spectrumInfo.position(i));
```
OwenArnold: Have you considered adding a method for this distance calculation for monitors?

`SpectrumInfo::lMonitor` or something similar.

Framework/Algorithms/src/SofQWCentre.cpp
```diff
@@ -111,24 +113,12 @@ void SofQWCentre::exec() {
...
-  // Get a pointer to the instrument contained in the workspace
-  Instrument_const_sptr instrument = inputWorkspace->getInstrument();
-
-  // Get the distance between the source and the sample (assume in metres)
-  IComponent_const_sptr source = instrument->getSource();
-  IComponent_const_sptr sample = instrument->getSample();
-  V3D beamDir = sample->getPos() - source->getPos();
+  const auto &detectorInfo = inputWorkspace->detectorInfo();
+  const auto &spectrumInfo = inputWorkspace->spectrumInfo();
+  V3D beamDir = detectorInfo.samplePosition() - detectorInfo.sourcePosition();
```
OwenArnold: This could also be made a public method of `DetectorInfo`

Framework/Algorithms/src/ModeratorTzeroLinear.cpp
SimonHeybrock: Well, `SpectrumInfo` defines L1+L2 = distance from source, and thus L2 is negative for monitors. That is what most algorithms use (and that is why `SpectrumInfo` does it). Here the algorithms wants something non-standard, so it is more explicit.

Framework/Algorithms/src/SofQWCentre.cpp
```diff
@@ -141,109 +131,97 @@ void SofQWCentre::exec() {
...
+
+    const auto &detIDs = inputWorkspace->getSpectrum(i).getDetectorIDs();
+    double numDets_d = static_cast<double>(detIDs.size());
+    const auto &Y = inputWorkspace->y(i);
+    const auto &E = inputWorkspace->e(i);
+    const auto &X = inputWorkspace->x(i);
+
+    // Loop over the detectors and for each bin calculate Q
+    for (const auto detID : detIDs) {
+      size_t idet;
```
OwenArnold: unintialized

Framework/Algorithms/src/SofQWCentre.cpp
SimonHeybrock: I am not sure about that yet. It has little to do with detectors. Maybe making it a method of the (no yet existing) new instrument would make more sense. The same goes for L1, which we *did* add, but that is used very frequently so it was useful. We should however also consider whether or not it should stay, or be moved to another class.

Framework/Algorithms/src/SofQWCentre.cpp
```diff
@@ -141,109 +131,97 @@ void SofQWCentre::exec() {
...
-                  "Incident energy of a neutron is negative. Are you trying to "
-                  "process Direct data in Indirect mode?\n Found for det ID: " +
-                  std::to_string(idet) + " bin No " + std::to_string(j) +
-                  " with efied=" + boost::lexical_cast<std::string>(efixed) +
-                  " and energy transfer: " +
-                  boost::lexical_cast<std::string>(deltaE);
-              throw std::runtime_error(mess);
-            }
+      // Calculate kf vector direction and then Q for each energy bin
+      V3D scatterDir =
```
OwenArnold: Maybe along a similar lines to my comment above about `beamDir`? Maybe a non-member function in `DetectorInfo` would be suitable?

SimonHeybrock: Yes, that would make sense, I think. We can add this once the basics are more settled.

OwenArnold:
Apart from some minor comments, this also looks good.

Framework/Algorithms/src/SofQWCentre.cpp
OwenArnold: 👍

Framework/Algorithms/src/ModeratorTzeroLinear.cpp
OwenArnold: 👍  I think it is better to keep things explicit when they are exceptional calculations local to a single algorithm.

OwenArnold:
:shipit:

PR #18362, opened 2017-01-04
Fix parameters without ties

### LINKED ISSUE #18312: Function parameter fixed to a number should not create a tie.
All tied parameters are marked as fixed which means that the minimizer doesn't vary it (directly). If the tie is a number there is no need to apply this tie as the value never changes. It means that it's enough to mark the parameter as fixed without creating the tie.

mantid-roman:
When a parameter is tied to a number no ParameterTie instance is stored.

**To test:**

Review code.

Fixes #18312.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


martyngigg:
@mantid-roman A conflict has developed here

Framework/API/src/IFunction.cpp
```diff
@@ -110,8 +111,14 @@ void IFunction::functionDeriv(const FunctionDomain &domain,
 ParameterTie *IFunction::tie(const std::string &parName,
                              const std::string &expr, bool isDefault) {
   auto ti = new ParameterTie(this, parName, expr, isDefault);
```
martyngigg: Can we avoid the raw pointer with `new` and use a unique_ptr?

mantid-roman: OK.

Framework/API/test/FunctionFactoryTest.h
```diff
@@ -433,9 +433,9 @@ class FunctionFactoryTest : public CxxTest::TestSuite {
   void test_MultiDomainFunction_creation_moreComplex() {
     const std::string fnString =
         "composite=MultiDomainFunction,NumDeriv=true;(name=FunctionFactoryTest_"
-        "FunctA,a0=0,a1=0.5;name=FunctionFactoryTest_FunctB,b0=0.1,b1=0.2,ties="
+        "FunctA,a0=0,a1=0.5;name=FunctionFactoryTest_FunctB,b0=0.1,ties="
```
martyngigg: It's not obvious to me why these values needed to change. Is something in the test dependent on them?

mantid-roman: This string is used in the check at line 447 and the output of `asString()` has changed so it outputs tied parameters only in the `ties=(...)` section.

martyngigg: Okay I see. I just to check that changing the string format isn't going to affect old scripts or saved nexus files or anything?

Framework/Crystal/test/OptimizeCrystalPlacementTest.h
```diff
@@ -105,7 +105,7 @@ class OptimizeCrystalPlacementTest : public CxxTest::TestSuite {
         d = angles5638[2] - fitInfoFix5637->Double(i, 1);
       else if (nm == "omega5638")
         d = angles5638[0] - fitInfoFix5637->Double(i, 1);
-      TS_ASSERT_DELTA(d, 0, .3);
+      TS_ASSERT_DELTA(d, 0, .4);
```
martyngigg: Was this necessary because of the changes made here?

Framework/API/test/FunctionFactoryTest.h
mantid-roman: It shouldn't break anything.

martyngigg:
I've just run these changes against the `PythonScriptsTest_CrystalFieldTest_Debug` and I see a much improved runtime and the debug version now doesn't hit the timeout limit on windows. :+1:

Framework/Crystal/test/OptimizeCrystalPlacementTest.h
mantid-roman: It was breaking on one of the platforms. I think it was for me only. So it was probably due to my changes.

martyngigg:
Ok. Once we may the switch to use a `unique_ptr` as commented then I'm happy for this to go in.

Framework/API/inc/MantidAPI/IFunction.h
```diff
@@ -583,6 +582,10 @@ class MANTID_API_DLL IFunction {
   Kernel::ProgressBase *m_progReporter;
 
 private:
+  /// No copying
+  IFunction(const IFunction&) = delete;
```
martyngigg: The preferred use of `= delete` is to put the declarations in the public section so that the compiler generates a better error message: http://stackoverflow.com/questions/18931133/must-a-deleted-constructor-be-private.

Can we move these?

Framework/API/inc/MantidAPI/ParameterReference.h
```diff
@@ -44,6 +44,8 @@ namespace API {
 class MANTID_API_DLL ParameterReference {
 public:
   ParameterReference();
+  ParameterReference(const ParameterReference&){}
```
martyngigg: What's the reason that we have a user-defined copy constructor that doesn't do anything along with an implicit copy-assignment operator?

Framework/API/src/IFunction.cpp
```diff
@@ -108,12 +109,16 @@ void IFunction::functionDeriv(const FunctionDomain &domain,
  * with this reference: a tie or a constraint.
  * @return newly ties parameters
  */
-ParameterTie *IFunction::tie(const std::string &parName,
-                             const std::string &expr, bool isDefault) {
-  auto ti = new ParameterTie(this, parName, expr, isDefault);
-  addTie(ti);
+void IFunction::tie(const std::string &parName, const std::string &expr,
+                    bool isDefault) {
+  auto ti = std::make_unique<ParameterTie>(this, parName, expr, isDefault);
```
martyngigg: I think you'll need to use `Kernel::make_unique` here. `std::make_unique` is not yet available on gcc 4.8 which is being used on RHEL6.

martyngigg:
This is good to go.

:shipit:

PR #18374, opened 2017-01-06
In & out of plane angle tool. 

### LINKED ISSUE #18120: In-plane and Out-of-plane angle calculation tool for Instrument View
Sub issue of #10144 

Another request from Pascal for alignment purposes: A tool on the instrument view to choose three peaks to define plane. Once the plane has been defined the user should be able to choose another peak and the tool will calculate both the in-plane and out-of-plane angles to the 4th peak.

samueljackson92:
Added a tool to the instrument view that allows the user to select three crystal peaks in detector space and use these to define a plane. A fourth peak can then be selected and the in plane and out of plane angle between the peaks will be calculated.

**To test:**

 - Load some single crystal data and open the instrument view
 - Create at least 4 single crystal peaks (or any points on the instrument)
 - Choose the new alignment tool
 - Select 3 peaks to form a plane
 - Select a 4th peak to show the in-plane (theta) and out-of-plane (phi) angle of that peak

Note: the phi angle should always be calculated relative to the beam direction for the instrument i.e. `phi==0` is the beam direction.

Fixes #18120 .

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


soininen:
@samueljackson92 Please resolve the merge conflicts.

soininen:
I tried testing this with some MANDI data that is available in Mantid's unit test data, but the 'Add single crystal peak' tool in the Instrument view does not seems to do anything. Any hints what I might be doing wrong here?

samueljackson92:
@soininen First click on a detector, then click on the 1D histogram at the bottom left of the screen where you see a peak to choose a position along the x axis.

soininen:
If I create several peaks and start clicking them with the new tool, the first three peaks get highlighted by a blue circle but afterwards the circles remain there while the phi and theta angles changes. Is this intended behavior? How do I know which peaks actually define the plane and which peak gives the angles?

soininen:
Another question: If I create a bunch of peaks close to each other so that the square selection(?) region of the angle tool covers several of them, I quickly run into theta: -nan phi: -nan, a state from which the tool does not seem to recover.

samueljackson92:
@soininen Yes. The three highlighted peaks are used as points which to define a plane in Q. Clicking a 4th peak should calculate the in plane and out of plane angles to the plane defined by those points. Start again you an right click -> clear plane.

I'll take a look at the second point. Are you seeing this when you're trying to define a plane (1st, 2nd, or 3rd click), or taking angles to it (4th click)?

soininen:
@samueljackson92 I did a bit more experimenting, and it seems the `-nan` issue is not about the closeness of the peaks. If I hold the mouse button and drag while selecting the either of first two reference plane peaks, the tool seems to get confused and give these `-nan` values.

samueljackson92:
@soininen I think you might be correct. I shall try...

MantidQt/MantidWidgets/src/InstrumentView/InstrumentWidgetPickTab.cpp
```diff
@@ -961,6 +983,60 @@ void ComponentInfoController::displayComparePeaksInfo(
...
 
+void ComponentInfoController::displayAlignPeaksInfo(
+    const std::vector<Mantid::Kernel::V3D> &planePeaks,
+    const Mantid::Geometry::IPeak *peak) {
+  if (planePeaks.size() < 3)
+    return;
+
+  auto pos1 = planePeaks[0];
+  auto pos2 = planePeaks[1];
+  auto pos3 = planePeaks[2];
```
soininen: Would it make sense to check here if the positions were unique? Otherwise you cannot define a plane.

soininen (changes requested):

MantidQt/MantidWidgets/src/InstrumentView/InstrumentWidgetPickTab.cpp
samueljackson92: I was actually thinking of doing it earlier on in ProjectionSurface, so that the event does not fire until you have uniquely defined points.

samueljackson92: But while we're here these could probably be `const`.

soininen:
By the way, would it make sense to highlight the 4th peak with similar circle as the first three but using a different color?

samueljackson92:
@soininen The `nan` issue should be corrected now, I fixed the merge conflicts, and I added a green highlight to the 4th peak.

MantidQt/MantidWidgets/src/InstrumentView/InstrumentWidgetPickTab.cpp
```diff
@@ -967,6 +989,61 @@ void ComponentInfoController::displayComparePeaksInfo(
...
+  auto v = pos3 - pos1;
+  auto n = u.cross_prod(v);
+
+  // find projection of beam direction onto plane
+  // this is so we always orientate to a common reference direction
+  const auto instrument = peak->getInstrument();
+  const auto samplePos = instrument->getSample()->getPos();
+  const auto sourcePos = instrument->getSource()->getPos();
+  const auto beam = samplePos - sourcePos;
+  const auto proj = beam - n * (beam.scalar_prod(n) / (pow(n.norm(), 2)));
```
AndreiSavici: use `n.norm2()`
Also, is it possible that `n` and `beam` are parallel?

docs/source/release/v3.9.0/ui.rst
```diff
@@ -13,6 +13,7 @@ Instrument View
 ###############
 
 - New peak comparison tool on the pick tab. The user can select two peaks and information relating to their properties and the angles between them.
+- New tool for computing in-plane and out-of-plane angles in Qlab space for crystal alignment.
```
AndreiSavici: Where is the documentation of what the angles mean? Do you have some formulas to compare to? My guess is that 4 peaks is too much. I would use 2 peaks + sample to define plane, but I am open to be persuaded (that's why I need documentation) that this is the correct approach

AndreiSavici (changes requested):

MantidQt/MantidWidgets/src/InstrumentView/InstrumentWidgetPickTab.cpp
samueljackson92: 1) Sure. 2) Good spot. It is but it's unlikely. I can put something in to handle that case.

docs/source/release/v3.9.0/ui.rst
samueljackson92: No and nothing specific. I was going to add an entry [here](http://www.mantidproject.org/MantidPlot:_Instrument_View) after the PR went through.

theta -> elevation angle from plane defined by the three peaks to the 4th peak
phi -> in plane angle with reference to the beam direction

Actually I think you're correct and you could just use 2 + sample here.

MantidQt/MantidWidgets/src/InstrumentView/InstrumentWidgetPickTab.cpp
```diff
@@ -967,6 +989,61 @@ void ComponentInfoController::displayComparePeaksInfo(
...
+
+  // now compute in plane & out of plane angles
+  const auto pos4 = peak->getQSampleFrame();
+  const auto x = pos4.scalar_prod(u);
+  const auto y = pos4.scalar_prod(v);
+  const auto z = pos4.scalar_prod(n);
+
+  Mantid::Kernel::V3D p(x, y, z);
+  // compute the elevation angle from the plane
+  auto R = p.norm();
```
soininen: Both `p` and `R` could be constants, I think.

samueljackson92: Yes.

soininen (approved):

MantidQt/MantidWidgets/src/InstrumentView/InstrumentWidgetPickTab.cpp
```diff
@@ -967,6 +989,69 @@ void ComponentInfoController::displayComparePeaksInfo(
...
+
+  // update in-plane vectors
+  v = u.cross_prod(n);
+
+  u.normalize();
+  v.normalize();
+  n.normalize();
+
+  // now compute in plane & out of plane angles
+  const auto pos4 = peak->getQSampleFrame();
```
AndreiSavici: Minor thing: you should rename this to pos3

AndreiSavici (approved):

AndreiSavici:
The changes are OK. Adding documentation to the instrument view on wiki is OK

AndreiSavici:
Ubuntu failures are unrelated.

martyngigg:
@AndreiSavici Is this ready to merge?

soininen:
@samueljackson92 The mouse dragging issue seems to be still there: if I hold down the mouse button, the same peak get selected twice for the reference plane and then all angles will be 0.

samueljackson92:
@soininen I can't reproduce this? For me, if I click & drag the peak is selected as blue. When clicking and dragging on the second peak it goes green. But that kind of makes sense, you selected it as a plane defining peak then also selected it (on the second second click event) to compare to the plane.

soininen:
@samueljackson92 On my system, if I drag the mouse over the first peak, I think the peak gets selected as the second peak as well. This does not make sense because then you cannot define the alignment plane and whatever peak you select as the third peak give zeros for theta and phi. If you mistakenly drag the mouse while selecting the first peak, you have to clear the alignment plane and start again.

samueljackson92:
@soininen Hmm. I agree that is probably what you are seeing, but I still can't reproduce even if I click and drag over the first peak. What instrument are you using? MANDI? Can you maybe send me a copy of the data and the peaks you're using?

soininen:
@samueljackson92 Can you first try it by just selecting the same peak twice (as the 1st and 2nd peak)? I'm indeed using a Mandi workspace here, and need to find a way to send it to you. I create the peaks quite randomly around the instrument view using the 'Add single crystal peak' tool.

soininen:
@samueljackson92 Hopefully the file is still available [here in Dropbox](https://www.dropbox.com/s/40kyog3a38n70en/MANDI_4376_event.nxs?dl=0)

samueljackson92:
@soininen That is what I'm doing. Both by individually (without holding) clicking on a peak, releasing,  then clicking again on the same peak immediately afterwards, and also separately by clicking, holding & dragging over a single peak. Neither way causes the issue you're seeing for me.

Thanks for the file!

samueljackson92:
@soininen Ah ha! Now I can reproduce :) - Seems that when MANDI data and selecting peaks you also don't get any red peak markers displayed either?

soininen:
@samueljackson92 I just checked that I'm running the correct commit and clean-build all, but the issue is still there. Does the tool accept the 2nd peak only if it differs from the 1st peak? Is there such a check?

soininen:
@samueljackson92 Red markers? I guess I don't :)

samueljackson92:
@soininen I think the issue is when I added the check to see if the peak is already in use I used Qlab instead of sample (doh!). On the dataset I was using these happened to be the same. In your dataset they are different. I will fix that now.

Yes you should see red markers indicating the position of the peak you selected and (if set) the HKL for the peak. But I think that might be a separate issue beyond scope of this work. I will check that also

samueljackson92:
@soininen Markers don't appear to show up in 3.8 either, so I assume that is an unrelated issue. I will make a new issue for that.

New commit should hopefully fix the problem you're seeing. Please kindly retest when you have time.

samueljackson92:
New issue for peak markers & event data is #18571

soininen:
The issues seem indeed be gone now. I think this is good to go :shipit:

martyngigg:
Is this still aiming for `3.9`?

samueljackson92:
Ideally.

martyngigg:
Ok. Can you change the base branch, please?

samueljackson92:
@martyngigg Done.

PR #18388, opened 2017-01-10
Store monitor flags in Beamline::DetectorInfo

### LINKED ISSUE #18372: Remove IDetector::markAsMonitor
Cleanup before refactoring with `DetectorInfo`. Use `Instrument::markAsMonitor` instead.

### LINKED ISSUE #18373: Python exports for monitor access via DetectorInfo
Part of refactoring for Instrument-2.0.

### LINKED ISSUE #18375: Move monitor flags into DetectorInfo
Consider making this read only -- should/could be defined at construction time.

SimonHeybrock:
`Beamline::DetectorInfo` now also stores monitor flags. The flags are no longer part of `Geometry::IDetector`. This is part of the Instrument-2.0 transition.
For the time being the flags are however also stored in the detector cache of the base instrument. This seemed to be the only reasonable way to support current code in Geometry and in particular to deal with the incremental initialization of `Geometry::Instrument` (adding detectors one by one), which would be very inefficient to obtain with `Beamline::DetectorInfo`.

**To test:**

Code review.
There should be no changes that affect loading and saving files, but to be on the safe side checking forward and backward compatibility would be a plus.

Fixes #18372.
Fixes #18373.
Fixes #18375.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/src/ExperimentInfo.cpp
+++ b/Framework/API/src/ExperimentInfo.cpp
@@ -166,7 +166,11 @@ makeDetectorInfo(const Instrument &oldInstr, const Instrument &newInstr) {
   } else {
     // If there is no DetectorInfo in the instrument we create a default one.
     const auto numDets = oldInstr.getNumberDetectors();
-    return Kernel::make_unique<Beamline::DetectorInfo>(numDets);
+    std::vector<size_t> monitors;
+    for (size_t i = 0; i < numDets; ++i)
+      if (newInstr.isMonitorViaIndex(i))
+        monitors.push_back(i);
+    return Kernel::make_unique<Beamline::DetectorInfo>(numDets, monitors);
   }
 }
 }
--- a/Framework/API/src/ExperimentInfo.cpp
+++ b/Framework/API/src/ExperimentInfo.cpp
@@ -166,6 +166,14 @@ makeDetectorInfo(const Instrument &oldInstr, const Instrument &newInstr) {
   } else {
     // If there is no DetectorInfo in the instrument we create a default one.
     const auto numDets = oldInstr.getNumberDetectors();
+    // Currently monitors flags are stored in the detector cache of the base
+    // instrument. The copy being made here is strictly speaking duplicating
+    // that data, but with future refactoring this will no longer be the case.
+    // Note that monitors will not change after creating a workspace.
+    // Instrument::markAsMonitor works only for the base instrument and it is
+    // not possible to obtain a non-const reference to the base instrument in a
+    // workspace. Thus we do not need to worry about the two copies of monitor
+    // flags running out of sync.
     std::vector<size_t> monitors;
     for (size_t i = 0; i < numDets; ++i)
       if (newInstr.isMonitorViaIndex(i))
--- a/Framework/Beamline/inc/MantidBeamline/DetectorInfo.h
+++ b/Framework/Beamline/inc/MantidBeamline/DetectorInfo.h
@@ -57,13 +57,17 @@ namespace Beamline {
 class MANTID_BEAMLINE_DLL DetectorInfo {
 public:
   DetectorInfo(const size_t numberOfDetectors);
+  DetectorInfo(const size_t numberOfDetectors,
+               const std::vector<size_t> &monitorIndices);
 
   size_t size() const;
 
+  bool isMonitor(const size_t index) const;
   bool isMasked(const size_t index) const;
   void setMasked(const size_t index, bool masked);
 
 private:
+  Kernel::cow_ptr<std::vector<bool>> m_isMonitor;
   Kernel::cow_ptr<std::vector<bool>> m_isMasked;
 };
 
--- a/Framework/Beamline/src/DetectorInfo.cpp
+++ b/Framework/Beamline/src/DetectorInfo.cpp
@@ -17,7 +17,11 @@ DetectorInfo::DetectorInfo(const size_t numberOfDetectors,
 
 /// Returns the size of the DetectorInfo, i.e., the number of detectors in the
 /// instrument.
-size_t DetectorInfo::size() const { return m_isMasked->size(); }
+size_t DetectorInfo::size() const {
+  if (!m_isMasked)
+    return 0;
+  return m_isMasked->size();
+}
 
 /// Returns true if the detector is a monitor.
 bool DetectorInfo::isMonitor(const size_t index) const {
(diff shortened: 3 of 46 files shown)


Framework/API/src/ExperimentInfo.cpp
```diff
@@ -166,7 +166,19 @@ makeDetectorInfo(const Instrument &oldInstr, const Instrument &newInstr) {
   } else {
     // If there is no DetectorInfo in the instrument we create a default one.
     const auto numDets = oldInstr.getNumberDetectors();
-    return Kernel::make_unique<Beamline::DetectorInfo>(numDets);
+    // Currently monitors flags are stored in the detector cache of the base
```
OwenArnold: Ok. This is a very small amount of data in any case. But I agree this should eventually live on the `DetectorInfo` rather than the `Instrument`

Framework/Beamline/inc/MantidBeamline/DetectorInfo.h
```diff
@@ -57,13 +57,17 @@ namespace Beamline {
...
 public:
   DetectorInfo(const size_t numberOfDetectors);
+  DetectorInfo(const size_t numberOfDetectors,
+               const std::vector<size_t> &monitorIndices);
 
   size_t size() const;
 
+  bool isMonitor(const size_t index) const;
   bool isMasked(const size_t index) const;
   void setMasked(const size_t index, bool masked);
```
OwenArnold: Do we also need to consider eventually `setMonitor` or would you do that at construction time? You could decide to use a detector or group of detectors as a monitor.

SimonHeybrock: Currently Mantid does not allow marking detectors as monitors in parameterized instrument. That is, once the instrument is in a workspace it is not possible to mark detectors as monitors, in can only be done at creation of the *base* instrument.

We might add `setMonitor` if required at a later time.

Framework/Beamline/src/DetectorInfo.cpp
```diff
@@ -5,13 +5,30 @@ namespace Mantid {
...
+    : DetectorInfo(numberOfDetectors) {
+  for (const auto i : monitorIndices)
+    m_isMonitor.access().at(i) = true;
+}
 
 /// Returns the size of the DetectorInfo, i.e., the number of detectors in the
 /// instrument.
-size_t DetectorInfo::size() const { return m_isMasked->size(); }
+size_t DetectorInfo::size() const {
+  if (!m_isMasked)
```
OwenArnold: How do you end up storing a null pointer for the masks? I don't see how that is currently allowed.

SimonHeybrock: `std::move` (move constructor is compiler generated).

Framework/Beamline/test/DetectorInfoTest.h
```diff
@@ -49,7 +61,30 @@ class DetectorInfoTest : public CxxTest::TestSuite {
...
+    TS_ASSERT(info.isMonitor(2));
+  }
+
+  void test_duplicate_monitors_ignored() {
+    std::vector<size_t> monitors{0, 0, 2, 2};
+    DetectorInfo info(3, monitors);
+    TS_ASSERT(info.isMonitor(0));
+    TS_ASSERT(!info.isMonitor(1));
+    TS_ASSERT(info.isMonitor(2));
   }
```
OwenArnold: 👍  for adding some simple tests.

Framework/CurveFitting/src/Algorithms/VesuvioCalculateMS.cpp
```diff
@@ -254,31 +255,28 @@ void VesuvioCalculateMS::cacheInputs() {
...
-  } else {
-    pixelShape = detPixel->shape();
-  }
+  // If is a detector group then take shape of first pixel
+  // All detectors in same bansk should be same shape anyway
+  // If the detector is a DetectorGroup, getID gives ID of first detector.
+  const auto &detectorInfo = m_inputWS->detectorInfo();
+  const size_t detIndex =
+      detectorInfo.indexOf(spectrumInfo.detector(index).getID());
+  const auto pixelShape = detectorInfo.detector(detIndex).shape();
```
OwenArnold: I think it would be nice to start hiding some of the `Detector` exposed functionality like `shape` and `getID` behind `DetectorInfo`, but I appreciate that that's not the purpose of this PR.

SimonHeybrock: `getID` should in my opinion not be part of `DetectorInfo`. We work only with indices internally.

Regarding shape, that is also a dependency issue: `Beamline` should in my opinion not depend on the old `Geometry` module. As a consequence `shape` cannot be added to `Beamline::DetectorInfo`. We could add it to `API::DetectorInfo`, but there would be follow up signature changes when eventually moving it into `Beamline::DetectorInfo`.

Framework/Beamline/inc/MantidBeamline/DetectorInfo.h
OwenArnold: Agreed. If we need it we can add it later. No reason to add it now.

Framework/Beamline/src/DetectorInfo.cpp
OwenArnold: Ok. That makes sense.

Framework/CurveFitting/src/Algorithms/VesuvioCalculateMS.cpp
SimonHeybrock: To elaborate on the IDs: Basically I would like to have not a single `detid_t` in the `Beamline` module. Detector IDs are a user/file thing and should be taken care of by the `Indexing` module (to be implemented).

OwenArnold: @SimonHeybrock. Yes. I agree, particularly to the extent that a `detid_t` cannot be used as a method argument.

Longer term, I think that the changes involved in the `FlatTree` and `ComponentProxy` way of mapping show that Parameter access and Geometry (Shape) access, should be things (IMHO) that are possible via the `DetectorInfo` although your dependency reasoning with Mantid is a very good reason not to do this at present.

OwenArnold:
:shipit:

PR #18394, opened 2017-01-10
Use git log rather than git describe to generate version number

martyngigg:
Description of work.

Updates the calculation of the version number to use `git log` only rather than `git describe` followed by `git log`. The previous state did not allow a build to function when using a shallow clone where no tags and been fetched. 

**To test:**

Make a new clone of mantid using the `--depth` option and try to run build kernel:

```
git clone --depth 50 https://github.com/mantidproject/mantid.git
cd mantid
mkdir build
cd build
cmake ../
cmake --build . --target Kernel
```

Before these changes the build would fail while compiling `MantidVersion.cpp`.

No issue number

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


SimonHeybrock:
Tried as suggested in PR, build failed. Downloaded and applied patch of this PR, compilation works. SHAs of both old and new git command are consistent. :shipit:

PR #18398, opened 2017-01-10
Fix unintentional modification of container workspace and make rebinning optional in ApplyPaalmanPingsCorrection

### LINKED ISSUE #18383: ApplyPaalmanPingsCorrection unexpectedly modifies the can input workspace
### Expected behavior
When running the script
```python
import numpy

xs = numpy.array([0.0, 1.0, 0.0, 1.1])
ys = numpy.array([2.2, 3.3])
sample_1 = CreateWorkspace(DataX=xs, DataY=ys, NSpec=2, UnitX='Wavelength')
ys = numpy.array([0.11, 0.22])
container_1 = CreateWorkspace(DataX=xs, DataY=ys, NSpec=2, UnitX='Wavelength')
corrected = ApplyPaalmanPingsCorrection(SampleWorkspace=sample_1, CanWorkspace=container_1)
```
one would expect the 'container_1' workspace to stay immutable over the call to `ApplyPaalmanPingsCorrection`.

### Actual behavior
`ApplyPaalmanPingsCorrection` will unconditionally run `RebinToWorkspace` on 'container_1' which, since the bins vary, will modify the binning of 'container_1'. This shouldn't happen as 'container_1' is an input property and as such should not be modified by the algorithm. Also, the modification will randomly delete the data behind the 'container_1' variable causing script failures.

soininen:
[`ApplyPaalmanPingsCorrection`](http://docs.mantidproject.org/nightly/algorithms/ApplyPaalmanPingsCorrection-v1.html) was accidentally applying [`RebinToWorkspace`](http://docs.mantidproject.org/nightly/algorithms/RebinToWorkspace-v1.html) to the input *CanWorkspace*. This PR fixes the behavior and also makes the rebinning optional as it is not always desirable. For example, if both *SampleWorkspace* and *CanWorkspace* have varying bins, `RebinToWorkspace` would rebin the container workspace incorrectly.

**To test:**

Before applying these changes, running
```python
import numpy

xs = numpy.array([0.0, 1.0, 0.0, 1.1])
ys = numpy.array([2.2, 3.3])
sample_1 = CreateWorkspace(DataX=xs, DataY=ys, NSpec=2, UnitX='Wavelength')
ys = numpy.array([0.11, 0.22])
container_1 = CreateWorkspace(DataX=xs, DataY=ys, NSpec=2, UnitX='Wavelength')
corrected = ApplyPaalmanPingsCorrection(SampleWorkspace=sample_1, CanWorkspace=container_1)
print(container_1)
```
would cause `container_1` to get rebinned, and sometimes raise an exception at the last `print` statement. This should not happen anymore.

Further, to test the optional rebinning, one could run the following script:
```python
import numpy

xs = numpy.array([0.0, 1.0, 0.0, 1.1])
ys = numpy.array([2.2, 3.3])
sample_1 = CreateWorkspace(DataX=xs, DataY=ys, NSpec=2, UnitX='Wavelength')
xs = numpy.array([-0.1, 0.5, 1.1, -0.1, 0.5, 1.2])
ys = numpy.array([0.11, 0.22, 0.33, 0.44])
container_1 = CreateWorkspace(DataX=xs, DataY=ys, NSpec=2, UnitX='Wavelength')
corrected = ApplyPaalmanPingsCorrection(SampleWorkspace=sample_1, CanWorkspace=container_1, RebinCanToSample='Rebin Can')
```
In the above, setting *RebinCanToSampe* to 'Do Not Rebin Can' should make the algorithm fail due to incompatible sample and container workspaces.

Fixes #18383.

*The release notes were updated accordingly.*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


SpencerHowells:
The RebinToWorkspace for the container is deliberately there to prevent the User from using different energy ranges and producing a rubbish result. Also, there is an option to shift the energy for the container which again requires a rebin. Making it optional is not what I want for our use.

soininen:
@SpencerHowells `RebinToWorkspace` does not work for our use cases where we have varying bins in the workspaces. `RebinToWorkspace` will take the binning of the first spectra of the sample workspace and apply it to all the spectra in the container workspace thus destroying the varying bins.

Further, *RebinCanToSample* is optional and its default value is set to enable rebinning so the algorithm works exactly as before, unless *RebinCanToSample* is deliberately set to 'Do Not Rebin Can'.

SpencerHowells:
I see your problem. You need to Rebin histogram by histogram. I've asked Nick Draper & Martyn Gigg about this.

soininen:
Good! Lets see if they have good ideas how to proceed or maybe we'll find another satisfactory solution. I can always revert the *RebinCanToSample* change.

soininen:
@SpencerHowells Any news from Nick or Martyn? Maybe I should also stress the point that we don't necessarily need rebinning at all. To my understanding, it doesn't make sense to measure the empty container using a different incident wavelength in a TOF instrument and thus the TOF axis should always match between the sample and the container workspaces.

SpencerHowells:
I agree with your comments. As I said I was forced to add it against by advice. historically I've kept it in because other routines would fail due to rounding errors in the x-units between sample & can. Nick was very wary about rebinning of what they call a ragged-workspace - that is, one in which the x-units for the histograms are different. Simple operations might work but many others will not. For your use I think an option would be the best solution.

soininen:
@SpencerHowells So, you agree that we can proceed with making the rebinning optional? As I said earlier, the change will not affect the current usage or behavior of this algorithm and rebinning has to be explicitly turned off if it is not needed.

SpencerHowells:
Yes I agree. Further to the ragged-ws problem, according to Nick it was originally thought of only for display/plotting use. If you need to rebin you will have to to do it hist by hist.

soininen:
@louisemccann I have fixed the flake8 warnings and all the tests seem to pass.

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/ApplyPaalmanPingsCorrection.py
```diff
@@ -49,8 +55,16 @@ def PyInit(self):
...
+        self.declareProperty(MatrixWorkspaceProperty('OutputWorkspace', '',
+                             direction=Direction.Output),
                              doc='The output corrections workspace.')
+        self.declareProperty(name='RebinCanToSample',
+                             defaultValue='Rebin Can',
+                             validator=StringListValidator([
+                                 'Rebin Can',
+                                 'Do Not Rebin Can']),
+                             doc=('Enable or disable RebinToWorkspace on ' +
+                                  'CanWorkspace.'))
```
louisemccann: Could this be done with a true/false checkbox instead of a string selector?

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/ApplyPaalmanPingsCorrection.py
```diff
@@ -65,7 +79,8 @@ def PyExec(self):
         prog_container.report('Starting algorithm')
 
         # Units should be wavelength
-        sample_unit = s_api.mtd[self._sample_ws_name].getAxis(0).getUnit().unitID()
+        sample_unit = \
+            s_api.mtd[self._sample_ws_name].getAxis(0).getUnit().unitID()
```
louisemccann: Is there any particular reason for this? Line length is 140, this should all be okay on one line.

louisemccann (changes requested):
It looks like the changes to the release notes got removed when you merged with master. 
Also could you add an option for this to the ApplyPaalmanPings interface

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/ApplyPaalmanPingsCorrection.py
soininen: Could be, but I rather use string selectors. It forces self-documenting code:
```python
ApplyPaalmanPingsCorrection('samplews', 'corrws', 'canws', 1.0, 0.0, 'outws', 'Rebin Can')
```
is easier to read than
```python
ApplyPaalmanPingsCorrection('samplews', 'corrws', 'canws', 1.0, 0.0, 'outws', True)
```
What is `True` here?

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/ApplyPaalmanPingsCorrection.py
soininen: This fails the 80 char limit on my `flake8` tool. Do we have a different standard in Mantid?

louisemccann: Here is the flake8 config file: https://github.com/mantidproject/mantid/blob/master/.flake8

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/ApplyPaalmanPingsCorrection.py
louisemccann: what is `1`?

louisemccann: When calling algorithms there is a tendency to use 
`ApplyPaalmanPingsCorrection(SampleWorkspace=sample_1, CanWorkspace=container_1, RebinCanToSample='True')`
so the code is still self-documented
It also means you can use get rid of:
```
if self.getProperty('RebinCanToSample').value == 'Rebin Can':
    self._rebin_container_ws = True
else:
    self._rebin_container_ws = False
```

soininen: Named arguments are not forced by Python so I'd expect there to be users who omit them. Anyway, I don't feel too strongly about this, so I'll make change the property to boolean.

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/ApplyPaalmanPingsCorrection.py
soininen: I see, thanks! I somehow thought `flake8` would pick up the config file automatically but it doesn't seem to.

soininen:
There are no release notes for 3.10, yet. Should we push this to 3.9 or add the release notes later, when they become available?

soininen:
Any hints what to do with the ApplyPaalmanPings interface? I could add a checkbox for container rebinning, but there is an additional check in `ApplyPaalmanPings::run()` if the binning matches, and if not, the user is shown a dialog and rebinning is done by manually running the `RebinToWorkspace` algorithm. 

As `ApplyPaalmanPingsCorrection` applies the rebinning anyway, should I just remove the additional rebinning step from the interface?

louisemccann:
@soininen wait until the release notes for 3.10. And yes, if the algorithm rebins the workspace anyway, there's no point calling `RebinToWorkspace` in the interface

louisemccann (approved):

louisemccann:
LGTM :shipit:

PR #18404, opened 2017-01-10
Automatically calculate ConvertToDiffractionMD extents

### LINKED ISSUE #18395: Automatically calculate extents in ConvertToDiffractionMDWorkspace
Pascal reported that he was getting weird extents in `BinMD` (+/-50 on each axis), even when using the calculate extents option. This is because `BinMD` calculates these extents on based on the size of the input workspace dimensions. 

Pascal had "weird" input workspace dimensions because he used the default option (+/-50) for the extents when converting to Q using `ConvertToDiffractionMDWorkspace`.

I think the best thing to do here is to ensure we use sensible limits when converting to Q in `ConvertToDiffractionMDWorkspace`. We already have an algorithm for this (`ConvertToMDMinMaxLocal`) so we should be able to just use that.

samueljackson92:
This changes the extents property on `ConvertToDiffractionMD2` to use `ConvertToMDMinMaxLocal` to calculate extents for the output MD workspace.

**To test:**
 - Check the unit & system tests are passing
 - Try running ConvertToDiffractionMD both with and entering values for the extents.
    - Check these work calculated extents are correct for the instrument attached to the workspace
 - Code review

Fixes #18395 .

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/MDAlgorithms/src/ConvertToDiffractionMDWorkspace2.cpp
```diff
@@ -100,12 +104,13 @@ void ConvertToDiffractionMDWorkspace2::init() {
       "order to merge them later\n");
   setPropertyGroup("MinRecursionDepth", getBoxSettingsGroupName());
 
-  std::vector<double> extents{-50, +50};
+  std::vector<double> extents;
   declareProperty(
```
OwenArnold: Do we want to create a new **version** of this algorithm for this change? This is going to give different results now.

Framework/MDAlgorithms/src/ConvertToDiffractionMDWorkspace2.cpp
```diff
@@ -247,5 +254,39 @@ void ConvertToDiffractionMDWorkspace2::exec() {
...
+  alg->setPropertyValue("dEAnalysisMode",
+                        dE_modes[Kernel::DeltaEMode::Elastic]);
+
+  std::string TargetFrame, Scaling;
+  convertFramePropertyNames(getPropertyValue("OutputDimensions"), TargetFrame,
+                            Scaling);
+  alg->setProperty("Q3DFrames", TargetFrame);
+  alg->setProperty("QConversionScales", Scaling);
+
+  alg->execute();
```
OwenArnold: How long does this typically take to run? This sub-algorithm is not reporting progress I think.

docs/source/release/v3.9.0/diffraction.rst
```diff
@@ -17,6 +17,10 @@ Crystal Improvements
 
 :ref:`algm-IntegrateEllipsoids` has a new option, AdaptiveQMultiplier, for the radius to vary as a function of the modulus of Q. If the AdaptiveQBackground option is set to True, the background radius also changes.  These are the same as the adaptive options in :ref:`algm-IntegratePeaksMD`.
 
+General Diffraction
+-------------------
+:ref:`algm-ConvertToDiffractionMD2` now automatically calculates extents using :ref:`algm-ConvertToMDMinMaxLocal` if none are provided.
+
 Single Crystal Diffraction
 --------------------------
 HB3A's IDF is modified to allow its detector center shifted from default position.
```
OwenArnold: The rst for the algorithms should also be updated. Very important that this information is explained in the help for this algorithm.

Framework/MDAlgorithms/src/ConvertToDiffractionMDWorkspace2.cpp
samueljackson92: On my machine and in elastic mode this takes ~0.12 which is negligible considering the time `ConvertToDiffractionMDWorkspace` takes to run.

Framework/MDAlgorithms/src/ConvertToDiffractionMDWorkspace2.cpp
samueljackson92: I was on the fence. On the one hand this isn't a huge change to the inner workings of the algorithm, but it does effect the algorithm's default parameters and therefore default behaviour. With fresh eyes this morning I'd be more inclined to move to v3.

docs/source/release/v3.9.0/diffraction.rst
samueljackson92: I did update the property docstring in the the algorithm itself. But perhaps I should add a note to the algorithm description as well?

Framework/MDAlgorithms/src/ConvertToDiffractionMDWorkspace2.cpp
OwenArnold: Ok. If you are confident that the processing time is not significant, sounds OK to me.

docs/source/release/v3.9.0/diffraction.rst
OwenArnold: Definitely. You can imagine an instrument scientist (at least I can) coming to you desk and asking you "how the hell is this algorithm calculating it's extents".

Framework/MDAlgorithms/src/ConvertToDiffractionMDWorkspace2.cpp
OwenArnold: I agree. I know it's a pain, but at least this way users can get the old behaviour back without installing an older version of Mantid. Try to share as much code as possible between the two version though.

samueljackson92: @OwenArnold agreed.

VickieLynch:
Works well.  I did find that the initial limits in the SliceViewer were not great, but this is not the fault of these changes. Left image is without limits, center is with limits (+/-25), right is ConvertToMD.
sliceviewer
:shipit:

samueljackson92:
@VickieLynch Is there an existing issue for the problem you've just pointed out?

PR #18405, opened 2017-01-10
Add WISH diffraction focussing reduction test

### LINKED ISSUE #17806: System Test for WISH diffraction focussing reduction
Pascal has provided a diffraction focussing reduction script for WISH processing. We should put this in as a system test so we can tell when stuff breaks.

samueljackson92:
Adding a system test for diffraction focussing for WISH.

**To test:**
 - Check system tests pass
 - Code review

Fixes #17806 .

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


gvardany (commented):
Is there a reason why you don't have a reference file to compare the output of the reduction and analysis to? Here you assert only the existence of the output and some of it's dimensions, but from system test I would expect to check also the values with the reference result with some reasonable tolerance.

samueljackson92:
@gvardany No justifiable reason. I think I originally copied from a test that only check dimensions, and I was being overly paranoid about files size, but to be honest the outputs of this are tiny compared to the input. I've added reference results for both tests.

gvardany:
@samueljackson92 Ok thanks, however I see you load and compare the reference file manually, whereas the stress test base class offers you this functionality for granted. Could you please make the validation following http://www.mantidproject.org/System_Tests ? There you will also find the correct location where to put the reference file.

samueljackson92:
@gvardany 👍  will move the results to reference folder. First test checks multiple files, but the `validate` method only appears to accept a single workspace to validate. How do we handle this? Save those reference workspaces as a group and load and compare that instead?

gvardany:
Yes, I think that's how it should work.

gvardany:
I am a bit confused with the files now: the "reference" is the expected output, so seems only `35991-foc-h00.nxs` and `WISHDiffractionFocussingResult.nxs` should be in reference folder.
While the rest are input data, so should stay in data. `35991-foc-h00.nxs` can be copied to both, seems it is a reference for one, and input for the other. Did you mean to run the reduction test for one run number? Seems there is a loop, but only one run.

samueljackson92:
@gvardany Yes I'm only using one file as input to the first test. The input file is very large. Ideally we'd run with three or four of them. We could remove the loop, but I didn't want to change this too far from the scientists original code.

gvardany:
Ok that's fine, the loop can stay. But please relocate the input files to avoid further confusion.

gvardany:
Happy to :shipit:

martyngigg:
Doc test failure can be ignored as it is the crystalfield heat capacity failure. All other pass.

PR #18418, opened 2017-01-12
Fix debug assertion when running tests in Windows debug mode

martyngigg:
Description of work

The MSVC standard library has quite strict checks in debug mode. Running the unit tests produced a
debug assertion for an iterator out of bounds error in `ExtractSpectraTest`. This fixes the code to avoid
that problem.

Note that the code was actually correct but this avoids it looking as if something is wrong.

**To test:**

Code review should suffice.

No issue number.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


quantumsteve:
LGTM. I have a slight preference for [std::next](http://en.cppreference.com/w/cpp/iterator/next), which would also address this issue. The result, however, is identical.

:shipit:

PR #18430, opened 2017-01-13
Stitch1DMany documentation improvement

### LINKED ISSUE #18200: Stitch1D and Stitch1DMany: more improvements to documentation
It was pointed out that these algorithms need to be documented better. They should include a note about restrictions/assumptions on the inputs (ascending x-axis, LHS, RHS ordering, partial overlap, etc). `Stitch1DMany` should include some explanation (probably with workflow diagrams) to indicate how it works under different inputs (i.e. no manual scale factor given, a manual scale factor given, scale factor from period, the difference between matrix workspaces and workspace groups, etc).

PranavBahuguna:
This PR updates and improves the documentation associated with the `Stitch1DMany` algorithm. Several of the new additions include:
- Updated descriptions for algorithm properties.
- Better description on how the various properties affect the function of the algorithm.
- Description of the constraints and assumptions made about the algorithm properties.
- Step-by-step description of the algorithm workflow.
- New workflow diagram.

Furthermore, some `Stitch1DMany` properties and validation have been altered and the corresponding unit tests were updated slightly.

Fixes #18200.

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/src/Stitch1DMany.cpp
+++ b/Framework/Algorithms/src/Stitch1DMany.cpp
@@ -39,11 +39,14 @@ void Stitch1DMany::init() {
 
   declareProperty(
       make_unique<ArrayProperty<double>>("StartOverlaps", Direction::Input),
-      "Start overlaps for stitched workspaces.");
+      "Start overlaps for stitched workspaces. If specified, the number of "
+      "StartOverlaps must be 1 less than the number of input workspaces. "
+      "Optional.");
 
   declareProperty(
       make_unique<ArrayProperty<double>>("EndOverlaps", Direction::Input),
-      "End overlaps for stitched workspaces.");
+      "End overlaps for stitched workspaces. If specified, the number of "
+      "EndOverlaps must be the same as the number of StartOverlaps. Optional.");
 
   declareProperty(make_unique<PropertyWithValue<bool>>("ScaleRHSWorkspace",
                                                        true, Direction::Input),
--- a/Framework/Algorithms/src/Stitch1DMany.cpp
+++ b/Framework/Algorithms/src/Stitch1DMany.cpp
@@ -206,9 +206,6 @@ std::map<std::string, std::string> Stitch1DMany::validateCommonInputs() {
     errors["EndOverlaps"] =
         "EndOverlaps must have the same number of entries as StartOverlaps.";
 
-  // if (m_params.empty())
-  //  errors["Params"] = "At least one parameter must be given.";
-
   m_useManualScaleFactor = this->getProperty("UseManualScaleFactor");
   m_manualScaleFactor = this->getProperty("ManualScaleFactor");
 
@@ -352,7 +349,8 @@ bool Stitch1DMany::processGroups() {
   std::string groupName = this->getProperty("OutputWorkspace");
   std::string outName;
 
-  // Determine whether or not we are using a global scale factor
+  // Determine whether or not we are scaling workspaces using scale factors from
+  // a specific period
   Property *manualSF = this->getProperty("ManualScaleFactor");
   bool usingScaleFromPeriod = m_useManualScaleFactor && manualSF->isDefault();
 
--- a/Framework/Algorithms/src/Stitch1DMany.cpp
+++ b/Framework/Algorithms/src/Stitch1DMany.cpp
@@ -76,7 +76,8 @@ void Stitch1DMany::init() {
                       "ScaleFactorFromPeriod", 1,
                       scaleFactorFromPeriodValidator, Direction::Input),
                   "Provided index of period to obtain scale factor from. "
-                  "Periods are indexed from 1.");
+                  "Periods are indexed from 1. Used only if stitching group "
+                  "workspaces and UseManualScaleFactor is false.");
 }
 
 /** Load and validate the algorithm's properties.
--- a/Framework/Algorithms/src/Stitch1DMany.cpp
+++ b/Framework/Algorithms/src/Stitch1DMany.cpp
@@ -154,7 +154,8 @@ void Stitch1DMany::validateGroupWorkspacesInputs() {
 
 /** Load and validate properties common to both group and non-group workspaces.
 */
-void Stitch1DMany::validateCommonInputs(std::map<std::string, std::string> &errors) {
+void Stitch1DMany::validateCommonInputs(
+    std::map<std::string, std::string> &errors) {
   const std::vector<std::string> inputWorkspacesStr =
       this->getProperty("InputWorkspaces");
   if (inputWorkspacesStr.size() < 2)
--- a/Framework/Algorithms/src/Stitch1DMany.cpp
+++ b/Framework/Algorithms/src/Stitch1DMany.cpp
@@ -177,14 +177,18 @@ std::map<std::string, std::string> Stitch1DMany::validateCommonInputs() {
     errors["InputWorkspaces"] = "At least 2 input workspaces required.";
 
   // Check that all the workspaces are of the same type
-  const std::string id = m_inputWSMatrix[0][0]->id();
-  for (auto &period : m_inputWSMatrix) {
-    for (auto &inputWS : period) {
-      if (inputWS->id() != id) {
-        errors["InputWorkspaces"] = "All workspaces must be the same type.";
-        break;
+  if (!m_inputWSMatrix.empty() && !m_inputWSMatrix[0].empty()) {
+    const std::string id = m_inputWSMatrix[0][0]->id();
+    for (auto &period : m_inputWSMatrix) {
+      for (auto &inputWS : period) {
+        if (inputWS->id() != id) {
+          errors["InputWorkspaces"] = "All workspaces must be the same type.";
+          break;
+        }
       }
     }
+  } else {
+    return errors; // Can't validate other properties without enough workspaces
   }
 
   m_startOverlaps = this->getProperty("StartOverlaps");
--- a/Framework/Algorithms/src/Stitch1DMany.cpp
+++ b/Framework/Algorithms/src/Stitch1DMany.cpp
@@ -76,7 +76,8 @@ void Stitch1DMany::init() {
                       scaleFactorFromPeriodValidator, Direction::Input),
                   "Provided index of period to obtain scale factor from. "
                   "Periods are indexed from 1. Used only if stitching group "
-                  "workspaces and UseManualScaleFactor is false.");
+                  "workspaces, UseManualScaleFactor is true and "
+                  "ManualScaleFactor is set to default.");
 }
 
 /** Load and validate the algorithm's properties.
--- a/Framework/Algorithms/src/Stitch1DMany.cpp
+++ b/Framework/Algorithms/src/Stitch1DMany.cpp
@@ -4,6 +4,7 @@
 #include "MantidAPI/WorkspaceProperty.h"
 #include "MantidKernel/ArrayProperty.h"
 #include "MantidKernel/BoundedValidator.h"
+#include "MantidKernel/MandatoryValidator.h"
 #include "MantidKernel/RebinParamsValidator.h"
 
 #include <boost/make_shared.hpp>
@@ -20,16 +21,19 @@ DECLARE_ALGORITHM(Stitch1DMany)
 void Stitch1DMany::init() {
 
   declareProperty(
-      make_unique<ArrayProperty<std::string>>("InputWorkspaces",
-                                              Direction::Input),
-      "Input Workspaces. List of histogram workspaces to stitch together.");
+      Kernel::make_unique<ArrayProperty<std::string>>(
+          "InputWorkspaces",
+          boost::make_shared<MandatoryValidator<std::vector<std::string>>>()),
+      "Input Workspaces. List of histogram workspaces to stitch together. At "
+      "least 2 workspaces must be supplied for stitching and all must be of "
+      "the same type.");
 
   declareProperty(make_unique<WorkspaceProperty<Workspace>>(
                       "OutputWorkspace", "", Direction::Output),
                   "Output stitched workspace.");
 
   declareProperty(make_unique<ArrayProperty<double>>(
-                      "Params", boost::make_shared<RebinParamsValidator>(true),
+                      "Params", boost::make_shared<RebinParamsValidator>(false),
                       Direction::Input),
                   "Rebinning Parameters. See Rebin for format.");
 
@@ -68,7 +72,8 @@ void Stitch1DMany::init() {
   declareProperty(make_unique<PropertyWithValue<int>>(
                       "ScaleFactorFromPeriod", 1,
                       scaleFactorFromPeriodValidator, Direction::Input),
-                  "Provided index of period to obtain scale factor from.");
+                  "Provided index of period to obtain scale factor from. "
+                  "Periods are indexed from 1.");
 }
 
 /** Load and validate the algorithm's properties.
@@ -168,18 +173,14 @@ std::map<std::string, std::string> Stitch1DMany::validateCommonInputs() {
     errors["InputWorkspaces"] = "At least 2 input workspaces required.";
 
   // Check that all the workspaces are of the same type
-  if (!m_inputWSMatrix.empty() && !m_inputWSMatrix[0].empty()) {
-    const std::string id = m_inputWSMatrix[0][0]->id();
-    for (auto &period : m_inputWSMatrix) {
-      for (auto &inputWS : period) {
-        if (inputWS->id() != id) {
-          errors["InputWorkspaces"] = "All workspaces must be the same type.";
-          break;
-        }
+  const std::string id = m_inputWSMatrix[0][0]->id();
+  for (auto &period : m_inputWSMatrix) {
+    for (auto &inputWS : period) {
+      if (inputWS->id() != id) {
+        errors["InputWorkspaces"] = "All workspaces must be the same type.";
+        break;
       }
     }
-  } else {
-    errors["InputWorkspaces"] = "Input workspaces must be given";
   }
 
   m_startOverlaps = this->getProperty("StartOverlaps");
@@ -201,8 +202,8 @@ std::map<std::string, std::string> Stitch1DMany::validateCommonInputs() {
     errors["EndOverlaps"] =
         "EndOverlaps must have the same number of entries as StartOverlaps.";
 
-  if (m_params.empty())
-    errors["Params"] = "At least one parameter must be given.";
+  // if (m_params.empty())
+  //  errors["Params"] = "At least one parameter must be given.";
 
   m_useManualScaleFactor = this->getProperty("UseManualScaleFactor");
   m_manualScaleFactor = this->getProperty("ManualScaleFactor");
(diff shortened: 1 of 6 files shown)


docs/source/algorithms/Stitch1DMany-v1.rst
```diff
@@ -13,22 +13,72 @@ Stitches single histogram :ref:`Matrix Workspaces <MatrixWorkspace>`
...
+The algorithm workflow is as follows:
+
+#. A check is performed to find out whether the input workspaces are group
+   workspaces or not. The algorithm handles matrix workspaces differently from
+   group workspaces.
+#. If matrix workspaces are supplied, the algorithm simply iterates over each
+   workspace and calls the :literal:`Stitch1D` algorithm, each RHS workspace to
+   the LHS workspace to form a single stitched workspace. The resultant
+   workspace and its scale factor are outputted.
+#. If group workspaces are supplied, the algorithm checks whether a global scale
```
raquelalvarezbanos: What do you mean by "global scale factor"? Do you mean "manual" or do you mean a factor that is the same for all pairs of workspaces?

docs/source/algorithms/Stitch1DMany-v1.rst
```diff
@@ -13,22 +13,72 @@ Stitches single histogram :ref:`Matrix Workspaces <MatrixWorkspace>`
...
+   returns a vector of period scale factors obtained by stitching workspaces
+   from a specific period.
+#. The algorithm iterates over each workspace for each period across all groups
+   and calls :literal:`Stitch1D`, passing the scale factor from period scale
+   factors for each period index. Like in step 4, the stitched workspaces are
+   stored in a vector while the out scale factors are appended and outputted.
+   Finally step 5 is called, grouping the workspaces into a single one that is
+   outputted.
+
+.. diagram:: Stitch1DMany-v1_wkflw.dot
```
raquelalvarezbanos: I am not sure about the workflow diagram either, see my previous comment.

Framework/Algorithms/src/Stitch1DMany.cpp
```diff
@@ -20,26 +21,32 @@ DECLARE_ALGORITHM(Stitch1DMany)
...
+      "least 2 workspaces must be supplied for stitching and all must be of "
+      "the same type.");
 
   declareProperty(make_unique<WorkspaceProperty<Workspace>>(
                       "OutputWorkspace", "", Direction::Output),
                   "Output stitched workspace.");
 
   declareProperty(make_unique<ArrayProperty<double>>(
-                      "Params", boost::make_shared<RebinParamsValidator>(true),
+                      "Params", boost::make_shared<RebinParamsValidator>(false),
```
PranavBahuguna: The `Params` property was effectively mandatory in the original code. I believed it would be better to change the `Params` validator to make entering it mandatory rather than let the `validateCommonInputs()` method do it, as it makes it more explicit in the documentation that this property is mandatory.

Framework/Algorithms/src/Stitch1DMany.cpp
```diff
@@ -168,18 +177,14 @@ std::map<std::string, std::string> Stitch1DMany::validateCommonInputs() {
     errors["InputWorkspaces"] = "At least 2 input workspaces required.";
 
   // Check that all the workspaces are of the same type
-  if (!m_inputWSMatrix.empty() && !m_inputWSMatrix[0].empty()) {
```
PranavBahuguna: The `InputWorkspaces` property was given a validator that made entering at least one workspace for it mandatory. Therefore it would not be possible for `m_inputWSMatrix` to be completely empty, so the check is unnecessary. Like `Params` the reason for attaching a validator in the first place was to make it explicit in the documentation that input workspaces were mandatory.

Framework/Algorithms/test/Stitch1DManyTest.h
```diff
@@ -175,30 +180,17 @@ class Stitch1DManyTest : public CxxTest::TestSuite {
...
-    AnalysisDataService::Instance().add("ws2", ws2);
+    AnalysisDataService::Instance().addOrReplace("ws1", ws1);
+    AnalysisDataService::Instance().addOrReplace("ws2", ws2);
     Stitch1DMany alg;
     alg.setChild(true);
     alg.initialize();
     alg.setProperty("InputWorkspaces", "ws1, ws2");
     alg.setProperty("Params", "0.1");
-    alg.setProperty("StartOverlaps", "-0.5, -0.6");
-    alg.setProperty("EndOverlaps", "0.5, 0.6");
```
PranavBahuguna: They weren't necessary for the tests. The same goes for the other ones.

docs/source/algorithms/Stitch1DMany-v1.rst
```diff
@@ -13,22 +13,72 @@ Stitches single histogram :ref:`Matrix Workspaces <MatrixWorkspace>`
...
+--------
+
+The algorithm workflow is as follows:
+
+#. A check is performed to find out whether the input workspaces are group
+   workspaces or not. The algorithm handles matrix workspaces differently from
+   group workspaces.
+#. If matrix workspaces are supplied, the algorithm simply iterates over each
+   workspace and calls the :literal:`Stitch1D` algorithm, each RHS workspace to
+   the LHS workspace to form a single stitched workspace. The resultant
```
PranavBahuguna: That's true, I should probably add that in.

Framework/Algorithms/test/Stitch1DManyTest.h
raquelalvarezbanos: :+1:

Framework/Algorithms/src/Stitch1DMany.cpp
```diff
@@ -201,8 +206,8 @@ std::map<std::string, std::string> Stitch1DMany::validateCommonInputs() {
     errors["EndOverlaps"] =
         "EndOverlaps must have the same number of entries as StartOverlaps.";
 
-  if (m_params.empty())
-    errors["Params"] = "At least one parameter must be given.";
+  // if (m_params.empty())
+  //  errors["Params"] = "At least one parameter must be given.";
```
raquelalvarezbanos: Dead code?

Framework/Algorithms/src/Stitch1DMany.cpp
raquelalvarezbanos: :+1:

Framework/Algorithms/src/Stitch1DMany.cpp
raquelalvarezbanos: Can you remind me what `false` means in this case for a `RebinParamsValidator`?

PranavBahuguna: Its a condition  if set true, that allows the array it is validating to be empty without throwing an error.

Framework/Algorithms/src/Stitch1DMany.cpp
PranavBahuguna: Yep, being removed.

Framework/Algorithms/src/Stitch1DMany.cpp
raquelalvarezbanos: OK, thanks. Perfect then.

docs/source/algorithms/Stitch1DMany-v1.rst
```diff
@@ -13,22 +13,72 @@ Stitches single histogram :ref:`Matrix Workspaces <MatrixWorkspace>`
...
+   workspace and its scale factor are outputted.
```
PranavBahuguna: Yes I think. With matrix workspaces, they are simply passed straight to the Stitch1D algorithm, whereas for group workspaces they are used to determine whether or not to use ScaleFactorFromPeriod.

raquelalvarezbanos: Could you check this? I am trying this script:

```
CreateWorkspace(OutputWorkspace='ws_1_1', DataX='1,2,3,4,5,6,7,8,9,10', DataY='1,1,1,1,1,1,1,1,1')
CreateWorkspace(OutputWorkspace='ws_1_2', DataX='3,4,5,6,7,8,9,10,11,12', DataY='2,2,2,2,2,2,2,2,2')
CreateWorkspace(OutputWorkspace='ws_1_3', DataX='5,6,7,8,9,10,11,12,13,14', DataY='4,4,4,4,4,4,4,4,4')
stitched, scale_factors = Stitch1DMany(InputWorkspaces='ws_1_1,ws_1_2,ws_1_3', Params='1', UseManualScaleFactor=True, ManualScaleFactor=1)
print scale_factors
stitched, scale_factors = Stitch1DMany(InputWorkspaces='ws_1_1,ws_1_2,ws_1_3', Params='1')
print scale_factors
```

and the output is:

```
[ 1.  1.]
[ 0.5   0.25]
```

and it seems to me that `UseManualScaleFactor` with `ManualScaleFactor` to have an effect, at least on the calculated scale factors.

docs/source/algorithms/Stitch1DMany-v1.rst
PranavBahuguna: I should probably rewrite this to make it more clear (the comments in the code were in fact wrong). There's no global scale factor, rather that if we are not using `ScaleFactorFromPeriod`, the workspaces for each period across all groups will be stitched together using Stitch1D. `UseManualScaleFactor` and `ManualScaleFactor` will be passed to them, but `UseManualScaleFactor` may not necessarily be true.

Framework/Algorithms/src/Stitch1DMany.cpp
PranavBahuguna: Actually I was wrong, its still possible for it to be empty. Placing the check back in now, though I will implement it differently from before.

raquelalvarezbanos: How can it be empty if it has a validator has been added?

PranavBahuguna: Another way it may be empty is if InputWorkspaces is set to an array of workspace name strings, but those workspaces don't actually exist in the ADS. It will pass the validator check, but won't be added in `m_inputWSMatrix` - I overlooked this previously.

raquelalvarezbanos: OK, that makes sense.

Framework/Algorithms/src/Stitch1DMany.cpp
```diff
@@ -68,7 +74,9 @@ void Stitch1DMany::init() {
   declareProperty(make_unique<PropertyWithValue<int>>(
                       "ScaleFactorFromPeriod", 1,
                       scaleFactorFromPeriodValidator, Direction::Input),
-                  "Provided index of period to obtain scale factor from.");
+                  "Provided index of period to obtain scale factor from. "
+                  "Periods are indexed from 1. Used only if stitching group "
+                  "workspaces and UseManualScaleFactor is false.");
```
raquelalvarezbanos: I am not sure that `ScaleFactorFromPeriod` is used if `UseManualScaleFactor` is `False`, I think it is the other way around: `UseManualScaleFactor` has to be `True` in order for the algorithm to use `ScaleFactorFromPeriod`.

The following script illustrates this:

```
import numpy as np

def straight_line(x, a, b):
  return a+b*x

#create two histograms
x1 = np.arange(-1, 1, 0.02)
x2 = np.arange(0.4, 1.6, 0.02)
x3 = np.arange(1.3, 3, 0.02)
ws_1_1 = CreateWorkspace(UnitX="1/q", DataX=x1, DataY=straight_line(x1[:-1], 1, 1)+1)
ws_2_1 = CreateWorkspace(UnitX="1/q", DataX=x2, DataY=straight_line(x2[:-1], 1, 1)+1)
ws_3_1 = CreateWorkspace(UnitX="1/q", DataX=x3, DataY=straight_line(x3[:-1], 1, 1)+1)
ws_1_2 = CreateWorkspace(UnitX="1/q", DataX=x1, DataY=straight_line(x1[:-1], 1, 1)+1)
ws_2_2 = CreateWorkspace(UnitX="1/q", DataX=x2, DataY=straight_line(x2[:-1], 1.5, 1)+1)
ws_3_2 = CreateWorkspace(UnitX="1/q", DataX=x3, DataY=straight_line(x3[:-1], 2, 1)+1)
ws_1_3 = CreateWorkspace(UnitX="1/q", DataX=x1, DataY=straight_line(x1[:-1], 1, 1)+1)
ws_2_3 = CreateWorkspace(UnitX="1/q", DataX=x2, DataY=straight_line(x2[:-1], 2, 1)+1)
ws_3_3 = CreateWorkspace(UnitX="1/q", DataX=x3, DataY=straight_line(x3[:-1], 3, 1)+1)
group1 = GroupWorkspaces('ws_1_1, ws_1_2, ws_1_3')
group2 = GroupWorkspaces('ws_2_1, ws_2_2, ws_2_3')
group3 = GroupWorkspaces('ws_3_1, ws_3_2, ws_3_3')

stitched_from_period, factors = Stitch1DMany('group1, group2, group3', Params=[0.02], UseManualScaleFactor=True, ScaleFactorFromPeriod=1)
print factors
stitched_from_period, factors = Stitch1DMany('group1, group2, group3', Params=[0.02], UseManualScaleFactor=False, ScaleFactorFromPeriod=1)
print factors
```

In the first `Stitch1DMany` I set `UseManualScaleFactor=True` and the rest of the periods are stitched with scale factor `1` (because data in the first period of each group is the same, so their scale factor is `1`). In the next line, `UseManualScaleFactor=True` and the output scale factors are different for the different periods.

Framework/Algorithms/src/Stitch1DMany.cpp
raquelalvarezbanos: Are you finally going to add this back?

docs/source/algorithms/Stitch1DMany-v1.rst
```diff
@@ -13,22 +13,74 @@ Stitches single histogram :ref:`Matrix Workspaces <MatrixWorkspace>`
...
+The algorithm workflow is as follows:
+
+#. A check is performed to find out whether the input workspaces are group
+   workspaces or not. The algorithm handles matrix workspaces differently from
+   group workspaces.
+#. If matrix workspaces are supplied, the algorithm simply iterates over each
+   workspace and calls the :literal:`Stitch1D` algorithm. This stitches each RHS
+   workspace to the LHS workspace to form a single stitched workspace (LHS to
+   RHS if ScaleRHSWorkspace is set to false). The resultant workspace and its
+   scale factor are outputted.
```
raquelalvarezbanos: Did you check if `UseManualScaleFactor` and `ManualScaleFactor` apply to matrix workspaces too? From this sentence it seems that they don't. The same seems from the workflow diagram.

docs/source/algorithms/Stitch1DMany-v1.rst
```diff
@@ -13,22 +13,74 @@ Stitches single histogram :ref:`Matrix Workspaces <MatrixWorkspace>`
...
+#. If group workspaces are supplied, the algorithm checks whether or not to
+   scale workspaces using scale factors from a specific period (given by
+   :literal:`ScaleFactorFromPeriod`). This is done if either
+   :literal:`UseManualScaleFactor` is false, or :literal:`UseManualScaleFactor`
```
raquelalvarezbanos: The same as above. I think `ScaleFactorFromPeriod` is not used if `UseManualScaleFactor` is `False`. I think `ScaleFactorFromPeriod` is used if `UseManualScaleFactor` is `True` and `ManualScaleFactor` is set to its default value.

docs/source/algorithms/Stitch1DMany-v1.rst
```diff
@@ -13,22 +13,74 @@ Stitches single histogram :ref:`Matrix Workspaces <MatrixWorkspace>`
...
+   is true and :literal:`ManualScaleFactor` is set to its default value.
+#. If not using :literal:`ScaleFactorFromPeriod`, the algorithm collects the
```
raquelalvarezbanos: I would say here "If UseManualScaleFactor is set to false, the algorithm collects...". The reason is that I think the process you are describing is what happens when `UseManualScaleFactor` is `False` (regardless of `ManualScaleFactor` and `ScaleFactorFromPeriod`).

raquelalvarezbanos (commented):
I am not still sure about a few things mentioned in the documentation, could you have a look and let me know if you disagree?

Framework/Algorithms/src/Stitch1DMany.cpp
PranavBahuguna: Yes that's a mistake, correcting it now.

raquelalvarezbanos: :+1:

docs/source/algorithms/Stitch1DMany-v1.rst
PranavBahuguna: All of the input parameters (except `ScaleFactorFromPeriod`) are simply passed straight to `Stitch1D` alg. The workflow diagram only shows the `UseManualScaleFactor`, `ManualScaleFactor` and `ScaleFactorFromPeriod` parameters as they are the only ones with any influence on the progression of the algorithm.

docs/source/algorithms/Stitch1DMany-v1.rst
PranavBahuguna: Agreed, changing it now.

docs/source/algorithms/Stitch1DMany-v1.rst
PranavBahuguna: This scenario (not using `ScaleFactorFromPeriod`) could still happen however if `UseManualScaleFactor` was `True`, but `ManualScaleFactor` was not set to its default value. I don't think this should be changed as the conditions for when it uses `ScaleFactorFromPeriod` had already been explained beforehand.

Framework/Algorithms/src/Stitch1DMany.cpp
PranavBahuguna: The new `ADSValidator` dealt with the problem I mentioned before, so it isn't needed anymore.

Framework/Algorithms/src/Stitch1DMany.cpp
raquelalvarezbanos: Was this corrected?

docs/source/algorithms/Stitch1DMany-v1.rst
raquelalvarezbanos: What you said above is perfect: "All of the input parameters (except `ScaleFactorFromPeriod`) are simply passed straight to `Stitch1D`". I'd just add this note to this point.

docs/source/algorithms/Stitch1DMany-v1.rst
raquelalvarezbanos: Was it changed?

docs/source/algorithms/Stitch1DMany-v1.rst
raquelalvarezbanos: OK.

raquelalvarezbanos:
Looks good :shipit:

martyngigg:
Docs test failure is due to current problem with CrystalField doc test that is being fixed.

PR #18433, opened 2017-01-13
Added an input property for temperature in ComputeCalibrationCoefVan

### LINKED ISSUE #18429: Temperature for ComputeCalibrationCoefVan is only read from the sample logs
The sample temperature information in [`ComputeCalibrationCoefVan`](http://docs.mantidproject.org/nightly/algorithms/ComputeCalibrationCoefVan-v1.html) is read from the sample log entry 'temperature'. If the log is not present, the temperature needed for the Debye-Waller correction is fixed to 293K. The temperature should be settable by an input property as well.

soininen:
This PR adds a *Temperature* input property to [`ComputeCalibratonCoefVan`](http://docs.mantidproject.org/nightly/algorithms/ComputeCalibrationCoefVan-v1.html). The temperature is used in correcting the integrated Vanadium data for the Debye-Waller factor. If a 'temperature' sample log exists in the *InputWorkspace*, the value given by the input property will override it.

**To test:**

1. Load `<mantid-build-dir>/ExternalData/Testing/Data/UnitTest/TOFTOFTestdata.nxs'
2. Check that the 'temperature' entry exists in the sample logs.
3. Run [`FindEPP`](http://docs.mantidproject.org/nightly/algorithms/FindEPP-v1.html) on the workspace.
4. Run `ComputeCalibrationCoefVan` without specifying *Temperature*. The algorithm should use the value in the sample logs.
5. Run `ComputeCalibrationCoefVan` with different *Temperature* than what is in the sample logs.

Fixes #18429.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/PythonInterface/plugins/algorithms/ComputeCalibrationCoefVan.py
+++ b/Framework/PythonInterface/plugins/algorithms/ComputeCalibrationCoefVan.py
@@ -58,8 +58,9 @@ class ComputeCalibrationCoefVan(PythonAlgorithm):
                              defaultValue=Property.EMPTY_DBL,
                              validator=FloatBoundedValidator(lower=0.0),
                              direction=Direction.Input,
-                             doc=("Temperature during the experiment, in " +
-                                  "Kelvins."))
+                             doc=("Temperature during the experiment (in " +
+                                  "Kelvins) if the 'temperature' sample log " +
+                                  "is missing or needs to be overriden."))
         return
 
     def validateInputs(self):
--- a/Framework/PythonInterface/plugins/algorithms/ComputeCalibrationCoefVan.py
+++ b/Framework/PythonInterface/plugins/algorithms/ComputeCalibrationCoefVan.py
@@ -58,7 +58,8 @@ class ComputeCalibrationCoefVan(PythonAlgorithm):
                              defaultValue=Property.EMPTY_DBL,
                              validator=FloatBoundedValidator(lower=0.0),
                              direction=Direction.Input,
-                             doc=("Temperature during the experiment."))
+                             doc=("Temperature during the experiment, in " +
+                                  "Kelvins."))
         return
 
     def validateInputs(self):
--- a/Framework/PythonInterface/plugins/algorithms/ComputeCalibrationCoefVan.py
+++ b/Framework/PythonInterface/plugins/algorithms/ComputeCalibrationCoefVan.py
@@ -3,7 +3,7 @@ from __future__ import (absolute_import, division, print_function)
 from mantid.api import (PythonAlgorithm, AlgorithmFactory,
                         MatrixWorkspaceProperty, Progress, InstrumentValidator,
                         ITableWorkspaceProperty)
-from mantid.kernel import Direction
+from mantid.kernel import Direction, FloatBoundedValidator, Property
 import numpy as np
 from scipy import integrate
 import scipy as sp
@@ -54,9 +54,16 @@ class ComputeCalibrationCoefVan(PythonAlgorithm):
                              direction=Direction.Output),
                              ("Name the workspace that will contain the " +
                               "calibration coefficients"))
+        self.declareProperty("Temperature",
+                             defaultValue=Property.EMPTY_DBL,
+                             validator=FloatBoundedValidator(lower=0.0),
+                             direction=Direction.Input,
+                             doc=("Temperature during the experiment."))
         return
 
     def validateInputs(self):
+        """ Validate the inputs
+        """
         issues = dict()
         inws = self.getProperty("VanadiumWorkspace").value
         run = inws.getRun()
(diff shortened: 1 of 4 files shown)


Framework/PythonInterface/plugins/algorithms/ComputeCalibrationCoefVan.py
```diff
@@ -54,9 +54,16 @@ def PyInit(self):
                              direction=Direction.Output),
                              ("Name the workspace that will contain the " +
                               "calibration coefficients"))
+        self.declareProperty("Temperature",
+                             defaultValue=Property.EMPTY_DBL,
+                             validator=FloatBoundedValidator(lower=0.0),
+                             direction=Direction.Input,
+                             doc=("Temperature during the experiment."))
```
NickDraper: We need the unit defined here, I guess it is Kelvin

Framework/PythonInterface/plugins/algorithms/ComputeCalibrationCoefVan.py
```diff
@@ -85,21 +92,23 @@ def validateInputs(self):
...
         run = self.vanaws.getRun()
         if not run.hasProperty('temperature'):
-            self.log().warning("Temperature sample log is not present in " +
+            self.log().warning("No Temperature given and the 'temperature' " +
+                               "sample log is not present in " +
                                self.vanaws.name() +
                                " T=293K is assumed for Debye-Waller factor.")
             return self.defaultT
         try:
             temperature = float(run.getProperty('temperature').value)
```
NickDraper: This will only work if the temperature is a single value, if it is a time series this will return a string with "Time  Value" which will fail the conversion to a float.  It might be checking if it is a Time series and then using timeAverageValue()

Framework/PythonInterface/plugins/algorithms/ComputeCalibrationCoefVan.py
```diff
@@ -54,9 +54,16 @@ def PyInit(self):
...
+                             direction=Direction.Input,
+                             doc=("Temperature during the experiment."))
         return
 
     def validateInputs(self):
+        """ Validate the inputs
+        """
         issues = dict()
         inws = self.getProperty("VanadiumWorkspace").value
         run = inws.getRun()
```
NickDraper: This line needs some protection, if you have a table workspace selected in the workspace list when you bring up the algorithm dialog box, then inws will be None, and will crash Mantid with a 'NoneType' object has no attribute 'getRun'.

This happened to me several times while testing.

NickDraper:
Other than the comments above this seems ok, it might not be worth worrying too much about time series temperature logs as this algorithm also requires a single value wavelength log.

Framework/PythonInterface/plugins/algorithms/ComputeCalibrationCoefVan.py
soininen: Huh? I cannot reproduce this on my system, Ubuntu 14.04... I'll see what I can do.

NickDraper: Interestingly, I'm having trouble recreating it now.  The first time I didn't have all the workspaces in the right format for a few times before I got it right.

I'd be  tempted just to put in 
``` python 
if inws is not None:
```

NickDraper: Interestingly, I'm having trouble recreating it now.  The first time I didn't have all the workspaces in the right format for a few times before I got it right.

I'd be  tempted just to put in 
``` python 
if inws is not None:
```

mganeva: @NickDraper It is very interesting what you have observed. Theoretically, it should not happen. The `VanadiumWorkspace` property is validated in `PyInit`:

```python
self.declareProperty(MatrixWorkspaceProperty(
                             "VanadiumWorkspace", "",
                             direction=Direction.Input,
                             validator=InstrumentValidator()),
                             "Input Vanadium workspace")
```
If one runs
```python
result = ComputeCalibrationCoefVan(VanadiumWorkspace=None, EPPTable='wsEppTable')
```
one should get the `RuntimeError: Some invalid Properties found` and the `validateInputs()` will not be executed.

Please, try to find the way to reproduce the error you have observed. It is really very interesting. I did not manage to reproduce it neither on Mac nor on Win7.

NickDraper: I've tried again and can't reproduce the crashes, just add the unit into the docs on line 61 and it will be good to go.

soininen:
We discussed about the time series with @mganeva and thought that it might be a good idea to take them into account. I didn't include the changes to this PR, thought.

- The 'temperature' log is to my knowledge only used at MLZ, and it is, to my understanding, always a single value.
- If [`MergeRuns`](http://docs.mantidproject.org/nightly/algorithms/MergeRuns-v1.html) will be used to merge multiple runs at some point, the merged time series log will be called 'temperature_time_series'. Thus the hard-coded log reading won't work.

I opened an issue #18455 so we can work on the time series stuff later.

NickDraper:
All passed and ready to go :shipit:

Framework/PythonInterface/plugins/algorithms/ComputeCalibrationCoefVan.py
```diff
@@ -54,9 +54,17 @@ def PyInit(self):
                              direction=Direction.Output),
                              ("Name the workspace that will contain the " +
                               "calibration coefficients"))
+        self.declareProperty("Temperature",
+                             defaultValue=Property.EMPTY_DBL,
+                             validator=FloatBoundedValidator(lower=0.0),
+                             direction=Direction.Input,
+                             doc=("Temperature during the experiment, in " +
+                                  "Kelvins."))
```
SimonHeybrock: Maybe it should be mentioned here (and in the `rst`) that this overrides the value in the sample log (and that leaving the field empty will use the sample log value)?

soininen: Good point.

SimonHeybrock:
:+1:

PR #18439, opened 2017-01-16
Log averaging muon analysis

### LINKED ISSUE #18061: MuonAnalysis results table: sample log averaging is wrong
#### Expected behaviour
When creating a results table in MuonAnalysis, if a time series log is selected for one column, the values in that column should be the time average over the run. (This average should automatically deal with pauses, aborted/restarted periods, etc). 

This averaging can readily be implemented using the methods already in `TimeSeriesProperty`.

#### Actual behaviour
The average does not use the methods on `TimeSeriesProperty`, but reimplements its own averaging. It will only be correct if the log values are evenly spaced in time, which is often not the case. It does not deal with pauses or aborted/restarted runs.

See [here](https://github.com/mantidproject/mantid/blob/ef0e24d7fa83bf33b566a31b7c2d3fb7eff0faf7/MantidQt/CustomInterfaces/src/Muon/MuonAnalysisResultTableTab.cpp#L494-L510) for how the averaging is done at the moment.

Problem was reported with HIFI runs 108370-108380 and log value `SEPTUM_A_CURR`.

(N.B. Need to wait for #17905 to be merged, as this fixes the methods on `TimeSeriesProperty` to deal correctly with filtering. It might be useful to use `LogFilterGenerator` from #17925.)

#### Reported by
James Lord, Mantid 3.8, Windows

AnthonyLim23:
Within the muon analysis code the time averaging was incorrect. It has been altered to use the timeAverageValue function and if the running log is present then it is filtered accordingly. 

**To test: If the new time averaging is correct**

 
1. Select: Interfaces -> Muon ->Muon Analysis  and a new window should open. 
2. Load some muon data (e.g.  HIFI108376.nxs). 
3. Click the Data Analysis tab. 
4. Under functions  right click and select "add function". In the new box select Background then LinearBackground. 
5. Click the Fit button and select Fit from the options.
6.  Click the Results Table tab. 
7. Select the values that you want to investigate (I have found ICP_SYS_TD to be good). 
8. Click Create Table. 

9. To create the comparison data right click on MuonAnalysis in the workspaces and select Sample Logs...
10. A window should appear. Select the value that is being investigated. 
11. The Time Avg: box should contain the same number as the results table. 

12. It is worth checking that the results are different between Status + Period and None. The results should match Status + Period (which is the same as Status) if the running log is present. 

Fixes #18061.

Release notes: see [here](https://github.com/mantidproject/mantid/pull/18439/commits/816bcc69da686feebd4d7b74c0f24a0d7368bf90)
---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


martyngigg (commented):
The general form of the changes here is good. The variable names are nice and it is readable. 

I think we're possibly over commenting a little bit and when comments are added they should be on the lines preceding the code and not at the end or this leads to logs of whitespace and oddly formatted lines. For example it's not necessary to comment on every line about what is intended. The comments will become out of date and we want readers to focus on the code.

MantidQt/CustomInterfaces/src/Muon/MuonAnalysisResultTableTab.cpp
```diff
@@ -497,40 +497,47 @@ void MuonAnalysisResultTableTab::populateLogsAndValues(
     // Get log information
     std::string wsName = fittedWsList[i].toStdString();
     auto ws = retrieveWSChecked<ExperimentInfo>(wsName + WORKSPACE_POSTFIX);
-
-    Mantid::Kernel::DateAndTime start = ws->run().startTime();
-    Mantid::Kernel::DateAndTime end = ws->run().endTime();
-
     const std::vector<Property *> &logData = ws->run().getLogData();
-
+    const TimeSeriesProperty<bool> *runningLog; // defined here to keep the
```
martyngigg: We should always initialize variables to some known value. This is not guaranteed to be a nullptr so we should make it one:
```
const TimeSeriesProperty<bool> *runningLog(nullptr);
```

AnthonyLim23:
I will edit the comments this afternoon and recommit it.

MantidQt/CustomInterfaces/src/Muon/MuonAnalysisResultTableTab.cpp
```diff
@@ -497,40 +497,47 @@ void MuonAnalysisResultTableTab::populateLogsAndValues(
...
+                                                // object in scope. If
+                                                // applicable it is assigned a
+                                                // value within an if statement.
+
+    std::string running = "running"; // define the running log via a string to
```
martyngigg: I would suggest making this a static string constant at the top of the file to avoid repeated evaluations: Placing

```
const std::string RUNNING_LOG_NAME = "running"
```

in an anonymous namespace at the top of the file will achieve this.

MantidQt/CustomInterfaces/src/Muon/MuonAnalysisResultTableTab.cpp
```diff
@@ -497,40 +497,47 @@ void MuonAnalysisResultTableTab::populateLogsAndValues(
...
+                                     // prevent typos between different
+                                     // occurances.
+    Property *runLog =
+        nullptr; // defined here to keep the object in scope. If applicable
+                 // it is assigned a value within an if statement.
+
+    bool foundRunning = ws->run().hasProperty(
```
martyngigg: We can make this `const`

martyngigg:
@AnthonyLim23 I've added a few other minor comments but the method is sound.

AnthonyLim23:
thank you. I will implement them this afternoon

MantidQt/CustomInterfaces/src/Muon/MuonAnalysisResultTableTab.cpp
AnthonyLim23: Do you mean like this: 

'namespace{
const std::string running = "running";
namespace MantidQt {
....code
}//close namespace 
`

martyngigg: Nearly, except the anonymous namespace should be closed straight after the string is defined:
```
namespace{
const std::string RUNNING_LOG_NAME = "running";
} // end anonymous
```

I have upper cased the variable name to follow the convention of static constants being defined this way.

AnthonyLim23:
updated, that should be all of the changes you suggested. It still works for me.

MantidQt/CustomInterfaces/src/Muon/MuonAnalysisResultTableTab.cpp
```diff
@@ -25,7 +25,9 @@
 #include <QMessageBox>
 
 #include <algorithm>
-
+namespace {
```
martyngigg: Sorry to be picky but can we have a new line between the `#include` & `namespace` and also between the closing `}` and the start of `namespace MantidQt`.

Also, please use uppercase letters for static constants - I would suggest renaming `running` to `RUNNING_LOG_NAME`.

MantidQt/CustomInterfaces/src/Muon/MuonAnalysisResultTableTab.cpp
```diff
@@ -497,40 +499,33 @@ void MuonAnalysisResultTableTab::populateLogsAndValues(
...
+    if (foundRunning) {
+      runLog = ws->run().getLogData(running);
+      runningLog = dynamic_cast<TimeSeriesProperty<bool> *>(runLog);
 
+    } else {
+      Mantid::Kernel::Logger g_log("MuonAnalysisResultTableTab");
+      g_log.warning(
+          "No running log found. Filtering will not be applied to the data.\n");
+    }
     for (const auto prop : logData) {
```
martyngigg: I know this isn't something you have changed but I would suggest adding a `&` qualifier to the `for` loop. It's only copying a pointer but I think it's good practice to avoid copies where possible:

```
for (const auto &prop : logData)
```

AnthonyLim23:
Updated the code

martyngigg:
The code changes look good now.

I've tested the functionality and the results table is now consistent with applying the necessary running/status filters.

Good to go. :shipit:

PR #18441, opened 2017-01-16
Fix angle definition in MayersSampleCorrection

martyngigg:
Fixes the azimuthal angle definition for the multiple scattering calculation in MayersSampleCorrection. Previously it was incorrectly defined as `det.getPhi` when translated from the fortran but the actual definition should be the angle away from the scattering plane. 

While this makes only a small numerical difference to the results it avoids an issue seen on POLARIS where the detectors at `phi=90 deg` (mantid) gave an infinite second path length as there is a `1/cos(phi)` dependence. Phi should be 45 deg above the scattering plane though so as to avoid `cos(90)=0`.

**To test:**

Run `MayersSampleCorrection` on an unfocussed POLARIS (ISIS) dataset and confirm that for all spectra >=55 there are no `inf` values. 

No issue number

[Release notes](/mantidproject/mantid/commit/ccd77803b721fd4423954b64d91192089138419b)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


DavidFair:
This works for me and resolves the issues we were seeing. :shipit:

PR #18460, opened 2017-01-18
Add Beamline::SpectrumDefinition

### LINKED ISSUE #18421: Add mechanism for defining spectra via detector indices and time indices in Beamline::SpectrumInfo.
Eventually this will replace the set of detector IDs in `ISpectrum`. This is also needed for scanning instruments.

SimonHeybrock:
This PR adds `Beamline::SpectrumDefinition`, which is used in the (also new) `Beamline::SpectrumInfo` to store detector groupings. This is part of Instrument-2.0.

`SpectrumDefinition` also stores a time index which currently always defaults to `0` and is ignored by all client code. This will be used to implement scanning support in the near future.

A fair bit of workarounds were required to initialize and update the `SpectrumDefinitions` when things change. The solution is quite hacky and I do not like it, but I think the alternative would be to do a full implementation and rollout of indexing changes (see https://github.com/mantidproject/documents/blob/master/Design/indexing_rollout_notes.md), which I estimate to take 12-24 months.

Detector IDs are now translated into detector indices upon setting them instead of for every access in `SpectrumInfo`. As a consequence we have fewer map lookups and might see some small speedup in various places.

Edit: Due to performance issues from unnecessary updates instead of copying the `Beamline::SpectrumInfo` I changes the mechanism: Setting detector IDs will now just set an `invalid` flag in `ExperimentInfo`. The rebuild happens upon access, triggered by `ExperimentInfo::spectrumInfo`. In the end, I think this is actually less of a hack (but still not perfect).

A follow up PR will cleanup the relationship between `API::SpectrumInfo` and `Beamline::SpectrumInfo`. I decided to keep this separate to keep the (already quite complex) PR smaller.

**To test:**

Thorough code review.

Fixes #18421.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Beamline/inc/MantidBeamline/SpectrumDefinition.h
```diff
@@ -0,0 +1,53 @@
...
+#define MANTID_BEAMLINE_SPECTRUMDEFINITION_H_
+
+#include "MantidBeamline/DllConfig.h"
+
+#include <vector>
+
+namespace Mantid {
+namespace Beamline {
+
+/** SpectrumDefinition : TODO: DESCRIPTION
```
OwenArnold: Description missing

SimonHeybrock: Hah, you are to fast, I pushed it this second.

Framework/Beamline/src/SpectrumDefinition.cpp
```diff
@@ -0,0 +1,30 @@
...
+}
+
+/// Adds a pair of detector index and time index to the spectrum definition. The
+/// time index defaults to zero when not specified.
+void SpectrumDefinition::add(const size_t detectorIndex,
+                             const size_t timeIndex) {
+  auto index = std::make_pair(detectorIndex, timeIndex);
+  auto it = std::lower_bound(m_data.begin(), m_data.end(), index);
+  if ((it == m_data.end()) || (*it != index))
+    m_data.emplace(it, index);
```
OwenArnold: So if we try to add a duplicate it just gets ignored. Is that to make things consistent with some of the existing behaviour?

Framework/Beamline/src/SpectrumDefinition.cpp
```diff
@@ -0,0 +1,30 @@
...
+const std::pair<size_t, size_t> &SpectrumDefinition::
+operator[](const size_t index) const {
+  return m_data[index];
+}
+
+/// Adds a pair of detector index and time index to the spectrum definition. The
+/// time index defaults to zero when not specified.
+void SpectrumDefinition::add(const size_t detectorIndex,
+                             const size_t timeIndex) {
+  auto index = std::make_pair(detectorIndex, timeIndex);
```
OwenArnold: Would it make sense to prevent the m_data having > 1 entries with different time indexes specified. i.e "temporal grouping"

Framework/Beamline/inc/MantidBeamline/SpectrumDefinition.h
```diff
@@ -0,0 +1,67 @@
...
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+class MANTID_BEAMLINE_DLL SpectrumDefinition {
+public:
+  size_t size() const;
+  const std::pair<size_t, size_t> &operator[](const size_t index) const;
+  void add(const size_t detectorIndex, const size_t timeIndex = 0);
+
+private:
+  std::vector<std::pair<size_t, size_t>> m_data;
```
OwenArnold: This is a little different from what I had in mind based on `Instrument 2.0` prototype. We don't want to duplicate time information across `SpectrumInfo` and `DetectorInfo`, but `DetectorInfo` must have time information since it does, for example, need to store a position (or an index to that position) for every detector-time combination. Better we talk about this I think.

Framework/Beamline/src/SpectrumInfo.cpp
```diff
@@ -0,0 +1,32 @@
+#include "MantidBeamline/SpectrumDefinition.h"
+#include "MantidBeamline/SpectrumInfo.h"
+#include "MantidKernel/make_cow.h"
+
+namespace Mantid {
+namespace Beamline {
+
+SpectrumInfo::SpectrumInfo(const size_t numberOfDetectors)
+    : m_spectrumDefinition(Kernel::make_cow<std::vector<SpectrumDefinition>>(
```
OwenArnold: So the vector of spectrum definitions initially is sized correctly but each `SpectrumDefinition` is empty.

Framework/Beamline/src/SpectrumDefinition.cpp
SimonHeybrock: `ISpectrum` stores detector IDs in a `set`, and I think that makes sense? I am basically using a `vector` to mimic a set.

Framework/Beamline/src/SpectrumInfo.cpp
SimonHeybrock: Yes. A default mapping will be created implicitly by `Workspace2d::init` and `EventWorkspace::init`.

Framework/Beamline/src/SpectrumDefinition.cpp
OwenArnold: Ok. I can't think of any specific scenarios where this would not be OK. I guess the client can always check the size if failure to add is some kind of error conditions.

Framework/Beamline/src/SpectrumDefinition.cpp
SimonHeybrock: I am copying an example for clarification that we discussed on Slack (short answer: we do not want to prevent different time indices):

Say we want a spectrum that covers a small region along the x axis, marked by `xxx`:

Static case:
```
time   x --->
0      0 1 2 3 4
           xxx
```
Detectors 2 and 3 are part of the spectrum (time index not needed, access `DetectorInfo.l2(2)` and `DetectorInfo.l2(3)`.


Scanning case:

```
time   x --->
0      0 1 2 3 4
1       0 1 2 3 4
2        0 1 2 3 4
           xxx
```
All in all, five detector positions (of 3 different detectors) contribute. Access, e.g., `DetectorInfo.l2(2,0)` or `DetectorInfo(1,2)`.

So in the former case, `SpectrumDefinition` contains `{{2,0},{3,0}}`. In the latter case it contains `{{2,0},{3,0}},{2,1},{1,2},{2,2}}`.

SimonHeybrock:
The index conversion still has some performance troubles, trying to find ways to fix those.

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -1018,39 +1085,54 @@ DetectorInfo &ExperimentInfo::mutableDetectorInfo() {
...
+  // count to the ParameterMap is typically not 1, so
+  // invalidateInstrumentReferences() is called.
+  // 2. invalidateInstrumentReferences() resets m_spectrumInfoWrapper, releasing
+  // any parameterized detectors and thus dropping any unneeded references to
+  // the ParameterMap.
+  // 3. Construction of SpectrumInfo continues and the result is assigned to
+  // m_spectrumInfoWrapper.
+
+  // No locking here since this non-const method is not thread safe.
+  m_spectrumInfoWrapper = Kernel::make_unique<SpectrumInfo>(*this);
```
OwenArnold: This whole mechanism looks very complicated. If we have time during the scanning meetup, can we look through this together?

SimonHeybrock: Yes we should. It is horrible and I'd remove this rather sooner than later but have not quite found a way to do it.

SimonHeybrock: @OwenArnold Note that this was just moved over from `MatrixWorkspace`, it is not an actual addition in the PR.

OwenArnold: @SimonHeybrock. I can't immediately see a better way of doing this. I've added it as an item to the agenda.

SimonHeybrock: I think once we store positions in `Beamline::DetectorInfo` we would be able to remove the caching of detectors in `API::SpectrumInfo` and `API::DetectorInfo`. That would require changing `API::SpectrumInfo::detector` to return a `shared_ptr` instead of a reference (and some refactoring). Once that is done, the only remaining hurdle is the caching of source and sample positions. If we find a way to solve also that, we can remove the weird reinitialization mechanism.

Framework/API/inc/MantidAPI/ExperimentInfo.h
```diff
@@ -74,11 +76,13 @@ class MANTID_API_DLL ExperimentInfo {
...
-  const std::set<detid_t> &getGroupMembers(const detid_t detID) const;
-  /// Get a detector or detector group from an ID
-  Geometry::IDetector_const_sptr getDetectorByID(const detid_t detID) const;
+  virtual void cacheDetectorGroupings(const det2group_map &mapping);
+
+  void setNumberOfDetectorGroups(const size_t count);
+  void updateCachedDetectorGrouping(const size_t index,
+                                    const std::set<detid_t> &detIDs);
+  void addDetectorToGroup(const size_t index, const detid_t detID);
+  void addDetectorsToGroup(const size_t index, const std::set<detid_t> &detIDs);
```
OwenArnold: This is slightly off on a tangent, but have you considered what the addition of these methods might mean to the `MultipleExperimentInfos` interface as used by the `MDWorkspace` types? I would expect that the `ExperimentInfos` as seen by `MDWorkspaces` are essentially immutable. I'm not sure that its worth worrrying about that though at this stage.

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -63,6 +66,9 @@ ExperimentInfo::ExperimentInfo()
  */
 ExperimentInfo::ExperimentInfo(const ExperimentInfo &source) {
   this->copyExperimentInfoFrom(&source);
+  if (source.m_spectrumInfo)
+    m_spectrumInfo =
+        Kernel::make_unique<Beamline::SpectrumInfo>(*source.m_spectrumInfo);
```
OwenArnold: 👍

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -414,46 +434,93 @@ void ExperimentInfo::swapInstrumentParameters(Geometry::ParameterMap &pmap) {
...
+void ExperimentInfo::setNumberOfDetectorGroups(const size_t count) {
   populateIfNotLoaded();
-  auto iter = m_det2group.find(detID);
-  if (iter != m_det2group.end()) {
-    return m_detgroups[iter->second];
-  } else {
-    throw std::runtime_error(
-        "ExperimentInfo::getGroupMembers - Unable to find ID " +
-        std::to_string(detID) + " in lookup");
+  m_spectrumInfo = Kernel::make_unique<Beamline::SpectrumInfo>(count);
```
OwenArnold: From what I can tell, at the point this is called you always need a new `m_spectrumInfo` so it would be wasteful to do a `count == m_spectrumInfo.size()` type check prior to the assignment?

Framework/API/inc/MantidAPI/ExperimentInfo.h
SimonHeybrock: You commented on an outdated version, only the first of these methods is still around.

I think in principle this method could also be used for `ExperimentInfo` when it is owned by a `MultipleExperimentInfo`, but that is not the case in reality.

Framework/API/src/ExperimentInfo.cpp
SimonHeybrock: Then we would also need something like `Beamline::SpectrumInfo::clearData`. I don't think we would gain much?

Framework/API/src/ISpectrum.cpp
```diff
@@ -173,16 +173,18 @@ ISpectrum &ISpectrum::operator=(ISpectrum &&other) {
...
+                                  const size_t index) {
   m_experimentInfo = experimentInfo;
+  m_index = index;
 }
 
 /// Updates detector IDs in the owning ExperimentInfo.
 void ISpectrum::updateExperimentInfo() const {
   if (m_experimentInfo)
-    m_experimentInfo->invalidateSpectrumDefinitions();
+    m_experimentInfo->invalidateSpectrumDefinition(m_index);
```
OwenArnold: 👍

Framework/API/src/SpectrumInfo.cpp
```diff
@@ -212,8 +210,8 @@ SpectrumInfo::getDetectorVector(const size_t index) const {
 }
 
 std::vector<size_t> SpectrumInfo::getDetectorIndices(const size_t index) const {
-  const auto &specDef =
-      m_experimentInfo.internalSpectrumInfo().spectrumDefinition(index);
+  m_experimentInfo.updateSpectrumDefinitionIfNecessary(index);
```
OwenArnold: I found it a little odd to see these calls from a const getter. Presumably this is part of the lazy initialisation?

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -219,7 +217,8 @@ void ExperimentInfo::setInstrument(const Instrument_const_sptr &instr) {
   // instrument may now suddenly be valid, so we have to reinitialize the
   // detector grouping. Also the index corresponding to specific IDs may have
   // changed.
-  invalidateSpectrumDefinitions();
+  std::fill(m_spectrumDefinitionNeedsUpdate.begin(),
```
OwenArnold: I'm not a big fan of how `ExperimentInfo` gets created and then setup like this, but I guess that's all part of a much larger issue that would be hard to change at this point.

Framework/API/src/SpectrumInfo.cpp
SimonHeybrock: Yes, lazy initialization but also the fact that detector IDs can be added at any time, so even if it was initialized once we may need to update at any point.

Framework/API/src/ExperimentInfo.cpp
SimonHeybrock: Yes, it is basically a problem of `MatrixWorkspace` which is initialized in many steps (some of which are optional):

- (copy) constructor
- `MatrixWorkspace::initialize`
- `ExperimentInfo::setInstrument` or `WorkspaceFactory::initializeFromParent`
- setting detector IDs for spectra

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -1962,18 +1966,14 @@ size_t MatrixWorkspace::groupOfDetectorID(const detid_t) const {
...
-  setNumberOfDetectorGroups(getNumberHistograms());
-  for (size_t i = 0; i < getNumberHistograms(); ++i) {
-    updateCachedDetectorGrouping(i, getSpectrum(i).getDetectorIDs());
-  }
+ * This method is called when the detector grouping stored in SpectrumDefinition
+ * at `index` in Beamline::SpectrumInfo is not initialized or outdated. Detector
+ * IDs are currently stored in ISpectrum, but grouping information needs to be
+ * available and updated in Beamline::SpectrumInfo. */
+void MatrixWorkspace::updateCachedDetectorGrouping(const size_t index) const {
+  setDetectorGrouping(index, getSpectrum(index).getDetectorIDs());
```
OwenArnold: This seems like quite a long path to set the detector groupings. I guess the only alternative would be to make getSpectrum part of `ExperimentInfo` which is most likely not a good option.

OwenArnold:
Overall I think the last set of changes have made things neater. Seems that there is now a reasonable strategy for syncs involving `ISpectrum` and `SpectrumDefinition`.

Framework/API/src/ExperimentInfo.cpp
OwenArnold: Yes, not a problem to solve as part of this.

Framework/API/src/MatrixWorkspace.cpp
SimonHeybrock: The problem is basically that there are two kinds of detector grouping convoluted into this: Grouping in `MatrixWorkspace` and the detector grouping caching mechanism used by some `MDAlgorithms`. So we need some kind of virtual function call and go via `MatrixWorkspace`.

Framework/Beamline/inc/MantidBeamline/SpectrumDefinition.h
ianbush: I think this is consistent with what we had in the [design document](https://github.com/mantidproject/documents/blob/17832_Design_document_for_scanning_instruments_at_ILL/Design/Instrument-2.0/ScanningInstruments_Design.md#spectruminfo). I can not remember what we discussed about using different types for each index - but is this a good time to introduce separate indexes for the detector index and time index?

Framework/Beamline/src/SpectrumDefinition.cpp
ianbush: I do not think this is a use case at the ILL, but I agree that it makes sense to allow it.

Framework/Beamline/inc/MantidBeamline/SpectrumDefinition.h
SimonHeybrock: I think the time indices will mainly be used internally in `SpectrumInfo` and `DetectorInfo`. If we notice that they see a lot of use outside, in client code, it would make sense to consider types for them.

Framework/Beamline/test/SpectrumInfoTest.h
```diff
@@ -0,0 +1,95 @@
...
+  void test_setSpectrumDefinition() {
+    SpectrumDefinition def;
+    def.add(7, 5);
+    TS_ASSERT_EQUALS(def.size(), 1);
+    SpectrumInfo info(3);
+    info.setSpectrumDefinition(1, def);
+    TS_ASSERT_EQUALS(info.spectrumDefinition(1)[0],
+                     (std::pair<size_t, size_t>(7, 5)));
+    TS_ASSERT_EQUALS(def.size(), 1);
+  }
```
ianbush: Should there be an explicit test for the copy on write behaviour - i.e. copy info, add a new `SpectrumDefinition`, verify they are different?

SimonHeybrock: Yes, good point.

Framework/Beamline/inc/MantidBeamline/SpectrumDefinition.h
OwenArnold: Yes, particularly for SpectrumInfo. Apart from some specialist cases, to algorithms any moved detector should just be an additional spectrum, so I would be against making the time index a requirement for access at the `SpectrumInfo` level.

ianbush:
It will be good to discuss this further next week, but overall this looks good to me.

SimonHeybrock:
I pushed an update that I consider as a cleanup of the mechanism for updating and transferring `SpectrumDefinitions` between workspaces. It also serves as a preparation for more work on `Indexing`, in particular compatibility with a future MPI implementation.

SimonHeybrock:
With `IndexInfo::setSpectrumDefinitions` it also provides a way to set up a scanning workspace, I think.

Framework/Beamline/inc/MantidBeamline/SpectrumDefinition.h
OwenArnold: A relatively minor point, but can we give this a better name than `m_data`. `m_detectorTimeIndexes`?

Framework/Beamline/inc/MantidBeamline/SpectrumInfo.h
```diff
@@ -0,0 +1,80 @@
...
+
+  size_t size() const;
+
+  const SpectrumDefinition &spectrumDefinition(const size_t index) const;
+  void setSpectrumDefinition(const size_t index, SpectrumDefinition def);
+  const Kernel::cow_ptr<std::vector<SpectrumDefinition>> &
+  sharedSpectrumDefinitions() const;
+
+private:
+  Kernel::cow_ptr<std::vector<SpectrumDefinition>> m_spectrumDefinition;
```
OwenArnold: 👍 - Did you decide against using/typedefs for part of this type definition.

Framework/API/inc/MantidAPI/ExperimentInfo.h
```diff
@@ -214,6 +223,13 @@ class MANTID_API_DLL ExperimentInfo {
   boost::shared_ptr<Beamline::DetectorInfo> m_detectorInfo;
   mutable std::unique_ptr<DetectorInfo> m_detectorInfoWrapper;
   mutable std::mutex m_detectorInfoMutex;
+
+  mutable std::unique_ptr<Beamline::SpectrumInfo> m_spectrumInfo;
+  mutable std::unique_ptr<SpectrumInfo> m_spectrumInfoWrapper;
+  mutable std::mutex m_spectrumInfoMutex;
+  // This vector stores boolean flags but uses char to do so since
+  // std::vector<bool> is not thread-safe.
+  mutable std::vector<char> m_spectrumDefinitionNeedsUpdate;
```
OwenArnold: 👍

Framework/Beamline/inc/MantidBeamline/SpectrumInfo.h
SimonHeybrock: What typedefs are you referring to?

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -204,6 +213,12 @@ void ExperimentInfo::setInstrument(const Instrument_const_sptr &instr) {
     m_parmap = boost::make_shared<ParameterMap>();
   }
   m_detectorInfo = makeDetectorInfo(*sptr_instrument, *instr);
+  // Detector IDs that were previously dropped because they were not part of the
+  // instrument may now suddenly be valid, so we have to reinitialize the
+  // detector grouping. Also the index corresponding to specific IDs may have
+  // changed.
+  std::fill(m_spectrumDefinitionNeedsUpdate.begin(),
```
OwenArnold: I think your storage type for the `m_spectrumDefinitionNeedsUpdate` is good. However i think it might make things slightly more readable of updates to that vector were made via some better named methods. So in this case something like `invalidateAllSpectrumDefinitions()`.

Framework/Indexing/src/IndexInfo.cpp
```diff
@@ -116,6 +119,19 @@ void IndexInfo::setDetectorIDs(
...
+    std::vector<Beamline::SpectrumDefinition>> spectrumDefinitions) {
+  m_spectrumDefinitions = spectrumDefinitions;
+}
+
+const Kernel::cow_ptr<std::vector<Beamline::SpectrumDefinition>> &
+IndexInfo::spectrumDefinitions() const {
+  return m_spectrumDefinitions;
 }
 
 } // namespace Indexing
```
OwenArnold: missing tests for this new functionality

Framework/Indexing/inc/MantidIndexing/IndexInfo.h
```diff
@@ -91,6 +100,9 @@ class MANTID_INDEXING_DLL IndexInfo {
 
   Kernel::cow_ptr<std::vector<specnum_t>> m_spectrumNumbers;
   Kernel::cow_ptr<std::vector<std::vector<detid_t>>> m_detectorIDs;
+
+  Kernel::cow_ptr<std::vector<Beamline::SpectrumDefinition>>
```
OwenArnold: Is COW the right thing here? We can maybe talk about this?

SimonHeybrock: Typically many workspaces have the same spectra (usually all workspaces for the same instrument), so we want to share as much as possible. What were your thoughts exactly?

SimonHeybrock:
@OwenArnold I pushed a small update based on some of your comments. I cannot test `IndexInfo::setSpectrumInfo` within the `Indexing` module, since right now `SpectrumDefinition` is part of `Beamline`, and I think there should be no dependency between the two (or maybe only in the other direction).  We can consider moving `SpectrumDefinition` to another module. Maybe a lower level module with some of the the very basic types would make sense?

SimonHeybrock:
Another thing you may notice in this context is that there is currently no check to enforce that the `SpectrumDefinitions` in `IndexInfo` are compatible with the groups of detector IDs held in the class. This is not so bad currently since `IndexInfo::setSpectrumDefinitions` is only used internally be `MatrixWorkspace, but ultimately we have to find a better solution (ideally by removing detector IDs from this context, but that would be long in the future). I am currently working on a major revision/update of things in the `Indexing` module (that will allow us to support MPI), maybe we can try and tackle this in that context.

Framework/Indexing/inc/MantidIndexing/IndexInfo.h
OwenArnold: My concern is in allowing that sharing to be broken. I think the sharing is essential, I'm not clear on the circumstances where we would want the breaking of that sharing to happen.

SimonHeybrock: As soon as a `SpectrumDefinition` is changed sharing will be and must be broken. Currently this will mainly happen when using, e.g., `ISpectrum::setDetectorID()`.

OwenArnold:
@SimonHeybrock Ah yes. Regarding your last comment I did notice that. Would it be so expensive to perform the compatibility check. I think we should have it and not depend on the current client usage keeping things together. This may cause nasty and hard to track down future bugs. For the timebeing, could we have `setSpectrumDefinitions` throw if the spectrum definition was in any way incompatible. I guess you could also add a unit test for this.

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -414,44 +428,63 @@ void ExperimentInfo::swapInstrumentParameters(Geometry::ParameterMap &pmap) {
...
-  auto iter = m_det2group.find(detID);
-  if (iter != m_det2group.end()) {
-    return m_detgroups[iter->second];
-  } else {
-    throw std::runtime_error(
-        "ExperimentInfo::getGroupMembers - Unable to find ID " +
-        std::to_string(detID) + " in lookup");
-  }
+  if (m_spectrumInfo)
+    m_spectrumDefinitionNeedsUpdate.clear();
```
OwenArnold: Maybe a similar thing to the `invalidateSpectrumDefinition` required here. Not necessary, but you might consider going further the `std::vector<char>` could be made into a separate class with these methods attached.

OwenArnold:
When builds pass :shipit:

martyngigg:
`APITest.SpectrumInfoTest.test_no_detector_twice` seems to be failing everywhere.

SimonHeybrock:
I will have a look, it is probably a badly resolved merged.

Framework/Types/inc/MantidTypes/SpectrumDefinition.h
```diff
@@ -0,0 +1,79 @@
...
+  /// The time index defaults to zero when not specified.
+  void add(const size_t detectorIndex, const size_t timeIndex = 0) {
+    auto index = std::make_pair(detectorIndex, timeIndex);
+    auto it = std::lower_bound(m_data.begin(), m_data.end(), index);
+    if ((it == m_data.end()) || (*it != index))
+      m_data.emplace(it, index);
+  }
+
+private:
+  std::vector<std::pair<size_t, size_t>> m_data;
```
martyngigg: Is it worth defining a POD `struct` with named members such as `detector_index` & `time_index` instead of having to remember what `first/second` refer to?

SimonHeybrock: @martyngigg I gave it a go, but then you have to implement all operators for comparisions, etc., and test those,...

Should I?

Typically this pair would be used directly, without obtaining the members, e.g.:

```cpp
Eigen::Vector3d SpectrumInfo::position(size_t i) {
  Eigen::Vector3d pos(0,0,0);
  for ( const auto &index: spectrumDefinition(i)) {
     pos += m_detectorInfo.position(index); // index is std::pair<det,time>
  }
  return pos/spectrumDefinition(i).size();
}
```

martyngigg: Good point, that is a lot more code to deal with.

Most people (developers) would then deal with the higher-level `SpectrumInfo::position` API rather than the lower-level `SpectrumDefinition` directly? If that's the case then I agree that what I suggested is probably too much overhead to maintain.

SimonHeybrock: @martyngigg Yes. Right now there are 185 files using `SpectrumInfo`, 68 using `DetectorInfo`.

Of those using `DetectorInfo`, many cases should be able to use something similar to by code example above, i.e., just get the compound index and forward it to `DetectorInfo`. 

There will probably be a few scanning-instrument-specific algorithms that will need to deal with the time indices in a more direct manner.

martyngigg: :+1:

martyngigg:
Looks good to me too. :+1:

Issue #18443, opened 2017-01-16
Add method to TimeSeriesProperty to get average value


DavidFair:
Currently the TimeSeriesProperty which is a templated class requires the caller to deduce the property type before getting the average value. This results in code that looks like the following:

``` c++
    if (type.std::string::find("vector") &&
        type.std::string::find("double") != std::string::npos) {
      const auto tsp = m_inputWS->run().getTimeSeriesProperty<double>(name);
      value = std::to_string(tsp->timeAverageValue());
    }

    else if (type.std::string::find("vector") &&
             type.std::string::find("int") != std::string::npos) {...}
    ....etc....
```

If we add a `timeAverageValue` method to the interface we get then get the property pointer, dynamically cast it to an ITimeSeriesProperty and call the method letting runtime polymorphism handle the rest. 

Following this `SaveOpenGenieAscii` can be refactored to utilise this method to remove trying to deduce the type from its `std::type_info`

DavidFair:
Added a method in the header in #18468

PR #18468, opened 2017-01-18
Rewrite SaveOpenGenieAscii to save valid files

### LINKED ISSUE #18084: SaveOpenGenieAscii saving all log values incorrectly
### Expected behavior
SaveOpenGenieAscii should save out an average of each log value in a way that Open Genie can load them.

### Actual behavior
The algorithm saves all log data with a date time stamp first and the value second. This means that it cannot be parsed by Open Genie trying to load it

Initially we need to get an interim fix as only the table positions in x,y,z, Θ and the stress/strain values. Later on we can look at which other log data is worth bringing across and the correct format to use.

DavidFair:
Description of work.
This PR rewrites `SaveOpenGenieAscii` to both clean it up and fix the format for ENGIN-X OpenGenie. At a later date a rarely used procedure could need a new parameter in the output file but at the moment this PR covers those needed in their `analyze_scan` routine which they need for some data processing.

Additionally it adds a unit test which just checks the output against a reference file. 

**To test:**
As a copy of ENGIN-X routines is difficult to setup and test these files with you will have to take my word the file (currently) works correctly in OpenGenie. 
- Code Review
- Run the unit test

Fixes #18084 and #18443 .

[Release notes](https://github.com/mantidproject/mantid/pull/18468/files#diff-2abf582e699070b9e379ddfafa4d1094)

~Additionally please do **not** merge this branch yet. There is a issue open with Github about the fact the author appears twice in compare commits using Git 2.11 that links to this branch. If the branch is merged the commits will not longer appear in the comparison.~

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


DavidFair:
Additionally to the person merging the branch please do *not* delete the branch. There is a issue open with Github about the fact the author appears twice in commits using Git 2.11 that links to this branch.

Framework/DataHandling/inc/MantidDataHandling/SaveOpenGenieAscii.h
```diff
@@ -33,44 +33,65 @@ class DLLExport SaveOpenGenieAscii : public Mantid::API::Algorithm {
   }
 
 private:
+  /// Typedef of a tuple containing the name, type and value as strings
+  using outputTuple = std::tuple<std::string, std::string, std::string>;
```
samueljackson92: Type names should being with a capital. It would also be better if this was more descriptive, what is an "Output Tuple"? See: http://llvm.org/docs/CodingStandards.html

Framework/DataHandling/inc/MantidDataHandling/SaveOpenGenieAscii.h
```diff
@@ -33,44 +33,65 @@ class DLLExport SaveOpenGenieAscii : public Mantid::API::Algorithm {
...
   /// Execution code
   void exec() override;
 
-  // write file header
-  void writeFileHeader(std::ofstream &outfile);
+  inline void addToOutputBuffer(const std::string outName,
+                                const std::string outType,
+                                const std::string outVal) {
+    m_outputVector.push_back(outputTuple(outName, outType, outVal));
+  }
```
samueljackson92: We should pass by ref, i.e. `const std::string &`

Framework/DataHandling/src/SaveOpenGenieAscii.cpp
```diff
@@ -49,337 +46,382 @@ void SaveOpenGenieAscii::init() {
   std::vector<std::string> header{"ENGIN-X Format", "Basic Format"};
   declareProperty("OpenGenieFormat", "ENGIN-X Format",
                   boost::make_shared<Kernel::StringListValidator>(header),
-                  "The format required to succesfully load the file to "
+                  "The format required to successfully load the file to "
                   "OpenGnie: ENGIN-X Format (default)");
```
samueljackson92: Typo "OpenGnie"

Framework/DataHandling/src/SaveOpenGenieAscii.cpp
```diff
@@ -49,337 +46,382 @@ void SaveOpenGenieAscii::init() {
...
-  // if true write file header
-  if (headers) {
-    writeFileHeader(outfile);
+  const std::string formatType = getProperty("OpenGenieFormat");
+  if (formatType == "ENGIN-X Format") {
+    progressBar.report("Generating ENGINX header");
+    applyEnginxFormat();
+  } else {
+    const std::string err_msg("Unrecognized format \"" + formatType + "\"");
+    throw std::runtime_error(err_msg);
```
samueljackson92: Should we really throw here if we're not using the ENGIN-X format? If the format is "Basic Format" do we just do nothing now? In which case does this break backwards compatibility? Do we need a version 2 of this alg?

Framework/DataHandling/src/SaveOpenGenieAscii.cpp
```diff
@@ -49,337 +46,382 @@ void SaveOpenGenieAscii::init() {
...
-      auto tsp = ws->run().getTimeSeriesProperty<bool>(name);
-      value = boost::lexical_cast<std::string>(tsp->timeAverageValue());
+  const auto &e = m_inputWS->e(0);
+  for (const auto eVal : e) {
+    if (eCount % 10 == 0) {
+      eValsOutput += newlineStr;
     }
+    eCount++;
+    eValsOutput += std::to_string(eVal) + ' ';
+  }
```
samueljackson92: These three for loops are identical. Consider a method and pass in x, y, or e?

Framework/DataHandling/src/SaveOpenGenieAscii.cpp
```diff
@@ -49,337 +46,382 @@ void SaveOpenGenieAscii::init() {
...
+  outDataStrings.push_back(std::make_tuple(eValsOutput, eCount));
+  return outDataStrings;
+}
 
-      auto tsp = ws->run().getTimeSeriesProperty<std::string>(name);
-      value = (tsp->lastValue());
-    }
+int SaveOpenGenieAscii::determineEnginXBankId() {
+  const auto &detectorIds = m_inputWS->getSpectrum(0).getDetectorIDs();
+  const std::string firstDetectorId = std::to_string(*detectorIds.cbegin());
```
samueljackson92: No docstring?

Framework/DataHandling/src/SaveOpenGenieAscii.cpp
```diff
@@ -49,337 +46,382 @@ void SaveOpenGenieAscii::init() {
...
+  * Mantid to the expected log name in OpenGenie. If any names are
+  * unrecognised they are skipped. If the logs have multiple values
+  * the time weighted average is taken instead. These strings are
+  * then stored in the output buffer.
+  */
+void SaveOpenGenieAscii::getSampleLogs() {
+  // Maps Mantid log names -> Genie save file name / type
+  const std::unordered_map<std::string, std::tuple<std::string, std::string>>
+      MantidGenieNameMap = {
+          {"x", std::make_tuple("x_pos", m_floatType)},
```
samueljackson92: Using upper case for variable name. LLVM code guide actually states that this is what we should be doing, but this is inconsistent with the rest of Mantid.

samueljackson92 (changes requested):
See comments below.

Framework/DataHandling/src/SaveOpenGenieAscii.cpp
DavidFair: I think the smartest option would be to remove the "Basic Format" option. This is because the algorithm is written for ENGIN-X OpenGenie which is the normal package plus several bespoke routines.
I suspect the original writer(s) foresaw the possibility of needing to support the vanilla version of the package and tried to keep the ENGIN-X specific stuff tided.

Framework/DataHandling/src/SaveOpenGenieAscii.cpp
DavidFair: I'm not too sure how to go about this - originally I wrote a small method which took the data as a parameter and was just called like so:
```c++
const auto &x = m_inputWS-x(0)
processWorkspaceData(x)
```
However I can't seem to see a universal histogram type for the signature - I have to either specify `HistogramX/Y/E`?

Framework/DataHandling/src/SaveOpenGenieAscii.cpp
DavidFair: Actually thinking about this as it is a constant mapping would it be worth naming it something along the lines of: `MANTID_GENIE_LOG_MAPPING`

Framework/DataHandling/src/SaveOpenGenieAscii.cpp
samueljackson92: So this is now just ENGIN-X specific? How certain are you no-one out there uses the basic format? Is the basic version of this just to not call `applyEnginxFormat`? If so shouldn't we not throw an exception and keep going?

Framework/DataHandling/src/SaveOpenGenieAscii.cpp
samueljackson92: Personally I'd hold off and just use screaming caps for Marcos. The type is already const which is more important.

Framework/DataHandling/src/SaveOpenGenieAscii.cpp
DavidFair: No previously if you tried to save in the basic format it would throw an exception [here](https://github.com/mantidproject/mantid/blob/master/Framework/DataHandling/src/SaveOpenGenieAscii.cpp#L115) hence I know for sure nobody is using the basic format.

The algorithm has never actually worked correctly since its initial release as it was missing some essential fields. However they don't use it very often as it is used when they are processing event workspaces as their bespoke routines handle everything with matrix workspaces. 
As they are gradually doing more with event workspaces and they have data waiting to be processed it became a priority when previously it wasn't an issue for them.

Framework/DataHandling/src/SaveOpenGenieAscii.cpp
samueljackson92: I see . Could use templates? Could even do something clever with `std::enable_if`?

Framework/DataHandling/src/SaveOpenGenieAscii.cpp
samueljackson92: If it already threw an exception for that case then I am satisfied. 👍

Framework/DataHandling/src/SaveOpenGenieAscii.cpp
DavidFair: I'll look into it thanks for the idea.

samueljackson92: Could almost do it with something like `std::accumulate` and a lambda function, but C++11 doesn't support `auto` in lambda definitions (C++14 does though...).

DavidFair: In the end I went with a plain old template as the `auto` iterator exists for the three types meaning we don't have to handle trying to deduce operators.

Framework/DataHandling/src/SaveOpenGenieAscii.cpp
```diff
@@ -46,340 +43,371 @@ void SaveOpenGenieAscii::init() {
...
+    outputString += std::to_string(val) + ' ';
+  }
 
-    else if (type.std::string::find("vector") &&
-             type.std::string::find("bool") != std::string::npos) {
-
-      auto tsp = ws->run().getTimeSeriesProperty<bool>(name);
-      value = boost::lexical_cast<std::string>(tsp->timeAverageValue());
-    }
+  return std::make_tuple(outputString, valueCount);
```
samueljackson92: Why not `std::pair`?

DavidFair: Ahh originally it was returning three values (the type as well) so a tuple made sense at the time. I'll go and change this as I know a few places that can be turned into pairs

samueljackson92:
Jenkins hung on the RHEL7 build. Once Jenkins is up again and that build passes then we're good to :shipit:

samueljackson92 (approved):

PR #18521, opened 2017-01-24
Crystal fields bugfixes

### LINKED ISSUE #18490: Crystal Fields bugfixes
This is a general issue for quick bugfixes to the Crystal Field calculations in Mantid, in preparation for release 3.9.

Identified issues from user testing are: (more issues will be added as they are identified).

1. No way to set `Lambda` parameter for `CrystalFieldSusceptibility` in Python.

2. Cannot remove ties previously set using the Python interface. [see also #18518 ]

3. Change Doctests to be more robust.

mducle:
Fixes some bugs found in testing of the new crystal field functionalities. In particular it fixes a bug in the doctests which uses a fitting example, but gave too exact output of the fitted variables.

**To test:**

Fixes #18490 .

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/docs/source/fitfunctions/CrystalFieldHeatCapacity.rst
+++ b/docs/source/fitfunctions/CrystalFieldHeatCapacity.rst
@@ -54,7 +54,7 @@ The `x`-axis is given in Kelvin, and the heat capacity (`y`-axis) is in Joules p
     EvaluateFunction(fun, ws, OutputWorkspace='data')
     
      # Change parameters slightly and fit to the reference data
-    fun = 'name=CrystalFieldHeatCapacity,Ion=Ce,Symmetry=C2v,B20=0.4,B22=0.04,B40=-0.03,B42=-0.1,B44=-0.1,'
+    fun = 'name=CrystalFieldHeatCapacity,Ion=Ce,Symmetry=C2v,B20=0.377,B22=0.0397,B40=-0.0318,B42=-0.116,B44=-0.125,'
     fun += 'ties=(B60=0,B62=0,B64=0,B66=0,BmolX=0,BmolY=0,BmolZ=0,BextX=0,BextY=0,BextZ=0)'
     
     # (set MaxIterations=0 to see the starting point)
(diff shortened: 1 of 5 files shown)


martyngigg:
@mducle Unfortunately 3 of the other tests have failed: http://builds.mantidproject.org/job/pull_requests-ubuntu/11869/

martyngigg:
@mducle I wonder if it might be better to not expect exact answers here and say something more like we did with `EstimateFitParameters` and accept anything and tell users that their output should look similar but not exactly the same: https://raw.githubusercontent.com/mantidproject/mantid/master/docs/source/algorithms/EstimateFitParameters-v1.rst

The doc tests are not there to verify numbers really more that the workflow still works.

mducle:
@martyngigg Yes. I think you're right. I'll make the changes.

martyngigg:
@mducle The last run of the doctests seems to give a different sign for one of the values: http://builds.mantidproject.org/job/pull_requests-ubuntu/11886/testReport/docs/fitfunctions_CrystalFieldHeatCapacity/ExampleCrystalFieldHeatCapacity/

mducle:
@martyngigg Yes. I was hopping that it would get the sign the same... The problem is that the heat capacity only depends on the energy levels, but in this case only two energy levels really contribute to the heat capacity below 300K, whilst there are three energy levels in total. So the position of the third energy level can vary by anything, which means the crystal field parameters which affects that energy level but not the one contributing to the heat capacity can vary wildly. I've just changed it so there's no more output to compare to. So as long as the code runs without error it will pass the test.

martyngigg:
@mducle Unfortunately now the tests fail because there is no output!

I think we'll either have to specify that absolutely any output is acceptable (using the ellipsis) or remove the print statements from the example.

mducle:
@martyngigg Doh! Sorry. I'll use the ellipsis.

martyngigg:
LGTM to fix the failing PR builds.

:shipit:

PR #18528, opened 2017-01-24
SpectrumInfo/DetectorInfo rollout

### LINKED ISSUE #18525: SpectrumInfo and DetectorInfo Rollout `AddPeak` to `DetectorDiagnostic`
SpectrumInfo and DetectorInfo rollout: algorithms <AddPeak to ScaleX>

```
Framework/Algorithms/src/AddPeak.cpp
Framework/Algorithms/src/ConvertDiffCal.cpp
Framework/Algorithms/src/ConvertSpectrumAxis.cpp
Framework/Algorithms/src/ConvertSpectrumAxis2.cpp
Framework/Algorithms/src/ConvertUnitsUsingDetectorTable.cpp
Framework/Algorithms/src/CreateLogTimeCorrection.cpp
Framework/Algorithms/src/DetectorDiagnostic.cpp
```

OwenArnold:
SpectrumInfo rollout: algorithms

## Description of work.
This PR is part of refactoring effort to use the new interface of the SpectrumInfo and DetectorInfo modules.

The following algorithms were refactored to use SpectrumInfo/DetectorInfo:
```
Framework/Algorithms/src/AddPeak.cpp
Framework/Algorithms/src/ConvertDiffCal.cpp
Framework/Algorithms/src/ConvertSpectrumAxis.cpp
Framework/Algorithms/src/ConvertSpectrumAxis2.cpp
Framework/Algorithms/src/ConvertUnitsUsingDetectorTable.cpp
Framework/Algorithms/src/CreateLogTimeCorrection.cpp
Framework/Algorithms/src/DetectorDiagnostic.cpp
```

Performance tests were added for:

The following algorithms have improved performance:

## To test.

Code review.

Fixes #18525


Framework/Algorithms/src/ConvertDiffCal.cpp
```diff
@@ -102,18 +105,18 @@ double calculateDIFC(OffsetsWorkspace_const_sptr offsetsWS,
...
   const detid_t detid = getDetID(offsetsWS, index);
   const double offset = getOffset(offsetsWS, detid);
-
-  double l1;
-  Kernel::V3D beamline, samplePos;
-  double beamline_norm;
-  instrument->getInstrumentParameters(l1, beamline, beamline_norm, samplePos);
-
-  Geometry::IDetector_const_sptr detector = instrument->getDetector(detid);
+  const Mantid::API::SpectrumInfo &spectrumInfo = offsetsWS->spectrumInfo();
```
ianbush: This seems to be done inside a loop. I think `spectrumInfo` can be passed in to this method instead of the workspace.

Framework/Algorithms/test/ConvertSpectrumAxisTest.h
```diff
@@ -168,4 +168,40 @@ class ConvertSpectrumAxisTest : public CxxTest::TestSuite {
   }
 };
 
+class ConvertSpectrumAxisTestPerformance : public CxxTest::TestSuite {
```
ianbush: The performance test will be best in version 2 of the algorithm only.

ianbush (commented):
`ConvertUnitsUsingDetectorTable` and `ConvertSpectrumAxis.cpp` still need updating to use `DetectorInfo`.

Framework/Algorithms/src/ConvertDiffCal.cpp
OwenArnold: Good catch!

Framework/Algorithms/src/ConvertSpectrumAxis2.cpp
```diff
@@ -153,13 +153,23 @@ void ConvertSpectrumAxis2::createElasticQMap(API::Progress &progress,
   else if (emodeStr == "Indirect")
     emode = 2;
 
-  auto &spectrumInfo = inputWS->spectrumInfo();
+  const auto &spectrumInfo = inputWS->spectrumInfo();
+  const auto &detectorInfo = inputWS->detectorInfo();
   for (size_t i = 0; i < nHist; i++) {
     double theta(0.0), efixed(0.0);
     if (!spectrumInfo.isMonitor(i)) {
```
ianbush: There seems to be an assumption here that a detector exists for a spectrum. Should there be a check? Perhaps not as there was not one before.

Framework/Algorithms/src/ConvertSpectrumAxis2.cpp
```diff
@@ -153,13 +153,23 @@ void ConvertSpectrumAxis2::createElasticQMap(API::Progress &progress,
...
       theta = 0.5 * spectrumInfo.twoTheta(i);
-      efixed =
-          getEfixed(spectrumInfo.detector(i), inputWS, emode); // get efixed
+      /*
+       * Two assumptions made in the following code.
+       * 1. Getting the detector index of the first detector in the spectrum
+       * defnition is enough (this should be completely safe).
```
ianbush: Typo in definition.

Framework/Algorithms/src/ConvertSpectrumAxis2.cpp
```diff
@@ -143,8 +145,6 @@ void ConvertSpectrumAxis2::createElasticQMap(API::Progress &progress,
                                              const std::string &targetUnit,
                                              API::MatrixWorkspace_sptr &inputWS,
                                              size_t nHist) {
```
ianbush: I think the `nHist` argument can be removed, and `spectrumInfo.size()` can be used instead.

OwenArnold:
Thanks for the comments. I'll see to all of that.

ianbush:
I ran the performance tests for `ConvertSpectrumAxis2` - there is no change.

Looks good, :shipit: once the tests pass.

Framework/Geometry/src/Instrument.cpp
```diff
@@ -1306,6 +1239,86 @@ void Instrument::setDetectorInfo(
...
+  }
 
+  // Now detPos will be set with respect to samplePos
+  Kernel::V3D relDetPos = detPos - samplePos;
+  // 0.5*cos(2theta)
+  double l2 = relDetPos.norm();
+  double halfcosTwoTheta =
+      relDetPos.scalar_prod(beamline) / (l2 * beamline_norm);
+  // This is sin(theta)
+  double sinTheta = sqrt(0.5 - halfcosTwoTheta);
```
SimonHeybrock: Why keep this weird way of getting `theta`? Can't you simply use `DetectorInfo::twoTheta`?

SimonHeybrock: Some client code would also get easier, since it does not have to mess around with computing the beamline and its norm.

OwenArnold: Yes. I originally decided to move and rename that block of code, but I should probably simplify it as you say.

SimonHeybrock: I would not have minded keeping it (much), but since you moved/renamed it I'd prefer cleaning it up.

Framework/Algorithms/src/ConvertSpectrumAxis2.cpp
```diff
@@ -153,13 +149,24 @@ void ConvertSpectrumAxis2::createElasticQMap(API::Progress &progress,
...
-  auto &spectrumInfo = inputWS->spectrumInfo();
+  const auto &spectrumInfo = inputWS->spectrumInfo();
+  const auto &detectorInfo = inputWS->detectorInfo();
+  const size_t nHist = spectrumInfo.size();
   for (size_t i = 0; i < nHist; i++) {
     double theta(0.0), efixed(0.0);
     if (!spectrumInfo.isMonitor(i)) {
       theta = 0.5 * spectrumInfo.twoTheta(i);
-      efixed =
-          getEfixed(spectrumInfo.detector(i), inputWS, emode); // get efixed
```
SimonHeybrock: Why this change?

Framework/Algorithms/src/ConvertUnitsUsingDetectorTable.cpp
```diff
@@ -147,7 +148,7 @@ MatrixWorkspace_sptr ConvertUnitsUsingDetectorTable::convertViaTOF(
 
   // Perform Sanity Validation before creating workspace
   size_t checkIndex = 0;
-  int checkSpecNo = inputWS->getDetector(checkIndex)->getID();
+  int checkSpecNo = inputWS->getSpectrum(checkIndex).getSpectrumNo();
```
SimonHeybrock: Doesn't this change the behavior of the algorithm? Does the column really contain spectrum numbers, or is the name `"spectra"` misleading?

Framework/Algorithms/src/ConvertSpectrumAxis2.cpp
OwenArnold: See below changed the `getEFixed` method to take a `DetectorInfo` rather than a `MatrixWorkspace`.

Framework/Algorithms/src/CreateLogTimeCorrection.cpp
```diff
@@ -77,49 +78,41 @@ void CreateLogTimeCorrection::exec() {
...
-  V3D samplepos = sample->getPos();
-
-  IComponent_const_sptr source = m_instrument->getSource();
-  if (!source) {
-    throw runtime_error("No source has been set.");
-  }
-  V3D sourcepos = source->getPos();
-  m_L1 = sourcepos.distance(samplepos);
+double
+CreateLogTimeCorrection::getInstrumentSetup(const DetectorInfo &detectorInfo) {
```
SimonHeybrock: This function seems a bit useless now. Have you considered removing `m_l2map` and `m_correctionMap` and simply using `DetectorInfo`? That would get rid of at least 2 map lookups.

Framework/Algorithms/src/ConvertSpectrumAxis2.cpp
SimonHeybrock: @OwenArnold My question was, why are we now getting EFixed for a detector instead of a spectrum? Isn't that a behaviour change?

Framework/Algorithms/src/CreateLogTimeCorrection.cpp
OwenArnold: I'll take a look.

Framework/Algorithms/src/ConvertUnitsUsingDetectorTable.cpp
OwenArnold: Agreed. It will. From the Algorithm documentation, I'm led to believe that this should actually be the spectrum number.

Framework/Algorithms/src/DetectorDiagnostic.cpp
```diff
@@ -559,9 +559,11 @@ DetectorDiagnostic::makeMap(API::MatrixWorkspace_sptr countsWS) {
...
   }
 
-  for (size_t i = 0; i < countsWS->getNumberHistograms(); i++) {
-    detid_t d = (*(countsWS->getSpectrum(i).getDetectorIDs().begin()));
-    auto anc = instrument->getDetector(d)->getAncestors();
+  const SpectrumInfo &spectrumInfo = countsWS->spectrumInfo();
+
+  for (size_t i = 0; i < countsWS->getNumberHistograms(); ++i) {
+
+    auto anc = spectrumInfo.detector(i).getAncestors();
```
SimonHeybrock: What is `getAncestors` doing for `DetectorGroup`? Unless it gives the ancestors of the first detector, this changes behavior.

Framework/Algorithms/src/ConvertSpectrumAxis2.cpp
OwenArnold: In a round about way it's the same thing. You're implementation in `SpectrumInfo` takes the first detector in the `SpectrumDefinition`. The code I wrote does the same thing. I review this again in any case.

SimonHeybrock: Oh, right I missed that changed bit in the implementation of `getEfixed`. It is fine then :+1: 

But actually you can then remove one `if`, I think?
https://github.com/mantidproject/mantid/pull/18528/files/e79bb8064b113dc8d02e8e3b8f8b6e26e2e8ad4f#diff-29d57758573e49db6d230ece2344892cR266

Framework/Algorithms/src/DetectorDiagnostic.cpp
OwenArnold: Looks like for a `DetectorGroup` the current implementation just returns an empty vector.

Framework/Algorithms/src/ConvertSpectrumAxis2.cpp
```diff
@@ -232,39 +239,40 @@ MatrixWorkspace_sptr ConvertSpectrumAxis2::createOutputWorkspace(
...
                                     "workspace. Please provide a value.");
       }
     } else if (emode == 2) {
-      std::vector<double> efixedVec = detector.getNumberParameter("Efixed");
+
+      const auto &detectorSingle = detectorInfo.detector(detectorIndex);
+
+      std::vector<double> efixedVec =
+          detectorSingle.getNumberParameter("Efixed");
       if (efixedVec.empty()) {
```
SimonHeybrock: @OwenArnold This is what I was referring to above. This branch covers the case of `DetectorGroup` (in that case `detector.getNumberParameter("Efixed");` would have returned something empty, I think.

OwenArnold: You're right about the `DetectorGroup`, but I think that `Component::getNumberParameter` is also capable of returning something empty for the plain `Detector`.

SimonHeybrock:
@OwenArnold Latest changes look good to me :+1: . According to code comments you are still working on `SaveDSpacemap`?

OwenArnold:
@SimonHeybrock I think both `SaveDspacemap` and `LoadDSpacemap` need an overhaul. I'm going to track down the original developer and talk to them about it. That's going to be a separate issue though.

SimonHeybrock:
@OwenArnold Ok, if you remove the comments here https://github.com/mantidproject/mantid/pull/18528/files#diff-d6adb85ec45b79d3b67b7e4432020684R78 I fine with merging.

@ianbush Since there were some substantial changes after your review, could you have another look to see if those are fine, before I merge.

ianbush:
@SimonHeybrock @OwenArnold These further changes still look good to me.

`ConvertUnitsUsingDetectorTable` is a bit of a mess - the name of the algorithm seems to contradict the documentation. As this is marked as deprecated I do not think this is important though.

PR #18549, opened 2017-01-25
Fixed locale issues in Sample Transmission Calculator

### LINKED ISSUE #18546: Locale issue in Sample Transmission Calculator
Sample Transmission Calculator fails to calculate anything if the system locale uses a different decimal separator than a dot.

soininen:
This PR fixes the locale issues in the Input Wavelength Range number fields when Type is set to Single. On a locale which doesn't use dots for decimal separators, the text which was directly read from the fields wasn't accepted by the Rebin algorithm. Now, the field contents are converted to strings using the 'C' locale.

**To test:**
Change the system locale to something which uses commas as decimal separators. For example, French, German or Finnish should do the job. Open the Sample Transmission Calculator and make sure the Wavelength range doesn't give errors in Mantid log. 

Fixes #18546.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidQt/CustomInterfaces/src/SampleTransmission.cpp
+++ b/MantidQt/CustomInterfaces/src/SampleTransmission.cpp
@@ -116,11 +116,11 @@ void SampleTransmission::calculate() {
   switch (wavelengthBinning) {
   // Single
   case 0: {
-    QStringList params;
-    params << m_uiForm.spSingleLow->text() << m_uiForm.spSingleWidth->text()
-           << m_uiForm.spSingleHigh->text();
-    QString binString = params.join(",");
-    transCalcAlg->setProperty("WavelengthRange", binString.toStdString());
+    // Convert values to binning params using the 'C' locale.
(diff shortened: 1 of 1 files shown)


KarlPalmen:
I've been unable reproduce the issue on an unfixed version of Mantid using Windows7. It might be, because I didn't change locale correctly or it might be that the issue only applies to some operating systems.

Please give the operating system in which the bug was observed and its locale settings.

KarlPalmen:
One curious thing I noticed (in Windows 7), is that if my date and time language is set to English, typing a comma into a wavelength box makes a dot, but if I do it with French date and time language, no character is made.
This applies to both fixed and unfixed and I think it is a Windows issue.

KarlPalmen:
I found that in Windows, I could explicitly change the decimal separator (rather than implicitly by changing locale). On doing so, I was able to reproduce the bug and verify it has been fixed. :shipit:

PR #18561, opened 2017-01-26
Add multiple time regime support for ISIS loading

### LINKED ISSUE #18560: Enable file loading for multiple file regimes
Rob is using Nexus files which have three time regimes. The file fails to load because we check the total number of monitors and detectors against the `vms_compat` entry `NSP1` which is not correct in the case of multiple time regimes.

If we see that the test fails, we should check if the file contains multiple time regimes and if so, then we should continue processing and not throw an exception.

This feature is high risk, since it is part of the core.

AntonPiccardoSelg:
These changes allow for multiple time regime loading of ISIS data. Our loading works correctly for files which have up to two time regimes. Rob has been using three time regimes which causes our LoadISISNexus2 algorithm to trip since it compares the number of detectors + monitors with the entry in `NSP1` of the `vms_compat`. This comparison is not valid for this scenario, therefore the comparison is disabled when loading  

Fixes #18560 

**To test:**

#### Test that can load a multiple time regime file

1. Take the file "LARMOR00009848.nxs" from the unit test repository. It contains three time regimes. If you try and load it into an old Mantid version without this fix, you should get an error
1. Load the file
    * Confirm that the file can be loaded without any issues

#### That that can load a normal file where the NSP1 is incorrect

You can get the sample file from here: \\olympic\Babylon5\Scratch\Anton\BAD_LOQ_FILE_WRONG_NSP1

1. Get the file "LOQ74019.nxs" from the specified location. Note that I tampered with the file and changed the NSP1 entry. Once you have used it for testing, it might be best to delete it.
1. Try to load it into Mantid
     * Confirm that this is not possible and an error occurs.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


DavidFair:
Is this worth adding to the release notes?

AntonPiccardoSelg:
@DavidFair Good point. In principle yes, but there will be most likely follow up work during this release which will make this "obsolete".

DavidFair:
Ok - I completed the tests locally so I'm happy to :shipit:

PR #18582, opened 2017-01-26
Update buildscript for python3

rosswhitfield:
When building a python3 job this adds `-python3` to the package_suffix which is then used to set the package_name and install_prefix thereby having unique named packages and allowing parallel installs of python 2 and 3 versions of Mantid on Ubuntu. This should also work with fedora rpms.

This also disables running docs-test for pull requests when using python 3 which is needed for when we add a python3 build to pull requests.

**To test:**
 * Code review
 * You can test the `buildscript` by running and comparing the output of running via `sh -x buildscript` after setting different environment variables, _e.g._ `export NODE_LABELS=ubuntu` and `export JOB_NAME=pull_requests-ubuntu`/`export JOB_NAME=pull_requests-ubuntu-python3` 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/buildconfig/Jenkins/buildscript
+++ b/buildconfig/Jenkins/buildscript
@@ -185,6 +185,11 @@ if [[ "$BUILDPKG" == true ]]; then
       PACKAGE_SUFFIX="unstable"
     fi
 
+    # Add '-python3' to package name and install path
+    if [[ ${JOB_NAME} == *python3* ]]; then
+	PACKAGE_SUFFIX=${PACKAGE_SUFFIX}-python3
+    fi
+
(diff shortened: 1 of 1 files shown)


martyngigg (approved):
Changes look sensible.

I ran them locally and checked that the package name is now `mantidunstable-python3` and the install prefix in `CPackConfig.cmake` is set to `/opt/mantidunstable-python3`.

At some point we will need to update the package dependencies so they are dependent on the `python3-` variants and not the standard `python-` ones.

I don't think we need to do that here though.

:shipit:

PR #18586, opened 2017-01-27
Update buildscript to avoid running system tests where possible

martyngigg:
Description of work.

This utilizes the `check_for_changes` script to decide whether there have been change that warrant
running the system tests.

**To test:**

Thorough review of changes.

*No issue number*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/buildconfig/Jenkins/check_for_changes
+++ b/buildconfig/Jenkins/check_for_changes
@@ -7,7 +7,7 @@ NOTFOUND=0
 TYPE=${1}
 # Assume the current commit is the merged commit created by GitHub.
 # The first and second parents will be the PR target branch & PR branch
-# that is being merged, resepectively. We want the files changes between the 
+# that is being merged, resepectively. We want the files changes between the
 # base and tail commits of the branch
 BRANCH_TIP=$(git rev-parse HEAD^2)
 BRANCH_BASE=$(git merge-base HEAD^2 HEAD^1)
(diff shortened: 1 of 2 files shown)


stuartcampbell:
I like the changes, I think it makes it clearer and puts all the logic on what build is actually being run in a central place so if you need to change it you don't have to worry about multiple `if`s.

I would say :shipit: but you may not care about my opinion anymore! 😆

martyngigg:
We will always value your opinion!

Thanks for looking over it.

martyngigg:
@rosswhitfield I know you've had concerns in the past over the robustness of checking for changes. What do think?

rosswhitfield:
Something isn't right, shouldn't the systemtests have run on *this* pull request since you made changes to the `buildconfig` directory which isn't one we are excluding? http://builds.mantidproject.org/job/pull_requests-rhel7/11506/console

buildconfig/Jenkins/check_for_changes
```diff
@@ -36,6 +36,16 @@ case "$TYPE" in
...
         exit $NOTFOUND
     ;;
+    docs-gui-only)
+	# FOUND=1 iff changes are limited to docs or GUI only
+	# Find all changed files and grep for required type. -v inverts match so grep=0 means
+	# there are other changes besides this
+	if git diff --name-only ${BRANCH_TIP} ${BRANCH_BASE} -- | grep -q -E -v '^docs/|^MantidQt/|^MantidPlot/'; then
+	    exit $NOTFOUND
+	else
+	    exit $FOUND
```
rosswhitfield: I think you just need to switch the `NOTFOUND` and `FOUND` here.

martyngigg: Thanks for catching that. I hadn't checked the system test builds yet.

Yes, you're right. We should `exit 1` if we have something other than those things changed so that then the outer `buildscript` will take the else branch. :+1:

PR #18593, opened 2017-01-27
SpectrumInfo rollout: algorithms SetInstrumentParameter to PredictPeaks

### LINKED ISSUE #18554: SpectrumInfo and DetectorInfo rollout: algorithms <SetInstrumentParameter to PredictPeaks>
[SetInstrumentParameter](http://docs.mantidproject.org/nightly/algorithms/SetInstrumentParameter-v1.html)
[RingProfile](http://docs.mantidproject.org/nightly/algorithms/RingProfile-v1.html)
[AddPeakHKL](http://docs.mantidproject.org/nightly/algorithms/AddPeakHKL-v1.html)
[PeaksInRegion](http://docs.mantidproject.org/nightly/algorithms/PeaksInRegion-v1.html)
[IntegratePeakTimeSlices](http://docs.mantidproject.org/nightly/algorithms/IntegratePeakTimeSlices-v1.html)
[FindSXPeaks](http://docs.mantidproject.org/nightly/algorithms/FindSXPeaks-v1.html)
[PredictPeaks](http://docs.mantidproject.org/nightly/algorithms/PredictPeaks-v1.html)

soininen:
## Description of work.
This PR is part of refactoring effort to use the new interface of the SpectrumInfo and DetectorInfo modules.

The following algorithms were refactored to use SpectrumInfo/DetectorInfo:
[SetInstrumentParameter](http://docs.mantidproject.org/nightly/algorithms/SetInstrumentParameter-v1.html) (no changes)
[RingProfile](http://docs.mantidproject.org/nightly/algorithms/RingProfile-v1.html) (no changes)
[AddPeakHKL](http://docs.mantidproject.org/nightly/algorithms/AddPeakHKL-v1.html) (no changes)
[PeaksInRegion](http://docs.mantidproject.org/nightly/algorithms/PeaksInRegion-v1.html)
[IntegratePeakTimeSlices](http://docs.mantidproject.org/nightly/algorithms/IntegratePeakTimeSlices-v1.html)
[FindSXPeaks](http://docs.mantidproject.org/nightly/algorithms/FindSXPeaks-v1.html)
[PredictPeaks](http://docs.mantidproject.org/nightly/algorithms/PredictPeaks-v1.html) (no changes)

## To test.

Code review.

Fixes #18554.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Crystal/inc/MantidCrystal/FindSXPeaks.h
+++ b/Framework/Crystal/inc/MantidCrystal/FindSXPeaks.h
@@ -9,7 +9,6 @@
 #include "MantidAPI/SpectrumInfo.h"
 #include "MantidDataObjects/PeaksWorkspace.h"
 
-
 namespace Mantid {
 namespace Crystal {
 
@@ -26,8 +25,7 @@ public:
   @param spectrumInfo: spectrum info of the original ws.
   */
   SXPeak(double t, double phi, double intensity,
-         const std::vector<int> &spectral,
-         const size_t wsIndex,
+         const std::vector<int> &spectral, const size_t wsIndex,
          const API::SpectrumInfo &spectrumInfo)
       : _t(t), _phi(phi), _intensity(intensity), _spectral(spectral),
         _wsIndex(wsIndex) {
@@ -40,8 +38,9 @@ public:
           "SXPeak: Cannot have zero sized spectral list");
     }
     if (!spectrumInfo.hasUniqueDetector(_wsIndex)) {
-      throw std::invalid_argument(
-            "SXPeak: Spectrum at ws index " + std::to_string(wsIndex) + " has more than a single detector");
+      throw std::invalid_argument("SXPeak: Spectrum at ws index " +
+                                  std::to_string(wsIndex) +
+                                  " has more than a single detector");
     }
     _th2 = spectrumInfo.twoTheta(_wsIndex);
     _Ltot = spectrumInfo.l1() + spectrumInfo.l2(_wsIndex);
--- a/Framework/Crystal/inc/MantidCrystal/FindSXPeaks.h
+++ b/Framework/Crystal/inc/MantidCrystal/FindSXPeaks.h
@@ -37,10 +37,10 @@ public:
       throw std::invalid_argument(
           "SXPeak: Cannot have zero sized spectral list");
     }
-    if (!spectrumInfo.hasUniqueDetector(_wsIndex)) {
+    if (!spectrumInfo.hasDetectors(_wsIndex)) {
       throw std::invalid_argument("SXPeak: Spectrum at ws index " +
                                   std::to_string(wsIndex) +
-                                  " has more than a single detector");
+                                  " doesn't have detectors");
     }
     _th2 = spectrumInfo.twoTheta(_wsIndex);
     _Ltot = spectrumInfo.l1() + spectrumInfo.l2(_wsIndex);
--- a/Framework/Crystal/inc/MantidCrystal/FindSXPeaks.h
+++ b/Framework/Crystal/inc/MantidCrystal/FindSXPeaks.h
@@ -6,7 +6,9 @@
 //----------------------------------------------------------------------
 #include "MantidAPI/Algorithm.h"
 #include "MantidAPI/IPeaksWorkspace_fwd.h"
+#include "MantidAPI/SpectrumInfo.h"
 #include "MantidDataObjects/PeaksWorkspace.h"
+
 
 namespace Mantid {
 namespace Crystal {
@@ -17,20 +19,18 @@ public:
   /**
   Constructor
   @param t : tof
-  @param th2 : 2 * theta angle
   @param phi : psi angle
   @param intensity : peak intensity
   @param spectral : contributing spectra
-  @param Ltot : detector-sample absolute distance
-  @param detectorId : id of the contributing detector
-  @param inst: geometry of the instrument
+  @param wsIndex : ws index of the contributing spectrum
+  @param spectrumInfo: spectrum info of the original ws.
   */
-  SXPeak(double t, double th2, double phi, double intensity,
-         const std::vector<int> &spectral, double Ltot,
-         Mantid::detid_t detectorId,
-         Mantid::Geometry::Instrument_const_sptr inst)
-      : _t(t), _th2(th2), _phi(phi), _intensity(intensity), _Ltot(Ltot),
-        _detectorId(detectorId), _inst(inst) {
+  SXPeak(double t, double phi, double intensity,
+         const std::vector<int> &spectral,
+         const size_t wsIndex,
+         const API::SpectrumInfo &spectrumInfo)
+      : _t(t), _phi(phi), _intensity(intensity), _spectral(spectral),
+        _wsIndex(wsIndex) {
     // Sanity checks
     if (intensity < 0) {
       throw std::invalid_argument("SXPeak: Cannot have an intensity < 0");
@@ -39,12 +39,28 @@ public:
       throw std::invalid_argument(
           "SXPeak: Cannot have zero sized spectral list");
     }
-    if (Ltot < 0) {
+    if (!spectrumInfo.hasUniqueDetector(_wsIndex)) {
+      throw std::invalid_argument(
+            "SXPeak: Spectrum at ws index " + std::to_string(wsIndex) + " has more than a single detector");
+    }
+    _th2 = spectrumInfo.twoTheta(_wsIndex);
+    _Ltot = spectrumInfo.l1() + spectrumInfo.l2(_wsIndex);
+    if (_Ltot < 0) {
       throw std::invalid_argument("SXPeak: Cannot have detector distance < 0");
     }
+    _detId = spectrumInfo.detector(_wsIndex).getID();
     npixels = 1;
-    _spectral.resize(spectral.size());
-    std::copy(spectral.begin(), spectral.end(), _spectral.begin());
+
+    const auto samplePos = spectrumInfo.samplePosition();
+    const auto sourcePos = spectrumInfo.sourcePosition();
+    const auto detPos = spectrumInfo.position(_wsIndex);
+    // Normalized beam direction
+    auto beamDir = samplePos - sourcePos;
+    beamDir.normalize();
+    // Normalized detector direction
+    auto detDir = (detPos - samplePos);
+    detDir.normalize();
+    _unitWaveVector = beamDir - detDir;
   }
   /**
   Object comparision
@@ -69,17 +85,6 @@ public:
   @return q vector
   */
   Mantid::Kernel::V3D getQ() const {
-    auto samplePos = _inst->getSample()->getPos();
-    auto sourcePos = _inst->getSource()->getPos();
-    auto detPos = _inst->getDetector(_detectorId)->getPos();
-
-    // Normalized beam direction
-    auto beamDir = samplePos - sourcePos;
-    beamDir /= beamDir.norm();
-    // Normalized detector direction
-    auto detDir = (detPos - samplePos);
-    detDir /= detDir.norm();
-
     double vi = _Ltot / (_t * 1e-6);
     // wavenumber = h_bar / mv
     double wi =
@@ -89,7 +94,7 @@ public:
     // wavevector=1/wavenumber = 2pi/wavelength
     double wvi = 1.0 / wi;
     // Now calculate the wavevector of the scattered neutron
-    return (beamDir - detDir) * wvi;
+    return _unitWaveVector * wvi;
   }
 
   /**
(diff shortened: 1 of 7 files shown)


Framework/Crystal/test/PeaksInRegionTest.h
```diff
@@ -30,7 +31,7 @@ class PeaksInRegionTest : public CxxTest::TestSuite {
                        double yMaxFromPeak, double zMinFromPeak,
                        double zMaxFromPeak) {
     PeaksWorkspace_sptr ws = WorkspaceCreationHelper::createPeaksWorkspace(1);
-    auto detectorIds = ws->getInstrument()->getDetectorIDs();
+    const auto detectorIds = ws->detectorInfo().detectorIDs();
```
SimonHeybrock: reference

Framework/Crystal/inc/MantidCrystal/FindSXPeaks.h
```diff
@@ -39,12 +37,29 @@ struct DLLExport SXPeak {
       throw std::invalid_argument(
           "SXPeak: Cannot have zero sized spectral list");
     }
-    if (Ltot < 0) {
+    if (!spectrumInfo.hasUniqueDetector(_wsIndex)) {
+      throw std::invalid_argument("SXPeak: Spectrum at ws index " +
+                                  std::to_string(wsIndex) +
+                                  " has more than a single detector");
```
SimonHeybrock: I can't find the equivalent of this check/exception in the original version?

Framework/Crystal/src/IntegratePeakTimeSlices.cpp
```diff
@@ -259,18 +259,14 @@ void IntegratePeakTimeSlices::exec() {
...
   TableWorkspace_sptr TabWS = boost::make_shared<TableWorkspace>(0);
 
   //----------------------------- get Peak extents
   //------------------------------
   try {
-
-    // Find the workspace index for this detector ID
-    detid2index_map::const_iterator it = m_wi_to_detid_map.find(detID);
-    size_t wsIndx = (it->second);
+    const size_t wsIndx = detectorInfo.indexOf(detID);
```
SimonHeybrock: This is not what you need. The index returned by `indexOf` is an index for the detector, but you need an index for the detector group (i.e., the index of spectrum which contains the detector). I think you need to keep the old way of obtaining the index.

Framework/Crystal/src/IntegratePeakTimeSlices.cpp
```diff
@@ -1570,13 +1566,14 @@ void IntegratePeakTimeSlices::SetUpData1(
...
+  const auto &detectorInfo = inpWkSpace->detectorInfo();
   for (int i = 2; i < m_NeighborIDs[1]; i++) {
     int DetID = m_NeighborIDs[i];
 
     size_t workspaceIndex;
-    if (m_wi_to_detid_map.count(DetID) > 0)
-      workspaceIndex = m_wi_to_detid_map.find(DetID)->second;
-    else {
+    try {
+      workspaceIndex = detectorInfo.indexOf(DetID);
```
SimonHeybrock: See above.

Framework/Crystal/test/IntegratePeakTimeSlicesTest.h
```diff
@@ -97,44 +99,36 @@ class IntegratePeakTimeSlicesTest : public CxxTest::TestSuite {
...
-    double dummy, phi;
-    pos1.getSpherical(dummy, phi, dummy);
-    double ScatAng = phi / 180 * M_PI;
+    const auto &spectrumInfo = wsPtr->spectrumInfo();
+    const auto &detectorInfo = wsPtr->detectorInfo();
+    const auto detID = pixelp->getID();
+    const auto wsIndex = detectorInfo.indexOf(detID);
+    const double L1 = spectrumInfo.l1();
+    double L2 = spectrumInfo.l2(wsIndex);
+    double ScatAng = detectorInfo.twoTheta(wsIndex) / 180 * M_PI;
```
SimonHeybrock: See above. `DetectorInfo::indexOf` does *not* return a workspace index. Doesn't `SpectrumInfo` simply give you what you need here?

Framework/Crystal/test/IntegratePeakTimeSlicesTest.h
```diff
@@ -97,44 +99,36 @@ class IntegratePeakTimeSlicesTest : public CxxTest::TestSuite {
...
-    const detid2index_map map = wsPtr->getDetectorIDToWorkspaceIndexMap(true);
-
     for (int row = 0; row < NRC; row++)
       for (int col = 0; col < NRC; col++) {
 
         boost::shared_ptr<Detector> detP = bankR->getAtXY(col, row);
-
-        detid2index_map::const_iterator it = map.find(detP->getID());
-        size_t wsIndex = (*it).second;
+        const auto wsIndex = detectorInfo.indexOf(detP->getID());
```
SimonHeybrock: See above.

Framework/Crystal/test/IntegratePeakTimeSlicesTest.h
```diff
@@ -237,18 +233,16 @@ class IntegratePeakTimeSlicesTest : public CxxTest::TestSuite {
...
    *   Calculates Q
    */
-  double calcQ(RectangularDetector_const_sptr bankP,
-               boost::shared_ptr<const Instrument> instPtr, int row, int col,
+  double calcQ(const DetectorInfo &detectorInfo, const detid_t detID,
                double time) {
-    boost::shared_ptr<Detector> detP = bankP->getAtXY(col, row);
-
-    double L2 = detP->getDistance(*(instPtr->getSample()));
+    const auto wsIndex = detectorInfo.indexOf(detID);
```
SimonHeybrock: This is not the workspace index, rename the variable.

Framework/Crystal/test/IntegratePeakTimeSlicesTest.h
```diff
@@ -237,18 +233,16 @@ class IntegratePeakTimeSlicesTest : public CxxTest::TestSuite {
...
+    const auto L2 = detectorInfo.l2(wsIndex);
 
     Kernel::Units::MomentumTransfer Q;
     std::vector<double> x;
     x.push_back(time);
-    double L1 = instPtr->getSample()->getDistance(*(instPtr->getSource()));
-    Kernel::V3D pos = detP->getPos();
+    const auto L1 = detectorInfo.l1();
+    const auto pos = detectorInfo.position(wsIndex);
     double ScatAng = fabs(asin(pos.Z() / pos.norm()));
```
SimonHeybrock: Probably you can use `DetectorInfo::twoTheta` to get this simpler?

SimonHeybrock (commented):
Some comments below:

Framework/Crystal/src/IntegratePeakTimeSlices.cpp
soininen: Will do.

Framework/Crystal/inc/MantidCrystal/FindSXPeaks.h
SimonHeybrock: The test you added forbids detector groups (spectra with more than 1 detector), so unless that is the behavior of the original code you cannot add this test. Maybe a check for `spectrumInfo.hasDetectors(i)` would make sense.

Framework/Crystal/test/IntegratePeakTimeSlicesTest.h
soininen: L1 and sample position, yes. I think I should leave the rest as-is as the L2 has to be calculated for an individual detector and they anyway need the phi angle which is not supplied by SpectrumInfo.

Framework/Crystal/test/IntegratePeakTimeSlicesTest.h
soininen: Not easily. They refer to the detector by its row and column in a detector bank. I could pass DetectorInfo to the method and get the L1 from it, though, but is it really worth it?

Framework/Crystal/test/IntegratePeakTimeSlicesTest.h
soininen: I confused the `index` in `DetectorInfo` with workspace indices as they have the same name and `size_t` type. Hopefully I got it more correct now.

Framework/Crystal/inc/MantidCrystal/FindSXPeaks.h
soininen: Good poin. I'll change the test to `hasDetectors()`, then.

Framework/Crystal/inc/MantidCrystal/FindSXPeaks.h
```diff
@@ -69,17 +84,6 @@ struct DLLExport SXPeak {
   @return q vector
   */
   Mantid::Kernel::V3D getQ() const {
-    auto samplePos = _inst->getSample()->getPos();
-    auto sourcePos = _inst->getSource()->getPos();
-    auto detPos = _inst->getDetector(_detectorId)->getPos();
```
SimonHeybrock: Looking at this gain, I think this class should actually use `DetectorInfo`, not `SpectrumInfo`, if we want to keep it consistent with the old behavior (but read also below first!). The index you pass would then be an index of the detector (obtained with `DetectorInfo::indexOf(detid_t)`).

Framework/Crystal/src/FindSXPeaks.cpp
```diff
@@ -155,34 +152,20 @@ void FindSXPeaks::exec() {
...
 
-    double th2 = det->getTwoTheta(sample, L1);
-
     std::vector<int> specs(1, i);
 
-    Mantid::Kernel::V3D L2 = det->getPos();
-    L2 -= sample;
-
-    SXPeak peak(tof, th2, phi, *maxY, specs, l1 + L2.norm(), det->getID(),
-                localworkspace->getInstrument());
```
SimonHeybrock: This is a weird: If the spectrum in the workspace has grouped detectors, `localworkspace->getDetector(static_cast<size_t>(i));` above returns a `DetectorGroup`. `det->getID()` will then return the ID of the first detector in the group, so `SXPeak` would use the position of only that first detector! All the other parameters that are passed to the constructor here, such as `th2` are for the whole group!

So in conclusion, I think the way you changed `SXPeak` (to use `SpectrumInfo`) makes sense, since actually there is a bug here (please double check yourself, maybe I am overlooking something!)? Maybe you can open an issue for this with a short description and mark it as fixed by this PR?

SimonHeybrock (commented):
I think this looks good now, but read below first. If you can confirm, then this is ready:

Framework/Crystal/src/FindSXPeaks.cpp
soininen: You are correct with the detector group vs. single detector stuff here. For me it seems this is a problem with `Mantid::Geometry::IPeak` which is not designed to work with detector groups. The algorithm outputs a `PeaksWorkspace`, which consists of `IPeak`s which in turn hold information like *Q*, *Intensity* but also detector ID. Now, with my changes, the *Q* and *Intensity* can be for a detector group but the ID is still for a single detector. This could be fixed by getting these values for single detectors, but I suppose there is no way of getting *Intensity* for a single detector in a group, is there? We could also check for `hasUniqueDetector()` here and give warning/bail out but would this change in behavior be undesirable?

With this single detector ID going to the output workspace, I don't think I actually fixed anything here. I could switch to `DetectorInfo`, but then again the intensity data would come from the spectrum itself, if I understood things right. Maybe it would be the right thing to keep as close to the old behavior as possible? Let me know your thoughts about this.

SimonHeybrock: Yes, good point, the intensity is only available for the group. Maybe try adding the check for `hasUniqueDetector()` that you suggested and see if any test fail?

@mantidproject/gatekeepers Could someone else please have a look at this problem, I don't think I have enough knowledge of what is going on with peaks in Mantid to make a final decision here.

soininen: All tests pass even if the spectra are tested for unique detectors.

martyngigg: I'm not most expert on the peaks side of things so some one else should comment for that but in the case of a group of detectors it does not make any sense to try and pull out an intensity from an individual detector - that information has been lost when the grouping is performed (whether hardware or software). 

@soininen Do the tests actually use any grouping anywhere?

soininen: @martyngigg By a quick look, the tests don't seem to be using detector groups.

SimonHeybrock:
@soininen Since no one else is commenting regarding `FindSXPeaks`, I would suggest to add the test for unique detectors, and then I'll put the shipit.

SimonHeybrock:
Looks good, :shipit: once builds green.

PR #18594, opened 2017-01-27
ISIS Powder Scripts - Refactor and Migration of Pearl 

### LINKED ISSUE #18108: ISIS_Powder intermediate refactoring work

Having completed the initial migration of PEARL and POLARIS: #17839 #17949 we are now at a stage where we need to refactor what we have and consolidate the code. 

The work to be completed:

---
- [x] Move logic out of instrument classes and into sub modules so that each class only holds settings
- [x] Migrate `instrument_version` into `RunDetails` to reduce the number of params being passed around
- [x] Reduce the number of hooks by consolidating different steps into separate functions
- [x] Rename some of the hooks to better reflect their behaviour 
- [x] Change from using private members to using public / private API in routines
- [x] Remove the empty hooks in `abstract_inst`
- [x] Put hooks into coherent order
- [x] Move `_set_from_config_kwargs` from POLARIS into common. Instruments should pass their expected kwargs in
- [x] Implement advanced configuration and move current configuration out of POLARIS Class
- [x] Make sure those parameters are read by the relevant functions when they are needed and not passed around all over the place
- [x] Create basic setup file for users to specify directories for POLARIS
- [x] Reorganise common as it no longer has any natural ordering
- [x] ~~Support various vanadium processing modes - currently we support spline only~~
- [x] Merge offset and grouping files together
- [x] ~~Change the grouping input file so its easily distinguished from the automatically generated offset cal fine.~~ 
- [x] Ron tells me the script currently attempts to compensate for noisy/quiet detectors which can cause issues with making faulty quiet detectors loud or masking valid data. Look into this and possibly remove the logic (as they mask faulty detectors in the grouping cal file themselves)
- [x] `LowerLambda` and `UpperLambda` are specified in the file. These are specific to an instrument but Ron wasn't sure what they did. ~~If I could figure out in the scripts what they do this would also make future migrations easier.~~ Edit: They are used in solid Angle Corrections and have no empirical effect on data so the params are removed
- [x] Create some sample configuration scripts to show how they work together
---
Following PR feedback from #18096 :
- [x] Abstract instrument needs tidying and ordering like PEARL and POLARIS were
- [x] We need to sort the public / private API within Abstract Instrument
- [x] Use `MergeRuns` instead of `Plus` in the several places we have used it. It should handle the sample logs better too
- [x] Look into using Mantid to process the run number strings using the following code suggested by soininen:
```python
numberProperty = mantid.kernel.IntArrayProperty('dummy_property', processed_string)
return numberProperty.value
```
---
Feedback from demo
- [x] Vanadium runs should be summed but focused runs need to be summed/individually
- [x] Focus mode should default to processing individually - the workspace left in Mantid should always be the final processed one
- [x] Basic config should hold the path to the calibration mapping
- [x] The calibration mapping should allow unbounded runs 
- [x] Exceptions in calibration mapping needs to be removed
- [x] Automatically generate vanadium corrections with a warning if they could not be found
- [x] Make Solid Angle default to off and move to advanced config
- [x] Divide SAC/No_SAC to determine if its an offset or linear correction from SAC
- [x] Add Multiple Scatter OR Absorption for Vanadium
- [x] Process advanced settings first as defaults and allow override from Basic config
- [x] Set maximum number of runs to 1000 so they can process all runs from a cycle
- [x] Override POLARIS to output to .GSAS instead of .GSS (Possibly applies for PEARL?)
- [x] Offsets should only be read from Vanadium/Focus routines make sure that is the case. Check against PEARL who generate these from silicon calibration
---
Pearl specific:
- [x] Remove save range from PEARL - depends on removing support for old instruments 
- [x] Remove _get_label_information in `PEARL` - Need to move to calibration mapping
- [x] Remove `cycle_dict` from `PEARL` - Need to move to calibration mapping
- [x] Change alg_save_ranges to be number of banks and move save ranges into `PEARL`
- [x] Split the old API out of PEARL into its own subclass
- [x] `PEARL` See if we can pull the monitor directly from the WS instead of creating a new one
- [x] Remove `spline_number` from the spline background hook and move into `PEARL` class
- [x] See what effect removing the PEARL specific rebin hooks have on the data and if possible unify where rebinning happens between PEARL and POLARIS
- [x] Basic config for POLARIS
- [x] See if we can get PEARL using `extract_bank_spectra` from common
- [x] System tests - Build system tests for PEARL which do not rely on `pearl_routines`
- [x] Migrate to a calibration mapping similar to POLARIS following discussion
- [x] Save out splined vanadium as part of calibrate vanadium (as per POLARIS)
- [x] ~~Combine grouping and calibration files~~
- [x] `get_run_details` has a lot of common elements between the two instruments - move these to common after YAML is implemented on `PEARL`
- [x] Confirm whether we need all the edge cases for older instruments - are they still doing data reduction with these older runs
- [x] Implement advanced configuration and move current configuration out of PEARL Class
---
Feedback from PEARL:
- [x] Remove `pearl_routines` they have their own copies of the script and these have further modifications. Therefore the Mantid copy is not needed (with all the workarounds)
- [x] Remove support for previous instrument generations (ditto about `pearl_routines`)
- [x] Create calibration mapping - they have two modes normal and long which need an empty/vanadium
- [x] Rename .gss to .gsas (with POLARIS above)
- [x] Unlike POLARIS, PEARL prefer their options to be set explicit every time. Focus uses: `f_mode` `do_atten` `tt_mode` whilst all modes use the `van_norm` and `long_mode` options
- [x] Attenuation file path is set once per script - move the file path out of the class and into a parameter
- [x] PEARL always sums data so set that parameter by default on call to `focus`

DavidFair:
Description of work.
This PR covers a large number of changes in order to bring the scripts up to standard whilst migrating Pearl onto them. The full list of changes is here #18108 . However some main points include:

- Pearl Instrument migrated on
- Initial documentation drafted - Note this is only enough to get a basic understanding of the scripts.
- Abstract and individual instrument classes tidied
- Better support and sanity checking for YAML config files
- All hard coded parameters moved out of the scripts into adv. configuration files
- Abstraction layer between parameter names and script parameter names. This will be utilised more and more as different instruments have different terminology. It helps abstract the script from changes in names
- `pearl_routines` completely removed. This is because the version Mantid has does not represent the local copies they are using so it is completely safe to complete this.

Still to do in future pull requests:
- Incorporate feedback / fixes following initial testing with Scientists.
- Remove `pearl_routines` documentation and links
- Tidying of instrument specific algorithms where needed
- Unit tests - This is high on the priority list. After the first round of changes with testers we should be far enough out of the territory of large changes to be able to implement this.
- System tests - These exist in some form however are disabled. Currently the output changes regularly but very subtly as we move around steps to make the algorithm correct following feedback. 
- Migration of additional instrument(s)
- Better documentation
- Create calibration mapping files that is representative of the data sets used 
- Tidy and include necessary files in system data directory

Due to the amount of changes in these scripts and setting up a calibration mapping with the offsets file this PR is not easy to test locally. 

Instead it is going to be tested and tweaked with the users getting custom / nightly builds (depending on the size of changes and the speed merging back into master) for the next few weeks. Therefore if it doesn't work 100% it will be picked up and fixed in future PRs

**To test:**

- Code review - Any and all suggestions / feedback is greatly appreciated 

Fixes #18108 .

Release notes. - Not included yet however added to #17831 when we are happy with the state of the scripts (e.g. Not still in active development)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


scripts/Diffraction/isis_powder/pearl_routines/pearl_output.py
```diff
@@ -50,49 +49,41 @@ def _focus_mode_mods(output_file_paths, calibrated_spectra):
...
-        no_att = output_file_paths["output_name"] + "_noatten"
-        mantid.CloneWorkspace(InputWorkspace=summed_ws, OutputWorkspace=no_att)
+def _focus_mode_all(output_file_paths, processed_spectra, perform_attenuation, instrument):
+    summed_spectra_name = output_file_paths["output_name"] + "_mods1-9"
+    summed_spectra = mantid.MergeRuns(InputWorkspaces=processed_spectra[:9], OutputWorkspace=summed_spectra_name)
 
-        summed_ws = mantid.ConvertUnits(InputWorkspace=summed_ws, Target="dSpacing")
-        summed_ws = instrument._attenuate_workspace(summed_ws)
-        summed_ws = mantid.ConvertUnits(InputWorkspace=summed_ws, Target="TOF")
+    summed_spectra = mantid.Scale(InputWorkspace=summed_spectra, Factor=0.111111111111111,
```
DimitarTasev: Do we know why it's 0.111111111111111 in two places, and then 0.3333333333333 for the last one?

scripts/Diffraction/isis_powder/polaris_routines/polaris_algs.py
```diff
@@ -0,0 +1,142 @@
...
+        chopper_config = yaml_dict["chopper_off"]
+
+    label = yaml_dict["label"]
+    empty_runs = chopper_config["empty_run_numbers"]
+    vanadium_runs = chopper_config["vanadium_run_numbers"]
+
+    splined_vanadium_name = _generate_splined_van_name(chopper_on=inst_settings.chopper_on,
+                                                       vanadium_run_string=vanadium_runs)
+
+    grouping_full_path = os.path.join(inst_settings.calibration_dir, inst_settings.grouping_file_name)
```
DimitarTasev: Same note as on the previous path comment, consider using expand and abspath if you think it's possible someone would pass `~` or env variables

dtasev (changes requested):
Looks like a very good improvement over the previous state of the scripts. This is just an initial overview of the changes, I now do some runs with it and have a play

scripts/Diffraction/isis_powder/routines/InstrumentSettings.py
```diff
@@ -0,0 +1,108 @@
...
+        if map_entry:
+            # User forgot to enter the param:
+            raise AttributeError("The parameter with name: '" + str(map_entry[0]) + "' is required but was not set or "
+                                 "passed.\nPlease set this configuration option and try again")
+        else:
+            # If you have got here from a grep or something similar this error message means the line caller
+            # has asked for a class attribute which does not exist. These attributes are set in a mapping file which
+            # is passed in whilst InstrumentSettings is being constructed. Check that the 'script name' (i.e. not user
+            # friendly name) is typed correctly in both the script(s) and mapping file.
+            raise AttributeError("The attribute in the script with name " + str(item) + " is unknown to the mapping."
```
DavidFair: I'm reluctant to change this one - I've worded it slightly but I want them to contact the development team as it means a variable name internal to the script is incorrect.
Above that I have put a long comment explaining to future developers if they do see this message what it means and what they need to do.

scripts/Diffraction/isis_powder/abstract_inst.py
```diff
@@ -39,238 +36,188 @@ def output_dir(self):
...
 
+    def generate_out_file_paths(self, run_details, output_directory=None):
+        """
+        Generates the various output paths and file names to be used during saving or as workspace names
+        :param run_details: The run details associated with this run
+        :param output_directory: The output directory to use as part of the output path
+        :return: A dictionary containing the various output paths and generated output name
+        """
+        if not output_directory:
+            output_directory = os.path.join(self._output_dir, run_details.label, self._user_name)
```
DavidFair: I've added a note in the roadmap to do the rest at a later date as low priority

scripts/Diffraction/isis_powder/routines/InstrumentSettings.py
```diff
@@ -0,0 +1,108 @@
...
+
+            # Update attributes from said dictionary
+            found_attribute = next((attr_tuple for attr_tuple in self._attr_mapping
+                                    if config_key == attr_tuple[0]), None)
+            if found_attribute:
+                # The first element of the attribute is the config name and the last element is the friendly name
+                self._update_attribute(attr_name=found_attribute[-1], attr_val=dict_to_parse[found_attribute[0]],
+                                       friendly_name=found_attribute[0], suppress_warnings=suppress_warnings)
+            elif not suppress_warnings:
+                warnings.warn("Ignoring unknown configuration key: " + str(config_key))
```
DavidFair: Good point I'll ask when I next see them and it can target the PR with initial fixes

scripts/Diffraction/isis_powder/pearl.py
```diff
@@ -1,557 +1,140 @@
...
+from isis_powder.routines.common_enums import InputBatchingEnum
 from isis_powder.abstract_inst import AbstractInst
-from isis_powder.pearl_routines import fmode_output, pearl_calib_factory, pearl_cycle_factory
-from isis_powder.routines.RunDetails import RunDetails
+from isis_powder.pearl_routines import pearl_algs, pearl_output, pearl_advanced_config, pearl_param_mapping
 
 
 class Pearl(AbstractInst):
+    def __init__(self, **kwargs):
+        basic_config_dict = yaml_parser.open_yaml_file_as_dictionary(kwargs.get("config_file", None))
```
DavidFair: So as all kwargs get forwarded to the `InstrumentSettings` class at construction it would subsequently warn (or throw is they choose that) the key was misspelt. 
Additionally the config_file is designed to move several keys out of the scripting window as they were becoming noise. If this key was misspelt those keys would be missing so it would fail too instead of trying to guess.

scripts/Diffraction/isis_powder/pearl_routines/pearl_output.py
DavidFair: Yes these sum up workspaces 1-9 (inclusive) so the value in the original scripts was 1/9 in decimal 
The other one sums up 3 workspaces so 1/3 in decimal

scripts/Diffraction/isis_powder/routines/InstrumentSettings.py
```diff
@@ -0,0 +1,108 @@
...
+        for expected_attr in expected_attr_names:
+            if not [attr_entry for attr_entry in self._attr_mapping if expected_attr == attr_entry[-1]]:
+                raise ValueError("Expected attribute '" + str(expected_attr) + "' is unknown to attribute mapping")
+
+        # Filter down the full mapping list
+        found_tuple_list = [tuple_entry for tuple_entry in self._attr_mapping if tuple_entry[-1] in expected_attr_names]
+        expected_params_dict = dict(found_tuple_list)
+        self._check_attribute_is_set(expected_params_dict)
+
+    def update_attributes(self, advanced_config=None, basic_config=None, kwargs=None, suppress_warnings=False):
```
DavidFair: Done - It will be in the final documentation as its an important concept but initial testers are being informed during training due to the amount we need to tell them

scripts/Diffraction/isis_powder/routines/InstrumentSettings.py
```diff
@@ -0,0 +1,108 @@
...
+        self._kwargs = kwargs if kwargs else self._kwargs
+
+        has_known_keys_already_been_printed = self._unknown_keys_found
+        # Only update if one in hierarchy below it has been updated
+        if advanced_config:
+            self._parse_attributes(self._adv_config_dict, suppress_warnings=suppress_warnings)
+        if advanced_config or basic_config:
+            self._parse_attributes(self._basic_conf_dict,
+                                   suppress_warnings=(not bool(basic_config or suppress_warnings)))
+        if advanced_config or basic_config or kwargs:
```
DavidFair: Added comment explaining why its done like this.
Tl;Dr - Its to ensure their overrides are preserved when reloading a config (e.g. executing focus twice)

scripts/Diffraction/isis_powder/abstract_inst.py
```diff
@@ -206,6 +206,7 @@ def generate_out_file_paths(self, run_details, output_directory=None):
         """
         if not output_directory:
             output_directory = os.path.join(self._output_dir, run_details.label, self._user_name)
+        output_directory.path.abspath(os.path.expanduser(output_directory))
```
DimitarTasev: Doesn't this throw `AttributeError: 'str' object has no attribute 'path'`?

dtasev (changes requested):

dtasev (approved):

DimitarTasev:
This PR covers the base work for the `isis_powder`, there are more fixes to come in #18666, and then additional work on merging more instruments. It looks good for now :shipit:

PR #18649, opened 2017-02-01
Support for old files that have masking parameters for non-detector components

### LINKED ISSUE #18647: Old nexus files with masks don't load
Old files generated by the code below are not readable by `LoadNexusProcessed`.

The error is:
```
Using instrument IDF XML text contained in nexus file.
Unable to find instrumentDefinition.vtpDirectory in the properties file
Error in execution of algorithm LoadNexusProcessed:
Found masking for a non-detector component. This is not possible
```

Code:
```python
SANSSensitivityCorrection(Filename=flood_field_file_raw,
    UseSampleDC=False,
    DarkCurrentFile=dark_curr_file_main,
    MinEfficiency=0.3,MaxEfficiency=1.7,
    MaskedFullComponent="detector1",
    OutputSensitivityWorkspace="sensitivitywing_1.0",
    ReductionProperties=ReductionSingleton().property_manager
)
SaveNexus(InputWorkspace="sensitivitywing_1.0", Filename=<NeXus File.nxs>)
```

Note this was done for HFIR CG3 (BioSANS) which has 2 detectors. `detector1` the main detector, the high angle detector `wing_detector`.

I'm copying the Slack conversation here:

> Simon Heybrock [11:56 AM] 
> @ricleal This fails due to a safety check I put in place. There is nothing major that would prevent us from loading such fils (we explicitly tested that old files can be loaded, as part of the PR review).
> 
> [11:57]  
> Basically the parameter name `"masked"` has (or rather had) a special meaning in Mantid: `Detector::isMasked()` would determine its return value based on that.
> 
> [11:58]  
> Masking is now stored in `DetectorInfo`, and to avoid any issues with having masking stored in two places I forbid accessing any parameter called `"masked"` in `ParameterMap`.
> 
> [11:58]  
> Apparently the file you are trying to load has a parameter called `"masked"` for a non-detector component.
> 
> [11:59]  
> When I refactored the way masking was done I did not come across a single case where masking of non-detector components was used.
> 
> [12:02]  
> I think there are three options to solve this:
> - Remove the check for insertion of `"masked"` in `ParameterMap`. I would like to avoid this since it compromises safety in all of Mantid.
> - Rename the parameter `"masked"` in `ExperimentInfo` (this is where the exception is originating) and insert it under a different name. The consequence would be that if you save the workspace again, the parameter would have a different name (unless you also rename it upon saving).
> - Drop the parameter `"masked"`: As far as I am aware there is no code reading that parameter, so maybe it was added by accident? This is something you would need to discuss with your users.

The NXS file is attached...

[wing467sens3.2deg.nxs.zip](https://github.com/mantidproject/mantid/files/742938/wing467sens3.2deg.nxs.zip)

**Edit:**

Freshly new generated `nxs` files work. I.e., without the `masked` tag, however, like my users pointed out, it is extremely important to keep back compatibility:

> backward compatibility is very important and we should push to have edits to the load function so there is backward compatibility to our old flood files.

SimonHeybrock:
@ricleal I noticed that the components are not *actually* masked> There is simple a parameter saying "this is not masked". I am now checking the mask flag value before throwing (and do not throw if masking is false).

I can load the file now. 

Questions:

- Is this sufficient?
- If not, what is the desired behavior? There is nothing in Mantid that reads masking of non-detector components.

**To test:**

Fixes #18647.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/src/ExperimentInfo.cpp
+++ b/Framework/API/src/ExperimentInfo.cpp
@@ -1331,17 +1331,21 @@ void ExperimentInfo::readParameterMap(const std::string &parameterStr) {
       paramValue += ";" + tokens[4];
 
     if (tokens[2].compare("masked") == 0) {
-      // Do not add masking to ParameterMap, it is stored in DetectorInfo
-      const auto det = dynamic_cast<const Detector *const>(comp);
-      if (!det)
-        throw std::runtime_error(
-            "Found masking for a non-detector component. This is not possible");
       const auto &type = tokens[1];
       const std::string name = "dummy";
       auto param = ParameterFactory::create(type, name);
       param->fromString(paramValue);
       bool value = param->value<bool>();
-      m_detectorInfo->setMasked(detectorInfo().indexOf(det->getID()), value);
+      if (value) {
+        // Do not add masking to ParameterMap, it is stored in DetectorInfo
+        const auto det = dynamic_cast<const Detector *const>(comp);
+        if (!det) {
+          throw std::runtime_error("Found masking for a non-detector "
+                                   "component. This is not possible");
+        } else
+          m_detectorInfo->setMasked(detectorInfo().indexOf(det->getID()),
+                                    value);
+      }
     } else {
       pmap.add(tokens[1], comp, tokens[2], paramValue);
     }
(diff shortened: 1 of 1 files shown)


ricleal:
Perfect! They just need it to be back compatible.
I tested and it loads the files with the respective components masked.
Thanks!
:shipit:

ricleal (approved):

PR #18653, opened 2017-02-01
Efixed mode for paalman pings corrections

### LINKED ISSUE #18652: Efixed mode for FlatPlatePaalmanPings and CylinderPaalmanPings
The above mentioned algorithms currently operate in these modes:

Direct: lambda_1 = lambda_fixed, lambda_2 = lambda_step
Indirect: lambda_1 = lambda_step, lambda_2 = lambda_fixed
Elastic: lambda_1 = lambda_2 = lambda_step

where lambda_1 is the incident wavelength, lambda_2 is the final wavelength, lambda_fixed is the wavelength corresponding to EFixed defined in the instrument, and lambda_step are the actual wavelength points along the x-axis of the input workspace.

Actually for `FlatPlatePaalmanPings` the properties exist, but they were not really in use.

New option is needed:

EFixed : lambda_1 = lambda_2 = lambda_fixed , in which case the workspace unit requirement can be waived.

See #18470 for discussion.

gvardany:
This adds an option to `FlatPlatePaalmanPings` and `CylinderPaalmanPings` as follows:

`Emode = Efixed`, in which case the workspace unit requirement can be waived, and the correction will be computed for a single wavelength corresponding to the given Efixed. Fixed value itself will be by default read from the instrument parameters, but can be overridden by the user input.

For `FlatPlatePaalmanPings`, the `Emode` property existed before, but was not actually in use.
So the previously existing 3 options (`Elastic, Direct, Indirect`) will be equivalent and they are left for legacy, and for consistency with cylinder algorithm. Only the `Efixed` new option will result in different behaviour.

**To test:**

1. Create a workspace with instrument defined, but with x unit other than wavelength.
2. Run the above mentioned algorithms with the option`Emode=Efixed` and specify some value for `Efixed`. For the chemical formula, use for example `H2-O`.  The rest of the options can be left to defaults. The algorithms should run and produce the correction workspace which should have one bin corresponding to the wavelength = sqrt(81.787 / Efixed) Angstrom. 
3. Use `SetInstrumentParameter` to add a parameter called `Efixed` of type `Number` with some value to the workspace created at step 1.
4. Repeat step 2. without giving the Efixed value.

Fixes #18652 .

Release notes to be added to 3.10.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


louisemccann:
@gvardany release notes can be updated

louisemccann:
The algorithm works as expected, but when attempting to use the Indirect>CalculatePaalmanPings there is an error: `Invalid value for property Input Workspace: the workspace must have units`

gvardany:
@louisemccann I will add the release notes. Regarding the GUI I was initially planning to do in a separate issue, but can do here as well. Btw, I noticed that the `NumberWavelengths` and `Interpolate` properties of the algorithms are also currently absent from the GUI. Should I add those as well?

gvardany:
Wrong merge, could not recover. See PR #18934 instead.

PR #18660, opened 2017-02-01
SpectrumInfo and DetectorInfo rollout: `LoadHelper.cpp` to `RotateInstrumentComponentTest.h`

### LINKED ISSUE #18569: SpectrumInfo and DetectorInfo rollout: `LoadHelper.cpp` to `RotateInstrumentComponentTest.h`
This issue is part of the SpectrumInfo and DetectorInfo rollout, see #17743.
Files covered in this issue are:

Framework/DataHandling/src/LoadHelper.cpp
Framework/DataHandling/src/LoadInstrumentFromNexus.cpp
Framework/DataHandling/src/SaveGSS.cpp
Framework/DataHandling/src/LoadVulcanCalFile.cpp
Framework/DataHandling/src/MoveInstrumentComponent.cpp
Framework/DataHandling/src/RotateInstrumentComponent.cpp
Framework/DataHandling/src/SaveAscii2.cpp
Framework/DataHandling/src/SaveFocusedXYE.cpp
Framework/DataHandling/test/LoadDetectorInfoTest.h
Framework/DataHandling/test/LoadEmptyInstrumentTest.h
Framework/DataHandling/test/GroupDetectors2Test.h
Framework/DataHandling/test/LoadIDFFromNexusTest.h
Framework/DataHandling/test/LoadISISNexusTest.h
Framework/DataHandling/test/LoadInstrumentFromRawTest.h
Framework/DataHandling/test/LoadInstrumentTest.h
Framework/DataHandling/test/RotateInstrumentComponentTest.h

### LINKED ISSUE #19000: Fix spectrum numbers saved to file with `SaveAscii2`
`SaveAscii2` does not seem to save spectrum numbers correctly.

### Expected behavior

The following output is expected after running the script below.

```
# X 	 Y 	 E
4
0.5	3	3
1.5	3	3
2.5	3	3
5
0.5	4	4
1.5	4	4
2.5	4	4
6
0.5	5	5
1.5	5	5
2.5	5	5
```

### Actual behavior

Output is:

```
# X 	 Y 	 E
1
0.5	3	3
1.5	3	3
2.5	3	3
1
0.5	4	4
1.5	4	4
2.5	4	4
1
0.5	5	5
1.5	5	5
2.5	5	5
```

Run this script:

```
x = [0, 1, 2, 3] * 64
y = [ j  for j in range(0, 64) for i in range(0, 3) ]
e = y
ws = CreateWorkspace(x, y, e, NSpec=64)
LoadInstrument(ws, RewriteSpectraMap=False, InstrumentName='MUSR')

SaveAscii(InputWorkspace=ws,
          Filename='/home/raquel/Desktop/test_ascii.txt',
          SpectrumList='3, 4, 5',
          WriteSpectrumID=True,
          Separator='Tab')
```

### Platforms affected

All

raquelalvarezbanos:
Description of work.

**To test:**

Fixes #18569 
Fixes #19000 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/DataHandling/test/LoadInstrumentTest.h
```diff
@@ -119,14 +106,15 @@ class LoadInstrumentTest : public CxxTest::TestSuite {
...
     TS_ASSERT_DELTA(samplepos->getPos().Z(), 0.0, 0.01);
 
-    boost::shared_ptr<const IDetector> ptrDet103 = i->getDetector(103);
-    TS_ASSERT_EQUALS(ptrDet103->getID(), 103);
-    TS_ASSERT_EQUALS(ptrDet103->getName(), "pixel");
-    TS_ASSERT_DELTA(ptrDet103->getPos().X(), 0.4013, 0.01);
-    TS_ASSERT_DELTA(ptrDet103->getPos().Z(), 2.4470, 0.01);
-    double d = ptrDet103->getPos().distance(samplepos->getPos());
+    const auto &detectorInfo = output->detectorInfo();
+    const auto &ptrDet103 = detectorInfo.detector(103);
```
raquelalvarezbanos: This makes the asserts below fail. For this to pass I have to use detector `102`.

Framework/DataHandling/test/LoadInstrumentTest.h
```diff
@@ -140,17 +128,18 @@ class LoadInstrumentTest : public CxxTest::TestSuite {
...
-    TS_ASSERT_EQUALS(det_from_ws->getID(), 602);
+    TS_ASSERT_EQUALS(det_from_ws.getID(), 602);
 
     // also a few tests on the last detector and a test for the one beyond the
     // last
-    boost::shared_ptr<const IDetector> ptrDetLast = i->getDetector(413256);
-    TS_ASSERT_EQUALS(ptrDetLast->getID(), 413256);
-    TS_ASSERT_EQUALS(ptrDetLast->getName(), "pixel");
-    TS_ASSERT_THROWS(i->getDetector(413257), Exception::NotFoundError);
+    const auto &ptrDetLast = detectorInfo.detector(413256);
```
raquelalvarezbanos: This causes a `vector subscript out of range`.

Framework/DataHandling/test/LoadInstrumentTest.h
```diff
@@ -140,17 +128,18 @@ class LoadInstrumentTest : public CxxTest::TestSuite {
...
+    TS_ASSERT_EQUALS(ptrDetLast.getID(), 413256);
+    TS_ASSERT_EQUALS(ptrDetLast.getName(), "pixel");
+    //TS_ASSERT_THROWS_ANYTHING(detectorInfo.detector(413257));
```
raquelalvarezbanos: Not sure how to do this with `DetectorInfo`

Framework/DataHandling/test/LoadInstrumentTest.h
```diff
@@ -226,17 +215,18 @@ class LoadInstrumentTest : public CxxTest::TestSuite {
     TS_ASSERT_EQUALS(samplepos->getName(), "nickel-holder");
     TS_ASSERT_DELTA(samplepos->getPos().Y(), 0.0, 0.01);
 
-    boost::shared_ptr<const IDetector> ptrDet = i->getDetector(101);
-    TS_ASSERT_EQUALS(ptrDet->getID(), 101);
+	const auto &detectorInfo = output->detectorInfo();
+    const auto &ptrDet = detectorInfo.detector(101);
+    TS_ASSERT_EQUALS(ptrDet.getID(), 101);
```
raquelalvarezbanos: The `ID` seems to be `259` now, and the assert fails.

Framework/DataHandling/test/LoadInstrumentTest.h
```diff
@@ -284,14 +274,13 @@ class LoadInstrumentTest : public CxxTest::TestSuite {
...
-    boost::shared_ptr<const Instrument> i = output->getInstrument();
-
-    boost::shared_ptr<const IDetector> ptrDet = i->getDetector(20201001);
-    TS_ASSERT_EQUALS(ptrDet->getName(), "det 1");
-    TS_ASSERT_EQUALS(ptrDet->getID(), 20201001);
-    TS_ASSERT_DELTA(ptrDet->getPos().X(), -0.0909, 0.0001);
-    TS_ASSERT_DELTA(ptrDet->getPos().Y(), 0.3983, 0.0001);
-    TS_ASSERT_DELTA(ptrDet->getPos().Z(), 4.8888, 0.0001);
+    const auto &detectorInfo = output->detectorInfo();
+    const auto &ptrDet = detectorInfo.detector(20201001);
```
raquelalvarezbanos: This now causes a `vector subscript out of range`

Framework/DataHandling/test/LoadInstrumentTest.h
```diff
@@ -389,12 +378,13 @@ class LoadInstrumentTest : public CxxTest::TestSuite {
...
 
     // Check the positions of the 6 detectors in the physical instrument
-    TS_ASSERT_EQUALS(physicalInst->getDetector(1000)->getPos(), V3D(0, 0, 0));
-    TS_ASSERT_EQUALS(physicalInst->getDetector(1001)->getPos(), V3D(0, 1, 0));
-    TS_ASSERT_EQUALS(physicalInst->getDetector(1002)->getPos(), V3D(1, 0, 0));
-    TS_ASSERT_EQUALS(physicalInst->getDetector(1003)->getPos(), V3D(1, 1, 0));
-    TS_ASSERT_EQUALS(physicalInst->getDetector(1004)->getPos(), V3D(2, 0, 0));
-    TS_ASSERT_EQUALS(physicalInst->getDetector(1005)->getPos(), V3D(2, 1, 0));
+    const auto &physicalDetectorInfo = ws->detectorInfo();
+    TS_ASSERT_EQUALS(physicalDetectorInfo.position(1000), V3D(0, 0, 0));
```
raquelalvarezbanos: This `physicalDetectorInfo.position(1000)` and the ones below also make the test crash.

Framework/DataHandling/test/LoadInstrumentTest.h
AntonPiccardoSelg: Can you try `   const auto &ptrDetLast = detectorInfo.detector(detectorInfo.indexOf(413256))`

raquelalvarezbanos: Yes, thanks Anton, Simon just told me that that is what I should be doing.

Framework/DataHandling/src/LoadHelper.cpp
```diff
@@ -105,22 +106,14 @@ double LoadHelper::calculateTOF(double distance, double wavelength) {
...
-  // Get a pointer to the instrument contained in the workspace
-  Geometry::Instrument_const_sptr instrument = workspace->getInstrument();
-  // Get the distance between the source and the sample (assume in metres)
-  Geometry::IComponent_const_sptr sample = instrument->getSample();
-  // Get the sample-detector distance for this detector (in metres)
-  double l2 =
-      workspace->getDetector(detId)->getPos().distance(sample->getPos());
-  return l2;
+  const auto &spectrumInfo = workspace->spectrumInfo();
+  return spectrumInfo.l2(detId);
```
SimonHeybrock: Now that these two functions are trivial, maybe they can be removed and instead the client code can use `SpectrumInfo` directly? I did not check myself, so I am not sure it is possible, but maybe you can have a look.

Framework/DataHandling/src/LoadVulcanCalFile.cpp
```diff
@@ -536,18 +537,17 @@ void LoadVulcanCalFile::convertOffsets() {
...
-    Geometry::IDetector_const_sptr det = m_tofOffsetsWS->getDetector(iws);
-    V3D detPos = det->getPos();
 
-    detid_t detid = det->getID();
+    detid_t detid = spectrumInfo.detector(iws).getID();
     int bankid = detid / static_cast<int>(NUMBERRESERVEDPERMODULE);
 
     double l2, twotheta, phi;
-    detPos.getSpherical(l2, twotheta, phi);
+	spectrumInfo.position(iws).getSpherical(l2, twotheta, phi);
```
SimonHeybrock: Have a look at the `//FIXME` comment above. I think if you use `SpectrumInfo::l2` and `SpectrumInfo::twoTheta` here you can remove that weird test there. Please verify yourself.

SimonHeybrock: You can then probably also remove the call to `getInstrumentParameters` and just use `SpectrumInfo::l1`.

Framework/DataHandling/src/SaveAscii2.cpp
```diff
@@ -443,11 +443,11 @@ void SaveAscii2::populateSpectrumNumberMetaData() {
 void SaveAscii2::populateAngleMetaData() {
   std::vector<std::string> angles;
   const size_t nHist = m_ws->getNumberHistograms();
+  const auto &spectrumInfo = m_ws->spectrumInfo();
   for (size_t i = 0; i < nHist; i++) {
     const auto specNo = m_ws->getSpectrum(i).getSpectrumNo();
     const auto workspaceIndex = m_specToIndexMap[specNo];
```
SimonHeybrock: Am I missing something here? Is there something special in `m_specToIndexMap` or are this two lines translating workspace-index -> spectrum-number -> workspace-index, i.e., do nothing?

Framework/DataHandling/test/LoadInstrumentTest.h
SimonHeybrock: ID or index?

Framework/DataHandling/test/LoadInstrumentTest.h
```diff
@@ -140,17 +128,18 @@ class LoadInstrumentTest : public CxxTest::TestSuite {
     TS_ASSERT_EQUALS(output->getAxis(1)->spectraNo(257), 258);
 
     auto ids_from_map = output->getSpectrum(257).getDetectorIDs();
-    IDetector_const_sptr det_from_ws = output->getDetector(257);
+	const auto &spectrumInfo = output->spectrumInfo();
+    const auto &det_from_ws = spectrumInfo.detector(257);
```
SimonHeybrock: ID or index?

SimonHeybrock (commented):
A few comments below:

Framework/DataHandling/src/LoadVulcanCalFile.cpp
raquelalvarezbanos: Yes, I think it can be done as you say.

Framework/DataHandling/test/LoadInstrumentTest.h
raquelalvarezbanos: Is this incorrect (the same for the code above)? I thought `output->getDetector(i);` should be replaced with `const auto &det = spectrumInfo.detector(i);`

Framework/DataHandling/src/SaveAscii2.cpp
raquelalvarezbanos: I don't know, I think we are just doing what you say but I am getting confused. I've made some changes, let me know if you think they are OK.

Framework/DataHandling/src/SaveAscii2.cpp
```diff
@@ -275,63 +275,14 @@ void SaveAscii2::exec() {
...
 }
 
-/**writes a spectra to the file using an iterator
-@param spectraItr :: a set<int> iterator pointing to a set of workspace IDs to
-be saved
+/** Writes a spectrum to the file using a workspace index
+@param wsIndex :: an integer relating to a workspace index
 @param file :: the file writer object
 */
-void SaveAscii2::writeSpectra(const std::set<int>::const_iterator &spectraItr,
```
SimonHeybrock: Where there simply two identical versions of the same method, one for an index, one for an iterator to an index?

raquelalvarezbanos: I think so... Code was not exactly the same, one was:

```
if (A & B)
   // Do something (I)
else if (!B)
   // Do something (II)
else
   // Do something (III)
```

and the other was:

```
if (!B)
  // Do something (II)
else if (A & B)
  // Do something (I)
else
  // Do something (III)
```

but both should do the same.

Framework/DataHandling/test/GroupDetectors2Test.h
```diff
@@ -132,11 +133,9 @@ class GroupDetectors2Test : public CxxTest::TestSuite {
       TS_ASSERT_DELTA(outputWS->e(0)[i], std::sqrt(double(2)), 0.0001);
     }
 
-    boost::shared_ptr<const IDetector> det;
-    TS_ASSERT_THROWS_NOTHING(det = outputWS->getDetector(0));
```
SimonHeybrock: Could test for `spectrumInfo.hasDetectors(0)`.

Framework/DataHandling/test/GroupDetectors2Test.h
```diff
@@ -132,11 +133,9 @@ class GroupDetectors2Test : public CxxTest::TestSuite {
...
-    TS_ASSERT(boost::dynamic_pointer_cast<const DetectorGroup>(det));
```
SimonHeybrock: Could test for `!spectrumInfo.hasUniqueDetector(0)`.

Framework/DataHandling/test/GroupDetectors2Test.h
```diff
@@ -132,11 +133,9 @@ class GroupDetectors2Test : public CxxTest::TestSuite {
...
-    TS_ASSERT_THROWS_ANYTHING(det = outputWS->getDetector(1));
```
SimonHeybrock: Test for `!spectrumInfo.hasDetectors(1)`?

Framework/DataHandling/test/GroupDetectors2Test.h
```diff
@@ -167,11 +166,9 @@ class GroupDetectors2Test : public CxxTest::TestSuite {
       TS_ASSERT_DELTA(outputWS->e(0)[i], std::sqrt(4.0), 0.0001);
     }
 
-    boost::shared_ptr<const IDetector> det;
-    TS_ASSERT_THROWS_NOTHING(det = outputWS->getDetector(0));
-    TS_ASSERT(boost::dynamic_pointer_cast<const DetectorGroup>(det));
-    TS_ASSERT_THROWS_ANYTHING(det = outputWS->getDetector(1));
```
SimonHeybrock: Same as above.

Framework/DataHandling/test/GroupDetectors2Test.h
```diff
@@ -202,10 +199,9 @@ class GroupDetectors2Test : public CxxTest::TestSuite {
       TS_ASSERT_DELTA(outputWS->e(0)[i], std::sqrt(double(NHIST)), 0.0001);
     }
 
-    boost::shared_ptr<const IDetector> det;
-    TS_ASSERT_THROWS_NOTHING(det = outputWS->getDetector(0));
-    TS_ASSERT(boost::dynamic_pointer_cast<const DetectorGroup>(det));
-    TS_ASSERT_THROWS_ANYTHING(det = outputWS->getDetector(1));
```
SimonHeybrock: See above

Framework/DataHandling/test/GroupDetectors2Test.h
```diff
@@ -266,20 +262,19 @@ class GroupDetectors2Test : public CxxTest::TestSuite {
...
-    TS_ASSERT_THROWS_NOTHING(det = outputWS->getDetector(1));
-    TS_ASSERT(boost::dynamic_pointer_cast<const Detector>(det));
-    TS_ASSERT_THROWS_NOTHING(det = outputWS->getDetector(2));
-    TS_ASSERT(boost::dynamic_pointer_cast<const Detector>(det));
-    TS_ASSERT_THROWS_NOTHING(det = outputWS->getDetector(3));
-    TS_ASSERT(boost::dynamic_pointer_cast<const Detector>(det));
-    TS_ASSERT_THROWS_NOTHING(det = outputWS->getDetector(4));
-    TS_ASSERT(boost::dynamic_pointer_cast<const Detector>(det));
+    const auto &spectrumInfo = outputWS->spectrumInfo();
+    TS_ASSERT_THROWS_NOTHING(spectrumInfo.detector(0));
```
SimonHeybrock: `hasDetectors`?

Framework/DataHandling/test/LoadInstrumentTest.h
```diff
@@ -380,55 +379,47 @@ class LoadInstrumentTest : public CxxTest::TestSuite {
...
-    TS_ASSERT_EQUALS(
-        physicalDetectorInfo.position(physicalDetectorInfo.indexOf(1003)),
-        V3D(1, 1, 0));
-    TS_ASSERT_EQUALS(
-        physicalDetectorInfo.position(physicalDetectorInfo.indexOf(1004)),
-        V3D(2, 0, 0));
-    TS_ASSERT_EQUALS(
-        physicalDetectorInfo.position(physicalDetectorInfo.indexOf(1005)),
-        V3D(2, 1, 0));
+    TS_ASSERT_EQUALS(physicalInst->getDetector(1000)->getPos(), V3D(0, 0, 0));
```
raquelalvarezbanos: I am not sure what to do with occurrences of `physicalInst->getDetector()`, this `physicalInst` is not attached to any workspace (I don't even know what it is and the difference with the "neutronic" instrument). Any ideas?

SimonHeybrock: @raquelalvarezbanos Basically this is a workaround that the current instrument code does to support analyzers or mirrors in the secondary flight path, since there is no direct support for this. The physical instrument has detector positions as we see them, the neutronic instrument has them moved around such that map the neutrons path to a straight line from the sample.

I think you should just leave this as it is. Instrument-.20 will support non-straight beam paths (eventually), we can clean this up then.

SimonHeybrock:
Thanks, latest changes look good :+1:, let us see if all builds pass.

Framework/DataHandling/test/LoadInstrumentTest.h
raquelalvarezbanos: 👍 Thanks for the explanation

raquelalvarezbanos:
@SimonHeybrock: could you double-check changes to `SaveAscii2`? More specifically, could you check if you agree with [this](https://github.com/mantidproject/mantid/pull/18660/files#diff-a37f7c701c12ca62076f8a56bc0a5432R578)? This assert corresponds to a test I added (`test_valid_SpectrumList`) to `SaveAscii2Test` and it fails before my changes, that is, if you added that test to master it would fail:

```
In SaveAscii2Test::test_valid_SpectrumList:
D:/mantid/mantid/Framework/DataHandling/test/SaveAscii2Test.h:575: Error: Expected (specID == 2), found (1 != 2)
Failed 1 of 1 test
Success rate: 0%
```

but I think the `ID` should be 2. With my changes it passes.

SimonHeybrock:
@raquelalvarezbanos There might be a bad naming issue here. There are input properties referring to "workspace index", which is used to build a set of indices here https://github.com/mantidproject/mantid/pull/18660/files#diff-60bfd869a2365c5ad3eee2efab6cd2b8R204. However, I think these "indices" are then translated from "spectrum numer" to "workspace index", for example here: https://github.com/mantidproject/mantid/pull/18660/files#diff-60bfd869a2365c5ad3eee2efab6cd2b8L289?

SimonHeybrock:
If that is the case, I think the range validations in the original code are also wrong.

raquelalvarezbanos:
I am starting to believe that there was a bug in `SaveAscii2` which I accidentally fixed when working on this.

I am trying this script:

```
x = [0, 1, 2, 3] * 64
y = [ j  for j in range(0, 64) for i in range(0, 3) ]
e = y
ws = CreateWorkspace(x, y, e, NSpec=64)
LoadInstrument(ws, RewriteSpectraMap=False, InstrumentName='MUSR')

SaveAscii(InputWorkspace=ws,
          Filename='/home/raquel/Desktop/test_ascii.txt',
          SpectrumList='3, 4, 5',
          WriteSpectrumID=True,
          Separator='Tab')
```

The workspace `ws` is a workspace with 64 spectra and instrument `MUSR`. When running the script above on master, the output file is:

```
# X 	 Y 	 E
1
0.5	3	3
1.5	3	3
2.5	3	3
1
0.5	4	4
1.5	4	4
2.5	4	4
1
0.5	5	5
1.5	5	5
2.5	5	5
```
The `X`, `Y` and `E` values are correct, but the first number before `X Y E` values, which is the `SpectrumID`, which I assume means `SpectrumNumber`, must be wrong because it is `1` for all the saved spectra. However, on this branch I get:

```
# X 	 Y 	 E
4
0.5	3	3
1.5	3	3
2.5	3	3
5
0.5	4	4
1.5	4	4
2.5	4	4
6
0.5	5	5
1.5	5	5
2.5	5	5
```
which I think is correct, because I am running `SaveAscii` with `SpectrumList=3, 4, 5` and, given that property `SpectrumList` refers to workspace indices, the spectrum numbers should be `3`, `4`, `5`.

What do you think?

SimonHeybrock:
Yes, the second version looks correct to me. I had a look at the old code and tried to see where things are going wrong, but I did not spot anything...?

raquelalvarezbanos:
I don't know, the old code looks correct to me too... I'll have to debug it next week to see what it is doing in more detail.

raquelalvarezbanos:
The algorithm seems to be working fine if you specify `SpectrumMetaData`, like this:

```
SaveAscii(InputWorkspace=ws,
          Filename='C:/Users/dsl37522/Desktop/test_ascii.txt',
          SpectrumList='3, 4, 5',
          WriteSpectrumID=True,
          Separator='Tab',
          SpectrumMetaData='SpectrumNumber')
```

In this case the output is the expected

```
# X 	 Y 	 E
4
0.5	3	3
1.5	3	3
2.5	3	3
5
0.5	4	4
1.5	4	4
2.5	4	4
6
0.5	5	5
1.5	5	5
2.5	5	5
```

What happens if you don't specify this `SpectrumMetaData` property (which is what I was doing above) is that the variable `m_specToIndexMap` is never initialized ([this](https://github.com/mantidproject/mantid/blob/master/Framework/DataHandling/src/SaveAscii2.cpp#L135) line is never executed) but `m_specToIndexMap` is later used when writing the spectra (see [here](https://github.com/mantidproject/mantid/blob/master/Framework/DataHandling/src/SaveAscii2.cpp#L286-L297)), therefore when we do:

```
const auto workspaceIndex = m_specToIndexMap[specNo];
```

in `SaveAscii2::writeSpectra()` because key `specNo` does not exist, `workspaceIndex` will be `0`, and this will happen for all the spectra we are writing. I think this is a bug in the code.

SimonHeybrock:
@raquelalvarezbanos Yes looks like a bug! Your changes fix it, right, so I can put the ship?

raquelalvarezbanos:
Yes, I think changes here fix it.

SimonHeybrock:
Code changes look good :shipit: !

@raquelalvarezbanos For documentation purposes, could you open an issue for this bug and mark it as fixed by this PR?

raquelalvarezbanos:
Done!

PR #18667, opened 2017-02-01
Add configuration dependent instrument path

peterfpeterson:
There was a bug in mantid related to loading `Facilities.xml` from the instrument download directory when instrument download was turned off. 

To reproduce (these are linux instructions):

1. Copy the [`Facilities.xml` from v3.8.0](https://github.com/mantidproject/mantid/blob/v3.8.0/instrument/Facilities.xml) into you instrument download directory `$HOME/.mantid/instrument/Facilities.xml`
2. Start `mantidpython`, `import mantid` then `print config.getInstrument('NOMAD').liveListener()` which will (incorrectly) be an empty string
3. Merge this branch
4. Do step 2 again, but see that the correct value of `SNSLiveEventDataListener` comes through

*Does not need to be in the release notes* because it is a bugfix to something that got in towards the end of the release cycle that wasn't seen by users.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Kernel/inc/MantidKernel/ConfigService.h
+++ b/Framework/Kernel/inc/MantidKernel/ConfigService.h
@@ -297,6 +297,8 @@ private:
   /// Empty the list of facilities, deleting the FacilityInfo objects in the
   /// process
   void clearFacilities();
+  /// Determine the name of the facilities file to use
+  std::string getFacilityFilename(const std::string &fName);
   /// Verifies the directory exists and add it to the back of the directory list
   /// if valid
   bool addDirectoryifExists(const std::string &directoryName,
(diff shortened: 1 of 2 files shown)


NickDraper:
So other than neatening things up a bit the major change here is to only consider the download directory if mantid is configured to perform the downloads.

This is an edge case as it will only affect installations where:
1. it has previously downloaded files
1. it is no longer configured to download them anymore

However while it is an edge case it might be an important one in some cases.

NickDraper:
Tested on Windows, all is fine, tested by manipulating the value of UpdateInstrumentDefinitions.OnStartup
with an old file in the download directory.  Should be safe in all situations. :shipit:

PR #18668, opened 2017-02-01
slice viewer thickness

### LINKED ISSUE #17956: The BinMD range is set to twice the thickness in SliceViewer
John:
 >in the slice viewer taking a cut from a MD workspace if i select a ‘thickness’ of say .5 when i look at the BinMD output  from the workspace history the dimension is reported as -.5 , .5 which seems to suggest the thickness is actually 1 rather than .5. Is this the expected behaviour? (edited)

Duc:
>  This is also the behaviour of SliceViewer for Histogram workspaces - the "width in plane" is actually the half-width. It's a bit counter-intuitive, but I guess I'm used to it now...

John:
> i didn’t check that but thanks for mentioning it, counter intuitive is probably not the correct phrase :slightly_smiling_face: at least with 2D workspaces the rubberband box gives some indication but with 3 or 4d data that isnt the case it could lead to issues, i.e. slicing a 3d volume and double counting. Which is what happened to us today.

The range that is passed to BinMD is min= slicePoint - thickness and max= slicePoint + thickness. It looks as if this is done by design. Since it has been mentioned that it is counterintuitive, we should change it to +/-thickness/2.

AndreiSavici:
The meaning of SliceViewer thickness is now `center+-(thickness/2)` instead of `center+-thickness`

**To test:**
Load an MDE file, show it in slice viewer. Rebin it using the GUI, and check if the thickness of the MDHisto is what you expect. Do the similar for the lineViewer

Fixes #17956 .

**Note:** I will open a new ticket to fix all python problems. The constructor works as expected, but setting  and getting the thickness in plane is still the old behaviour. Wiki pages on sliceviewer will need to be changed

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidPlot/pymantidplot/proxies.py
+++ b/MantidPlot/pymantidplot/proxies.py
@@ -734,11 +734,12 @@ class SliceViewerWindowProxy(QtProxyObject):
         # Set the width.
         if not width is None:
             liner.setThickness(width)
+            liner.setPlanarWidth(width*0.5)
         else:
-            liner.setPlanarWidth(planar_width)
+            liner.setPlanarWidth(planar_width*0.5)
             if not thicknesses is None:
                 for d in xrange(len(thicknesses)):
-                    liner.setThickness(d, thicknesses[i])
+                    liner.setThickness(d, thicknesses[d])
         # Bins
         liner.setNumBins(num_bins)
         liner.apply()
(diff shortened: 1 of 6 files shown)


peterfpeterson:
LGTM :shipit:

PR #18669, opened 2017-02-01
Add Eigen as an external project

### LINKED ISSUE #18540: Add Eigen as an External Project
Add Eigen as an external project to Mantid, for use in Instrument 2.0. Using an external project will ensure we get the same version on all platforms and no extra dependencies will be required.

~~Wrapping `Matrix3d` and `Vector3d` with `boost:python` should also be investigated as part of this ticket.~~

ianbush:
Add Eigen as an external dependency.

**To test:**

Checkout the branch and re-run `cmake` however you normally would. Check that there are no errors.

Try running `cmake` in a clean directory and check there are no errors.

You should also checkout out commit 9de61cf and try building the `Beamline` target. You can see the build history to see that this test passed on the build servers (`Eigen` will soon be used in `Beamline`).

Check no extra targets are created, e.g. `ninja -t targets` or check the list in VS.

Fixes #18540.

*Internal change, does not need to be in the release notes - but developer documentation should be updated once this is merged.*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


LamarMoore:
Works fine for me on Windows

Framework/Beamline/inc/MantidBeamline/EigenTest.h
```diff
@@ -0,0 +1,13 @@
+#ifndef MANTID_EIGENTEST_H
```
SimonHeybrock: Can we remove the header and simply put everything in the test? That would avoid polluting the library and no one will mind keeping the test around.

buildconfig/CMake/Eigen.cmake
```diff
@@ -0,0 +1,25 @@
+include(ExternalProject)
+
+# Use version 3.2.10 of Eigen
+set(eigen_version "3.2.10")
```
SimonHeybrock: Owen suggest to add a comment here why you went back to the old version (compiler warnings in the latest builds).

SimonHeybrock:
Tested on Ubuntu, cmake in clean dir works without errors, and so does rerunning cmake.

Framework/Beamline/inc/MantidBeamline/EigenTest.h
ianbush: I was planning to remove `EigenTest.h` and `EigenTest.cpp` completely before removing 'In Progress'. It was just to test the build servers can actually build something using `Eigen`.

SimonHeybrock: Also fine :+1:

SimonHeybrock:
Owen also tested this on OSX, works -> :shipit:

Issue #18580, opened 2017-01-26
Update reflectometry release notes


raquelalvarezbanos:
Scientists pointed out a few things that were not clear in the release notes.

Issue #18489, opened 2017-01-20
`LoadEventNexus` can't load some ISIS datasets in debug mode


raquelalvarezbanos:
Some event datasets at ISIS have `event_id = 0` (this could mean that the encoding signal was corrupted somehow or that there's a detector which wasn't assigned a spectrum to it) which causes Mantid to crash at [this](https://github.com/mantidproject/mantid/blob/master/Framework/DataHandling/src/ProcessBankData.cpp#L59) line. This issue is to fix this for the next release. The reason why this is not urgent is that when loading the problematic files in release mode, those events are just skipped, so that the data we end up with is correct.

### Expected behavior

We should be able to load file:

```
ws = Load(Filename=r'\\isis\inst$\ndxinter\instrument\data\cycle_16_2\INTER00038415.nxs')
```
in debug mode.

### Actual behavior

Mantid crashes.

### Steps to reproduce the behavior

Run the line above.

### Platforms affected

Windows, debug mode.

gemmaguest:
Closed as a duplicate of #19103

PR #18690, opened 2017-02-02
Fix some issues in the Reflectometry GUI

### LINKED ISSUE #18677: Fix reflectometry issues found during beta testing
Scientists have reported the following:

1. Combined transmission is stitched the wrong way round 42115, 42114 instead of 42114, 42115 - or something else that works differently from the old GUI.
2. The interface does not output combined transmission run in lambda.
3. In the script sent by Max that reduces event data, there were some hard-coded values when rebinning the event workspace. We should use `RebinToWorkspace` and `Rebin` to monitors workspace instead.
4. There seems to be a problem with the normalization of slices, there is a shift in Y.

raquelalvarezbanos:
Fixes some issues reported by scientists during beta testing.

**To test:**

- Open the Reflectometry (polref) interface and load `INTER_NR_test2.tbl`, make sure the selected instrument is `INTER`. Go to `Settings` and give values for `StartOverlap` and `EndOverlap`, for instance 10 and 12. Select the first row and hit `Process`. Check the history of `TRANS_13463_13464` and check that the algorithm was run with values 10 and 12 for `StartOverlap` and `EndOverlap` respectively.
- Check that the above also works if you specify the transmission runs from the `Settings` tab. For this, remove the transmission runs from the table and enter `13463,13464` in `Settings`.
- Review code changes. Check that `Rebin` algorithms in `ReflDataProcessorPresenter` were replaced by a `RebinToWorkspace` which rebins the detectors to monitors.
- In a new table, enter run `38415`, angle `0.5` and transmission run `38393`. In `Event Handling` enter `0, 200, 400`. Hit process. Plot workspaces `IvsLam_38415_0_200` and `IvsLam_38415_200_400`, curves should be similar (i.e. there should be no shift in Y, you may want to try this on a nightly build to see the difference). The other workspaces (`IvsQ_` and `IvsQ_binned`) should be similar too.

Fixes #18677.

Release notes will be updated as part of #18580

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/IReflSettingsView.h
+++ b/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/IReflSettingsView.h
@@ -2,8 +2,8 @@
 #define MANTID_CUSTOMINTERFACES_IREFLSETTINGSVIEW_H
 
 #include "MantidQtCustomInterfaces/DllConfig.h"
-#include <vector>
 #include <map>
+#include <vector>
 
 namespace MantidQt {
 
@@ -56,6 +56,8 @@ public:
   virtual std::string getAnalysisMode() const = 0;
   virtual std::string getDirectBeam() const = 0;
   virtual std::string getTransmissionRuns() const = 0;
+  virtual std::string getStartOverlap() const = 0;
+  virtual std::string getEndOverlap() const = 0;
   virtual std::string getPolarisationCorrections() const = 0;
   virtual std::string getCRho() const = 0;
   virtual std::string getCAlpha() const = 0;
--- a/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/QtReflSettingsView.h
+++ b/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/QtReflSettingsView.h
@@ -52,6 +52,10 @@ public:
   std::string getDirectBeam() const override;
   /// Return transmission runs
   std::string getTransmissionRuns() const override;
+  /// Return start overlap for transmission runs
+  std::string getStartOverlap() const override;
+  /// Return end overlap for transmission runs
+  std::string getEndOverlap() const override;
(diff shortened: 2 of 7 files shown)


PranavBahuguna:
After review of this PR, I have found the following:

- The code looks fine, no issues there.
- I entered values 10 and 12 for start/end overlaps and processed them with the first row of `INTER_NR_test2.tbl` - the history of the output ws `TRANS_13463_13464` does indeed show that these values were used for those properties.
- When transmission runs were removed from table and entered from `Settings` tab, the last three TRANS workspaces (`TRANS_13463, TRANS_13463_13464, TRANS_13464`) were replaced with two group workspaces (TRANS_13463, TRANS_13464). The stitched workspace `TRANS_13463_13464` is nowhere to be seen. This definitely needs addressing as the output workspaces are supposed to be the same.
- I attempted to process a run with the given parameters. While it processes the run alone fine, when processed with Event Handling parameters it throws a critical error.

raquelalvarezbanos:
Thanks for testing this @PranavBahuguna . Regarding the issues you found:

1. `TRANS_13463_13464` not being created when transmission run is specified from `Settings`: Max noticed this and he agreed that it is OK to fix it for the next release. See https://github.com/mantidproject/mantid/issues/18677

2. Critical error processing event runs: Were you testing this in debug mode? I think this could be due to https://github.com/mantidproject/mantid/issues/18489. Could you try in release mode? It should work there. The alternative if you test in debug mode is to add this fix: https://github.com/mantidproject/mantid/commit/583f8c04f69af48abe21b99049ba312f17984b36

PranavBahuguna:
Alright, I've gone over the two issues now and have found the following:

- `CreateTransmissionWorkspace` in `IvsLam` shows the values of Start/End Overlaps being used as expected.
- Processing event runs works under Release mode. The plotted workspaces for `IvsLam_38415_0_200` and `IvsLam_38415_200_400` (and their IvsQ, IvsQ_binned and TOF equivalents) have similar curves to each other.

As all tests are successful, I see no further issues with this :shipit:

PR #18701, opened 2017-02-03
Fix bug in transmission multiperiod reduction ISIS SANS

### LINKED ISSUE #18700: Bug when reducing multiperiod transmission data
Diego found a bug when trying to reduce in the ISIS SANS GUI with multiperiod transmission data. Basically it does not work, because the method is `move2next` has never been implemented. This bug appears to have been there for 4 years or so. The output is:
multiperiod

Fixing this is easy. The correct method is `move2ws`. Nevertheless th reduction falls over with the data set that diego has provided, because the transmission monitor did not collect any data, hence it is not really the code's fault that this does not work. Nevertheless it might be usefull to provide a better warning message in order to let the user know that potentially the transmission monitor is faulty.

single period

AntonPiccardoSelg:
Fixes #18700 

This PR replaces a non-existent method name with the correct one. It also provides a better exception message.

**To test:**

Please find the sample data here: \\olympic\Babylon5\Scratch\Anton\LARMOR_MULTI_PERIOD_REDUCTION
Make sure that it you make this available to your Mantid path.

A quick note. The reduction will fail with the data set which is provided here. The user does not have any other data set. We test here for two things:
  * We don't get an error saying: `LoadRun object has no attribute `move2next`
  * We get an error which makes it clear that the transmission calculation has failed and mentions that missing signal might be the cause.

1. Open the ISIS SANS GUI
2. Select LARMOR as the instrument
3. Load the user file: USER_Bender_RB1610214_162DMagnet_parallel_r10182.txt
4. For the data insert

|           |  Scatter     |  Transmission   | Direct   |
|--------|---------------|-------------------|----------|
| SANS  | LARMOR00013538| LARMOR00013537| LARMOR00013510|
| CAN    |  LARMOR00013509| LARMOR00013510|LARMOR00013510

5. Presss `Load Data`
6. Press `1D Reduce`
    * Confirm that the reduction fails. Confirm that the error message mentions a failure in the transmission calculation and explains that maybe missing signal might be the cause of it.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/scripts/SANS/ISISCommandInterface.py
+++ b/scripts/SANS/ISISCommandInterface.py
@@ -644,7 +644,7 @@ def _WavRangeReduction(name_suffix=None):
         for trans in [ReductionSingleton().samp_trans_load, ReductionSingleton().can_trans_load]:
             if trans and trans.direct.periods_in_file > 1 and trans.trans.periods_in_file > 1:
                 trans.direct.move2ws(period)
-                trans.trans.move2next(period)
+                trans.trans.move2ws(period)
         return
 
     def _applySuffix(result, name_suffix):
--- a/scripts/SANS/isis_reduction_steps.py
+++ b/scripts/SANS/isis_reduction_steps.py
@@ -2388,8 +2388,14 @@ class TransmissionCalc(ReductionStep):
         calc_trans_alg.execute()
 
         # Set the y axis label correctly for the transmission ratio data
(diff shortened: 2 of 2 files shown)


martyngigg:
Simple improvement to give users a better hint at what went wrong.

LGTM :shipit:

PR #18720, opened 2017-02-06
Correct system package name for eigen3

peterfpeterson:
The name of the package is actually `eigen3`

**To test:**

Ideally, you would try building with the system packages. In reality you could just look at another project's [cmake requirements](*https://github.com/SpinWaveGenie/SpinWaveGenie/blob/master/CMakeLists.txt#L68)

*Does not need to be in the release notes* because it is a change for the developers

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/buildconfig/CMake/Eigen.cmake
+++ b/buildconfig/CMake/Eigen.cmake
@@ -8,7 +8,8 @@ option(USE_SYSTEM_EIGEN "Use the system installed Eigen - v${eigen_version}?" OF
 
 if(USE_SYSTEM_EIGEN)
   message(STATUS "Using system Eigen")
-  find_package(Eigen3 ${eigen_version} EXACT REQUIRED)
+  find_package(Eigen3 ${eigen_version} EXACT REQUIRED
+               PATH_SUFFIXES eigen3 eigen )
 else()
   message(STATUS "Using Eigen in ExternalProject")
 
--- a/buildconfig/CMake/Eigen.cmake
+++ b/buildconfig/CMake/Eigen.cmake
@@ -8,7 +8,7 @@ option(USE_SYSTEM_EIGEN "Use the system installed Eigen - v${eigen_version}?" OF
 
 if(USE_SYSTEM_EIGEN)
   message(STATUS "Using system Eigen")
-  find_package(Eigen ${eigen_version} EXACT REQUIRED)
+  find_package(Eigen3 ${eigen_version} EXACT REQUIRED)
 else()
   message(STATUS "Using Eigen in ExternalProject")
 


ianbush:
Tested this on Ubuntu 14.04 & 16.04, where it does not seem to be enough to determine the correct version of `eigen3`. However it does not make anything worse, and should help on some platforms, so :shipit:.

buildconfig/CMake/Eigen.cmake
```diff
@@ -8,7 +8,8 @@ option(USE_SYSTEM_EIGEN "Use the system installed Eigen - v${eigen_version}?" OF
 
 if(USE_SYSTEM_EIGEN)
   message(STATUS "Using system Eigen")
-  find_package(Eigen ${eigen_version} EXACT REQUIRED)
+  find_package(Eigen3 ${eigen_version} EXACT REQUIRED
```
martyngigg: I think the problem on Ubuntu is that we'll never be able to match the version exactly. I have `3.2.92` on `16.04`.

Also it doesn't actually seem to pick up the version of the system one correctly anyway:

```
The following configuration files were considered but not accepted:

    /usr/lib/cmake/eigen3/Eigen3Config.cmake, version: unknown
```

I think that's a problem with the `Eigen3` cmake config that doesn't set a `EIGEN3_VERSION` variable that cmake expects.

Shall we just drop the version requirement?

peterfpeterson: To make this useful to redhat7, we'd have to change it to a minimum version starting at 3.2.5. Dropping the version requirement might not be a big deal as build servers should pick up errors in our usage and most systems will use the external projects.

ianbush: I was discussing this with @peterfpeterson, and were thinking to leave it as it is as the download is small and it will ensure the correct version, as opposed to the developer making sure it is compatible. I think either approach should be fine, and in either case I expect it would be rarely used.

martyngigg: :+1: - I'll merge it

ianbush: I hadn't seen the reply from @peterfpeterson when I commented, should we drop it then?

martyngigg: I have no strong preference either way. As you say, I think it is unlikely to be used in this way anyway.

ianbush: Then let us go for it as it is :+1: .

rosswhitfield:
I can't build the docs when the build directory is in the source directory, I expect the nightly builds to starting failing for the same reason.
```
Exception occurred:
  File "/home/rwp/mantid/docs/sphinxext/mantiddoc/doctest.py", line 520, in doctest_to_xunit
    raise exception
SourceLinkError: Found multiple possibilities for Transpose.h
Possible matches[u'/home/rwp/mantid/Release/clang/eigen-src/Eigen/src/Core/Transpose.h', u'/home/rwp/mantid/Release/py3/eigen-src/Eigen/src/Core/Transpose.h', u'/home/rwp/mantid/Framework/Algorithms/inc/MantidAlgorithms/Transpose.h']
Specify one using the h option
e.g. 
.. sourcelink:
      :h: /Release/clang/eigen-src/Eigen/src/Core/Transpose.h
```

PR #18725, opened 2017-02-07
Fix indexing errors in NumPy 1.12

rosswhitfield:
In 1.12 numpy changed some DeprecationWarnings to errors related to indexing, see https://docs.scipy.org/doc/numpy/release.html#deprecationwarning-to-error

This only fixes the errors where unit tests were failing, mostly using floats to index arrays, see http://builds.mantidproject.org/job/master_clean-archlinux/748/console

**To test:**
Code review should be sufficient, I don't expect anyone else to be using NumPy 1.12 currently

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/PythonInterface/plugins/algorithms/MatchPeaks.py
+++ b/Framework/PythonInterface/plugins/algorithms/MatchPeaks.py
@@ -19,18 +19,18 @@ def mask_ws(ws_to_mask, xstart, xend):
 
     if xstart > 0:
         logger.debug('Mask bins smaller than {0}'.format(xstart))
-        MaskBins(InputWorkspace=ws_to_mask, OutputWorkspace=ws_to_mask, XMin=x_values[0], XMax=x_values[xstart])
+        MaskBins(InputWorkspace=ws_to_mask, OutputWorkspace=ws_to_mask, XMin=x_values[0], XMax=x_values[int(xstart)])
     else:
         logger.debug('No masking due to x bin <= 0!: {0}'.format(xstart))
     if xend < len(x_values) - 1:
         logger.debug('Mask bins larger than {0}'.format(xend))
-        MaskBins(InputWorkspace=ws_to_mask, OutputWorkspace=ws_to_mask, XMin=x_values[xend + 1], XMax=x_values[-1])
+        MaskBins(InputWorkspace=ws_to_mask, OutputWorkspace=ws_to_mask, XMin=x_values[int(xend) + 1], XMax=x_values[-1])
     else:
         logger.debug('No masking due to x bin >= len(x_values) - 1!: {0}'.format(xend))
 
     if xstart > 0 and xend < len(x_values) - 1:
         logger.information('Bins out of range {0} {1} [Unit of X-axis] are masked'.format
-                           (x_values[xstart],x_values[xend + 1]))
+                           (x_values[int(xstart)],x_values[int(xend) + 1]))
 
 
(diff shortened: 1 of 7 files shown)


Framework/PythonInterface/plugins/algorithms/MatchPeaks.py
```diff
@@ -19,18 +19,18 @@ def mask_ws(ws_to_mask, xstart, xend):
 
     if xstart > 0:
         logger.debug('Mask bins smaller than {0}'.format(xstart))
-        MaskBins(InputWorkspace=ws_to_mask, OutputWorkspace=ws_to_mask, XMin=x_values[0], XMax=x_values[xstart])
+        MaskBins(InputWorkspace=ws_to_mask, OutputWorkspace=ws_to_mask, XMin=x_values[0], XMax=x_values[int(xstart)])
```
AndreiSavici: You could instead replace the `xstart` with an integer variable. To do that, look for `np.ones`, and add`dtype=int`

Framework/PythonInterface/plugins/algorithms/VelocityAutoCorrelations.py
```diff
@@ -308,7 +308,7 @@ def auto_correlation(self,u):
 
     def fold_correlation(self,w):
         # Folds an array with symmetrical values into half by averaging values around the centre
-        right_half=w[len(w)/2:]
+        right_half=w[len(w)//2:]
         left_half=w[:int(np.ceil(len(w)/2.0))][::-1]
```
AndreiSavici: `int(np.ceil(len(w)/2.0))` is equal to `len(w)-len(w)//2`

Framework/PythonInterface/plugins/algorithms/MatchPeaks.py
rosswhitfield: Unfortunately that won't work because we later assume it's a double when writing it to a TableWorkspace, see https://github.com/mantidproject/mantid/blob/master/Framework/PythonInterface/plugins/algorithms/MatchPeaks.py#L210

Framework/PythonInterface/plugins/algorithms/VelocityAutoCorrelations.py
rosswhitfield: That's correct but it's not related to what I'm fixing here. _i.e._ I don't care.

Framework/PythonInterface/plugins/algorithms/MatchPeaks.py
rosswhitfield:
```
580: ======================================================================
580: ERROR: testValidatorInput2 (__main__.MatchPeaksTest)
580: ----------------------------------------------------------------------
580: Traceback (most recent call last):
580:   File "/home/rwp/mantid/Framework/PythonInterface/test/python/plugins/algorithms/MatchPeaksTest.py", line 163, in testValidatorInput2
580:     alg_test = run_algorithm('MatchPeaks', **self._args)
580:   File "/home/rwp/mantid/Release/py3/bin/testhelpers/__init__.py", line 25, in run_algorithm
580:     alg.execute()
580: RuntimeError: Wrong datatype <int64> for column <MinBin> (expected <double>)
580:   at line 212 in '/home/rwp/mantid/Framework/PythonInterface/plugins/algorithms/MatchPeaks.py'
```

AndreiSavici:
OK :shipit:

PR #18751, opened 2017-02-08
Fix template usage

### LINKED ISSUE #18748: Correct template implementation
The template `getPropertyFromRun` implemented in `LoadHelper` cannot be used.
- its definition need to be moved to the header file to make it usable

Developers seem to copy this template (which I think is not desired) for example in:
- ConvertEmptyToTof
- LoadILLrefelctometryTest

### Platforms affected

All

vreimund:
The template function `getPropertyFromRun` contained in `LoadHelper.cpp` could not be used. Instead, it was copied to other files. 
- moved the template implementation to the header
- the MatrixWorkspace is not const anymore for more flexible usage that I need for the loader
- the include MatrixWorkspace.h must move to the header file as well 

**To test:**

Test possibility 1:
- checkout the different branch 18235_update_load_ill_reflectometry, where I used the template (test and source).
- make sure everything compiles (unit tests will pass for 80% due to current development)

Test possibility 2:
choose a source of a loader, then use the template for example like
Mantid::DataHandling::LoadHelper loadHelper;
or
#include "MantidDataHandling/LoadHelper.h"
LoadHelper loadHelper;
then
type sample_angle = loadHelper.getPropertyFromRun<type>(ws, "your_log_entry");
where ws is a MatrixWorkspace, the type should be the expected type of your_log_entry and your_log_entry should name a sample log entry.

Fixes #18748.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


AntonPiccardoSelg:
Refactoring looks good to me. Makes sense.
I went for test possibility 1 and built branch `18235_update_load_ill_reflectometry`. That worked fine. 

One point though:
The same code seems to exist in [`ConvertEmptyToTof`](https://github.com/mantidproject/mantid/blob/master/Framework/Algorithms/src/ConvertEmptyToTof.cpp#L442:L453) Would it be possible to have code sharing here?

reimundILL:
That's true, I will try code sharing.

reimundILL:
Unfortunately, LoadHelper.h lives in DataHandling and cannot be seen in Algorithms. But the template does not really belong to the class LoadHelper, it is much more general and can be moved to MantidAPI or MantidKernel. I did not find a really good place for it yet.

AntonPiccardoSelg:
@reimundILL It might be worth asking on the `#tech-qa` channel on Slack to see what people think regarding  a good location for this template.

reimundILL:
The tip using `getPropertyValueAsType` is from Ian Bush.

Framework/DataHandling/inc/MantidDataHandling/LoadHelper.h
```diff
@@ -4,6 +4,7 @@
 #include "MantidAPI/Run.h"
 #include "MantidKernel/System.h"
 #include "MantidNexus/NexusClasses.h"
+#include "MantidAPI/MatrixWorkspace.h"
```
AntonPiccardoSelg: Is this `include` still needed?

reimundILL: Correct, I forgot to push

AntonPiccardoSelg:
The removal of `getPropertyFromRun` from Mantid is a good improvement. The changes look good to me. Once all tests have passed, please :shipit:

PR #18752, opened 2017-02-08
Ensure MantidPlot runs with python's builtin range command by default

### LINKED ISSUE #18746: plotSpectrum doesn’t take a range() object
from james lord:

Just noticed this change in the latest nightly build (3.8.20170207.2221). My script wants to plot multiple spectra from a workspace.

w=mtd["workspacename"] # this workspace has 5 spectra or more
plotSpectrum(w,range(5))

Expected behavior

Plot appears with several spectra

Actual behavior

Error message:
TypeError: Incorrect type passed as index argument "xrange(5)"
 at line 2 in 'New script'
caused by line 249 in 'C:\MantidInstall\bin\pymantidplot__init__.py'
 caused by line 1156 in 'C:\MantidInstall\bin\pymantidplot__init__.py'

Steps to reproduce the behavior

See above

Note if I use list(range(5)) or tuple(range(5)) then it works as intended. Something is being picky about the exact types of list objects.

Platforms affected

Tried on Windows, maybe others?

martyngigg:
Description of work.

Deletes `six.moves.range` that was imported into the `MantidPlot` global namespace and so reverts back to python's built-in `range` command.

*This was a regression introduced since v3.8.0.*

**To test:**

The following breaks in nightly but should succeed after these changes:

```python
w = CreateSampleWorkspace()
plotSpectrum(w, range(5))
```

Fixes #18746

*Does not need to be in the release notes*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidPlot/mantidplotrc.py
+++ b/MantidPlot/mantidplotrc.py
@@ -22,6 +22,12 @@ if __name__ == '__main__':
     # Import MantidPlot python commands
     import mantidplot
     from mantidplot import *
+    try:
+        # The MantidPlot namespace is not ready for the python3-style range function
+        # so we ensure we revert back to the current built-in version
+        del range
+    except NameError:
+        pass
 
     # Make Mantid available
     import mantid
--- a/MantidPlot/test/MantidPlot1DPlotTest.py
+++ b/MantidPlot/test/MantidPlot1DPlotTest.py
@@ -52,6 +52,10 @@ class MantidPlot1DPlotTest(unittest.TestCase):
         g = plotSpectrum("fake", [0, 1])
         self.g = g
 
+    def test_plotSpectrum_range_command(self):
+        g = plotSpectrum("fake", range(0, 2))
+        self.g = g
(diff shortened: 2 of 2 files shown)


NickDraper:
tested with

``` python
w = CreateSampleWorkspace()
plotSpectrum(w, range(5))

a=Load("MAR11060")
plotSpectrum("a",range(5))

print type(range)
```

All is good the unit test should protect us into the future.
:shipit: once the builds are complete

rosswhitfield:
@martyngigg MantidPlot1DPlotTest now fails in Python 3 after this change http://builds.mantidproject.org/job/master_clean-ubuntu-16.04-python3/270/console

martyngigg:
Interesting. This is the same error that we were seeing before this change with python 2. I think this actually points to the real limitation of the code being that the `plotSpectrum` stuff doesn't understand the new iterator-style `range` and they need to be upgraded.

PR #18757, opened 2017-02-08
Expose FABADA minimizer options in ConvFit

### LINKED ISSUE #18745: Expose FABADA options in ConvFit
Currently the interface only has some of the FABADA options exposed. This ticket is to add a new checkable option called "advanced", that, when checked, shows the advanced options.

raquelalvarezbanos:
FABADA options should be accessible from `Indirect Data Analysis` -> `ConvFit`.

**To test:**

- Open `Indirect Data Analysis` interface and go to `ConvFit`.
- Check `Use FABADA`, there should be a new checkable option called `Advanced`. When this option is checked, all of the FABADA options should be visible.
- Ensure that all the FABADA minimizer options (except output workspace names) are visible in the interface (you can check the minimizer properties [here](https://github.com/mantidproject/mantid/blob/master/Framework/CurveFitting/src/FuncMinimizers/FABADAMinimizer.cpp#L60-L111))
- Check the minimizer: load `irs26173_graphite002_red` as `Sample`, `irs26173_graphite002_res` as `Resolution`, select `One Lorentzian`, select `Use FABADA`, and set the number of iterations to `2e+06`. Hit `Fit Single Spectrum`. It will take a long time but the fit should finish successfully, although fit results may not be good.

Fixes #18745.

Release notes: https://github.com/mantidproject/mantid/pull/18757/commits/58350a7fdb2f7ad6ed46248cc89336ec73063dee

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


KarlPalmen:
I found all the properties, except `Jump Acceptance Rate`
```
"ChainLength", static_cast<size_t>(10000) FOUND
"StepsBetweenValues", 10, FOUND (advanced)
"ConvergenceCriteria", 0.01, FOUND
"InnactiveConvergenceCriterion", static_cast<size_t>(5), FOUND (advanced)
"JumpAcceptanceRate", 0.6666666, NOT FOUND
// Simulated Annealing properties 
"SimAnnealingApplied", false, FOUND
"MaximumTemperature", 10.0, FOUND
"NumRefrigerationSteps", static_cast<size_t>(5), FOUND
"SimAnnealingIterations", static_cast<size_t>(10000), FOUND
"Overexploration", false, FOUND
// Output Properties 
"NumberBinsPDF", 20, FOUND
```
It is possible that this is `Acceptence Rate` but it has a different default of `0.25`.

raquelalvarezbanos:
Yes, `Jump Acceptance Rate` is called `Acceptance Rate` in the interface for some reason and used with a different default value. I am not sure why, but I have kept it as it was.

KarlPalmen:
Fit worked in 45 minutes.

KarlPalmen:
:shipit:

MantidQt/CustomInterfaces/src/Indirect/ConvFit.cpp
```diff
@@ -222,6 +209,72 @@ void ConvFit::setup() {
   updatePlotOptions();
 }
 
+/** Setup FABADA minimizer options
+*
+*/
+void ConvFit::initFABADAOptions() {
```
martyngigg: I realise this is how it was before but would it be possible to cut down how much this has to know about the actual properties on the minimizer?

If we used the `FuncMinimizerFactory` to create an instance of the `FABADA` minimizer we could then just iterate over the minimizer properties returned by `minimizer->getProperties()` to get the information. If new properties were added to the minimizer then they would automatically get pull over.

Would this be possible?

raquelalvarezbanos: I've checked and I can rewrite this method as:

```
void ConvFit::initFABADAOptions() {

  // Map from minimizer property name to interface property name
  std::map<std::string, std::string> interfaceNames = {
      {"FABADAOutputChain", "Output Chain"},
      {"FABADAChainLength", "Chain Length"},
      {"FABADAConvergenceCriteria", "Convergence Criteria"},
      {"FABADAJumpAcceptanceRate", "Acceptance Rate"},
      {"FABADAStepsBetweenValues", "Steps Between Values"},
      {"FABADAInactiveConvergenceCriterion", "Inactive Convergence Criterion"},
      {"FABADASimAnnealingApplied", "Sim Annealing Applied"},
      {"FABADAMaximumTemperature", "Maximum Temperature"},
      {"FABADANumRefrigerationSteps", "Num Refrigeration Steps"},
      {"FABADASimAnnealingIterations", "Sim Annealing Iterations"},
      {"FABADAOverexploration", "Overexploration"},
      {"FABADANumberBinsPDF", "Number Bins PDF"}};

  m_properties["FABADA"] = m_grpManager->addProperty("Bayesian");
  m_properties["UseFABADA"] = m_blnManager->addProperty("Use FABADA");
  m_properties["FABADA"]->addSubProperty(m_properties["UseFABADA"]);

  const auto &minimizer =
      Mantid::API::FuncMinimizerFactory::Instance().createMinimizer("FABADA");

  for (const auto &prop : minimizer->getProperties()) {

    if (prop->type() == "boolean") {

      const auto name = QString::fromStdString("FABADA" + prop->name());
      m_properties[name] = m_blnManager->addProperty(
          QString::fromStdString(interfaceNames[prop->name()]));
      m_blnManager->setValue(m_properties[name],
                             (prop->getDefault() == "1") ? true : false);

    } else if (prop->type() == "number") {

      const auto name = QString::fromStdString("FABADA" + prop->name());
      m_properties[name] = m_dblManager->addProperty(
          QString::fromStdString(interfaceNames[prop->name()]));
      m_dblManager->setValue(m_properties[name],
                             boost::lexical_cast<double>(prop->getDefault()));

    } else if (prop->type() == "MatrixWorkspace" ||
               prop->type() == "TableWorkspace") {
      // Don't expose these properties
    } else {
      // Assume size_t

      const auto name = QString::fromStdString("FABADA" + prop->name());
      m_properties[name] = m_dblManager->addProperty(
          QString::fromStdString(interfaceNames[prop->name()]));
      m_dblManager->setValue(m_properties[name],
                             boost::lexical_cast<double>(prop->getDefault()));
      m_dblManager->setDecimals(m_properties[name], 0);
    }
  }

  // Advanced options
  m_properties["FABADAAdvanced"] = m_blnManager->addProperty("Advanced");
  m_blnManager->setValue(m_properties["FABADAAdvanced"], false);

  m_cfTree->addProperty(m_properties["FABADA"]);
}
```

The reason why I need the `interfaceNames` map is because some properties are not called the same in the interface and the minimizer (`JumpAcceptanceRate` is for instance `Acceptance Rate` in the interface, and the others have white spaces). 
On the other hand, I need to deal with property types, as some of them are added to `m_blnManger` and others are added to `m_dblManager`, whereas `MatrixWorkspace` and `TableWorkspace` should not be exposed. If new `double`, `boolean` or `integer` properties are added to the minimizer, they will automatially get added to the interface, but if someone adds other types of properties, for instance a vector of numbers, they'll have to add another `if` statement.

Do you think I should change the method? I find the code more complicated, although perhaps there's a simpler approach.

martyngigg: You're right, I hadn't really appreciated that the names will need to be updated anyway and the types make it difficult to work with generically.

Let's leave it as it is.

PR #18771, opened 2017-02-09
VSI: Fix crash in splatter plot view 

martyngigg:
Description of work.

Adds a check of an unsigned is `> 0` before subtracting `1` to avoid wrapping. 

**To test:**

The dataset `MDEvent_SXD.nxs` seems to demonstrate the bug nicely. Load the data set and click show VSI. Before these changes MantidPlot would crash immediately.

No issue number

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


AntonPiccardoSelg:
### Load into StandardView and switch to the SplatterplotView

Works fine and did so previously. It produces:
image

### Load directly into SplatterplotView

Does not crash any longer :+1:, but produces a different result:

image

We also get a warning message:

image

PR #18780, opened 2017-02-09
Fix two issues with absolute rotations in RotateInstrumentComponent (via ComponentHelper::rotateComponent)

### LINKED ISSUE #18750: ComponentHelper: Rotating wrong if any parent component apart from immidiate parent is rotated
This should be the absolute rotation, not the relative rotation: https://github.com/mantidproject/mantid/blob/master/Framework/Geometry/src/Instrument/ComponentHelper.cpp#L81

Example:
```cpp
Load(Filename='MAR11015.raw', OutputWorkspace='MAR11015')
RotateInstrumentComponent(Workspace='MAR11015', ComponentName='wide angle bank', Z=1, Angle=-10, RelativeRotation=False)
RotateInstrumentComponent(Workspace='MAR11015', ComponentName='W1-1', Z=1, Angle=60, RelativeRotation=False)
RotateInstrumentComponent(Workspace='MAR11015', DetectorID=1101, Z=1, Angle=90, RelativeRotation=False)
```

The result is that detector 1101 is rotated to `-10` degree, instead of 90.

Note that the first two rotate calls are just for setting up a "bad" situtation: If those components have any previously set relative rotations (e.g., from the IDF) this will show the bug as well.

### LINKED ISSUE #18811: ComponentHelper: Absolute rotations wrong if any of the parents is rotated.
This should be the absolute rotation, not the relative rotation: https://github.com/mantidproject/mantid/blob/master/Framework/Geometry/src/Instrument/ComponentHelper.cpp#L83

Example:
```cpp
LoadEmptyInstrument(InstrumentName='MAR', OutputWorkspace='ws')
RotateInstrumentComponent(Workspace='ws', ComponentName='wide angle bank', Y=1, Angle=90)
RotateInstrumentComponent(Workspace='ws', ComponentName='W1-1', Z=1, Angle=90, RelativeRotation=False)
```

The result is `W1-1` at a weird location instead of being simple rotated around Z.

Note that the first rotate call is just for setting up a "bad" situtation: If those components have any previously set relative rotations (e.g., from the IDF) this will show the bug as well.

SimonHeybrock:
See issues.

**To test:**

Code review, and check if the following gives the expected result (in comparison to the old rotation computation, which did not):

For testing the first issue:

```python
LoadEmptyInstrument(InstrumentName='MAR', OutputWorkspace='ws')
RotateInstrumentComponent(Workspace='ws', ComponentName='wide angle bank', Z=1, Angle=-20, RelativeRotation=False)
RotateInstrumentComponent(Workspace='ws', ComponentName='W1-1', Z=1, Angle=60, RelativeRotation=False)
RotateInstrumentComponent(Workspace='ws', DetectorID=1101, Z=1, Angle=90, RelativeRotation=False)
```

For testing the second issue (need non-planar rotations, the second issue is not visible in the previous example):

```python
LoadEmptyInstrument(InstrumentName='MAR', OutputWorkspace='ws')
RotateInstrumentComponent(Workspace='ws', ComponentName='wide angle bank', Y=1, Angle=90)
RotateInstrumentComponent(Workspace='ws', ComponentName='W1-1', Z=1, Angle=90, RelativeRotation=False)
```

Fixes #18750.
Fixes #18811.

Release notes have been updated. This change affects the behavior of `RotateInstrumentComponent` when absolute rotations are enabled.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Geometry/test/ComponentHelperTest.h
+++ b/Framework/Geometry/test/ComponentHelperTest.h
@@ -8,6 +8,9 @@
 
 #include <boost/make_shared.hpp>
 
+using namespace Mantid::Kernel;
+using namespace Mantid::Geometry;
+
 class ComponentHelperTest : public CxxTest::TestSuite {
 public:
   // This pair of boilerplate methods prevent the suite being created statically
@@ -105,6 +108,59 @@ public:
     TS_ASSERT_DELTA(newRot.imagI(), expectedRot.imagI(), 1e-12);
     TS_ASSERT_DELTA(newRot.imagJ(), expectedRot.imagJ(), 1e-12);
     TS_ASSERT_DELTA(newRot.imagK(), expectedRot.imagK(), 1e-12);
+  }
+
+  void test_absolute_planar_rotation_takes_into_account_grandparent() {
+    const int banks = 1;
+    const int pixels = 1;
+    auto base =
+        ComponentCreationHelper::createTestInstrumentRectangular(banks, pixels);
+    auto pmap = boost::make_shared<ParameterMap>();
+    Instrument instrument(base, pmap);
+
+    const auto bank = instrument.getComponentByName("bank1");
+    const auto pixel = instrument.getDetector(1);
+
+    Quat rootRot(100.0, V3D(1, 2, 3));
+    Quat bankRot(110.0, V3D(1, 2, 3));
+    Quat pixelRot(111.0, V3D(1, 2, 3));
+
+    const auto type = ComponentHelper::TransformType::Absolute;
+    ComponentHelper::rotateComponent(instrument, *pmap, rootRot, type);
+    TS_ASSERT_EQUALS(instrument.getRotation(), rootRot);
+    TS_ASSERT_EQUALS(bank->getRotation(), rootRot);
+    TS_ASSERT_EQUALS(pixel->getRotation(), rootRot);
+    ComponentHelper::rotateComponent(*bank, *pmap, bankRot, type);
+    TS_ASSERT_EQUALS(instrument.getRotation(), rootRot);
+    TS_ASSERT_EQUALS(bank->getRotation(), bankRot);
+    TS_ASSERT_EQUALS(pixel->getRotation(), bankRot);
+    ComponentHelper::rotateComponent(*pixel, *pmap, pixelRot, type);
+    TS_ASSERT_EQUALS(instrument.getRotation(), rootRot);
+    TS_ASSERT_EQUALS(bank->getRotation(), bankRot);
+    TS_ASSERT_EQUALS(pixel->getRotation(), pixelRot);
+  }
+
+  void test_absolute_rotation_takes_into_account_parent_correctly() {
+    const int banks = 1;
+    const int pixels = 1;
+    auto base =
+        ComponentCreationHelper::createTestInstrumentRectangular(banks, pixels);
+    auto pmap = boost::make_shared<ParameterMap>();
+    Instrument instrument(base, pmap);
+
+    const auto bank = instrument.getComponentByName("bank1");
+    const auto pixel = instrument.getDetector(1);
+
+    Quat bankRot(10.0, V3D(1, 0, 0));
+    Quat pixelRot(20.0, V3D(0, 1, 0));
+
+    const auto type = ComponentHelper::TransformType::Absolute;
+    ComponentHelper::rotateComponent(*bank, *pmap, bankRot, type);
+    TS_ASSERT_EQUALS(bank->getRotation(), bankRot);
+    TS_ASSERT_EQUALS(pixel->getRotation(), bankRot);
+    ComponentHelper::rotateComponent(*pixel, *pmap, pixelRot, type);
+    TS_ASSERT_EQUALS(bank->getRotation(), bankRot);
+    TS_ASSERT_EQUALS(pixel->getRotation(), pixelRot);
   }
 
 private:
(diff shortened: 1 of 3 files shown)


SimonHeybrock:
@OwenArnold I decided to also fix the second rotation issue in this branch, see updated issue description for testing instructions.

SimonHeybrock:
@OwenArnold I think the reason is probably that most instruments are fixed: They are just loaded from IDF or Nexus files, and in that case parameters are inserted directly into the `ParameterMap`, so those will never exercise this code path.

OwenArnold:
We have two manual tests here that should be quite easy to create as unit tests in `ComponentHelper` via a Mocked or Faked `Component`/`CompAssembly`. I'd like to see that this gets enshrined as the normal behaviour now. Do you think you could write the tests?

SimonHeybrock:
@OwenArnold Added new tests. Running the new tests locally on master fails.

Framework/Geometry/test/ComponentHelperTest.h
```diff
@@ -107,6 +110,59 @@ class ComponentHelperTest : public CxxTest::TestSuite {
...
+    TS_ASSERT_EQUALS(pixel->getRotation(), rootRot);
+    ComponentHelper::rotateComponent(*bank, *pmap, bankRot, type);
+    TS_ASSERT_EQUALS(instrument.getRotation(), rootRot);
+    TS_ASSERT_EQUALS(bank->getRotation(), bankRot);
+    TS_ASSERT_EQUALS(pixel->getRotation(), bankRot);
+    ComponentHelper::rotateComponent(*pixel, *pmap, pixelRot, type);
+    TS_ASSERT_EQUALS(instrument.getRotation(), rootRot);
+    TS_ASSERT_EQUALS(bank->getRotation(), bankRot);
+    TS_ASSERT_EQUALS(pixel->getRotation(), pixelRot);
+  }
```
OwenArnold: This testing should be sufficient.

Framework/Geometry/test/ComponentHelperTest.h
```diff
@@ -107,6 +110,59 @@ class ComponentHelperTest : public CxxTest::TestSuite {
...
+    const int pixels = 1;
+    auto base =
+        ComponentCreationHelper::createTestInstrumentRectangular(banks, pixels);
+    auto pmap = boost::make_shared<ParameterMap>();
+    Instrument instrument(base, pmap);
+
+    const auto bank = instrument.getComponentByName("bank1");
+    const auto pixel = instrument.getDetector(1);
+
+    Quat rootRot(100.0, V3D(1, 2, 3));
```
OwenArnold: `Quat` has quite a few constructors. A minor improvement might be to comment or provide named variables so that it's clear the construction arguments are in this case rotation angle and axis.

OwenArnold:
This is definitely the right direction with the fix. One oddity I don't understand is that the rotation of the detector does not come out as exactly 90 degrees around `z` in the example provided.

```
LoadEmptyInstrument(InstrumentName='MAR', OutputWorkspace='ws')
# workspace index 3 is detector id 1101
print(mtd['ws'].getDetector(3).getRotation())
RotateInstrumentComponent(Workspace='ws', ComponentName='wide angle bank', Z=1, Angle=-20, RelativeRotation=False)
RotateInstrumentComponent(Workspace='ws', ComponentName='W1-1', Z=1, Angle=60, RelativeRotation=False)
RotateInstrumentComponent(Workspace='ws', DetectorID=1101, Z=1, Angle=90, RelativeRotation=False)
print(mtd['ws'].getDetector(3).getRotation())
```

or to simplify:
```
LoadEmptyInstrument(InstrumentName='MAR', OutputWorkspace='ws')
# workspace index 3 is detector id 1101
print(mtd['ws'].getDetector(3).getRotation())
RotateInstrumentComponent(Workspace='ws', DetectorID=1101, Z=1, Angle=90, RelativeRotation=False)
print(mtd['ws'].getDetector(3).getRotation())
```
This isn't the PI/2 rotation I would expect to see. Maybe this is a misunderstanding on my part? But I would expect this should rotate around the detector centre by 90 degrees.

Other comments are that the InstrumentView can be made to crash, and it doesn't seem to properly display the rotations, but these are separate issues I will write.

SimonHeybrock:
@OwenArnold What are you getting? Looks like a correct rotation to me, but I may be wrong. After setting it to 90 degree around Z the quaternion is `[0.707107,0,0,0.707107]`.

OwenArnold:
@SimonHeybrock Yes that's 81.02 degree rotation I think.

SimonHeybrock:
@OwenArnold This is printing the parameters of the quaternion. The first one is `cos(theta/2)`.

OwenArnold:
@SimonHeybrock Ok. This makes sense. I misread that as being Euler angles.

OwenArnold:
My conclusion from some testing directly in Eigen is that you're changes here are correct. https://github.com/DMSC-Instrument-Data/instrument-prototype/pull/45/files However, I'd still like to find out why I get some bizarre rotations reported for Mari. This could well be down to the the IDF parsing. I've already spotted some unpleasant implementation in `InstrumentDefinitionParser::setLocation`, but I'd like to investigate this further.

OwenArnold:
I found out what was going wrong with those Euler angles. Looking at the code, there's some kind of bizarre default convention. Here's an example I coded.

```python
from mantid.kernel import Quat
import math
theta = math.pi/2
quat = Quat(math.cos(theta/2), 0, 0, math.sin(theta/2))
print(quat.getEulerAngles())
print(quat.getEulerAngles(convention="XYZ"))
```
Gives
```
[0,90,-0]
[0,0,90]
```
The rotation (quaternion) itself looks OK.

OwenArnold:
:shipit:

PR #18799, opened 2017-02-10
Convert table to matrix workspace bug with numeric values

### LINKED ISSUE #18782: ConvertTableToMatrixWorkspace does not always convert numeric values.
### Expected behavior
A table can be converted into a matrix workspace if all of the elements are numeric. 
### Actual behavior
For some fields an error is produced stating that the data is non-numeric.

```
Error in execution of algorithm ConvertTableToMatrixWorkspace:
Cannot convert class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > to double.
```
Open muon analysis and load some data. 
Generate the results table with run_number selected. Try to create a matrix workspace of the results table (including run_number) and the error is thrown. I have also seen this error for sample_temp and temp_sample). 

### Platforms affected
Windows

AnthonyLim23:
Sometimes the ConvertTableToMatrixWorkspace was failing due to a string, however inspection of the elements revealed that all of the numbers were numeric. A check has been added to see if a string is passed and it then attempts to convert it to a double. Unit tests have also been added to check this scenario and if it will error when a non-numeric string is given. 

**To test:**
1. Run the AlgorithmsTest_ConvertTableToMatrixWorkspaceTest and make sure it passes. 

2. Load a file (e.g. HIFI0008447) and create a table. Then use ConvertTableToMatrixWorkspace on a few different columns (run number usually causes an error prior to the fix). See if the workspace is produced. 

Fixes https://github.com/mantidproject/mantid/issues/18782

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Algorithms/test/ConvertTableToMatrixWorkspaceTest.h
```diff
@@ -139,6 +139,88 @@ class ConvertTableToMatrixWorkspaceTest : public CxxTest::TestSuite {
     TS_ASSERT_THROWS_NOTHING(m_converter->setPropertyValue("ColumnX", "A"));
     TS_ASSERT_THROWS_NOTHING(m_converter->setPropertyValue("ColumnY", "B"));
   }
+  // test if it can convert a string to a double, if the string is numeric
+  void testStringToDouble() {
+
+    ITableWorkspace_sptr tws = WorkspaceFactory::Instance().createTable();
```
AntonPiccardoSelg: You could use `auto`

Framework/Algorithms/test/ConvertTableToMatrixWorkspaceTest.h
```diff
@@ -139,6 +139,88 @@ class ConvertTableToMatrixWorkspaceTest : public CxxTest::TestSuite {
...
+    tws->addColumn("str", "A");
+    tws->addColumn("double", "B");
+    tws->addColumn("double", "C");
+
+    size_t n = 10;
+    for (size_t i = 0; i < n; ++i) {
+      TableRow row = tws->appendRow();
```
AntonPiccardoSelg: `auto`

Framework/Algorithms/test/ConvertTableToMatrixWorkspaceTest.h
```diff
@@ -139,6 +139,88 @@ class ConvertTableToMatrixWorkspaceTest : public CxxTest::TestSuite {
...
+      std::string x = "1";
+      double y = double(i) * 1.1;
```
AntonPiccardoSelg: In general we avoid functional-style casts (and especially c-style casts) and use the new-style casting. Clearly here it won't make a huge difference, but you could consider it consistent with the rest of Mantid.

Framework/Algorithms/test/ConvertTableToMatrixWorkspaceTest.h
```diff
@@ -139,6 +139,88 @@ class ConvertTableToMatrixWorkspaceTest : public CxxTest::TestSuite {
...
+    auto &X = mws->x(0);
+    auto &Y = mws->y(0);
+    auto &E = mws->e(0);
+
+    for (size_t i = 0; i < tws->rowCount(); ++i) {
+      TableRow row = tws->getRow(i);
+      std::string x;
+      double y, e;
+      row >> x >> y >> e;
+      TS_ASSERT_EQUALS(boost::lexical_cast<double>(x), X[i]);
```
AntonPiccardoSelg: Consider using [this](http://en.cppreference.com/w/cpp/string/basic_string/stof)

Framework/Algorithms/test/ConvertTableToMatrixWorkspaceTest.h
```diff
@@ -139,6 +139,88 @@ class ConvertTableToMatrixWorkspaceTest : public CxxTest::TestSuite {
...
+      TS_ASSERT_EQUALS(y, Y[i]);
+      TS_ASSERT_EQUALS(e, E[i]);
+    }
+
+    boost::shared_ptr<Units::Label> label =
```
AntonPiccardoSelg: `auto`

Framework/DataObjects/inc/MantidDataObjects/TableColumn.h
```diff
@@ -162,7 +162,17 @@ template <class Type> class TableColumn : public API::Column {
...
             DoubleType;
-    return boost::numeric_cast<double, DoubleType>(m_data[i]);
+    if (typeid(m_data[i]) == typeid(std::string)) {
+      std::string trimmedString = boost::lexical_cast<std::string>(
+          m_data[i]); // so not to alter the original data
+
+      if (char(trimmedString.back()) == ':') {
+        trimmedString.pop_back(); // so not to alter the original data
+      }
+      return boost::lexical_cast<double>(trimmedString);
```
AntonPiccardoSelg: See above

Framework/DataObjects/inc/MantidDataObjects/TableColumn.h
```diff
@@ -162,7 +162,17 @@ template <class Type> class TableColumn : public API::Column {
         typename boost::mpl::if_c<boost::is_convertible<double, Type>::value,
                                   Type, InconvertibleToDoubleType>::type
             DoubleType;
-    return boost::numeric_cast<double, DoubleType>(m_data[i]);
+    if (typeid(m_data[i]) == typeid(std::string)) {
+      std::string trimmedString = boost::lexical_cast<std::string>(
+          m_data[i]); // so not to alter the original data
+
+      if (char(trimmedString.back()) == ':') {
```
AntonPiccardoSelg: Why do you need a conversion here?

AntonPiccardoSelg (changes requested):
There are some cases where `auto` could be used which I have not marked.

Framework/DataObjects/inc/MantidDataObjects/TableColumn.h
AnthonyLim23: I saw on one dataset a stray ":" on the end of the number, so created a way to handle it

Framework/Algorithms/test/ConvertTableToMatrixWorkspaceTest.h
AnthonyLim23: It is not clear to me where to use this. Could you please give a bit more detail?

AntonPiccardoSelg: You are trying to convert an `std::string` to a `double`. Your method is correct and works, but since `C++11` you can use a whole new family of functions which allow you to perform similar conversions to `lexical_cast`. The functions are `std::stod`, etc. An example would be:
```cpp
  std::string doubleAsString = "123.123";
  auto doubleAsDouble = std::stod(doubleAsString);
```
It is best to use as much as possible form the standard library.

Framework/DataObjects/inc/MantidDataObjects/TableColumn.h
AntonPiccardoSelg: Can you use `if (trimmedString.back() == ':') {...`? without the functional cast?

Framework/Algorithms/test/ConvertTableToMatrixWorkspaceTest.h
AnthonyLim23: ah ok, thank you.

Framework/DataObjects/inc/MantidDataObjects/TableColumn.h
AnthonyLim23: I will double check, but if I recall correctly there was an error about comparing a string to a char.

Framework/Algorithms/test/ConvertTableToMatrixWorkspaceTest.h
AnthonyLim23: I believe that I should keep the size_t, and if I replace TableRow with auto then I get a compiler error.

AntonPiccardoSelg: Yes the `size_t` is required. I actually thought that `TableRow` could be replaced by `auto`. Is it templated? I will have a look.

AntonPiccardoSelg: Ok I had a look. `appendRow` actually returns a `TableRowHelper` object. `TableRow` has a constructor ` TableRow(const TableRowHelper &trh)` hence we have an assignment initialization here. So yes, having `auto` there would give a compilation error because the stream operators are not overloaded for the `TableRowHelper` type.

AnthonyLim23:
made the code changes. Let me know if I missed anything.

AntonPiccardoSelg (approved):

Framework/DataObjects/inc/MantidDataObjects/TableColumn.h
```diff
@@ -162,7 +162,17 @@ template <class Type> class TableColumn : public API::Column {
         typename boost::mpl::if_c<boost::is_convertible<double, Type>::value,
                                   Type, InconvertibleToDoubleType>::type
             DoubleType;
-    return boost::numeric_cast<double, DoubleType>(m_data[i]);
+    if (typeid(m_data[i]) == typeid(std::string)) {
```
AntonPiccardoSelg: I am wondering if we should have something like
```cpp
  auto isStringType = std::is_convertible<Type, std::string>::value;
  if (isStringType) {
     ...
```
 here and avoid the `typeid`. To be fair it is probably a compile-time operation anyway in this case. I just don't come across the need to use `typeid` very often. @martyngigg  do you have an opinion on this?

AntonPiccardoSelg: Saying that, the code around this portion of Mantid relies heavily on `typeid`.

martyngigg: To be honest I don't really like the reliance on `typeid` at all. It's a known problem to use them across DLL boundaries and the code ends up with lots of `if` statements like this. It would be better to use some type of `traits` approach but I understand that overhauling that is outside of the scope of a fix for this.

I would like us to remove the usage of `boost::mpl` and switch to using the `std` `type_traits`.

Framework/DataObjects/inc/MantidDataObjects/TableColumn.h
```diff
@@ -162,7 +162,17 @@ template <class Type> class TableColumn : public API::Column {
...
+      std::string trimmedString = boost::lexical_cast<std::string>(
+          m_data[i]); // so not to alter the original data
+
+      if (trimmedString.back() == ':') {
```
martyngigg: I don't understand the need for this?

Our cases are essentially whether we can directly cast from `Type` to a `double` or if we have a string type that we need to use `lexical_cast` on. If Type is `string` and that string cannot be interpreted as a number then we should just throw shouldn't we?

AntonPiccardoSelg: 1. I agree there is no need for the lexical cast here. The element is already a string. 
2. As I understand it the fix addresses also the issue where the string cannot be converted because of a trailing colon. This colon is removed here and the string is further processed.

martyngigg: Hmm. It seems too low-level for that type of fix here (unless `TableColumn` is the one responsible for putting it there in the first place). It feels wrong that `TableColumn` has to understand the the string might contain any random characters.

AntonPiccardoSelg:
@AnthonyLim23  How do I create the table workspace with the string values?

AnthonyLim23:
I have sent you a message on slack, I am sorry but I have been away from my desk today.

AntonPiccardoSelg:
There are a few comments which we should clarify.

1. Why do we have that `lexical_cast` in the `toDouble` function?
2. Let's replace the `typeid` for a `type_trait`
3. When does the scenario occur with the colon?  Martyn is right, the handling of the colon should not happen on this level.

AnthonyLim23:
the lexical cast was part of 3. I cannot find the file that originally caused the error, so I will remove the code for dealing with a trailing ;

AntonPiccardoSelg:
Functional testing is fine. The error which that gets produced by `ConvertTableToMatrixWorkspace` has been solved. 

Using the overload as a replacement for the template specialization is a good solution here, else you would have to use `enable_if` which would be more complicated in this particular case.

There are two small points left:
1. There is some commented out code in `toDouble`
2. Can't the `boost::lexical_cast` be a `std::stod`? 

Else it is fine :+1:

Framework/DataObjects/inc/MantidDataObjects/TableColumn.h
```diff
@@ -155,14 +155,32 @@ template <class Type> class TableColumn : public API::Column {
...
+  /**
+   * Cast an string to double if possible. If it's impossible
+   * boost::numeric::bad_lexical_cast
+   * is throw. In case of an overflow boost::numeric::positive_overflow or
+   * boost::numeric::negative_overflow
+   * is throw.
+   * @param value :: The value of the element.
+   */
+
+  double convertToDouble(std::string value) const {
```
martyngigg: Can this be 'cons std::string &'?

AnthonyLim23: your right it should be (the one above is)

Framework/DataObjects/inc/MantidDataObjects/TableColumn.h
```diff
@@ -155,14 +155,32 @@ template <class Type> class TableColumn : public API::Column {
...
+  /**
+   * Cast an string to double if possible. If it's impossible
+   * boost::numeric::bad_lexical_cast
+   * is throw. In case of an overflow boost::numeric::positive_overflow or
+   * boost::numeric::negative_overflow
+   * is throw.
+   * @param value :: The value of the element.
+   */
+
+  double convertToDouble(const std::string value) const {
```
martyngigg: This is better but we're still missing the reference qualifier.

Also, please remove the commented out lexical_cast.

Framework/DataObjects/inc/MantidDataObjects/TableColumn.h
```diff
@@ -155,14 +155,32 @@ template <class Type> class TableColumn : public API::Column {
    * is throw. In case of an overflow boost::numeric::positive_overflow or
    * boost::numeric::negative_overflow
    * is throw.
-   * @param i :: The index to an element.
+   * @param value :: The value of the element.
    */
-  double toDouble(size_t i) const override {
+  template <typename T> double convertToDouble(const T &value) const {
     typedef
```
martyngigg: I think we should get rid of this `mpl::if` completely and just have `return numeric_cast<double, Type>(value)`. This way we will get a compile-time error straight from the `numeric_cast` if the Type is not convertible to a `double`. There is a boost mpl header that will also need to go.

Framework/DataObjects/inc/MantidDataObjects/TableColumn.h
AnthonyLim23: do you mean &value instead of value?

martyngigg: That's right.

AnthonyLim23:
is there anything else that I need to do for this?

Framework/DataObjects/inc/MantidDataObjects/TableColumn.h
martyngigg: @AnthonyLim23 Have you look into whether we are able to get rid of this mpl stuff?

AnthonyLim23: Sorry I do not understand what you mean by the mpl::if

AnthonyLim23: oh, I see on the line below. Never mind.

AnthonyLim23: we need to keep the code this way because 

``/home/osboxes/work/mantid2/Framework/DataObjects/src/TableColumn.cpp:19:1:   required from here
/usr/include/boost/numeric/conversion/converter_policies.hpp:187:95: error: invalid static_cast from type ‘const Mantid::Kernel::V3D’ to type ‘boost::numeric::raw_converter<boost::numeric::conversion_traits<double, Mantid::Kernel::V3D> >::result_type {aka double}’
``

AntonPiccardoSelg: To put this into context, we allow `V3D` in TableColumns, see [here](https://github.com/mantidproject/mantid/blob/master/Framework/DataObjects/src/TableColumn.cpp#L16), hence we have to defer the error to runtime (should this method with a `V3D` ever be called).

martyngigg: Ah yes I see. Can we then at least move to using the `std::` versions of the `type_traits` and remove the boost code? `boost::if_c` should be able to be replaced with [`std::conditional`](http://en.cppreference.com/w/cpp/types/conditional) and `boost::is_convertible` by `std::is_convertible.`

martyngigg (commented):
This is very nearly there. The final thing is to tidy up the headers. There is still several boost headers included which we should no longer need. Can we remove those?

AnthonyLim23:
The boost library are used elsewhere in the code: 

mpl - line 195
lexical_Cast line 213

AnthonyLim23:
I have removed two of the boost libraries

AntonPiccardoSelg:
Followed the test instructions and crated a table workspace via the Muon interface (and fitting). When applying the `ConvertTableToMatrixWorkspace` and selecting `run_number` I don't get the error. It seems to be fixed. Please :shipit:

PR #18800, opened 2017-02-10
Use system provided FindEigen3

peterfpeterson:
Dropping the extra arguments from `find_package` causes cmake to use the system provided one. This still doesn't work with ubuntu, but that appears to be a problem with how that package was created. It does fix system eigen3 on fedora.

Also note that the eigen3 version is now a minimum rather than exact match.

**To test:**

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/buildconfig/CMake/Eigen.cmake
+++ b/buildconfig/CMake/Eigen.cmake
@@ -8,8 +8,7 @@ option(USE_SYSTEM_EIGEN "Use the system installed Eigen - v${eigen_version}?" OF
 
 if(USE_SYSTEM_EIGEN)
   message(STATUS "Using system Eigen")
-  find_package(Eigen3 ${eigen_version} EXACT REQUIRED
-               PATH_SUFFIXES eigen3 eigen )
+  find_package(Eigen3 ${eigen_version} REQUIRED)
 else()
   message(STATUS "Using Eigen in ExternalProject")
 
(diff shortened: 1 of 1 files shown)


SimonHeybrock:
@peterfpeterson Any higher `Eigen` version causes compilation warnings, doesn't it?

ianbush:
Ah yes, when I tested on the 3.3.2 version there were some warnings. You will not see them now as nothing is actually using Eigen yet though.

peterfpeterson:
@SimonHeybrock We should fix library issues upstream.

quantumsteve:
By default, this will still build with the external project, but I do like the option for intrepid developers to use the system headers.

Do we need a specific bug-fix in 3.2.10, or would it be better to just require eigen 3.2 or later?

ianbush:
@quantumsteve - I think the system package was desired for use on the build servers, but there it would probably be nicer to stick to exactly 3.2.10.

If this is just for intrepid developers then I would agree making this 3.2.* would be fine.

SimonHeybrock:
I am fine with this as long as it does not default to the system provided one.

ianbush:
@peterfpeterson Do you want to make this PR for 3.2 minimum, then I'll mark it as ready to go.

ianbush:
:shipit: - this does not need to wait for the builds as none of them have this option enabled.

PR #18813, opened 2017-02-13
Indirect Absorption - Update Interface

### LINKED ISSUE #18577: Indirect Absorption GUI : improve the layout of the options
This is a follow-up of the PR #18282 .

GUI view should be upgraded to look like attached.

Make sure the unit is specified for each input. e.g. it is not clear that sample angle is in degrees.

Optionally, most of the validation of the input properties should also be handled in GUI, before passing to the underlying algorithms.

screen shot 2017-01-26 at 09 53 05

louisemccann:
The Indirect Absorptions interface has been refactored to improve the ordering of input parameters. Default beam dimensions are now obtained in the interface rather than the algorithm. 

**To test:**
Interfaces>Indirect>Corrections>Absorption
Interface should be ordered in a similar way to that shown in the issue and units should be appropriate.
Input Files: [Corrections.zip](https://github.com/mantidproject/mantid/files/770704/Corrections.zip)

IRIS test:
Sample: `26176`
Container:`26174`
Beam should be picked up as: Width: `2.1cm` Height: `3.2cm`
Test with `Flat Plate` and use `4cm` for all fields apart from angle which should be `0`
Mass Density `1` Chemical Formula `H2-O`

ILL test:
Sample: `sample_red`
No need for container.
Beam should be: Width `3.5cm` Height `3.0cm`
Test with `Flat Plate` and use `4cm` for all fields apart from angle which should be `0`
Mass Density `1` Chemical Formula `H2-O`

Beam dimensions can also be checked in the IRIS/IN16B instrument parameter files.

Fixes #18577

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


KarlPalmen:
IRIS test did not work. Unexpected exception occurred and mass density of 1 was rejected.

KarlPalmen:
Branch needs merging with master so it can build.

KarlPalmen:
I ran the IRIS test and after several minutes of Monte Carlo adsorption I got
```
Error in execution of algorithm IndirectFlatPlateAbsorption:
Left hand side smaller than right hand side.
  at line 278 in 'C:/Mantid/staging_mantid/Framework/PythonInterface/plugins/algorithms\WorkflowAlgorithms\IndirectFlatPlateAbsorption2.py'
```
error occurring.

louisemccann:
@KarlPalmen try with 
[Corrections.zip](https://github.com/mantidproject/mantid/files/790453/Corrections.zip)

KarlPalmen:
Tests ran without error on new data file. Spectrum plots of workspaces looked OK. :shipit: 

Another issue, is that is it possible to do a quickly check of the data files, so the correction can end with error if the data files are not OK before a time consuming Monte-Carlo algorithm?

martyngigg:
I don't think it is a good idea to remove the units from the absorption correction fields - they are important to understand what values to enter. How about including them in the label text instead of in the text field if it's not easy to extract the number out?

gvardany:
Initially I meant to put the missing units in labels, but to me it looks nicer as suffix now. Is there a problem to extract the numbers?

martyngigg:
I agree it reads better. Can this be done, @louisemccann ?

gvardany:
It seems that it is already done, is it not?

louisemccann:
It's not the numbers. Mass density and number density have different units (g/cm3) and (atoms/amstrong3) respectively and I'm not sure how to change the suffix when changing the option. 
Also the text in the option gets passed to the algorithm, so it ought to be "Mass Density" or "Number Density" as this is what the algorithm expects

gvardany:
How about this ? If you like can you do for the can density as well? I could not put the proper symbol for angstrom though.

KarlPalmen:
This works for me. I now see that the units for mass-density occur after the number as in the widgets above. One thing I noticed is that in the `Container Details` box not used in testing, there is a Mass Density widget without mass units in it.

louisemccann:
@gvardany yes, that works. I'll try and get it working for container density too

gvardany:
Great, regarding number density I am not sure it should be `1/A3`, or just `/A3`. Maybe the latter is less confusing.

martyngigg:
@louisemccann It looks like some conflicts have developed with the recent movement of the python algorithm tests.

martyngigg:
@KarlPalmen Could you retest this, please?

KarlPalmen:
Builds now OK. I go ahead with testing again.

KarlPalmen:
In the IRIS test I got
```
MonteCarloAbsorption successful, Duration 2.58 seconds
Error in execution of algorithm IndirectFlatPlateAbsorption:
Left hand side smaller than right hand side.
  at line 278 in 'C:/Mantid/staging_mantid/Framework/PythonInterface/plugins/algorithms\WorkflowAlgorithms\IndirectFlatPlateAbsorption2.py'
```

louisemccann:
@KarlPalmen try with 
[Corrections.zip](https://github.com/mantidproject/mantid/files/824738/Corrections.zip)

KarlPalmen:
Tests worked with corrected files.

martyngigg:
:+1:

PR #18393, opened 2017-01-10
Add missing synchronization in ParameterMap::setDetectorInfo.

SimonHeybrock:
This fixes an issue with caching a pointer to `DetectorInfo` in `ParameterMap`. This is mainly used for providing legacy Python wrappers. This bug causes sporadic segfaults in `SCDCalibratePanelsTest`

**To test:**

Code review.

Fixes #18390.

No release notes, this bug was introduced very recently.

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)? Is it well structured with small focussed classes/methods/functions?
- [ ] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant documentation been added/updated?
- [ ] Is user-facing documentation written in a user-friendly manner?
- [ ] Has developer documentation been updated if required?

- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

PR #18831, opened 2017-02-13
Avoid synchronization issues with ParameterMap::setDetectorInfo.

### LINKED ISSUE #18390: SCDCalibratePanelsTest segfaults sporadically on master.
On my system (Ubuntu 16.04, gcc) `SCDCalibratePanelsTest` sporadically segfaults for the current master branch. Backtraces indicates that this could be related to recent changes for Instrument-2.0, but I could not reproduce it systematically enough so far.

SimonHeybrock:
After https://github.com/mantidproject/mantid/pull/18393 SCDCalibratePanels would still show sporadic segfaults, this is an attempt to fix this.

I now set the `DetectorInfo` in the `ParameterMap` just after creation, which avoids threading issues.

I have been running `SCDCalibratePanelsTest` in a loop for 5 hours now. It segfaulted twice, which is much less than previously. Also, previously I had observed two kinds of segfaults, now I see only one in relation to `tbb::concurrent_unordered_map`. I do not know if that is related to any changes with `DetectorInfo`, or if it exists independently.

In conclusion, this might not fix all issues, but it should be a significant improvement.

**To test:**

Code review

Fixes #18390.

No release notes, internal change.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/src/MatrixWorkspace.cpp
+++ b/Framework/API/src/MatrixWorkspace.cpp
@@ -72,11 +72,6 @@ MatrixWorkspace::MatrixWorkspace(const MatrixWorkspace &other)
   m_isCommonBinsFlag = other.m_isCommonBinsFlag;
   m_masks = other.m_masks;
   // TODO: Do we need to init m_monitorWorkspace?
-
-  // This call causes copying of m_parmap (ParameterMap). The constructor of
-  // ExperimentInfo just kept a shared_ptr to the same map as in other, which
-  // is not enough as soon as the maps in one of the workspaces it edited.
-  instrumentParameters();
 }
 
 /// Destructor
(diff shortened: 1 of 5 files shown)


SimonHeybrock:
This still segfaults, but less frequently. So this needs more work. :(

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -72,11 +72,6 @@ MatrixWorkspace::MatrixWorkspace(const MatrixWorkspace &other)
   m_isCommonBinsFlag = other.m_isCommonBinsFlag;
   m_masks = other.m_masks;
   // TODO: Do we need to init m_monitorWorkspace?
-
```
martyngigg: Am I correct in saying that this is no longer required because of the new copy behaviour in `setInstrument`?

SimonHeybrock: Yes: The copy constructor is using `copyExperimentInfoFrom`, which is using `setInstrument`, which as you point out is not doing the copy directly.

martyngigg (approved):
I'm happy with the changes.

:shipit:

Issue #18832, opened 2017-02-13
Sliceviewer axis incorrect when showing HKL axes orthogonally


lottiegreenwood:
When opening HKL workspaces in the sliceviewer, it attempts to show a nonorthogonal axes but does not skew the data. This results in an inaccurate axis.

PR #18833, opened 2017-02-13
Added check for if nonorthogonal axes should be displayed

lottiegreenwood:
Fixes https://github.com/mantidproject/mantid/issues/18832

To test, open a workspace with HKL axes. Change the dimensions shown and ensure the axes remain consistent. Switch back and forth between nonorthogonal and orthogonal modes and also check this.


AntonPiccardoSelg:
@lottiegreenwood Which data set did you use? 

The change makes sense and I saw the issue on your screen, but I am having trouble reproducing it. It works quite well with and without the fix. Maybe it is more visible with some data sets.

AntonPiccardoSelg:
The changes make sense and I was able to reproduce the bug without the fix. With the fix the dimensions are displayed correctly. 
Please :shipit:

Issue #12270, opened 2015-03-25
Create some basic reduction scripts for users


DanNixon:
This issue was originally [TRAC 11431](http://trac.mantidproject.org/mantid/ticket/11431)

Create some scripts to do a QENS reduction and diffraction reduction.

The QENS script should (AFAIK):
- Take a list of raw files, and optionally a calibration workspace and run `ISISIndirectEnergyTransfer`
- Optionally run `SofQW`
- Optionally run `ElasticWindowMultiple` and `MSDFit`

I'm not entirely sure what the diffraction script should do, check with Sanghamitra but I think it just needs to run the appropriate diffraction algorithm (either `OSIRISDiffractioReduction` or `ISISIndirectDiffractionReduction`).

I don't think the TOSCA scientists want anything similar for INS, since their reduction is pretty much a one step process already.

Issue #18216, opened 2016-12-07
Indirect QuickRun Gui


louisemccann:
Create an interface for all Indirect QuickRun algorithms. Currently algorithms included would be:

- [x] EnergyWindowScan - already implemented
- [x] SofQWMomentsScan - #17986 - merged
- [x] IndirectDiffusionReduction - #18223 
- [x] SampleChanger - #18844 

This interface would follow the setup of the other Indirect interfaces, under a new category named "QuickRun" and would have different tabs for each algorithm.
This interface will be written in python using `PyQt` and automatic testing will be implemented.
Also see plan:
[QuickRuns_scheme.pdf](https://github.com/mantidproject/mantid/files/837984/QuickRuns_scheme.pdf)

Mock ups will be produced and agreed first before proceeding with any development.

PR #18845, opened 2017-02-14
Indirect QuickRun - add diffusion option

### LINKED ISSUE #18223: Indirect QuickRun - Diffraction
Add an option to run through IndirectDiffractionReduction to existing EnergyWindowScan algorithm. The diffraction should only be in diffspec mode and it is assumed that default inputs will be used, so the users shouldn't have to modify them.

louisemccann:
EnergyWindowScan has been extended to include diffraction reduction and renamed to IndirectReductionAndAnalysis to better represent what the algorithm does.

**To test:**
InputFiles: [Diff_inputs.zip](https://github.com/mantidproject/mantid/files/774174/Diff_inputs.zip)

1
Interfaces>Indirect>Diffraction
Input: `26181` 
(this should find the appropriate run in the data archive otherwise browse to `IRS26181` in files provided)
Ensure Instrument = `ISIS` and mode = `diffspec`
Press `Run`

2
```Python
IndirectReductionAndAnalysis(InputFiles='26176',
                             Instrument='IRIS',
                             Analyser='graphite',Reflection='002',
                             SpectraRange='3, 50',
                             ElasticRange='-0.5, 0',
                             InelasticRange='0, 0.5',
                             GroupingMethod='All',
                             MSDFit=False,
                             Diffraction=True,
                             DiffractionSpectraRange=[105,112])
```
Ensure irs26176_diffspec_red is the same for both methods

Fixes #18223 #12270

[Release Notes](https://github.com/mantidproject/mantid/blob/be1b1486734488150de3569407f561772a855007/docs/source/release/v3.10.0/indirect_inelastic.rst)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


samueljackson92:
@louisemccann Question: `ElasticWindowScan` is already an algorithm in 3.9 right? So by renaming the algorithm to `IndirectReductionAndAnalysis` does this break the Python API? `ElasticWindowScan` is pretty niche, but shouldn't we deprecate for a version then remove?

One option would be to just point the deprecated `ElasticWindowScan` at the new algorithm.

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/IndirectReductionAndAnalysis.py
```diff
@@ -213,8 +247,22 @@ def PyExec(self):
...
+            diff_red_alg.setProperty('ContainerFiles',self._can_files)
+            diff_red_alg.setProperty('SumFiles', self._sum_files)
+            diff_red_alg.setProperty('LoadLogFiles', self._load_logs)
+            diff_red_alg.setProperty('Instrument', self._instrument_name)
+            diff_red_alg.setProperty('Mode', self._diff_mode)
+            diff_red_alg.setProperty('SpectraRange',self._diff_spec)
+            diff_red_alg.setProperty('GroupingPolicy', self._grouping_method)
+            diff_red_alg.setProperty('OutputWorkspace', self._diff_workspace)
+            diff_red_alg.execute()
+            self.setProperty("DiffractionWorkspace", diff_red_alg.getProperty("OutputWorkspace").value)
```
samueljackson92: All of the stuff under the if statement could be refactored into a method.

samueljackson92: In fact I'd go for one method per algorithm i.e. something like `self.run_diffraction_reduction()`, `self.run_msg()` ...

samueljackson92 (changes requested):
See suggested changes above, or come talk to me if you disagree.

SpencerHowells:
We need EnergyWindowScan for use by the Interface in #18216

samueljackson92:
@SpencerHowells I agree the algorithm needs to exist. My problem is with renaming. Renaming would potentially break any user scripts that we have no control over that use the current name. Hence we should really deprecate the name `EnergyWindowScan` name for one version of Mantid, then remove any reference of the algorithm called `EnergyWindowScan` in favour of `IndirectReductionAndAnalysis`.

louisemccann:
It has been decided that EnergyWindowScan and the DiffractionScan should be 2 different algs so I am going to close this.

PR #18847, opened 2017-02-14
LoadILLTOF: add monitor spectra to the end of the workspace

### LINKED ISSUE #18842: Spectrum numbers don't match to detector IDs in IN4, IN5 and IN6 workspaces
The instrument scientists have requested that the spectrum numbers should match the detector IDs on ILL's time-of-flight instruments.

soininen:
This PR modifies [`LoadILLTOF`](http://docs.mantidproject.org/nightly/algorithms/LoadILLTOF-v2.html) so that the monitor data is at the end of the loaded workspace, not at the beginning. The change makes the spectrum numbers match with the detector IDs, as requested by the instrument scientists at ILL.

**To test:**

Load IN4, IN5 and IN6 data. Some test data is available under `ExternalData/Testing/Data/U
nitTest/ILL/` in Mantid's build directory. Check that

- Monitor data is at the end of the workspaces.
- Spectrum numbers and detector IDs match except for monitors, which have IDs such as 0, 1001, 1002 and 1003 (instrument dependent)

Fixes #18842.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


AntonPiccardoSelg:
The code changes make sense. 
In the new output the monitor is at the end:
image

I have found though that in the detector table, there is an issue with the monitor. It is not shown as such:
image

The last two fields seem to be wrong. I traced the issue and found that it originates [here](https://github.com/mantidproject/mantid/blob/master/MantidPlot/src/Mantid/MantidUI.cpp#L1369:L1383).
When  `double usignTheta = spectrumInfo.twoTheta(wsIndex) * 0.5;` is executed it throws an `std::logic_error` which is not caught by this `try-catch` block , but by the outer one, which then sets the monitor flag to a strange value.
Would catching the `std::logic_error` together with the `std::runtime_error` be the right approach? Would an entry in the `Q` column of `No Efixed` be the right thing for monitors? 

`

SimonHeybrock:
@AntonPiccardoSelg Yes, `SpectrumInfo::twoTheta` throws an exception for monitors, since a scattering angle does not make sense for monitors, I think?

soininen:
Also the *R* in the spherical coordinates for the monitor is negative in the detector table. Is this somehow related?
There can be monitors in the beam axis before and after the sample. For these, the scattering angle is either 0 or 180 degrees. I don't know if anyone is using this information, though.

SimonHeybrock:
@soininen That is the old definition, `SpectrumInfo` throws an exception. Or do you consider backscattering from the sample into a monitor before the sample?

soininen:
@SimonHeybrock I think it is OK to throw the exception. We just need to fix the detector table. I'll open an issue for it.

AntonPiccardoSelg:
The changes are fine. The discovered bug will be dealt with in separate PR. Please :shipit:

AndreiSavici:
Next time check if the release note link is fine

PR #18848, opened 2017-02-14
Crystal Field - Allow arbitrary J (Ion type)

### LINKED ISSUE #17984: Extend crystal field module to calculate for arbitrary J multiplets
This is a feature request. At present, the crystal field fitting functions (port of VMS Focus program) are restricted to trivalent rare earth ions with particular values of the total angular momentum quantum number J. Some users would also like to use it for calculations with other magnetic ions, with different values of J (or equivalently the orbital angular momentum L or spin S). 

Essentially, this requires a few small amendments of the code to allow arbitrary J, and to omit certain factors which are only required for rare earths - the current code to calculate the crystal field Hamiltonian matrix elements and to diagonalise this and to calculate the neutron cross-sections should suffice without any changes.

mducle:
Changes the CrystalField code to allow non-rare earth ions (e.g. transition metal ions, assuming spin-only / orbitally quenched) to be calculated. For these ions, the `g_J` value is fixed to 2 (spin-only value).

**To test:**

Initialise a `CrystalField` object with `Ion=S<n>` where `<n>` is an integer or half-integer number (decimal notation, e.g. `Ion=S1.5`), and set `B20` to 1. Print out the Hamiltonian and check that it has `2<n>+1` rows/columns and the diagonal elements follow the sequence: `3*Jz^2-J(J+1)`, where the Jz runs from -J to +J. E.g.

```
J = 1
cf = CrystalField.fitting.CrystalField('J'+str(J), 'C2v', B20=1)
print cf.getHamiltonian()
checksum = np.sum(np.sum( np.diag(3*(np.arange(-J,J+1)**2) - J*(J+1)) - cf.getHamiltonian() ))
print "checksum = " + str(checksum)
```

should give the output:

```
[[ 1.+0.j  0.+0.j  0.+0.j]
 [ 0.+0.j -2.+0.j  0.+0.j]
 [ 0.+0.j  0.+0.j  1.+0.j]]
checksum = 0j
```

Check that checksum is always `0` for any value of `J`.

Also, check that all the usual operations work without errors, e.g.

```
cf = CrystalField('J'+str(J), 'C2v', B20=1, FWHM=0.1,Temperature=10)
spec = cf.getSpectrum();
plot(*spec)
plot(*cf.getSusceptibility(range(1,300), Hdir=[1, 1, 1], Lambda=-10))
ws = CreateWorkspace(*spec)
cf1 = CrystalField('J'+str(J), 'C2v', B20=0.9, FWHM=0.2,Temperature=10)
fit = CrystalFieldFit(Model=cf, InputWorkspace=ws, MaxIterations=100)
fit.fit()
```

Fixes #17984

In release notes.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/CurveFitting/src/Algorithms/CrystalFieldEnergies.cpp
```diff
@@ -46,7 +46,7 @@ const std::string CrystalFieldEnergies::summary() const {
 void CrystalFieldEnergies::init() {
 
   // Input
-  auto bounds = boost::make_shared<Kernel::BoundedValidator<int>>(1, 13);
+  auto bounds = boost::make_shared<Kernel::BoundedValidator<int>>(-99, 13);
   declareProperty("Nre", 1, bounds, "A rare earth ion. Possible values are: "
```
mantid-roman: For the future: can this property be made a string instead of int?

mducle: Yes. I guess we can refactor the code in `CrystalElectricField` to use a string instead. And then we can remove all instances of `nre`. I just implemented as it is now because it was the easiest way to go... but I can change it.

mantid-roman: I don't think we need to do it now. Some time in the future.

Framework/CurveFitting/src/Functions/CrystalFieldPeaksBase.cpp
```diff
@@ -336,11 +336,37 @@ void CrystalFieldPeaksBase::calculateEigenSystem(DoubleFortranVector &en,
...
+          auto J2 = std::stof(ion.substr(1)) * 2.;
+          if (fabs(J2 - (int)J2) < 0.001) {
+            nre = -(int)J2;
+            break;
+          }
+          // Catch exceptions thrown by stof so we get a more meaningful error
+        } catch (const std::invalid_argument &) {
+        } catch (const std::out_of_range &) {
+        }
+      }
```
mantid-roman: Maybe an exception with a different message should be thrown here warning that the `J` value is wrong

Framework/CurveFitting/src/Functions/CrystalFieldPeaksBase.cpp
```diff
@@ -336,11 +336,37 @@ void CrystalFieldPeaksBase::calculateEigenSystem(DoubleFortranVector &en,
...
-    throw std::runtime_error("Unknown ion name passed to CrystalFieldPeaks.");
+    // If 'Ion=S2', or 'Ion=J2.5' etc, interpret as arbitrary J values with gJ=2
+    // Allow lower case, but values must be half-integral. E.g. 'Ion=S2.4' fails
+    switch (ion[0]) {
+    case 'S':
+    case 's':
+    case 'J':
+    case 'j': {
+      std::string numbers("0123456789");
+      if (numbers.find(ion[1]) != std::string::npos) {
```
mantid-roman: Can `std::isdigit` be used here?
Are we certain that `ion.size() >= 2`?

Framework/CurveFitting/src/Functions/CrystalFieldPeaksBase.cpp
```diff
@@ -336,11 +336,37 @@ void CrystalFieldPeaksBase::calculateEigenSystem(DoubleFortranVector &en,
...
+        // Need to store as 2J to allow half-integer values
+        try {
+          auto J2 = std::stof(ion.substr(1)) * 2.;
+          if (fabs(J2 - (int)J2) < 0.001) {
+            nre = -(int)J2;
```
mantid-roman: Should `nre` be checked against `maxNre`?

Framework/CurveFitting/src/Functions/CrystalFieldPeaksBase.cpp
mducle: Yes. And no... I'll change the code to check.

Framework/CurveFitting/src/Functions/CrystalFieldPeaksBase.cpp
mducle: Ok. Will do.

mantid-roman:
When I run the test with too high `J` the error message is not very helpful:

`ValueError: When converting parameter "Nre": Selected value -200 is < the lower bound (-99)`

Could it say instead that `J` is too high?

mducle:
Yes. I'll change the error messages to be more informative.

mantid-roman:
Looks good. :shipit:

PR #18873, opened 2017-02-15
SpectrumInfo and DetectorInfo rollout: InstrumentView

### LINKED ISSUE #18688: SpectrumInfo and DetectorInfo rollout: InstrumentView
```
MantidQt/MantidWidgets/.../InstrumentView/ComponentActor.cpp
MantidQt/MantidWidgets/.../InstrumentView/InstrumentActor.cpp
MantidQt/MantidWidgets/.../InstrumentView/InstrumentWidgetPickTab.cpp
MantidQt/MantidWidgets/.../InstrumentView/ObjComponentActor.cpp
MantidQt/MantidWidgets/.../InstrumentView/ProjectionSurface.cpp
MantidQt/MantidWidgets/.../InstrumentView/RotationSurface.cpp
MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/InstrumentView/ComponentActor.h
MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/InstrumentView/InstrumentActor.h
MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/InstrumentView/ProjectionSurface.h
```

mantid-roman:
Changed code to use DetectorInfo to access detectors in InstrumentView.

**To test:**

Code review.

Fixes #18688.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/InstrumentView/UnwrappedSurface.h
+++ b/MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/InstrumentView/UnwrappedSurface.h
@@ -57,7 +57,7 @@ public:
   Mantid::detid_t detID;  ///< Detector ID
   Mantid::Kernel::V3D position;              ///< Detector position
   Mantid::Kernel::Quat rotation;             ///< Detector orientation
-  Mantid::Geometry::Object_const_sptr shape; ///< Shape of the detector
+  boost::shared_ptr<const Mantid::Geometry::Object> shape; ///< Shape of the detector
   Mantid::Kernel::V3D scaleFactor;           ///< Detector's scale factor
 };
 
--- a/MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/InstrumentView/UnwrappedSurface.h
+++ b/MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/InstrumentView/UnwrappedSurface.h
@@ -44,7 +44,7 @@ class UnwrappedDetector {
 public:
   UnwrappedDetector();
   UnwrappedDetector(const unsigned char *c,
-                    boost::shared_ptr<const Mantid::Geometry::IDetector> det);
+                    const Mantid::Geometry::IDetector &det);
   UnwrappedDetector(const UnwrappedDetector &other);
   UnwrappedDetector &operator=(const UnwrappedDetector &other);
   unsigned char color[3]; ///< red, green, blue colour components (0 - 255)
@@ -54,7 +54,11 @@ public:
   double height;          ///< detector height in units of v
   double uscale;          ///< scaling factor in u direction
   double vscale;          ///< scaling factor in v direction
-  Mantid::Geometry::IDetector_const_sptr detector;
+  Mantid::detid_t detID;  ///< Detector ID
+  Mantid::Kernel::V3D position;              ///< Detector position
+  Mantid::Kernel::Quat rotation;             ///< Detector orientation
+  Mantid::Geometry::Object_const_sptr shape; ///< Shape of the detector
+  Mantid::Kernel::V3D scaleFactor;           ///< Detector's scale factor
 };
 
 /**
--- a/MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/InstrumentView/UnwrappedSurface.h
+++ b/MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/InstrumentView/UnwrappedSurface.h
@@ -47,6 +47,7 @@ public:
                     const Mantid::Geometry::IDetector &det);
   UnwrappedDetector(const UnwrappedDetector &other);
   UnwrappedDetector &operator=(const UnwrappedDetector &other);
+  bool isValid() const;
   unsigned char color[3]; ///< red, green, blue colour components (0 - 255)
   double u;               ///< horizontal "unwrapped" coordinate
   double v;               ///< vertical "unwrapped" coordinate
--- a/MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/InstrumentView/UnwrappedSurface.h
+++ b/MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/InstrumentView/UnwrappedSurface.h
@@ -56,10 +56,11 @@ public:
   double uscale;          ///< scaling factor in u direction
   double vscale;          ///< scaling factor in v direction
   Mantid::detid_t detID;  ///< Detector ID
-  Mantid::Kernel::V3D position;              ///< Detector position
-  Mantid::Kernel::Quat rotation;             ///< Detector orientation
-  boost::shared_ptr<const Mantid::Geometry::Object> shape; ///< Shape of the detector
-  Mantid::Kernel::V3D scaleFactor;           ///< Detector's scale factor
+  Mantid::Kernel::V3D position;  ///< Detector position
+  Mantid::Kernel::Quat rotation; ///< Detector orientation
+  boost::shared_ptr<const Mantid::Geometry::Object>
+      shape;                       ///< Shape of the detector
+  Mantid::Kernel::V3D scaleFactor; ///< Detector's scale factor
 };
 
 /**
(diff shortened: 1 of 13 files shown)


MantidQt/MantidWidgets/src/InstrumentView/RotationSurface.cpp
```diff
@@ -191,8 +192,6 @@ void RotationSurface::findUVBounds() {
   m_v_max = -DBL_MAX;
   for (size_t i = 0; i < m_unwrappedDetectors.size(); ++i) {
     const UnwrappedDetector &udet = m_unwrappedDetectors[i];
-    if (!udet.detector)
-      continue;
```
SimonHeybrock: Why exactly can we remove this?

MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/InstrumentView/UnwrappedSurface.h
```diff
@@ -44,17 +44,23 @@ class UnwrappedDetector {
...
   double height;          ///< detector height in units of v
   double uscale;          ///< scaling factor in u direction
   double vscale;          ///< scaling factor in v direction
-  Mantid::Geometry::IDetector_const_sptr detector;
+  Mantid::detid_t detID;  ///< Detector ID
+  Mantid::Kernel::V3D position;  ///< Detector position
+  Mantid::Kernel::Quat rotation; ///< Detector orientation
+  boost::shared_ptr<const Mantid::Geometry::Object>
+      shape;                       ///< Shape of the detector
+  Mantid::Kernel::V3D scaleFactor; ///< Detector's scale factor
```
SimonHeybrock: The values cached here may actually be different from what is found in the instrument? Or is this just a caching mechanism for speed?

`DetectorInfo` would eventually be a fast way for accessing these paramters, but currently it is not yet, so I think I agree that it makes sense having this.

SimonHeybrock (commented):
@mantid-roman Looks good, some comments below.

Generally, I think it would be beneficial to remove as many uses of `detid_t` as possible and use detector indices for direct access to `DetectorInfo`. This does not necessarily have to be part of this PR, but maybe you can keep it in mind for further refactoring. I think it would significantly reduce the juggling of index maps.

In that context it is worth pointing out that Instrument-2.0 will also have `ComponentInfo`, which is a bit like `DetectorInfo`, but will provide access to *all* components. In this case as well access will be via a plain index, so the interface would become more uniform.

MantidQt/MantidWidgets/src/InstrumentView/InstrumentActor.cpp
```diff
@@ -757,9 +761,9 @@ void InstrumentActor::setupPickColors() {
 void InstrumentActor::cacheDetPos() const {
   if (m_detPos.size() != m_detIDs.size()) {
     m_detPos.clear();
-    for (size_t i = 0; i < m_detIDs.size(); i++) {
-      IDetector_const_sptr det = this->getDetector(i);
-      m_detPos.push_back(det->getPos());
+    for (size_t pickID = 0; pickID < m_detIDs.size(); pickID++) {
```
mantid-roman: `m_detIDs` contains all detector ids. I can try to replace them with the indices.
`pickID`s are like detector indices but I am not sure at the moment if they can be easily replaced with them without significant restructuring of the code. It's one of the things I wanted to think about.

MantidQt/MantidWidgets/src/InstrumentView/InstrumentWidgetPickTab.cpp
```diff
@@ -876,7 +868,7 @@ QString ComponentInfoController::displayDetectorInfo(Mantid::detid_t detid) {
       wsIndex = "None";
     }
     text += "Workspace index: " + wsIndex + '\n';
-    Mantid::Kernel::V3D pos = det->getPos();
+    Mantid::Kernel::V3D pos = det.getPos();
```
mantid-roman: OK. I'll try to use `position`.

MantidQt/MantidWidgets/src/InstrumentView/PanelsSurface.cpp
```diff
@@ -231,9 +219,8 @@ void PanelsSurface::addFlatBank(ComponentID bankId,
     m_unwrappedDetectors.reserve(m_unwrappedDetectors.size() + nelem);
     for (int i = 0; i < nelem; ++i) {
       // setup detector info
-      Mantid::Geometry::IDetector_const_sptr det =
-          boost::dynamic_pointer_cast<const Mantid::Geometry::IDetector>(
-              assembly->getChild(i));
+      auto det = boost::dynamic_pointer_cast<const Mantid::Geometry::IDetector>(
+          assembly->getChild(i));
       if (!doneRotation) {
         pos0 = det->getPos();
```
OwenArnold: I understand that you are iterating over a bank here, but we should not get positions or rotations from the detector directly. Maybe we could extract the `detector ids` from the detector, create a list of these and then use `DetectorInfo::indexOf` to find the index to get the position for? Maybe this would be a suitable half-way solution until we have some kind of `ComponentInfo`. @SimonHeybrock what do you think?

mantid-roman: Sound complicated. Should we wait until we have all `Infos` in place?

MantidQt/MantidWidgets/src/InstrumentView/InstrumentActor.cpp
mantid-roman: On the other hand `pickID`s should probably distinguish detectors at different time steps, so they cannot be identical to the detector indices.

SimonHeybrock: Good point about time steps. I guess there are two options: (1) keep `pickID` as a linear index that can then map to detectors with time index (a `std::pair` of detector index and time index, see class `SpectrumDefinition`), or (2) eventually add a time index to `pickID`.

I do not like option (2) since it is exposing something about the instrument that does not need to be known at this level. Maybe, for hiding the fact that there are time indices, would it make sense to use `SpectrumInfo`? Such a `SpectrumInfo` could be set up with a 1:1 map from spectrum to `pair(detector index, time index)` (i.e., this could be different from a mapping in a workspace). We would then have a plain linear index, without the need to mess around with time indices.

MantidQt/MantidWidgets/src/InstrumentView/PanelsSurface.cpp
SimonHeybrock: @OwenArnold I think @mantid-roman has a point that that would be too complicated. On the other hand the other `Infos` will probably not be available for many months.

However, I actually need a very similar functionality in my changes to `DetectorInfo` that are updating the positions of detectors when a parent is moved. Currently I am using `Instrument::getDetectorInBank`, and then obtain detectors IDs, which I then translate to indices. This is slow and painful.
Maybe there is a way to add `ComponentInfo` without any functionality for accessing components, but with a mechanism that provides easy and fast access to a list of all detector indices in a subtree? Basically what we want is

```cpp
std::vector<size_t> API::ComponentInfo::detectorIndices(const IComponent &comp) {
  // find and return all detector indices in subtree given by `comp`
}

for (const auto &detIndex : ws.componentInfo().detectorIndices(comp)) {
  ws.detectorInfo().position(detIndex);
}
```

This requires some more thought, in particular how we work with time indices in an elegant (i.e., ideally hidden) way.

OwenArnold: Ok @SimonHeybrock maybe we should look at this as a separate item. I think we have some of this functionality in the prototype already.

SimonHeybrock: @OwenArnold Yes definitely separate.

SimonHeybrock:
I just notice that this PR is still around and is green now. Was there anything to be added here, are should be get this merged as it is?

SimonHeybrock:
I take that as a no. :shipit:

PR #18891, opened 2017-02-16
Fix bug in gd_prtn_chrg for chunked data

peterfpeterson:
There was a bug in the `gd_prtn_chrg` if the data was processed with chunking. Adding a call to `wksp.run().integrateProtonCharge()` before `NormaliseByCurrent` fixes the issue since it forces the workspace to re-integrate the proton charge based on the logs.

**To test:**

Process a dataset with chunking.

This does not have release notes because it the patch release needs to be setup and hasn't been yet.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/inc/MantidAlgorithms/NormaliseByCurrent.h
+++ b/Framework/Algorithms/inc/MantidAlgorithms/NormaliseByCurrent.h
@@ -71,8 +71,8 @@ private:
   void init() override;
   void exec() override;
   // Extract the charge value from the logs.
-  double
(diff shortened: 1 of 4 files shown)


wdzhou:
I did test by running an auto reduction script of a recent run.  The corrected code can generate the reduced data consistent with different max chunk size, while the previous code cannot.  So the bug is well resolved. 

It is based from release 3.9.  And the failure of Flake8 seems not to be caused by the change of this PR. 

:shipit:

rosswhitfield:
This has broken the performance tests http://builds.mantidproject.org/job/master_performancetests/3934/

PR #18907, opened 2017-02-17
Fix zeropadding for IMAT in Facilities.XML

dtasev:
A quick change that fixes being able to load IMAT workspaces through the data archive.

No related issue.

### DIFF
--- a/instrument/Facilities.xml
+++ b/instrument/Facilities.xml
@@ -78,6 +78,7 @@
 
   <instrument name="IMAT">
     <technique>Neutron Imaging</technique>
+    <zeropadding size="8"/>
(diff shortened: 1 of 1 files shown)


DavidFair:
Tested locally and it now correctly picks up IMAT filenames :shipit:

martyngigg:
@DavidFair Have their filenames always been padded to 8 digits or is it a recent change?

martyngigg:
I've just checked the runs in cycle 15_1 and they are all padded to 8 digits.

DavidFair:
Sorry - yes they have always had 8 digits as it is a (relatively) new instrument.

PR #18914, opened 2017-02-17
Fix mass ws deletion bug

### LINKED ISSUE #17729: Fix Mantid crash when closing many workspaces via MantidDock

NickDraper:
Fixes #17729

Added a DeleteWorkspaces algorithm, use it in GUI when deleting multiple workspaces, all it does is iterate over the list deleting workspaces.

**To test:**

*\* Test that deleting many workspaces does not crash**
1. Create many workspaces with:

``` python
for i in range(3000):
    name = "WS_" + str(i)
    CreateSampleWorkspace(OutputWorkspace=name)
```
1. Select all in the ADS
2. Press the delete button
   - Confirm that Mantid does not crash and that the time it takes to clear the 3000 workspaces is  acceptable.

*\* Test that deleting few very large data sets does not disturb the user experience (too much)**
1. Find a sample data set here: smb://olympic/babylon5/Scratch/Anton/SampleWS . The relevant file is MDEvent_ellipPeaks_new.nxs
1. If you have a 16GB computer run the follow script, else you have to modify the times you can load the file.

``` python
for i in range(10):
    name = "WS_" + str(i)
    Load(Filename='PATH_TO/MDEvent_ellipPeaks_new.nxs', OutputWorkspace=name)
```
1. Select all workspace and delete them.
   - Confirm that the time taken is acceptable and that the screen does not freeze. It will be unusable though.

### RELEASE NOTES

docs/source/release/framework.rst

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidPlot/src/Mantid/MantidUI.cpp
+++ b/MantidPlot/src/Mantid/MantidUI.cpp
@@ -459,7 +459,7 @@ void MantidUI::deleteWorkspaces(const QStringList &wsNames) {
       alg->setLogging(false);
       std::vector<std::string> vecWsNames;
       vecWsNames.reserve(wsNames.size());
-      foreach(auto wsName, wsNames) {
+      foreach (auto wsName, wsNames) {
         vecWsNames.push_back(wsName.toStdString());
       }
       alg->setProperty("WorkspaceList", vecWsNames);
--- a/MantidPlot/src/Mantid/MantidUI.cpp
+++ b/MantidPlot/src/Mantid/MantidUI.cpp
@@ -455,9 +455,15 @@ void MantidUI::deleteWorkspaces(const QStringList &wsNames) {
 
   try {
     if (!wsNames.isEmpty()) {
-      for (auto &ws : wsNames) {
-        deleteWorkspace(ws);
+      auto alg = createAlgorithm("DeleteWorkspaces");
+      alg->setLogging(false);
+      std::vector<std::string> vecWsNames;
+      vecWsNames.reserve(wsNames.size());
+      foreach(auto wsName, wsNames) {
+        vecWsNames.push_back(wsName.toStdString());
       }
+      alg->setProperty("WorkspaceList", vecWsNames);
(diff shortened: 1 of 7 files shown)


MantidPlot/src/Mantid/MantidUI.cpp
```diff
@@ -455,9 +455,15 @@ void MantidUI::deleteWorkspaces(const QStringList &wsNames) {
...
-        deleteWorkspace(ws);
+      auto alg = createAlgorithm("DeleteWorkspaces");
+      alg->setLogging(false);
+      std::vector<std::string> vecWsNames;
+      vecWsNames.reserve(wsNames.size());
+      foreach (auto wsName, wsNames) {
+        vecWsNames.push_back(wsName.toStdString());
       }
+      alg->setProperty("WorkspaceList", vecWsNames);
+      executeAlgorithmAsync(alg);
```
samueljackson92: Maybe we should put all this in a private method? `void MantidUI::deleteWorkspaces()` or something?

samueljackson92 (changes requested):
Tested with example scripts and played with tying to pass invalid input (e.g. no workspaces) to algorithm. All seems well. See comment below from code review but this is basically ready to go I think.

samueljackson92:
@NickDraper Also checked this with Giovanni's Mantid project and I cannot reproduce.

MantidPlot/src/Mantid/MantidUI.cpp
NickDraper: it already is!

samueljackson92: So it is!

samueljackson92 (approved):

samueljackson92:
LGTM :shipit:

Issue #15435, opened 2016-02-24
Super-WISH instrument definition


raquelalvarezbanos:
From Pascal:

> I think I talked before about getting a super-WISH instrument definition so we could see all the reflections that we are missing. Ideally we would be able to load the actual WISH data onto it but it would havbe much more out of plane coverage. 
> One feedback from my users is that it would be nice to give the angles between reflections selected by clicking. This would be the actual angles and also in plane and out of plane angles.

samueljackson92:
To further clarify this: this is basically just expanding the instrument definition file for WISH to include a fuller range of angles. Contact him for the specifics.

samueljackson92:
With the above two PRs merged this is now finished.

PR #18921, opened 2017-02-20
Groundwork for visualising predicted peaks

### LINKED ISSUE #18920: Ground work for extended detector space visulisation
This is a sub issue of #15435.

Pascal want's to be able to visualise predicted peaks that do not necessarily intersect with real detector bank in an instrument's geometry. But for this we need to know some measure of distance from the sample that a peak is likely to hit.

Planned changes:
 - Using an extra element in the IDF `extended-detector-space` we can define a simple shape roughly approximating the instrument. For example, WISH can be approximated as being roughly cylindrical.

 - Add code to `Peak` to compute approximate the detector distance from the extra IDF element.
 - Add an option to `PredictPeaks` to predict peaks that fall off detectors if checked.

samueljackson92:
This is a sub issue of #15435.

This adds the groundwork for visualising predicted peaks which fall outside of an instrument's detector geometry.

**To test:**

 - Check key tests pass
   - `PredictPeaks`
   - `PeakTest`

 - You can enable the new feature by adding an entry to an IDF similar to the following example for WISH:

```xml
   <!-- extended detector space definition for SCP projection -->
   <component type="extended-detector-space">
       <location /> 
   </component>

   <type name="extended-detector-space">
       <properties />
       <infinite-cylinder id="detector-surface">
           <centre x="0.0" y="0.0" z="0.0" />
           <axis x="0.0" y="1.0" z="0.0" /> 
           <radius val="2.2018" />
       </infinite-cylinder> 
       <algebra val="detector-surface" />
   </type>
```
 - Check that adding that definition and running `PredictPeaks` with the new option `PredictPeaksOutsideDetectors` will allow the prediction of peaks that fall outside of defined detectors and that these are in the correct place (relative to the component defined above).

 - These peaks should also show up in an unwrapped view on the instrument view, but proper visualisation has not yet been implemented and will be handled outside the scope of this ticket (see #18923).

 - Code review

Fixes #18920.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Crystal/src/PredictPeaks.cpp
```diff
@@ -101,6 +101,12 @@ void PredictPeaks::init() {
                   "values in the HKLPeaksWorkspace to the nearest integers if "
                   "checked.\n"
                   "Keep unchecked to use the original values");
+
+  declareProperty("PredictPeaksOutsideDetectors", false,
+                  "Use an extended detector space (if defined for the"
+                  " instrument) to predict peaks which do not fall onto any"
+                  "detector. This may produce a very high number of results.");
+
```
soininen: Will inserting the new property here change the algorithm's Python signature so that old scripts might stop working? I mean, somebody might have
```python
PredictPeaks(inputWS, 0.1, 100, 1, 100, Primitive, False, hklWS, True, False, 'outputWS')
```
in their scripts.

Framework/DataObjects/test/PeakTest.h
```diff
@@ -363,6 +375,124 @@ class PeakTest : public CxxTest::TestSuite {
...
+  void test_setQSampleFrameVirtualDetectorWithQLab() {
+    constexpr auto radius = 10.;
+    auto sphereInst =
+        ComponentCreationHelper::createTestInstrumentRectangular(5, 100);
+    auto extendedSpaceObj =
+        ComponentCreationHelper::createSphere(10., V3D(0, 0, 0));
+    auto extendedSpace = new ObjComponent("extended-detector-space",
+                                          extendedSpaceObj, sphereInst.get());
+    extendedSpace->setPos(V3D(0.0, 0.0, 0.0));
+    sphereInst->add(extendedSpace);
```
soininen: Would using a `unique_ptr` for and `release()` here help to see the intended transfer of ownership more easily?

soininen (changes requested):

Framework/Crystal/src/PredictPeaks.cpp
```diff
@@ -252,6 +258,13 @@ void PredictPeaks::exec() {
 
     size_t allowedPeakCount = 0;
 
+    bool useExtendedDetectorSpace = getProperty("PredictPeaksOutsideDetectors");
+    if (useExtendedDetectorSpace &&
+        !m_inst->getComponentByName("extended-detector-space")) {
+      g_log.warning() << "Attempting to find peaks outside of detectors but"
+                         "no extended detector space has been defined";
```
soininen: A space is missing between 'but' and 'no'

soininen:
When the new property is turned on, I get log messages like
```
Out of 39 allowed peaks within parameters, 39 were found to hit a detector.
```
when actually only 9 hit a real detector while the rest were on the extended detector space. Would outputting something like
```
Out of 39 allowed peaks within parameters, 9 were found to hit a detector and 30 the extended detector space.
```
be a bit more informative?

soininen:
I tested this with a Python script stolen from the algorithm's documentation:
```python
ws = CreateSimulationWorkspace(Instrument='WISH',
                               BinParams='0,10000,20000',
                               UnitX='TOF')

SetUB(ws, a=5.5, b=6.5, c=8.1, u='12,1,1', v='0,4,9')

cs = CrystalStructure('5.5 6.5 8.1', 'P b c a',
                      """Ni 0.232 0.114 0.543 1.0 0.00843;
                         Al 0.434 0.041 0.854 1.0 0.01120""")
ws.sample().setCrystalStructure(cs)

predicted = PredictPeaks(InputWorkspace=ws,
                         CalculateStructureFactors=True,
                         MinDSpacing=0.5,
                         PredictPeaksOutsideDetectors=True,
                         WavelengthMin=2.0, WavelengthMax=2.2)
```
However, when I try to visualize this using
```python
iView = getInstrumentView('ws')
iView.overlay('predicted')
iView.show()
```
I see no peaks. What am I doing wrong here?

BTW, calling `overlay('predicted')` on an open instrument window with *Full 3D* view results in `Aborted (core dumped)`.

samueljackson92:
@soininen Hi, I'm out of the office most if today. I agree with most of your comments and I'll address them soon. 

About the last point: are you defining an appropriate component in your IDF such as the example one provided for WISH in the description? No change to the IDF was made in this PR so you need to do it manually yourself.  Also peaks only show in a 2D view and not with event data (i remember we had a problem with MANDI data last time)

soininen:
I copy-pasted the XML-stuff from the PR description to the IDFs. After this, the warning
```
Attempting to find peaks outside of detectors butno extended detector space has been defined
```
did not show up. Also, the extended detector peaks  (detector ID -1) appeared in the output peak workspace.

In the instrument view, I can create new peaks manually and they show up in the 2D views. These peaks get added to the peaks workspace, but in contrary to the ones created by `PredictPeaks`, these have nonzero 'SigInt'. I don't know if that's the reason they get drawn and the predicted ones don't, though.

Framework/Crystal/src/PredictPeaks.cpp
samueljackson92: Yes you are right. It is easy to forget about ordering.

Framework/DataObjects/test/PeakTest.h
samueljackson92: Yes. I think I copy-pasted this from elsewhere and didn't update it to unique _ptr

samueljackson92:
> Also, the extended detector peaks (detector ID -1) appeared in the output peak workspace.

This is to be expected. A predicted peak that does not land on a detector by definition cannot have a detector associated with it. Ideally we'd refactor this code to better handle the requirement to be associated with a detector. I originally tried creating a new peak type inheriting from `IPeak` that wouldn't have a link to any detector, but found that lots of places seem to explicitly use `Peak` instead, meaning I'd have to change more code than the scope of this ticket should allow. If you have a better idea about how to handle this rather than just having it in an undefined state I would be interested.

> I don't know if that's the reason they get drawn and the predicted ones don't, though.

Just to check: did you drag the workspace output by `PredictPeaks` onto the 2D instrument view? They won't show unless you overlay. Also, depending on what unit cell/orientation you used you might not actually see anything unless the peaks are just outside the detectors in the 2D view. In #18936 I have added the ability to zoom out in a 2D view so you can actually see where things are.

soininen (approved):

Framework/Crystal/src/PredictPeaks.cpp
```diff
@@ -261,7 +261,7 @@ void PredictPeaks::exec() {
     bool useExtendedDetectorSpace = getProperty("PredictPeaksOutsideDetectors");
     if (useExtendedDetectorSpace &&
         !m_inst->getComponentByName("extended-detector-space")) {
-      g_log.warning() << "Attempting to find peaks outside of detectors but"
+      g_log.warning() << "Attempting to find peaks outside of detectors but "
                          "no extended detector space has been defined";
```
soininen: There is still `\n` missing at the end of this warning.

soininen (changes requested):

soininen:
I'm sorry, but I've touched the Peak stuff only briefly and thus cannot provide any opinion on that matter.

I didn't know about the drag n' drop functionality so it's nice to find out that it exists. I just used the `overlay()` Python call as described in my earlier comment. Seems that draggin and dropping gives the same result as Python, though. No peak constructed by `PredictPeaks` gets drawn in the InstrumentWindow, only the ones I add there manually. You can try it out yourself using my scripts above, if you wish.

samueljackson92:
@soininen We should make an issue for the `.overlay` method. I did not know that was an option!

I think the problem is that the TOF of the predicted peaks is much larger than the range you've set in your original script. If I change the first line to `ws = CreateSimulationWorkspace(Instrument='WISH',
                               BinParams='0,10000,300000',
                               UnitX='TOF')` the peaks get plotted. Also, your script only seems to generate peaks that fall on detectors, perhaps you can rotate the sample a little to make some of them deliberately fall off.

soininen (approved):

soininen:
I see, so it was the TOF range. Now the peaks show up in the instrument window both by dragging and dropping and using Python's `overlay()`.

samueljackson92:
@soininen Did you manage to produce a test that caused a peak to fall off of the detectors?

samueljackson92:
That's also interesting that the python command does not fail. Perhaps it is making some assumptions about the TOF range?

soininen:
@samueljackson92 The overlay worked only after I extended the TOF range in the script, as you suggested, regardless of whether I used python's `overlay()` or drag n' drop. I do see both the peaks that fall on the detectors, and peaks that don't.

soininen:
The algorithm predicts the peaks outside the detector, and basic visualization works. The changes look good to me, so :shipit:

PR #18943, opened 2017-02-21
Send error notifications for parent algorithms

### LINKED ISSUE #18820: A failed algorithm can't be canceled
### Expected behavior

Mantid should be able to cancel an algorithm that failed and it shouldn't hang if you try to close MantidPlot.

### Actual behavior

A failed algorithm can not be canceled. Then, MantidPlot hangs if you try to close it.

- Load `POLREF14882.nxs` which is a multi-period workspace
- Open `ConvertUnits` from the algorithm dialog. Select `POLREF14882` and set `Target=MomentumTransfer` and `AlignBins=True`.
- Try to cancel the algorithm from `Details`.

### Platforms affected

All? This seems to work on release 3.8.

gemmaguest:
This PR fixes a problem where error notifications are not sent from a parent algorithm if one of its child algorithms throws an exception. This causes a problem where the algorithm monitor still thinks the algorithm is running and it cannot be cancelled from the Details dialog. MantidPlot also hangs when you try to close it because it is waiting for the algorithm to finish.

This PR also adds some additional tests to check that the correct notifications are received when running algorithms on workspace groups.

**To test:**

Run the following script:
```
inWS=Load(Filename=r'POLREF00014882.NXS')
IvsQ, IvsQ_unbinned, IvsLam = ReflectometryReductionOneAuto(InputWorkspace=inWS)
```
MantidPlot should report a runtime error about theta being 0. Open the Details dialog - the algorithm should have finished and should not be shown in the Details dialog. You should be able to close MantidPlot without it hanging.

Fixes #18820 .

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/src/Algorithm.cpp
+++ b/Framework/API/src/Algorithm.cpp
@@ -1200,8 +1200,24 @@ bool Algorithm::doCallProcessGroups(Mantid::Kernel::DateAndTime &startTime) {
   startTime = Mantid::Kernel::DateAndTime::getCurrentTime();
   // Start a timer
   Timer timer;
-  // Call the concrete algorithm's processGroups method
-  const bool completed = processGroups();
+
+  bool completed = false;
+  try {
+    // Call the concrete algorithm's processGroups method
+    completed = processGroups();
+  } catch (std::exception &ex) {
+    // The child algorithm will already have logged the error etc.,
+    // but we also need to update flags in the parent algorithm and
+    // send an ErrorNotification (because the child isn't registered with the
+    // AlgorithmMonitor).
+    setExecuted(false);
+    m_runningAsync = false;
+    m_running = false;
+    notificationCenter().postNotification(
+        new ErrorNotification(this, ex.what()));
+    throw;
+  }
+
   // Check for a cancellation request in case the concrete algorithm doesn't
   interruption_point();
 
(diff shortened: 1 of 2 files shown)


martyngigg (changes requested):
This is good and does fix the problem by ensuring that some kind of notification does get send in the case or success or failure.

I think, just to be on the safe side, we should also add a `catch(...)` block that also sends an error notification much like the final catch all in the [main exec](https://github.com/mantidproject/mantid/blob/084fcde2d28027aa19850ce2afa67d903683a2f4/Framework/API/src/Algorithm.cpp#L619). It's quite possibly someone could define a custom exception type that doesn't derive from std::exception.

martyngigg (changes requested):
The changes that related to the issue are now good.

I was nearly going to mark it for merging but then I notced a piece of [bad practice]( https://github.com/mantidproject/mantid/pull/18943/files#diff-09ae871bda8c189fdc40fb4a954f03aaR18) in the existing code (this has been there for quite some time). We shouldn't use `#define` in this manner as it subverts the type system. Can we replace `#define NoofLoops 10` with something a bit more modern:

```
namespace {
   constexpr int NO_OF_LOOPS = 10;
}
```

It will hopefully avoid copying bad practice around too.

martyngigg (approved):

martyngigg:
LGTM :shipit:

PR #18906, opened 2017-02-17
Store positions and rotations of detectors in DetectorInfo

SimonHeybrock:
This PR moves position and rotation information of detectors into `DetectorInfo`. This is a major part of changes for Instrument-2.0.

- Position/rotation changes are no longer stored in the `ParameterMap` (except for non-detector components).
- Positions/rotations of detectors are updated when a parent rotation is changed. Thus, moving components is now more costly than it used to be. In turn, getting positions of detectors is now much cheaper.
- Files are kept compatible by creating a "legacy-style" `ParameterMap` including position and rotation updates for saving.
- Extraction of position/rotation information is done initially when loading an instrument (from IDF or from a parameter string of a processed Nexus file). As a consequence, loading is now more expensive.
- `CompareWorkspaces` now uses `DetectorInfo::isEquivalent` when comparing instruments. Since the internals of position and rotation math have changed slight differences are possible. The tolerance has been set to 1e-9m.

Minor fixed/changes:
- Update cached L1 in `DetectorInfo` when components are moved or rotated.
- Don't use `twoTheta` for monitors in `LoadDspacemap`.
- `MoveInstrumentComponent` now ignores changes for `RectangularDetectorPixel` and prints a warning. Previously parameters where added to the `ParameterMap` but took no effect, since `RectangularDetectorPixel::getRelativePos()` ignores parameters in the map.

This PR may have major impact on the performance of some algorithm, both positive and negative:

- There is currently extra cost when loading and saving files, for conversion from and to the old way of storing the instrument and its parameters (note that there is no new way yet).
- Moving/rotating components, in particular large banks is now more expensive, since the position of all contained pixels are updated. In turn, reading positions/rotations is cheaper.
- I had identified one candidate with major issues, `SCDCalibratePanels`. After some optimizations it seems to be mostly back to its old speed. The unit test (not a performance test) is a fair bit slower, but a more realistic example that can be found [here](https://github.com/mantidproject/mantid/pull/17153) runs in roughly the same time on my system (rather beefy 12 core system).

Note a rather big change to the deltas in the asserts in `SCDCalibratePanels` test. As experienced recently in another PR, minor differences in floating point math seem to be yielding massive differences in the resulting fit parameters. It is hard to prove that this is the issue, I tried to verify the resulting parameters instead. For the example linked above I get for `master`:

```
# 2017-03-08T10:42:37.660479000
6         L1     T0_SHIFT
7  1800.0339        0.000
[...]

PR #18945, opened 2017-02-21
Functional quaternions

OwenArnold:
As a follow on from @SimonHeybrock's #18780. `Quaternions` have **mutable/stateful** behaviour where they really ought to have **functional** behaviour. 

Look at the current `Quat`. This makes coding more error prone. Consider taking the `inverse` for example - That modifies the original! This is logically incorrect. In future we will very likely start using `Eigen` so this small PR should be a helpful step in that direction.

Methods changed to be `const` on `Quat` are:

* `inverse()`
* `normalize()`
* `conjugate()`

All return a new `Quat`

WIP. **DO NOT MERGE**. Cannot be considered for merging until #18780 is merged.

**TESTER**. A code review and verification that all tests still pass will be sufficient. This is an under-the-hood change, so no need to update release notes.


OwenArnold:
**BEWARE** changes as part of unmerged #18906 to `DetectorInfo::setRotation` which would become logically incorrect, but would still compile with these changes.

SimonHeybrock:
@OwenArnold `Eigen::Quaterniond::normalize()` normalizes the quaterniond. `normalized()` returns a normalized variant, so this is inconsistent.

I am not sure how beneficial changing this now is. We are risking of breaking things, and have to learn two new interfaces instead of 1. Or was your intention to make the interfaces 100% equivalent, such that we can later just use search & replace to use `Eigen`?

OwenArnold:
@SimonHeybrock I did have the forthcoming Eigen replacement in mind. I think this step would make that easier. Good point on `Quaterniond::normalize` I'd need to change that back. I don't think these changes are necessary to merge, but the stateful behaviour of `Quat` is wrong for a number of it's methods such as `inverse`, and this makes the current implementation ugly and error prone. Shall we just shelve this PR for now?

SimonHeybrock:
@OwenArnold I think that might be a good idea. We can have another look once we have a clearer idea of how (and when) a rollout of `Eigen` could happen (I suspect this will not be in the near future).

OwenArnold:
Eigen will replace this functionality eventually.

PR #18952, opened 2017-02-21
Fix an issue in `TimeSeriesProperty`

### LINKED ISSUE #18944: Fix crash with OFFSPEC data
Mantid crashes with some `OFFSPEC` runs

### Expected behavior

It shouldn't crash

### Actual behavior

It crashes

Run this script:

```
ws_1 = LoadISISNexus(Filename=r'OFFSPEC00038084.nxs')
ws_2 = LoadISISNexus(Filename=r'OFFSPEC00038087.nxs')
dbp = Plus(LHSWorkspace='ws_1', RHSWorkspace='ws_2')
```

### Platforms affected

Windows, debug mode

raquelalvarezbanos:
Description of work.

**To test:**

- This has to be tested on Windows, debug mode.
- Run the script in the issue description, Mantid shouldn't crash.
- Make sure that the new unit test fails without this fix.
- I could have added a similar check for the property values before [this](https://github.com/mantidproject/mantid/blob/291847a5e65364b6933fcf325f91d15932b99f4c/Framework/Kernel/src/TimeSeriesProperty.cpp#L191) line, but I think it is unnecessary. Make sure you agree with me.

Fixes #18944.

No need to mention this in the release notes, as the bug only happens on Windows in debug mode.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Kernel/src/TimeSeriesProperty.cpp
+++ b/Framework/Kernel/src/TimeSeriesProperty.cpp
@@ -176,6 +176,9 @@ operator==(const TimeSeriesProperty<TYPE> &right) const {
   { // so vectors can go out of scope
     std::vector<DateAndTime> lhsTimes = this->timesAsVector();
     std::vector<DateAndTime> rhsTimes = right.timesAsVector();
+    if (lhsTimes.size() != rhsTimes.size()) {
+      return false;
+    }
     if (!std::equal(lhsTimes.begin(), lhsTimes.end(), rhsTimes.begin())) {
       return false;
     }
--- a/Framework/Kernel/src/TimeSeriesProperty.cpp
+++ b/Framework/Kernel/src/TimeSeriesProperty.cpp
@@ -175,20 +175,15 @@ operator==(const TimeSeriesProperty<TYPE> &right) const {
 
   if (this->realSize() != right.realSize()) {
     return false;
-  }
-
-  { // so vectors can go out of scope
-    std::vector<DateAndTime> lhsTimes = this->timesAsVector();
-    std::vector<DateAndTime> rhsTimes = right.timesAsVector();
+  } else {
+    const std::vector<DateAndTime> lhsTimes = this->timesAsVector();
+    const std::vector<DateAndTime> rhsTimes = right.timesAsVector();
     if (!std::equal(lhsTimes.begin(), lhsTimes.end(), rhsTimes.begin())) {
       return false;
     }
-  }
 
-  {
-    // so vectors can go out of scope
-    std::vector<TYPE> lhsValues = this->valuesAsVector();
-    std::vector<TYPE> rhsValues = right.valuesAsVector();
+    const std::vector<TYPE> lhsValues = this->valuesAsVector();
+    const std::vector<TYPE> rhsValues = right.valuesAsVector();
     if (!std::equal(lhsValues.begin(), lhsValues.end(), rhsValues.begin())) {
       return false;
     }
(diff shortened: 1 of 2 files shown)


Framework/Kernel/src/TimeSeriesProperty.cpp
```diff
@@ -176,6 +176,9 @@ operator==(const TimeSeriesProperty<TYPE> &right) const {
   { // so vectors can go out of scope
     std::vector<DateAndTime> lhsTimes = this->timesAsVector();
     std::vector<DateAndTime> rhsTimes = right.timesAsVector();
+    if (lhsTimes.size() != rhsTimes.size()) {
```
peterfpeterson: Shouldn't you use  `TimeSeriesProperty::size()` or `TimeSeriesProperty::realSize()` and copy the vectors if they aren't equal?

peterfpeterson (changes requested):

Framework/Kernel/src/TimeSeriesProperty.cpp
DimitarTasev: :+1: maybe extend the `if (this->m_size != right.m_size) ` above with an ` || this->realSize() != right.realSize()`, then the check will be applied for both below

peterfpeterson: The greatest effect would be had by putting this as an `if` on line 176.

DimitarTasev: @peterfpeterson Could you tell me why an additional `if` is preferred to extending an existing one?

DimitarTasev:
I've tested without the fix works and mantid longer crashes. After applying the fix there is no crash.
I've ran the unit tests without the fix and the new test fails. After applying the fix it passes successfully.
If @peterfpeterson doesn't have any comments I'm happy to :shipit:

Framework/Kernel/src/TimeSeriesProperty.cpp
raquelalvarezbanos: Shouldn't they be the same?

Framework/Kernel/src/TimeSeriesProperty.cpp
```diff
@@ -173,12 +173,13 @@ operator==(const TimeSeriesProperty<TYPE> &right) const {
     return false;
   }
 
+  if (this->realSize() != right.realSize()) {
+    return false;
+  }
+
```
peterfpeterson: add the word `else` here and I'm happy

peterfpeterson:
I think a clearer version of the code (with added `const` below) is:
```c
if (this->realSize() != right.realSize()) {
     return false;
} else {
     const std::vector<DateAndTime> lhsTimes = this->timesAsVector();
      const std::vector<DateAndTime> rhsTimes = right.timesAsVector();
     if (!std::equal(lhsTimes.begin(), lhsTimes.end(), rhsTimes.begin())) {
       return false;
     }

     const std::vector<TYPE> lhsValues = this->valuesAsVector();
     const std::vector<TYPE> rhsValues = right.valuesAsVector();
     if (!std::equal(lhsValues.begin(), lhsValues.end(), rhsValues.begin())) {
       return false;
     }
}

return true;
```
I don't think the `const` buys anything since those functions return copies of the vectors. Without checking against a compiler, you can probably declare the various variables as `const auto` as well, but I don't think it really adds.

peterfpeterson (approved):

martyngigg:
Tests are the usual unrelated failures.

PR #18955, opened 2017-02-21
Fix crash in MonitorDlg

### LINKED ISSUE #18951: Monitor dialog crashes when clicking Details button
### Expected behavior
The monitor dialog should not cause mantid to crash.

### Actual behavior
There is an intermittent crash when clicking on the Details button to open the monitor dialog.

The crash is intermittent and not easily reproducible. It happens quite often for offspec users when using live data.

### Platforms affected
Windows 10, possibly all.

gemmaguest:
This PR avoids a potential crash when clicking the Details button to open the monitor dialog. It does not solve the root cause of the problem (an algorithm pointer being null) but stops mantid crashing if a null pointer is found.

**To test:**

The crash is intermittent and is not easily reproducible. The problem has been seen frequently by offspec users and we have managed to reproduce it under the debugger on their machine. They are testing a dev version with the same fix as this PR and so far have not seen the crash again.

Fixes #18951 .

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidPlot/src/Mantid/AlgorithmMonitor.cpp
+++ b/MantidPlot/src/Mantid/AlgorithmMonitor.cpp
@@ -206,6 +206,9 @@ void MonitorDlg::update() {
        itr != iend; ++itr) {
     IAlgorithm_sptr alg =
         Mantid::API::AlgorithmManager::Instance().getAlgorithm(*itr);
+    if (!alg) {
+      continue;
+    }
     // m_algorithms << alg;
     QStringList iList;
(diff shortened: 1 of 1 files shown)


martyngigg (approved):
Adds a preventative check for a null pointer. This is actually required as the `AlgorithmManager` can in fact return a null pointer if the algorithm is not found it's managed list which could happen if the algorithm being queried has been knocked off the list.

:shipit:

PR #18875, opened 2017-02-15
U correction not correctly applied to viewport

samueljackson92:
This correctly applies the U correction to the viewport when viewing a rotation surface in the instrument view.

**To test:**
 - Follow issue steps to reproduce bug. Apply fix and retest to confirm it is resolved.
 - Code review

Fixes #18869.

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [x] Is the code of an acceptable quality?
- [x] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)? Is it well structured with small focussed classes/methods/functions?
- [ ] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [x] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant documentation been added/updated?
- [ ] Is user-facing documentation written in a user-friendly manner?
- [ ] Has developer documentation been updated if required?

- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

PR #18959, opened 2017-02-21
Blank instrument view with U correction

### LINKED ISSUE #18869: U correction makes half of WISH's detectors disappear
Pascal has noted that this is broken. On further investigation it appears it was also broken in 3.8.

### Expected behavior
Applying a U correction of `min = 0, max = 6.28` should cause the front & backscattering detectors to be flipped.

### Actual behavior
Applying a U correction of `min = 0, max = 6.28` causes half of WISH's detectors to disappear.

 - Load WISH file
 - Open instrument view
 - Display settings -> U correction
 - Enter values `min = 0, max = 6.28`
 - Apply
 - Observe bug

### Platforms affected
All

samueljackson92:
It appears the start up the instrument view doesn't draw the correct viewport unless a U correction is applied. This appears to have been missed in the original PR, probably because the U correction gets cached in the QSettings.

**To test:**
Follow the same instructions as in PR #18875.  

Fixes #18869 .

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidQt/MantidWidgets/src/InstrumentView/RotationSurface.cpp
+++ b/MantidQt/MantidWidgets/src/InstrumentView/RotationSurface.cpp
@@ -162,7 +162,7 @@ void RotationSurface::init() {
                             udet.u = applyUCorrection(udet.u);
                           }
                         }
-    updateViewRectForUCorrection();
+                        updateViewRectForUCorrection();
 }
 
 /** Update the view rect to account for the U correction
--- a/MantidQt/MantidWidgets/src/InstrumentView/RotationSurface.cpp
+++ b/MantidQt/MantidWidgets/src/InstrumentView/RotationSurface.cpp
@@ -162,6 +162,7 @@ void RotationSurface::init() {
                             udet.u = applyUCorrection(udet.u);
                           }
                         }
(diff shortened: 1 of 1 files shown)


MantidQt/MantidWidgets/src/InstrumentView/RotationSurface.cpp
```diff
@@ -162,6 +162,7 @@ void RotationSurface::init() {
                             udet.u = applyUCorrection(udet.u);
                           }
                         }
+    updateViewRectForUCorrection();
```
AndreiSavici: Does the text need to be indented so much from line 89 to 164

peterfpeterson:
Works for me.

MantidQt/MantidWidgets/src/InstrumentView/RotationSurface.cpp
samueljackson92: No, I tried to fix this in the previous PR but the code has detector info changes which I don't think we want in the patch release so I took that commit out.

AndreiSavici:
:shipit:

PR #18962, opened 2017-02-21
Expose SpectrumInfo functionality to python

### LINKED ISSUE #18961: Expose SpectrumInfo functions to python
Currently on `SpectrumInfo` only `isMonitor`, `isMasked`, and `hasDetectors` functions are exposed to python. For geometry calculations, `l2`, `twoTheta`, and `position` should be exposed to python as well. In addition, SpectrumInfo should have a function to retun `phi` as well.

AndreiSavici:
**To test:**
Load a file, get the `spectrumInfo` object something like `si=ws.getSpectrumInfo()`.
Check if `position`, `l2`, `twoTheta`, and `phi` functions return sensible values. Remember that angles are in radians.

Fixes #18961.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/src/SpectrumInfo.cpp
+++ b/Framework/API/src/SpectrumInfo.cpp
@@ -137,7 +137,7 @@ double SpectrumInfo::phi(const size_t index) const {
   if (isMonitor(index))
     throw std::logic_error("phi is not defined for monitors.");
   Kernel::V3D pos = position(index);
-  return std::atan2(pos[1],pos[0]);
+  return std::atan2(pos[1], pos[0]);
 }
 
 /// Returns true if the spectrum is associated with detectors in the instrument.
--- a/Framework/API/src/SpectrumInfo.cpp
+++ b/Framework/API/src/SpectrumInfo.cpp
@@ -137,9 +137,7 @@ double SpectrumInfo::phi(const size_t index) const {
   if (isMonitor(index))
     throw std::logic_error("phi is not defined for monitors.");
   Kernel::V3D pos = position(index);
-  double p, dummy;
-  pos.getSpherical(dummy, dummy, p);
-  return p * M_PI / 180.;
+  return std::atan2(pos[1],pos[0]);
 }
 
 /// Returns true if the spectrum is associated with detectors in the instrument.
--- a/Framework/API/src/SpectrumInfo.cpp
+++ b/Framework/API/src/SpectrumInfo.cpp
@@ -132,6 +132,18 @@ Kernel::V3D SpectrumInfo::position(const size_t index) const {
   return newPos / static_cast<double>(dets.size());
 }
 
+/// Returns the phi angle (in radians) for the spectrum with given index
+double SpectrumInfo::phi(const size_t index) const {
+  if (isMonitor(index))
+    throw std::logic_error(
+        "phi is not defined for monitors.");
+  Kernel::V3D pos = position(index);
+  double p, dummy;
+  pos.getSpherical(dummy, dummy, p);
+  return p * M_PI / 180.;
+}
+
+
 /// Returns true if the spectrum is associated with detectors in the instrument.
 bool SpectrumInfo::hasDetectors(const size_t index) const {
   // Workspaces can contain invalid detector IDs. Those IDs will be silently
--- a/Framework/API/src/SpectrumInfo.cpp
+++ b/Framework/API/src/SpectrumInfo.cpp
@@ -135,14 +135,12 @@ Kernel::V3D SpectrumInfo::position(const size_t index) const {
 /// Returns the phi angle (in radians) for the spectrum with given index
 double SpectrumInfo::phi(const size_t index) const {
   if (isMonitor(index))
-    throw std::logic_error(
-        "phi is not defined for monitors.");
+    throw std::logic_error("phi is not defined for monitors.");
   Kernel::V3D pos = position(index);
   double p, dummy;
   pos.getSpherical(dummy, dummy, p);
   return p * M_PI / 180.;
 }
-
 
 /// Returns true if the spectrum is associated with detectors in the instrument.
 bool SpectrumInfo::hasDetectors(const size_t index) const {
--- a/Framework/PythonInterface/mantid/api/src/Exports/SpectrumInfo.cpp
+++ b/Framework/PythonInterface/mantid/api/src/Exports/SpectrumInfo.cpp
@@ -14,5 +14,14 @@ void export_SpectrumInfo() {
            "masked.")
       .def("hasDetectors", &SpectrumInfo::hasDetectors, (arg("self")),
            "Returns true if the spectrum is associated with detectors in the "
-           "instrument.");
+           "instrument.")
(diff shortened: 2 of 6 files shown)


Framework/API/src/SpectrumInfo.cpp
```diff
@@ -132,6 +132,16 @@ Kernel::V3D SpectrumInfo::position(const size_t index) const {
...
 }
 
+/// Returns the phi angle (in radians) for the spectrum with given index
+double SpectrumInfo::phi(const size_t index) const {
+  if (isMonitor(index))
+    throw std::logic_error("phi is not defined for monitors.");
+  Kernel::V3D pos = position(index);
+  double p, dummy;
+  pos.getSpherical(dummy, dummy, p);
+  return p * M_PI / 180.;
```
quantumsteve: Would it be better to replace lines 140-142 with `return atan2(pos[1], pos[0]);`?

Framework/API/src/SpectrumInfo.cpp
```diff
@@ -132,6 +132,16 @@ Kernel::V3D SpectrumInfo::position(const size_t index) const {
   return newPos / static_cast<double>(dets.size());
 }
 
+/// Returns the phi angle (in radians) for the spectrum with given index
+double SpectrumInfo::phi(const size_t index) const {
+  if (isMonitor(index))
+    throw std::logic_error("phi is not defined for monitors.");
+  Kernel::V3D pos = position(index);
+  double p, dummy;
+  pos.getSpherical(dummy, dummy, p);
```
quantumsteve: I'm debating whether using the same dummy arguments for `R` and `theta` is genius (avoids calculating theta) or just evil aliasing.

quantumsteve (changes requested):
Just a few thoughts on the implementation of `SpectrumInfo::phi()`. In this case is the additional encapsulation worthwhile?

quantumsteve (approved):

quantumsteve:
LGTM. :shipit:

Framework/API/src/SpectrumInfo.cpp
```diff
@@ -132,6 +132,14 @@ Kernel::V3D SpectrumInfo::position(const size_t index) const {
   return newPos / static_cast<double>(dets.size());
 }
 
+/// Returns the phi angle (in radians) for the spectrum with given index
+double SpectrumInfo::phi(const size_t index) const {
```
SimonHeybrock: This method is deliberately not part of `SpectrumInfo`. The problem is that its naming and functionality is ambiguous, confusing, and misleading. `Detector::getPhi` returns the spherical component (angle around Z), but there are several algorithms the interpret it as the angle around the beam direction. For many instruments this is ok, since the beam is aligned with Z, but in general it is not. Those algorithm will thus give wrong results.

We are considering to later add a method to obtain the angle around the beam direction but it will definitely not be named `phi()`.

Please remove this method.

Framework/PythonInterface/mantid/api/src/Exports/SpectrumInfo.cpp
```diff
@@ -14,5 +14,15 @@ void export_SpectrumInfo() {
...
+      .def("position", &SpectrumInfo::position, (arg("self"), arg("index")),
+           "Returns the position of the detector(s) associated with the "
+           "spectrum.")
+      .def("l2", &SpectrumInfo::l2, (arg("self"), arg("index")),
+           "Returns the distance from the sample of the detector(s) "
+           "associated with the spectrum.")
+      .def("phi", &SpectrumInfo::phi, (arg("self"), arg("index")),
+           "Returns phi angle of the detector(s) associated with the spectrum.")
+      .def("twoTheta", &SpectrumInfo::twoTheta, (arg("self"), arg("index")),
+           "Returns twoTheta of the detector(s) associated with the spectrum.");
```
SimonHeybrock: We are considering renaming this method, since there is some risk of confusing with the `theta` in sperical coordinates. I this did not expose it to Python, since it may be removed, and it would not make sense to deprecate functions even before the interface is complete.

docs/source/release/v3.10.0/framework.rst
```diff
@@ -61,6 +61,7 @@ Python
       3.0
       3.0
 
+- Exposed more `SpectrumInfo` functionality to Python.
```
SimonHeybrock: We have not "officially" release `SpectrumInfo` to Python (it was not included in previous release notes and there is no documentation apart form very basic Doxygen), so this line does not make sense here.

Framework/PythonInterface/mantid/api/src/Exports/SpectrumInfo.cpp
```diff
@@ -14,5 +14,15 @@ void export_SpectrumInfo() {
            "masked.")
       .def("hasDetectors", &SpectrumInfo::hasDetectors, (arg("self")),
            "Returns true if the spectrum is associated with detectors in the "
-           "instrument.");
+           "instrument.")
+      .def("position", &SpectrumInfo::position, (arg("self"), arg("index")),
+           "Returns the position of the detector(s) associated with the "
+           "spectrum.")
```
SimonHeybrock: We will probably change the return type of this from `Kernel::V3D` to `Eigen::Vector3d`, so I had not exposed this to Python for now. Probably ok, but worth considering to wait more before making it available.

SimonHeybrock (commented):
Unfortunately you were a bit too fast with merging, please see below for several comments:

AndreiSavici:
@SimonHeybrock I agree with some of your comments. `phi` and `twoTheta` are probably not the best names. But since `twoTheta` is one that is implemented, that's what I expose. `phi` is what we have as legacy, including in `V3D` class. I am more than happy to do something different. The return type of `position` is fine as `Eigen::Vector3d` as long as we have the python bindings for that. As it is mentioned, `SpectrumInfo` is the way forward, and other ways will be deprecated, so there is no point in doing something different. **We need this functionality in Python**. I had users requesting them. If  you have better solution to get this information from within Python, I am willing to make the changes.

SimonHeybrock:
@AndreiSavici Sure, just use the old interface, like all other Python code. The old interface is 100% functional.

It does not make sense to provide a Python interface to a class that is just in development and has an incomplete interface.

Once `SpectrumInfo` has settled to a stable state we will provide a Python interface, update the release notes and documentation, and notify developers that it is ready for use in Python.

AndreiSavici:
The old interface knows how to deal with one detector, not with a detector group. Unfortunately, grouping detectors is the default way of working at ISIS. The MantidPlot GUI computes these values for the detector group, same as `SpectrumInfo`. So I would need to write some code in Mantid framework to do all the same thing that `SpectrumInfo` already does, with the knowledge that is going to be deprecated anyway. It seems like a waste of time

SimonHeybrock:
@AndreiSavici `SpectrumInfo` provides nothing that you cannot do with the old interface, it is just a wrapper class and does not add any functionality. If you want to work with `DetectorGroup`, just use `MatrixWorkspace::getDetector`.

Issue #18963, opened 2017-02-22
RemoveBins equivalent for Point Data workspaces


NickDraper:
From jameslord:

The RemoveBins algorithm (and its partner MaskBins) only work on histogram data. An equivalent RemovePoints for point data would be useful. It could simply delete the unwanted bins from the middle of the spectra, with no zeroing or interpolation, as there are no left over bin edges to worry about. Or could RemoveBins be expanded to have that behaviour if presented with point data?
A likely use is if I do a sequential fit in Muon Analysis, create a results table, extract a fit parameter column versus a log value as a Workspace (for subsequent fitting) and notice there are some "bad" data points. If I'd done the fitting as individual fits then I could un-tick a few of them and recreate the results table but there's no equivalent editing possible for a sequential fit.

NickDraper:
We plan to make RemoveBins work for point data as well as Histogram data

PR #18970, opened 2017-02-22
Neatenup RemoveBins and add tests

NickDraper:
1. Added a missing test for RangeUnits
1. Consolidated cross parameter validation
1. Fixed a broken check for when units are the same
1. Only require the ws to have a unit when RangeUnits is set
1. Removed some units from the RangeUnits list as they made no sense

*No changes to how the guts of how removebins works!*

**To test:**

1. Test the validation of RemoveBins
 1. Xmax < Xmin
 1. WorkspaceIndex out of range
 1. Removals using RangeBins
 1. Removals using a workspace without units (use CreateWorkspace) with RangeUnits set to AsInput
 1. RangeUnits no longer contains the following entries ("Empty", "Label", "Time", "Degrees")

Re #18963.

### RELEASE NOTES

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


DavidFair:
I've found some issues with the validation - 

- Accepts negative numbers for `XMin` and `XMax`
- Using a sample workspace without units (`AsInput`) and all other settings default. Then setting the RangeUnit to something like `dSpacing` gives an incorrect error message about converting from TOF
- If you set `XMax` to a high value (say 999999) the log incorrectly states `XMin` is greater than the largest value

DavidFair:
@NickDraper Is this for 3.10 release or the next maintenance cycle?

NickDraper:
The improvements here have not changed the previous validation, just added missing tests, and moved the previous validation code to the standard validateInputs location, so it can be checked before exec.

Moved to next release as this is only a minor benefit, and not worth any risk at this point in the release.

PR #18983, opened 2017-02-23
Add histogram specific integration ranges to Integration

### LINKED ISSUE #18657: New algorithm: IntegrateElasticPeaks
The direct geometry time-of-flight reduction workflow at ILL would benefit from an algorithm which integrates the spectra around the elastic peak positions specified by the table produced by [`FindEPP`](http://docs.mantidproject.org/nightly/algorithms/FindEPP-v1.html).

soininen:
This PR adds two new properties to the [`Integration`](http://docs.mantidproject.org/nightly/algorithms/Integration-v1.html) algorithm: *RangeLowerList* and *RangeUpperList* can be used to give integration limits to each spectrum individually. This will help, for example, the data reduction at ILL on spectrometers where the L2 distance and thus the elastic peak position in TOF axis is different for each detector.

**To test:**

Try out different combinations of parameters, also *IncludePartialBins*. Test that there are no hidden surprises and that the algorithm fails gracefully if upper range is smaller than the lower range. Check that the X values correspond to the integration limits. Try out ragged workspaces, if possible.

Fixes #18657.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


KarlPalmen:
I've found a problem using the upper ranges on a cropped MUSR workspace (with 5 spectra). It assumes the blank lower ranges are a very high number, warns about this and sets the integration to zero.

KarlPalmen:
Also Mantid crashes when I attempt to copy this warning message.

soininen:
@KarlPalmen Could you sent me (or attach here) the failing workspace and perhaps a python script to reproduce the problem?

KarlPalmen:
Load `MUSR00022725.nxs` from the unit test data. Crop the workspace to 5 spectra (e.g. 11-15). Run integration on this cropped workspace with five upper range values such as `6,7,8,9,10` and leave other optional properties as default.

KarlPalmen:
Also I ran it on Windows 7.

soininen:
Nice catch! The very high number in the blank lower range is actually `EMPTY_DBL` and when compared to whatever upper limit you might have, the lower limit is always greater. This behavior actually originates from before these changes: if you gave only *RangeUpper*, the algorithm would always integrate from [xmin, 0] regardless of given upper limit. The warning message wasn't very helpful either:
`Range_upper is less than Range_lower. Will integrate up to frame maximum.`

I'll fix this such that if no lower limit is given (either in *RangeLower* or *RangeLowerList*), the integration is done from xmin to the upper limit or xmax.

KarlPalmen:
When I tried the same test as before but also set RangeUpper to 10, I expected no difference, because the maximum in RangeUpperList is 10, but find I get very small values as though no spectrum were counted in the integration.

KarlPalmen:
It has behaved as I expected when I tried it again, so may not be a serious problem.

KarlPalmen:
:shipit:

soininen:
I can reproduce this problem, and I think it *is* serious. I'll try to investigate it.

soininen:
I could see the problem with the `MUSR00022725.nxs` file a few times but it happened too infrequently to grab it by the gills. I might be completely wrong, but I suspect the muon loader might be corrupting the workspace randomly; at least Mantid segfaults at exit if the file is loaded during the session so there is something wrong with it.

Anyway, I took the opportunity to fine-tune the algorithm's behavior if you mix *RangeUpper* and *RangeUpperList* or the lower limits. Before, the lists always took precedence, so giving *RangeUpper*=10 and *RangeUpperList*=[11, 12, 13] would always integrate up to 11, 12 and 13. Now, the more strict limit of the two is applied, so that in the above case `Integration` will always integrate to 10. I think this is closer to what a user would expect.

KarlPalmen:
It's what I'd expect.

KarlPalmen:
Works OK. I think the problem arises when one integrates for a second time having viewed the data of the first integration. Then one can accidentally integrate the integrated workspace instead of the original workspace

KarlPalmen:
:shipit:

PR #19003, opened 2017-02-27
Avoid SIGABRT in CompAssembly::getComponentByName

### LINKED ISSUE #19002: LoadHelper should throw if component name does not exist (getComponentByName)
### Expected behavior

Mantid should not crash when an error occurs loading a file with incompatible IDF, incompatible in terms of unexpected component names.
`CompAssembly::getComponentByName` takes a string holding the component name as input. In case no component of that name exist, an exception should be thrown.

### Actual behavior

If a component of the given name does not exist in the IDF, `LoadHelper` tries to catch a runtime_error or a NotFoundError never thrown by `CompAssembly::getComponentByName`.

### Platforms affected

All

vreimund:
`CompAssembly::getComponentByName` takes a string holding the component name as input. In case no component of that name exist, an exception should be thrown rather than to stop Mantid executing (sigabrt).

**To test:**

Choose a file (likely a loader located in `DataHandling`), which uses `CompAssembly::getComponentByName` and its corresponding instrument definition file, then try loading.

For example:
- choose `LoadILLReflectrometry` and replace the component name  `uniq_detector` of its instrument definition file `D17_Definition.xml` with another name
- execute `LoadILLReflectometry`, it should result in an error but should not stop Mantid executing

Fixes #19002.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Geometry/src/Instrument/CompAssembly.cpp
```diff
@@ -344,7 +345,8 @@ CompAssembly::getComponentByName(const std::string &cname, int nlevels) const {
   } // while-end
 
   // If we have reached here then the search failed
-  return boost::shared_ptr<const IComponent>();
+  throw std::runtime_error("Error when trying to find the component " + cname +
+                           " , this component may not exist. \n");
 }
```
KarlPalmen: One unit test requires invalid argument exception to be thrown if the component is not found (not runtime error).

KarlPalmen: It is `SpecularReflectionCalculateThetaTest::test_throws_if_DetectorComponentName_unknown`.

reimundILL: Thanks

SimonHeybrock:
I am not sure that changing the behavior of `getComponentByName` is the right approach. I think there is client code that relies on the current behavior (returning a `nullptr`). Have you considered changing the code that causes the `SIGABRT` (is it attempting to dereference the `nullptr`?)?

reimundILL:
I will change `LoadHelper.cpp` where the error occurs. I will leave the `getComponentByName` as it is. I was not sure which way to go, thank you for your opinion

reimundILL:
Well, I did something wrong in git

PR #19014, opened 2017-02-28
ISIS Reflectometry (Polref) interface: Implemented uniform time slicing options in 'Event Handling' tab

### LINKED ISSUE #18500: Implement uniform time slice options
This issue is to implement:

- Uniform slicing
- ~~Slicing by log value~~ (will be done as a separate issue)

eventhandling

Groups should be exclusive.

PranavBahuguna:
This PR adds two new slicing options to the `Event Handling` tab in the interface called`'UniformEven` and `Uniform` respectively. A third option `LogValue` has been added, but is currently non-functional and will be implemented in another issue.

**To test:**

- If using windows, this must be tested in Release mode.
- The interface may be accessed via Interfaces->Reflectometry->ISIS Reflectometry (Polref). From there, go to the `Event Handling` tab.
- Each option can be separately activated by clicking on its radio button. Ensure that only one option is activated at any time.
- Go to the `Runs` tab and load `INTER_NR_test2.tbl` (from the sample INTER dataset) and click on `Process`. This should reduce the table and produce several workspaces. Ensure that these are the same as committing the same process on master.
- In `Event Handling`, select `Custom slicing`  and enter values `0, 10, 20, 30`. Go back to `Runs`, select the first group and process. This should produce sliced workspaces with time ranges of `0-10`, `10-20` and `20-30` seconds respectively. Ensure that the results match that of master (although the workspaces will be named by slice index, rather than time range).
- Now select `Uniform even slicing` and enter `5`. Go back to `Runs`, select the first group and process. This should  produce 5 evenly-sized sliced workspaces of each type.
- Finally select `Uniform slicing` and enter `10`. Runs `13460` and `13460` have a duration of 52 and 21 seconds respectively. Therefore this should produce 6 and 3 sliced workspaces of each type for each run. 
As only common slices are used for the stitched IvsQ workspaces, there should only be three sliced workspaces for them.
- Open a new table and add run `38415` in the first row, with `38393` as transmission run. Process with each slicing option and ensure that the `plot rows` and `plot group` buttons work correctly and produce the desired result.
- Repeat previous step (using one of the slicing options) and delete `TOF_38415` from the workspaces dock. Ensure that when attempting to plot, the interface gives an error message.
- Ensure documentation is accurate and up to date.

Fixes #18500.

*Release notes have been updated*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/ReflEventWidget.ui
```diff
@@ -41,7 +138,45 @@
...
+            <item>
+              <widget class="QComboBox" name="logValueComboBox">
+                <item>
+                  <property name="text">
+                    <string>T1</string>
+                  </property>
+                </item>
+              </widget>
+            </item>
+            <item>
```
raquelalvarezbanos: If slicing by log value is not implemented yet, can we remove the related widgets?

MantidQt/CustomInterfaces/src/Reflectometry/QtReflEventView.cpp
```diff
@@ -34,12 +48,49 @@ IReflEventPresenter *QtReflEventView::getPresenter() const {
...
+  std::string values;
+
+  if (m_sliceType == m_slicingTypes[0])
+    values = m_ui.uniformEvenEdit->text().toStdString();
+  else if (m_sliceType == m_slicingTypes[1])
+    values = m_ui.uniformEdit->text().toStdString();
+  else if (m_sliceType == m_slicingTypes[2])
+    values = m_ui.customEdit->text().toStdString();
+  else if (m_sliceType == m_slicingTypes[3])
+    values = m_ui.logValueComboBox->currentText().toStdString();
```
raquelalvarezbanos: Just a very minor point: I think this would be easier to read if we replaced `m_slicingTypes` with the actual values we are using for the comparisons. I had to go back to `QtReflEventView.h` to check how the vector was initialized. Is there any reason not to do this?

MantidQt/CustomInterfaces/test/ReflEventPresenterTest.h
```diff
@@ -26,14 +26,26 @@ class ReflEventPresenterTest : public CxxTest::TestSuite {
...
+    TS_ASSERT(Mock::VerifyAndClearExpectations(&mockView));
+  }
+
+  void test_get_slicing_type() {
+    MockEventView mockView;
+    ReflEventPresenter presenter(&mockView);
+
+    EXPECT_CALL(mockView, getTimeSlicingType())
+        .Times(Exactly(1))
+        .WillOnce(Return("UniformEven"));
```
raquelalvarezbanos: Just for information, if the value returned by the mocked view is not going to be used, you can just do:

```
EXPECT_CALL(mockView, getTimeSlicingType()).Times(Exactly(1));
```

raquelalvarezbanos:
Something is not right in terms of functionality. The following two cases should give the same results (for run `INTER38415` with transmission `INTER38393`):

- Custom slicing with Python list `0, 122, 244, 366, 488, 610`
- Uniform slicing with `5` even time slices

This is because the run duration (according to sample logs) is 610 seconds. The problem seems to be [here](https://github.com/mantidproject/mantid/pull/19014/files#diff-c72c951c1d262cd463f8d2ccf3ef702fR304), when debugging, I can see that `totalDurationSec = 6`, are we sure that `getFirstPulseTime()` and `getLastPulseTime()` are the methods we need to use?

MantidQt/CustomInterfaces/src/Reflectometry/QtReflEventView.cpp
PranavBahuguna: It was mostly just a little more convenient, but I see that it makes it less readable so I will change it.

MantidQt/CustomInterfaces/src/Reflectometry/ReflDataProcessorPresenter.cpp
```diff
@@ -392,34 +464,53 @@ std::string ReflDataProcessorPresenter::takeSlice(const std::string &runNo,
...
+  // Num of slices can be predetermined with uniform even or custom slicing
+  if (timeSlicingType == "UniformEven") {
+    const auto &row = items.at(0).at(0);
+    const std::string wsName = getReducedWorkspaceName(row, "TOF_");
+    parseUniform(timeSlicingValues, timeSlicingType, wsName, startTimes,
+                 stopTimes);
+  } else if (timeSlicingType == "Custom") {
+    parseCustom(timeSlicingValues, startTimes, stopTimes);
+  }
+
```
PranavBahuguna: After some investigation, I think we only need to store how many slices are used for each run in each group. We don't actually need the start/stop times after slicing is done.

raquelalvarezbanos:
When I try to reduce run `38415` (transmission `38393`) with a custom python list `0, 122, 244, 366, 488, 610`, I get an error:

```
Logic Error in execution of algorithm FilterByTime
You need to specify either the StartTime or StopTime parameters; or both the AbsoluteStartTime and AbsoluteStopTime parameters; but not other combinations.
```

I think we should be able to reduce this run with this parameters.

MantidQt/CustomInterfaces/src/Reflectometry/ReflDataProcessorPresenter.cpp
```diff
@@ -305,9 +305,8 @@ void ReflDataProcessorPresenter::parseUniform(const std::string &timeSlicing,
     return;
   }
 
-  const auto minTime = mws->getFirstPulseTime();
-  const auto maxTime = mws->getLastPulseTime();
-  const auto totalDuration = maxTime - minTime;
+  const auto run = mws->run();
+  const auto totalDuration = run.endTime() - run.startTime();
   double totalDurationSec = totalDuration.seconds();
```
raquelalvarezbanos: This is wrong, it should be

`double totalDurationSec = totalDuration.total_seconds();`

as suggested in a comment above. Having just `seconds()` does not produce the correct results.

In general, when working on a new piece of work/fixing something you should think about the best way of checking that results are correct. In this case it is very easy, you should check that the following three cases give the same `IvsQ` and `IvsLam` workspaces:

1. Custom slicing with Python list `0, 122, 244, 366, 488, 610`
2. Uniform slicing with `5` even time slices
3. Uniform slicing with `122` seconds

The reason is that run `INTER38415` has a total duration of `610` seconds (which is something you can see from the logs), and `610` is divisible by `5`, which means that the above cases should produce the same time slices. Check this when you change the line above and you'll see that, for instance, the `IvsQ` workspaces match.

raquelalvarezbanos:
[This](https://github.com/mantidproject/mantid/pull/19014/files#diff-c72c951c1d262cd463f8d2ccf3ef702fR312) line is wrong, it should be

`double totalDurationSec = totalDuration.total_seconds();`

as suggested in a comment above. Having just `seconds()` does not produce the correct results.

In general, when working on a new piece of work/fixing something you should think about the best way of checking that results are correct. In this case it is very easy, you should check that the following three cases give the same result workspaces:

1. Custom slicing with Python list `0, 122, 244, 366, 488, 610`
2. Uniform slicing with `5` even time slices
3. Uniform slicing with `122` seconds

The reason is that run `INTER38415` has a total duration of `610` seconds (which is something you can see from the logs), and `610` is divisible by `5`, which means that the above cases should produce the same time slices. Check this when you change that line and you'll see that, for instance, the `IvsQ` workspaces match exactly.

raquelalvarezbanos:
Results look good now.

:shipit:

PR #19027, opened 2017-03-01
Multifile limit in facility info

### LINKED ISSUE #19013: Multifilelimit in Facilities.xml
The multiple file loading has a limit of `100`, when first installing mantid. It is possible to configure it via config service,  but the initial value could be more reasonable and facility-tailored, e.g. for ILL, `1000` is a better default. This can be implemented through `FacilityInfo`.

gvardany:
Config entry `loading.multifilelimit` is used in `MultiFileNameParser` to prevent too large file ranges.
It was previously set to 100 by default, however it is not a reasonable default for `ILL`.
This PR makes the default value of the limit facility dependent through Facilities.xml.
It is 100 for the facilities which don't specify the limit explicitly.
When the limit is set explicitly by the user, it will take priority over the facility dependent defaults.

**To test:**
1. Delete `Mantid.user.properties`
2. Build and run MantidPlot, set the log level to error
3. Set `ISIS`/`MARI` as facility/instrument
4. `print config['loading.multifilelimit']` should be empty
5. Open `Load` dialog, and put some range above 100, e.g. `100000:100200`
6. Tab out from the filename field of the dialog, and observe the error log
7. Set `ILL`/`IN16B` and repeat 4>6, there should be no error
8. Repeat 5>6 with larger range, e.g. `100000:101010`, see the error log
9. `config['loading.multifilelimit'] = '10'`
10. Repeat 3>8 checking the error message

Fixes #19013 .

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


DavidFair:
This works correctly and detects n+1 runs above the configured limit. Additionally it preserves the current behaviour by default so is opt-in.
:shipit:

PR #19064, opened 2017-03-06
Sanitise Invalid Nexus Log Characters

### LINKED ISSUE #19062: LoadNexusLogs triggers debug assertion on bad log values
### Expected behavior
The loading method within `LoadNexusLogs` should strip and warn on any invalid characters. Whilst the invalid characters shouldn't be there we should handle them better
### Actual behavior
We are passing them on to the c library - this comes with additional risks as correctly detected by MSVC
Load `IMAT00003680.nxs` in debug mode on MSVC to see the assertion when loading in the `EPICS_PUTLOG` property
### Platforms affected
All

DavidFair:
Description of work.
This PR fixes an issue where invalid characters were being passed to `iscntrl`. We now check that the character is within a valid range (0 - 255) for the c style locale we use by default then continue to call `iscntrl`.

Additionally I've swapped out the c style library `cctypes` with `locale` and switched to `std::iscntrl`.

**To test:**
- Build on MSVC in debug mode (for the assertion to work):
- Load 
[IMAT00003680](https://github.com/mantidproject/mantid/files/821190/IMAT00003680.zip) on Master, ensure it throws a assertion error (note if you run without debugging Mantid will appear to close gracefully)
- Load the same file on this branch - it should load successfully
- Go to log data. Check the value for `EPICS_PUTLOG` does not contain special characters

Fixes #19062 .

Does not need to be in the release notes. - Assertion is only enabled for debug mode.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/DataHandling/src/LoadNexusLogs.cpp
```diff
@@ -654,5 +656,33 @@ LoadNexusLogs::createTimeSeries(::NeXus::File &file,
   }
 }
 
+/**
+  * Checks whether the specified character is invalid or a control
+  * character. If it is invalid (i.e. negative) or a control character
+  * the method returns true. If it is valid and not a control character
+  * it returns false. Additionally if the character is invalid is
```
AnthonyLim23: small thing, but I think there is a typo at the end of the above line

Framework/DataHandling/src/LoadNexusLogs.cpp
```diff
@@ -654,5 +656,33 @@ LoadNexusLogs::createTimeSeries(::NeXus::File &file,
...
+  * @param c :: Character to check
+  * @param propName :: The name of the property currently being checked for
+  *logging
+  * @param log :: Reference to logger to print out to
+  * @return :: True if control character OR invalid. Else False
+  */
+bool LoadNexusLogs::isControlValue(const char &c, const std::string &propName,
+                                   Kernel::Logger &log) {
+  // Have to check it falls within range accepted by c style check
+  if (c <= -1) {
```
AnthonyLim23: I think it would be safer if we used a static_cast on c. Then we know it is comparing two integers.

DavidFair: So I double checked this as I was unsure what the compiler would do. According to the C++ standard this is up to the compiler:
http://en.cppreference.com/w/cpp/language/implicit_conversion#Numeric_promotions
So the compiler could compare `char` types or `int` types (using promotion).

Checking the compiler output confirms my suspicions - both the compiler chosen type and static_cast types are treated identically because of this:
https://godbolt.org/g/7lFs4p

For this reason I think a static_cast wouldn't be appropriate here as it implies we can check for a range larger than char (which is untrue) and it may force the compiler to use `int` types when a `char` type is appropriate

Framework/DataHandling/src/LoadNexusLogs.cpp
DavidFair: 👍

Framework/DataHandling/src/LoadNexusLogs.cpp
AnthonyLim23: something about comparing a char and an int makes me uneasy.

DavidFair: Your right - looking around whilst for `int` and other larger types the default sign is specified in the standard for a `char` it is implementation defined (based on architecture)

In this case if we cast up from an unsigned `char` where that is used on that architecture to a `signed int` we should be able to both hold the value and sign regardless. Additionally if we do this we need to check the value is not greater than 255 as well (if using unsigned types)

Looking again at the compiler output this doesn't add any extra overhead but should ensure we are doing the correct thing : https://godbolt.org/g/Y5ewMz

AnthonyLim23: sounds good to me

AnthonyLim23:
Behaves as expected. :shipit:

Framework/DataHandling/inc/MantidDataHandling/LoadNexusLogs.h
```diff
@@ -97,6 +97,11 @@ class DLLExport LoadNexusLogs : public API::Algorithm {
   /// Create a time series property
   Kernel::Property *createTimeSeries(::NeXus::File &file,
                                      const std::string &prop_name) const;
+
+  /// Returns true if the character is a control value or invalid
+  static bool isControlValue(const char &c, const std::string &propName,
```
martyngigg: Is there any reason this needs to be a static on the class? Can it just be a simple anonymous function defined only in the `.cpp` file?

DavidFair: Completely forgot about using anonymous namespaces good point.

martyngigg (approved):

PR #19076, opened 2017-03-07
MergeRuns sum option and error control

### LINKED ISSUE #17520: MergeRuns to allow error message if bins are different
#16928 made some changes to allow warning and failing for `MergeRuns` where sample logs are different. It would also be useful to allow runs to be skipped with an error message where the binning is different.

Currently `MergeRuns` automatically rebins when bins are different, but this should be made optional (but still default).

### LINKED ISSUE #19042: Addition to MergeRuns for summing
Add additional option to `MergeRuns` to allow summing the sample logs. For things like elapsed time this makes more sense than creating a time series, where the sum is not easily obtained.

Also, allow overwriting of the old sample log value instead of creating a new one.

ianbush:
Adds the following changes:
 1. Sum merge type - this allows a sample log to be simply summed
 2. Sample logs are now overwritten, and not renamed
 3. Option to skip a file if binning is different
 4. Option to throw an error instead of skipping

**To test:**

Load 2 data files that make sense to merge.

 1. Try setting a sum property in `MergeRuns`, either via the algorithm or instrument parameters file, check the output is as expected
 2. Sample logs should be overwitten with new type, should be noticable for TimeSeries and List options. Check a TimeSeries/List is created instead of any existing numeric type. Check merging two workspaces that were already merged once also works.
 3. Rebin one workspace, check that workspace is skipped in the merge
 4. Check the algorithm throws when using `RebinBehaviour=Skip` and `FailBehaviour=Stop`. Also check for the `SampleLogsFail` option and `FailBehaviour=Stop`.

Also try a few unscripted tests, and check code looks good.

Fixes #19042 
Fixes #17520 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/inc/MantidAlgorithms/MergeRuns/SampleLogsBehaviour.h
+++ b/Framework/Algorithms/inc/MantidAlgorithms/MergeRuns/SampleLogsBehaviour.h
@@ -34,8 +34,9 @@ namespace Algorithms {
 */
 class MANTID_ALGORITHMS_DLL SampleLogsBehaviour {
 public:
-  enum class MergeLogType { TimeSeries, List, Warn, Fail };
+  enum class MergeLogType { Sum, TimeSeries, List, Warn, Fail };
 
+  static const std::string SUM_MERGE;
   static const std::string TIME_SERIES_MERGE;
   static const std::string LIST_MERGE;
   static const std::string WARN_MERGE;
@@ -53,6 +54,7 @@ public:
   };
 
   SampleLogsBehaviour(API::MatrixWorkspace &ws, Kernel::Logger &logger,
+                      const std::string sampleLogsSum,
                       const std::string sampleLogsTimeSeries,
                       const std::string sampleLogsList,
                       const std::string sampleLogsWarn,
@@ -94,6 +96,8 @@ private:
   createTolerancesVector(const size_t numberNames,
                          const std::vector<std::string> &tolerances);
 
+  void updateSumProperty(double addeeWSNumber, double outWSNumber,
+                         API::MatrixWorkspace &outWS, const std::string name);
   void updateTimeSeriesProperty(API::MatrixWorkspace &addeeWS,
                                 API::MatrixWorkspace &outWS,
                                 const std::string name);
--- a/Framework/Algorithms/inc/MantidAlgorithms/MergeRuns/SampleLogsBehaviour.h
+++ b/Framework/Algorithms/inc/MantidAlgorithms/MergeRuns/SampleLogsBehaviour.h
@@ -7,7 +7,7 @@
 namespace Mantid {
 namespace Algorithms {
 
-/** MergeRunsSampleLogsBehaviour : This class holds information relating to the
+/** SampleLogsBehaviour : This class holds information relating to the
   behaviour of the sample log merging. It holds a map of all the sample log
   parameters to merge, how to merge them, and the associated tolerances.
 
@@ -43,9 +43,6 @@ public:
   static const std::string FAIL_MERGE;
   static const std::string WARN_MERGE_TOLERANCES;
   static const std::string FAIL_MERGE_TOLERANCES;
-
-  static const std::string TIME_SERIES_SUFFIX;
-  static const std::string LIST_SUFFIX;
 
   struct SampleLogBehaviour {
     std::shared_ptr<Kernel::Property> property;
@@ -103,7 +100,6 @@ private:
                                 const std::string name);
   void updateListProperty(API::MatrixWorkspace &addeeWS,
                           API::MatrixWorkspace &outWS,
-                          Kernel::Property *addeeWSProperty,
                           const std::string name);
   void checkWarnProperty(const API::MatrixWorkspace &addeeWS,
                          Kernel::Property *addeeWSProperty,
--- a/Framework/Algorithms/inc/MantidAlgorithms/MergeRuns/SampleLogsBehaviour.h
+++ b/Framework/Algorithms/inc/MantidAlgorithms/MergeRuns/SampleLogsBehaviour.h
@@ -63,6 +63,7 @@ public:
   void mergeSampleLogs(API::MatrixWorkspace &addeeWS,
                        API::MatrixWorkspace &outWS);
   void setUpdatedSampleLogs(API::MatrixWorkspace &outWS);
+  void removeSampleLogsFromWorkspace(API::MatrixWorkspace &addeeWS);
   void readdSampleLogToWorkspace(API::MatrixWorkspace &addeeWS);
   void resetSampleLogs(API::MatrixWorkspace &ws);
 
--- a/Framework/Algorithms/inc/MantidAlgorithms/MergeRuns/SampleLogsBehaviour.h
+++ b/Framework/Algorithms/inc/MantidAlgorithms/MergeRuns/SampleLogsBehaviour.h
@@ -48,7 +48,6 @@ public:
     std::shared_ptr<Kernel::Property> property;
     double tolerance;
     bool isNumeric;
-    MergeLogType mergeLogType;
   };
 
   SampleLogsBehaviour(API::MatrixWorkspace &ws, Kernel::Logger &logger,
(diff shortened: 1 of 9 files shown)


Framework/Algorithms/src/MergeRuns/SampleLogsBehaviour.cpp
```diff
@@ -254,21 +255,22 @@ std::shared_ptr<Property> SampleLogsBehaviour::addPropertyForTimeSeries(
   std::shared_ptr<Property> returnProp;
 
   try {
-    // See if property exists already - merging an output of MergeRuns
-    returnProp = std::shared_ptr<Property>(
-        ws.getLog(item + TIME_SERIES_SUFFIX)->clone());
+    // See if property exists as a TimeSeriesLog already - merging an output of
+    // MergeRuns
+    ws.run().getTimeSeriesProperty<double>(item);
+    returnProp = std::shared_ptr<Property>(ws.getLog(item)->clone());
```
soininen: Would
```
returnProp.reset(ws.getLog(item)->clone());
```
save a construction and an assignment?

Framework/Algorithms/src/MergeRuns/SampleLogsBehaviour.cpp
```diff
@@ -371,6 +376,22 @@ void SampleLogsBehaviour::mergeSampleLogs(MatrixWorkspace &addeeWS,
...
+ * Perform the update for a sum property, adding a new value to the existing
+ *one. Skipped if the time series log entry is in the addeeWS.
+ *
+ * @param addeeWSNumber the sample log value from the workspace being merged
+ * @param outWSNumber the sample log value for the merged workspace
+ * @param outWS the workspace the others are merged into
+ * @param name the name of the property
+ */
+void SampleLogsBehaviour::updateSumProperty(double addeeWSNumber,
+                                            double outWSNumber,
```
soininen: The names bring to my mind runNumbers. Would `valueInAddeeWS` or something similar be better?

Framework/Algorithms/src/MergeRuns/SampleLogsBehaviour.cpp
```diff
@@ -371,6 +376,22 @@ void SampleLogsBehaviour::mergeSampleLogs(MatrixWorkspace &addeeWS,
...
+                                            MatrixWorkspace &outWS,
+                                            const std::string name) {
+  outWS.mutableRun().addProperty(name, addeeWSNumber + outWSNumber, name, true);
```
soininen: Why `name` is here twice? Isn't this the
```
addProperty(const std::string &name, const TYPE &value,
                   const std::string &units, bool overwrite = false)
```
signature, in which case you are giving quite a strange units for the property.

soininen (changes requested):

soininen:
An example on how to use the sum functionality in instrument parameter files could be added to the documentation, as well.

Framework/Algorithms/inc/MantidAlgorithms/MergeRuns/SampleLogsBehaviour.h
```diff
@@ -34,25 +34,25 @@ namespace Algorithms {
...
-
   struct SampleLogBehaviour {
     std::shared_ptr<Kernel::Property> property;
     double tolerance;
     bool isNumeric;
+    MergeLogType mergeLogType;
   };
 
   SampleLogsBehaviour(API::MatrixWorkspace &ws, Kernel::Logger &logger,
+                      const std::string sampleLogsSum,
```
ianbush: This is now fixed.

Framework/Algorithms/src/MergeRuns.cpp
```diff
@@ -40,6 +46,9 @@ void MergeRuns::init() {
   declareProperty(make_unique<WorkspaceProperty<Workspace>>(
                       "OutputWorkspace", "", Direction::Output),
                   "Name of the output workspace");
+  declareProperty(
+      "SampleLogsSum", "",
+      "A comma separated list of the sample logs to sum into a single entry.");
```
ianbush: Indeed it would, I have moved this so it should not break the signature now.

Framework/Algorithms/src/MergeRuns/SampleLogsBehaviour.cpp
ianbush: Yes, I have changed this elsewhere where the same pattern occurs too.

Framework/Algorithms/src/MergeRuns/SampleLogsBehaviour.cpp
ianbush: I wanted to convey it is a number not, say, a string value. I have gone for `addeeWSNumericValue`.

Framework/Algorithms/src/MergeRuns/SampleLogsBehaviour.cpp
ianbush: Whoops, yes, that would have some strange units!

ianbush:
I have pushed the code changes, and also added a doc test for the sum option.

soininen (approved):

ianbush:
@soininen I realise I have missed the point on the doc changes - I will add the instrument parameter file bits for the sum option too.

soininen:
The algorithm does a bit too strict checking: if I list a sample log entry such as 'sample.temperature' in _SampleLogsTimeSeries_ and _SampleLogsWarn_ at the same time, I get:
```
Error when making list of merge items, sample log "sample.temperature" defined more than once!
```
Here, I would like to merge the temperatures as time series but at the same time make sure the values do not differ too much.

soininen:
Why running this:
```python
mantid.config.appendDataSearchDir('/net4/serdon/illdata/162/in4/exp_4-01-1529/rawdata/')

ws1 = Load('086490')
ws2 = Load('086491')

merged = MergeRuns('ws1, ws2', SampleLogsTimeSeries='sample.temperature')
```
gives this warning:
```
TimeSeriesProperty sample.temperature could not be added to another property of the same name but incompatible type.
```
?

ianbush:
The new restriction is there because sum, time series and list now can not be used together. It does make sense to allow fail/warn in combination with these so I try and loosen that restriction.

I'll take a look at that warning.

ianbush:
@soininen The changes we discussed are here now, you can use sum, time series and list merge in combination with warn/fail options. The warning is gone too. Please try testing lots of permutations again and let me know if things look good now.

soininen:
Where does the Duration info in time series come from? When running this:
```python
wss = Load(Filename = '/net4/serdon/illdata/162/in4/exp_4-01-1529/rawdata/086490:086491')
merged = MergeRuns(InputWorkspaces='086490, 086491', SampleLogsTimeSeries='sample.temperature')
```
I get Duration = 1800s for the 'sample.temperature' time series, but neither workspace has such duration. In fact, both input workspaces have 'duration' = 0, and the real duration is in the `actual_time` log.

soininen:
The algorithm gives an error, if I want to merge nonexistent logs as time series when giving the logs as input properties. This is all fine, but it seems that the algorithm silently ignores nonexistent logs when they are specified in the IPF. Maybe a warning should be shown in this case?

soininen:
In my IPF I have this:
```
<parameter name="sample_logs_warn_tolerances" type="string">
	<value val="kaljaa" />
</parameter>
```
This kill the algorithm, since "kaljaa" is not a number, but the error message is cryptic:
```
stod
```

ianbush:
In the time series logs actual date-time is used, taken from `start_time` in the workspace. The files seem to be 30 minutes apart so the result is as I would expect.

ianbush:
I have just pushed changes to give better error messages when tolerances can not be converted to doubles, and supressed the warning when looking at time series logs (the `running` log is something specific to ISIS).

When trying to merge a non-existent time series log you should see `Could not merge sample log "kaljaa", does not exist in workspace! This sample log will be ignored.` Do you not get that?

soininen:
I do get the error message when I mention a non-existent log for `MergeRuns`.

soininen:
This might be something to consider for another issue/PR, but I'll leave it here anyway:

I'm wondering if we could handle the 'start_time' and 'end_time' logs in a graceful way, as well. Preferably one would choose the minimum (earliest date/time) of 'start_time's and the maximum (last date/time) of 'end_time's for the merged workspace. These are string logs, and though they seems to be in some standard format, I don't know if their values can be parsed reliably.

ianbush:
Yes, let's leave the start and end times for a separate PR, that will need some thought. Parsing the values is not an issue, that is done to construct the time series. Depending on how you treat other information you might really want a vector of start/end times to account for all the data.

soininen:
@ianbush The algorithm seem to happily accept negative fail and warn tolerances.

ianbush:
@soininen - an error is now thrown if a negative tolerance is given.

soininen:
Tested this with different combinations of sums, time_series, warnings, failings and tolerances. Everything seems to work as expected. The code changes look good to me, as well. Nice work! :shipit:

Issue #18989, opened 2017-02-24
Offset in Asymmetry 


AnthonyLim23:
### Expected behavior
MUSR data should osciallte around a value of zero for the asymmetry. 
### Actual behavior
There is a non-zero offset in the data 
### Steps to reproduce the behavior
Load a file (e.g. MUSR00062260.nxs)  via the muon analysis. The data will be offset. This is easily visable if you fit a flat backgound and exposc functions to the data. The flat background will have a non-zero value. 
### Platforms affected
All

AnthonyLim23:
The offset is numeric only when the flatBackground and the GausOsc functions are used together. If just the flatBackground is fitted to the data then it returns a value of zero (except it is visible that there is an offset). Further investigation has shown that this is due to the errors in the fitting routine. Fitting a flatBackground without errors improves the answer but the value is still larger than the error.

AnthonyLim23:
Fitting a GausOsc + flatBackground during the normalization (use the value for the flatBackground as the normalization constant) does give a value for zero for the intercept, in most cases. However, the method fails for 0 field e.g. 57980 (musr). Using just a flatBackground to determine the normalization constant does remove it if the scenario is the same (e.g. no errorbars are used in both the normalization and fit of the output data). This suggests that the problem is that the method we are using is inappropriate for the data, as the current methods do what they are supposed to.

AnthonyLim23:
Talked to the scientists and got the following request: 

> On loading / grouping the data, generate the "normalised counts" by multiplying by exp(t/muon lifetime) and dividing by the number of good frames (perhaps also by the number of detectors in the group). Probably store this in another workspace (called something like MUSR01234_<groupname>_NormCounts ).

Do an initial guess of "N0" by summing the raw counts in the group, over the time range to be used for fitting, and dividing by the integral of the muon lifetime exponential over that range.

To plot the data, generate an Asymmetry workspace with Asymmetry = (NormCounts/N0)-1. Perhaps use a child algorithm for this as it's needed again later.

For fitting: have a composite function called something like "TFAsymmetry" which takes a numeric parameter N0 and also sums the user's choice of fit function(s) included within it, just as the implied CompositeFunction does now. The collection of muon and other functions that might be included within do not need to be changed at all. This function would be automatically entered in the function dialog, not to be deleted, and its parameter N0 initialised to the guess above.

[...]

PR #19083, opened 2017-03-07
Offset in asymmetry calculations: new algorithms 

AnthonyLim23:
The asymmetry calculations in muon analysis were not centered on zero. Two new algorithms have been made to calculate the asymmetry (an estimate and a full calculation). 

**To test:**

The easiest way to test the data is to open the muon analysis interface and load a MUSR file. Change the Plot type to counts (from asymmetry). The plot should update. 

Then run EstimateAsymmetryFromCounts and the input workspace should be MuonAnalysisGrouped_1 (there are only 4 spectra). The result should be a spectrum that oscillates around a value close to zero (unless a zero field case has been used, which should increase over time). I find that an x range of 0 to 15 and a y range of -30 to 30 works well.  

Then run CalculateAsymmetry and the input workspace should be MuonAnalysisGrouped_1. You will then be able to enter a function to use (can do a list) and the default is good for MUSR data. The result should be a spectrum that oscillates around a value of zero, test this by fitting a flatbackground and a GaussOsc to the output. This will not work for zero field. 

Is the first stage of fixing https://github.com/mantidproject/mantid/issues/18989 and the next stage will be to add these algorithms to the muon analysis gui. 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Algorithms/test/CalculateAsymmetryTest.h
```diff
@@ -0,0 +1,333 @@
...
+        "FittingFunction",
+        "name=UserFunction,Formula=A*cos(omega*x+phi),A=10,omega=3.0,phi=0.0");
+    TS_ASSERT_THROWS_NOTHING(alg->execute());
+    TS_ASSERT(alg->isExecuted());
+
+    MatrixWorkspace_sptr outWS = alg->getProperty("OutputWorkspace");
+
+    // First spectrum
+    // Test some X values
+    TS_ASSERT_DELTA(outWS->x(0)[10], 2.000, 0.0001);
```
gvardany: const double delta = 0.0001

Framework/Algorithms/test/CalculateAsymmetryTest.h
```diff
@@ -0,0 +1,333 @@
...
+    TS_ASSERT_EQUALS(out1->x(1).rawData(), out2->x(1).rawData());
+    TS_ASSERT_EQUALS(out1->y(1).rawData(), out2->y(1).rawData());
+    TS_ASSERT_EQUALS(out1->e(1).rawData(), out2->e(1).rawData());
+
+    // Compare non-selected spectra, the should match the input ones
+    TS_ASSERT_EQUALS(ws->x(0).rawData(), out2->x(0).rawData());
+    TS_ASSERT_EQUALS(ws->y(0).rawData(), out2->y(0).rawData());
+    TS_ASSERT_EQUALS(ws->e(0).rawData(), out2->e(0).rawData());
+  }
+  void test_yUnitLabel() {
```
gvardany: This can be merged with test_Execute

Framework/Algorithms/test/CalculateAsymmetryTest.h
```diff
@@ -0,0 +1,333 @@
...
+        AlgorithmManager::Instance().create("CalculateAsymmetry");
+    alg->initialize();
+    TS_ASSERT(alg->isInitialized())
+  }
+
+  void test_Execute() {
+
+    auto ws = createWorkspace(1, 50);
+
+    IAlgorithm_sptr alg =
```
gvardany: These lines are repeated for each test, one could have a single `alg` for all the tests with common properties set, and in each test, one could only reset the specific properties, and execute

Framework/Algorithms/test/CalculateAsymmetryTest.h
```diff
@@ -0,0 +1,333 @@
...
+        "name=UserFunction,Formula=A*cos(omega*x+phi),A=10,omega=3.0,phi=0.0");
+    TS_ASSERT_THROWS_NOTHING(alg2->execute());
+    TS_ASSERT(alg2->isExecuted());
+    MatrixWorkspace_sptr out2 = alg2->getProperty("OutputWorkspace");
+
+    // Both output workspaces should have 2 spectra
+    TS_ASSERT_EQUALS(out1->getNumberHistograms(), ws->getNumberHistograms());
+    TS_ASSERT_EQUALS(out2->getNumberHistograms(), ws->getNumberHistograms());
+
+    // Compare results, they should match for the selected spectrum
```
gvardany: put in a loop i = 0 > 1

Framework/Algorithms/test/CalculateAsymmetryTest.h
```diff
@@ -0,0 +1,333 @@
...
+    alg->setPropertyValue("OutputWorkspace", outputName);
+    alg->setProperty("StartX", 0.1);
+    alg->setProperty("EndX", 10.);
+    TS_ASSERT_THROWS_NOTHING(alg->execute());
+    TS_ASSERT(alg->isExecuted());
+
+    MatrixWorkspace_sptr outWS = alg->getProperty("OutputWorkspace");
+
+    // First spectrum
+    // Test some X values
```
gvardany: again, take the tolerance out

Framework/Algorithms/test/CalculateAsymmetryTest.h
```diff
@@ -0,0 +1,333 @@
...
+    alg->setProperty("EndX", 0.9);
+    alg->setProperty("OutputWorkspace", outputName);
+    alg->setProperty(
+        "FittingFunction",
+        "name=UserFunction,Formula=A*cos(omega*x+phi),A=10,omega=3.0,phi=0.0");
+    TS_ASSERT_THROWS_NOTHING(alg->execute());
+    TS_ASSERT(alg->isExecuted())
+  }
+
+  void test_noRange() {
```
gvardany: The case of the test method names does not seem to be consistent.

Framework/Algorithms/test/EstimateAsymmetryFromCountsTest.h
```diff
@@ -0,0 +1,292 @@
+#ifndef MUONESTIMATEASYMMETRYFROMCOUNTSTEST_H_
+#define MUONESTIMATEASYMMETRYFROMCOUNTSTEST_H_
+
+#include <cxxtest/TestSuite.h>
+#include "MantidAlgorithms/EstimateAsymmetryFromCounts.h"
+#include "MantidKernel/PhysicalConstants.h"
+#include "MantidAPI/FrameworkManager.h"
+#include "MantidAPI/AlgorithmManager.h"
```
gvardany: move up

Framework/Algorithms/test/EstimateAsymmetryFromCountsTest.h
```diff
@@ -0,0 +1,292 @@
+#ifndef MUONESTIMATEASYMMETRYFROMCOUNTSTEST_H_
```
gvardany: The comments for the `CalculateAsymmetryTest` apply also here

docs/source/release/v3.10.0/CalculateAsymmetry.rst
```diff
@@ -0,0 +1,8 @@
+============
```
gvardany: This is not the correct way of adding release notes. Remove these files. You should edit the muon.rst file, under Algorithms heading. Also put the links to the algorithm documentations (see e.g. Framework.rst how it is done). No need to mention the unit test and the documentation in the release note (they are supposed to be there).

gvardany (changes requested):
I have read the code and have some remarks inline. Most importantly pay attention on moving some parts out of the parallel loops. In general, I think the algorithm names don't speak about Muons (at least for me), so I would call them CalculateMuonAsymmetry and so on. The unit tests can also be better organised by sharing the same input workspace, and the algorithm object. I have not tried to run the algorithms yet. I will try to do shortly.

Framework/Algorithms/src/EstimateAsymmetryFromCounts.cpp
```diff
@@ -0,0 +1,157 @@
...
+#include "MantidAlgorithms/AsymmetryHelper.h"
+#include "MantidAlgorithms/EstimateAsymmetryFromCounts.h"
+#include "MantidAPI/IFunction.h"
+#include "MantidAPI/MatrixWorkspace.h"
+#include "MantidAPI/Run.h"
+
+#include "MantidAPI/Workspace_fwd.h"
+#include "MantidAPI/WorkspaceFactory.h"
+#include "MantidKernel/PhysicalConstants.h"
+#include "MantidKernel/ArrayProperty.h"
```
AnthonyLim23: do you mean to move ArrayProperty to be between WorkspaceFactorty and PhysicalConstants?

gvardany: yes, just to keep alphabetical order

gvardany:
Tested with `MUSR00015189.nxs`.
After loading it into Interfaces>Muon>Muon Analysis, there is no option to change the plot type from asymmetry to counts. Am I doing something wrong?
The algorithms do run and produce something that by eye oscillates around zero, at least up to 20microsec. Beyond that they diverge, not sure what to expect though.
When trying to fit with GausOsc+FlatBackground, I get an error: `Error in execution of algorithm Fit: Integration is not implemented for this function.`

Framework/Algorithms/src/CalculateAsymmetry.cpp
```diff
@@ -0,0 +1,240 @@
...
+double CalculateAsymmetry::getNormConstant(API::MatrixWorkspace_sptr ws,
+                                           int wsIndex,
+                                           const double estNormConstant,
+                                           const double startX,
+                                           const double endX) {
+  double retVal = 1.0;
+
+  API::IAlgorithm_sptr fit;
+  fit = createChildAlgorithm("Fit", -1, -1, true);
+  std::string tmpString = getProperty("FittingFunction");
```
AnthonyLim23: up to here 22/03/2017 14:00

AnthonyLim23:
Some of the data does not have counts. The example that I use is MUSR00062260. The algorithms tend to only be used up to about 15 microseconds, even if the range is larger it wont have mush effect as the errors are large. 

Sorry, but I do not understand that error message.

AnthonyLim23:
If you change the group/Group Pair to fwd or bwd, then you will be able to select counts.

Framework/Algorithms/src/CalculateAsymmetry.cpp
AnthonyLim23: Several parts of these lines change depending on the spectrum being analysed. Each one will have a slightly different normalized counts (input workspace), as a result the FittingFunction will be different as the estimate for the normalization will be different.

gvardany:
OK I see. Is the `MUSR00062260` in unit tests data? 
With `MUSR00015189.nxs` (now in counts) I get this when running `CalculateAsymmetry` with the default options.
```
Fit falled. Status = Failed to converge after 500 iterations.
For workspace index 0
Asym norm constant set to 1.0
```

Btw, these algorithms calculate/estimate the asymmetry, and then normalise to that, so the names as they are now are slightly misleading, since one could expect to get the asymmetry out, but not the input workspace normalised to it. I don't have a good suggestion though, maybe something like NormaliseToCalculatedAsymmetry or CalculateAsymmetryAndNormalise could be more strict to the point.

Framework/Algorithms/src/CalculateAsymmetry.cpp
gvardany: The FittingFunction at this point, when you just get the property from the algorithm, is the same for all. So this should be moved out. And then in the loop when you get the estimate, you append the necessary change. I would also do the check if this string is empty as the first thing. So if it is empty, return the estimate, and don't even instantiate the fit algorithm.

gvardany: Even more, I would check at L147 if the fitting function is empty. If so, I would not even call to `getNormConstant` but just use the estimate.

AnthonyLim23:
I cannot recall where I got the data from. It is one that I have picked up. 

For the naming, the definition of asymmetry from the counts requires a normalization to be calculated. The main feature seems to be that it is centered on 0 and the easiest way to achieve that  is to normalize the data and then subtract 1. 

It does not converge as the data is not of the right type. You want data that has oscillations in the asymmetry (when plotted in muon analysis).

gvardany:
Can you send that file then?
One more general remark, these two algorithms seem to have a lot in common, do you think it wouldn't be more compact to have one algorithm, with option Estimate/Calculate ?

Framework/Algorithms/src/CalculateAsymmetry.cpp
AnthonyLim23: To be honest I could remove L147, as the function has a default value and this will ensure that it is never empty. 

I could try creating the fit algorithm at the top of the exec (outside the loop) and passing it into the function along with a base string (for the first part of the function).

gvardany: OK, indeed this will never be empty, so the check is redundant then.

AnthonyLim23: So I have had a play with the code and all of the fit algorithm has to stay within the loop. If I just move the createChildAlgorithm outside of the loop then the first spectrum is correct, but any after that are wrong. I think this is because it adds the new data to the fit function each time the normalization is called, without removing the previous data. In the current set up the fit function is automatically cleared as it goes out of scope between different spectra.

Framework/Algorithms/src/CalculateAsymmetry.cpp
```diff
@@ -0,0 +1,240 @@
...
+  std::vector<std::string> paramnames = result->getParameterNames();
+  // Check order of names
+  if (paramnames[0].compare("f0.A0") != 0) {
+    g_log.error() << "Parameter 0 should be f0.A0, but is " << paramnames[0]
+                  << '\n';
+    throw std::invalid_argument(
+        "Parameters are out of order @ 0, should be f0.A0");
+  }
+
+  if (!fitStatus.compare("success")) {
```
AnthonyLim23: The ! is correct in that line of code because of how compare works. However, I agree that it is confusing so I have changed it to 

`fitStatus=="success" `

to be explicit.

Framework/Algorithms/test/CalculateAsymmetryTest.h
```diff
@@ -0,0 +1,333 @@
...
+  void testInit() {
+    IAlgorithm_sptr alg =
+        AlgorithmManager::Instance().create("CalculateAsymmetry");
+    alg->initialize();
+    TS_ASSERT(alg->isInitialized())
+  }
+
+  void test_Execute() {
+
+    auto ws = createWorkspace(1, 50);
```
AnthonyLim23: Some of the tests are interested in multiple spectra. To save computation I have used as few spectra as possible each time. I think there are a couple that could be reduced to one spectrum.

AnthonyLim23:
I believe that I have made the changes that you have suggested, with the exception of renaming the files. 

I will have a think  about merging them. 

If you switch to fwd in groups in the home tab it should then let you pick counts. Any MUSR data could be used then.

AnthonyLim23:
I did not merge the algorithms, because I have extended the calculate algorithm (I will need the new features for the GUI that I will be developing next). I feel that merging the two will make the code too difficult to follow.

AnthonyLim23:
I think that is now everything, please let me know if I have missed anything.

AnthonyLim23:
I believe that is everything

Framework/Algorithms/src/MuonAsymmetryHelper.cpp
```diff
@@ -0,0 +1,130 @@
...
+                        exp(-endX / MUON_LIFETIME_MICROSECONDS));
+  return summation * Delta / denominator;
+}
+/**
+* Finds the first index in bin edges that is after
+* the start time.
+* @param xData :: [input] Input HistogramData as bin edges
+* @param startX :: [input] the start time
+* @returns :: The index to start calculations from
+* @throw :: runtime error if the range is zero.
```
gvardany: remove @throw

Framework/Algorithms/src/MuonAsymmetryHelper.cpp
```diff
@@ -0,0 +1,130 @@
...
+      std::upper_bound(xData.rawData().begin(), xData.rawData().end(), startX);
+  return std::distance(xData.rawData().begin(), upper);
+}
+/**
+* find the first index in bin edges that is after
+* the endtime.
+* @param xData :: [input] HistogramData as bin edges
+* @param endX :: [input] the end time
+* @returns :: The last index to  include in calculations
+* @throw :: runtime error if the range is zero.
```
gvardany: remove @throw

Framework/Algorithms/src/MuonAsymmetryHelper.cpp
```diff
@@ -0,0 +1,130 @@
...
+    Mantid::PhysicalConstants::MuonLifetime * MICROSECONDS_PER_SECOND};
+}
+
+namespace Mantid {
+
+using namespace Kernel;
+using API::Progress;
+using std::size_t;
+
+/** a
```
gvardany: remove a

gvardany (changes requested):
Thanks for the update, it looks better now. See few minor edits/typos inline.

Still, I see that "estimate" algorithms is almost fully "contained" inside the "calculate" counterpart, which only does the fitting on top. So I am in favour of merging into one to avoid maintaining duplicated parts. I don't insist on this, but let's hear the gatekeeper's opinion. Otherwise it is fine.

AnthonyLim23:
Thank you. I have had a chat with the scientists and they would like one property (the normalisation) added to each as an output. I will do that now and will commit it.

AnthonyLim23:
I have now done the updates that were requested by the scientists. A future version of the CalculateMuonAsymmetry algorithm will be able to do multiple fitting.

Framework/Algorithms/test/CalculateMuonAsymmetryTest.h
```diff
@@ -72,6 +74,24 @@ IAlgorithm_sptr setUpAlg() {
   return asymmAlg;
 }
 
+std::vector<double> convertToVec(std::string list){
```
gvardany: Use boost::split here, pass the argument by const &.

AnthonyLim23:
I have now used boost::split an std::for_each to do the conversion

Framework/Algorithms/test/CalculateMuonAsymmetryTest.h
```diff
@@ -74,21 +74,11 @@ IAlgorithm_sptr setUpAlg() {
...
-    vec.push_back(std::stod(tmp.c_str()));
-    start = end + 1;
-  }
-  // record last value
-  tmp.assign(list, start, end - start);
-  vec.push_back(std::stod(tmp.c_str()));
-
+  std::vector<std::string> tmpVec;
+  boost::split(tmpVec,list,boost::is_any_of(","));  
+  std::for_each(tmpVec.begin(),tmpVec.end(),[&vec](std::string const &element){vec.push_back(std::stod(element));});
```
gvardany: std::transform could be nicer here

AnthonyLim23:
I have changed it to transform and altered the arguments as necessary.

gvardany:
OK, I think that was all :shipit:

gvardany (approved):

PR #19111, opened 2017-03-10
Reflectometry GUI now outputs stitched transmission run in wavelength

### LINKED ISSUE #18707: Reflectometry GUI does not output combined transmission run in wavelength
The reflectometry GUI should output combined (stitched) transmission run in wavelength. This doesn't happen when transmission runs are specified via the `Settings` tab.

PranavBahuguna:
The Reflectometry GUI was supposed to output a stitched transmission in wavelength if transmission runs are specified before processing. This happens if transmission runs were specified directly in each row being processed, but not if they were specified in the `Settings` tab. This PR fixes this very issue.

**To test:**

- Open the Reflectometry interface from Interfaces->Reflectometry->ISIS Reflectometry
- Load `INTER_NR_test2.tbl` (available from sample ISIS dataset) by navigating to Reflectometry->Import .TBL
- Opening the first group reveals they have transmission runs `13463` and `13464`. Processing the group should produce a stitched transmission workspace `TRANS_13463_13464`.
- Remove the transmission runs from the table. Go to the `Settings` tab and input `13463,13464` into `Transmission run(s)`. Go back to the `Runs` tab and process the group again and ensure you have the same results as before.

Fixes #18707.

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


gemmaguest (changes requested):
The changes look good, but I have found one small problem; if you remove the transmission runs from both the Runs and Settings tabs, then you cannot run the reduction - it gives the error `File "INTER" not found`. It looks like it's trying to load a transmission run even though none was specified.

gemmaguest:
I've noticed another problem while testing this. It's not related to this PR, but might make sense to fix alongside it. If you enter different transmission runs in the `Runs` and `Settings` tabs (say 13463 in `Runs` and 13464 in `Settings`) then only one of them gets used, but both of them get loaded into workspaces, so it's not immediately obvious which one got used. Could we skip loading the one that is not used?

gemmaguest (approved):

gemmaguest:
I haven't had time to fully test the new changes yet, but it looks like there is a bug where you get an `Invalid Filename ""` error if there is no transmission run specified in the `Settings` tab (even if there is one in the `Runs` tab).

gemmaguest:
I'm now seeing the following error with this: Invalid value for property InputWorkspaces (str list) "": Select a value

PranavBahuguna:
Could you be more specific? I cannot seem to be able to replicate this problem you are seeing.

gemmaguest:
Ah, sorry, it's working now. I think I had the wrong instrument selected or something...

gemmaguest:
This looks good now and behaves as expected. Please :shipit:

PR #19112, opened 2017-03-10
ISIS Powder - Migration of GEM

### LINKED ISSUE #18643: ISIS Powder Diffraction Scripts - Migration of GEM

This ticket holds the current status of the migration of the GEM instrument. As it has a similar construction and workflow to POLARIS this should be it the next candidate to migrate over.

The plan for this branch is to only contain the necessary changes to migrate GEM - it should not contain fixes to the `isis_powder` scripts. Instead we will make those changes in a separate branch as needed and merge the changes back in.

----

### Work to do:
- [x] Create skeleton instrument file
- [x] Meet and assess requirements - what parameters do they need, what already exists...etc?
- [x] Create calibration mapping file following discussion
- [x] Create advanced configuration file for instrument specific properties
- [x] Absorption corrections
- [x] Bring any remaining params from existing parameter files
- [x] Scattering Cross Section....etc. need to be options as they use VNb (94.86% V 5.14% NB)
- [x] Sample empty as an optional field on a per run basis -
- [x] We need to find out where to perform the subtraction but the change can be added at a later date as it is very rarely used at the moment. 
- [x] Keep one unit of choice in ADS (d-spacing / TOF)
- [x] Add PDF and Rietveld modes
- [x] Separate d-spacing and grouping files

## Misc Refactoring Work
- [x] Move instances of the pattern - get run numbers, get first element of them into common
- [x] All the existing instruments which used the `cry` scripts have the same output processing, move this to a common method

DavidFair:
Description of work.
This PR adds the GEM instrument into the ISIS Powder scripts. 

It contains two .cal files which normally should be added as external data. However the users specifically requested I add an example of the file structure in the respective instrument examples folder. This means they can copy the example folder out, mimic the structure and copy the relevant files in on their machines.

**To test:**
Download the following .zip file which contains all files required for testing: https://uploadfiles.io/da5bb1
- Extract files
- There is a minor mistake in the mapping file I created for this PR. In `<extracted_folder>/Calibration/Gem_Mapping.yaml` the value `rietveld` needs to be changed to `Rietveld` (capital R). The script will correctly inform you of this mistake if you forget to make this change
- Add `<extracted_folder>/InputData` to the list of user directories in Mantid so it can find the data
- Within the extracted folder are two scripts - one for creating the calibration and one for focusing. In both of them make sure you set the `path_to_extracted_file` variable to your extracted folder path.
- Try running focus first - it should fail and prompt you to run the calibration
- Run calibration then focus, they should both succeed 
- Try messing around with the various kwargs. Try removing, misspelling and adding additional keyword args. It should handle them all appropriately and give sane error messages
- Check the data looks sane. The focused silicon run should look like this (dSpacing and zoomed in): 
gem si all (notice the sharp peaks)
- Code review

As always any and all suggestions and feedback is very welcome during this process. 

Fixes #18643 .

*Does not need to be in the release notes. - ISIS Powder scripts are still in development*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


samueljackson92 (changes requested):
A few minor comments. I'm guessing unit/system tests are added in the next PR.

samueljackson92:
As discussed gave it a good test with variety of parameter combinations. Seems to work ok. Output data appears to match the example given above.

scripts/Diffraction/isis_powder/routines/common.py
```diff
@@ -119,6 +126,36 @@ def generate_run_numbers(run_number_string):
...
+    :param vanadium_string: The name of this vanadium run
+    :param args: Any identifying properties to append to the name
+    :return: The splined vanadium name
+    """
+    out_name = "VanSplined" + '_' + str(vanadium_string)
+    for value in args:
+        out_name += '_' + str(value)
+
+    out_name += ".nxs"
+    return out_name
```
DavidFair: Typically 40-60 characters at the most. These are used internally for the script. Basically we generate a unique fingerprint for a splined vanadium containing the offset name, tt_mode (on Pearl) and vanadium numbers. 
This also means users can see something of the format
`VanSplined_12345-12350_offsets_16_5.cal.nxs`

scripts/Diffraction/isis_powder/routines/common.py
```diff
@@ -119,6 +126,36 @@ def generate_run_numbers(run_number_string):
...
+    """
+    Takes a run number string and returns the first user specified run from that string
+    :param run_number_string: The string to parse to find the first user rnu
+    :return: The first run for the user input of runs
+    """
+    run_numbers = generate_run_numbers(run_number_string=run_number_string)
+    if isinstance(run_numbers, list):
+        run_numbers = run_numbers[0]
+
+    return run_numbers
```
DavidFair: It should always return a list - this is a leftover from when it didn't. 👍

scripts/Diffraction/isis_powder/routines/common.py
samueljackson92: That's probably ok. I've seen longer. I just wary sometimes of have workspace names that are too long, but this is fine.

scripts/Diffraction/isis_powder/routines/common.py
```diff
@@ -109,7 +116,7 @@ def generate_run_numbers(run_number_string):
     :return: A list of run numbers generated from the string
     """
     # Check its not a single run
-    if isinstance(run_number_string, int) or run_number_string.isdigit():
+    if isinstance(run_number_string, int):
```
samueljackson92: This does not make sense anymore. You're checking it's an `int` then casting it to an `int`. So I'm guess this fails when it's a `str`.

Is this function still expected to take strings? If so I think we still need another case here? 
1) is string run number
2) is int run number
3) is a list of run numbers

If not then we should update the docstring, which also needs a `:param:` entry anyway and the variable name.

samueljackson92 (changes requested):
One more minor thing:

samueljackson92 (approved):

samueljackson92:
Changes now LGTM. :shipit:

PR #19118, opened 2017-03-13
Add algorithm that computes basic statistics for HKLs without intensities

### LINKED ISSUE #19049: Completeness, redundancy, missing reflections from SortHKL
SortHKL calculates (among other quantities) completeness and redundancy, but only if the supplied PeaksWorkspace has intensities.

In order to be useful after a PredictPeaks run, there should be a way to calculate these quantities for peaks without intensities. Another quantity that comes out of this are the missing unique reflections, it should be possible to obtain them from somewhere.

MichaelWedel:
A new algorithm has been added: `CountReflections` takes a PeaksWorkspace and a few other parameters and calculates statistics based on counting reflections. There is some overlap with SortHKL, but SortHKL critically relies on having reflection intensities available to calculate agreement indicators. The new algorithm does not take intensities into account, making it more useful in combination with `PredictPeaks`.

`SortHKL` had some helper classes defined its files, these have been moved out into a separate namespace, so `SortHKL` and `CountReflections` use the same code for organizing reflections based on point group symmetry. Unfortunately that made the PR quite a bit larger than I had originally anticipated.

Documentation has been updated, release notes are still missing. I'm marking the PR `In Progress` so that we can discuss whether that's what required and a few other minor things (naming, algorithm properties).

**To test:**

You could look at the doctest and modify it a bit or you could load actual data from some external source and compare whether the algorithm produces the same numbers as other tools.

Fixes #19049.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


samueljackson92 (commented):
Sorry this has taken me so long to take a look at. Here's some initial code review comments, but on the whole it looks good. I'm going to have a little play.

Framework/Crystal/src/CountReflections.cpp
```diff
@@ -0,0 +1,217 @@
...
+/** Initialize the algorithm's properties.
+ */
+void CountReflections::init() {
+  declareProperty(Kernel::make_unique<WorkspaceProperty<PeaksWorkspace>>(
+                      "InputWorkspace", "", Direction::Input),
+                  "A workspace with peaks to calculate statistics for. Sample "
+                  "with valid UB-matrix is required.");
+
+  auto centeringSymbols = getAllReflectionConditionSymbols();
+  declareProperty("LatticeCentering", centeringSymbols[0],
```
samueljackson92: `PredictPeaks` outputs the reflection condition name, but here we are using the symbol. Perhaps we should make this consistent between the two?

Framework/Crystal/inc/MantidCrystal/PeakStatisticsTools.h
```diff
@@ -0,0 +1,144 @@
...
+  double m_meanIOverSigma;
+  double m_dspacingMin;
+  double m_dspacingMax;
+  double m_chiSquared;
+
+  std::vector<DataObjects::Peak> m_peaks;
+
+private:
+  void calculatePeaksStatistics(
+      std::map<Kernel::V3D, UniqueReflection> uniqueReflections);
```
MichaelWedel: In principle I'd agree, but the method actually modifies the map (it removes outliers) so actually want it to work on a copy. There's probably a better way to do the removal of outliers...maybe this would be a good chance to implement that as well.

Framework/Crystal/src/PeakStatisticsTools.cpp
```diff
@@ -0,0 +1,330 @@
...
+    const UnitCell &cell, const std::pair<double, double> &dLimits,
+    const PointGroup_sptr &pointGroup,
+    const ReflectionCondition_sptr &centering)
+    : m_reflections(), m_pointgroup(pointGroup) {
+  HKLGenerator generator(cell, dLimits.first);
+  HKLFilter_const_sptr dFilter = boost::make_shared<const HKLFilterDRange>(
+      cell, dLimits.first, dLimits.second);
+  HKLFilter_const_sptr centeringFilter =
+      boost::make_shared<const HKLFilterCentering>(centering);
+  HKLFilter_const_sptr filter = dFilter & centeringFilter;
```
MichaelWedel: Have not tried it, but it would probably be a lot shorter :)

Framework/Crystal/src/CountReflections.cpp
MichaelWedel: Yes, that is one thing we should discuss. All other algorithms use the full name, but it seems very verbose and in this context it should be sufficient with the symbol. It's the same for the point group - all other algorithms use the "name", which is the symbol plus the lattice system (if I remember correctly).

"Breaking the rule" here was meant to start that discussion.

Framework/Crystal/inc/MantidCrystal/PeakStatisticsTools.h
samueljackson92: Ah I see the issue. Yes perhaps a refactor is the solution.

MichaelWedel: `UniqueReflection` could get a `remove_outliers` parameter in the constructor. Then a new `UniqueReflection` would be constructed with the parameter set. There'd still be a copy involved, but it would be slightly less bad I suppose. Do you have another idea?

Framework/Crystal/src/CountReflections.cpp
samueljackson92: I thought that might be the case. I guess the question comes down to whether the average user will be more at home with a verbose description or the symbols. I think I side with you in that the shorter symbol should be used. It will also make it easier to work with Python side.

MichaelWedel: Another option would be dedicated property types which accept both forms (one for each of centering, point group, space group). That would definitely be better for backwards compatibility. It would also eliminate the need to do this "get a list of all x, make a validator, check what string the user supplied, create object accordingly", which is found anywhere point group and/or lattice centering are used.

samueljackson92: Yes that could be useful, I agree that it ideally should not care. Can we make a separate issue for this?

Framework/Crystal/inc/MantidCrystal/PeakStatisticsTools.h
samueljackson92: Now that I look again I think it is probably ok as it is.

Framework/Crystal/src/PeakStatisticsTools.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+/// Removes peaks whose intensity deviates more than sigmaCritical from the
+/// intensities' mean.
+void UniqueReflection::removeOutliers(double sigmaCritical) {
+  if (sigmaCritical <= 0.0) {
+    throw std::invalid_argument(
+        "Critical sigma value has to be greater than 0.");
+  }
+
+  if (m_peaks.size() > 2) {
```
MichaelWedel: I changed this entire thing to return a new `UniqueReflection` instead. That also solves the problem with the `const` map in the calculation function and it's generally nicer to not change state. Maybe the `setIntensities` thing should work the same way.

Framework/Crystal/src/CountReflections.cpp
MichaelWedel: I've opened #19442.

MichaelWedel:
I don't quite understand the RHEL7 error - looks unrelated.

Framework/Crystal/test/PeakStatisticsToolsTest.h
```diff
@@ -0,0 +1,356 @@
...
+    TS_ASSERT_THROWS(reflection.removeOutliers(-10.0), std::invalid_argument);
+  }
+
+  void test_UniqueReflectionRemoveOutliersFewPeaks() {
+    std::vector<Peak> peaks = getPeaksWithIandSigma({30.0, 34.0}, {4.5, 6.5});
+
+    UniqueReflection reflection(V3D(2, 3, 4));
+    reflection.addPeak(peaks[0]);
+
+    TS_ASSERT_THROWS_NOTHING(reflection.removeOutliers());
```
MichaelWedel: I just realized this doesn't really test anything anymore because the reflection is not modified by `removeOutliers`.

samueljackson92 (commented):
Couple of places where we could use `auto`, but other than that and the test case this looks good.

Framework/Crystal/src/PeakStatisticsTools.cpp
```diff
@@ -256,11 +256,11 @@ void PeaksStatistics::calculatePeaksStatistics(
...
         // The original algorithm sets the intensities and sigmas to the mean.
         double sqrtOfMeanSqrSigma = getRMS(sigmas);
-        unique.second.setPeaksIntensityAndSigma(meanIntensity,
-                                                sqrtOfMeanSqrSigma);
+        outliersRemoved.setPeaksIntensityAndSigma(meanIntensity,
+                                                  sqrtOfMeanSqrSigma);
       }
 
-      const std::vector<Peak> &reflectionPeaks = unique.second.getPeaks();
+      const std::vector<Peak> &reflectionPeaks = outliersRemoved.getPeaks();
```
MichaelWedel: I'll go over everything in that file and make everything `auto` :)

VickieLynch:
:shipit: Results match StatisticsOfPeaksWorkspace using full dataset of Monoclinic C TOPAZ peaks.

PR #19119, opened 2017-03-13
Add time index for DetectorInfo methods -> scanning support

### LINKED ISSUE #19016: Add methods that accept time index to `Beamline::DetectorInfo`.
This is needed for scanning and will also make code that uses `SpectrumInfo::spectrumDefinition` cleaner.

SimonHeybrock:
This PR adds an (optional) time index to methods of `DetectorInfo`. It can now be accessed with the complete index stored in `SpectrumDefinition` (obtained from `SpectrumInfo::spectrumDefinition()`). With this time index, we support scanning, with certain limitations:

- Scanning of detectors is support, but not of any higher-level method.
- Code that accesses the position or the rotation of an `IDetector` for a scanning workspace will fail. Currently this is `solidAngle`, `getBoundingBox`, and low level code such as ray tracing. Only access via `DetectorInfo` or `SpectrumInfo` is supported (position, rotation, L2, twoTheta).
- All scan positions will yield the same parameters obtained from the parameter map.

`DetectorInfo` will now do a more complex index translation, however, a mechanism that should ensure close to zero overhead for non-scanning cases was put in place.

**To test:**

Thorough code review.

Fixes #19016.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/inc/MantidAPI/DetectorInfo.h
+++ b/Framework/API/inc/MantidAPI/DetectorInfo.h
@@ -76,18 +76,30 @@ public:
   size_t size() const;
 
   bool isMonitor(const size_t index) const;
+  bool isMonitor(const std::pair<size_t, size_t> index) const;
   bool isMasked(const size_t index) const;
+  bool isMasked(const std::pair<size_t, size_t> index) const;
   double l2(const size_t index) const;
+  double l2(const std::pair<size_t, size_t> index) const;
   double twoTheta(const size_t index) const;
+  double twoTheta(const std::pair<size_t, size_t> index) const;
   double signedTwoTheta(const size_t index) const;
+  double signedTwoTheta(const std::pair<size_t, size_t> index) const;
   Kernel::V3D position(const size_t index) const;
+  Kernel::V3D position(const std::pair<size_t, size_t> index) const;
   Kernel::Quat rotation(const size_t index) const;
+  Kernel::Quat rotation(const std::pair<size_t, size_t> index) const;
 
   void setMasked(const size_t index, bool masked);
+  void setMasked(const std::pair<size_t, size_t> index, bool masked);
   void clearMaskFlags();
 
   void setPosition(const size_t index, const Kernel::V3D &position);
+  void setPosition(const std::pair<size_t, size_t> index,
+                   const Kernel::V3D &position);
   void setRotation(const size_t index, const Kernel::Quat &rotation);
+  void setRotation(const std::pair<size_t, size_t> index,
+                   const Kernel::Quat &rotation);
 
   void setPosition(const Geometry::IComponent &comp, const Kernel::V3D &pos);
   void setRotation(const Geometry::IComponent &comp, const Kernel::Quat &rot);
--- a/Framework/API/inc/MantidAPI/DetectorInfo.h
+++ b/Framework/API/inc/MantidAPI/DetectorInfo.h
@@ -122,7 +122,7 @@ public:
   std::pair<Kernel::DateAndTime, Kernel::DateAndTime>
   scanInterval(const std::pair<size_t, size_t> &index) const;
   void setScanInterval(
-      const std::pair<size_t, size_t> &index,
+      const size_t index,
       const std::pair<Kernel::DateAndTime, Kernel::DateAndTime> &interval);
 
   void merge(const DetectorInfo &other);
--- a/Framework/API/inc/MantidAPI/DetectorInfo.h
+++ b/Framework/API/inc/MantidAPI/DetectorInfo.h
@@ -77,29 +77,29 @@ public:
   size_t size() const;
 
   bool isMonitor(const size_t index) const;
-  bool isMonitor(const std::pair<size_t, size_t> index) const;
+  bool isMonitor(const std::pair<size_t, size_t> &index) const;
   bool isMasked(const size_t index) const;
-  bool isMasked(const std::pair<size_t, size_t> index) const;
+  bool isMasked(const std::pair<size_t, size_t> &index) const;
   double l2(const size_t index) const;
-  double l2(const std::pair<size_t, size_t> index) const;
+  double l2(const std::pair<size_t, size_t> &index) const;
   double twoTheta(const size_t index) const;
-  double twoTheta(const std::pair<size_t, size_t> index) const;
+  double twoTheta(const std::pair<size_t, size_t> &index) const;
   double signedTwoTheta(const size_t index) const;
-  double signedTwoTheta(const std::pair<size_t, size_t> index) const;
+  double signedTwoTheta(const std::pair<size_t, size_t> &index) const;
   Kernel::V3D position(const size_t index) const;
-  Kernel::V3D position(const std::pair<size_t, size_t> index) const;
+  Kernel::V3D position(const std::pair<size_t, size_t> &index) const;
   Kernel::Quat rotation(const size_t index) const;
-  Kernel::Quat rotation(const std::pair<size_t, size_t> index) const;
+  Kernel::Quat rotation(const std::pair<size_t, size_t> &index) const;
 
   void setMasked(const size_t index, bool masked);
-  void setMasked(const std::pair<size_t, size_t> index, bool masked);
+  void setMasked(const std::pair<size_t, size_t> &index, bool masked);
   void clearMaskFlags();
 
   void setPosition(const size_t index, const Kernel::V3D &position);
-  void setPosition(const std::pair<size_t, size_t> index,
+  void setPosition(const std::pair<size_t, size_t> &index,
                    const Kernel::V3D &position);
   void setRotation(const size_t index, const Kernel::Quat &rotation);
-  void setRotation(const std::pair<size_t, size_t> index,
+  void setRotation(const std::pair<size_t, size_t> &index,
                    const Kernel::Quat &rotation);
 
   void setPosition(const Geometry::IComponent &comp, const Kernel::V3D &pos);
@@ -120,10 +120,10 @@ public:
 
   size_t scanCount(const size_t index) const;
   std::pair<Kernel::DateAndTime, Kernel::DateAndTime>
-  scanInterval(const std::pair<size_t, size_t> index) const;
-  void
-  setScanInterval(const std::pair<size_t, size_t> index,
-                  std::pair<Kernel::DateAndTime, Kernel::DateAndTime> interval);
+  scanInterval(const std::pair<size_t, size_t> &index) const;
+  void setScanInterval(
+      const std::pair<size_t, size_t> &index,
+      const std::pair<Kernel::DateAndTime, Kernel::DateAndTime> &interval);
 
   void merge(const DetectorInfo &other);
 
--- a/Framework/API/inc/MantidAPI/DetectorInfo.h
+++ b/Framework/API/inc/MantidAPI/DetectorInfo.h
@@ -2,6 +2,7 @@
 #define MANTID_API_DETECTORINFO_H_
 
 #include "MantidAPI/DllConfig.h"
+#include "MantidKernel/DateAndTime.h"
 #include "MantidKernel/Quat.h"
 #include "MantidKernel/V3D.h"
 
@@ -116,6 +117,15 @@ public:
   /// Returns the index of the detector with the given detector ID.
   /// This will throw an out of range exception if the detector does not exist.
   size_t indexOf(const detid_t id) const { return m_detIDToIndex.at(id); }
+
+  size_t scanCount(const size_t index) const;
+  std::pair<Kernel::DateAndTime, Kernel::DateAndTime>
+  scanInterval(const std::pair<size_t, size_t> index) const;
+  void
+  setScanInterval(const std::pair<size_t, size_t> index,
+                  std::pair<Kernel::DateAndTime, Kernel::DateAndTime> interval);
+
+  void merge(const DetectorInfo &other);
 
   friend class SpectrumInfo;
 
--- a/Framework/API/src/DetectorInfo.cpp
+++ b/Framework/API/src/DetectorInfo.cpp
@@ -279,6 +279,11 @@ void DetectorInfo::setPosition(const Geometry::IComponent &comp,
     const auto index = indexOf(det->getID());
     setPosition(index, pos);
   } else {
+    const auto &detIndices = getAssemblyDetectorIndices(comp);
+    if ((!detIndices.empty()) && m_detectorInfo.isScanning())
+      throw std::runtime_error("Cannot move parent component containing "
+                               "detectors since the beamline has "
+                               "time-dependent (moving) detectors.");
     // This will go badly wrong if the parameter map in the component is not
     // identical to ours, but there does not seem to be a way to check?
     const auto oldPos = comp.getPos();
@@ -288,7 +293,6 @@ void DetectorInfo::setPosition(const Geometry::IComponent &comp,
     // If comp is a detector cached positions stay valid. In all other cases
     // (higher level in instrument tree, or other leaf component such as sample
     // or source) we flush all cached positions.
-    const auto &detIndices = getAssemblyDetectorIndices(comp);
     if (detIndices.size() == 0 || detIndices.size() == size()) {
       // Update only if comp is not a bank (not detectors) or the full
       // instrument (all detectors). The should make this thread-safe for
@@ -321,6 +325,11 @@ void DetectorInfo::setRotation(const Geometry::IComponent &comp,
     const auto index = indexOf(det->getID());
     setRotation(index, rot);
   } else {
+    const auto &detIndices = getAssemblyDetectorIndices(comp);
+    if ((!detIndices.empty()) && m_detectorInfo.isScanning())
+      throw std::runtime_error("Cannot move parent component containing "
+                               "detectors since the beamline has "
+                               "time-dependent (moving) detectors.");
     // This will go badly wrong if the parameter map in the component is not
     // identical to ours, but there does not seem to be a way to check?
     const auto pos = toVector3d(comp.getPos());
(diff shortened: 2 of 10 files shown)


Framework/API/src/DetectorInfo.cpp
```diff
@@ -133,21 +173,68 @@ double DetectorInfo::signedTwoTheta(const size_t index) const {
   return angle;
 }
 
+/// Returns signed 2 theta (signed scattering angle w.r.t. to beam direction).
+double
+DetectorInfo::signedTwoTheta(const std::pair<size_t, size_t> index) const {
+  if (isMonitor(index))
```
OwenArnold: 👍  Good additional checks.

Framework/Beamline/inc/MantidBeamline/DetectorInfo.h
```diff
@@ -68,41 +68,87 @@ class MANTID_BEAMLINE_DLL DetectorInfo {
...
+  void initScanIntervals();
+  void initIndices();
+
   Kernel::cow_ptr<std::vector<bool>> m_isMonitor{nullptr};
   Kernel::cow_ptr<std::vector<bool>> m_isMasked{nullptr};
   Kernel::cow_ptr<std::vector<Eigen::Vector3d>> m_positions{nullptr};
   Kernel::cow_ptr<std::vector<Eigen::Quaterniond>> m_rotations{nullptr};
+
+  Kernel::cow_ptr<std::vector<size_t>> m_scanCounts{nullptr};
+  Kernel::cow_ptr<std::vector<std::pair<int64_t, int64_t>>> m_scanIntervals{
```
OwenArnold: Some comment about using the same underlying type as `Kernel::DataAndTime` is needed here.

Framework/API/src/DetectorInfo.cpp
SimonHeybrock: Note that there is some code duplication here. Should be fixed if we move these secondary position methods into `Beamline`, but I think that is currently not possible.

Framework/Beamline/src/DetectorInfo.cpp
```diff
@@ -78,28 +90,236 @@ bool DetectorInfo::isEquivalent(const DetectorInfo &other) const {
...
+  if (!m_scanIntervals || !other.m_scanIntervals)
+    failMerge("scan intervals not defined");
+  if (!(m_isMonitor == other.m_isMonitor) &&
+      (*m_isMonitor != *other.m_isMonitor))
+    failMerge("monitor flags mismatch");
+  // TODO If we make masking time-independent we need to check masking here.
+
+  std::vector<bool> merge(other.m_positions->size(), true);
+
+  for (size_t index1 = 0; index1 < other.m_positions->size(); ++index1) {
```
OwenArnold: In the following there's some very confusing mixing of the `index` terminology. i.e. `index1` should be something like `linearIndex1`

Framework/Beamline/src/DetectorInfo.cpp
```diff
@@ -78,28 +90,236 @@ bool DetectorInfo::isEquivalent(const DetectorInfo &other) const {
...
+    const size_t detIndex = getIndex(other.m_indices, index1).first;
+    const auto &interval1 = (*other.m_scanIntervals)[index1];
+    for (size_t timeIndex = 0; timeIndex < scanCount(detIndex); ++timeIndex) {
+      const auto index2 = linearIndex({detIndex, timeIndex});
+      const auto &interval2 = (*m_scanIntervals)[index2];
+      if (interval1 == interval2) {
+        if ((*m_isMasked)[index2] != (*other.m_isMasked)[index1])
+          failMerge("matching scan interval but mask flags differ");
```
OwenArnold: Good checking. We might need to go further and report which indexes are not matching.

Framework/Beamline/src/DetectorInfo.cpp
```diff
@@ -78,28 +90,236 @@ bool DetectorInfo::isEquivalent(const DetectorInfo &other) const {
...
 
+/// Set the mask flag of the detector with given index. Not thread safe.
+void DetectorInfo::setMasked(const std::pair<size_t, size_t> index,
+                             bool masked) {
+  m_isMasked.access()[linearIndex(index)] = masked;
+}
+
+/// Returns the position of the detector with given index.
+Eigen::Vector3d
+DetectorInfo::position(const std::pair<size_t, size_t> index) const {
```
OwenArnold: The behaviour here seems to rely more on type safety aspects rather than throwing as in the non-scanning cases. `checkNoTimeDependence`.

SimonHeybrock: @OwenArnold I am not sure I understand this comment?

OwenArnold: Looks like access to a scanning `DetectorInfo` via `position` causes an exception to be raised. Looks like there is no equivalent runtime check for access to a non-scanning `DetectorInfo`. I assume this is deliberate.

SimonHeybrock: Yes, it allows for generic code:

```cpp
for(const auto &index : spectrumInfo.spectrumDefinition(i))
  pos + =detectorInfo.position(index);
```

The time index would always be zero in the non-scanning case, but there is no need for client code to know that.

Framework/API/src/SpectrumInfo.cpp
```diff
@@ -174,14 +174,12 @@ const Geometry::IDetector &SpectrumInfo::getDetector(const size_t index) const {
   return *m_lastDetector[thread];
 }
 
-std::vector<size_t> SpectrumInfo::getDetectorIndices(const size_t index) const {
-  std::vector<size_t> detIndices;
-  for (const auto &def : spectrumDefinition(index))
-    detIndices.push_back(def.first);
-  if (detIndices.empty())
+const SpectrumDefinition &
+SpectrumInfo::getDetectorIndices(const size_t index) const {
```
ianbush: Refactor method name too?

Framework/API/test/MatrixWorkspaceTest.h
```diff
@@ -1419,6 +1422,59 @@ class MatrixWorkspaceTest : public CxxTest::TestSuite {
...
+    BoundingBox bb;
+    TS_ASSERT_THROWS(det.getBoundingBox(bb), std::runtime_error);
+    // Moving parent not possible since non-detector components do not have time
+    // indices and thus DetectorInfo cannot tell which set of detector positions
+    // to adjust.
+    TS_ASSERT_THROWS(detInfo.setPosition(*det.getParent(), V3D(1, 2, 3)),
+                     std::runtime_error);
+    TS_ASSERT_THROWS(detInfo.setRotation(*det.getParent(), Quat(1, 2, 3, 4)),
+                     std::runtime_error);
+  }
```
ianbush: The first scanning workspace, a nice milestone!

ianbush (commented):
This looks great, good to see the first scanning workspaces constructed.

Here every detector is effectively able to do an independent scan - independent start/end times for each detector. This makes sense for flexibility, but I expect is not the most common case. Does it make sense to also support a single set of start/end times for each time index, or is this not really a big performance overhead?

For scanning workspace construction should we allow methods other than merging `DetectorInfo`, or make it easier via some helper methods?

Framework/API/inc/MantidAPI/DetectorInfo.h
```diff
@@ -76,18 +77,30 @@ class MANTID_API_DLL DetectorInfo {
   size_t size() const;
 
   bool isMonitor(const size_t index) const;
+  bool isMonitor(const std::pair<size_t, size_t> index) const;
```
SimonHeybrock: Passing pairs by reference now, though I am not sure if that is really better (cannot exclude that it is actually slower).

I don't think a typedef makes things better. Client code mostly uses `auto` I guess, and if it accessing the members the typedef does not help in any case.

Framework/Beamline/src/DetectorInfo.cpp
```diff
@@ -78,28 +90,250 @@ bool DetectorInfo::isEquivalent(const DetectorInfo &other) const {
...
+ *
+ * The interval start and end values would typically correspond to nanoseconds
+ * since 1990, as in Kernel::DateAndTime. */
+void DetectorInfo::setScanInterval(const std::pair<size_t, size_t> index,
+                                   std::pair<int64_t, int64_t> interval) {
+  if (!m_scanIntervals)
+    initScanIntervals();
+  // We forbid this since we (currently?) can not verify that the new interval
+  // has no collisions with any of the other intervals for this detector.
+  checkNoTimeDependence();
```
SimonHeybrock: Maybe I should change this to accept only a detector index?

Framework/Beamline/src/DetectorInfo.cpp
```diff
@@ -78,28 +90,250 @@ bool DetectorInfo::isEquivalent(const DetectorInfo &other) const {
...
+ * incremented by the scan count of that detector in `this`. The relative order
+ * of time indices added from `other` is preserved. If the interval for a time
+ * index in `other` is identical to a corresponding interval in `this`, it is
+ * ignored, i.e., no time index is added. */
+void DetectorInfo::merge(const DetectorInfo &other) {
+  if (size() != other.size())
+    failMerge("size mismatch");
+  if (!m_scanIntervals || !other.m_scanIntervals)
+    failMerge("scan intervals not defined");
+  if (!(m_isMonitor == other.m_isMonitor) &&
```
SimonHeybrock: No it is in principle the same but `cow_ptr` does not support that. I think I actually have another open PR that adds this operator, if that gets merged first we can clean this up here.

Framework/Beamline/test/DetectorInfoTest.h
```diff
@@ -253,6 +254,241 @@ class DetectorInfoTest : public CxxTest::TestSuite {
...
+    DetectorInfo info(PosVec(1), RotVec(1));
+    info.setScanInterval({0, 0}, {1, 2});
+    TS_ASSERT_EQUALS(info.scanInterval({0, 0}),
+                     (std::pair<int64_t, int64_t>(1, 2)));
+  }
+
+  void test_setScanInterval_failures() {
+    DetectorInfo info(PosVec(1), RotVec(1));
+    TS_ASSERT_THROWS(info.setScanInterval({0, 0}, {1, 1}), std::runtime_error);
+    TS_ASSERT_THROWS(info.setScanInterval({0, 0}, {2, 1}), std::runtime_error);
```
SimonHeybrock: Can that be done? I think I tried to do that a while back but did not find a way to do that with `cxxtest`. Is there?

SimonHeybrock:
*Here every detector is effectively able to do an independent scan - independent start/end times for each detector. This makes sense for flexibility, but I expect is not the most common case. Does it make sense to also support a single set of start/end times for each time index, or is this not really a big performance overhead?*

There is some memory overhead for storing all the scan intervals, but probably that is negligible compared to all the other data in the workspace, such as histograms. There might be some performance overhead when merging, but I would rather wait for benchmarks of the first merge of real workspaces, so we see if this is an actual issue. If it is, I'd be happy to optimize the existing `DetectorInfo::merge` (there are a few things than can be improved, even if we keep independent scans for each detector.

*For scanning workspace construction should we allow methods other than merging DetectorInfo, or make it easier via some helper methods?*

I am not sure I understand?

Framework/Beamline/src/DetectorInfo.cpp
```diff
@@ -78,28 +90,251 @@ bool DetectorInfo::isEquivalent(const DetectorInfo &other) const {
...
+}
+
+void DetectorInfo::initScanCounts() {
+  checkNoTimeDependence();
+  m_scanCounts = Kernel::make_cow<std::vector<size_t>>(size(), 1);
+}
+
+void DetectorInfo::initScanIntervals() {
+  checkNoTimeDependence();
+  m_scanIntervals = Kernel::make_cow<std::vector<std::pair<int64_t, int64_t>>>(
```
OwenArnold: So we have as many scan intervals as we have detectors `size()`?

SimonHeybrock: Um, I am not sure I understand?

OwenArnold: Do we rely on `merge` to allow us to have multiple intervals per detector?

SimonHeybrock: Yes, currently `merge` is the only way of creating a `DetectorInfo` with multiple intervals.

OwenArnold: In the prototype I had a unique set of scan intervals and  the `linearindex` could be used with a `timeindex` vector to address into this. What we currently have seems a little wasteful. given that you end up with `n`*`m` intervals where `n` is the number of detectors and `m` is the number of scan positions.

SimonHeybrock: Yes, currently each detector has independent scan intervals (see also discussion with Ian above).

I am not sure how the indexing would work with `linearindex` and `timeindex` as you describe?

OwenArnold: We had some discussions about this with the prototype, and generally considered it better to avoid the dense `n` * `m` scenarios so in the prototype the `m_linearIndex` map provides a linear index into positions etc using `detectorIndex` and `scanIndex`. https://github.com/DMSC-Instrument-Data/instrument-prototype/blob/master/cow_instrument/DetectorInfo.h#L356-L361

SimonHeybrock: But this is not dense?

OwenArnold: `m_linearIndex` is dense but it's the only part of this data structure that doesn't has to be. My point is that we should consider whether we need to store such a large amount of duplicates for the `intervals` for the mantid solution.

SimonHeybrock: I suppose there are two alternatives:
- Store a vector of unique intervals and for each `(detector index, time index)` pair we store an index into that vector of unique intervals. This would save us `8 Byte` per spectrum.
- Store a vector of unique vectors of intervals. For each detector index we store an index into that vector. This saves use `8 * scanCount Byte` per spectrum.

The cost is some extra complexity. The memory saving might be negligible compared to the histogram or event list stored for each spectrum.

I do not mind doing this if it turns out that the current performance is affected significantly by having scan intervals for each individual detector, but I would prefer waiting until we know, i.e., did some benchmarking for merging workspaces that contain real amounts of data.

Framework/Beamline/src/DetectorInfo.cpp
ianbush: :+1: this seems like a good solution.

Framework/Beamline/src/DetectorInfo.cpp
OwenArnold: Delaying this would be OK on the provision that.

1. We don't end up coding this information into file formats. I guess this could easily happen if we forget to revisit this.
2. It doesn't leak heavily into the rest of Mantid (little danger of this at this stage) and result in lots of additional refactoring.

Framework/Beamline/test/DetectorInfoTest.h
ianbush: Yes, you can use either `TS_ASSERT_THROWS_ASSERT` or `TS_ASSERT_THROWS_EQUAL`. It is not the nicest but it works.

For example:
``` cpp
TS_ASSERT_THROWS_EQUALS(info.setScanInterval(0, {1, 1}), const std::runtime_error &e, std::string(e.what()), "DetectorInfo: cannot set scan interval with start > end");
```

SimonHeybrock: :+1: :+1: :+1:

Framework/Beamline/src/DetectorInfo.cpp
SimonHeybrock: @OwenArnold Regarding file formats, I think we should first figure out if Nexus actually supports such a complex case. Maybe several detectors can be linked to the same table of intervals? See also the joint slack channel with Ian, I had talked to Tobias regarding scanning in Nexus files.

OwenArnold: Ok. I can see that work is underway. My point is just to be careful about doing anything that would make revisiting the interval design hard. Aside from that, I have no problem delaying the decision as you suggest.

Framework/Beamline/src/DetectorInfo.cpp
```diff
@@ -78,28 +90,250 @@ bool DetectorInfo::isEquivalent(const DetectorInfo &other) const {
...
+    scanCounts.access()[detIndex]++;
+    m_indexMap.access()[detIndex].push_back((*m_indices).size());
+    m_indices.access().push_back(newIndex);
+    m_isMasked.access().push_back((*other.m_isMasked)[linearIndex]);
+    m_positions.access().push_back((*other.m_positions)[linearIndex]);
+    m_rotations.access().push_back((*other.m_rotations)[linearIndex]);
+    m_scanIntervals.access().push_back((*other.m_scanIntervals)[linearIndex]);
+  }
+  m_scanCounts = std::move(scanCounts);
+}
```
OwenArnold: Two minor comments on this. It might be worth reserving the memory for masks, positions and rotations, given that you know how much additional space you'll need in advance of the push back.

This is a relatively long and important method. It might be worth breaking it up to improve readability. For example extracting the warning into a separate function.

SimonHeybrock: I feel that reserving would be irrelevant. Given that `std::vector` typically(?) doubles its size when it grows I would expect that we have typically have at most one single reallocation per merge. If you merge many objects you might even do worse with `reserve`, since it might cause *more* reallocations.

OwenArnold: I think doubling is approximately right for most implementations. It is not standardised though. Might be worth looking at if we find performance issues. As I said though, a minor comment.

ianbush:
Just to confirm I am happy with these changes :+1: .

OwenArnold:
When the builds are all green :shipit:

PR #19120, opened 2017-03-13
Add translation and partitioning mechanism to IndexInfo.

### LINKED ISSUE #18522: Basic translation and partitioning classes for Indexing module.

SimonHeybrock:
This PR adds some classes for handling indexing, translation between index types, and partitioning capabilities. The latter is needed for MPI support. `IndexInfo` does not yet support MPI, but this PR deals with some cleanup and other preparations for it.

The added Doxygen of `IndexInfo` summarizes the purpose of many of the changes done here:
```
IndexInfo provides mapping from spectrum numbers to spectrum indices, and
  grouping information that defines a spectrum as a group of detectors.

  The interface of IndexInfo is designed to hide an underlying partitioning of
  data as in the case of MPI. There are three interconnected index types:
  - Spectrum numbers are user-defined (instrument-specific) identifiers for a
    spectrum. In principle these must be unique, but for legacy support this is
    currently not guaranteed. Most of the key functionality of IndexInfo is not
    available unless spectrum numbers are unique.
  - Global spectrum indices are a contiguous way of indexing all spectra,
    starting at zero. In particular, this index spans all partitions. If there
    is only a single partition the global spectrum index is equivalent to the
    index (see next item). Note that in the user interface this is termed
    `workspace index`.
  - A contiguous index that is used to access data in workspaces. This index
    refers only to spectra on this partition and is thus used in all client code
    when accessing a partitioned workspace.
  Typically, input from users or files would be in terms of spectrum numbers or
  global spectrum indices. IndexInfo is then used to translate these into a set
  of indices, whereby IndexInfo internally takes care of including all indices
  in question in the set, such that the union of sets on all partitions
  corresponds to the requested spectrum numbers or global spectrum indices.
  Client code that treats each spectrum on its own can thus be written without
  concern or knowledge about the underlying partitioning of the data.

  Note that currently IndexInfo also carries information about grouping in terms
  of detector IDs. This mechanism has been added for supporting the current
  legacy in Mantid: Grouping is stored in terms of detector IDs in ISpectrum (in
  addition to the new way of storing grouping with the class SpectrumDefinition
  in terms of indices). We also store detector IDs such that grouping
  information can be transferred to another workspace, without the need for a
  costly (and potentially impossible, since invalid IDs are allowed) rebuild of
  the sets of detector IDs that need to be part of ISpectrum.
```

**To test:**

Thorough code review.

Fixes #18522.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Indexing/src/SpectrumNumberTranslator.cpp
```diff
@@ -0,0 +1,92 @@
...
+  static_cast<void>(m_partitions.at(min));
+  static_cast<void>(m_partitions.at(max));
+
+  // The order of spectrum numbers can be arbitrary so we need to iterate.
+  std::vector<size_t> indices;
+  for (const auto &index : m_indices) {
+    if (index.first >= min && index.first <= max)
+      indices.push_back(index.second);
+  }
+  return SpectrumIndexSet(indices, m_indices.size());
```
LamarMoore: Here for example you supply `indices` and `indices.size()` could that not be retrieved internally? Then you could compare the size of the set you create with the size of indices which would tell you if there were duplicate indices. Or the argument could be a set so the user is unable to provide duplicate indices?

Framework/Indexing/src/SpectrumNumberTranslator.cpp
```diff
@@ -0,0 +1,92 @@
...
+}
+
+SpectrumIndexSet
+SpectrumNumberTranslator::makeIndexSet(GlobalSpectrumIndex min,
+                                       GlobalSpectrumIndex max) const {
+  if (min > max)
```
LamarMoore: Do you also want to check if `min < 0`

LamarMoore (commented):
Looks good, just a few questions.

Framework/Indexing/src/RoundRobinPartitioner.cpp
```diff
@@ -0,0 +1,13 @@
+#include "MantidIndexing/RoundRobinPartitioner.h"
+
+namespace Mantid {
+namespace Indexing {
+
+PartitionIndex
+RoundRobinPartitioner::doIndexOf(const GlobalSpectrumIndex index) const {
+  return PartitionIndex(static_cast<int>(static_cast<size_t>(index) %
+                                         numberOfNonMonitorPartitions()));
+}
```
SimonHeybrock: `doIndexOf` is not called or monitors in that case? So monitors are on a dedicated partition, whereas it is round-robin allocation for everything else.

Framework/Indexing/inc/MantidIndexing/IndexSet.h
```diff
@@ -0,0 +1,152 @@
...
+IndexSet<T>::IndexSet(size_t fullRange)
+    : m_size(fullRange) {}
+
+/// Constructor for a set covering the range from min to max. Range is verified
+/// at construction time.
+template <class T>
+IndexSet<T>::IndexSet(int64_t min, int64_t max, size_t fullRange) {
+  if (min < 0 || min > max)
+    throw std::logic_error("IndexSet: specified min or max values are invalid");
+  if (max >= static_cast<int64_t>(fullRange))
```
SimonHeybrock: Just for validation, you would pass the workspace size and `max` is checked automatically. You can of course still break things by passing something invalid to `fullRange`, but if the range is correct, it is guaranteed that any index in the set will be a valid workspace index. I agree that is not strictly necessary.

Framework/Indexing/inc/MantidIndexing/IndexSet.h
```diff
@@ -0,0 +1,152 @@
...
+
+/// Constructor for a set containing all specified indices. Range is verified at
+/// construction time and duplicates are removed.
+template <class T>
+IndexSet<T>::IndexSet(const std::vector<size_t> &indices, size_t fullRange)
+    : m_isRange(false) {
+  // We use a set to create unique and ordered indices.
+  std::set<size_t> index_set(indices.cbegin(), indices.cend());
+  if (!index_set.empty() && *(index_set.rbegin()) >= fullRange)
+    throw std::out_of_range("IndexSet: specified index is out of range");
```
SimonHeybrock: Empty sets are supported, so there is no error there?

Framework/Indexing/src/SpectrumNumberTranslator.cpp
SimonHeybrock: `m_indices.size()`, not `indices.size()`. I need the size of the latter for a *range* check. What you suggest would only work for a check of *duplicates*.

Framework/Indexing/src/SpectrumNumberTranslator.cpp
SimonHeybrock: No, `GlobalSpectrumIndex` is `unsigned` (`size_t`).

LamarMoore:
If @OwenArnold  has no further comments :shipit:

Framework/Indexing/inc/MantidIndexing/DetectorID.h
```diff
@@ -34,16 +33,13 @@ namespace Indexing {
...
 */
-template <class T1, class T2,
-          class = typename std::enable_if<std::is_integral<T1>::value &&
-                                          std::is_integral<T2>::value>::type>
-std::vector<T2> makeRange(T1 first, T2 last) {
-  std::vector<T2> vec(last - static_cast<T2>(first) + 1);
-  std::iota(vec.begin(), vec.end(), static_cast<T2>(first));
-  return vec;
-}
+struct MANTID_INDEXING_DLL DetectorID
```
OwenArnold: @SimonHeybrock I like it. I assume this will now give us type safety, say for catching incompatible operations at compile times?

Framework/Indexing/inc/MantidIndexing/IndexType.h
```diff
@@ -0,0 +1,74 @@
...
+
+private:
+  Int m_data;
+};
+
+} // namespace detail
+} // namespace Indexing
+} // namespace Mantid
+
+#endif /* MANTID_INDEXING_INDEXTYPE_H_ */
```
OwenArnold: Looks good!

Framework/Indexing/inc/MantidIndexing/DetectorID.h
SimonHeybrock: @OwenArnold Yes, up to a point. Conversion from `DetectorID -> int` is explicit, but I am currently allowing implicit conversion from `int -> DetectorID`. I initially forbid both, but it lead to a lot of bloat.

OwenArnold: That's a shame.

However I think having proper types for the different index types is conceptually much better and clearer. I would hope this change doesn't affect performance but I think that would be worth checking.

Framework/Indexing/src/IndexInfo.cpp
```diff
@@ -133,10 +123,81 @@ void IndexInfo::setSpectrumDefinitions(
...
+void IndexInfo::makeSpectrumNumberTranslator() const {
+  // To support legacy code that creates workspaces with duplicate spectrum
+  // numbers we are currently creating the SpectrumNumberTranslator only when
+  // needed.
+  // Note that the lazy initialization implies that no sharing of
+  // SpectrumNumberTranslator will be set up for workspaces created before the
+  // first access.
+  // TODO We are not setting monitors currently. This is ok as long as we have
+  // exactly one partition.
+  PartitionIndex partition = 0;
```
OwenArnold: I take it this hard-coded value will be calculated somehow at a later date. If you didn't have a follow up PR, I would have said that a comment to that effect would be needed here.

OwenArnold:
This also looks Ok to me.

LamarMoore:
@Owen, there is something approaching that [here](https://github.com/DMSC-Instrument-Data/documents/blob/master/meeting_notes/March_2017/indexing.JPG)

PR #19123, opened 2017-03-13
Reflectometry Interface: Fixed bug to prevent global options being applied when disabled

### LINKED ISSUE #18708: Don't apply global options if settings are disabled.
When `Instrument` settings and/or `Experiment` settings are disabled, these options should not be applied in the reduction.

- Fix this.
- Add unit tests.
- Update the interface documentation.

PranavBahuguna:
PR is as described in the related issue #18708.

**To test:**

- The reflectometry interface can be accessed from Interfaces->Reflectometry->ISIS Reflectometry.
- Go to the `Settings` tab and enter some values in the entries for both `Experiment` and `Instrument` sections (or just use their respective `Get Defaults` button).
- Uncheck the two sections, disabling all entries.
- In the `Runs` tab,  load `INTER_NR_test2.tbl` (from the sample INTER dataset) and process the first group. Ensure that none of the inputs to the settings from before are being applied in the reduction.

Fixes #18708.

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


raquelalvarezbanos:
Functionality works as described. However, looking at code changes, I see you are adding the checks for enabled settings in `QtReflSettingsView`. I think it would be better if this check was added to `ReflSettingsPresenter`, specifically to methods `ReflSettingsPresenter::getTransmissionOptions()`, `ReflSettingsPresenter::getReductionOptions()` and `ReflSettingsPresenter::getStitchOptions()`, all you'll need is to define two methods in the view, something like `bool experimentSettingsEnabled()` and `bool instrumentSettingsEnabled()` and use them in the presenter to check whether the settings are enabled or not. This way, we can avoid a bunch of `if`s when we need to retrieve the pre-processing/processing/post-processing settings.

PranavBahuguna:
I think I tried to do something like that. The problem is that some properties (like `AnalysisMode` and `Start/EndOverlaps`) are retrieved in multiple methods (`getReductionOptions` and `getTransmissionOptions`). Under your suggestion, if 'Experiment Settings' was disabled, but `Instrument Settings` was not, `getTransmissionOptions()` would be prevented from running despite there being non-disabled options it could still use.

raquelalvarezbanos:
I am not sure I understand the problem, I think something like the following should work:

```
std::string ReflSettingsPresenter::getTransmissionOptions() const {

  std::vector<std::string> options;

  if (m_view->experimentSettingsEnabled()) {

      // Add analysis mode
      auto analysisMode = m_view->getAnalysisMode();
      if (!analysisMode.empty())
        options.push_back("AnalysisMode=" + analysisMode);

      // Other options in experiment settings
      // ...
    }

    if (m_view->instrumentSettingsEnabled()) {

      // Add monitor integral min
      auto monIntMin = m_view->getMonitorIntegralMin();
      if (!monIntMin.empty())
          options.push_back("MonitorIntegrationWavelengthMin=" + monIntMin);

      // Other options in instrument settings
      // ...
    }
}
```

the same for `ReflSettingsPresenter::getReductionOptions()` and `ReflSettingsPresenter::getStitchOptions()`.

If as you say, `Experiment Settings` was disabled, but `Instrument Setting`s was not, `getTransmissionOptions(`) would still be able to run with the non-disabled `Instrument Settings`, unless I am missing something?

PranavBahuguna:
The approach you gave seems workable but I realised there is another problem - some of the other settings (such as the polarisation constants `CRho`, `CAlpha`, `CAp`, `CPp`) can also be disabled externally, so checking if `Experiment/Instrument Settings` is checked is insufficient. It just seems a lot easier to me to just check if each individual setting is disabled or not before trying to use it.

raquelalvarezbanos:
What do you mean by "disabled externally"?

PranavBahuguna:
They can be disabled depending on what instrument is currently selected in the 'Runs' tab.

raquelalvarezbanos:
I think checking if `Experiment/Instrument Settings` is checked should be enough with just a little extra work. I am going to add a commit to illustrate what I am suggesting, so you can have a look and let me know if it doesn't work as expected and I'll revert it.

Here are a few cases to test the behaviour of polarization corrections:

1. `Experiment Settings` are disabled and:
  a. You change from `INTER` to `POLREF` -> nothing happens, polarization corrections stay disabled
  b. You change from `POLREF` to `INTER` -> nothing happens, polarization corrections stay disabled

2. `Experiment Settings` are enabled and:
  a. You change from `INTER` to `POLREF` -> polarization corrections are enabled
  b. You change from `POLREF` to `INTER` -> polarization corrections are disabled

If possible, I think it's better if we add the checks to the presenter for two reasons:

- We can avoid ~hundreds of `if`s when users reduce hundreds of runs
- We can write unit tests for the presenter class to check that if experiment/instrument settings are disabled they are not returned as global options.

raquelalvarezbanos:
Sorry I didn't read your last comment... I added a commit, please check the changes and if you are happy with functionality, fill in the new unit test and fix clang format and other possible build errors.

PranavBahuguna:
I've discovered a problem with the way you implemented this. Suppose starting on `INTER` you disable experiment settings.  If you change instrument to say `POLREF` and check experiment settings back again, the polarisation correction options will still be disabled.

I will try to find a solution for this.

PranavBahuguna:
Actually the interface was buggy in a similar way beforehand (due to polarisation corrections being enabled/disabled by instrument changes).

raquelalvarezbanos:
Looks good.

:shipit:

PR #19127, opened 2017-03-13
Indirect Sample Changer

### LINKED ISSUE #18844: Indirect QuickRun - Sample Changer
A sample changer has been implemented on IRIS and OSIRIS that holds 3 samples. When the temperature of the environment is changed, a run is recorded on each of the three samples in turn as the sample changer rotates the position. The temperature and position of the sample is recorded in the logs. MSG have asked for an algorithm similar to [EnergyWindowScan](http://docs.mantidproject.org/nightly/algorithms/EnergyWindowScan-v1.html) which can reduce and perform basic analysis on the samples, given the position of the sample changer.

louisemccann:
A new algorithm to streamline the workflow required to reduce and analyse data from a sample changer with multiple different samples.

**To test:**

Input if no access to ISIS data archive:
[SampleChanger.zip](https://github.com/mantidproject/mantid/files/839003/SampleChanger.zip)

```python
IndirectSampleChanger(FirstRun=72462, LastRun=72465, NumberSamples=2, Instrument='IRIS',
                      Analyser='graphite', Reflection='002', SpectraRange='3, 50', ElasticRange='-0.5, 0',
                      InelasticRange='0, 0.5', GroupingMethod='All')
```

Fixes #18844.

_Release Notes have been updated_

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/IndirectSampleChanger.py
```diff
@@ -0,0 +1,171 @@
...
+                             validator=StringListValidator(['IRIS', 'OSIRIS']),
+                             doc='Instrument name')
+        self.declareProperty(name='Analyser', defaultValue='',
+                             validator=StringListValidator(['graphite', 'mica', 'fmica']),
+                             doc='Analyser bank used during run.')
+        self.declareProperty(name='Reflection', defaultValue='',
+                             validator=StringListValidator(['002', '004', '006']),
+                             doc='Reflection number for instrument setup during run.')
+
+        self.declareProperty(name='FirstRun', defaultValue=1,
```
martyngigg: The `defaultValue` here seems a little confusing. Are they set to these just so that they are integer properties?

If so can we do something like `self.declareProperty("FirstRun",-1,InttBoundedValidator(lower=0)) (
https://www.mantidproject.org/Algorithm_Property_Validation) instead?

martyngigg:
I think this is nearly there.

One thing that I think would be a good addition is if we could add a `validateInputs` method? For example if I set `FirstRun=72462` & `LastRun=72461` then I get an error:

```python
Error in execution of algorithm IndirectSampleChanger:
list index out of range
  at line 82 in '/media/data1/source/github/mantidproject/mantid/Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/IndirectSampleChanger.py'
```

It's not too obvious at first glance what this means. I think the easiest thing to get wrong here is how the parameters `FirstRun`, `LastRun` & `NumberSamples` relate to each other so I'd focus on sanity checking those if we can.

SpencerHowells:
Good idea! Obviously LastRun must be > FirstRun. Number of runs also > NumberSamples. I also thought that number of runs should be a multiple of NumberSamples but then users may stop the sequence early and it would not be true.

Framework/PythonInterface/test/python/plugins/algorithms/WorkflowAlgorithms/IndirectSampleChangerTest.py
```diff
@@ -0,0 +1,52 @@
...
+        Basic test for sample changer
+        """
+        IndirectSampleChanger(FirstRun=72462, LastRun=72465, NumberSamples=1, Instrument='IRIS', Analyser='graphite',
+                              Reflection='002', SpectraRange='3, 50', ElasticRange='-0.5, 0',
+                              InelasticRange='0, 0.5', GroupingMethod='All')
+        scan_ws = mtd['iris72462_to_72465_s0_scan_eisf']
+        self.assertEqual(round(scan_ws.readY(0)[0], 7), 1.0519965)
+        self.assertEqual(round(scan_ws.readY(1)[0], 7), 1.0452707)
+
+    def test_multiple_samples(self):
```
martyngigg: We've got some good tests here. It's good practice to also add tests that verify failure behaviour.

Can we add some tests that test the invalid scenarios being checked by `validateInputs`?

martyngigg:
The new tests look good.

Happy to :shipit:

PR #19147, opened 2017-03-15
Bad double checked locking in ExperimentInfo::detectorInfo().

SimonHeybrock:
Locking in this way without memory fences is broken. We fix this by simply avoiding it.

Note that `ExperimentInfo::spectrumInfo()` suffers from the same problem. It will be dealt with in another pull request, since it will be easier to deal with once some unrelated work on Instrument-2.0 is finished.

**To test:**

Code review.

No issue, not release notes.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/src/ExperimentInfo.cpp
+++ b/Framework/API/src/ExperimentInfo.cpp
@@ -63,6 +63,8 @@ ExperimentInfo::ExperimentInfo()
       sptr_instrument(new Instrument()),
       m_detectorInfo(boost::make_shared<Beamline::DetectorInfo>()) {
   m_parmap->setDetectorInfo(m_detectorInfo);
+  m_detectorInfoWrapper = Kernel::make_unique<DetectorInfo>(
+      *m_detectorInfo, getInstrument(), m_parmap.get());
 }
 
 /**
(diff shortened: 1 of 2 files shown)


Framework/API/src/ExperimentInfo.cpp
```diff
@@ -1115,17 +1118,11 @@ ExperimentInfo::getInstrumentFilename(const std::string &instrumentName,
 
 /** Return a const reference to the DetectorInfo object.
  *
- * Any modifications of the instrument or instrument parameters will invalidate
+ * Setting a new instrument via ExperimentInfo::setInstrument will invalidate
```
soininen: Just a question: wouldn't this be a legitimate use case for `std::weak_ptr` or similar?

SimonHeybrock: You could probably figure out a way to avoid invalid references, but the result would still be nonsense: If you have two threads, where one sets a new instrument (invalidating the reference to `DetectorInfo` in the current implementation), and the other thread loops over all detectors, to e.g., read the position, what should the outcome be?

- The positions of the old instrument?
- The positions of the new instrument?
- Some positions from the old positions, the rest from the new instrument?

What I am saying is that doing such things is a data race with unpredictable outcome (unless you do global locking of all instrument access), so I would rather have crashed from accessing invalid references: that at least tell me that something is wrong.

soininen:
These changes seem to touch the code in all the relevant places. :shipit:

PR #19153, opened 2017-03-15
ComponentInfo

### LINKED ISSUE #19015: Basic ComponentInfo required
As indicated on a recent slack discussion around scanning. We need to start introducing a basic `ComponentInfo` into `Beamline`. Specifically we need an analogue to the `getDetectorsInBank` functionality currently contained in `Instrument`. 

A good starting point might be something like this.

```
std::vector<size_t> Beamline::ComponentInfo::getDetectorIndexes(size_t componentIndex)
``` 

I would prefer to avoid using names as in the current signature unless this proves too cumbersome.

Some of this functionality is already available in the prototype. https://github.com/DMSC-Instrument-Data/instrument-prototype/blob/master/cow_instrument/ComponentInfo.h#L161-L162 and the `LinkedTreeParser` may also prove very useful for compatibility with `Instrument1.0`.

OwenArnold:
Basic `ComponentInfo` addition

**To test:**

Fixes #19015

No release notes additions needed.

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/API/src/ExperimentInfo.cpp
```diff
@@ -176,6 +184,24 @@ void checkDetectorInfoSize(const Instrument &instr,
...
+  InfoComponentVisitor visitor(detectorInfo);
+  std::vector<size_t> detectorIndexes; // Not strictly needed at this level
+
+  // Register everything via visitor
+  oldInstr.registerContents(visitor, detectorIndexes);
+  // Extract component ids. We need this for the ComponentInfo wrapper.
+  componentIds = visitor.componentIds();
+
+  return boost::make_shared<Mantid::Beamline::ComponentInfo>(
+      visitor.componentDetectorIndexes());
```
SimonHeybrock: Minor point, but by having `componentIds` and `componentDetectorIndices` as members of the visitor we are doing an extra copy here.

Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
```diff
@@ -0,0 +1,48 @@
...
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+class MANTID_BEAMLINE_DLL ComponentInfo {
+
+private:
+  std::vector<std::vector<size_t>> m_detectorIndexes;
```
SimonHeybrock: - Use a `cow_ptr`, this can be shared between all workspaces.
- Using a vector of vectors feels very inefficient. We would have every detector index at least twice (once in the root component, once in a bank, more often if there are more levels), and an empty vector for each detector, i.e., potentially many millions! Have you considered using a flat vector?

Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
```diff
@@ -0,0 +1,48 @@
...
+
+public:
+  ComponentInfo() = default;
+  ComponentInfo(const std::vector<std::vector<size_t>> &detectorIndexes);
+  std::vector<size_t> detectorIndexes(size_t componentIndex) const;
```
SimonHeybrock: `detectorIndices` (can we use `indices` everywhere, instead of `indexes`?).

SimonHeybrock: Returning a vector by value seems inefficient. In combination with my suggestion above regarding using a flat vector, this could return something similar to `gsl::span` (i.e., some iterable object that holds a start and end index into `m_detectorIndices`).

Framework/Geometry/inc/MantidGeometry/Instrument/ComponentVisitor.h
```diff
@@ -0,0 +1,51 @@
...
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+class ComponentVisitor {
+public:
+  virtual void
+  registerComponentAssembly(const ICompAssembly &assembly,
+                            std::vector<size_t> &detectorIndexes) = 0;
```
SimonHeybrock: I presume you chose to use the visitor pattern to avoid a ton of `dynamic_cast`?

Framework/Geometry/src/Instrument/DetectorGroup.cpp
```diff
@@ -454,5 +455,15 @@ size_t DetectorGroup::index() const {
   throw std::runtime_error("A DetectorGroup cannot have an index");
 }
 
+void DetectorGroup::registerContents(
```
SimonHeybrock: As far as I know a `DetectorGroup` is never part of an instrument, should we simply throw here?

Framework/API/src/InfoComponentVisitor.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+  /*
+   * Add the detector index to the parent collection, but as each detector
+   * is a leaf this components detector indexes are empty.
+   */
+  parentDetectorIndexes.push_back(m_detectorInfo.indexOf(detector.getID()));
+  m_componentDetectorIndexes.emplace_back(std::vector<size_t>());
+  m_componentIds.emplace_back(detector.getComponentID());
+}
+InfoComponentVisitor::~InfoComponentVisitor() {}
```
SimonHeybrock: Empty -> remove?

SimonHeybrock (commented):
@OwenArnold Looks good, apart from smaller comments I have only one key concern, regarding the data structure (and return values) for component-detector indices, see below:

Framework/Geometry/inc/MantidGeometry/Instrument/ComponentVisitor.h
OwenArnold: I actually implemented as a load of dynamic_casts originally. I then refactored it to this. I put in some timings before I did the refactoring and found it to be more than x2 faster doing it this way. That was for a relatively small instrument, for larger instruments it would probably make an even bigger difference.

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -233,7 +259,6 @@ makeDetectorInfo(const Instrument &oldInstr, const Instrument &newInstr) {
         std::move(positions), std::move(rotations), monitors);
   }
 }
-}
```
OwenArnold: 👍

Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
OwenArnold: Agree on the `cow_ptr`. Not sure what you have in mind to get around this?

I did consider only storing the immediate child detectors of each bank, but I think that's going to tie the queries too much into how each instrument is structured.

SimonHeybrock: @OwenArnold 

Simple version:

```cpp
std::vector<size_t> m_detectorIndices;
std::vector<std::pair<size_t, size_t>> m_ranges;

span detectorIndices(const size_t index) const {
  return span(m_detectorIndices, m_ranges(index));
}
```

`m_ranges` would have size equal to the number of components. We can make that much much shorter by not storing ranges for detectors, because those would be empty. That can be accomplished by defining component indices such that all detectors come first (we discussed that a while ago, for other reasons such as having only one vector of positions in `ComponentInfo`, but not in `DetectorInfo`). The index to access `m_ranges` is then obtained by a simple offset:

```cpp
span detectorIndices(const size_t index) const {
  if(index < m_numberOfDetectors)
    return span();
  return span(m_detectorIndices, m_ranges(index - m_numberOfDetectors));
}
```

SimonHeybrock: The additional advantage of reordering would be that for detectors we have `detectorIndex == componentIndex`.

SimonHeybrock: Note that you need to walk the instrument tree depth-first, otherwise you cannot define a flat vector `m_detectorIndices`.

Framework/API/src/InfoComponentVisitor.cpp
```diff
@@ -0,0 +1,75 @@
...
+   */
+  m_componentDetectorIndexes.emplace_back(std::vector<size_t>());
+  m_componentIds.emplace_back(component.getComponentID());
+}
+void InfoComponentVisitor::registerDetector(
+    const IDetector &detector, std::vector<size_t> &parentDetectorIndexes) {
+
+  /*
+   * Add the detector index to the parent collection, but as each detector
+   * is a leaf this components detector indexes are empty.
```
SimonHeybrock: Is this a good definition? Shouldn't a component include itself, i.e., this "component's detector indices" is simply the detector index?

Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
SimonHeybrock: If we define `detectorIndices` a bit differently (see my new comment on registering detectors), i.e., include the detector itself, we have to change this a bit, but it should still be possible (maybe `span` can then reference the vector of detector indices in `DetectorInfo`?).

Framework/Geometry/src/Instrument/DetectorGroup.cpp
OwenArnold: Yes. I think you're right.

Framework/API/src/InfoComponentVisitor.cpp
OwenArnold: I think that depends slightly on what the client expects when using `detectorIndexes`, but I think you're right.

OwenArnold:
Made some of the minor changes suggested by @SimonHeybrock. Still need to look at the range suggestion.

Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
OwenArnold: Interesting idea. I think this should be worth a try.

SimonHeybrock:
Latest changes look good.

ianbush (commented):
This looks good, and code wise I do not have any specific comments to add.

I realise it is work in progress, but it would be good to add more documentation, for example in the class descriptions and doxygen comments. I am not sure the motivation for some parts (like the `ComponentVisitor`) are so clear to someone not familiar with the instrument prototype.

It would be good to add some independent tests for `InfoComponentVisitor` too.

Framework/Geometry/CMakeLists.txt
```diff
@@ -199,6 +199,7 @@ set ( INC_FILES
 	inc/MantidGeometry/Instrument/CompAssembly.h
 	inc/MantidGeometry/Instrument/Component.h
 	inc/MantidGeometry/Instrument/ComponentHelper.h
+  inc/MantidGeometry/Instrument/ComponentVisitor.h
```
ianbush: Indentation has gone wonky here.

OwenArnold:
Thanks for the comments @ianbush I'll address those.

Framework/API/src/InfoComponentVisitor.cpp
```diff
@@ -4,74 +4,113 @@
...
+ */
+void InfoComponentVisitor::registerDetector(const IDetector &detector) {
+
+  const auto detectorIndex =
+      m_detectorIdToIndexMapperFunction(detector.getID());
+
+  /* Already allocated we just need to index into the inital front-detector
+   * part of the collection
+  */
+  m_componentIds[m_detectorCounter++] = detector.getComponentID();
```
SimonHeybrock: I suppose this does not guarantee that the order of detector indices is the same as the order of component indices? That is detector ID 17 might be detector index 4, but component index 8?

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -177,6 +185,24 @@ void checkDetectorInfoSize(const Instrument &instr,
...
+makeComponentInfo(const Instrument &oldInstr,
+                  const API::DetectorInfo &detectorInfo,
+                  std::vector<Geometry::ComponentID> &componentIds) {
+
+  InfoComponentVisitor visitor(
+      detectorInfo.size(),
+      std::bind(&DetectorInfo::indexOf, &detectorInfo, std::placeholders::_1));
+
+  // Register everything via visitor
+  oldInstr.registerContents(visitor);
```
SimonHeybrock: Have you tried using the bare instrument (without parameter map) here? That might be faster, since `IComponent::getChildren` will then not need to do allocation via `ParComponentFactory`.

SimonHeybrock (commented):
@OwenArnold Looks nice, apart from two comments below, I am still a bit concerned about returning by value from `ComponentInfo::detectorIndices`. I think the best way to test this would be to use `ComponentInfo` in `DetectorInfo::setPosition`/`setRotation` and then run some `SCDCalibratePanels` example -- that seemed to be the heaviest user of moving/rotating parents.

Framework/API/src/InfoComponentVisitor.cpp
OwenArnold: This guarantee is not currently provided. I think it could be allowed for. What specific use case do you have for this?

SimonHeybrock: It would make translation between detector index and component index trivial.

Framework/API/src/ExperimentInfo.cpp
OwenArnold: @SimonHeybrock Thanks for the tip. Yes we don't need any parameterised information for this at present. I'll make that change.

OwenArnold:
@SimonHeybrock regarding you comment with `ComponentInfo::detectorIndices` returning by value. This seems to be right from the API perspective, but I agree for a full instrument traversal, it could be expensive. I don't want to have `ComponentInfo` unnecessarily caching anything to return a reference, that essentially would lead to the same problem anyway. Maybe we should talk about this as I have some concerns.

OwenArnold:
@SimonHeybrock. Regarding performance, I've done some work looking at the use of detectorinfo, setting positions in a bank, via `CalibrationHelpers` with real data.

Compared to the current implementation, I can get this about x2 faster using the new methods by adding a method to detectorInfo, that accepts a list of detectorIndices to adjust the position for by reference. 

```cpp
auto detIndices = componentInfo.detectorIndices(componentInfo.indexOf(bank>getComponentID()));
detectorInfo.setPositions(detIndices, rotatedPos); // Trial only
```

In any case the generation of the detector indices represents only a very small fraction of this according to the timings, so I don't think we should worry about the return by copy.

OwenArnold:
Regarding the guaranteed ordering, i.e. detectorIndices == componentIndices for the first n component indices. I think this would be difficult to implement with the current state of things. Most notably the construction of the ranges of detector indices in the bank. This can currently be done in a very straight forward way by noting the size of the detectindices list before and after the registration of the bank. (see snippet from InfoComponentVisitor below)
```cpp 
  const size_t detectorStart = m_detectorIndices.size();
  for (const auto &child : assemblyChildren) {
    // register everything under this assembly
    child->registerContents(*this);
  }
  const size_t detectorStop = m_detectorIndices.size();
```

I would suggest we start by **not** providing this guarantee to start with and adding it at a later stage. One way I think this could be possible is by harmonising the creation of the `DetectorInfo` and `ComponentInfo` to both happen via the `InfoComponentVisitor`. This is essentially what is done in the prototype. 

https://github.com/DMSC-Instrument-Data/instrument-prototype/blob/master/cow_instrument/LinkedTreeParser.cpp#L37-L42

SimonHeybrock:
@OwenArnold If we return by value, why do you need to construct ranges for detectors? For detectors, you would simply return `std::vector<size_t>{componentIndex}`, automatically get the correct ordering, and save setup cost and memory since you do not need to set up ranges for detectors.

Framework/API/src/ExperimentInfo.cpp
SimonHeybrock: @OwenArnold Did you try this?

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -273,9 +304,11 @@ void ExperimentInfo::setInstrument(const Instrument_const_sptr &instr) {
 Instrument_const_sptr ExperimentInfo::getInstrument() const {
   populateIfNotLoaded();
   checkDetectorInfoSize(*sptr_instrument, *m_detectorInfo);
+  // checkComponentInfoSize(*sptr_instrument, *m_componentInfo);
   auto instrument = Geometry::ParComponentFactory::createInstrument(
       sptr_instrument, m_parmap);
   instrument->setDetectorInfo(m_detectorInfo);
+  // instrument->setComponentInfo(m_componentInfo);
```
SimonHeybrock: Leftover commented code?

Framework/API/src/InfoComponentVisitor.cpp
```diff
@@ -0,0 +1,128 @@
...
+void InfoComponentVisitor::registerDetector(const IDetector &detector) {
+
+  const auto detectorIndex =
+      m_detectorIdToIndexMapperFunction(detector.getID());
+
+  /* Unfortunately Mantid supports having detectors attached to an
+   * instrument that have an an invalid or duplicate detector id.
+   * We do not register detectors with the same id twice.
+   */
+  if (m_componentIds[detectorIndex] == nullptr) {
```
SimonHeybrock: I suppose this breaks if the order in which detectors were added to the instrument is not equal to the order in which components are visited. Did you have to fix any tests due to this?

OwenArnold: @SimonHeybrock. That is correct in the case of duplicate detector ids, the first one visited would be registered, and subsequent ones would be rejected. I didn't need to fix anything because of this. I would say that.

1. Longer term, the `InfoComponentVisitor` should be used to create all our info layers. That would give a nice seam with `Instrument1.0` and would ensure that we can create things in a well defined order.
2. Hopefully the cases where this nullptr check are actually required is minimal. I consider IDFs which support duplication of detector indexes to be invalid. I'm asking developers to fix these as I find them (as in the BASIS case).

Framework/API/src/ExperimentInfo.cpp
OwenArnold: Yes I'll remove the dead code.

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include <utility>
+#include <vector>
+
+namespace Mantid {
+namespace Beamline {
+
+ComponentInfo::ComponentInfo()
+    : m_detectorIndices(boost::make_shared<std::vector<size_t>>(0)),
+      m_ranges(
+          boost::make_shared<const std::vector<std::pair<size_t, size_t>>>(0)),
```
SimonHeybrock: No need to allocate?

Framework/API/src/InfoComponentVisitor.cpp
```diff
@@ -0,0 +1,128 @@
...
+const std::vector<std::pair<size_t, size_t>> &
+InfoComponentVisitor::componentDetectorRanges() const {
+  return m_ranges;
+}
+
+/**
+ * @brief InfoComponentVisitor::detectorIndices
+ * @return detector indices in the order in which they have been visited
+ */
+const std::vector<size_t> &InfoComponentVisitor::detectorIndices() const {
```
SimonHeybrock: Shouldn't this be removed now that the ordering is guaranteed?

Framework/API/src/ExperimentInfo.cpp
OwenArnold: Hmm. I thought I had made that change, but you're right it's using the parameterised one. Yes we don't need the parameterised information. I'll change it.

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -0,0 +1,42 @@
...
+          boost::make_shared<std::vector<size_t>>(detectorIndices)),
+      m_ranges(boost::make_shared<const std::vector<std::pair<size_t, size_t>>>(
+          ranges)),
+      m_size(detectorIndices.size() + ranges.size()) {}
+
+std::vector<size_t>
+ComponentInfo::detectorIndices(const size_t componentIndex) const {
+  if (componentIndex < m_detectorIndices->size()) {
+    // This is a single detector. Just return the corresponding index.
+    return std::vector<size_t>{(*m_detectorIndices)[componentIndex]};
```
SimonHeybrock: Now that ordering is guaranteed, should this be `return {componentIndex};`?

Framework/API/src/ExperimentInfo.cpp
SimonHeybrock: @OwenArnold Oh, I see, you changed it at the call site below, now passing `sptr_instrument`.

Framework/Beamline/test/ComponentInfoTest.h
```diff
@@ -0,0 +1,78 @@
...
+  void test_empty_size() {
+    ComponentInfo info;
+    TS_ASSERT_EQUALS(info.size(), 0);
+  }
+
+  void test_empty_throw_out_of_range() {
+    ComponentInfo info;
+    TSM_ASSERT_THROWS(
+        "Empty ComponenetInfo, so any attempt to use this should throw",
+        info.detectorIndices(0), std::out_of_range &);
```
SimonHeybrock: I am not doing any range checks in `DetectorInfo` or `SpectrumInfo` (behaving basically like `std::vector::operator[]`). Should we keep that consistent?

Framework/API/src/ExperimentInfo.cpp
OwenArnold: @SimonHeybrock `makeComponentInfo` now uses `sptr_instrument`. I think you might have been looking at out of date code?

Framework/Beamline/test/ComponentInfoTest.h
OwenArnold: Agreed. this was introduced as an initialisation check, given that the ComponentInfo has a default constructor.

SimonHeybrock: @OwenArnold Yes, so does `DetectorInfo`, but that yields `size() == 0` so that is covered.

Framework/API/src/InfoComponentVisitor.cpp
OwenArnold: No. I've renamed this to `assemblySortedDetectorIndices()` in order for the range (start, stop) to work, we need to have an assembly grouped collection of detector indexes. We do not however need to use that when we are return detector indices (your point below).

OwenArnold:
The system tests are broken until #19262 can be merged

SimonHeybrock:
Latest changes look good. The additions in this PR are all clean and a good amount of tests has been added. :shipit:

NickDraper:
All looks good, in depth code review and conversations have occured, happy to merge.

PR #19168, opened 2017-03-17
Add function to compare workspaces in Python

### LINKED ISSUE #19167: Add function for workspace object comparison in Python
Sometimes we want to know if two workspaces in Python are equal, but this is not possible at the moment since both `is` and `==` will not perform an address comparison of the underlying workspace.

We should provide a free function which takes two workspace pointers and performs a address comparison.

AntonPiccardoSelg:
Fix #19167

For workflow algorithms which can be defined in Python we somtimes want to check if the input workspace is the same as the output workspace (as is done sometimes on the `c++` side). A naive `ws1 is ws2` or `ws1 == ws2` does not produce the correct result since it only compares the Python wrapper and not the underlying workspace. Overloading `==` is not a good option since `==` implies a value comparison we want to actually know if the we are dealing with the same workspace and not with a clone. Overloading `is` is not possible, hence a free function seems to be the clearest option here.

### To Test

Run a the Python script below

```python
ws1 = CreateSampleWorkspace()
ws2 = CreateSampleWorkspace()
print("Expect to be the same, ie should be True and is: {}".format(isSameWorkspaceObject(ws1, ws1)))
print("Expect to be not the same, ie should be False and is: {}".format(isSameWorkspaceObject(ws1, ws2)))
# comment added to aid cut and paste
```
Play around and check if the comparisons make sense.

*Does not need to be in the release notes, since this is a developer facing change*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/PythonInterface/mantid/api/src/api.cpp.in
+++ b/Framework/PythonInterface/mantid/api/src/api.cpp.in
@@ -5,12 +5,15 @@
 #include <boost/python/def.hpp>
 #include <boost/python/module.hpp>
 #include <boost/python/docstring_options.hpp>
+#include <boost/python/args.hpp>
+
 
 // See http://docs.scipy.org/doc/numpy/reference/c-api.array.html#PY_ARRAY_UNIQUE_SYMBOL
 #define PY_ARRAY_UNIQUE_SYMBOL API_ARRAY_API
 #include <numpy/arrayobject.h>
 
 #include "MantidAPI/AlgorithmFactory.h"
+#include "MantidAPI/Workspace.h"
 #include "MantidPythonInterface/api/Algorithms/RunPythonScript.h"
 
 namespace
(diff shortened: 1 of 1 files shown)


KarlPalmen:
I tested with the suggested script and
```python
base = CreateSampleWorkspace(Function='Powder Diffraction', NumBanks=4, BankPixelWidth=1)

wsNames = []

for t in range (10,110,10):
    wsName = "temp" + str(t)
    wsNames.append(wsName)
    MultiplyRange(base,75,99,t/100.0,OutputWorkspace=wsName)
    AddSampleLog(wsName,"Temp",str(t),"Number","Kelvin")

TestGroup = GroupWorkspaces(wsNames)
print("Expect to be the same, ie should be True and is: {}".format(isSameWorkspaceObject(TestGroup, TestGroup)))
print("Expect to be the diferent, ie should be False and is: {}".format(isSameWorkspaceObject(base, TestGroup)))
print("Expect to be the same, ie should be True and is: {}".format(isSameWorkspaceObject(base, base)))
#
```

KarlPalmen:
:shipit:

PR #19169, opened 2017-03-17
Added volume() method to Geometry::Object

### LINKED ISSUE #19154: Shape volume calculation
A method to calculate the volume of a shape could be added to `Mantid::Geometry::Object`. For arbitrary shapes, the volume can be calculated from the triangulated mesh if it is available.

soininen:
This PR adds a new method `Mantid::Geometry::Object::volume()`. As the name states, the method calculates the volume of the object. In case of simple shapes, the volume is calculated exactly, otherwise the method falls back to Monte Carlo integration. Calculating the volume from the triangulated mesh was also experimented during the development but discarded due to problems with inconsistent face normals.

**To test:**

The method is exported to Python which should ease testing. Unfortunately, at the moment it seems one can only examine detectors:
```python
in4 = LoadEmptyInstrument(InstrumentName='IN4')
volume = in4.getInstrument().getDetector(0).shape().volume()
print('Volume of monitor in IN4: {0}'.format(volume))
```

Fixes #19154.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Geometry/src/Objects/Object.cpp
```diff
@@ -1475,6 +1476,108 @@ double Object::ConeSolidAngle(const V3D &observer,
...
+}
+
+/**
+ * The volume is calculated using the Monte Carlo method.
+ * @returns The volume of this object.
+ */
+double Object::monteCarloVolume() const {
+  const auto &boundingBox = getBoundingBox();
+  if (boundingBox.isNull()) {
+    return 0;
```
quantumsteve: Would one need to check for a return value of zero? Would an exception be a better choice?

Framework/Geometry/src/Objects/Object.cpp
```diff
@@ -1475,6 +1476,108 @@ double Object::ConeSolidAngle(const V3D &observer,
...
+  const double boundingDy = boundingBox.yMax() - boundingBox.yMin();
+  const double boundingDz = boundingBox.yMax() - boundingBox.yMin();
+  PARALLEL {
+    const auto threadCount = PARALLEL_NUMBER_OF_THREADS;
+    const auto currentThreadNum = PARALLEL_THREAD_NUMBER;
+    size_t blocksize = maxIterations / threadCount;
+    if (currentThreadNum == threadCount - 1) {
+      // Last thread may do some extra work.
+      blocksize = maxIterations - (threadCount - 1) * blocksize;
+    }
```
quantumsteve: What about adding one iteration to multiple threads?

Framework/Geometry/src/Objects/Object.cpp
```diff
@@ -1475,6 +1476,108 @@ double Object::ConeSolidAngle(const V3D &observer,
...
+    // The Mersenne twister randon number engine has poor performance
+    // with discard(). Use more suitable engine for parallelization.
+    std::ranlux48 rnEngine;
```
quantumsteve: We generally prefer using boost for random numbers so that the results are identical on each platform.

Framework/Geometry/src/Objects/Object.cpp
```diff
@@ -1475,6 +1476,108 @@ double Object::ConeSolidAngle(const V3D &observer,
...
+      rnd = rnDistribution(rnEngine);
+      const double z = boundingBox.zMin() + rnd * boundingDz;
+      if (isValid(V3D(x, y, z))) {
+        ++hits;
+      }
+    }
+    // Collect results.
+    PARALLEL_CRITICAL(monteCarloVolume) {
+      total += localTotal;
+      totalHits += hits;
```
quantumsteve: make `total` and `totalHits` atomic and remove the `PARALLEL_CRITICAL` section.

Framework/Geometry/src/Objects/Object.cpp
```diff
@@ -1475,6 +1476,108 @@ double Object::ConeSolidAngle(const V3D &observer,
...
+  const auto &boundingBox = getBoundingBox();
+  if (boundingBox.isNull()) {
+    return 0;
+  }
+  int total = 0;
+  int totalHits = 0;
+  // The maximum number of iterations is chosen by the
+  // Stetson-Harrison method. This means that the following
+  // number may or may not work in every scenario.
+  const int maxIterations = 1000000;
```
quantumsteve: How much work would it be to calculate an uncertainty in the volume and iterate until it reaches some criteria?

Framework/Geometry/test/ObjectTest.h
```diff
@@ -859,6 +859,17 @@ class ObjectTest : public CxxTest::TestSuite {
...
+    Object_sptr cube = createUnitCube();
+    TS_ASSERT_DELTA(cube->volume(), 1.0, 1e-6)
+  }
+
+  void testVolumeSmallCappedCylinder() {
+    Object_sptr cylinder = createSmallCappedCylinder();
+    const double correctVolume = M_PI * 0.005 * 0.005 * 0.003;
+    TS_ASSERT_DELTA(cylinder->volume(), correctVolume, 1e-6)
+  }
+
```
quantumsteve: Any tests for `Object::monteCarloVolume()`?

quantumsteve (changes requested):

Framework/Geometry/src/Objects/Object.cpp
soininen: I was thinking that a null bounding box refers to a zero sized one, but reading the documentation proves me wrong. It is indeed better to throw here.

Framework/Geometry/src/Objects/Object.cpp
soininen: Calculating an uncertainty would definitely be the best convergence criteria, but I wasn't sure how to implement it in a nice way. I'll have another look at it.

Framework/Geometry/src/Objects/Object.cpp
soininen: Could you elaborate a bit?

Framework/Geometry/src/Objects/Object.cpp
soininen: Good to know. I'll have a look what they have to offer there, then.

Framework/Geometry/test/ObjectTest.h
soininen: The test for the unit cube actually runs the Monte Carlo volume, but I'll add a more complex case just in case.

Framework/Geometry/src/Objects/Object.cpp
quantumsteve: In the worst case, the last thread could have significantly more iterations as any other thread. It would be nice to see these split up blocksize never differs by more than 1.

Rounding up `maxIterations / threadCount` would be the simplest option, but may add a few extra iterations.

Framework/Geometry/test/ObjectTest.h
quantumsteve: unit cube isn't using `GluGeometryHandler::GeometryType::CUBOID`?

Framework/Geometry/src/Objects/Object.cpp
soininen: I'm trying to make this run the same number of iterations regardless of the number of threads so the results don't depend on how many cores one has. I'll have a look if the workload could be balanced more evenly, but I don't think we should spend too much time on it as this is not an hour-long simulation.

Framework/Geometry/test/ObjectTest.h
soininen: The test unit cube is generated using six planes in `ObjectTest::createUnitCube()`, if I read the code correctly, and this prevents `Object::volume()` recognizing it as a cuboid.

Framework/Geometry/src/Objects/Object.cpp
```diff
@@ -1475,6 +1482,153 @@ double Object::ConeSolidAngle(const V3D &observer,
...
+      const double y = boundingBox.yMin() + rnd * boundingDy;
+      rnd = rnDistribution(rnEngine);
+      const double z = boundingBox.zMin() + rnd * boundingDz;
+      if (isValid(V3D(x, y, z))) {
+        ++hits;
+      }
+    }
+    // Collect results.
+    PARALLEL_ATOMIC
+    totalHits += hits;
```
quantumsteve: @mantidproject/tech-steering Do we have a preference between std::atomic and openmp pragmas?

quantumsteve (approved):

quantumsteve:
Changes look good, additional tests we added and they are passing on all supported platforms. :shipit:

Framework/Geometry/src/Objects/Object.cpp
martyngigg: We're leaning more to openmp in most places so I'm happy with this. At least they're easy to `ack` for...

PR #19172, opened 2017-03-17
Minimizer failure reports

### LINKED ISSUE #18203: Improve reporting of minimiser failure reports
### Expected behavior
If a fit minimizer failed to a reason we should make it easily visible in the UI (if it is a manually started fit in the UI) or throught the Fit algorithm status property.

We need to decide a mantid list of texts to use to report failure and success, and map those of the various minimisers to it.

### Actual behavior
At the moment we ignore the return statuses of minimisers.

mantid-roman:
Made various minimizers to report their failures. Display the error/success message in the fitting gui.

**To test:**

Run some fits from the GUI and scripts. The GUIs must display the fit status string and the number of completed iterations. In scripts: Fit returns error messages in its `OutputStatus` property and the number of completed iterations in `OutputNIterations`.

Fixes #18203.

*Release notes are in the UI section.*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


peterfpeterson:
The currently failing system tests are:
* `SNSPowderRedux.PG3StripPeaks` - Update the reference file. The differences are in some "tiny" peaks that may or may not be there.
* `SNSPowderRedux.PG3Analysis` - Update the reference file. The immediately differences are in the high-d region where the data is noisy. The regions where there are differences from `PG3StripPeaks` are also in this result and minor.
* `CalibrateRectangularDetector_Test.PG3Calibration` - Not sure what the right answer is. The calibration results move all the data to lower d-spacing. However, more pixels are calibrated and the peak widths aren't really changed. Also, it only affects future calibrations since most powder diffraction instruments will use an existing calibration when they re-process data I'm not sure if it will be seen by a user in the wild.

These all rely on `FindPeaks`. It might be beneficial to add more testing there so issues that arise in system. Also, merge `master` into the branch and resolve the conflicts before they build up.

mantid-roman:
@peterfpeterson Hi Pete. Would you mind updating those files for me please? I cannot generate them myself.

martyngigg:
@mantid-roman I'm bring my usual cautious self at this point. I'm concerned that this is a lot of change to core stuff this late on the release cycle, residually now the beta testing had started. Should we move this to `3.11`?

Framework/CurveFitting/src/Algorithms/Fit.cpp
```diff
@@ -212,8 +209,16 @@ void Fit::finalizeMinimizer(size_t nIterations) {
...
 
+  if (errorString.empty()) {
+    errorString = "success";
+  }
+
   // return the status flag
   setPropertyValue("OutputStatus", errorString);
+  setProperty("OutputNIterations", static_cast<int>(nIterations));
+  if (!this->isChild()) {
+    g_log.notice(errorString);
```
soininen: I wonder, if it would make sense to be a bit more wordy for the log, for example write
```
Fit status: success
```
In case of `errorString != "success"`, the error message could be written to `g_log.error()` instead of `g_log.notice()`.

Framework/CurveFitting/src/FuncMinimizers/DampedGaussNewtonMinimizer.cpp
```diff
@@ -30,6 +30,7 @@ DECLARE_FUNCMINIMIZER(DampedGaussNewtonMinimizer, Damped GaussNewton)
 DampedGaussNewtonMinimizer::DampedGaussNewtonMinimizer(double relTol)
     : IFuncMinimizer(), m_relTol(relTol) {
   declareProperty("Damping", 0.0, "The damping parameter.");
+  declareProperty("Debug", false, "Turn on the debug output.");
```
soininen: Does the explicit debugging property really make sense? We do have a `debug` channel in the logger already.

Framework/CurveFitting/src/FuncMinimizers/DampedGaussNewtonMinimizer.cpp
```diff
@@ -70,37 +70,53 @@ bool DampedGaussNewtonMinimizer::iterate(size_t) {
 
   for (size_t i = 0; i < n; ++i) {
     double tmp = H.get(i, i) + damping;
+    if (tmp == 0.0) {
+      m_errorString = "Function doesn't depend on parameter " +
+                      m_leastSquares->parameterName(i);
+      g_log.information(m_errorString);
```
soininen: Should this message be on a warning logging level?

Framework/CurveFitting/src/FuncMinimizers/LevenbergMarquardtMDMinimizer.cpp
```diff
@@ -70,8 +70,8 @@ bool LevenbergMarquardtMDMinimizer::iterate(size_t) {
   }
 
   if (m_mu > muMax) {
-    // m_errorString = "Failed to converge, maximum mu reached";
-    // g_log.warning() << m_errorString << '\n';
+    m_errorString = "Failed to converge, maximum mu reached.";
+    g_log.information(m_errorString);
```
soininen: Should failures like this be recorded at warning or error log level rather than just information?

Framework/CurveFitting/src/GSLMatrix.cpp
```diff
@@ -13,6 +14,23 @@ namespace CurveFitting {
...
+/// @param ilist :: Initialisation list as a list of rows:
+///      {{M00, M01, M02, ...},
+///       {M10, M11, M12, ...},
+///             ...
+///       {Mn0, Mn1, Mn2, ...}}
+GSLMatrix::GSLMatrix(std::initializer_list<std::initializer_list<double>> ilist)
+    : GSLMatrix(ilist.size(), ilist.begin()->size()) {
+  for (auto row = ilist.begin(); row != ilist.end(); ++row) {
+    auto i = static_cast<size_t>(std::distance(ilist.begin(), row));
+    for (auto cell = row->begin(); cell != row->end(); ++cell) {
```
soininen: Would it make sense to check that the number of cells on each row is the same?

Framework/CurveFitting/src/Algorithms/Fit.cpp
```diff
@@ -48,6 +48,9 @@ void Fit::initConcrete() {
       ->setDocumentation("Whether the fit was successful");
   declareProperty("OutputChi2overDoF", 0.0, "Returns the goodness of the fit",
                   Kernel::Direction::Output);
+  declareProperty("OutputNIterations", 0, Kernel::Direction::Output);
```
soininen: This breaks the Python interface and potentially user scripts. Can this be avoided somehow?

To me it looks like if **OutputNormalisedCovarianceMatrix** and **OutputParameters** properties were declare in `init()` and not in `createOutput()`, **OutputNIterations** could be added as the very last output property and keep the Python interface compatible. I might be overlooking some details here, though.

soininen (changes requested):
What concerns me most here is that this will break the Python API, as a new output property is introduced in the midst of existing outputs. I wonder if this could be avoided somehow, for example by declaring all output properties in `init()`. There were a few other minor issues mainly concerning logging, too.

Otherwise, I tested the changes in MantidPlot, and the Status and Iterations lines show up in the Fit dock as well as in the multidataset interface, as promised. The status-dependent coloring of the text is a nice touch IMHO.

Framework/CurveFitting/src/Algorithms/Fit.cpp
mantid-roman: This is done for performance reasons: if you don't need the table workspace with the parameters or the covariance matrix you may choose not to create them. You can access the optimised parameters via Function itself which is an `InOut` property. How does `OutputNIterations` break the python interface?

Framework/CurveFitting/src/Algorithms/Fit.cpp
mantid-roman: To be honest I am not sure which level to use here but I don't want it to be `error` for now. Some of the unsuccessful results may be still acceptable to the user and I wouldn't want to annoy them by saying it's an error especially if we couldn't offer a better solution that returns `success`.

Framework/CurveFitting/src/FuncMinimizers/DampedGaussNewtonMinimizer.cpp
mantid-roman: Maybe `Verbose` is a better name for this property.

Framework/CurveFitting/src/GSLMatrix.cpp
mantid-roman: You are right.

Framework/CurveFitting/src/FuncMinimizers/LevenbergMarquardtMDMinimizer.cpp
mantid-roman: Maybe the minimizers shouldn't log their error strings at all because `Fit` repeats them in `status`?

Framework/CurveFitting/src/Algorithms/Fit.cpp
soininen: You can see the breakage in the Python algorithms you had to modify to accommodate the changes in Fit: if a script expects the covariance matrix to be at index 3 in the tuple returned by Fit, the addition of OutputNIterations moves the index to 4. This breaks every script relaying on these indices.

I understand the trick done with the output workspaces, but I don't know if there is a nice way around this. How big of a deal is this really? Maybe someone at @mantidproject/gatekeepers has an opinion on this?

Framework/CurveFitting/src/Algorithms/Fit.cpp
soininen: Then I recommend using the warning level, but in the end, it is your call.

Framework/CurveFitting/src/FuncMinimizers/DampedGaussNewtonMinimizer.cpp
soininen: Shouldn't the verbosity be controlled by the logging level anyhow?

Framework/CurveFitting/src/Algorithms/Fit.cpp
mantid-roman: Python returns a `namedtuple` now, there are no indices any more.

There is even more interesting case of dynamically created output properties by `Fit`: `FABADAMinimizer` defines its own outputs (probability distributions, etc)

Framework/CurveFitting/src/Algorithms/Fit.cpp
mantid-roman: OK, let's try warning, see how it goes.

Framework/CurveFitting/src/Algorithms/Fit.cpp
martyngigg: The tuples still allow index access (or we would have broken scripts when doing that).

I think we should move this to the end of output list to keep things compatible.

Framework/CurveFitting/src/FuncMinimizers/DampedGaussNewtonMinimizer.cpp
mantid-roman: Debug logging level is for debugging the code. This flag is for the user to observe what `Fit` is doing internally and maybe to help to figure out why a fit went wrong. I don't think the user would want to be bombarded by all the irrelevant information from the debug level.

Framework/CurveFitting/src/Algorithms/Fit.cpp
mantid-roman: OK

Framework/CurveFitting/src/FuncMinimizers/DampedGaussNewtonMinimizer.cpp
soininen: What you describe sounds awfully lot like debugging, and there is the *Notices* logging level above *Debug* as well, if you get overwhelmed with the amount of messages. Anyway, I don't feel strongly about this. If you think the verbose option will help users to better grasp what happened during the fitting, then I guess the property is justified.

mantid-roman: In any case I think it is a question for a separate issue. Please feel free to open one. Here if you don't mind I'll rename the property and keep the rest as it is.

Framework/PythonInterface/mantid/simpleapi.py
```diff
@@ -302,6 +302,39 @@ def fitting_algorithm(f):
     When applied to a function definition this decorator replaces its code with code of
     function 'wrapper' defined below.
     """
+    def gather_fitting_returns(function_name, lhs, algm):
```
soininen: Why not declare the 'OutputNIterations' property in `Fit::createOutput()` after the other special output properties ('OutputNormalisedCovarianceMatrix', 'OutputParameters', etc.)? If I understood correctly, this would take care of the ordering automatically. Changing the ordering here looks to me, frankly speaking, like a hack.

mantid-roman: Output parameters should have never been ordered in the first place

soininen: Yes, it would be nice if a `dict()` or similar was returned instead of `tuple()`. However, things are as they are and I am a bit concerned about these Fit specific changes to simpleapi, and was just wondering if the same effect could be reached by changing `Fit.cpp` only.

mantid-roman: I have a few objections:
1. An error made in the python layer should be fixed in the python layer and not spread it to others.
2. Fit specific changes were made in the Fit (fitting) specific area of simpleapi.
3. `OutputNIterators` is a permanent property (not optional as `OutputNormalisedCovarianceMatrix` is) and should be declared by `init()`.

soininen: In my opinion, this is not an error of the Python layer, but rather in the design of Fit. There really should not be any 'temporary' output properties in any algorithm. The optional output workspaces should have been declared in `Fit::init()` as non-mandatory properties, not post-mortem in `Fit::createOutput()`.

Anyway, as you say, this change is limited to the Fit specific part of simpleapi, and I'm not feeling too strongly about it. Most importantly, the Python API breakage gets fixed, so I'm OK with this.

soininen (approved):

PR #19174, opened 2017-03-17
Reflectometry interface: Event mode handling now avoids loading workspaces already present in ADS.

### LINKED ISSUE #18536: Reflectometry GUI: avoid loading if workspace is already in ADS
This refers to event mode handling: if a workspace is already in the ADS we shouldn't need to load it again.

PranavBahuguna:
In the Reflectometry interface, the `GenericDataProcessorPresenter` has facilities for being able to access input runs from the AnalysisDataService rather than reloading them if possible. This PR allows the `ReflDataProcessorPresenter` class (which handles processing event workspaces) to access runs from the ADS in a similar manner.

**To test:**

This should be tested on `Release` mode. Alternatively if using `Debug` mode, go to `LoadEventNexus.cpp` and replace the following lines at lines 289-292 with this:

```
if (temp < 0) {
    if (temp < m_min_id)
        m_min_id = temp;
    if (temp > m_max_id)
        m_max_id = temp;
}
```

There are a variety of test cases that must pass for both event and non-event processing, To start:
- Interface can be accessed from Interfaces->Reflectometry->ISIS Reflectometry
- Enter runs `38415` and `38417` in the first two rows of the first group with transmission run `38393`.

For non-event testing, do not insert any values in the `Event Handling` tab. Then try the following:
- Process `38415` - this should load the run without failure.
- Load `38415` with `LoadISISNexus` as `TOF_38415` and process - this should successfully use run from ADS.
- Load `38415` with `LoadEventNexus` as `TOF_38415` and process - this should **not** use run from ADS and loads run instead.

For event testing, insert some values in the `Event Handling` tab (e.g. custom slicing: `0,10,20`). Then try these:
- Process `38415` - this should load the run without failure.
- Load `38415` with `LoadISISNexus` as `TOF_38415` and process - this should **not** use run from ADS and loads run instead.
- Load `38415` with `LoadEventNexus` as `TOF_38415` without loading monitors and process - this should **not** use run from ADS.
- Load `38415` with `LoadEventNexus` as `TOF_38415` with monitors and process - this should successfully use run from ADS.
- Repeat steps (apart from event workspaces) with `38417` - this is not an event workspace unlike `38415`. Ensure it uses the ADS when a loaded workspace for it exists.

Fixes #18536.

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


MantidQt/MantidWidgets/src/DataProcessorUI/GenericDataProcessorPresenter.cpp
```diff
@@ -520,53 +521,104 @@ std::string GenericDataProcessorPresenter::getPostprocessedWorkspaceName(
...
+
+  bool runFound;
+  std::string outName;
+  std::string fileName = instrument + run;
+
+  outName = findRunInADS(run, prefix, runFound);
+  if (!runFound ||
+      AnalysisDataService::Instance().retrieveWS<IEventWorkspace>(outName) !=
+          NULL) {
+    // Workspace must not be an event workspace
```
raquelalvarezbanos: I'd suggest that, if possible, we should keep all functionality related to event workspaces outside the `GenericDataProcessorPresenter`. The reason is that this class is only meant to deal with non-event runs, that is why we added event functionality to a separate class `ReflGenericDataProcessorPresenter`. I think we should try to avoid mixing event and non-event workspaces here. Is there any reason why this is not possible?

PranavBahuguna: I think one of the major problems I ran into while doing this was that when loading runs from the ADS, it was perfectly possible for `GenericDataProcessorPresenter` to load a run with the right name, but was an event workspace which is unusable by this class. A way this could happen was if we used set some slicing values (thus using `ReflDataProcessorPresenter` and loading runs as event workspaces), then remove slicing values (using `GenericDataProcessorPresenter`) which would attempt to use the event workspace loaded previously in the ADS. I believe this was a problem before I worked on this issue too.

raquelalvarezbanos: OK, I understand. What happens when the `GenericDataProcessorPresenter` attempts to run the reduction on an event workspace? Does it show an error message, does Mantid crash, ...?

PranavBahuguna: It shows an error in the results log when executing `ReflectometryReductionOneAuto` while using an event workspace.

raquelalvarezbanos: OK. I still think we should remove any reference to event workspaces from `GenericDataProcessorPresenter`. One way of doing that would be to add a check in `ReflGenericDataProcessorPresenter` to make sure that when we call `GenericDataProcessorPresenter` the workspaces that exist in the ADS are of the right type. Would this be possible?

PranavBahuguna: I think this can be done. I'll opt for warning users that the existing workspaces will be reloaded.

raquelalvarezbanos: OK 👍

gemmaguest:
In most of the tests, where the algorithm needs to re-load the workspace, I get a warning message telling me that it will overwrite the existing workspace, which is good. However, in this particular test there is no warning message: "Load `38415` with `LoadEventNexus` as `TOF_38415` without loading monitors and process - this should **not** use run from ADS."

I think this is overwriting an existing workspace, so should there be a warning message here too?

gemmaguest (approved):
The changes look good now and behave as described. Please :shipit:

PR #19186, opened 2017-03-20
Prepare IndexInfo for MPI support

### LINKED ISSUE #19151: Prepare IndexInfo for MPI support.

SimonHeybrock:
This PR is a follow up to https://github.com/mantidproject/mantid/pull/19120. Changes are mainly motivated by requirements for MPI. The most notable change is that `IndexInfo` does no longer store and track grouping via detector IDs, but only via `SpectrumDefinition`. This has a few consequences:

- Setting an `IndexInfo` on a workspace that does not have a matching instrument will fail.
- Setting an `IndexInfo` on a workspace will build the corresponding legacy way of storing groupings via detector IDs (accessed via `ISpectrum::getDetectorIDs`).
- For compliance with legacy code, setting invalid detector IDs in `ISpectrum` is still possible. Internally, when updating the corresponding `SpectrumDefinition`, bad IDs are silently ignored, just as before.

As a consequence of the above, getting `IndexInfo` from one workspace and setting it on another workspace will not copy any of the bad detector IDs from the original workspace, they are silently dropped.

`IndexInfo` (and code in `Indexing` in general) in the state provided by this PR is by no means complete. A fair number of features required for implementing algorithms with MPI support is still missing, but to avoid implementing things we do not need in the end (or need in a different way), implementation is postponed. Examples are:
- `IndexInfo::makeLocalSubvector`, or similar, to convert a vector of data item for all spectra into a vector of the corresponding data items for spectra on the local partition.
- `Indexing::extract()` and `Indexing::group()` do not support MPI and potentially even their interface should be changed. The main difficulty here is that we need to create a new `IndexInfo` with a subset of spectra, but their distribution should not change, i.e., we cannot create the new `IndexInfo` from scratch with a default partitioner, but instead need to keep the partition assignment of the source.

**To test:**

Code review.

Fixes #19151.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -2006,6 +2013,60 @@ void MatrixWorkspace::updateCachedDetectorGrouping(const size_t index) const {
...
+  size_t globalSpecIndex = 0;
+  for (size_t detIndex = 0; detIndex < detInfo.size(); ++detIndex) {
+    for (size_t time = 0; time < detInfo.scanCount(detIndex); ++time) {
+      if (m_indexInfo->isOnThisPartition(
+              Indexing::GlobalSpectrumIndex(globalSpecIndex++)))
+        specDefs[specIndex++].add(detIndex, time);
+    }
+  }
+  m_indexInfo->setSpectrumDefinitions(std::move(specDefs));
+}
```
OwenArnold: This looks reasonable.

OwenArnold:
Starting to look good.

Framework/Indexing/src/IndexInfo.cpp
```diff
@@ -13,99 +14,72 @@ namespace Indexing {
...
+  return m_spectrumNumberTranslator->spectrumNumber(index);
 }
 
 /// Set a spectrum number for each index.
 void IndexInfo::setSpectrumNumbers(
     std::vector<SpectrumNumber> &&spectrumNumbers) {
-  if (size() != spectrumNumbers.size())
+  if (m_spectrumNumberTranslator->globalSize() != spectrumNumbers.size())
     throw std::runtime_error(
         "IndexInfo: Size mismatch when setting new spectrum numbers");
```
LamarMoore: Is this error enough to explain what went wrong?

Framework/Indexing/src/IndexInfo.cpp
```diff
@@ -13,99 +14,72 @@ namespace Indexing {
...
 
 /// Set a contiguous range of spectrum numbers.
 void IndexInfo::setSpectrumNumbers(const SpectrumNumber min,
                                    const SpectrumNumber max) {
-  if (static_cast<int64_t>(size()) !=
-      static_cast<int32_t>(max) - static_cast<int32_t>(min) + 1)
+  auto newSize = static_cast<int32_t>(max) - static_cast<int32_t>(min) + 1;
+  if (static_cast<int64_t>(m_spectrumNumberTranslator->globalSize()) != newSize)
     throw std::runtime_error(
         "IndexInfo: Size mismatch when setting new spectrum numbers");
```
LamarMoore: Similarly here. I feel it would be good to indicate a mismatch to the global size.

Framework/Indexing/src/IndexInfo.cpp
```diff
@@ -178,25 +148,37 @@ SpectrumIndexSet IndexInfo::makeIndexSet(
...
+}
+
+void IndexInfo::makeSpectrumNumberTranslator(
+    std::vector<SpectrumNumber> &&spectrumNumbers) const {
+  PartitionIndex partition;
+  int numberOfPartitions;
+  if (m_storageMode == StorageMode::Distributed) {
+    partition = PartitionIndex(m_communicator.rank);
+    numberOfPartitions = m_communicator.size;
+  } else if (m_storageMode == StorageMode::Cloned) {
```
LamarMoore: I am confused here. Is it not possible to have a cloned storage mode with multiple partitions? I thought the data would be duplicated on each partition.

Framework/Indexing/src/SpectrumNumberTranslator.cpp
```diff
@@ -88,5 +109,21 @@ SpectrumIndexSet SpectrumNumberTranslator::makeIndexSet(
...
+  // can (currently) be smaller than the actual size due to duplicate spectrum
+  // numbers.
+  return m_globalToLocal.size() != m_globalSpectrumNumbers.size();
+}
+
+void SpectrumNumberTranslator::checkUniqueSpectrumNumbers() const {
+  // To support legacy code that creates workspaces with duplicate spectrum
+  // numbers we check for bad spectrum numbers only when needed, i.e., when
+  // accessing index maps.
+  if (m_globalSpectrumNumbers.size() != m_partitions.size())
```
LamarMoore: Could `m_partitions` do with a better name? Something like `m_spectrumToPartionMap`? I got confused reading this because I keep thinking `m_partitions.size()` tells you the number of partitions.

LamarMoore (commented):
Looks good to me. Just a few minor things which need clarification.

Framework/Indexing/src/IndexInfo.cpp
SimonHeybrock: You can have cloned storage mode with multiple MPI ranks. The workspace is cloned, there is only 1 partition, and every MPI rank has a copy of that partition.

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -145,7 +135,24 @@ void MatrixWorkspace::setIndexInfo(const Indexing::IndexInfo &indexInfo) {
...
+  // by rebuildDetectorIDGrouping).
+  setSpectrumDefinitions(m_indexInfo->spectrumDefinitions());
+}
+
+/// Variant of setIndexInfo, used by WorkspaceFactoryImpl.
+void MatrixWorkspace::setIndexInfoWithoutISpectrumUpdate(
+    const Indexing::IndexInfo &indexInfo) {
+  // Comparing the *local* size of the indexInfo.
+  if (indexInfo.size() != getNumberHistograms())
+    throw std::runtime_error("MatrixWorkspace::setIndexInfo: IndexInfo size "
```
OwenArnold: Minor point, but i would say that the exception you are raising here is predicable and underived consequence of input arguments. I would therefore suggest using `std::invalid_argument` in this case.

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -2006,6 +2013,60 @@ void MatrixWorkspace::updateCachedDetectorGrouping(const size_t index) const {
   setDetectorGrouping(index, getSpectrum(index).getDetectorIDs());
 }
 
+void MatrixWorkspace::buildDefaultSpectrumDefinitions() {
+  const auto &detInfo = detectorInfo();
+  size_t numberOfDetectors{detInfo.size()};
+  size_t numberOfSpectra{0};
+  if (detInfo.isScanning()) {
+    for (size_t i = 0; i < numberOfDetectors; ++i)
+      numberOfSpectra += detInfo.scanCount(i);
```
OwenArnold: I was initially thinking that this behaviour could like in `DetectorInfo` but I nowthink having it in the client code is actually correct as you have done.

Framework/Indexing/inc/MantidIndexing/IndexInfo.h
```diff
@@ -78,20 +68,32 @@ class SpectrumNumberTranslator;
 */
 class MANTID_INDEXING_DLL IndexInfo {
 public:
-  explicit IndexInfo(const size_t globalSize);
-  IndexInfo(std::vector<SpectrumNumber> &&spectrumNumbers,
-            std::vector<std::vector<DetectorID>> &&detectorIDs);
+  // StorageMode and Communicator are temporary helpers provided here for
+  // testing that will be removed (or rather replaced) once we have proper MPI
+  // support.
+  enum class StorageMode { Cloned, Distributed, MasterOnly };
```
OwenArnold: 👍

Framework/Indexing/test/IndexInfoTest.h
```diff
@@ -109,18 +109,82 @@ class IndexInfoTest : public CxxTest::TestSuite {
...
+  void test_setSpectrumDefinitions_cow_ptr() {
     IndexInfo info(3);
     const auto defs = Kernel::make_cow<std::vector<SpectrumDefinition>>(3);
     TS_ASSERT_THROWS_NOTHING(info.setSpectrumDefinitions(defs));
     TS_ASSERT_EQUALS(info.spectrumDefinitions().get(), defs.get());
   }
+
+  void test_StorageMode_Cloned() {
+    IndexInfo rank0(3, IndexInfo::StorageMode::Cloned,
+                    IndexInfo::Communicator{2, 0});
```
OwenArnold: Minor point. Might be nice to see some named arguments or inline comments for the arguments for the IndexInfo::Communicator.

Framework/Indexing/test/IndexInfoTest.h
```diff
@@ -109,18 +109,82 @@ class IndexInfoTest : public CxxTest::TestSuite {
...
+    // Current default is RoundRobinPartitioner
+    for (size_t i = 0; i < rank0.globalSize(); ++i) {
+      if (i % 2 == 0) {
+        TS_ASSERT(rank0.isOnThisPartition(i));
+      } else {
+        TS_ASSERT(rank1.isOnThisPartition(i));
+      }
+    }
+  }
 };
```
OwenArnold: Nice simple tests. Looks good.

LamarMoore:
If there are no more comments @OwenArnold  I would say :shipit:

OwenArnold:
Looks like you've addressed the minor concerns I had. :shipit:

PR #18936, opened 2017-02-20
Visualise off detector peaks from Predict Peaks

samueljackson92:
**Note: This must not be merged before #18921**

This adds the ability to view predicted peaks which do not land on any detector to the instrument view. The main changes are:

 - Update the IDF for WISH to include an extended detector space definition.
 - Add the ability to zoom out on an unwrapped view in the instrument view.
 - Add the ability to reset the zoom in an unwrapped view.

**To test:**

For testing with WISH you'll need a machine with lots of memory. Alternatively you can add a extended detector space component to an instrument of your choice.

 - Load some single crystal raw data for WISH and open the instrument view
 - Add some single crystal peaks
 - Use `FindUBUsingFFT` to get a UB matrix
 - Use `PredictPeaks` to predict the position of peaks and check the `PredictPeaksOffDetectors` option.
 - Drag the generated workspace onto the instrument view
 - Zoom in/out with left/right click & drag to view peaks.
 - Ctrl+Right Click resets the zoom.

Fixes #18923.

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [x] Is the code of an acceptable quality?
- [x] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)? Is it well structured with small focussed classes/methods/functions?
- [ ] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [x] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [x] Do changes function as described? Add comments below that describe the tests performed?
- [x] How do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant documentation been added/updated?
- [ ] Is user-facing documentation written in a user-friendly manner?
- [ ] Has developer documentation been updated if required?

- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

PR #19207, opened 2017-03-22
Speed up PredictPeaks for non-rectangular detectors

### LINKED ISSUE #19131: Reduce ray tracing to speed up PredictPeaks
The ray tracing performed when creating a peak in `setQLabFrame` is very slow for instruments with non-rectangular detectors. When looking up many, many peaks (such as in `PredictPeaks`) we can use the nearest neighbours to figure out this faster.

samueljackson92:
~~**Must not be merged before #18936**~~

This speeds up `PredictPeaks` for non-rectangular detectors by using a KDTree of detector positions rather than doing a full ray tracing search over all instrument components.

**To test:**

There are three cases that need testing:
 - Instruments with Rectangular detectors (e.g. TOPAZ)
 - Instruments with non-rectangular detectors (e.g. WISH)
 - Instruments with non-rectangular detectors and the tube-gap parameter (currently just CORELLI)

Here's a python script that will run PredictPeaks on all three cases. Note that the time taken to run `PredictPeaks` in each case should remain the same for TOPAZ and decrease for WISH and CORELLI with these changes applied.

**Note:** CORELLI took >1hr to run prior to these changes, so maybe get a cup of coffee while that one is running.

```python
ws = CreateSimulationWorkspace(Instrument='CORELLI',
                               BinParams='0,10000,20000',
                               UnitX='TOF')              
SetUB(ws, a=5.5, b=6.5, c=8.1, u='12,1,1', v='0,4,9')
cs = CrystalStructure('5.5 6.5 8.1', 'P b c a',
                      """Ni 0.232 0.114 0.543 1.0 0.00843;
                         Al 0.434 0.041 0.854 1.0 0.01120""")
ws.sample().setCrystalStructure(cs)
predicted = PredictPeaks(InputWorkspace=ws,
                         CalculateStructureFactors=False,
                         MinDSpacing=0.1,
                         MaxDSpacing=15,
                         PredictPeaksOutsideDetectors=False,
                         WavelengthMin=0.5, WavelengthMax=10.2)

ws2 = CreateSimulationWorkspace(Instrument='TOPAZ',
                               BinParams='0,10000,20000',
                               UnitX='TOF')
SetUB(ws2, a=5.5, b=6.5, c=8.1, u='12,1,1', v='0,4,9')
ws2.sample().setCrystalStructure(cs)                    
predicted2 = PredictPeaks(InputWorkspace=ws2,
                         CalculateStructureFactors=False,
                         MinDSpacing=0.1,
                         MaxDSpacing=15,
                         PredictPeaksOutsideDetectors=False,
                         WavelengthMin=0.5, WavelengthMax=10.2)
                         
ws2 = CreateSimulationWorkspace(Instrument='WISH',
                               BinParams='0,10000,20000',
                               UnitX='TOF')
SetUB(ws2, a=5.5, b=6.5, c=8.1, u='12,1,1', v='0,4,9')
ws2.sample().setCrystalStructure(cs)                    
predicted2 = PredictPeaks(InputWorkspace=ws2,
                         CalculateStructureFactors=False,
                         MinDSpacing=0.1,
                         MaxDSpacing=15,
                         PredictPeaksOutsideDetectors=False,
                         WavelengthMin=0.5, WavelengthMax=10.2)
```

Fixes #19131 .

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Kernel/inc/MantidKernel/NearestNeighbours.h
```diff
@@ -0,0 +1,196 @@
...
+ * takes care of the proper allocation and deallocation of memory.
+ */
+class NNDataPoints {
+public:
+  /** Construct a new set of data points
+   *
+   * @param nPts :: the number of data points
+   * @param nElems :: the number of elements for each point
+   */
+  NNDataPoints(const size_t nPts, const size_t nElems) : m_nPts(nPts) {
```
DavidFair: Could we use `int` in the signature as the `size_t` type implies we aren't casting down to int.

docs/source/release/v3.10.0/framework.rst
```diff
@@ -32,6 +32,7 @@ Improved
 - :ref:`FilterEvents <algm-FilterEvents-v1>` now splits all the sample logs if the input splitters are given by MatrixWorkspace or a general TableWorkspace.
 - Two new properties were added to :ref:`algm-Integration` *RangeLowerList* and *RangeUpperList* can be used to give histogram-specific integration ranges.
 - :ref:`algm-FindEPP` does not output the two extra workspaces from the :ref:`algm-Fit` anymore.
+- :ref:`PredictPeaks <algm-PredictPeaks-v1>` is now faster on instruments that do not have rectangular detectors
```
DavidFair: Can we say roughly how much faster as a percentage?

DavidFair (changes requested):

Framework/API/src/DetectorSearcher.cpp
```diff
@@ -0,0 +1,224 @@
...
+    auto E1 =
+        V3D(-std::sin(tt1) * std::cos(ph1), -std::sin(tt1) * std::sin(ph1),
+            1. - std::cos(tt1)); // end of trajectory
+    E1 = E1 * (1. / E1.norm());  // normalize
+
+    Eigen::Vector3d point(E1[0], E1[1], E1[2]);
+
+    // Ignore nonsensical points
+    if (point.hasNaN())
+      continue;
```
samueljackson92: This should only be `NaN` when the computed Q is non physical. For example is the Q value was directly along the beam direction.

Framework/API/src/DetectorSearcher.cpp
```diff
@@ -0,0 +1,224 @@
...
+  m_rayTracer->traceFromSample(direction);
+  const auto det = m_rayTracer->getDetectorResult();
+
+  if (!det)
+    return std::make_tuple(false, 0);
+
+  const auto detIndex = m_detInfo.indexOf(det->getID());
+
+  if (m_detInfo.isMasked(detIndex) || m_detInfo.isMonitor(detIndex))
+    return std::make_tuple(false, 0);
```
samueljackson92: This is a chicken & egg problem. We need to know which detector we hit before we know whether to discard it because the detector was masked/a monitor.

Framework/API/src/DetectorSearcher.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+  const auto result = checkInteceptWithNeighbours(detectorDir, neighbours);
+  const auto hitDetector = std::get<0>(result);
+  const auto index = std::get<1>(result);
+
+  if (hitDetector)
+    return std::make_tuple(true, m_indexMap[index]);
+
+  // Tube Gap Parameter specifically applies to CORELLI
+  if (!hitDetector && m_instrument->hasParameter("tube-gap")) {
```
samueljackson92: I have ideas for the future that this can be done by other means, but yes for now let's move it to another method.

docs/source/release/v3.10.0/framework.rst
samueljackson92: Depends how many detectors you have and whether you use the `tube-gap` parameter.

DavidFair: 🤔 maybe if we said something along the lines of 
"The performance increase will vary instrument to instrument, Instrument X has been shown to be  Y% faster"
So people still get a relative idea of the speedup esp. with the results you saw dropping runtimes down from hours is worth putting in.

Framework/API/test/DetectorSearcherTest.h
```diff
@@ -0,0 +1,243 @@
...
+        V3D(-std::sin(tt1) * std::cos(ph1), -std::sin(tt1) * std::sin(ph1),
+            1. - std::cos(tt1));  // end of trajectory
+    return E1 * (1. / E1.norm()); // normalize
+  }
+};
+
+class DetectorSearcherTestPerformance : public CxxTest::TestSuite {
+public:
+  static DetectorSearcherTestPerformance *createSuite() {
+    return new DetectorSearcherTestPerformance();
```
samueljackson92: I think this was just down to copy-paste :)

samueljackson92:
@DavidFair I believe I have addressed all of your issues now. Pending the builds passing can you please review again?

DavidFair (approved):

DavidFair:
The changes look good to me I'm happy to ship it but I'll wait on @VickieLynch and @MichaelWedel to give the go ahead.

DavidFair:
I'm going to mark this ready to ship so it can be merged.
:shipit:

VickieLynch:
Much faster for CORELLI data and the same for TOPAZ.  I am ready for merging.

MichaelWedel (changes requested):
A few small things that would be good to clarify/discuss.

In addition I think it would be a very good idea if @SimonHeybrock took a brief look at `DetectorSearcher` and how that fits conceptually with the rest of the Instrument/Detector info infrastructure.

Framework/API/src/DetectorSearcher.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+    // calculate a Q vector for each detector
+    // This follows a method similar to that used in IntegrateEllipsoids
+    const auto &det = m_detInfo.detector(pointNo);
+    const auto tt1 = det.getTwoTheta(V3D(0, 0, 0), V3D(0, 0, 1)); // two theta
+    const auto ph1 = det.getPhi();                                // phi
+    auto E1 =
+        V3D(-std::sin(tt1) * std::cos(ph1), -std::sin(tt1) * std::sin(ph1),
+            1. - std::cos(tt1)); // end of trajectory
+    E1 = E1 * (1. / E1.norm());  // normalize
```
SimonHeybrock: What is this? A unit vector in some direction, but I don't really see which one?

Framework/API/src/DetectorSearcher.cpp
```diff
@@ -0,0 +1,251 @@
...
+    const V3D &detectorDir,
+    const Kernel::NearestNeighbours<3>::NearestNeighbourResults &neighbours) {
+  std::vector<double> gaps = m_instrument->getNumberParameter("tube-gap", true);
+  if (!gaps.empty()) {
+    const auto gap = static_cast<double>(gaps.front());
+    // try adding and subtracting tube-gap in 3 q dimensions to see if you can
+    // find detectors on each side of tube gap
+    for (int i = 0; i < 3; i++) {
+      auto gapDir = V3D(0., 0., 0.);
+      gapDir[i] = gap;
```
samueljackson92: This code was effectively copied from what `Peak` was already doing during the ray tracing step, except now we only look at the nearest neighbours instead of the whole instrument tree. As far as I understand it, this was a quick hack by @VickieLynch to catch missing reflections falling between tubes.

I'd argue that it is *not* correct for a couple of reasons. 
 1. You're correct in that I think the units are not 100% accurate, it's just the difference between the centre of two (random) pixels minus their width. It does not account for individual values or proper  "curved" interpolation between tubes.
2. This will also potentially catch vectors that fall outside of detectors as well as between tubes.

However, I'd argue that this is still necessary because users do notice when reflections they can clearly see appear to be missing. I think the longer term solution is to come up with and agree on what the proper fix for this case should be. I am very open to suggestions.

Framework/Crystal/src/PredictPeaks.cpp
```diff
@@ -461,37 +470,86 @@ void PredictPeaks::calculateQAndAddToOutput(const V3D &hkl,
...
+      throw std::runtime_error("PredictPeaks: user requested use of a extended "
+                               "detector space to predict peaks but there is no"
+                               "definition in the IDF");
+
+    // find where this Q vector should intersect with "extended" space
+    Geometry::Track track(detInfo.samplePosition(), detectorDir);
+    if (!c->interceptSurface(track))
+      return;
+
+    const auto magnitude = track.back().exitPoint.norm();
```
samueljackson92: As I understand it when you trace a ray to the surface of some object, if the object was hit you will get a vector from the starting point to the exit point (where it hit). I believe it is possible to "chain" rays more than once, but in "interceptSurface" it only adds one vector hence I use `back()` here.

Perhaps I should add a comment in the code to clarify as well.

SimonHeybrock (commented):
Small comment on instrument access:

Framework/Kernel/inc/MantidKernel/NearestNeighbours.h
```diff
@@ -0,0 +1,197 @@
...
+              const std::unique_ptr<ANNdist[]> nnDistList) {
+    NearestNeighbourResults results;
+    results.reserve(k);
+
+    for (size_t i = 0; i < k; ++i) {
+      // create Eigen array from ANNpoint
+      auto pos = m_dataPoints->mutablePoint(nnIndexList[i]);
+      VectorType point = Eigen::Map<VectorType>(pos, N, 1);
+      auto item = std::make_tuple(point, nnIndexList[i], nnDistList[i]);
+      results.push_back(item);
```
samueljackson92: Maybe. I might experiment.

TEST-APITest.DetectorSearcherTest.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
samueljackson92: No! I shall remove. :)

Framework/API/src/DetectorSearcher.cpp
samueljackson92: A unit vector in Q but corresponding to the direction of a detector. This is mostly taken from code in`IntegrateEllipsoids`.

SimonHeybrock: Ok, maybe this can be explained in the code a bit? It looks *almost* like the negative detector position (with normalized `R`), which is quite confusing otherwise.

samueljackson92: Perhaps some more detailed commenting required? Or if you have suggestions for how I can more clearly write this so it's self descriptive that would be good too.

samueljackson92:
I think this is ready for another review. @MichaelWedel's comment above should be addressed in another issue as that problem existed before this PR.

VickieLynch (approved):
This works with SNS data

martyngigg:
@MichaelWedel As Sam said, is it acceptable to address your issues separately?

MichaelWedel (approved):

PR #19210, opened 2017-03-22
Sanitize Histogram1D::setHistogram() input

### LINKED ISSUE #19204: setHistogram() can inject invalid data into a workspace causing segfault
It is possible to have invalid Y data in a workspace if one provides `setHistogram()` only the bin edges. So, having something like
```c++
Mantid::API::MatrixWorkspace_sptr WS =  Mantid::API::WorkspaceFactory::Instance().create("Workspace2D", 1, 2, 1);
HistogramData::BinEdges binEdges{0.0, 1.0};
WS->setHistogram(0, binEdges);
const auto Y = WS->y(0);
```
in a program will segfault Mantid.
This seems to be due to the `Histogram` constructor
```c++
template <class TX, class TY = Counts, class TE = CountVariances>
explicit Histogram(const TX &x, const TY &y = Counts(), const TE &e = CountVariances());
```
Apparently, the default values for the `y` and `e` parameters will create invalid objects which, when accessed later, will segfault.

There is a test algorithm `SetHistBug` available in a forked GitHub repository [here](https://github.com/mganeva/mantid). Look for a branch 'set_hist_bug'. The algorithm takes an event workspace as *InputWorkspace*. This can be created by the `CreateSampleWorkspace` algorithm.

soininen:
This PR adds input data checking to `Histogram1D::setHistogram()`. Basically, NULL Y data will cause an exception. This prevents `Workspace2D` objects to have invalid data which might segfault Mantid.

**To test:**

Code review.

Fixes #19204.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/src/ISpectrum.cpp
+++ b/Framework/API/src/ISpectrum.cpp
@@ -146,7 +146,6 @@ void ISpectrum::checkAndSanitizeHistogram(HistogramData::Histogram &histogram) {
         "Histogram1D: invalid input: Y data set to nullptr");
   }
   if (!histogram.sharedE()) {
-    std::cout << "Failed histogram sanity check." << std::endl;
     throw std::invalid_argument(
         "Histogram1D: invalid input: E data set to nullptr");
   }
--- a/Framework/API/src/ISpectrum.cpp
+++ b/Framework/API/src/ISpectrum.cpp
@@ -135,22 +135,6 @@ bool ISpectrum::hasDx() const { return bool(histogramRef().sharedDx()); }
  */
 void ISpectrum::resetHasDx() { mutableHistogramRef().setSharedDx(nullptr); }
 
-/**
- * Makes sure the histogram has valid Y data.
- * @param histogram A histogram to process.
- * @throw std::invalid_argument if Y data is NULL.
- */
-void ISpectrum::checkAndSanitizeHistogram(HistogramData::Histogram &histogram) {
-  if (!histogram.sharedY()) {
-    throw std::invalid_argument(
-        "Histogram1D: invalid input: Y data set to nullptr");
-  }
-  if (!histogram.sharedE()) {
-    throw std::invalid_argument(
-        "Histogram1D: invalid input: E data set to nullptr");
-  }
-}
-
 /// Copy constructor.
 ISpectrum::ISpectrum(const ISpectrum &other)
     : m_specNo(other.m_specNo), detectorIDs(other.detectorIDs) {
--- a/Framework/API/src/ISpectrum.cpp
+++ b/Framework/API/src/ISpectrum.cpp
@@ -135,6 +135,23 @@ bool ISpectrum::hasDx() const { return bool(histogramRef().sharedDx()); }
  */
 void ISpectrum::resetHasDx() { mutableHistogramRef().setSharedDx(nullptr); }
 
+/**
+ * Makes sure the histogram has valid Y data.
+ * @param histogram A histogram to process.
+ * @throw std::invalid_argument if Y data is NULL.
+ */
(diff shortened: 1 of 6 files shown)


Framework/DataObjects/src/Histogram1D.cpp
```diff
@@ -65,5 +65,18 @@ const MantidVec &Histogram1D::dataDx() const { return m_histogram.dataDx(); }
...
 const MantidVec &Histogram1D::readDx() const { return m_histogram.readDx(); }
 
+/**
+ * Makes sure the histogram has valid Y data.
+ * @param histogram A histogram to process.
+ * @throw std::invalid_argument if Y data is NULL.
+ */
+void Histogram1D::checkAndSanitizeHistogram(
+    HistogramData::Histogram &histogram) {
+  if (!histogram.sharedY()) {
```
DavidFair: Would this test be better in `ISpectrum` so all histograms get it by default (unless they override)?
Currently we do no checking here:
https://github.com/mantidproject/mantid/blob/master/Framework/API/inc/MantidAPI/ISpectrum.h#L246

Additionally what about E data being null?

DavidFair (changes requested):

Framework/DataObjects/src/Histogram1D.cpp
soininen: There was a reason the checking ended up in `Histogram1D` but I cannot remember it anymore. Anyway, I made that decision for an implementation that didn't throw but instead silently filled Y and E with zeros. Now, I've moved the throwing version to `ISpectrum` and don't see problems with it. The only downside is that unit tests need to be done in `Histogram1D` as ISpectrum is too virtual for anything.

I also added check for null E data. I thought it's not possible to have null E if you give Y to `Histogram`'s constructor, but there are other ways to construct invalid `Histogram`s.

Framework/API/src/ISpectrum.cpp
```diff
@@ -135,6 +135,22 @@ bool ISpectrum::hasDx() const { return bool(histogramRef().sharedDx()); }
  */
 void ISpectrum::resetHasDx() { mutableHistogramRef().setSharedDx(nullptr); }
 
+/**
+ * Makes sure the histogram has valid Y data.
+ * @param histogram A histogram to process.
+ * @throw std::invalid_argument if Y data is NULL.
+ */
+void ISpectrum::checkAndSanitizeHistogram(HistogramData::Histogram &histogram) {
```
SimonHeybrock: Semantically, it would probably make more sense to have this implementation in `Histogram1D`, since `EventList` (which also inherits `ISpectrum`) has a different implementation. Having this in the common base class is a bit confusing.

DavidFair: Originally I suggested it because the interface in the base class `ISpectrum` directly constructs the Histogram [here](https://github.com/mantidproject/mantid/blob/master/Framework/API/inc/MantidAPI/ISpectrum.h#L104) but relies on the derived class to provide the override for any checks to happen. 
Imagining we had future workspace types which inherited `ISpectrum` would not having Y and/or E data be an exception to the rule or something that is worth checking on a case by case basis?

SimonHeybrock: I think having further types is very unlikely.

DavidFair: Ok in that case I agree we should move this back to `Histogram1D` - sorry @soininen

soininen: No problem. I'm also of the opinion that this fits Histogram1D better.

DavidFair (approved):

DavidFair:
LGTM :shipit:

PR #19213, opened 2017-03-22
Fix calling Load with None returns bad error

### LINKED ISSUE #19190: Calling load with None argument gives unhelpful error
When calling Load from the Python interface with a `None` argument for the Filename:
```python
Load(Filename=None)
```
Gives the following error - we should explicitly tell the user they are passing a `None` value instead.
```
ValueError: Problem when setting Filename. This is the detailed error description: When converting 
parameter "Filename": TypeRegistry::retrieve(): No PropertyValueHandler registered for type 'class 
std::vector<class std::vector<class std::basic_string<char,struct std::char_traits<char>,class 
std::allocator<char> >,class std::allocator<class std::basic_string<char,struct std::char_traits<char>,class 
std::allocator<char> > > >,class std::allocator<class std::vector<class std::basic_string<char,struct 
std::char_traits<char>,class std::allocator<char> >,class std::allocator<class std::basic_string<char,struct 
std::char_traits<char>,class std::allocator<char> > > > > >'
```

DavidFair:
Description of work.
This PR fixes an issue with `Load` being called in Python with `Filename=None` passed. Previously it would print a long error message describing the Python container. Now it prints a valid error saying that a `None` is not acceptable.

Additionally I tided `simpleapi` to fix the PEP8 white space issues (which should reduce the number of warnings in flake8) ~~and remove `import os` from the `simpleapi` which was unused into the various scripts which relied on this implicit import.~~ this was too much work so in the next maintenance cycle I'll open another PR to finish this properly. 

**To test:**
Ensure all builds pass
Run the following and check it gives a sane error message
`Load(Filename=None)`

Fixes #19190 .

*Does not need to be in the release notes. - Only noticed internally*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/PythonInterface/mantid/simpleapi.py
+++ b/Framework/PythonInterface/mantid/simpleapi.py
@@ -37,7 +37,7 @@ from . import apiVersion, __gui__
 from .kernel._aliases import *
 from .api._aliases import *
 
-#------------------------ Specialized function calls --------------------------
+# ------------------------ Specialized function calls --------------------------
 # List of specialized algorithms
 __SPECIALIZED_FUNCTIONS__ = ["Load", "StartLiveData", "CutMD", "RenameWorkspace"]
 # List of specialized algorithms
@@ -101,7 +101,8 @@ def _create_generic_signature(algm_object):
     arg_str = ','.join(arg_list)
     # Calling help(...) will put a * in front of the first parameter name,
     # so we use \b to delete it
-    return ("\b%s" % arg_str, "\b\bVersion=%d" % algm_object.version())
+    return "\b%s" % arg_str, "\b\bVersion=%d" % algm_object.version()
+
 
 def Load(*args, **kwargs):
     """
@@ -139,6 +140,12 @@ def Load(*args, **kwargs):
       Load('INSTR00001000.nxs',OutputWorkspace='run_ws')
     """
     filename, = _get_mandatory_args('Load', ["Filename"], *args, **kwargs)
+    if not filename:
+        # If we try to set property with a None type we get a unhelpful error about allocators
+        # so check up front here
+        raise ValueError("Problem with supplied Filename. The value given was a 'None' "
+                         "type and cannot be used. Please ensure the Filename is set"
+                         " to the path of the file.")
 
     # Create and execute
     (_startProgress, _endProgress, kwargs) = extract_progress_kwargs(kwargs)
--- a/Framework/PythonInterface/mantid/simpleapi.py
+++ b/Framework/PythonInterface/mantid/simpleapi.py
@@ -863,31 +863,25 @@ def _gather_returns(func_name, lhs, algm_obj, ignore_regex=None):
                 raise RuntimeError('Internal error. Unknown property type encountered. "%s" '
                                    'on algorithm "%s" is not understood by '
                                    'Python. Please contact development team' % (name, algm_obj.name()))
-    # For reference this ensures multiple returns works eg
-    # foo, bar = myFunc(...)
-    # In the case foo and bar mean we have 2 on LHS and myFunc should return 2 member vars
+    nvals = len(retvals)
+    nlhs = lhs[0]
 
-    number_of_returned_values = len(retvals)
-    number_of_vars_lhs_of_func = lhs[0]
+    if nlhs == 0:
+        return None
 
-    if number_of_returned_values != number_of_vars_lhs_of_func:
-        if number_of_vars_lhs_of_func == 0:
+    if nvals != nlhs:
+        if nlhs == 1:
             pass  # This is fine as we do not need to unpack
         else:
             # There is a discrepancy in the number whilst unpacking variables
             # Let's not have the more cryptic unpacking error raised
             raise RuntimeError("%s is trying to return %d output(s) but you have provided %d variable(s). "
-                               "These numbers must match." % (func_name, number_of_returned_values, number_of_vars_lhs_of_func))
-
-    if number_of_returned_values == 0:
-        # Nothing was returned so return None type
-        return None
+                               "These numbers must match." % (func_name, nvals, nlhs))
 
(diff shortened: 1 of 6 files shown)


gvardany:
Loading none throws the expected error. But should you not remove in this PR the `import os` from the 5 algorithms touched then?

DavidFair:
I'm in two minds about this - technically I should do them all in a separate PR which I will do in the future and normally I would completely agree with you. 
However equally this makes the scripts correct without affecting them. Reverting the changes - then putting them back again in a later PR seems wasteful to me.

gvardany:
OK :shipit: then

Framework/PythonInterface/mantid/simpleapi.py
```diff
@@ -806,68 +850,72 @@ def ignore_property(name, ignore_regex):
...
-
+                raise RuntimeError('Internal error. Unknown property type encountered. "%s" '
+                                   'on algorithm "%s" is not understood by '
+                                   'Python. Please contact development team' % (name, algm_obj.name()))
     nvals = len(retvals)
     nlhs = lhs[0]
-    if nlhs > 1 and nvals != nlhs:
+
+    # If we have more than one value but not the same number of values throw
+    if 1 < nlhs != nvals:
```
martyngigg: Is this correct? It doesn't read very well to me as it takes too long to think and remember the operator precedence rules to see if it is correct? Was this a suggestion from the flake8 warning or somewhere else?

DavidFair: Chaining expressions is something Python encourages you to do so my IDE warns me on this line. I'm not sure whether it is a flake warning but it tends to be flake warnings are a subset of the warnings I get locally. 

How about tidying it up to both remove the warning and make the intention a bit clearer:
Original:
```python
    if nlhs > 1 and nvals != nlhs:
        # There is a discrepancy in the number are unpacking variables
        # Let's not have the more cryptic unpacking error raised
        raise RuntimeError("%s is trying to return %d output(s) but you have provided %d variable(s). "
                           "These numbers must match." % (func_name, nvals, nlhs))
    if nvals > 0:
        ret_type=namedtuple(func_name+"_returns",retvals.keys())
        ret_value=ret_type(**retvals)
        if nvals==1:
            return ret_value[0]
        else:
            return ret_value
    else:
        return None
```
Into something like:
```python
if nlhs == 0:
    return None

if nvals != nlhs:
   if nlhs == 1:
        pass   # This is fine as we do not need to unpack
   else:
               # There is a discrepancy in the number are unpacking variables
        # Let's not have the more cryptic unpacking error raised
        raise RuntimeError("%s is trying to return %d output(s) but you have provided %d variable(s). "
                           "These numbers must match." % (func_name, nvals, nlhs))

# Unpack and return
ret_type=namedtuple(func_name+"_returns",retvals.keys())
ret_value=ret_type(**retvals)
return ret_value[0] if nvals == 1 else ret_value

Comment (martyngigg) [inline Framework/PythonInterface/mantid/simpleapi.py]: I'll have to live with disagreeing with Python over `1 < nlhs != nvals` then I think but to be honest it just looks weird to me.

:+1: For your new version. This reads better and makes the intention of the code much clearer.

**Comment by peterfpeterson** [comment]
I changed the ship to a squirrel so it doesn't accidentally match a filter for things that are ready to be merged.

**Comment by DavidFair** [comment]
It turns out there is a subtle bug in the original implementation of unpacking returned values from a function simple API executes. 
If there is a function called `myFunc` which returns 3 values in Python we should return a tuple of three values in the following case
`foo=myFunc(...)`
Currently we return just the first result.

I'm going to revert commits  `39a24b1` and `f7cab6b` and cherry pick them into their own PR as they will require some additional work.

File: Framework/PythonInterface/mantid/simpleapi.py (hunk 1 above)
Comment (martyngigg) [inline Framework/PythonInterface/mantid/simpleapi.py]: Can we put this back to what it was originally: `if nlhs > 1 and nvals != nlhs:` ?

File: Framework/PythonInterface/mantid/simpleapi.py [hunk 2]
```diff
@@ -806,68 +850,72 @@ def ignore_property(name, ignore_regex):
...
         # Let's not have the more cryptic unpacking error raised
         raise RuntimeError("%s is trying to return %d output(s) but you have provided %d variable(s). "
                            "These numbers must match." % (func_name, nvals, nlhs))
     if nvals > 0:
-        ret_type=namedtuple(func_name+"_returns",retvals.keys())
-        ret_value=ret_type(**retvals)
-        if nvals==1:
+        ret_type = namedtuple(func_name+"_returns", retvals.keys())
+        ret_value = ret_type(**retvals)
+        if nvals == 1:
```
martyngigg: Also, I'm not sure I see the where the bug is? 

This checks if `nvals==1` which is if there is `1` value provided not if there is 1 value on the left hand side. I don't think the usage examples for [`GetEi`](http://docs.mantidproject.org/nightly/algorithms/GetEi-v2.html#usage) would work if this was broken.

Framework/PythonInterface/mantid/simpleapi.py
DavidFair: 👍

Framework/PythonInterface/mantid/simpleapi.py
DavidFair: Looking at master it seems my indentation went wrong which was throwing me off.

I definitely think this needs a clean up considering the confusion that has arisen from it but I agree its for a separate PR so I'll revert it back to the original form

DavidFair:
Just a gentle nudge could we look at this PR again?

martyngigg:
I'm happy with the this now. @gvardany Any other comments?

martyngigg (approved):

gvardany:
Not really, please go ahead.

PR #19218, opened 2017-03-22
Use clamp algorithm instead of custom code.

quantumsteve:
Description of work.

This uses the [clamp algorithm](http://www.boost.org/doc/libs/1_63_0/libs/algorithm/doc/html/algorithm/Misc.html#the_boost_algorithm_library.Misc.clamp) instead of a more verbose comparison. The behavior for values < 0 is slightly different, but still consistent with the documentation.

**To test:**

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


rosswhitfield:
It states just above the function [here](https://github.com/mantidproject/mantid/blob/397cc4fc2119902444c81a84038bac373eb44466/Vates/VatesAPI/src/vtkSplatterPlotFactory.cpp#L586-L589) that it must be more than 0, in the interval (0,100]. The `clamp` algorithm will return `0` for `percentToUse<=0`. Are you sure this is the behaviour wanted?

quantumsteve:
@rosswhitfield I updated the minimum value.

rosswhitfield:
LGTM :shipit:

Issue #19208, opened 2017-03-22
Waterfall Plot may not display legend


KarlPalmen:
### Expected behavior

Waterfall plot always shows its legend

### Actual behaviour

Waterfall plot can occur without legend being shown or more precisely it is shown then quickly removed.
Sometimes the legend is shown far up, so only its bottom is visible.

### Steps to reproduce the behaviour

Run script
```
base = CreateSampleWorkspace(Function='Powder Diffraction', NumBanks=4, BankPixelWidth=1)

wsNames = []

for t in range (10,110,10):
    wsName = "temp" + str(t)
    wsNames.append(wsName)
    MultiplyRange(base,75,99,t/100.0,OutputWorkspace=wsName)
    AddSampleLog(wsName,"Temp",str(t),"Number","Kelvin")

TestGroup = GroupWorkspaces(wsNames)
```
Then do waterfall plot of one spectrum of the TestGroup workspace.

### Platforms affected

Windows 7 at least.

KarlPalmen:
I found that the legend was moved up almost out of view and could be moved back into view.

KarlPalmen:
Hopefully will be fixed by change to Mantid 4.0

PR #19237, opened 2017-03-23
Harmonise plotting of workspaces from right-button click

### LINKED ISSUE #18159: Harmonize advanced plotting options
### Current behavior

We have a few plotting options that are "advanced"
 1. waterfall plots - option on plot spectrum dialog
 2. tiled plots - option on plot spectrum dialog
 3. Plot Surface from group - option on workspace group context menu if 3 or more members 
 4. Plot Contour from group - option on workspace group context menu if 3 or more members
image

As you can see the approaches are a bit varied as to how you get to them (and the last two can be hard to find),but there is a lot of similarity in the dialog box they create.
image

image

### Actual behavior
We want to harmonize all of this without disrupting the well known and used plot Spectrum (+- errors) usage.

We want to replace the context menu options with 
 1. Plot Spectrum ...
 1. Plot Spectrum with errors ...
 1. Plot Special ...
 1. Color Fill plot (leave as before)

The two plot spectrum options had special behaviour if the workspace contained only one spectrum in that it would not show the dialog box and just plot immediately.  We want to retain that.

We want to harmonize the dialog box that the other options use, and plot special should be able to be used for waterfall and tiled plots as well as the surfaces and contours, and there is no reason it cannot also be used for normal plots.  Here is a suggested mockup.

mockup

The same dialog could be used for Plot Spectrum, just with the group box hidden and the form shrunk appropriately.

You might want to have a second level of context menus under plot special, they should just fill the plot type out in the dialog box for you.

All of the underlying plotting for this should already be in place. The only new aspect is allowing waterfall plots to be labelled using a log value or custom labels.  The default label approach or [workspace]-sp[wpectrum no] should be maintained as an option.

### Platforms affected
All

KarlPalmen:
Until this, surface and contour plotting was done in quite a different way to the plotting of spectra with or without errors in simple 1D, waterfall or tiled plot. Now they should all be accessible from a `Plot Spectrum Advanced` right-click menu option and basic plot spectrum functionality should be available as before.

The issue #18159 shows this in more detail.

**To test:**

The issue shows a python script and file to load that would be useful in testing also use your own file or scripts as you see fit.

Test the right click menu for `Plot Spectrum`, `Plot Spectrum with Errors` and `Plot Spectrum Advanced` in the various options available. It has been noted that the legend may not always appear for a waterfall plot #19208 and when 1D or waterfall plots are done with log values instead of workspace index, the items may not be shown in numerical order in the legend.

Also check that surface and contour plots are done in the same manner as before this change.

Fixes #18159.

Release notes to do.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


MantidPlot/src/Mantid/MantidGroupPlotGenerator.cpp
```diff
@@ -25,70 +26,97 @@ MantidGroupPlotGenerator::MantidGroupPlotGenerator(MantidDisplayBase *mantidUI)
...
+ * @param logName :: [input] log name
+ * @param customLogValues :: [input] custom log values
+ * @param workspaces :: [input] set of workspaces forming the group to be
+ * plotted
  */
 void MantidGroupPlotGenerator::plotSurface(
-    const WorkspaceGroup_const_sptr &wsGroup,
-    const MantidSurfacePlotDialog::UserInputSurface &options) const {
-  plot(Type::Surface, wsGroup, options);
+    bool accepted, int plotIndex, const QString &axisName,
```
AntonPiccardoSelg: The class is not only used for groups any longer, so we should consider changing the name of the class.

MantidPlot/src/Mantid/MantidGroupPlotGenerator.cpp
```diff
@@ -101,31 +129,35 @@ void MantidGroupPlotGenerator::plot(
...
-  const auto index = static_cast<size_t>(
-      options.plotIndex);                // which spectrum to plot from each WS
-  const auto &logName = options.logName; // Log to read for axis of XYZ plot
+    Type graphType,
+    const std::vector<Mantid::API::MatrixWorkspace_const_sptr> &workspaces,
+    int plotIndex, const QString &logName,
+    const std::set<double> &customLogValues) const {
+  const auto index =
+      static_cast<size_t>(plotIndex); // which spectrum to plot from each WS
+  // const auto &logName = logName; // Log to read for axis of XYZ plot
```
AntonPiccardoSelg: If the commented out code is not being used, then it should be removed.

MantidPlot/src/Mantid/MantidGroupPlotGenerator.cpp
```diff
@@ -274,21 +261,24 @@ std::string MantidGroupPlotGenerator::validatePlotOptions(
...
+ * @param workspaces :: [input] list of workpaces containing data for graph.
+ * The title is generated from the first of these workspaces.
  * @returns :: Title for X axis of graph
  */
 QString MantidGroupPlotGenerator::getXAxisTitle(
-    const boost::shared_ptr<const Mantid::API::WorkspaceGroup> wsGroup) const {
-  if (wsGroup->getNumberOfEntries() <= 0) {
+    const std::vector<Mantid::API::MatrixWorkspace_const_sptr> &workspaces)
+    const {
+  if (workspaces.size() <= 0) {
```
AntonPiccardoSelg: `workspaces.empty()` is a better option here

MantidPlot/src/Mantid/MantidPlotUtilities.cpp
```diff
@@ -0,0 +1,71 @@
...
+ * @throws invalid_argument if log is wrong type or not present
+ */
+double getSingleWorkspaceLogValue(
+    size_t wsIndex, const Mantid::API::MatrixWorkspace_const_sptr &matrixWS,
+    const QString &logName) {
+  if (logName == MantidWSIndexWidget::WORKSPACE_INDEX || logName == "") {
+    return wsIndex;
+  } else {
+    // MatrixWorkspace is an ExperimentInfo
+    if (auto ei = boost::dynamic_pointer_cast<const ExperimentInfo>(matrixWS)) {
```
AntonPiccardoSelg: `MatrixWorkspace` inherits from `ExperimentInfo`, hence you don't have to perform a `dynamic_pointer_cast` here. You should be able to get the run via `matrixWS->run()`.

MantidPlot/src/Mantid/MantidPlotUtilities.cpp
```diff
@@ -0,0 +1,71 @@
...
+      auto log = ei->run().getLogData(logName.toStdString());
+      if (log) {
+        if (dynamic_cast<Mantid::Kernel::PropertyWithValue<int> *>(log) ||
+            dynamic_cast<Mantid::Kernel::PropertyWithValue<double> *>(log)) {
+          return std::stod(log->value());
```
AntonPiccardoSelg: Good use of  `C++11` feature

MantidPlot/src/Mantid/MantidPlotUtilities.cpp
```diff
@@ -0,0 +1,71 @@
...
+* i.e. the nth in order from smallest to largest.
+* If the index is outside the range, returns 0.
+* @param wsIndex :: [input] Index of log value to use
+* @param logValues :: [input] User-provided set of log values
+* @returns Numeric log value
+*/
+double getSingleWorkspaceLogValue(size_t wsIndex,
+                                  const std::set<double> &logValues) {
+  double value = 0;
+  if (wsIndex < static_cast<int>(logValues.size())) {
```
AntonPiccardoSelg: Why do you `static_cast` here? `size` returns a `size_t` and `wsIndex` is a `size_t`.

MantidPlot/src/Mantid/MantidPlotUtilities.cpp
```diff
@@ -0,0 +1,71 @@
...
+    auto it = logValues.begin();
+    std::advance(it, wsIndex);
```
AntonPiccardoSelg: :+1:

MantidPlot/src/Mantid/MantidUI.cpp
```diff
@@ -3090,11 +3108,28 @@ MultiLayer *MantidUI::plot1D(const QMultiMap<QString, int> &toPlot,
...
   for (QMultiMap<QString, int>::const_iterator it = toPlot.begin();
        it != toPlot.end(); ++it) {
+
     try {
-      auto *wsCurve = new MantidMatrixCurve(it.key(), g, it.value(), indexType,
-                                            errs, plotAsDistribution, style);
+      if (!log.isEmpty()) { // Get log value from workspace
+        double logVal;
+        if (customLogValues.size() > 0) {
+          logVal = getSingleWorkspaceLogValue(i++, customLogValues);
```
AntonPiccardoSelg: Why are we incrementing the iterator here?

MantidPlot/src/Mantid/MantidUI.h
```diff
@@ -18,6 +19,7 @@
 #include "MantidAPI/IPeaksWorkspace_fwd.h"
 #include "MantidAPI/MatrixWorkspace_fwd.h"
 #include "MantidAPI/Workspace_fwd.h"
+#include "MantidAPI/WorkspaceGroup_fwd.h"
```
AntonPiccardoSelg: Do we need to the forward declaration of the `WorkspaceGroup` any longer?

MantidQt/MantidWidgets/src/MantidWSIndexDialog.cpp
```diff
@@ -243,21 +410,152 @@ void MantidWSIndexWidget::initSpectraBox() {
...
+* in every workspace)
+*/
+void MantidWSIndexWidget::populateLogComboBox() {
+  // First item should be "Workspace index"
+  m_logSelector->addItem(WORKSPACE_INDEX);
+
+  // Create a table of all single-value numeric log names versus
+  // how many workspaces they appear in
+  std::map<std::string, int> logCounts;
+  for (auto wsName : m_wsNames) {
```
AntonPiccardoSelg: take this by reference, ie `auto& wsName : m_wsNames`. This avoids the copy of each element

MantidQt/MantidWidgets/src/MantidWSIndexDialog.cpp
```diff
@@ -243,21 +410,152 @@ void MantidWSIndexWidget::initSpectraBox() {
...
+    auto ws = getWorkspace(wsName);
+    if (ws) {
+      const std::vector<Mantid::Kernel::Property *> &logData =
+          ws->run().getLogData();
+      for (auto log : logData) {
```
AntonPiccardoSelg: See above

MantidQt/MantidWidgets/src/MantidWSIndexDialog.cpp
```diff
@@ -243,21 +410,152 @@ void MantidWSIndexWidget::initSpectraBox() {
...
+            logCounts[name] = 1;
+          }
+        }
+      }
+    }
+  }
+
+  // Add the log names to the combo box if they appear in all workspaces
+  const int nWorkspaces = m_wsNames.size();
+  for (auto logCount : logCounts) {
```
AntonPiccardoSelg: See above

MantidQt/MantidWidgets/src/MantidWSIndexDialog.cpp
```diff
@@ -367,25 +665,28 @@ bool MantidWSIndexWidget::usingSpectraNumbers() const {
  * @param wsNames :: the names of the workspaces to be plotted
  * @param showWaterfallOption :: If true the waterfall checkbox is created
  * @param showPlotAll :: If true the "Plot all" button is created
- * @param showTiledOption :: If true the "Tiled" checkbox is created
+ * @param showTiledOption :: If true the "Tiled" option is created
+ * @param isAdvanced :: true if adanced plotting dialog is created
  */
 MantidWSIndexDialog::MantidWSIndexDialog(QWidget *parent, Qt::WFlags flags,
                                          QList<QString> wsNames,
```
AntonPiccardoSelg: This is not your change, but we should check if taking this via `const &` is the right thing to do here.

MantidQt/MantidWidgets/test/WorkspacePresenter/WorkspacePresenterTest.h
```diff
@@ -531,17 +531,23 @@ class WorkspacePresenterTest : public CxxTest::TestSuite {
...
+    EXPECT_CALL(*mockView.get(), plotSpectrum("Errors")).Times(Exactly(1));
     presenter->notifyFromView(ViewNotifiable::Flag::PlotSpectrumWithErrors);
     TS_ASSERT(Mock::VerifyAndClearExpectations(&mockView));
   }
 
+  void testPlotSpectrumAdvanced() {
+    EXPECT_CALL(*mockView.get(), plotSpectrum("Advanced")).Times(Exactly(1));
+    presenter->notifyFromView(ViewNotifiable::Flag::PlotSpectrumAdvanced);
+    TS_ASSERT(Mock::VerifyAndClearExpectations(&mockView));
+  }
```
AntonPiccardoSelg: :+1:

AntonPiccardoSelg (changes requested):

MantidPlot/src/Mantid/MantidGroupPlotGenerator.cpp
KarlPalmen: Renamed MantidSurfaceContourPlotGenerator as implied by opening comment in .h file.

MantidPlot/src/Mantid/MantidPlotUtilities.cpp
KarlPalmen: The code in MantidplotUtilities was moved by me from MantidGroupPlotGenerator to make it available elsewhere. I'm happy to make simple improvements in its code.

AntonPiccardoSelg: That was just a thumbs up for good use of `std::advance`

KarlPalmen: I was referring to all comments about `MantidPlotUtilities`.

MantidQt/MantidWidgets/src/MantidWSIndexDialog.cpp
KarlPalmen: The `wsnames` is set by the user and so `const` is not appropriate, but perhaps `&` is. This is also the same for the `MantidWSIndexWidget`.

MantidPlot/src/Mantid/MantidUI.cpp
KarlPalmen: Why not?

AntonPiccardoSelg: Hmm, I read it wrong. I read `i` as `it`. It is only a counter, that is fine.  This makes more sense now to me.

KarlPalmen:
I've added issue #19259 for the Release Notes.

AntonPiccardoSelg:
Issues found while testing:

### Single workspace with single spectrum does not open when clicking `Plot Spectrum` (or `Plot Spectrum with Errors`
 You can create the workspace with 
```python
CreateSampleWorkspace(OutputWorkspace='ws3', Function='Flat background', NumBanks=1, BankPixelWidth=1, XUnit='MomentumTransfer', XMax=10, BinWidth=1)
```

KarlPalmen:
I had noticed that if one put in too few custom log values, the remaining log values are shown as 0 in a waterfall or 1D plot. I didn't consider this an important issue, because the user should input as many custom log values as workspaces.

AntonPiccardoSelg:
Yes, the user should provide one log value per workspace. Should we warn the user if this is not being provided or should we just default to 0?

AntonPiccardoSelg:
Some more issues:

1. Run the temp script where the workspaces have multiple spectra
2. Select spectrum 1-2
3. Select waterfall plot and plot
  * Confirm that the labels are duplicates. I recall that we clarified with Nick that the label should be something like *templable*-spX.

KarlPalmen:
When I tried this I got `temp10-sp-1`, `temp10-sp-2`, `temp100-sp-1`, `temp100-sp2`, `temp20-sp-1`, `temp20-sp2` etc. Each label is unique.

KarlPalmen:
I then realised I did this for 1D-plot, when I tried waterfall plot I found the legend had been moved up so only the bottom part was showing. I could move it down and see the same labels as for 1D-plot.

AntonPiccardoSelg:
I get: 

image

KarlPalmen:
You didn't mention selecting `temp` as the log.

AntonPiccardoSelg:
You are right, the instructions have been updated.

NickDraper:
If you use this script
``` python
base = CreateSampleWorkspace(Function='Powder Diffraction', NumBanks=1, BankPixelWidth=1)

wsNames = []
for t in range (10,110,10):
    wsName = "temp" + str(t)
    wsNames.append(wsName)
    MultiplyRange(base,75,99,t/100.0,OutputWorkspace=wsName)
    AddSampleLog(wsName,"Temp",str(t),"Number","Kelvin")
TestGroup = GroupWorkspaces(wsNames)
```

You get workspaces with only one spectrum, in this case you do not get the plot Advanced gui, and I think you should.

NickDraper:
Log values are being sorted in text order rather than numeric order
image

AntonPiccardoSelg:
Hi, 

Nick just tested the PR and there is a set of requirements and changes that we should consider. All of this refers to `Plot Spectrum Advanced`

1. `Plot Spectrum Advanced` should be renamed to `Plot Advanced`
2. When the user forgets to enter a spectrum number or a workspace index in the two line edit fields in the top, a warning box should appear
3. The custom input field is not disabled when opening the `Plot Spectrum Advanced` the first time.
4. The GUI itself needs to be reorder slightly. We should place the `Show Error Bars` checkbox underneath the combo box which contains `1D Plot`, etc. Next to this combobox we should add a lable with the name `Plot Type`
5. The elements concerning labels, ie everything below the `Show Error Bars`  should be grouped.
6. When the user tries to open `Plot Advanced` with workspaces that have only one spectrum, the values are plotting immediately without prompting a GUI. This is correct for `Plot Spectrum` and `Plot Spectrum With Errors` but in the case of `Plot Advanced` we should get the GUI.
7. The last settings for anything but the workspace index line edit field and the spectrum number line edit field should be persisted via QSettings.
8. WSIndex for labelling does not make much sense for the `Plot1D` family. In fact we need to check if it makes sense at all
9. When dealing with log values (or custom values) we need to order the labels in an ascending order. This means we would have an order of, for example, `10-sp1, 10-sp2, 20-sp1, 20-sp2, 100-sp1, 100-sp2` instead of `10-sp1, 10-sp2, 100-sp1, 100-sp2, 20-sp1, 20-sp2` which is currently being used.

KarlPalmen:
I've found out that the surface plot graph depends on how the workspaces in the group are named. This situation does not apply before this issue. The following script demonstrates it
```python
base = CreateSampleWorkspace(Function='Powder Diffraction', NumBanks=4, BankPixelWidth=1)

wsNames = []

for t in range (10,110,10):
    wsName = "temp" + str(t)
    wsNames.append(wsName)
    MultiplyRange(base,75,99,t/100.0,OutputWorkspace=wsName)
    AddSampleLog(wsName,"Temp",str(t),"Number","Kelvin")

TestGroup = GroupWorkspaces(wsNames)

# Now the same again, but the numbers in the names have 100 added
wsNames100 = []

for t in range (10,110,10):
    wsName = "temp" + str(t+100)
    wsNames100.append(wsName)
    MultiplyRange(base,75,99,t/100.0,OutputWorkspace=wsName)
    AddSampleLog(wsName,"Temp",str(t),"Number","Kelvin")

TestGroup100 = GroupWorkspaces(wsNames100)
```

AntonPiccardoSelg:
### Retesting the change requests from above:

1. `Plot Spectrum Advanced` should be renamed to `Plot Advanced` -> Open
2. When the user forgets to enter a spectrum number or a workspace index in the two line edit fields at the top, a warning box should appear. -> Open
3. The custom input field is not disabled when opening the `Plot Spectrum Advanced` for the first time. -> Open
4. The GUI itself needs to be reordered slightly. We should place the `Show Error Bars` check box underneath the combo box which contains `1D Plot`, etc. Next to this combo box we should add a label with the name `Plot Type`. -> Open
5. The elements concerning labels, ie everything below the `Show Error Bars`  should be grouped. -> Open
6. ~~When the user tries to open `Plot Advanced` with workspaces that have only one spectrum, the values are plotted immediately without prompting a GUI. This is correct for `Plot Spectrum` and `Plot Spectrum With Errors` but in the case of `Plot Advanced` we should get the GUI.~~ -> Done
7. The last settings for anything but the workspace index line edit field and the spectrum number line edit field should be persisted via QSettings. -> Open
8. WSIndex for labelling does not make much sense for the `Plot1D`, `Waterfall Plot` and `Tiled Plot`. In fact we need to check if it makes sense at all. -> Open
9. When dealing with log values (or custom values) we need to order the labels in an ascending order. This means we would have an order of, for example, `10-sp1, 10-sp2, 20-sp1, 20-sp2, 100-sp1, 100-sp2` instead of `10-sp1, 10-sp2, 100-sp1, 100-sp2, 20-sp1, 20-sp2` which is currently being used. -> Cannot check since I don't see the legend

AntonPiccardoSelg:
Going through the remaining points:

- [x] Plot Spectrum Advanced should be renamed to Plot Advanced
- [x] When the user forgets to enter a spectrum number or a workspace index in the two line edit fields at the top, a warning box should appear. **Comment** the input field shows now a red star which is better than a pop ip.
- [x] The custom input field is not disabled when opening the Plot Spectrum Advanced for the first time.
- [x] The GUI itself needs to be reordered slightly. We should place the Show Error Bars check box underneath the combo box which contains 1D Plot, etc. Next to this combo box we should add a label with the name Plot Type.
- [x] The elements concerning labels, ie everything below the Show Error Bars should be grouped.
- [x] When the user tries to open Plot Advanced with workspaces that have only one spectrum, the values are plotted immediately without prompting a GUI. This is correct for Plot Spectrum and Plot Spectrum With Errors but in the case of Plot Advanced we should get the GUI.
- [ ]  The last settings for anything but the workspace index line edit field and the spectrum number line edit field should be persisted via QSettings. **Comment** This will addressed in a separate issue. This does not break any old functionality
- [ ] WSIndex for labelling does not make much sense for the Plot1D, Waterfall Plot and Tiled Plot. In fact we need to check if it makes sense at all. **Comment** Name is changed to `Worspace name` for the standard plots. There is still a small inconsistency to be addressed when switching between the plot options. I will show you on my machine.
- [x] When dealing with log values (or custom values) we need to order the labels in an ascending order. This means we would have an order of, for example, 10-sp1, 10-sp2, 20-sp1, 20-sp2, 100-sp1, 100-sp2 instead of 10-sp1, 10-sp2, 100-sp1, 100-sp2, 20-sp1, 20-sp2 which is currently being used. :+1: 
image

Some minor points to address, before we send it to other people to test.

MantidPlot/src/Mantid/MantidPlotUtilities.h
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+* These utilities assist with plotting in Mantid
+*/
+// Structure to aid odering of plots
+typedef struct {
+  double logVal;
+  QString wsName;
+  int index;
+} CurveSpec;
```
AntonPiccardoSelg: You should be able to use `C++`-style here, ie `struct CurveSpect {...`

MantidPlot/src/Mantid/MantidSurfaceContourPlotGenerator.cpp
```diff
@@ -305,23 +287,24 @@ QString MantidGroupPlotGenerator::getXAxisTitle(
...
  * @throw std::logic_error if spectrum index not contained in workspace, or if
  * wsGroup contains workspaces other than MatrixWorkspaces
  */
-bool MantidGroupPlotGenerator::groupContentsHaveSameX(
-    const Mantid::API::WorkspaceGroup_const_sptr &wsGroup, const size_t index) {
-  if (!wsGroup) {
+bool MantidSurfaceContourPlotGenerator::groupContentsHaveSameX(
+    const std::vector<Mantid::API::MatrixWorkspace_const_sptr> &workspaces,
+    const size_t index) {
+  if (workspaces.size() == 0) {
```
AntonPiccardoSelg: better to use `workspaces.empty()`

MantidPlot/src/Mantid/MantidSurfaceContourPlotGenerator.h
```diff
@@ -0,0 +1,87 @@
...
+      int nWorkspaces);
+
+  /// Tests if WorkspaceGroup contents all have same X for given spectrum
+  static bool groupContentsHaveSameX(
+      const std::vector<Mantid::API::MatrixWorkspace_const_sptr> &workspaces,
+      const size_t index);
+
+private:
+  /// Type of graph to plot
+  enum class Type { Surface, Contour };
```
AntonPiccardoSelg: :+1: good use of `enum classs`

MantidPlot/src/Mantid/MantidUI.cpp
```diff
@@ -3085,21 +3109,30 @@ MultiLayer *MantidUI::plot1D(const QMultiMap<QString, int> &toPlot,
...
+
     try {
-      auto *wsCurve = new MantidMatrixCurve(it.key(), g, it.value(), indexType,
-                                            errs, plotAsDistribution, style);
-      if (!firstCurve) {
-        firstCurve = wsCurve;
-        g->setNormalizable(firstCurve->isNormalizable());
-        g->setDistribution(firstCurve->isDistribution());
+      if (!log.isEmpty()) { // Get log value from workspace
+        if (customLogValues.size() > 0) {
```
AntonPiccardoSelg: why not `!customLogValues.empty()` ?

MantidPlot/src/Mantid/MantidUI.cpp
```diff
@@ -3935,72 +3995,45 @@ void MantidUI::updateRecentFilesList(const QString &fname) {
...
+ * @param isAdvanced :: [input] Whether to do advanced plotting
  * @returns :: New dialog
  */
-MantidWSIndexDialog *MantidUI::createWorkspaceIndexDialog(int flags,
-                                                          QStringList wsNames,
-                                                          bool showWaterfall,
-                                                          bool showPlotAll,
-                                                          bool showTiledOpt) {
+MantidWSIndexDialog *
+MantidUI::createWorkspaceIndexDialog(int flags, QStringList wsNames,
```
AntonPiccardoSelg: Should we pass the `wsNames` as `const &` here?

MantidQt/MantidWidgets/src/MantidTreeWidget.cpp
```diff
@@ -189,24 +189,33 @@ QList<MatrixWorkspace_const_sptr>
...
+  std::set<QString> selectedWsNameSet;
+  std::vector<QString> selectedWsNameList;
   foreach (const QString wsName, this->getSelectedWorkspaceNames()) {
     const auto groupWs = boost::dynamic_pointer_cast<const WorkspaceGroup>(
         m_ads.retrieve(wsName.toStdString()));
     if (groupWs) {
       const auto childWsNames = groupWs->getNames();
       for (auto childWsName : childWsNames) {
-        selectedWsNames.insert(QString::fromStdString(childWsName));
+        if (selectedWsNameSet.find(QString::fromStdString(childWsName)) ==
```
AntonPiccardoSelg: We should iterate over the names via `const &` then we avoid copying the whole container.

AntonPiccardoSelg:
- [x]`Ubuntu 16.04`. Comparing the behaviour of this PR and the latest release seems to be ok.
- [ ] OSX
- [ ] Windows

NickDraper:
When adding custom values that are not numeric you get en error message, so "a,b,c,d" fails

KarlPalmen:
On testing an old version of Mantid, I found that non-numeric custom log values were never accepted.

AntonPiccardoSelg (approved):

KarlPalmen:
Nick's third item of feedback from testing was easy to fix and has been done. The other two would be harder to fix, because log values are currently conveyed as double and not strings and the correct number of custom values is not readily available as the dialog box.

KarlPalmen:
This PR is ready for testing again.

KarlPalmen:
Pull Request #19485 is for the release notes of this issue.

NickDraper:
Custom value handling is much better now

peterfpeterson:
@quantumsteve Please try this out on osx.

MantidPlot/src/Mantid/MantidPlotUtilities.cpp
```diff
@@ -0,0 +1,80 @@
...
+ * @param matrixWS :: [input] Workspace to find log from
+ * @param logName :: [input] Name of log
+ * @returns log value as a double, or workspace index
+ * @throws invalid_argument if log is wrong type or not present
+ */
+double getSingleWorkspaceLogValue(
+    size_t wsIndex, const Mantid::API::MatrixWorkspace_const_sptr &matrixWS,
+    const QString &logName) {
+  if (logName == MantidWSIndexWidget::WORKSPACE_INDEX || logName == "") {
+    return (double)wsIndex; // cast for plotting
```
samueljackson92: Don't use a C style cast. Use `static_cast<double>`

samueljackson92: We also `return` here, so we can use an [early exit](http://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code) to remove the else statement and reduce code complexity

MantidPlot/src/Mantid/MantidPlotUtilities.cpp
```diff
@@ -0,0 +1,80 @@
...
+* i.e. the nth in order from smallest to largest.
+* If the index is outside the range, returns 0.
+* @param wsIndex :: [input] Index of log value to use
+* @param logValues :: [input] User-provided set of log values
+* @returns Numeric log value
+*/
+double getSingleWorkspaceLogValue(size_t wsIndex,
+                                  const std::set<double> &logValues) {
+  double value = 0;
+  if (wsIndex < logValues.size()) {
```
samueljackson92: Could also use an early exit here.

MantidPlot/src/Mantid/MantidPlotUtilities.h
```diff
@@ -0,0 +1,28 @@
...
+*/
+// Structure to aid ordering of plots
+struct CurveSpec {
+  double logVal;
+  QString wsName;
+  int index;
+};
+
+// Compare to sort according to log value
+bool byLogValue(const CurveSpec &lhs, const CurveSpec &rhs);
```
samueljackson92: Three slashes (`///`) needed for doxygen comment.

MantidPlot/src/Mantid/MantidUI.cpp
```diff
@@ -3108,6 +3141,33 @@ MultiLayer *MantidUI::plot1D(const QMultiMap<QString, int> &toPlot,
...
+    auto *wsCurve = new MantidMatrixCurve(
+        logValue, curveSpecList[i].wsName, g, curveSpecList[i].index, indexType,
+        errs, plotAsDistribution, style, multipleSpectra);
+    if (!firstCurve) {
+      firstCurve = wsCurve;
+      g->setNormalizable(firstCurve->isNormalizable());
+      g->setDistribution(firstCurve->isDistribution());
+    }
+  }
+
```
samueljackson92: Could we split this block for adding curves into a new method?

MantidPlot/src/Mantid/MantidUI.cpp
```diff
@@ -3085,21 +3109,30 @@ MultiLayer *MantidUI::plot1D(const QMultiMap<QString, int> &toPlot,
...
+                  it.key().toStdString());
+          curveSpec.logVal = getSingleWorkspaceLogValue(1, workspace, log);
+        }
+      } else {
+        curveSpec.logVal = 0.1234; // This should not be used.
       }
+      curveSpec.wsName = it.key();
+      curveSpec.index = it.value();
+      curveSpecList.push_back(curveSpec);
+
```
samueljackson92: Code is quite long and complex. Perhaps we could split it into a new method?

MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/WorkspacePresenter/QWorkspaceDockView.h
```diff
@@ -129,7 +124,7 @@ class EXPORT_OPT_MANTIDQT_MANTIDWIDGETS QWorkspaceDockView
   void showWorkspaceData() override;
   void saveToProgram() override;
   void showInstrumentView() override;
-  void plotSpectrum(bool showErrors) override;
+  void plotSpectrum(std::string type) override;
```
samueljackson92: `const std::string&`

samueljackson92 (dismissed):
A few minor code comments added below. I also noticed a couple of issues when testing on OSX.

1. The plot legend is missing on waterfall plots
screen shot 2017-05-08 at 14 24 52
2. The "Show Error Bars" text appears squished on OSX.
screen shot 2017-05-08 at 14 25 28

3. Custom log values seem to work a little inconsistently. For example if I put in `1,2,1,2` I get back labels `0, 0, 1, 2`

4. It appears I cannot edit the label for the plot axis. Even when the log value is set to custom

KarlPalmen:
Thank you @samueljackson92 for reviewing this. The point about the waterfall plot legend not always being shown is addressed in issue #19208 and applies already. The axis name only has effect in surface and contour plots and should be enabled for such plots. I'll address other issues later on.

samueljackson92:
@KarlPalmen Ahhh I misunderstood the purpose of the axis name.

KarlPalmen:
I looked into the issues arising from duplicate custom log values. In doing so, I checked the behaviour of Mantid without this PR and found that the validation of the custom logs was done against the number of _distinct_ log values.  This suggests that the validation is not correct for duplicate custom log values and Samuel's case should be rejected, because there are only 2 log values. However with waterfall and 1D plots different considerations may apply.

MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/WorkspacePresenter/QWorkspaceDockView.h
KarlPalmen: Function is called like this ``  view->plotSpectrum("Simple");`` so can't use`` string&``.

MantidPlot/src/Mantid/MantidUI.cpp
KarlPalmen: I've put this one into a new method (with a little help from you).

MantidPlot/src/Mantid/MantidUI.cpp
KarlPalmen: It would require 6 arguments for just 20 lines.

AntonPiccardoSelg:
Please :shipit:

PR #19243, opened 2017-03-24
Add experimental MPI build

### LINKED ISSUE #19205: Create MPI build for Mantid
Create MPI for generic MPI support in Mantid.

`Framework`-only?.

Should be separate from the SNS MPI build? -> Yes, since at least temporarily our changes will break existing SNS builds.

SimonHeybrock:
This PR adds an MPI build for Mantid. There is an existing MPI build (used at SNS) that can be enabled with `cmake` flag `-DMPI_BUILD`. However, changes for MPI support (for workspaces and algorithms) that will be done as follow-ups to this PR will not be compatible with the existing MPI functionality. Therefore, we are here adding a new MPI build, currently enabled with `-DMPI_EXPERIMENTAL`.

Just like the existing MPI build this uses `boost::mpi` internally. The main addition in this PR is however a (incomplete) wrapper for `boost::mpi`. The main purpose of this is to allow testing without running tests with MPI. To this end, if Mantid is built without MPI, the wrapper is internally using threading to mimic MPI and code running on several ranks.

**To test:**

Code review.

Fixes #19205.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


quantumsteve:
@SimonHeybrock I'm uncomfortable with changing between threads and processes. Our five year plan for visualization includes starting a `pvserver` in each MPI process. The threading backend appears to prevent this.

SimonHeybrock:
@quantumsteve I am not sure I understand your question, but if you compile with MPI enabled the threading backend is not used. This is simply for unit testing MPI code in non-MPI builds.

SimonHeybrock:
@quantumsteve I admit that this is probably not useful everywhere. If all client code that uses MPI calls is wrapped in `#ifdef MPI...` the threading backend is useless. How are you unit testing your MPI code?

quantumsteve:
@SimonHeybrock The idea is part of the [5 year plan](https://github.com/mantidproject/documents/blob/cf3ad5edaafbd309a5adaafbaaf988784e842d0a/Design/ORNL_Mantid_5yearplan.pdf) I haven't tried moving the VSI to use MPI, but want to ensure such a design remains possible.

SimonHeybrock:
@quantumsteve I don't think what I do here would cause issues. You can still write code that directly calls MPI. The threading backend is really just meant for unit testing, it is *not* an attempt to provide something like hybrid MPI/threading or anything like that.

Framework/CMakeLists.txt
```diff
@@ -57,6 +57,10 @@ if ( ${CMAKE_SYSTEM_NAME} MATCHES "Linux" )
   endif ( MPI_BUILD )
 endif ()
 
+if ( MPI_EXPERIMENTAL )
```
OwenArnold: Long-term I don't think that `MPI_EXPERIMENTAL` is what we would want to call it. Could we try to select a suitable flag name now. `MPI_ALGORITHMS` etc?

Framework/CMakeLists.txt
```diff
@@ -70,6 +74,9 @@ add_subdirectory (Types)
 include_directories (Kernel/inc)
 add_subdirectory (Kernel)
 
+include_directories (Parallel/inc)
+add_subdirectory (Parallel)
```
OwenArnold: Wouldn't it be better to wrap this in the condition `if (MPI_EXPERIMENTAL)` rather than the large block condition in `Framework/Parallel/CMakeLists.txt`?

Framework/Parallel/CMakeLists.txt
```diff
@@ -0,0 +1,71 @@
...
+	${MPI_TEST_FILES}
+)
+
+if (COVERALLS)
+  foreach( loop_var ${SRC_FILES} ${INC_FILES})
+    set_property(GLOBAL APPEND PROPERTY COVERAGE_SRCS "${CMAKE_CURRENT_SOURCE_DIR}/${loop_var}")
+  endforeach(loop_var)
+endif()
+
+if(UNITY_BUILD)
```
OwenArnold: You can omit this, I think we need to go through and remove the remaining `UNITY_BUILD` macros and implementations at some point as they are unused.

Framework/Parallel/inc/MantidParallel/Communicator.h
```diff
@@ -0,0 +1,95 @@
...
+  template <typename... T> void recv(T &&... args) const;
+  template <typename... T> Request isend(T &&... args) const;
+  template <typename... T> Request irecv(T &&... args) const;
+
+private:
+  Communicator(boost::shared_ptr<detail::ThreadingBackend> backend,
+               const int rank);
+
+  boost::mpi::communicator m_communicator;
+  boost::shared_ptr<detail::ThreadingBackend> m_backend;
```
OwenArnold: How would the `m_backend` be swapped in/out?

OwenArnold (commented):
Some initial comments. I'd like to review this again later.

Framework/CMakeLists.txt
SimonHeybrock: It has nothing to do with algorithms. My intention is to merge it with the existing SNS MPI build, which is named `MPI_BUILD`, but as explained in the PR I cannot do that right now.

Framework/CMakeLists.txt
SimonHeybrock: There are (or rather will be) bits in `Parallel` that are available also in non-MPI builds, I think.

Framework/Parallel/inc/MantidParallel/Communicator.h
SimonHeybrock: It can only be initialized by a private constructor, accessible by `ParallelRunner`. This is not intented for generic use but only unit tests, so I do not provide another way for access.

OwenArnold: I missed the fact you were checking `m_backend` every time before you used it.

LamarMoore (commented):
Looks good. Only one comment.

LamarMoore:
I still think it might be a good idea to get the third-party repo configured for boost::mpi if it is possible.

Framework/Parallel/inc/MantidParallel/ThreadingBackend.h
```diff
@@ -0,0 +1,134 @@
...
+    oa.operator<<(std::forward<T>(args)...);
+  }
+  std::lock_guard<std::mutex> lock(m_mutex);
+  m_buffer[std::make_tuple(source, dest, tag)].push_back(std::move(buf));
+}
+
+template <typename... T>
+void ThreadingBackend::recv(int dest, int source, int tag, T &&... args) {
+  const auto key = std::make_tuple(source, dest, tag);
+  while (true) {
```
SimonHeybrock: Yes, that is what MPI does. I think most implementations just hang indefinitely. Maybe it is possible to configure a timeout, but I am not sure.

LamarMoore:
MPI build on windows issue here #19274.

LamarMoore:
:shipit:

PR #19247, opened 2017-03-24
Fix WorkspaceCreation::create(const P &parent) for parent workspaces with varying bins

### LINKED ISSUE #19242: WorkspaceCreation::create(const P &) doesn't clone the entire X data
### Expected behavior
`WorkspaceCreation::create(const P &parent)` should return a clone of the parent workspace with Y and E data zeroed.

### Actual behavior
The method copies the X values from the first histogram over the entire output workspace. This is incorrect if the parent has varying bins over each histogram.

Change the `buildDir` variable to your liking and run the following script:
```python
buildDir = <path-to-Mantid-build-directory>
ws = Load(Filename=buildDir + 'ExternalData/Testing/Data/UnitTest/ILL/IN4/084446.nxs')
ws = ConvertUnits(ws, Target='Wavelength', EMode='Direct')
geometry = {
    'Shape': 'Cylinder',
    'Height': 10.0,
    'Radius': 0.5,
    'Center': [0.0, 0.0, 0.0]
}
material = {
    'ChemicalFormula': 'O2',
    'SampleMassDensity': 1
}
SetSample(ws, Geometry=geometry, Material=material)
corrections = MonteCarloAbsorption(ws, 
    NumberOfWavelengthPoints=2, 
    EventsPerPoint=1)
```
Compare the X data of `ws` and `corrections` for workspace index > 300. The data should match but it does not.

soininen:
`WorkspaceCreation::create(const P &parent)` always copied the X data of the first histogram in the parent workspace to the created one. This would lead to incorrect binning in the created workspace if the parent had varying bins.

**To test:**

The symptoms can be seen e.g. when using the [MonteCarloAbsorption](http://docs.mantidproject.org/nightly/algorithms/MonteCarloAbsorption-v1.html) algorithm. In the following script, change the `buildDir` variable to point to Mantid's build dir and run.
```python
buildDir = <path-to-Mantid-build-directory>
ws = Load(Filename=buildDir + 'ExternalData/Testing/Data/UnitTest/ILL/IN4/084446.nxs')
ws = ConvertUnits(ws, Target='Wavelength', EMode='Direct')
geometry = {
    'Shape': 'Cylinder',
    'Height': 10.0,
    'Radius': 0.5,
    'Center': [0.0, 0.0, 0.0]
}
material = {
    'ChemicalFormula': 'O2',
    'SampleMassDensity': 1
}
SetSample(ws, Geometry=geometry, Material=material)
corrections = MonteCarloAbsorption(ws, 
    NumberOfWavelengthPoints=2, 
    EventsPerPoint=1)
```
After the unit conversion `ws` has varying bins for workspace indices > 300. Binning should be the same for `corrections` after this PR has been applied.

Fixes #19242.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/src/DiffractionEventCalibrateDetectors.cpp
+++ b/Framework/Algorithms/src/DiffractionEventCalibrateDetectors.cpp
@@ -292,8 +292,7 @@ void DiffractionEventCalibrateDetectors::exec() {
   // Get some stuff from the input workspace
   // We make a copy of the instrument since we will be moving detectors in
   // `inputW` but want to access original positions (etc.) via `detList` below.
-  const auto &dummyW = create<EventWorkspace>(*inputW, 1);
-  Instrument_const_sptr inst = dummyW->getInstrument();
+  Instrument_const_sptr inst(inputW->getInstrument()->baseInstrument()->clone());
 
   // Build a list of Rectangular Detectors
   std::vector<boost::shared_ptr<RectangularDetector>> detList;
--- a/Framework/Algorithms/src/DiffractionEventCalibrateDetectors.cpp
+++ b/Framework/Algorithms/src/DiffractionEventCalibrateDetectors.cpp
@@ -292,7 +292,8 @@ void DiffractionEventCalibrateDetectors::exec() {
   // Get some stuff from the input workspace
   // We make a copy of the instrument since we will be moving detectors in
   // `inputW` but want to access original positions (etc.) via `detList` below.
-  Instrument_const_sptr inst(inputW->getInstrument()->baseInstrument()->clone());
+  Instrument_const_sptr inst(
+      inputW->getInstrument()->baseInstrument()->clone());
 
   // Build a list of Rectangular Detectors
   std::vector<boost::shared_ptr<RectangularDetector>> detList;
--- a/Framework/Algorithms/src/DiffractionEventCalibrateDetectors.cpp
+++ b/Framework/Algorithms/src/DiffractionEventCalibrateDetectors.cpp
@@ -292,8 +292,9 @@ void DiffractionEventCalibrateDetectors::exec() {
   // Get some stuff from the input workspace
   // We make a copy of the instrument since we will be moving detectors in
   // `inputW` but want to access original positions (etc.) via `detList` below.
-  Instrument_const_sptr inst(
-      inputW->getInstrument()->baseInstrument()->clone());
+  const auto &dummyW = create<EventWorkspace>(*inputW, 1,
+                                              inputW->binEdges(0));
+  Instrument_const_sptr inst = dummyW->getInstrument();
 
   // Build a list of Rectangular Detectors
   std::vector<boost::shared_ptr<RectangularDetector>> detList;
--- a/Framework/Algorithms/src/DiffractionEventCalibrateDetectors.cpp
+++ b/Framework/Algorithms/src/DiffractionEventCalibrateDetectors.cpp
@@ -292,8 +292,7 @@ void DiffractionEventCalibrateDetectors::exec() {
   // Get some stuff from the input workspace
   // We make a copy of the instrument since we will be moving detectors in
   // `inputW` but want to access original positions (etc.) via `detList` below.
-  const auto &dummyW = create<EventWorkspace>(*inputW, 1,
-                                              inputW->binEdges(0));
+  const auto &dummyW = create<EventWorkspace>(*inputW, 1, inputW->binEdges(0));
   Instrument_const_sptr inst = dummyW->getInstrument();
 
   // Build a list of Rectangular Detectors
--- a/Framework/DataObjects/inc/MantidDataObjects/WorkspaceCreation.h
+++ b/Framework/DataObjects/inc/MantidDataObjects/WorkspaceCreation.h
@@ -175,7 +175,8 @@ template <class T, class P, class IndexArg,
 std::unique_ptr<T> create(const P &parent, const IndexArg &indexArg,
                           const HistogramData::Histogram &histogram) {
   std::unique_ptr<T> ws = detail::createUninitialized<T, P>(parent);
-  ws->initialize(indexArg, HistogramData::Histogram(detail::stripData(histogram)));
+  ws->initialize(indexArg,
+                 HistogramData::Histogram(detail::stripData(histogram)));
   detail::initializeFromParent(parent, *ws);
 
   return ws;
@@ -195,7 +196,8 @@ template <class T, class IndexArg,
           typename std::enable_if<
               !std::is_base_of<API::MatrixWorkspace, IndexArg>::value>::type * =
               nullptr>
-std::unique_ptr<T> create(const IndexArg &indexArg, const HistogramData::Histogram &histogram) {
+std::unique_ptr<T> create(const IndexArg &indexArg,
+                          const HistogramData::Histogram &histogram) {
   auto ws = Kernel::make_unique<T>();
   HistogramData::Histogram histogramTemplate(histogram);
   if (std::is_base_of<DataObjects::EventWorkspace, T>::value) {
--- a/Framework/DataObjects/inc/MantidDataObjects/WorkspaceCreation.h
+++ b/Framework/DataObjects/inc/MantidDataObjects/WorkspaceCreation.h
@@ -132,6 +132,20 @@ template <class T, class P, class = typename std::enable_if<std::is_base_of<
                                 API::MatrixWorkspace, P>::value>::type>
 std::unique_ptr<T> createUninitialized(const P &parent) {
   std::unique_ptr<T> ws;
+  return ws;
+}
+
+MANTID_DATAOBJECTS_DLL void
+initializeFromParent(const API::MatrixWorkspace &parent,
+                     API::MatrixWorkspace &ws);
+}
+
+template <class T, class P, class IndexArg, class HistArg,
+          class = typename std::enable_if<
+              std::is_base_of<API::MatrixWorkspace, P>::value>::type>
+std::unique_ptr<T> create(const P &parent, const IndexArg &indexArg,
+                          const HistArg &histArg) {
+  std::unique_ptr<T> ws;
   // Figure out (dynamic) target type:
   // - Type is same as parent if T is base of parent
   // - If T is not base of parent, conversion may occur. Currently only
@@ -149,23 +163,8 @@ std::unique_ptr<T> createUninitialized(const P &parent) {
       ws = detail::createConcreteHelper<T>();
     }
   }
-  return ws;
-}
-
-MANTID_DATAOBJECTS_DLL void
-initializeFromParent(const API::MatrixWorkspace &parent,
-                     API::MatrixWorkspace &ws);
-}
-
-template <class T, class P, class IndexArg, class HistArg,
-          class = typename std::enable_if<
-              std::is_base_of<API::MatrixWorkspace, P>::value>::type>
-std::unique_ptr<T> create(const P &parent, const IndexArg &indexArg,
-                          const HistArg &histArg) {
-  std::unique_ptr<T> ws = detail::createUninitialized<T, P>(parent);
   ws->initialize(indexArg, HistogramData::Histogram(histArg));
   detail::initializeFromParent(parent, *ws);
-
   return ws;
 }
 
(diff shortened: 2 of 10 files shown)


SimonHeybrock:
I wrote these new factory methods a while ago, and considered exactly this problem. In the end, the reason I had decided against copying all the bins from the parent was the case of creating a smaller workspaces, e.g., with `create<Workspace2D>(parent, parent->getNumberHistograms()/2)`. In that case it could be undefined which `X` (i.e., which spectra) should be used, so I decided to default.

As this is fixed now, I think we are running exactly into this problem again: Consider, with the changes from this PR:

```cpp
ws1 = create<Workspace2D>(parent);
ws2 = create<Workspace2D>(parent, parent->getNumberHistograms());`
```

I think as it stands now, `ws2` will have `X` equal to `parent->x(0)`, whereas `ws1` has `X` identical to `parent`, for all spectra. Is that a good behavior? Should there be as size check?

I am not trying to say that I do not like copying the bins (I would actually prefer it :+1), but we have to be careful how to define edge cases, and make sure that the differences in behavior are documented.

SimonHeybrock:
Maybe one option would be to remove the variants that accept a size but not a `Histogram`?

```cpp
  create<T>(ParentWS, NumSpectra)
  create<T>(ParentWS, IndexInfo)
```

soininen:
These are very good points. I think forcing the clients to be explicit on how to fill the histogram data is the way to go, i.e. removing the size-only variants. I will try it out and see what will come.

Framework/Algorithms/src/DiffractionEventCalibrateDetectors.cpp
```diff
@@ -292,8 +292,7 @@ void DiffractionEventCalibrateDetectors::exec() {
   // Get some stuff from the input workspace
   // We make a copy of the instrument since we will be moving detectors in
   // `inputW` but want to access original positions (etc.) via `detList` below.
-  const auto &dummyW = create<EventWorkspace>(*inputW, 1);
-  Instrument_const_sptr inst = dummyW->getInstrument();
+  Instrument_const_sptr inst(inputW->getInstrument()->baseInstrument()->clone());
```
SimonHeybrock: I don't think this is correct, this way you are losing all position information stored in parameter map and in `DetectorInfo`. `DetectorInfo` is currently owned by `ExperimentInfo` (since Instrument-2.0 is not complete yet), so I think you need to keep creating a workspace.

Framework/DataObjects/src/WorkspaceCreation.cpp
```diff
@@ -8,8 +8,12 @@ namespace Mantid {
...
 namespace detail {
 HistogramData::Histogram stripData(HistogramData::Histogram histogram) {
-  histogram.setSharedY(nullptr);
-  histogram.setSharedE(nullptr);
+  if (histogram.ptrY()) {
+    histogram.setSharedY(nullptr);
+  }
+  if (histogram.ptrE()) {
+    histogram.setSharedE(nullptr);
+  }
```
SimonHeybrock: Why this change?

Framework/DataObjects/inc/MantidDataObjects/WorkspaceCreation.h
```diff
@@ -158,13 +149,39 @@ std::unique_ptr<T> create(const P &parent, const IndexArg &indexArg,
...
   detail::initializeFromParent(parent, *ws);
 
   return ws;
 }
 
+template <class T, class P, class IndexArg,
+          class = typename std::enable_if<
+              std::is_base_of<API::MatrixWorkspace, P>::value>::type>
+std::unique_ptr<T> create(const P &parent, const IndexArg &indexArg,
+                          const HistogramData::Histogram &histogram) {
```
SimonHeybrock: Why this new variant? I think it will actually break things since there is code that creates workspaces based on a histogram including data (via the templated version above).

Framework/DataObjects/inc/MantidDataObjects/WorkspaceCreation.h
```diff
@@ -175,19 +192,40 @@ std::unique_ptr<T> create(const IndexArg &indexArg, const HistArg &histArg) {
   return std::move(ws);
 }
 
+template <class T, class IndexArg,
+          typename std::enable_if<
+              !std::is_base_of<API::MatrixWorkspace, IndexArg>::value>::type * =
+              nullptr>
+std::unique_ptr<T> create(const IndexArg &indexArg,
+                          const HistogramData::Histogram &histogram) {
```
SimonHeybrock: Why this new variant? (see above)

Framework/DataObjects/inc/MantidDataObjects/WorkspaceCreation.h
```diff
@@ -175,19 +192,40 @@ std::unique_ptr<T> create(const IndexArg &indexArg, const HistArg &histArg) {
...
+
 template <class T, class P,
           typename std::enable_if<std::is_base_of<API::MatrixWorkspace,
                                                   P>::value>::type * = nullptr>
 std::unique_ptr<T> create(const P &parent) {
-  return create<T>(parent, parent.getNumberHistograms(),
-                   detail::stripData(parent.histogram(0)));
-}
+  const auto numHistograms = parent.getNumberHistograms();
+  if (parent.isCommonBins()) {
```
SimonHeybrock: I think the `isCommonBins` method is a bit broken, I would not use it. Simply remove this `if` and set up sharing X below, that should be both safer and faster.

Framework/DataObjects/inc/MantidDataObjects/WorkspaceCreation.h
```diff
@@ -175,19 +192,40 @@ std::unique_ptr<T> create(const IndexArg &indexArg, const HistArg &histArg) {
...
+    return create<T>(parent, numHistograms,
+                     detail::stripData(parent.histogram(0)));
+  }
+  std::unique_ptr<T> ws = detail::createUninitialized<T, P>(parent);
+  const auto YLength = parent.blocksize();
+  const auto XLength = parent.isHistogramData() ? YLength + 1 : YLength;
+  ws->initialize(numHistograms, XLength, YLength);
+  for (size_t i = 0; i < numHistograms; ++i) {
+    ws->mutableX(i) = parent.x(i);
```
SimonHeybrock: `ws->setSharedX(i, parent.sharedX(i));` sets up sharing and is much faster.

Framework/DataObjects/inc/MantidDataObjects/WorkspaceCreation.h
```diff
@@ -175,19 +192,40 @@ std::unique_ptr<T> create(const IndexArg &indexArg, const HistArg &histArg) {
...
-}
+  const auto numHistograms = parent.getNumberHistograms();
+  if (parent.isCommonBins()) {
+    return create<T>(parent, numHistograms,
+                     detail::stripData(parent.histogram(0)));
+  }
+  std::unique_ptr<T> ws = detail::createUninitialized<T, P>(parent);
+  const auto YLength = parent.blocksize();
+  const auto XLength = parent.isHistogramData() ? YLength + 1 : YLength;
+  ws->initialize(numHistograms, XLength, YLength);
```
SimonHeybrock: I think you can simply call the `create(parent, numHistograms, detail::stripData(parent.histogram(0)));` method here, and then set the correct X afterwards, the result is identical. As a result, you do not need to split out the `createUninitialized` method from the original `create` method. There should be no performance penalty to this.

SimonHeybrock (commented):
Some comments below. I think most of the changes to the templated methods are not necessary, and you can get the same result with less code (and much fewer changes):

Framework/DataObjects/inc/MantidDataObjects/WorkspaceCreation.h
soininen: `EventWorkspace::initialize()` won't accept histograms with non-NULL y or e data. Check the new unit test `test_create_event_from_event()`: it will throw at `create()` when using the original implementation above.
I added this variant to fix the problem. However, as you note, it erroneously  strips the data from histogram workspaces.
Maybe this variant (and the other one below) could be merged to the corresponding `histArg` accepting `create()`:
```c++
template <class T, class P, class IndexArg, class HistArg,
          class = typename std::enable_if<
              std::is_base_of<API::MatrixWorkspace, P>::value>::type>
std::unique_ptr<T> create(const P &parent, const IndexArg &indexArg,
                          const HistArg &histArg) {
  std::unique_ptr<T> ws = detail::createUninitialized<T, P>(parent);
  HistogramData::Histogram templateHist(histArg);
  if (std::is_base_of<DataObjects::EventWorkspace, T>::value) {
    templateHisto = detail::stripData(templateHist);
  }  
  ws->initialize(indexArg, templateHist);
  detail::initializeFromParent(parent, *ws);
  return ws;
}
```
This way we would get rid of the variants and handle EventWorkspaces gracefully. Or am I missing something?

SimonHeybrock: I think client code should simply use a `Histogram` with `null` y and e data in this case. Removing it silently might actually hide issues.

Framework/DataObjects/inc/MantidDataObjects/WorkspaceCreation.h
soininen: Good idea! `isCommonBins` is indeed not the most reliable of tests.

Framework/DataObjects/inc/MantidDataObjects/WorkspaceCreation.h
SimonHeybrock: @soininen Note that it is trivial to do that, since client code can simply use

```cpp
create<EventWorkspace>(parent, BinEdges{1,2,3,4});
```

The histogram with `null` Y and E is created automatically, so there is no burden on the client code.

Framework/DataObjects/src/WorkspaceCreation.cpp
soininen: I was just being to careful here. I'll remove it.

Framework/DataObjects/inc/MantidDataObjects/WorkspaceCreation.h
soininen: In the code above, the e and y data are removed only for EventWorkspaces, which don't use this data anyhow. However, if it is OK that
```c++
create<WorkspaceType>(parent, 2, parent.histogram(0));
```
works only if `WorkspaceType != EventWorkspace`, then I think I can completely revert the variants. Personally, I don't care either way, it is just that somebody might be surprised when EventWorkspaces suddenly cause runtime errors with something that works fine with Workspace2D.

SimonHeybrock: I prefer runtime errors over behavior that silently removes data --  correctness is the most important aspect I think!

Framework/DataObjects/inc/MantidDataObjects/WorkspaceCreation.h
```diff
@@ -130,6 +128,13 @@ template <>
 MANTID_DATAOBJECTS_DLL std::unique_ptr<API::HistoWorkspace>
 createConcreteHelper();
 
+template <class T, class P, class = typename std::enable_if<std::is_base_of<
+                                API::MatrixWorkspace, P>::value>::type>
+std::unique_ptr<T> createUninitialized(const P &parent) {
```
SimonHeybrock: Now unused?

docs/source/release/v3.10.0/framework.rst
```diff
@@ -57,7 +57,8 @@ As a consequence of these changes, :ref:`CopyInstrumentParmeters <algm-CopyInstr
 Bugs
 ----
 
-- We have fixed a bug where Mantid could crash when deleteing a large number of workspaces.
+- We have fixed a bug where Mantid could crash when deleting a large number of workspaces.
+- A bug was fixed where sometimes a workspace created using another one as a template would have incorrect X data. This happened only when the template workspace had varying bins between the histograms. In these cases the binning was not copied correctly from the template. The bug was discovered in :ref:`algm-MonteCarloAbsorption` but other algorithms may have been affected as well.
```
SimonHeybrock: These workspace factory methods are not exposed to Python, so I am not sure it makes sense to be so specific in the release notes. Maybe only report the fix for the affected algorithm?

soininen: I wasn't sure if this affected other algorithms/user scripts, so I wrote the entire story here. But perhaps you are right, this is not so big deal. I'll revise the release notes.

Framework/Algorithms/src/DiffractionEventCalibrateDetectors.cpp
```diff
@@ -292,8 +292,8 @@ void DiffractionEventCalibrateDetectors::exec() {
   // Get some stuff from the input workspace
   // We make a copy of the instrument since we will be moving detectors in
   // `inputW` but want to access original positions (etc.) via `detList` below.
-  const auto &dummyW = create<EventWorkspace>(*inputW, 1);
-  Instrument_const_sptr inst = dummyW->getInstrument();
+  Instrument_const_sptr inst(
+      inputW->getInstrument()->baseInstrument()->clone());
```
SimonHeybrock: @soininen I had previously commented on this, but I think it was lost in `clang-format`:

I don't think this is correct, this way you are losing all position information stored in parameter map and in `DetectorInfo`. `DetectorInfo` is currently owned by `ExperimentInfo` (since Instrument-2.0 is not complete yet), so I think you need to keep creating a workspace.

soininen: I see - I'll fix it. Sorry for missing your comment!

SimonHeybrock:
Latest changes look good. Overall I think this is a nice improvement of these factory methods :+1: 
:shipit: once builds pass.

PR #19262, opened 2017-03-29
Fix wrong component comparison in SpecularReflationPos.Cor.

SimonHeybrock:
The old code compared the parent `shared_ptr`. However, `IComponent::getParent` returns a newly allocated `IComponent`, so this gives nonsense results. Instead the *base* component must be compared.

Miraculously, the old implementation mostly worked. The reason is most likely that the old pointer immediately went out of scope and the next loop iteration would typically allocate the next parent in the same memory location.

Marked as high priority since it influences builds of all PRs.

**To test:**

Code review.

No issue, fixes regular test failures on Linux build servers.

No release notes.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/src/SpecularReflectionPositionCorrect.cpp
+++ b/Framework/Algorithms/src/SpecularReflectionPositionCorrect.cpp
@@ -40,9 +40,9 @@ getParentComponent(IComponent_const_sptr &currentComponent) {
  */
 bool hasCommonParent(const std::vector<IDetector_const_sptr> &detectors) {
   bool sameParentComponent = true;
-  IComponent const *lastParentComponent = detectors[0]->getParent().get();
+  auto lastParentComponent = detectors[0]->getParent()->getComponentID();
   for (size_t i = 1; i < detectors.size(); ++i) {
-    IComponent const *currentParentComponent = detectors[i]->getParent().get();
+    auto currentParentComponent = detectors[i]->getParent()->getComponentID();
     if (lastParentComponent != currentParentComponent) {
       sameParentComponent = false; // Parent components are different.
       break;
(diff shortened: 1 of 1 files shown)


OwenArnold:
Based on the behaviour of `getComponentID()` I think that this will not be correctly comparing the same pointers. i.e. always the base pointer.

SimonHeybrock:
@OwenArnold I am not sure I understand your comment?

OwenArnold:
Based on the behaviour of getComponentID() I think that this will NOW be correctly comparing the same pointers. i.e. always the base pointer. Which is what should happen. If tests pass :shipit:

rosswhitfield:
:+1: Also confirming that the `SpecularReflectionPositionCorrectTest` now passes consistently.

Framework/Algorithms/src/SpecularReflectionPositionCorrect.cpp
```diff
@@ -39,17 +39,17 @@ getParentComponent(IComponent_const_sptr &currentComponent) {
  * @return True only if all detectors have the same immediate parent.
  */
 bool hasCommonParent(const std::vector<IDetector_const_sptr> &detectors) {
-  bool sameParentComponent = true;
-  auto lastParentComponent = detectors[0]->getParent()->getComponentID();
+  auto firstParent = detectors[0]->getParent();
+  if (!firstParent)
```
OwenArnold: nullptr check looks to make things safer. Was this causing things to fail?

SimonHeybrock: @OwenArnold Yes, the `OFFSPECReflRedOneAutoPolarizationCorrection` system test.

OwenArnold: 👍

OwenArnold:
Latest round of changes look sensible. If tests pass :shipit: when green system tests.

PR #19271, opened 2017-03-30
Engg: Remove hard coded EnginX strings in preparation for IMAT

### LINKED ISSUE #17433: Engg GUI fitting tab: Refactoring umbrella issue
Moving the ideas from #17301 to a new issue following the work on that issue. 
Here are some of the improvements which can be completed on the fitting tab to make maintenance/future development easier:

**Maintenance Work**
- [x] Look for 'ENGINX_' or 'bank_' instead of hard coding expected file name positions 
- [ ] Return mapping from `findFilePathFromBaseName` of base names onto paths to avoid multiple calls
- [ ] Minimize/remove usage of `g_multi_run_directories` and `g_fitting_runno_counter`
- [ ] As part of #17285 the error in `browsedFile` needs to distinguish between directories that don't exist and if the user has not set their focused directory in settings
- [ ] Refactor `enableMultiRun` to either make recursion obvious or try to remove it and remove some levels of nesting (Idea: Create vector with all run numbers between range and pass it to `findFilePathFromBaseName` as part of first issue).
- [ ] Currently we handle browsed, multi-run and single files using different methods with different logic. This need to be changed so a common core logic is used and `fittingRunNoChanged` can complete simple calls to a multi-run, full path or single run entry in the input box
- [ ] Following this refactoring to different methods with shared common logic we should make the methods public so we can unit test them
- [ ] Each method seems to update the UI in its own way and the UI code can also trigger other UI updates as a side effect. This can be seen if you put a breakpoint in `EnggDiffFittingViewQtWidget::setFittingRunNo(...)`
- [ ] The scopes of any remaining try...catch blocks need to be reduced to the lines they are intended to wrap
- [ ] Ensure all unit tests call `VerifyAndClearExpectations` explicitly instead of letting the framework do it implicitly and the call is wrapped in an assert
- [ ] Where possible move the mocking over from `NiceMock` to `StrictMock`
- [ ] Move all `QString` usage into the view and make the interface convert to standard strings

**General Improvements**
- [ ] On the warning: "Unable to find one or more files with a name matching the run number or range input. Please check the files are present in the focused output directory and the input is correct." we should log or display the files which couldn't be found

**Problems which will be resolved by said refactoring**
- [x] When inputting multi-runs the input is not preserved (e.g. 23663-23664 becomes 23663) - _Fixed in refactor as it was causing confusion during PR testing_
- [ ] When using a numerical only file name if other files are present it will incorrectly use them. So `246043` loads `ENGINX_246043_bank_1.dat`
- [ ] The Plot Bank drop down box will not distinguish between textured and normally focused files.

dtasev:
Description of work.

Removed most hard coded EnginX name in preparation for adding IMAT.
Some are left because it's not clear to me what to do with them, for example:
`std::string EnggDiffractionViewQtGUI::guessGSASTemplatePath() const` has `template_ENGINX_241391_236516_North_and_South_banks.par`, and `std::string EnggDiffractionViewQtGUI::guessDefaultFullCalibrationPath() const` has `ENGINX_full_pixel_calibration_vana194547_ceria193749.csv`.

**To test:**
- Play around with GUI
- Run unit tests

Partially Fixes #17433.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/EnggDiffraction/EnggDiffFittingViewQtWidget.h
+++ b/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/EnggDiffraction/EnggDiffFittingViewQtWidget.h
@@ -159,6 +159,12 @@ public:
 
   void resetView();
 
+  std::string getCurrentInstrument() const override { return m_currentInst; }
+
+  void setCurrentInstrument(const std::string &newInstrument) override {
+    m_currentInst = newInstrument;
+  }
+
 protected:
   void initLayout();
 
@@ -240,6 +246,10 @@ private:
 
   /// presenter as in the model-view-presenter
   boost::scoped_ptr<IEnggDiffFittingPresenter> m_presenter;
+
+  /// current selected instrument
+  /// updated from the EnggDiffractionPresenter processInstChange
+  std::string m_currentInst = "";
 };
 
 } // namespace CustomInterfaces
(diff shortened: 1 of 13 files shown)


DavidFair (approved):

DavidFair:
The changes LGTM and remove a lot of the hard-coded EnginX strings
:shipit:

PR #18633, opened 2017-01-31
Histogram Rollout - SaveDetectorGroup to SaveMask

DavidFair:
HistogramData rollout: algorithms SaveDetectorGroup to SaveMask

## Description of work.
This PR is part of refactoring effort to use the new interface of the HistogramData module.
See https://github.com/mantidproject/mantid/issues/17641 for details.

`SaveFocussedXYETest` was renamed to `SaveFocusedXYETest` to match the algorithm name.

Headers were tided and c4250 warning suppression was removed from some headers - the rest have been completed in this PR with more details: #18632

The following algorithms were refactored to use HistogramData:

```
Framework/DataHandling/src/SaveDetectorsGrouping.cpp
Framework/DataHandling/src/SaveFITS.cpp
Framework/DataHandling/src/SaveFocusedXYE.cpp
Framework/DataHandling/src/SaveGSS.cpp
Framework/DataHandling/src/SaveMask.cpp
```

Performance tests were added for:

`SaveFocusedXYE`

The following algorithms have improved performance:
None - No algorithms copied histogram data so this is expected.
## To test.

Code review.

Fixes #18625 .
Internal change, no release notes (potential performance improvements will be added to release notes as part of the umbrella issue).

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [x] Is the code of an acceptable quality?
- [x] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)? Is it well structured with small focussed classes/methods/functions?
- [ ] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant documentation been added/updated?
- [ ] Is user-facing documentation written in a user-friendly manner?
- [ ] Has developer documentation been updated if required?

- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

PR #19276, opened 2017-03-31
SaveGSSTest - Removed duplicate code and Restore Performance Test Difficulty

### LINKED ISSUE #19267: SaveGSSTest Code Duplication and Performance Test Change
The SaveGSS Performance Test was originally written [here](https://github.com/mantidproject/mantid/issues/18625) and modified [here](https://github.com/mantidproject/mantid/issues/18569). I fear the characteristic of the unit test has changed which leaves the performance test results unreliable. There is also a duplicated method `generateTestMatrixWorkspace`, at first glance I cannot see why this needs to be different. It is currently unclear whether changes to this algorithm have resulted in a performance regression affecting other Algorithms.

http://builds.mantidproject.org/job/master_performancetests2/Master_branch_performance_tests/AlgorithmsTest.ConvertSpectrumAxisTestPerformance.test_exec_performance.htm

http://builds.mantidproject.org/job/master_performancetests2/Master_branch_performance_tests/CrystalTest.FilterPeaksTestPerformance.testPerformance.htm

DavidFair:
Description of work.
This PR removes some duplicate code which entered SaveGSSTest and restores the performance difficulty to be the original number of bins and iterations. 

The original seen [here](https://github.com/mantidproject/mantid/pull/18633/files#diff-79012422933dee560c132ece4d5e0014R366) had 100,000 bins to save this was memory intense so was changed the save n number of times [here](https://github.com/mantidproject/mantid/pull/18660/files#diff-79012422933dee560c132ece4d5e0014R401) but we unintentionally changed the total number of bins saved. 

In this PR we take both and save 100,000 bins (5 * 20,000) as per the original PR but using far less memory. 

**To test:**
Code review - Does the number of bins match the original PR (first link)
Run the performance test - does it complete in the same amount of time

Fixes #19267 .

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/DataHandling/test/SaveGSSTest.h
```diff
@@ -398,7 +370,10 @@ class SaveGSSTestPerformance : public CxxTest::TestSuite {
 private:
   std::vector<SaveGSS *> saveAlgPtrs;
 
-  const int numberOfIterations = 5;
+  // Controls the speed of the test
+  const int m_numberOfBinsToSave = 20000;
+  // Multiplied by n times specified below
+  const int m_numberOfIterations = 5;
```
LamarMoore: @DavidFair the major issue with the performance test was this iterative run of the algorithm. When you initially wrote the unit test, there were no iterations. It would be better to get rid of this and just run the algorithm once with a workspace with 100000 bins as per the original implementation.

LamarMoore:
This is better. The performance test should not be returning sensible metrics for comparison with previous runs. :shipit:

PR #19279, opened 2017-03-31
ComponentInfo positions and rotations

### LINKED ISSUE #19278: ComponentInfo with readable positions and roatations
Follow on from #19015. Provides positions and rotations built from Instrument 1.0.

OwenArnold:
Fixes #19278. Puts positions and rotations into the `Beamline::ComponentInfo`. detector indices forwarded to the `Beamline::DetectorInfo`. This PR will not involve rolling out the changes or providing write access.

**WIP NOT READY FOR MERGING** This is currently blocked by #19468 which (along with some recent changes to master) contain some of the caching facilities needed to complete this PR.

**To test:**

* Code review required.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/API/inc/MantidAPI/ComponentInfo.h
```diff
@@ -58,6 +59,8 @@ class MANTID_API_DLL ComponentInfo {
                 std::vector<Mantid::Geometry::IComponent *> &&componentIds);
   std::vector<size_t> detectorIndices(size_t componentIndex) const;
   size_t size() const;
+  Eigen::Vector3d position(const size_t componentIndex) const;
+  Eigen::Quaterniond rotation(const size_t componentIndex) const;
```
SimonHeybrock: Can we use `Eigen` here or do we need to convert to the types in `Kernel`, as in `DetectorInfo?

Framework/API/inc/MantidAPI/InfoComponentVisitor.h
```diff
@@ -64,9 +71,19 @@ class MANTID_API_DLL InfoComponentVisitor
   /// Only Assemblies and other NON-detectors yield ranges
   std::vector<std::pair<size_t, size_t>> m_ranges;
 
+  /// Positions
+  boost::shared_ptr<std::vector<Eigen::Vector3d>> m_positions;
+
+  /// Rotations
+  boost::shared_ptr<std::vector<Eigen::Quaterniond>> m_rotations;
```
SimonHeybrock: Why `shared_ptr`?

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -69,7 +69,7 @@ ExperimentInfo::ExperimentInfo()
       m_run(new Run()), m_parmap(new ParameterMap()),
       sptr_instrument(new Instrument()),
       m_detectorInfo(boost::make_shared<Beamline::DetectorInfo>()),
-      m_componentInfo(boost::make_shared<Beamline::ComponentInfo>()) {
+      m_componentInfo(boost::shared_ptr<Beamline::ComponentInfo>(nullptr)) {
```
SimonHeybrock: Why this change? What if no instrument is set?

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -187,23 +187,27 @@ void checkDetectorInfoSize(const Instrument &instr,
                              "instrument");
 }
 
-boost::shared_ptr<Beamline::ComponentInfo>
-makeComponentInfo(const Instrument &oldInstr,
-                  const API::DetectorInfo &detectorInfo,
-                  std::vector<Geometry::ComponentID> &componentIds) {
+boost::shared_ptr<Beamline::ComponentInfo> makeComponentInfo(
```
SimonHeybrock: Personally, I use `unique_ptr` wherever possible (such as in factory methods).

Framework/API/src/InfoComponentVisitor.cpp
```diff
@@ -36,11 +58,14 @@ void InfoComponentVisitor::registerComponentAssembly(
...
   const size_t detectorStop = m_assemblySortedDetectorIndices.size();
-
   m_ranges.emplace_back(std::make_pair(detectorStart, detectorStop));
 
   // For any non-detector we extend the m_componetIds from the back
   m_componentIds.emplace_back(assembly.getComponentID());
+
+  m_positions->emplace_back(Kernel::toVector3d(assembly.getPos()));
+  m_rotations->emplace_back(Kernel::toQuaterniond(assembly.getRotation()));
+  clearPositionAndRotationsParameters(m_pmap, assembly);
```
SimonHeybrock: This clears positions and rotations immediately. Thus, once you visit the children of the assembly, you will not see their original position and rotation anymore, right? Or am I missing something that takes that into account?

SimonHeybrock: Do we have a test for this? Move a parent component and then check if children are still correctly parsed.

Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
```diff
@@ -40,14 +44,22 @@ class MANTID_BEAMLINE_DLL ComponentInfo {
...
+  boost::shared_ptr<DetectorInfo> m_detectorInfo;
+  bool isDetectorDomain(const size_t componentIndex) const;
 
 public:
-  ComponentInfo();
   ComponentInfo(const std::vector<size_t> &assemblySortedDetectorIndices,
-                const std::vector<std::pair<size_t, size_t>> &ranges);
+                const std::vector<std::pair<size_t, size_t>> &ranges,
+                boost::shared_ptr<std::vector<Eigen::Vector3d>> positions,
+                boost::shared_ptr<std::vector<Eigen::Quaterniond>> rotations,
```
SimonHeybrock: Why `shared_ptr`?

SimonHeybrock (commented):
@OwenArnold Some comments below.

There are also several other missing things:
- Changes to `Component::getPos()` (and similar) that forward to getting the position from the `ComponentInfo`.
- Changes to the two methods in `ExperimentInfo` that deal with parsing IDFs and parameter strings and intercept changes to positions/rotations.
- Changes to `Instrument:saveNexus()` the reinsert positions/rotations into the parameter map (extend `Instrument::createLegacyParameterMap`).

Framework/API/inc/MantidAPI/ComponentInfo.h
OwenArnold: You're right. we should use `Kernel::V3D` etc for consistency.

Framework/API/inc/MantidAPI/InfoComponentVisitor.h
OwenArnold: Ultimately we want `ComponentInfo` to store positions/rotations as `cow_ptr`. Creating as `shared_ptr` allow us to pass the same objects over from the `InfoComponentVisitor` to the `ComponentInfo`. When the temporary `InfoComponentVisitor` goes out of scope, the `ComponentInfo` becomes the sole owner (at least until copies etc are made).

Framework/API/src/ExperimentInfo.cpp
OwenArnold: I was trying to avoid a default constructor for a `ComponentInfo`. But you're probably right here. I hadn't considered it if `setInstrument` was not called.

Framework/API/src/ExperimentInfo.cpp
OwenArnold: I did this to keep it consistent to the storage policy for `Beamline::DetectorInfo`. presumably this was driven by the need to facilitate copies.

Framework/API/src/InfoComponentVisitor.cpp
OwenArnold: I'm not 100% sure what your concern is here. But the parameter map will be cleared depth first. So any parent information will always still be in the parameter map at the point that child (derived) information is fetched. 

Probably having a test for this is still a good idea.

Framework/API/src/ExperimentInfo.cpp
SimonHeybrock: Yeah, but I still create `unique_ptr` and then dump it into `shared_ptr`. `DetectorInfo` is currently held as a `shared_ptr` only for legacy workarounds, namely for storing it in `Instrument` in `ExperimentInfo::getInstrument`. Eventually it will not be shared, since it does all sharing internally.

Framework/API/src/InfoComponentVisitor.cpp
SimonHeybrock: Hmm, sounds ok, as long as you ensure that `makeDetectorInfo` is called *before* you visit.

OwenArnold: Yes. I'll add a test for this in any case. 

As a side point, you'll notice that in visiting detectors, no positions/rotation are stored. One improvement that I'd like to make (in a future PR) is to have the visitor create the `DetectorInfo` as well as the `ComponentInfo`. There are several potential benefits to that.

1. It would consolidate the place that all relevant `IComponent` subtypes are iterated through and used to create info layers. 
2. We visit the `IDetectors` already anyway, so the `DetectorInfo` generation should slot in nicely.
3. `ComponentInfo` depends on `DetectorInfo` now, but I have to expose lots of the getters of the `InfoComponentVisitor` just so that I can bind the two together. Doing this internally would simplify that API quite a lot.

SimonHeybrock: Yes, I saw that :+1:

OwenArnold:
@SimonHeybrock I'm unsure about how to proceed with this without making some major changes around the way that `DetectorInfo` is used. I think it would be best to skype about this. But some of the points needing consideration are:

* I've now made it possible for `Component` to forward to `ComponentInfo` (example below not committed). We would want to delete `m_pos`/`m_rot` from `Component` but I think this could be too much of a big jump. Not clear the easiest way to proceed from here. Ideas welcome.
* With the changes you have introduced, `MoveInstrumentComponent`, `RotateInstrumentComponent` hook directly into `DetectorInfo` now, which in turn calls `ComponentHelper` for non-detectors and adds items to the `ParameterMap`. Clearly now `DetectorInfo` is performing functions that `ComponentInfo` should be providing. I think this requires some thought.

```
void Component::setPos(double x, double y, double z) {
    if(this->m_map->hasComponentInfo()){
        const auto& componentInfo = m_map->componentInfo();
        size_t currentIndex = m_map->componentIndex(this->getComponentID());
        componentInfo.setPosition(currentIndex, Eigen::Vector3d{x, y, z});
    } else {
        throw std::runtime_error("No ComponentInfo onto which to set the position");
    }
}
```

Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
```diff
@@ -40,14 +44,26 @@ class MANTID_BEAMLINE_DLL ComponentInfo {
 private:
   boost::shared_ptr<std::vector<size_t>> m_assemblySortedDetectorIndices;
   boost::shared_ptr<const std::vector<std::pair<size_t, size_t>>> m_ranges;
+  Mantid::Kernel::cow_ptr<std::vector<Eigen::Vector3d>> m_positions;
+  Mantid::Kernel::cow_ptr<std::vector<Eigen::Quaterniond>> m_rotations;
   const size_t m_size = 0;
+  boost::shared_ptr<DetectorInfo> m_detectorInfo;
+  bool isDetectorDomain(const size_t componentIndex) const;
```
SimonHeybrock: Maybe not relevant immediately, but eventually this could be made public, and named `isDetector`?

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -41,5 +61,46 @@ ComponentInfo::detectorIndices(const size_t componentIndex) const {
...
+}
+
+void ComponentInfo::setPosition(const size_t componentIndex,
+                                const Eigen::Vector3d &position) {
+  if (isDetectorDomain(componentIndex)) {
+    m_detectorInfo->setPosition(componentIndex, position);
+  }
+  const auto rangesIndex =
+      componentIndex - m_assemblySortedDetectorIndices->size();
+  m_positions.access()[rangesIndex] = position;
```
SimonHeybrock: This should update the positions of all components in the subtree, I think.

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -41,5 +61,46 @@ ComponentInfo::detectorIndices(const size_t componentIndex) const {
...
+
+void ComponentInfo::setRotation(const size_t componentIndex,
+                                const Eigen::Quaterniond &rotation) {
+
+  if (isDetectorDomain(componentIndex)) {
+    m_detectorInfo->setRotation(componentIndex, rotation);
+  }
+  const auto rangesIndex =
+      componentIndex - m_assemblySortedDetectorIndices->size();
+  m_rotations.access()[rangesIndex] = rotation; // TODO adjust position too.
```
SimonHeybrock: This should update the positions and rotations of all components in the subtree, I think.

Framework/Beamline/test/ComponentInfoTest.h
```diff
@@ -14,18 +16,192 @@ class ComponentInfoTest : public CxxTest::TestSuite {
...
+        Eigen::AngleAxisd(M_PI / 3, Eigen::Vector3d::UnitZ()));
+    auto detectorInfo =
+        boost::make_shared<DetectorInfo>(detPositions, detRotations);
+    std::vector<size_t> bankSortedDetectorIndices{0, 2, 1};
+
+    std::vector<std::pair<size_t, size_t>> ranges;
+    ranges.push_back(std::make_pair(0, 3));
+    ranges.push_back(std::make_pair(0, 2));
+
+    auto compPositions = boost::make_shared<std::vector<Eigen::Vector3d>>();
```
SimonHeybrock: By having `shared_ptr` for the constructor arguments in `ComponentInfo` you get a lot of cumbersome client code. Consider passing a vector (by value or `&&`) and change the visitor getters to move its members when accessed for construction of `ComponentInfo` (i.e., you can get the positions from the visitor only once). Maybe having a one-shot factory method on this visitor to create the `ComponentInfo`?

Framework/Geometry/src/Instrument.cpp
```diff
@@ -1233,19 +1239,87 @@ const Beamline::DetectorInfo &Instrument::detectorInfo() const {
...
 }
 
+/**
+ * @brief Instrument::componentIndex
+ * Note that this index can be used with ComponentInfo. Throw std::runtime_error
+ * if the ComponentID does not exist.
+ * @param componentId : ComponentID to find the index for.
+ * @return the Component index associated with this Component.
+ */
+size_t Instrument::componentIndex(const ComponentID componentId) const {
```
SimonHeybrock: Hmmm, the update mechanism looks broken, e.g., when adding components to a child component (also there are probably thread-safety issues). Can we simply forward to `ComponentInfo::indexOf()`?

SimonHeybrock (commented):
@OwenArnold Yes, we can discuss this later on skype. Regarding your questions:

- I think we are far from the point where `m_pos`/`m_rot` can be removed from `Component`. We still cannot do anything without having a stand-alone legacy instrument.
- The functionality in `DetectorInfo` you mention is already marked as to-be-replaced by `ComponentInfo`. You simply need to do recursive updates (see also my comments below).

Framework/API/src/ComponentInfo.cpp
```diff
@@ -39,8 +42,30 @@ ComponentInfo::detectorIndices(size_t componentIndex) const {
   return m_componentInfo.detectorIndices(componentIndex);
 }
 
+std::vector<Geometry::IComponent *> ComponentInfo::componentIds() const {
```
OwenArnold: Agreed

Framework/API/src/ComponentInfo.cpp
```diff
@@ -12,17 +13,19 @@ namespace API {
...
+ * @brief ComponentInfo::ComponentInfo
+ * @param componentInfo
+ * @param componentIds : ComponentIDs ordered by component
+ * index
  */
-ComponentInfo::ComponentInfo(const Beamline::ComponentInfo &componentInfo,
-                             std::vector<Geometry::ComponentID> &&componentIds)
+ComponentInfo::ComponentInfo(
+    Mantid::Beamline::ComponentInfo &componentInfo,
+    std::vector<Mantid::Geometry::IComponent *> componentIds)
```
OwenArnold: Do you mean ref to `Beamline::ComponentInfo`? Now needs to be non-const so that `setPosition`, `setRotation` can be propagated

Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
OwenArnold: Agreed

Framework/Beamline/src/ComponentInfo.cpp
OwenArnold: 👍 it should

Framework/Beamline/src/ComponentInfo.cpp
OwenArnold: Yes

Framework/API/src/ComponentInfo.cpp
SimonHeybrock: No, I mean `componentIds`.

Framework/Geometry/src/Instrument.cpp
OwenArnold: Yes, there are issues. I think I outlined some of these in detail in one of the commit messages. The crux of the issue as I see it is that Instrument1.0 is never "finalised" and mixes up construction with behaviour. so `Instrument::add` can occur at any time and that forces us to also rebuild or ComponentID->componentIndex caches too.

I don't see how `ComponentInfo::indexOf` would fix the problem. For one thing, you need to have the `Instrument` associated with an `ExperimentInfo` to get that information set properly.

Framework/Beamline/test/ComponentInfoTest.h
OwenArnold: Good point. I'll take a look at this.

Framework/Geometry/src/Instrument.cpp
SimonHeybrock: But `componentIndex()` is only called to access `ComponentInfo`, so if there is no `ComponentInfo`, `componentIndex()` will never get called? Or am I missing something here?

NickDraper:
I'm going to assume this is not really intended for R3.10 and move the milestone.

Please tell me if I'm wrong.

SimonHeybrock:
@OwenArnold Should we close this, I think it has been replaced by another PR, right?

PR #14880, opened 2016-01-04
MPI support with splitting based on spectra IDs

SimonHeybrock:
This pull request is meant for discussing design and implementation of the new MPI model.

PR #19280, opened 2017-03-31
Add mechanism for MPI support to Workspace and Algorithm

### LINKED ISSUE #19246: Add mechanism for MPI support to workspaces and algorithms.
Revive basic features from old MPI proof-of-concept PR: https://github.com/mantidproject/mantid/pull/14880.

Basically we want to add `StorageMode` and `ExecutionMode` here.

SimonHeybrock:
This implements a cleaned-up version of the MPI mechanism that has a proof-of-concept implementation in https://github.com/mantidproject/mantid/pull/14880.

First examples and detailed documentation are part of the follup-up PR https://github.com/mantidproject/mantid/pull/19510. Having a look at the documentation added there may be helpful in reviewing this: https://github.com/mantidproject/mantid/blob/02a5e4852fd0abb60cc2f95e7abba722dbd16eaa/docs/source/development/ImplementingMPISupport.rst.

An overview of the concept can also be found in the slides from the 2016 developer meeting: https://github.com/mantidproject/documents/blob/master/Presentations/DevMeetings/2016-01/mantid-mpi-new.pdf.

**To test:**

Code review.

Fixes #19246.

Internal change, no release notes since we do not yet have any algorithms that could be used with MPI.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/API/inc/MantidAPI/Algorithm.h
```diff
@@ -304,12 +307,26 @@ class MANTID_API_DLL Algorithm : public IAlgorithm,
...
 
+  void exec(Parallel::ExecutionMode executionMode);
+  virtual void execDistributed();
+  virtual void execMasterOnly();
+  virtual void execNonMaster();
+
+  virtual Parallel::ExecutionMode getParallelExecutionMode(
+      const std::map<std::string, Parallel::StorageMode> &storageModes) const;
+
+  Parallel::ExecutionMode
```
OwenArnold: This is deliberately not virtual?

SimonHeybrock: Yes, I think it should just provide a generic mapping. Might actually make more sense as a non-member.

Framework/API/inc/MantidAPI/IMDHistoWorkspace.h
```diff
@@ -44,6 +44,10 @@ class DLLExport IMDHistoWorkspace : public IMDWorkspace,
   IMDHistoWorkspace_uptr clone() const {
     return IMDHistoWorkspace_uptr(doClone());
```
OwenArnold: I think this was discussed on a review a long time ago. Minor point - since the covariance is unnecessary. Maybe just `make_unique` inplace and drop the `doClone` method?

SimonHeybrock: Why is the covariance unnecessary?

OwenArnold: At the public interface you are creating a `IMDHistoWorkspace_uptr`. You could just make this method virtual and do away with the `doClone`. As I say this is a minor point.

Framework/API/inc/MantidAPI/Algorithm.h
OwenArnold: Non member sounds good. I wondered if there would be any difference for MDAlgorithms, but that's a future issue to worry about.

Framework/API/inc/MantidAPI/IMDHistoWorkspace.h
SimonHeybrock: But then I cannot have covariance?

OwenArnold: I mean covariance in you actual return type. Subtypes of `IMDHistoWorkspace` must implement a clone that returns a `IMDHistoWorkspace_uptr`. Return a uptr is the right thing to do. I suggest leaving things as they are to avoid unnecessary work.

SimonHeybrock: No, we *need* covariance, i.e., subtypes of `IMDHistoworkspace` must return a `unique_ptr` to their own type, not to the base class type.

Framework/API/inc/MantidAPI/WorkspaceGroup.h
```diff
@@ -124,6 +124,9 @@ class MANTID_API_DLL WorkspaceGroup : public Workspace {
   WorkspaceGroup *doClone() const override {
     throw std::runtime_error("Cloning of WorkspaceGroup is not implemented.");
   }
+  WorkspaceGroup *doCloneEmpty() const override {
+    throw std::runtime_error("Cloning of WorkspaceGroup is not implemented.");
```
OwenArnold: Is this disabled, for any particular reason?

Framework/API/inc/MantidAPI/IMDHistoWorkspace.h
SimonHeybrock: Example:
If you have a `Workspace2D`, and you call `ws->clone()`, you want to get back a `Workspace2D`. Without covariance you would get a `Workspace`.

Framework/API/inc/MantidAPI/WorkspaceGroup.h
OwenArnold: `WorkspaceGroup` should just be a container around a `std::vector<Workspace_sptr>`

Framework/API/inc/MantidAPI/IMDHistoWorkspace.h
OwenArnold: But you would get that with `make_unique`.

Framework/API/inc/MantidAPI/WorkspaceGroup.h
SimonHeybrock: I do not know why it was not implemented.

Framework/API/inc/MantidAPI/IMDHistoWorkspace.h
SimonHeybrock: http://stackoverflow.com/questions/6924754/return-type-covariance-with-smart-pointers

OwenArnold: Ok. Sorry my misunderstanding of what you were trying to do here.

Framework/API/src/Algorithm.cpp
```diff
@@ -545,7 +551,7 @@ bool Algorithm::execute() {
       // Start a timer
       Timer timer;
       // Call the concrete algorithm's exec method
-      this->exec();
+      this->exec(executionMode);
```
OwenArnold: I don't think there's any need for the named variable here.

Framework/API/test/ScopedWorkspaceTest.h
```diff
@@ -22,6 +22,9 @@ class MockWorkspace : public Workspace {
   MockWorkspace *doClone() const override {
```
OwenArnold: This type should be called `FakeWorkspace` not `MockWorkspace`, but you only extended this type, so I don't suggest fixing that.

Framework/Parallel/inc/MantidParallel/StorageMode.h
```diff
@@ -0,0 +1,51 @@
...
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+enum class StorageMode { Cloned, Distributed, MasterOnly };
+
+MANTID_PARALLEL_DLL std::string toString(StorageMode mode);
+MANTID_PARALLEL_DLL std::string
+toString(const std::map<std::string, StorageMode> &map);
```
OwenArnold: Good addition of utility free function.

Framework/Parallel/test/StorageModeTest.h
```diff
@@ -0,0 +1,35 @@
...
+  static void destroySuite(StorageModeTest *suite) { delete suite; }
+
+  void test_toString() {
+    TS_ASSERT_EQUALS(toString(StorageMode::Cloned),
+                     "Parallel::StorageMode::Cloned");
+    TS_ASSERT_EQUALS(toString(StorageMode::Distributed),
+                     "Parallel::StorageMode::Distributed");
+    TS_ASSERT_EQUALS(toString(StorageMode::MasterOnly),
+                     "Parallel::StorageMode::MasterOnly");
+  }
```
OwenArnold: Good. Would have been easy to forget this.

OwenArnold:
@SimonHeybrock Looks good.

Framework/API/src/Algorithm.cpp
SimonHeybrock: Will have to think about that one, since `getExecutionMode()` can throw.

SimonHeybrock:
@OwenArnold Regarding `TriviallyParallelAlgorithm`: I will add those bits in follow-up PRs, once we actually support MPI in actual algorithms.

OwenArnold:
@SimonHeybrock That sounds sensible. This PR is large enough.

Framework/API/src/Algorithm.cpp
```diff
@@ -1616,6 +1620,153 @@ void Algorithm::setAlgStartupLogging(const bool enabled) {
...
+
+/** By default execMasterOnly() runs this in `master-only` execution mode on all
+ * but rank 0.
+ *
+ * The default implementation creates dummy workspaces for all pure output
+ * workspaces. Classes inheriting from Algorithm can re-implement this if they
+ * support execution with multiple MPI ranks and require a special behavior on
+ * non-master ranks in master-only execution. */
+void Algorithm::execNonMaster() {
+  if (m_inputWorkspaceProps.size() == 1 &&
```
OwenArnold: The logic here is harder to follow than it might be. Could this be written in a way to make it more readable. Maybe splitting some of the checks out into other functions.

Framework/API/src/Algorithm.cpp
```diff
@@ -1616,6 +1620,153 @@ void Algorithm::setAlgStartupLogging(const bool enabled) {
...
+        const auto &wsProp = m_pureOutputWorkspaceProps.front();
+        // This is the reverse cast of what is done in
+        // cacheWorkspaceProperties(), so it should never fail.
+        const Property &prop = dynamic_cast<Property &>(*wsProp);
+        setProperty(prop.name(), ws->cloneEmpty());
+        return;
+      }
+    }
+  }
+  throw std::runtime_error(
```
OwenArnold: Is this level of diagnostics enough? There are four conditions in this function that must be failed to be met before getting to this.

Framework/API/src/Algorithm.cpp
OwenArnold: Might also be useful to help devs overriding this to reuse bits that still make sense.

Framework/API/src/Algorithm.cpp
SimonHeybrock: Maybe the opposite: The message should maybe simply say "The algorithm does not support ExecutionMode::MasterOnly".

My point here is that there is nothing about the algorithm inputs the user can do to get around this issue. There is no other way but write code that supports this execution mode.

Framework/API/src/Algorithm.cpp
```diff
@@ -1616,6 +1620,153 @@ void Algorithm::setAlgStartupLogging(const bool enabled) {
...
+        wsProp->setIsMasterRank(false);
+}
+
+/// Returns a const reference to the (MPI) communicator of the algorithm.
+const Parallel::Communicator &Algorithm::communicator() const {
+  return *m_communicator;
+}
+
+/// Sets the (MPI) communicator of the algorithm.
+void Algorithm::setCommunicator(const Parallel::Communicator &communicator) {
```
OwenArnold: This is currently never used. We currently just use the default constructed `Communicator` on the `Algorithm`?

SimonHeybrock: It is used a lot in unit tests, I think.

SimonHeybrock: Later, the idea is that the `AlgorithmManager` can set this, in case we want to run algorithms only in a sub-communicator.

Framework/API/src/Algorithm.cpp
SimonHeybrock: I did not really see better way to do this. Not sure where such bits could be reused, if this ever is the case we can refactor. For now I just added some code comments. I hope this clarifies things?

OwenArnold:
@SimonHeybrock. Looks good to me.

OwenArnold (approved):
Several rounds of improvements. Looks good to me.

martyngigg:
@SimonHeybrock There's a conflict in one of the cmake files after a recent merge.

SimonHeybrock:
@martyngigg Thanks, should be resolved.

Framework/API/src/Algorithm.cpp
```diff
@@ -1616,6 +1622,168 @@ void Algorithm::setAlgStartupLogging(const bool enabled) {
...
+    return exec();
+  case Parallel::ExecutionMode::Distributed:
+    return execDistributed();
+  case Parallel::ExecutionMode::MasterOnly:
+    return execMasterOnly();
+  default:
+    throw(std::runtime_error("Algorithm " + name() +
+                             " does not support execution mode " +
+                             Parallel::toString(executionMode)));
+  }
```
martyngigg: Out of curiosity, what is the use case for the `Parallel::ExecutionMode::Identical` mode?

SimonHeybrock: @martyngigg If the storage mode of the workspace is `StorageMode::Cloned` the corresponding execution mode would usually be `ExecutionMode::Identical`. You could for example imagine loading a monitor spectrum into a cloned workspace and then use it normalize a distributed workspace (I am not certain we will actually handle monitors in this case, but there are also other use cases, I think).

martyngigg: @SimonHeybrock I see. Thanks for the clarification.

PR #19301, opened 2017-04-04
Implement slicing by log value

### LINKED ISSUE #19012: Implement slicing by log value
Design sent by Max:

eventhandling

I believe we could use `FilterByLogValue`, for which we would have to expose properties `MinimumValue` and `MaximumValue`. The first step before implementing this in the interface is to understand how the algorithm works (for instance, what happens if two non-consecutive blocks of data are in the allowed range of log value?). The second step is to write a script similar to the script originally sent by Max (can be found [here](https://github.com/mantidproject/mantid/files/639344/EventRef.zip)), discuss it with scientists and implement the functionality in the interface once they are happy with it.

PranavBahuguna:
This PR adds a fourth slicing option to the event handling tab in the ISIS Reflectometry interface - log value slicing. This new slicing option allows the user to provide a selection of values (in a manner identical to custom slicing) and the name of a log. This allows a run to be filtered by the log values for each start/stop value, creating several slices.

~~NOTE: The `FilterByLogValue` algorithm currently has the property `TimeTolerance` set to `1.0` to avoid collecting no data when only one log is found. This is temporary however and we should talk to the instrument scientists about what they wish `TimeTolerance` to actually be set to.~~ As scientists are still undecided on its value, it will be left at a value of 1.0 for now.

**To test:**

This should be tested on `Release` mode. Alternatively if using `Debug` mode, go to `LoadEventNexus.cpp` and replace the following lines at lines 289-292 with this:

```
if (temp < 0) {
    if (temp < m_min_id)
        m_min_id = temp;
    if (temp > m_max_id)
        m_max_id = temp;
}
```

- Code review
- Access the interface from Interfaces->Reflectometry->ISIS Reflectometry
- Enter run `38415` with transmission run `38393` in the `Runs` tab
- In the `Event Handling` tab, ensure that omitting either slicing values or log name and processing causes the run to be processed as a non-event workspace (it will be processed via `GenericDataProcessorPresenter.cpp`).
- Enter values `0,100,1000,2000 and log name `raw_frames` and process. This should produce 3 sliced workspaces of each kind. As there are no positive-time `raw_frames` logs with values less than 100, plotting the first slice for IvsQ workspaces should show no data while the others will. Similarly, the first TOF slice should contain no data.
- You can check the log values for `raw_frames` by right clicking on the original TOF workspace, opening 'sample logs' and opening the logs for `raw_frames`.

Fixes #19012.

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


PranavBahuguna:
@AntonPiccardoSelg Merge conflicts are resolved and this issue is no longer in progress, can you review this please?

MantidQt/CustomInterfaces/src/Reflectometry/ReflDataProcessorPresenter.cpp
```diff
@@ -445,42 +469,54 @@ std::string ReflDataProcessorPresenter::loadRun(const std::string &run,
...
+* @param stopTime :: Stop time
+* @param logFilter :: The log filter to use if slicing by log value
 * @return :: the name of the sliced workspace (without prefix 'TOF_')
 */
 std::string ReflDataProcessorPresenter::takeSlice(const std::string &runNo,
                                                   size_t sliceIndex,
                                                   double startTime,
-                                                  double stopTime) {
+                                                  double stopTime,
+                                                  std::string logFilter) {
```
AntonPiccardoSelg: `const &`?

AntonPiccardoSelg:
@PranavBahuguna This is stuck at `ClangFormat`

AntonPiccardoSelg:
Works as described. Please :shipit:

quantumsteve (changes requested):
A few things which I think should be considered before merging

MantidQt/CustomInterfaces/src/Reflectometry/QtReflEventView.cpp
```diff
@@ -80,13 +80,8 @@ std::string QtReflEventView::getTimeSlicingValues() const {
...
 */
 std::string QtReflEventView::getTimeSlicingType() const {
-  std::map<SliceType, std::string> sliceTypeStrMap = {
-      {Uniform, "Uniform"},
-      {UniformEven, "UniformEven"},
-      {Custom, "Custom"},
-      {LogValue, "LogValue"}};
 
-  return sliceTypeStrMap[m_sliceType];
+  return m_sliceTypeStrMap.at(m_sliceType);
```
quantumsteve: Why did you switch to `std::map::at`? It's a relatively expensive check that in this case serves no purpose.

PranavBahuguna: As m_sliceTypeStrMap is now defined in the header file, it is essentially `const` after initialisation. We cannot use `[]` as that can also be used to insert new elements into the map.

I feel it is also worth mentioning that the code in question is only called once in reduction and is not performance-critical at all.

quantumsteve (approved):

PR #19308, opened 2017-04-05
Fix ComponentInfo performance Problems 

### LINKED ISSUE #19307: Fix ComponentInfo performance Problems
Changes introduced as part of issue #19015 PR #19153 have caused a dramatic slow down to the performance tests. Particularly for `DiffractionFocussing2TestPerformance` which currently times out due to the extent of the slow down.

OwenArnold:
Use a cache and reuse approach similar to that employed for `DetectorInfo`. So we do not create the `ComponentInfo` parsing the whole instrument tree unless we really need to.

The approach here is to reuse the `ComponentInfo` and `DetectorInfo` on the incoming instrument if it has already been provisioned with them. 

**To test:**

* Code review.
* We expect that performance tests, particularly `DiffractionFocussing2TestPerformance` execute in a similar time to before. Though there will be additional overhead on the first instrument creation when loaded.

Fixes #19307 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Geometry/inc/MantidGeometry/Instrument/ParameterMap.h
```diff
@@ -381,8 +388,13 @@ class MANTID_GEOMETRY_DLL ParameterMap {
   /// Pointer to the DetectorInfo object. NULL unless the instrument is
   /// associated with an ExperimentInfo object.
   boost::shared_ptr<const Beamline::DetectorInfo> m_detectorInfo{nullptr};
+  /// Pointer to the ComponentInfo object. NULL unless the instrument is
+  /// associated with an ExperimentInfo object.
+  boost::shared_ptr<const Beamline::ComponentInfo> m_componentInfo{nullptr};
+  /// Component ids.
+  std::vector<Geometry::ComponentID> m_componentIds;
```
OwenArnold: I think that the storage mode here would be something to review as part of another issue. But those internals can be dealt with another time provided they are not causing big performance issues.

OwenArnold:
Release Build OSX. Using a 100*100 detector rectangular bank instrument (`ExperimentInfoTestPerformance`). Here are the timings in microseconds for the relevant parts of `ExperimentInfo::setInstrument`

```
Time to construct Beamline DetectorInfo: 789233
Time to construct API Detectorinfo: 139414
Time to info visit instrument: 309634
Time to construct Beamline ComponentInfo: 5480
```
And if the instrument is provisioned....
```
Time to copy construct Beamline DetectorInfo: 2
Time to construct API Detectorinfo: 161801
Time to copy construct Beamline ComponentInfo: 1
```

Framework/Geometry/inc/MantidGeometry/Instrument/ParameterMap.h
```diff
@@ -346,9 +347,15 @@ class MANTID_GEOMETRY_DLL ParameterMap {
 
   bool hasDetectorInfo() const;
   const Beamline::DetectorInfo &detectorInfo() const;
+  bool hasComponentInfo() const;
+  const Beamline::ComponentInfo &componentInfo() const;
   size_t detectorIndex(const detid_t detID) const;
+  const std::vector<Geometry::ComponentID> &componentIds() const;
   void
   setDetectorInfo(boost::shared_ptr<const Beamline::DetectorInfo> detectorInfo);
+  void setComponentInfo(
```
martyngigg: Is it intentional to take `componentIds` by value?

martyngigg (commented):
A few questions...

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -110,6 +111,7 @@ void ExperimentInfo::copyExperimentInfoFrom(const ExperimentInfo *other) {
   //   not here since we just copy the experiment data).
   // - For cached groupings (for MDWorkspaces), grouping was not copied in the
   //   old implementation either.
+  *m_componentInfo = *other->m_componentInfo;
```
OwenArnold: This was a bug I noticed.

martyngigg: :+1:

Framework/API/src/ComponentInfo.cpp
```diff
@@ -13,16 +13,39 @@ namespace API {
...
+ * Constructor
+ * @brief ComponentInfo::ComponentInfo
+ * @param componentInfo
+ * @param componentIds : r-value reference to ComponentIDs ordered by component
+ * index
+ */
+ComponentInfo::ComponentInfo(
+    const Mantid::Beamline::ComponentInfo &componentInfo,
+    std::vector<Mantid::Geometry::IComponent *> &&componentIds)
+    : m_componentInfo(componentInfo),
```
OwenArnold: I'm not sure that that would work with the reference collapsing rules, or without adding a universal reference. I've added a `std::move`.  Let me know if you have a better idea.

Framework/Geometry/inc/MantidGeometry/Instrument/ParameterMap.h
OwenArnold: It's a good point. As I've commented already, this is something I'd like to revisit at a later point along side some other ownership issues. I doubt this will have much impact on the performance at present, so I would suggest leaving it the way that it is.

martyngigg: :+1:

Framework/API/src/ComponentInfo.cpp
martyngigg: Inside the r-value reference constructor `componentIds` is now an lvalue which should be able to bind to a non-const lvalue reference I think?

This seems to work, for example: 

```
struct Foo {
  Foo(std::vector<int> & ids) : cids(ids) {
     ids = {1,2,3,4,5};
  }
  Foo(std::vector<int> && ids) : Foo(ids) {} 
  std::vector<int> cids;
};

int main() {
    std::vector<int> my_ids;
    {
     Foo f(my_ids);
    }
    std::copy(std::begin(my_ids), std::end(my_ids),
              std::ostream_iterator<int>(std::cout, ","));   
}
```

Demo on [wandbox](https://wandbox.org/permlink/GMO6FlH7vVBpM19R).

OwenArnold: @martyngigg I'm not convinced the above is doing what you think. My demo on [wandbox](https://wandbox.org/permlink/72jhJCeYgIxzRcVV)
A slight variant of yours...
```
struct Arg {

    Arg(){std::cout << "default constructor called" << std::endl;};
    Arg(const Arg& ){std::cout << "copy constructor called" << std::endl;}
    Arg(Arg&& ){std::cout << "move constructor called" << std::endl;}
};

struct Foo {
  Foo(const Arg & arg) : m_arg(arg) {
     
  }
  Foo(Arg && arg) : Foo(arg) 
  {} 
    
  /*
  // Uncomment. This avoids the copy!
  Foo(Arg && arg) : m_arg(std::move(arg)) //Foo(arg) 
  {} 
  */

  Arg m_arg;
};

int main() {
    
    Arg arg;
    Foo f1{arg};
    
    Foo f2{Arg{}};
}
```

OwenArnold: Arg&& & becomes Arg&

martyngigg:
@OwenArnold I think is good to go once we fix the formatting.

martyngigg:
:shipit:

LGTM when Jenkins gives its okay.

Framework/API/src/ComponentInfo.cpp
SimonHeybrock: Why not just pass by value and move, instead of having two constructor + `init()`?

Framework/API/src/ExperimentInfo.cpp
SimonHeybrock: I think actually this could be removed, as well as the line `*m_detectorInfo = *other->m_detectorInfo;` above. `setInstrument()` already handles this, I think?

Framework/API/src/ComponentInfo.cpp
OwenArnold: 👍

Framework/API/src/ExperimentInfo.cpp
OwenArnold: Ok I'll take a look

SimonHeybrock:
RHEL7 build took a bit long, we should jeep an eye on this. Merging for now.

OwenArnold:
@SimonHeybrock Yes. Something is happening in `VesuvioCommandsTest.BankByBankForwardSpectraNoBackground`. I'm investigating that.

PR #19347, opened 2017-04-11
FindEPP rewrite in c++

### LINKED ISSUE #19133: Parallelize FindEPP?
[`FindEPP`](http://docs.mantidproject.org/nightly/algorithms/FindEPP-v1.html) runs the [`Fit`](http://docs.mantidproject.org/nightly/algorithms/Fit-v1.html) algorithm in a single thread. Perhaps fitting could be parallelized?

gvardany:
`FindEPP` is used when finding elastic TOF channel. The existing algorithm was written in python and has been found to be very slow for an instrument like IN5 at ILL, where one has 10ˆ5 spectra. This PR provides C++ version of the same algorithm, implemented as version 2. It is now 10 to 50 times faster (dependent on number of cores and whether OpenMP is enabled), while providing identical results. A tiny difference is introduced in the behaviour of the algorithm concerning the cases when the fit fails (as described in the documentation), however this should be transparent.

**To test:**
1. Run this script and make sure the answer is positive.
```
ws = CreateSampleWorkspace(Function="User Defined", UserDefinedFunction="name=LinearBackground, A0=0.3, A1=0.001;name=Gaussian, PeakCentre=6000, Height=5, Sigma=75", NumBanks=10,BankPixelWidth=10, XMin=4005.75, XMax=7995.75, BinWidth=2.1, Random=True)
new = FindEPP(ws)
old = FindEPP(ws, Version=1)
new_peaks = ConvertTableToMatrixWorkspace(new,ColumnX="WorkspaceIndex",ColumnY="PeakCentre", ColumnE="PeakCentreError")
old_peaks = ConvertTableToMatrixWorkspace(old,ColumnX="WorkspaceIndex",ColumnY="PeakCentre", ColumnE="PeakCentreError")
result = CompareWorkspaces(new_peaks,old_peaks,Tolerance=0.00001)
print result[0]
```
2. Open the histories of new and old table workspaces, and compare the execution times.
3. Optionally, load the attached IN5 raw data, and run FindEPP version 1 and 2 on a real life example. Report the execution times. If you have OpenMP enabled, follow that the loop over the spectra in the  new version of the algorithm is indeed run in parallel.

[106948.tar.zip](https://github.com/mantidproject/mantid/files/913582/106948.tar.zip)

Fixes #19133 .

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


gvardany:
@mganeva Can you please check if your scripts remain operational with the version 2?

Framework/Algorithms/src/FindEPP.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+/* Call Fit as child algorithm for each spectra
+ * @param index : the workspace index
+ */
+void FindEPP::fitGaussian(int64_t index) {
+  size_t spectrum = static_cast<size_t>(index);
+  const auto &x = m_inWS->x(spectrum).rawData();
+  const auto &y = m_inWS->y(spectrum).rawData();
+  const auto &e = m_inWS->e(spectrum).rawData();
+
```
AnthonyLim23: I think some comment should be added between here and L104 to explain why the algorithm is doing these operations. I had to go and read the source code for the .py version to understand this one.

Framework/Algorithms/src/FindEPP.cpp
```diff
@@ -0,0 +1,196 @@
...
+  if (height > 0) {
+    size_t leftHalf = maxIndex, rightHalf = x.size() - maxIndex - 1;
+    for (auto it = maxIt; it != y.end(); ++it) {
+      if (*it < 0.5 * height) {
+        rightHalf = it - maxIt - 1;
+        break;
+      }
+    }
+    for (auto it = maxIt; it != y.begin(); --it) {
```
AnthonyLim23: You could reverse the start and end iterators and then the break is not needed. Or does that alter the performance?

AnthonyLim23:
Code works as expected and the examples in the script above had about a factor 6 difference on my machine.

AnthonyLim23:
The tests pass too

Framework/Algorithms/src/FindEPP.cpp
gvardany: You are absolutely right, I will add comments here.

Framework/Algorithms/src/FindEPP.cpp
gvardany: I will try with reverse iterators.

gvardany:
I have added more comments. Regarding the reverse iterators, I think the way it is now is anyway the most compact.

AnthonyLim23:
The comments are great. I'm happy to :shipit: once it passes

martyngigg:
@gvardany Should we wait to check with @mganeva before merging?

gvardany:
It would be nice to be confirmed, just in case.

mganeva:
@gvardany @martyngigg I will test it now and let you know.

mganeva:
The TOFTOF workflow works fine with the new FIndEPP. There are some small differences in the result, but they are in an acceptable range:

diff_peak_center

diff_sigma

Since typical TOF is at least 10^3 and typical Sigma for TOFTOF data is in the range of 10, this difference can be considered as negligible small.

mganeva:
To my opinion, it can be merged.

martyngigg:
@mganeva Thanks for double checking this. I'll merge it now.

PR #18450, opened 2017-01-17
ISIS Kafka Listener Implementation

LamarMoore:
**NB. This PR Must be merged with the following third-party [PR](https://github.com/mantidproject/thirdparty-msvc2015/pull/15)!!**

This is the implementation of the a new Live Listener for Mantid which makes use of the [Kafka](https://kafka.apache.org/) and google [flatbuffers](https://google.github.io/flatbuffers/) technologies. This coincides with the development of a kafka based live streaming system which is under development at ISIS. This design will go hand-in-hand with the live stream set-up at the ESS. There were several levels to this development work which included:
 
 - Creating a [cluster](http://shizune:9000/clusters/isis_test_cluster/topics) which could act as a broker for the data stream.
- Creating a [producer](https://github.com/ScreamingUdder/isis_nexus_streamer_for_mantid) which streams large nexus files over the cluster which was used to test the live data stream.
- Laying the groundwork for the listener in mantid.
- Dependency configuration and functionality tests were also performed across all supported platforms along with ensuring packages were correctly built.

**To test:**

- Extensive Code Review.
- Ideally test that builds work on all platforms as well as package generation.

Fixes #xxxx.

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [x] Is the code of an acceptable quality?
- [x] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)? Is it well structured with small focussed classes/methods/functions?
- [ ] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [x] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant documentation been added/updated?
- [ ] Is user-facing documentation written in a user-friendly manner?
- [ ] Has developer documentation been updated if required?

- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

PR #19356, opened 2017-04-12
Do not sort events in LoadLiveData

### LINKED ISSUE #19353: Do not sort events in LoadLiveData
`LoadLiveData` is responsible for loading event data from buffer workspaces in live listeners. It is run periodically with the interval set as a property in `StartLiveData`. Currently event data in the main output workspace is sorted every time `LoadLiveData` is run. This should not be necessary as whether data is sorted or not is recorded in the workspace and sorting is carried out when required, for example before plotting.

Sorting the data is a huge processing overhead during data streaming. This chart shows `LoadLiveData` (set to run once per second) taking increasingly long times to complete due to accumulation of events in a workspace during streaming using the `KafkaEventListener`. The same volume of data is accumulated in the top and bottom row of the chart.

gantt

matthew-d-jones:
`LoadLiveData` executes periodically whilst a live listener is running in order to accumulate data in an output workspace. I've removed event sorting from it as it can have a huge impact on performance (see #19353) and sorting is done whenever it is required anyway. This could save a lot of processing time, worst case scenario is that it just moves the computation to the listener post-processing step.

**To test:**

Could be tested by checking there are no errors when running an event data live listener (ADARA at SNS, Kafka at ISIS).
If at ISIS follow procedure here: https://github.com/mantidproject/mantid/pull/18450#issuecomment-283654382

Fixes #19353.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/LiveData/inc/MantidLiveData/LoadLiveData.h
+++ b/Framework/LiveData/inc/MantidLiveData/LoadLiveData.h
@@ -1,10 +1,10 @@
 #ifndef MANTID_LIVEDATA_LOADLIVEDATA_H_
 #define MANTID_LIVEDATA_LOADLIVEDATA_H_
 
-#include "MantidKernel/System.h"
 #include "MantidAPI/Algorithm.h"
-#include "MantidLiveData/LiveDataAlgorithm.h"
 #include "MantidAPI/Workspace_fwd.h"
+#include "MantidKernel/System.h"
+#include "MantidLiveData/LiveDataAlgorithm.h"
 
 namespace Mantid {
 namespace LiveData {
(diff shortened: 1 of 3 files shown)


matthew-d-jones:
@peterfpeterson Might be worth someone at SNS checking they are happy with this?

peterfpeterson:
I think removing the `SortEvents` is fine. However, I think that the individual listeners should be modified to set the sort type when they bring a chunk in.

matthew-d-jones:
Streamed events are assumed to be unsorted, individual listeners should be modified to set the sort type (to `EventSortType::PULSETIME_SORT` for example) where possible to avoid calling `std::sort` on already sorted data. This PR does not make the situation worse or better than it is already in that respect, so it is a separate issue.

peterfpeterson:
Looks good and shouldn't affect actual usage. :shipit: 

Please make an issue to set the sorting on `SNSLiveEventDataListener`.

PR #19370, opened 2017-04-12
Fix PerformanceTests DiffractionFocussing2TestPerformance

### LINKED ISSUE #19367: Failing Performance Tests
`DiffractionFocussing2TestPerformance` is failing because of the additional memory utilisation from `ComponentInfo`. The tests now push the memory slightly over the 4GB (4.3GB) available on the machine. We need to get things working again.

OwenArnold:
Fixes #19367

I don't mean to be complacent, but until `Instrument2.0` is fully implemented, we are going to have to live with the fact that we have bits of `Instrument2.0` and `Instrument1.0` simultaneously in memory. Looking at `ComponentInfo` and it's construction, and running it through a profiler, there seem to be few gains to be made (I've applied some but they won't actually reduce the memory overhead).

The solution here is to compensate for the increase in the instrument parts by reducing the data parts of the workspaces. 4GB is pretty tight these days for an analysis machine, and the Mantid restrictions of managed workspaces, low-end laptops etc, have been dropped. One thing to note is that this is changing the performance test itself, so is going to make the history prior to these changes hard to follow. In general, I don't like doing this kind of thing.

The changes here reduce the resident memory of the process down to a peak of ~3.8GB. They exceeded 4.3GB prior to these changes (current state of master), and against branch 3.9.0 (no ComponentInfo) it peaked at ~3GB

**Tester**
* Code review,
* Run the performance test `DiffractionFocussing2TestPeformance` under `AlgorithmsTest` and watch the memory usage. Should be below about 4 now.
* If possible run the test above on `mantidlnx02` to verify it runs without crashing.

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/src/ComponentInfo.cpp
+++ b/Framework/API/src/ComponentInfo.cpp
@@ -31,7 +31,7 @@ ComponentInfo::ComponentInfo(
    * Currently that check would break too much in Mantid.
    */
   const auto infoSize = m_componentInfo.size();
-  m_compIDToIndex.reserve(infoSize);
+  m_compIDToIndex->reserve(infoSize);
   for (size_t i = 0; i < infoSize; ++i) {
     (*m_compIDToIndex)[(*m_componentIds)[i]] = i;
   }
--- a/Framework/API/src/ComponentInfo.cpp
+++ b/Framework/API/src/ComponentInfo.cpp
@@ -30,8 +30,9 @@ ComponentInfo::ComponentInfo(
    * m_componentInfo.size().
    * Currently that check would break too much in Mantid.
    */
-
-  for (size_t i = 0; i < m_componentInfo.size(); ++i) {
+  const auto infoSize = m_componentInfo.size();
+  m_compIDToIndex.reserve(infoSize);
+  for (size_t i = 0; i < infoSize; ++i) {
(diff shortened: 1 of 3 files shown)


OwenArnold:
@martyngigg I have requested a review from you because the other possibilities are:

1. Stick more RAM in `mantidlnx02`
2. Retire `mantidlnx02` and use the newer performance test platform instead.

The two changes above would leave the newer performance test job history intact.

VickieLynch:
I tried to do an instrument view of TOPAZ with nightly and it never popped up the view.  Tried mantidplot3.9 on the same machine at the same time and it popped up in a few seconds.  Also LoadIsawPeaks for 6000 peaks now takes 5 minutes for TOPAZ peaks.  Not sure these two problems are related to the original issue.  If the problem is instrument 2.0, remember TOPAZ has approximately 1.5 million pixels.

OwenArnold:
There are a few useful runtime speed additions made here. But #19370 may provide the correct fix to reduce the memory overhead.

SimonHeybrock:
I suppose this PR is superseded by https://github.com/mantidproject/mantid/pull/19403 and can be closed?

PR #19376, opened 2017-04-13
Grey out masked bins in table view

### LINKED ISSUE #17820: Masked bins to gray out in workspace table view.
Masked bins could be highlighted in the workspace table view similar to masked spectra.

gvardany:
This PR adds colour highlighting to masked bins in the table view of matrix workspaces.
Also optimises the caching of masked and monitor spectra by holding only the sets of indices, that are masked or monitor, instead of maps of spectra to boolean flags, no matter true or false.

**To test:**

1. Create sample workspace, mask some bins.
2. See how they appear in table view, mouse hover to check the tooltip. 
3. Make some masked bins in a masked or monitor (or both) spectrum to see if the priority ordering makes sense.

Fixes #17820 .

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


AntonPiccardoSelg:
This does not seem to be working for event-mode workspaces. 

The script to see the difference is:
```python
ws = CreateSampleWorkspace("Event", NumMonitors=4)

# Convert to Histo
rebinned_1 = Rebin(ws, "0, 2000, 20000", PreserveEvents=False)
masked_1 = MaskBins(rebinned_1,XMin=0, XMax=10000)

# Preserve Evenents
rebinned_2 = Rebin(ws, "0, 2000, 20000", PreserveEvents=True)
masked_2 = MaskBins(rebinned_2,XMin=0, XMax=10000)
```

image

gvardany:
I had a look, it seems `MatrixWorkspace::hasMaskedBins()` always returns false for event workspace (had to add the python export, it's missing). So either MaskBins mishandles the event workspace case, or there is some problem in EventWorkspace itself concerning holding/setting/getting the masked bin information. I can try to investigate further, but I think this is not a concern of this ticket.

AntonPiccardoSelg:
Hmm, I think that you mask a TOF range in the case of `EventWorkspace`s, hence seems to be behaving completely differently. I agree with you that fixing this would be too much work for this ticket.  We should mention this in the release notes though. Also we should mention this [here](https://www.mantidproject.org/MBC_The_Workspace_Matrix).

AntonPiccardoSelg:
There is an edge case with partially masked bins: 
```python
ws = CreateSampleWorkspace("Event", NumMonitors=4)

# Convert to Histo
rebinned_1 = Rebin(ws, "0, 2000, 20000", PreserveEvents=False)
masked_1 = MaskBins(rebinned_1,XMin=0, XMax=10000)
rebinned_2 = Rebin(masked_1, "0, 4000, 20000", PreserveEvents=False)
```
image

Should this be highlighted differently?

gvardany:
I added the exception for event mode in release notes, and wiki.
Regarding the partial masking, I think it is up to Rebin to handle if the bins stay masked after split/merge. Seems they still stay flagged as masked, if a masked bin is merged with a non-masked bin, and the highlighting just reflects that logic, which is sensible I think.

AntonPiccardoSelg:
We could have a different colour for partially masked bins. But this is not necessarily part of this PR. 

The functionality is very helpful and works as expected. Please :shipit:

PR #19390, opened 2017-04-17
Made some changes for not copying.

### LINKED ISSUE #19389: Create a MatrixWorkspace from its parent without cloning sample logs
### Expected behavior

Create a MatrixWokspace from its parent without cloning Run object.

### Actual behavior

Mantid does not support this.

wdzhou:
Adding methods such that a workspace can be created from its parent without copying run object.

**To test:**

Fixes #19389.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


SimonHeybrock:
I am not so happy with these changes. The workspace factory methods are already very complex and adding yet another argument makes things even worse. Have you considered using a copy-on-write mechanism for `Run` stored in `ExperimentInfo`? That is, use `Kernel::cow_ptr<Run>`?

SimonHeybrock:
@wdzhou I meant a bit like this: https://github.com/mantidproject/mantid/commit/05579ccdc31c172e22ec861a1ff489103750ea40 (I have not compiled/tested this, it is just a draft).

However, looking a `ExperimentInfo` and `WorkspaceFactory::initializeFromParent` (in `master`, not my branch) **the `Run` object is already shared** (the copy constructor does not share but the factory does). Given that, I am not sure how to interpret https://github.com/mantidproject/mantid/issues/19389?

wdzhou:
@SimonHeybrock I think it might be easy to explain with my use case.  In algorithm FilterEvents, a set of child EventWorkspaces will be created from their parent, i.e., the input EventWorkspace.  The current create() method clones all the sample logs.  Later in FilterEvents, all the TimeSeriesProperty sample logs in the child EventWorkspaces will be replaced by the processed (i.e., split) TimeSeriesProperties.  It is not efficient at all. 
This issue is not maximized by an experiment on VULCAN.  It has 3 sample logs that are in the size of GB.  The users decide not to copy them to the child EventWorkspaces.  Thus, I need a method to create a workspace with empty Run object such that I can add the needed sample logs back later.

SimonHeybrock:
I still think that adding more options/arguments to the already quite complex factory methods and functions is not a clean way of solving this:
- The reasons for the change seems rather special, whereas the `create` functions and `initializeFromParent` are used many times, e.g., close to 40 times for the latter.
- `bool` arguments in general should be avoided.
- `create` is heavily templated and adding another argument increases risk of confusion.
- `initializeFromParent` already has another `bool` argument, having two would be even worse.

I don't see why this cannot be solved by adding `ExperimentInfo::setSharedRun(boost::shared_ptr<Run> run)` (or `Kernel::cow_ptr` instead of `shared_ptr`). If factory functions are essential, I believe they should have a distinct name (as you partially did in your latest commit), but they should *not* be extending existing `create` methods by means of new (default) arguments.

NickDraper:
Given the discussion about the correct way to handle this, I suspect that this will not make R3.10.

Please let me know if I'm wrong, but I'm going to move the milestone for now.

peterfpeterson:
Creating new methods rather than adding extra options to existing methods seems like a good middle road. I like the new names. :shipit:

AndreiSavici:
No release notes necessary since this is an internal change

Framework/API/src/WorkspaceFactory.cpp
```diff
@@ -76,27 +77,25 @@ WorkspaceFactoryImpl::create(const MatrixWorkspace_const_sptr &parent,
...
     const MatrixWorkspace &parent, MatrixWorkspace &child,
     const bool differentSize) const {
   child.setTitle(parent.getTitle());
   child.setComment(parent.getComment());
   child.setInstrument(parent.getInstrument()); // This call also copies the
                                                // SHARED POINTER to the
                                                // parameter map
+
+  //  g_log.warning() << child.run().getProperites() << "\n";
+
```
SimonHeybrock: Left over commented code.

Framework/API/src/WorkspaceFactory.cpp
```diff
@@ -76,27 +77,25 @@ WorkspaceFactoryImpl::create(const MatrixWorkspace_const_sptr &parent,
...
   // This call will (should) perform a COPY of the parameter map.
   child.instrumentParameters();
   child.m_sample = parent.m_sample;
-  child.m_run = parent.m_run;
+  child.m_run = boost::make_shared<API::Run>();
+  //  if (!noproperty)
+  //    child.m_run = parent.m_run;
+  //  else
+  //    child.m_run = boost::make_shared<API::Run>();
```
SimonHeybrock: Left over commented code.

Framework/DataObjects/inc/MantidDataObjects/WorkspaceCreation.h
```diff
@@ -137,8 +137,14 @@ createConcreteHelper();
...
 initializeFromParent(const API::MatrixWorkspace &parent,
                      API::MatrixWorkspace &ws);
+
+MANTID_DATAOBJECTS_DLL void
+initializeFromParentWithoutLogs(const API::MatrixWorkspace &parent,
+                                API::MatrixWorkspace &ws);
 }
 
+/** This is the create() method that all the other create() methods call
+ */
```
SimonHeybrock: This is misleading, the method is also called directly.

Framework/DataObjects/inc/MantidDataObjects/WorkspaceCreation.h
```diff
@@ -163,7 +169,8 @@ std::unique_ptr<T> create(const P &parent, const IndexArg &indexArg,
     }
   }
 
-  // The instrument is also copied by initializeFromParent, but if indexArg is
+  // The instrument is also copied by initializeFromParent or
+  // initializeFromParentWithoutLogs, but if indexArg is
```
SimonHeybrock: This change to the comment is confusing since there is no call to `initializeFromParentWithoutLogs` in this function.

Framework/DataObjects/test/WorkspaceCreationTest.h
```diff
@@ -172,6 +172,15 @@ class WorkspaceCreationTest : public CxxTest::TestSuite {
     check_instrument(*ws);
   }
 
+  void test_create_parent_without_logs() {
+    const auto parent = create<Workspace2D>(m_instrument, make_indices(),
+                                            Histogram(BinEdges{1, 2, 4}));
+    const auto ws = createWithoutLogs<Workspace2D>(*parent);
+    check_indices(*ws);
+    check_zeroed_data(*ws);
+    check_instrument(*ws);
```
SimonHeybrock: You are not testing anything here? I could replace line 178 by `create` and the test would still pass, I think?

PR #19391, opened 2017-04-17
Fix min range in Integration for 0 events

rosswhitfield:
When working with live data you sometimes get chucks with zero events and `Integration` fails when this happens with something like
`Invalid value for property Params (dbl list) "1.79769e+308,-1.79769e+308,1000": Bin boundary values must be given in order of increasing value`

It shouldn't fail like this if you provide `RangeLower` and `RangeUpper`. This also makes the handling of `minRange` and `maxRange` consistent.

This reverts the changes to this line which was part of #18983 

**To test:**
This should now work.
```
ws = CreateSampleWorkspace(WorkspaceType="Event",NumEvents=0)
out=Integration(ws, RangeLower=0, RangeUpper=1000)
```

*As this was only changed in this release it does not need to be in the release notes.*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/src/Integration.cpp
+++ b/Framework/Algorithms/src/Integration.cpp
@@ -142,7 +142,8 @@ void Integration::exec() {
           "Range lists not supported for EventWorkspaces.");
     }
     // Get the eventworkspace rebinned to apply the upper and lowerrange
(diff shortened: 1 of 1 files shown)


peterfpeterson:
Works as advertised. :shipit:

PR #19398, opened 2017-04-18
Time indexes for detector table

### LINKED ISSUE #19380: Add Time Indexes to Detector Table
To make testing for scanning workspaces a little easier time index information should be added to the detector table in Mantid. The behaviour should be similar to that for detector IDs, with multiple time indexes being shown where appropriate.

ianbush:
This adds a column showing the time indexes to the detector table. The logic is the same as for detector IDs - if there are more than 10 entries then the indexes are just displayed as `1, 2...(100 more)...102, 103`.

**To test:**

Please use the changes in #19728 to test this branch. `CreateSampleWorkspace` now has the option to create a workspace with mulitple time indexes, and `GroupDetectors` can be used to test workspaces with more than one time index.

1. A workspace with no time indexing should not show the column at all
1. A workspace with less a single time index should show the time index
1. A workspace with less than 10 time indexes should show in a comma separated list
1. A workspace with more than 10 time indexes should show in a list such as `1, 2...(100 more)...102, 103`

Fixes #19380.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Types/CMakeLists.txt
```diff
@@ -14,5 +14,11 @@ if (COVERALLS)
   endforeach(loop_var)
 endif()
 
+# Add the target for this directory
+add_library ( Types ${INC_FILES} )
+# Set the name of the generated library
+set_target_properties ( Types PROPERTIES OUTPUT_NAME MantidTypes
+        COMPILE_DEFINITIONS IN_MANTID_TYPES LINKER_LANGUAGE CXX )
```
SimonHeybrock: `Types` is a header-only module so I am not sure why this should be required...?

MantidPlot/src/Mantid/MantidUI.cpp
```diff
@@ -1377,6 +1361,15 @@ Table *MantidUI::createDetectorTable(
...
       colValues << QVariant(specNo) << QVariant(detIds);
+      if (isScanning) {
+        std::set<int> timeIndexSet;
+        for (auto def : spectrumInfo.spectrumDefinition(wsIndex)) {
+          timeIndexSet.insert(int(def.second));
+        }
+
+        QString timeIndexes = createTruncatedList(timeIndexSet);
+        colValues << QVariant(timeIndexes);
+      }
```
SimonHeybrock: So if there is more than one detector for the spectrum you would get duplicate time indices and cannot associate them to detectors? Probably this is not a common case, but just to make sure this is not overlooked. Should it simply be ignored? Or generate a warning? Throw?

ianbush: Just to be clear, are you saying that you could have a spectrum with 2 detectors, one contributing time indexes [1, 2] and another contributing time indexes [2, 3]. You would then see the time indexes as [1, 2, 3] without knowing which detectors had which time indexes?

I would think we can ignore that, as it should be easy to see when you have multiple time indexes and multiple detectors. The detector table does not always give the complete information in any case, truncating long lists.

The other alternative would be to do something like "1 (1, 2), "2 (2, 3)" in the detector part of the table. I think that would get more difficult to read clearly though.

SimonHeybrock: Yes that is what I meant. I suppose you would see the time indices as [1,2,2,3].

ianbush: But it is coming from a set, so you would see [1,2,3] ?

Framework/Types/CMakeLists.txt
ianbush: Is this what you would expect?

SimonHeybrock: I am not sure. Why are you even creating a library? Maybe making it an interface library would make sense? ```add_library( Types INTERFACE ...)```

ianbush: `add_library ( Types INTERFACE )` seems to be sufficent for the build to work. Without this I get the error: `cannot find -lTypes`.

SimonHeybrock:
Looks good now, :shipit:

PR #19403, opened 2017-04-19
API::ComponentInfo, API::DetectorInfo std::unorderd_map insert performance

### LINKED ISSUE #19385: BinMD too slow for Sliceviewer (300 times slower than 3.9)
Reported from TOPAZ current experiment user:
### Expected behavior
For 8 Gb MDEvent file, BinMD takes 0.05 seconds for each slice rebinned with MantidPlot3.9
### Actual behavior
For the same file with nightly, BinMD takes 15 seconds for each slice which makes SliceViewer with auto rebin too slow to use.  Profiler shows the time is mostly in ExperimentInfo.cpp line 99 set/get instrument.
LoadMD with large MDEvent file.  Open in SliceViewer and select auto rebin and then change slices.
### Platforms affected
All

OwenArnold:
Upon completion BinMD copies `ExperimentInfo`s from input to output workspaces. Since BinMD is used interactively as part of the SliceViewer, it is executed many, many times, and the slow down in this copying process has become rather noticeable. 

The vast majority of the time is spent inserting mapping key-value pairs for `ComponentID`->`ComponentIndex` and `DetectorID`->`DetectorIndex` into the respective `std::unordered_maps` in the constructors of `API::ComponentInfo` and `API::DetectorInfo`. 

**While the problem is quite clear. I am happy to debate the exact fix** as there does not seem to be an approach that is completely satisfactory. **Here are things I have tried**.

* Reserving the memory for the `std::unordered_map`s. This does make a small ~10% improvement, and I have left this small optimization in place, since we do know the sizes of each of these mappings aprori.
* Tried different container `std::map` on the basis that insertion is less costly. However, you cannot pre-allocate this data structure, and this caused a further slowdown.
* Tried different contatiner `boost::container::flat_tree`. This allows pre-allocation of the mapping, and sorts the contents, so insertion should be good, but this was again significantly slower.
* When an Instrument is set (or reset) on an `ExperimentInfo` we do not expect that base instrument to be different. We should therefore expect that `ComponentID`->`ComponentIndex` and  `DetectorID`->`DetectorIndex`  are immutable! Could we therefore cache the mappings outside of the `ComponentInfo` and `DetectorInfo` for reuse? Code for this starts looking quite ugly and we end up leaking a lot that is currently encapsulated.

What has been implemented here is lazy calculation of the mappings. These don't get calculated until `indexOf` is first called. This does dramatically improve the performance, however, thread synchronization now needs to be considered and this has an overhead.

**To test:**

- Fuzzy test - run BinMD in-situ such as in in the SliceViewer. It should be as interactive as v3.9.0
- Time `ExperimentInfo::setInstrument` this should be significantly faster than master and closer to v3.9.0 speeds.

Fixes #19385 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


DavidFair:
Something that has just come into my head whilst reading this PR could we use a `future` from the std library and create this mapping asynchronously?
`indexOf` will block if called immediately after constructing the object but I'm sure most algorithms could put the construction of `ComponentInfo` as one of the first things they do and perform various other ops whilst the index mapping is generated in a separate thread. 

According to http://www.cplusplus.com/reference/future/future/ this is also synchronised correctly across multiple threads too. 

This should help hide the worst of the effects whilst we decide on a better container.

SimonHeybrock:
I would strongly prefer the last option you mention (carrying the maps along with `Instrument`). If the execution of a pice of code can be avoided (or the execution count drastically reduced as in this case) I think that is always preferable to optimizing it. Optimized and threaded code is much harder to debug.

Regarding uglyness and leaking encapsulation of that last option: We are already storing `Beamline::DetectorInfo`, `Beamline::ComponentInfo`, and a vector of component IDs in the `Instrument`. Now we want to add two more maps to that list. I think this would be the right time to introduce a little `struct` holding these objects together, which might eventually develop into a new `Instrument` class (i.e., the `Instrument` equivalent for Instrument-2.0).

OwenArnold:
@DavidFair Thanks for the suggestion. I actually wasn't really aware that futures had made their way into c++11. However, I don't think that threading is the right solution in this case.

VickieLynch:
7.3Gb MD file in SliceViewer
v.3.9:
LoadMD started
LoadMD successful, Duration 1 minutes 13.90 seconds
BinMD started
BinMD successful, Duration 0.10 seconds
BinMD started
BinMD successful, Duration 0.03 seconds
BinMD started
BinMD successful, Duration 0.03 seconds
BinMD started
BinMD successful, Duration 0.03 seconds
BinMD started
BinMD successful, Duration 0.03 seconds
BinMD started
BinMD successful, Duration 0.04 seconds
BinMD started
BinMD successful, Duration 0.04 seconds
this version:
LoadMD started
LoadMD successful, Duration 1 minutes 51.54 seconds
BinMD started
BinMD successful, Duration 10.71 seconds
BinMD started
BinMD started
BinMD started
BinMD: Execution terminated by user.
BinMD started
BinMD successful, Duration 3.44 seconds
BinMD successful, Duration 3.54 seconds
BinMD successful, Duration 3.43 seconds
BinMD started
BinMD started
BinMD started
BinMD successful, Duration 2.95 seconds
BinMD successful, Duration 3.29 seconds
BinMD successful, Duration 3.19 seconds

Framework/API/inc/MantidAPI/ComponentInfo.h
```diff
@@ -59,10 +59,21 @@ class MANTID_API_DLL ComponentInfo {
       const Mantid::Beamline::ComponentInfo &componentInfo,
       boost::shared_ptr<const std::vector<Mantid::Geometry::IComponent *>>
           componentIds);
+
+  ComponentInfo(
+      const Mantid::Beamline::ComponentInfo &componentInfo,
+      boost::shared_ptr<const std::vector<Mantid::Geometry::IComponent *>>
+          componentIds,
+  boost::shared_ptr<std::unordered_map<Geometry::IComponent *, size_t>>
```
SimonHeybrock: Maybe this could be made `const std::unordered_map`, to prevent accidential changes.

Framework/API/inc/MantidAPI/ExperimentInfo.h
```diff
@@ -229,7 +230,7 @@ class MANTID_API_DLL ExperimentInfo {
   boost::shared_ptr<Beamline::ComponentInfo> m_componentInfo;
   std::unique_ptr<API::ComponentInfo> m_componentInfoWrapper;
   boost::shared_ptr<const std::vector<Geometry::ComponentID>> m_componentIds;
-
+  boost::shared_ptr<std::unordered_map<Geometry::ComponentID, size_t>> m_componentIdToIndexMap;
```
SimonHeybrock: Why does `ExperimentInfo` need this?

Framework/API/src/InfoComponentVisitor.cpp
```diff
@@ -39,6 +40,8 @@ void InfoComponentVisitor::registerComponentAssembly(
 
   m_ranges.emplace_back(std::make_pair(detectorStart, detectorStop));
 
+  // Record the ID -> index mapping
+  m_componentIdToIndexMap[assembly.getComponentID()] = m_componentIds.size();
```
SimonHeybrock: Why are you creating this in the visitor now? It is still also done in `ComponentInfo` itself?

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -300,12 +306,17 @@ void ExperimentInfo::setInstrument(const Instrument_const_sptr &instr) {
...
   m_parmap->setComponentInfo(m_componentInfo);
-  m_componentInfoWrapper =
-      Kernel::make_unique<ComponentInfo>(*m_componentInfo, m_componentIds);
-  // Detector IDs that were previously dropped because they were not part of the
+
+    /*
+     * If the ID -> index map has already been created. Reuse it.
+     * */
+    m_componentInfoWrapper = Kernel::make_unique<ComponentInfo>(
+        *m_componentInfo, m_componentIds, m_componentIdToIndexMap);
```
SimonHeybrock: I am not quite sure I understand why all these changes to `ExperimentInfo` are necessary. Can't you keep creating the IDs in `API::ComponentInfo` and simple extract them in `getInstrument()`, and here cover 2 cases:
- If the instrument has `ComponentInfo` (and thus the map), use the new constructor you added to `API::ComponentInfo`.
- Otherwise, use the older constructor and let `ComponentInfo` build the map.

SimonHeybrock (commented):
@OwenArnold Some comments below. Some changes look a bit redundant, but maybe I do not understand an underlying issue?

Framework/API/src/InfoComponentVisitor.cpp
OwenArnold: It won't be. Doing it in the visitor removes the need for the `ComponentInfo` to have to return the map. Not complete yet. I still have a lot of changes to push.

Framework/API/inc/MantidAPI/ComponentInfo.h
OwenArnold: 👍

VickieLynch:
Loading 9500 TOPAZ peaks is improved from nightly!
mantidplot3.9
LoadIsawPeaks started
LoadIsawPeaks successful, Duration 2.82 seconds
nightly:
progess bar predicts over 3 min, but it crashes every time.
this branch:
LoadIsawPeaks started
LoadIsawPeaks successful, Duration 8.69 seconds

Framework/API/inc/MantidAPI/ExperimentInfo.h
```diff
@@ -12,6 +12,7 @@
 
 #include <list>
 #include <mutex>
+#include <unordered_map>
```
SimonHeybrock: Leftover include.

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -300,12 +306,20 @@ void ExperimentInfo::setInstrument(const Instrument_const_sptr &instr) {
...
+      componentIdToIndexMap;
+  std::tie(m_componentInfo, componentIds, componentIdToIndexMap) =
       makeComponentInfo(*instr, detectorInfo());
   m_parmap->setComponentInfo(m_componentInfo);
-  m_componentInfoWrapper =
-      Kernel::make_unique<ComponentInfo>(*m_componentInfo, m_componentIds);
-  // Detector IDs that were previously dropped because they were not part of the
+
+    /*
+     * If the ID -> index map has already been created. Reuse it.
```
SimonHeybrock: Outdated comment? Map is now always provided by `makeComponentInfo`?

Framework/API/src/InfoComponentVisitor.cpp
```diff
@@ -137,5 +144,9 @@ size_t InfoComponentVisitor::size() const {
   return m_componentIds.size() - m_droppedDetectors;
 }
 
+const std::unordered_map<Mantid::Geometry::IComponent *, size_t> &
+InfoComponentVisitor::componentIdToIndexMap() const {
+  return m_componentIdToIndexMap;
```
SimonHeybrock: For extra speedup return by value here and use `return std::move(m_componentIdToIndexMap`, and the same for `componentIds()` above. That avoids two copies all in all.

Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
```diff
@@ -48,6 +48,8 @@ class MANTID_BEAMLINE_DLL ComponentInfo {
                 const std::vector<std::pair<size_t, size_t>> &ranges);
   std::vector<size_t> detectorIndices(const size_t componentIndex) const;
   size_t size() const;
+  bool operator==(const ComponentInfo& other) const;
+  bool operator!=(const ComponentInfo& other) const;
```
SimonHeybrock: I am not sure what these are needed for. Also keep in mind that it will soon not be possible to provide this anymore -- once you add positions and rotations to `ComponentInfo` exact comparison are not useful, see `DetectorInfo::isEquivalent`.

SimonHeybrock (commented):
Looks good so far, just small comments below.

Given that Vickie already sees a good speedup: if you now do the same for the map in `DetectorInfo` we should be getting closer to where we need to be.

Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
OwenArnold: @SimonHeybrock yes, I'll move these to `DetectorInfo::isEquivalent` as part of the positions/rotations. I did have this in place to support some of the unit tests. I'm going to leave it in for now.

OwenArnold:
@SimonHeybrock I'm working through changes for `DetectorInfo` shouldn't take long now.

Framework/API/src/InfoComponentVisitor.cpp
OwenArnold: Hmm. I'm not sure about this @SimonHeybrock. Maybe we can talk about this.

OwenArnold:
Running ```MantidPlot -x scripts/Vates/Diffraction_Workflow.py 
``` and playing with BinMD. I'm getting `BinMD` execution times in the slice viewer back down at approx 0.03s

File: Framework/API/src/DetectorInfo.cpp [hunk 9]
```diff
@@ -29,11 +46,44 @@ DetectorInfo::DetectorInfo(
   // Note: This does not seem possible currently (the instrument objects is
```
SimonHeybrock: Can we just use destructor delegation to avoid a lot of code duplication here?

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -297,14 +304,30 @@ void ExperimentInfo::setInstrument(const Instrument_const_sptr &instr) {
...
 
-  std::tie(m_componentInfo, m_componentIds) =
+    // Reuse the ID -> index map for the detector ids
+    m_detectorInfoWrapper = Kernel::make_unique<DetectorInfo>(
+        *m_detectorInfo, getInstrument(), m_parmap.get(),
+        instr->detIdToIndexMap());
+  } else {
+    m_detectorInfoWrapper = Kernel::make_unique<DetectorInfo>(
+        *m_detectorInfo, getInstrument(), m_parmap.get(),
+        makeDetIdToIndexMap(instr->getDetectorIDs()));
```
SimonHeybrock: If you call this here to create the map, why do we still need the "old" constructor for `DetectorInfo` that creates the map itself?

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -322,8 +345,19 @@ Instrument_const_sptr ExperimentInfo::getInstrument() const {
   checkDetectorInfoSize(*sptr_instrument, *m_detectorInfo);
   auto instrument = Geometry::ParComponentFactory::createInstrument(
       sptr_instrument, m_parmap);
-  instrument->setDetectorInfo(m_detectorInfo);
-  instrument->setComponentInfo(m_componentInfo, m_componentIds);
+  // We can only set the DetectorInfo on the return Instrument if the API
+  // wrapper is fully constructed
+  if (m_detectorInfoWrapper) {
+    instrument->setDetectorInfo(m_detectorInfo,
+                                m_detectorInfoWrapper->detIdToIndexMap());
```
SimonHeybrock: What is happening here (or rather, why was this change neessary?)? This looks dangerous -- we might lose information stored in `Beamline::DetectorInfo`, such as positions?

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -322,8 +345,19 @@ Instrument_const_sptr ExperimentInfo::getInstrument() const {
...
+  }
+  // We can only set the ComponentInfo on the return Instrument if the API
+  // wrapper is fully constructed
+  if (m_componentInfoWrapper) {
```
SimonHeybrock: Same comment as for `DetectorInfo`.

Framework/API/src/InfoComponentVisitor.cpp
```diff
@@ -126,6 +134,18 @@ InfoComponentVisitor::assemblySortedDetectorIndices() const {
...
+   * Detectors can be dropped (see above). This is a workaround of Instrument
+   * 1.0 for
+   * the fact that in some cases we have IDFs that provide an invalid ID for a
+   * detector.
+   * However, we use a fixed offset for the non-detector component indexes
+   * calculated to be the
+   * exact size of the number of detectors we expect to get. When we drop
+   * detectors we introduce gaps in
+   * an otherwise contiguous range.
+  */
```
SimonHeybrock: I suppose this comment is unrelated to the changes done in this PR but existed previously? Does this cause issues in some tests?

Framework/API/src/InfoComponentVisitor.cpp
SimonHeybrock: I just thought that it is an optimization you get for free without even adding an extra line of code. Probably not super relevant though.

SimonHeybrock (commented):
Looks good. A few small comments below:

Framework/API/src/ExperimentInfo.cpp
OwenArnold: This is necessary given the way that `getInstrument` currently ties in with `ExperimentInfo::ExperimentInfo` and `setInstrument`. I'm sure this can be done a better way, but would appreciate your input here.

Framework/API/src/ExperimentInfo.cpp
OwenArnold: Yes. That's correct it's now redundant. It needs to be removed. I was going to do this as a latter step so that we could get the fix out and available for Vickie asap.

Framework/API/src/InfoComponentVisitor.cpp
OwenArnold: Correct. This is a note for the future. Doesn't currently cause an issue. I think the correct fix should be at the IDF level, I don't like this dropped detector thing. Failing that being possible, we would have to "shift down" or index ranges before actually returning the containers we create.

Framework/API/src/ExperimentInfo.cpp
SimonHeybrock: I was just wondering why this chage was necessary (it seems to have worked before).

SimonHeybrock:
:shipit: (smaller cleanup discussed in comments will be dealt with in follow-up PR).

PR #19407, opened 2017-04-20
Modify camera toolbar to snap to nonorthogonal axes.

quantumsteve:
Description of work.

This modifies the ParaView camera toolbar to align the view along the crystallographic instead of cartesian axes. This required reworking the `pqCameraToolbar` and `pqCameraReaction`classes and adding them to the VSI. 

I'm allowing VTK to orthogonalize the up direction because transforming it doesn't seem to produce a better result.

The icons still show X, Y, Z. We may want to change these before the release.

**To test:**

1. Open a MDHistoWorkspace with nonorthogonal axes in the VSI. I used a hexagonal cell (symm.nxs from @AndreiSavici), so another file with different symmetry would be ideal.
2. Click the icons in the camera toolbar and verify that the views are now along the crystallographic axes defined by the change-of-basis matrix.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


quantumsteve:
An alternative would be to view perpendicular to the other two axes. I'm not sure which option is better.

AndreiSavici (approved):
Works as expected. :shipit:

Issue #19386, opened 2017-04-14
BASIS instrument view reverted to showing the neutronic positions


jmborr:
Pull request #18906 has caused BASIS to display neutronic positions when User requests `Show Instrument`.
The current BASIS IDF in master only contains neutronic positions but the problem persists when using and IDF containing both neutronic and real positions. Thus, there are two issues here.

SimonHeybrock:
@jmborr After an initial look: https://github.com/mantidproject/mantid/blob/master/Framework/Geometry/src/Instrument.cpp#L164

The changes in the PR you linked extract (i.e., remove) information from that parameter map. For example positions of detectors are now stored in the class `DetectorInfo`, instead of having relative positions changes in the parameter map.
However, the physical instrument would have a different detector, so the parameters for it should be untouched.

I will have to have a closer look, but first I need to track down some other issues, so this is just my initial thought.

SimonHeybrock:
I think I understand now what is happening, it seems indeed due to sharing of the parameter map. In particular, the problem is that the parameter map forwards to `DetectorInfo` when positions are read via the old legacy interface (such as `Detector::getPos()`). However, `DetectorInfo` of course contains positions for the neutronic instrument, and due to the shared parameter map with the physical instrument we get the wrong positions.

Will need to think a bit about a clean solution.

SimonHeybrock:
@jmborr There is a probably a quick fix that can be done in `Instrument`, but that will make the `InstrumentView` very slow. However, that "fix" does *not* work with the latest `BASIS` definition, due to changes done here: https://github.com/mantidproject/mantid/pull/19273. The changes there remove the tag `"indirect-neutronic-positions"`. Without that, Mantid does not create a physical instrument (see `InstrumentDefinitionParser`).

I will try and implement a proper fix (without `InstrumentView` slowdown) on Thursday (I am on leave tomorrow). Maybe you can fix the BASIS definition in the meantime?

jmborr:
@SimonHeybrock Thanks a million for taking the time and for spotting the bug in the latest instrument definition file. I'll fix the IDF right away

jmborr:
Pull request #19409 has addressed the first issue so now real space positions are shown when using an IDF containing both neutronic and real space positions.
However, the BASIS IDF in master has to be corrected to include real space positions as before issue #19258

jmborr:
Resolved _via_ #19421

PR #19409, opened 2017-04-20
Fix parameter map access for physical instrument

SimonHeybrock:
Recent changes to position access via the parameter map incorrectly caused the physical instrument to read its positions from `DetectorInfo`. `DetectorInfo` contains only positions for the neutronic instrument. This is fixed here. The solution should be seen as an temporary one, since once Instrument-2.0 has progressed further, it is planned to deal with the physical/neutronic distinction in a different way. However, "temporary" probably will still be a rather long time, so when reviewing please make sure that this solution is reasonable.

**To test:**

- Code review.
- Check that InstrumentView now correctly shows the *physical* instrument instead of the neutronic one (as it used to be). Cannot be tested with BASIS currently, until its IDF is fixed.

No related issue, but this solves parts of #19386 (but also the BASIS IDF needs to be fixed for that).

No release notes, issue was introduced after last release.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Geometry/src/Instrument/Detector.cpp
+++ b/Framework/Geometry/src/Instrument/Detector.cpp
@@ -172,8 +172,8 @@ void Detector::registerContents(ComponentVisitor &componentVisitor) const {
 
 bool Detector::hasDetectorInfo() const {
   const IComponent *root = this;
-  while (this->m_parent)
-    root = this->m_parent;
+  while (auto parent = root->getBareParent())
+    root = parent;
   auto instrument = dynamic_cast<const Instrument *>(root);
   return m_map->hasDetectorInfo(instrument);
 }
--- a/Framework/Geometry/src/Instrument/Detector.cpp
+++ b/Framework/Geometry/src/Instrument/Detector.cpp
@@ -171,7 +171,7 @@ void Detector::registerContents(ComponentVisitor &componentVisitor) const {
 }
 
 bool Detector::hasDetectorInfo() const {
-  const IComponent *root = this;
+  const IComponent *root = m_base;
   while (auto parent = root->getBareParent())
     root = parent;
   auto instrument = dynamic_cast<const Instrument *>(root);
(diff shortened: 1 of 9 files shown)


Framework/Geometry/src/Instrument/Detector.cpp
```diff
@@ -169,5 +170,13 @@ void Detector::registerContents(ComponentVisitor &componentVisitor) const {
   componentVisitor.registerDetector(*this);
 }
 
+bool Detector::hasDetectorInfo() const {
+  const IComponent *root = this;
+  while (this->m_parent)
```
OwenArnold: This looks suspect. Is this supposed to be `root->m_parent` or something similar?

Framework/Geometry/src/Instrument/ParameterMap.cpp
```diff
@@ -1152,14 +1152,20 @@ ParameterMap::create(const std::string &className,
...
-/// contains a DetectorInfo.
-bool ParameterMap::hasDetectorInfo() const {
+/** Only for use by ExperimentInfo. Returns returns true if this instrument
+ contains a DetectorInfo.
+
+ The `instrument` argument is needed for the special case of having a neutronic
+ *and* a physical instrument. `Instrument` uses the same parameter map for both,
+ but the DetectorInfo is only for the neutronic instrument. */
+bool ParameterMap::hasDetectorInfo(const Instrument *instrument) const {
+  if (instrument != m_instrument)
```
OwenArnold: Would be good to add a comment around `m_instrument` in the header regarding the fact that `m_instrument` is taken to be the neutronic one.

jmborr:
@SimonHeybrock  I have a BASIS instrument with both neutronic and real space positions which I'll use for testing.

jmborr:
This is BASIS view from branch **master**  using an IDF containing both neutronic and real space positions. It shows the neutronic positions:
before
This is BASIS view from branch **fix_ParameterMap_access_for_physical_instrument** using the previous IDF. It shows the real space positions:
after

Looks good to me.
:shipit:

jmborr:
@OwenArnold Can you merge the pull-request?

PR #19432, opened 2017-04-24
Change openssl function in gsoap

rosswhitfield:
Mantid fails to compile with openssl 1.1.0 This fixes the "`M_ASN1_STRING_data’ was not declared in this scope" error with openssl 1.1.0.

This switches to using `ASN1_STRING_data` which is itself deprecated in `1.1.0` and will some point need to be changed to the new function [ASN1_STRING_get0_data](https://www.openssl.org/docs/manmaster/man3/ASN1_STRING_get0_data.html) :unamused: but this will work for now.

**To test:**
As long as everything still compiles and passes the tests it should be good.

I also started a build for this branch on Arch which has openssl 1.1.0 [here](http://builds.mantidproject.org/job/master_clean-archlinux/853/)

The better solution is to update gsoap version (openssl 1.1.0 support was added in [2.8.41](https://www.genivia.com/changelog.html)) but that is more work.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/ICat/src/GSoap/stdsoap2.cpp
+++ b/Framework/ICat/src/GSoap/stdsoap2.cpp
@@ -4547,7 +4547,7 @@ again:
               break;
             name = X509_NAME_ENTRY_get_data(X509_NAME_get_entry(subj, i));
             if (name) {
-              if (!soap_tag_cmp(host, (const char *)M_ASN1_STRING_data(name)))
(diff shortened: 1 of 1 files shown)


quantumsteve:
This change is the easiest way to fix the above build error with OpenSSL 1.1.0. I created a separate issue for migrating to the system version of gSOAP.

:shipit:

PR #19451, opened 2017-04-27
Improve parallel scaling of MDNormSCD

quantumsteve:
Description of work.

A user noticed that the MDNormSCD algorithm runs much slower on our 12-32 core analysis computers than a 2-core laptop. A lot of time is spent in the critical section synchronizing updates to the signal array instead of doing work.

In this PR, I use `std::atomic::compare_exchange_weak` as suggested in [this reddit thread](https://www.reddit.com/r/cpp/comments/338pcj/atomic_addition_of_floats_using_compare_exchange/) to update the values of a temporary signal array before copying them into the MDHistoWorkspace. This appears to make a huge difference, reducing the execution time in the user's script from several minutes to 15 seconds (ndav3.sns.gov) and ~45 seconds (topaz.sns.gov)

**To test:**

@AndreiSavici Would you try this on one or more inputs and verify that it
  - Shows a considerable performance improvement
  - Still gives the correct result

@SimonHeybrock I would appreciate a second opinion on the atomic operations

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/MDAlgorithms/src/MDNormSCD.cpp
+++ b/Framework/MDAlgorithms/src/MDNormSCD.cpp
@@ -363,6 +363,16 @@ void MDNormSCD::cacheDimensionXValues() {
   }
 }
 
+namespace {
+template <typename T, typename BinaryOp>
+void AtomicOp(std::atomic<T> &f, T d, BinaryOp _Op) {
+  T old = f.load();
+  T desired = _Op(old, d);
+  while (!f.compare_exchange_weak(old, desired))
+    desired = _Op(old, d);
+}
+} // namespace
+
 /**
  * Computed the normalization for the input workspace. Results are stored in
  * m_normWS
@@ -401,6 +411,8 @@ void MDNormSCD::calculateNormalization(
       integrFlux->getDetectorIDToWorkspaceIndexMap();
   const detid2index_map solidAngDetToIdx =
       solidAngleWS->getDetectorIDToWorkspaceIndexMap();
+
+  std::vector<std::atomic<signal_t>> signalArray(m_normWS->getNPoints());
 
   auto prog = make_unique<API::Progress>(this, 0.3, 1.0, ndets);
   PARALLEL_FOR_IF(Kernel::threadSafe(*integrFlux))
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
Had only a short look at this so far, will have to dig into `compare_exchange_weak` more to really verify this (but it looks reasonable, I think). Before I do that, two questions:

- Have you considered using, e.g., `std::atomic::fetch_add` (maybe with a custom memory order)?
- I do not really know how big `m_normWS->getNPoints()` is. If it is not too big, have you considered simply accumulating into a separate vector in each thread and adding them at the end? That would completely avoid any synchronization.

Apart from that, there is also some inefficiency due to the use of `SpectrumInfo::detector()`. If you can avoid that you might see some further speedup. I am also not sure that all the the `detid_t` to workspace index maps a required (but I did not check in detail).

quantumsteve:
@SimonHeybrock 
- `std::atomic::fetch_add` is only available with integral and pointer types.
- The signal array is quite large (The user's example contained 701^3 doubles = 2.6 GB). Making 64 copies dramatically increases memory requirements.

Framework/MDAlgorithms/src/MDNormSCD.cpp
```diff
@@ -363,6 +363,16 @@ void MDNormSCD::cacheDimensionXValues() {
   }
 }
 
+namespace {
+template <typename T, typename BinaryOp>
+void AtomicOp(std::atomic<T> &f, T d, BinaryOp _Op) {
+  T old = f.load();
+  T desired = _Op(old, d);
+  while (!f.compare_exchange_weak(old, desired))
+    desired = _Op(old, d);
```
SimonHeybrock: Comparing this to the example in http://en.cppreference.com/w/cpp/atomic/atomic/compare_exchange:
Why is the loop body needed?

quantumsteve: It's possible that between lines 369 and 371 `f` was updated by another thread. The while loop updates the value of `desired` and tries again.

AndreiSavici (approved):
Works as advertised. On my computer I went from 26 seconds to 6. I checked the output workspaces, and they are the same.
:shipit:

Framework/MDAlgorithms/src/MDNormSCD.cpp
SimonHeybrock: Oh right, it is accumulation. :+1:

PR #19459, opened 2017-04-28
Option to remove edge in PredictPeaks

### LINKED ISSUE #19456: PredictPeaks should have option to not predict on edges

VickieLynch:
Description of work.

**To test:**
```python
LoadEmptyInstrument(Filename='TOPAZ_Definition_2017-02-24.xml', OutputWorkspace='topaz')
LoadIsawUB(InputWorkspace='topaz', Filename='ls5637.mat')
PredictPeaks(InputWorkspace='topaz', WavelengthMin=0.5, WavelengthMax=3.5, MinDSpacing=0.5, OutputWorkspace='topaz_peaks_20edge', EdgePixels=20)
PredictPeaks(InputWorkspace='topaz', WavelengthMin=0.5, WavelengthMax=3.5, MinDSpacing=0.5, OutputWorkspace='topaz_peaks_0edge')
```
Show Instrument for topaz workspace and drag both topaz_peaks_0edge and topaz_peaks_20edge into plot.  Check that edge peaks are not in topaz_peaks_20edge.

Fixes #19456 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Crystal/src/PredictPeaks.cpp
+++ b/Framework/Crystal/src/PredictPeaks.cpp
@@ -11,6 +11,7 @@
 #include "MantidKernel/ListValidator.h"
 #include "MantidKernel/EnabledWhenProperty.h"
 #include "MantidGeometry/Instrument/RectangularDetector.h"
+#include "MantidKernel/BoundedValidator.h"
 
 using Mantid::Kernel::EnabledWhenProperty;
 
@@ -49,7 +50,7 @@ PredictPeaks::PredictPeaks()
 /** Initialize the algorithm's properties.
  */
 void PredictPeaks::init() {
-  declareProperty(make_unique<WorkspaceProperty<Workspace>>(
+  declareProperty(make_unique<WorkspaceProperty<Workspace> >(
                       "InputWorkspace", "", Direction::Input),
                   "An input workspace (MatrixWorkspace, MDEventWorkspace, or "
                   "PeaksWorkspace) containing:\n"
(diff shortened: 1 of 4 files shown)


Framework/Crystal/src/PredictPeaks.cpp
```diff
@@ -489,10 +492,58 @@ void PredictPeaks::calculateQAndAddToOutput(const V3D &hkl,
...
-
+//-----------------------------------------------------------------------------------------
+/**
+  @param  bankName     Name of bank containing peak
+  @param  col          Column number containing peak
+  @param  row          Row number containing peak
+  @param  Edge         Number of edge points for each bank
+  @return True if peak is on edge
+*/
+bool PredictPeaks::edgePixel(std::string bankName, int col, int row, int Edge) {
```
MichaelWedel: I did a search for `edgePixel` in the code base and as far as I can see this is the exact same function as in `SCDCalibratePanels.cpp`. `CentroidPeaks` and `OptimizeLatticeForCellType` contain a slightly different version (it could be just explicit types instead of auto, I have not checked in detail).

The fact that this comes up in four different places suggests that it might be a more general concept that should probably be implemented once in some location where it can be used by all these four algorithms.

I haven't thought about it in detail, but would it be an option to mask the edge pixels and simply check whether a detector is masked? I'm not sure about the best way to implement this masking though. Having an algorithm `MaskEdgePixels` would be one option, it could create a proper mask given some instrument. Can such a mask be reverted easily?

VickieLynch:
There is an algorithm, MaskBTP, that does what you want in python, but python child algorithms do not run in the unit tests.

VickieLynch:
Over the weekend I got the same request  from Feng for FindPeaksMD to not find peaks on the edge of detectors.  Since Xiaoping needs this immediately and Feng has a Matlab script to remove edge peaks, could you pass this and I will use a common function for the now five places we need this when I implement in FindPeaksMD?

Framework/Crystal/src/PredictPeaks.cpp
```diff
@@ -489,10 +492,58 @@ void PredictPeaks::calculateQAndAddToOutput(const V3D &hkl,
   if (m_sfCalculator) {
     p.setIntensity(m_sfCalculator->getFSquared(hkl));
   }
-
+  int edge = this->getProperty("EdgePixels");
```
MichaelWedel: Maybe this could be moved a bit further up - before the matrix is set and before potential calculation of the structure factor, that would save some more calculations for the peaks affected by this check.

MichaelWedel (commented):
Yes, no problem...I guess 5 places make an even stronger argument than 4 :)

I've made one additional comment. It does not change much though.

When @samueljackson92's PR for speeding up `PredictPeaks` is through we should maybe go back to this "inner method" and move out things that don't need to be in there (like for example getting the properties - those won't ever change between two peaks).

I'll check with the test script you provided.

Framework/Crystal/src/PredictPeaks.cpp
MichaelWedel: Ah, sorry, bad choice of line for this comment - I mean the entire check & early return.

Framework/Crystal/src/PredictPeaks.cpp
```diff
@@ -125,6 +126,8 @@ void PredictPeaks::init() {
                   "Use an extended detector space (if defined for the"
                   " instrument) to predict peaks which do not fall onto any"
                   "detector. This may produce a very high number of results.");
+  declareProperty("EdgePixels", 0,
```
MichaelWedel: Values less than 0 are simply ignored now. Maybe a validator could be useful here, but depending on how urgent this PR is it can also be added at a later point.

MichaelWedel:
The test script works as described. Pending the build servers and opening a ticket to refactor the `edgePixel` function I'd say :shipit:

VickieLynch:
Refactor will be done in # #19461

PR #19462, opened 2017-05-01
Refactor MDNormSCD

quantumsteve:
Description of work.

I replaced temporary variables that allocate memory on the heap in each iteration with thread local variables that are reused. The largest gains are in

`MDNormSCD::calculateIntersections` : Reuse the intersections array and use `std::array<double,4>` to minimize additional allocations during sorting. 

Adding method `Matrix<T>::multiplyPoint`, which avoids allocating `std::vector<T>` for each point.

I tried running the same script as #19451 on ndav3, and got 13-17 seconds vs 17-25 seconds with nightly.

I also noticed that the lock in `MatrixWorkspace::flagMasked` is causing some overhead. `std::map::operator[]` should do the same thing but avoid an additional binary search.

**To test:**

Try this on one or more inputs and verify that it
- Shows a performance improvement
- Still gives the correct result

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/src/MatrixWorkspace.cpp
+++ b/Framework/API/src/MatrixWorkspace.cpp
@@ -1027,11 +1027,7 @@ void MatrixWorkspace::flagMasked(const size_t &index, const size_t &binIndex,
     // First get a reference to the list for this spectrum (or create a new
     // list)
     MaskList &binList = m_masks[index];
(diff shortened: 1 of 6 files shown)


AndreiSavici (approved):
Code review looks fine. :shipit:

PR #19467, opened 2017-05-02
boost::serialization build test

SimonHeybrock:
This is just a test for dependencies on all build systems, not intended for merging!


SimonHeybrock:
All builds pass now (unstable due to unused test function that I added, but compilation and linking are ok), so we have no more need for this.

PR #19468, opened 2017-05-02
Clean up Infos in ExperimentInfo

### LINKED ISSUE #19441: Clean up of API::Infos in ExperimentInfo
Follow on from Issue #19385 & PR #19403. Some things were not addressed there that should be dealt with.

**Mandatory**

* Remove check for `API::DetectorInfo` in `ExperimentInfo::getInstrument`
* Add default constructor for `API::ComponentInfo` and use in `ExperimentInfo`. Make similar to `API::DetectorInfo` usage
* Remove the need to construct `DetectorIDs` from `Instrument` twice in `API::DetectorInfo`
* Remove old constructor from `API::DetectorInfo` whereby the map is constructed internally. This is no longer used.

**If Possible**

* Use `InfoComponentVisitor` to create the detector id map instead of getting the information from `Instrument`.

OwenArnold:
Fix issues highlighted in #19441.

We need to determine if these changes should make it to master prior to release 3.10

**To test:**

Code review. Ensure that performance is within acceptable limits.

Fixes #19441

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Geometry/inc/MantidGeometry/Instrument/ParameterMap.h
```diff
@@ -386,9 +381,6 @@ class MANTID_GEOMETRY_DLL ParameterMap {
   /// Pointer to the DetectorInfo object. NULL unless the instrument is
   /// associated with an ExperimentInfo object.
   boost::shared_ptr<const Beamline::DetectorInfo> m_detectorInfo{nullptr};
-  /// Pointer to the ComponentInfo object. NULL unless the instrument is
-  /// associated with an ExperimentInfo object.
-  boost::shared_ptr<const Beamline::ComponentInfo> m_componentInfo{nullptr};
```
SimonHeybrock: With this gone, how are you going to implement `Component::getPos()` to forward to `ComponentInfo`?

SimonHeybrock (commented):
Looks like a really nice cleanup! Some small comments:

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -277,9 +277,9 @@ std::unique_ptr<Geometry::InfoComponentVisitor>
 ExperimentInfo::makeOrRetrieveVisitor(const Instrument &instrument) const {
 
   if (!instrument.hasInfoVisitor() || instrument.isEmptyInstrument()) {
-    const auto nDetectors =
-        instrument.getNumberDetectors(false /*do not skip monitors*/);
-    auto visitor = Kernel::make_unique<InfoComponentVisitor>(nDetectors);
+    const auto nDetectors = instrument.getNumberDetectors();
```
SimonHeybrock: Unused?

Framework/Geometry/src/Instrument/InfoComponentVisitor.cpp
```diff
@@ -81,7 +95,18 @@ void InfoComponentVisitor::registerGenericComponent(
...
+  size_t detectorIndex = 0;
+  try {
+    detectorIndex = m_detectorIdToIndexMap->at(detector.getID());
+  } catch (std::out_of_range &) {
+    /*
+     Do not register a detector with an invalid id. if we can't determine
+     the index, we cannot register it in the right place!
+    */
+
+    throw std::runtime_error("Duplicate detector IDs!");
```
SimonHeybrock: Not sure I understand this error message. Isn't the check above failing only if the ID is *not* found?

SimonHeybrock (commented):
Minor comments:

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -70,11 +70,16 @@ ExperimentInfo::ExperimentInfo()
...
-      m_componentInfo(boost::make_shared<Beamline::ComponentInfo>()) {
+      m_componentInfo(boost::make_shared<Beamline::ComponentInfo>()),
+      m_infoVisitor(Kernel::make_unique<InfoComponentVisitor>(
+          sptr_instrument->getNumberDetectors())) {
+  auto parInstrument = makeParameterizedInstrument();
   m_parmap->setDetectorInfo(m_detectorInfo);
-  m_parmap->setComponentInfo(m_componentInfo);
   m_detectorInfoWrapper = Kernel::make_unique<DetectorInfo>(
-      *m_detectorInfo, getInstrument(), m_parmap.get());
+      *m_detectorInfo, sptr_instrument, m_infoVisitor->detectorIds(),
```
OwenArnold: This is an empty instrument. No detectors or any other sub-components. Why would we need this to be Parameterised? There may be reasons, I'm just not clear what they would be.

SimonHeybrock: True, it seems to be irrelevant.

Framework/Geometry/inc/MantidGeometry/Instrument/ParameterMap.h
OwenArnold: I will need it back. I just don't need it yet. I'd rather keep this PR a moderate size without any dangling features that are unused.

Framework/API/test/ExperimentInfoTest.h
```diff
@@ -901,14 +907,9 @@ class ExperimentInfoTest : public CxxTest::TestSuite {
...
-    const Mantid::API::ComponentInfo &compInfo = expInfo.componentInfo();
-
-    TSM_ASSERT_EQUALS("Should be a valid component index", 0,
-                      compInfo.indexOf(det1->getComponentID()));
-    TSM_ASSERT_THROWS("Should throw. Duplicate should have been rejected",
-                      compInfo.indexOf(det2->getComponentID()),
-                      std::out_of_range &);
+    TSM_ASSERT_THROWS(
+        "Component count and detector count will be out of sync. Should throw.",
+        expInfo.setInstrument(instrument), std::invalid_argument &);
```
OwenArnold: Good point. But I'm going to leave this for now. I need to explore how strict we can be about dealing with this kind of error condition. I'm going to lean on the system tests as part of this PR to get some kind of indication.

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -302,31 +304,17 @@ void ExperimentInfo::setInstrument(const Instrument_const_sptr &instr) {
   }
   const auto parInstrument = Geometry::ParComponentFactory::createInstrument(
       sptr_instrument, m_parmap);
+
+  m_infoVisitor = makeOrRetrieveVisitor(*sptr_instrument);
```
SimonHeybrock: @OwenArnold Looking at this again: Why does the visitor ever need to live outside the `Instrument`? Couldn't it simply be created internally in `Instrument`? That would avoid messing around with the visitor in `ExperimentInfo` (apart from passing a reference to the `ComponentInfo` factory function. That might make the code in `ExperimentInfo` a bit easier to understand.

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -335,29 +323,22 @@ void ExperimentInfo::setInstrument(const Instrument_const_sptr &instr) {
...
-  }
-  // We can only set the ComponentInfo on the return Instrument if the API
-  // wrapper is fully constructed
-  if (m_componentInfoWrapper) {
-    instrument->setComponentInfo(
-        m_componentInfo, m_componentInfoWrapper->componentIds(),
-        m_componentInfoWrapper->componentIdToIndexMap());
-  }
+  auto instrument = makeParameterizedInstrument();
+  instrument->setDetectorInfo(m_detectorInfo);
```
SimonHeybrock: I suppose we could eventually also put `DetectorInfo` into the visitor (similar to what you did with `ComponentInfo`)? I think it is good though to not do this in this PR but keep it for the next one.

SimonHeybrock (commented):
Had another look through this after the latest changes. Looks like a good cleanup, but probably this is not the last step yet. I have two comments, but I would also be happy if those were addressed in a follow up PR, in case you do want to get this merged:

Framework/API/src/ExperimentInfo.cpp
OwenArnold: Yes. I deliberately kept it out for the time being. But yes, all this information could be collected. Ideally, it would have just two construction related public functions (others are required due to `ComponentVisitor`).

```cpp
class InfoComponentVisitor : public ComponentVisitor {
public:
.....
std::unique<API::ComponentInfo> componentInfo() const;
std::unique<API::DetectorInfo> detectorinfo() const;
};
```

Framework/API/src/ExperimentInfo.cpp
OwenArnold: @SimonHeybrock Interesting... yes I agree. I think to get the full benefit from this we would need to have the `InfoComponentVisitor` capable of fully constructing `DetectorInfo`, otherwise we are back to exposing all the methods via `Instrument` to get/set the information for the `DetectorInfo`. So basically, I think this is a good idea, but we can't do this properly now.

OwenArnold:
@SimonHeybrock. Given the proximity to the release, myself and @LamarMoore will monitor the effect on performance of these changes to ensure there are not any performance edge cases making it into the official release.

SimonHeybrock:
Comments regarding further cleanup will be addressed in the future =>
:shipit:

VickieLynch (commented):
Nightly:
LoadMD-[Notice] LoadMD started
LoadMD-[Notice] LoadMD successful, Duration 1 minutes 29.47 seconds
BinMD-[Notice] BinMD started
BinMD-[Notice] BinMD successful, Duration 7.10 seconds
#19468 
LoadMD started
LoadMD successful, Duration 1 minutes 42.36 seconds
BinMD started
BinMD successful, Duration 28.36 seconds

DO NOT MERGE

VickieLynch (changes requested):
Need to improve performance for TOPAZ data

Framework/API/src/ExperimentInfo.cpp
SimonHeybrock: Could this alleviate the performance regression that @VickieLynch observed? Currently, if the same instrument without visitor is set on multiple `ExperimentInfos`, `Instrument::registerContents` will be called multiple times?

On the other hand, what you are doing right now does not seem to do more work than what we had before, so I do not understand why there is a slowdown?

martyngigg:
I assume this should go back to `3.11`?

VickieLynch:
Yes, 3.11 unless someone has time to improve performance.

OwenArnold:
@VickieLynch this can go to 3.11 i have improved the performance based on your review. Retry this and you should find faster than  => master in speed.

SimonHeybrock:
I hd also reviewed the latest changes, which look good to me and I would have expected better performance from just looking at the code, as confirmed by Owen's performance testing.
:shipit: once Vickie confirms this.

VickieLynch:
Testing now

VickieLynch (approved):

VickieLynch:
This is back to 7 seconds, so OK to merge in 3.10

PR #19481, opened 2017-05-04
Validate spectra list is valid in several algs

### LINKED ISSUE #19289: Mask Detectors Fails with Invalid Unordered_map key
### Expected behavior
Masking detectors with a spectrum list (e.g. `0-58`) should work
### Actual behavior
Algorithm fails with
```
Error in execution of algorithm MaskDetectors:
invalid unordered_map<K, T> key
```

Following script:
```
CreateSampleWorkspace(OutputWorkspace='test', Function='Flat background', NumBanks=1, XMax=2, BinWidth=1, BankPixelWidth=10)
MaskDetectors(Workspace="test", SpectraList=range(0, 51))
```

### Platforms affected
All

DavidFair:
Description of work.
This PR resolves an issue where several algorithms did not have validation for their detector/spectra list inputs. It adds validation to these algorithms and fixes any tests that fail because of these changes.

If there are any algorithms I have missed please let me know and I'll check / add validators to them as well.

**To test:**
Code review
Open a changed algorithm such as `MaskDetectors` ensure SpectraList only allows input >= 1 and DetectorList >= 0 and a sane error message is given

Fixes #19289 .

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/PythonInterface/plugins/algorithms/IndirectTransmission.py
+++ b/Framework/PythonInterface/plugins/algorithms/IndirectTransmission.py
@@ -173,7 +173,7 @@ class IndirectTransmission(PythonAlgorithm):
         except ValueError:
             # If that fails then get it by taking from group of all detectors
             wsHandle = mtd[workspace]
-            spectra_list = list(range(0, wsHandle.getNumberHistograms()))
+            spectra_list = list(range(1, wsHandle.getNumberHistograms() + 1))
             GroupDetectors(InputWorkspace=workspace,
                            OutputWorkspace=workspace,
                            SpectraList=spectra_list)
--- a/Framework/PythonInterface/plugins/algorithms/LoadSINQFile.py
+++ b/Framework/PythonInterface/plugins/algorithms/LoadSINQFile.py
@@ -68,8 +68,7 @@ class LoadSINQFile(PythonAlgorithm):
 
         if inst == "POLDI":
             if ws.getNumberHistograms() == 800:
-                ws.maskDetectors(SpectraList=list(range(0,800))[::2])
-
+                ws.maskDetectors(SpectraList=list(range(2, 800, 2)))
                 config.appendDataSearchDir(config['groupingFiles.directory'])
                 grp_file = "POLDI_Grouping_800to400.xml"
                 ws = mantid.simpleapi.GroupDetectors(InputWorkspace=ws,
--- a/Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/TOSCABankCorrection.py
+++ b/Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/TOSCABankCorrection.py
@@ -218,7 +218,7 @@ class TOSCABankCorrection(DataProcessorAlgorithm):
         # Recalculate the sum spectra from corrected banks
         GroupDetectors(InputWorkspace=self._output_ws,
                        OutputWorkspace='__sum',
-                       SpectraList=[0,1],
+                       SpectraList=[1,2],
                        Behaviour='Average')
 
         # Add the new sum spectra to the output workspace
--- a/scripts/Diffraction/isis_powder/polaris_routines/polaris_algs.py
+++ b/scripts/Diffraction/isis_powder/polaris_routines/polaris_algs.py
@@ -9,7 +9,7 @@ from isis_powder.polaris_routines import polaris_advanced_config
 
 
 def calculate_absorb_corrections(ws_to_correct, multiple_scattering):
-    mantid.MaskDetectors(ws_to_correct, SpectraList=list(range(0, 55)))
+    mantid.MaskDetectors(ws_to_correct, SpectraList=list(range(1, 55)))
 
     absorb_dict = polaris_advanced_config.absorption_correction_params
     ws_to_correct = absorb_corrections.run_cylinder_absorb_corrections(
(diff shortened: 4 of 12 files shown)


scripts/Diffraction/isis_powder/polaris_routines/polaris_algs.py
```diff
@@ -9,7 +9,7 @@
 
 
 def calculate_absorb_corrections(ws_to_correct, multiple_scattering):
-    mantid.MaskDetectors(ws_to_correct, SpectraList=list(range(0, 55)))
+    mantid.MaskDetectors(ws_to_correct, SpectraList=list(range(1, 55)))
```
DTasev: Is this supposed to cover less range after the change? Before it was 0->55 (length 54), now it's 1->55 (length 53)?

DavidFair: This is correct. I had already correctly taken into account the fact the last invalid spectra was detector 54 however failed to notice they start a 1 not 0 originally.

DTasev:
Everything looks good, all of the unit tests pass locally, I'm happy to :shipit:

martyngigg:
I completely agree with the sentiment here but I'm not sure on the validation for `SpectraList` & `DetectorList`. Spectra starting at `1` is more of an ISIS tradition that is generally followed but the number could well be set to `0` and these algorithms would now not work. There are also cases where some detectors have negative IDs and this would now not work with these changes.

I think we may have to leave the out the `DetectorList` validation but we are probably safe to change the `SpectrumList` validation to be `>=0`.

Framework/PythonInterface/plugins/algorithms/IndirectTransmission.py
```diff
@@ -173,7 +173,7 @@ def _get_efixed(self, workspace):
         except ValueError:
             # If that fails then get it by taking from group of all detectors
             wsHandle = mtd[workspace]
-            spectra_list = list(range(0, wsHandle.getNumberHistograms()))
+            spectra_list = list(range(1, wsHandle.getNumberHistograms() + 1))
```
martyngigg: Doesn't this change the behaviour?

Framework/PythonInterface/plugins/algorithms/LoadSINQFile.py
```diff
@@ -68,8 +68,7 @@ def PyExec(self):
 
         if inst == "POLDI":
             if ws.getNumberHistograms() == 800:
-                ws.maskDetectors(SpectraList=list(range(0,800))[::2])
-
+                ws.maskDetectors(SpectraList=list(range(2, 800, 2)))
```
martyngigg: Again, doesn't this change behaviour?

martyngigg (changes requested):

Framework/PythonInterface/plugins/algorithms/LoadSINQFile.py
DavidFair: In this case no, as it attempted to mask spectra `0, 2, 4...` `MaskDetectors` silently skipped spectra 0, it was only because we emitted a key error I noticed these problems.

Framework/PythonInterface/plugins/algorithms/IndirectTransmission.py
DavidFair: In this case yes, I was going on the comment that it takes a group of *all* detectors whereas previously it only did `1...n-1` detectors. For that reason I changed it to end use all the spectra which may change behaviour slightly.

DavidFair:
@martyngigg 
The only gotcha with setting spectra list validation to `>= 1` is many of the algorithms then assume the spectra starts at `0` and subsequently fail if that assumption doesn't hold true such as on ISIS data sets. I'm not sure where to go forward with this unless we check the spectra at execution rather that letting the validators handle it?

I agree with the changes to the detector IDs, I'll remove the validators for them as it doesn't make sense to constrain those.

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/TOSCABankCorrection.py
```diff
@@ -218,7 +218,7 @@ def _apply_correction(self, bank_1_sf, bank_2_sf):
         # Recalculate the sum spectra from corrected banks
         GroupDetectors(InputWorkspace=self._output_ws,
                        OutputWorkspace='__sum',
-                       SpectraList=[0,1],
+                       SpectraList=[1,2],
```
martyngigg: Same question on change of behaviour here.

Framework/PythonInterface/plugins/algorithms/IndirectTransmission.py
martyngigg: Ah yes I see. I think in this case it would then be better to just use SumSpectra without any arguments to sum everything. @louisemccann Do you think this makes sense?

martyngigg:
@DavidFair For a given workspace the spectrum number & detector ID can start wherever they like. For example if you load a file and then crop the workspace starting at spectrum `100` the first one is `100` so no algorithm should assume they start anywhere really and I think any validation would have to be done at runtime to confirm the spectra requested are actually in the workspace.

DavidFair:
Ok, in that case I'll go ahead and tweak the algorithms so that we check the starting spectrum number and give a useful error message rather than a key errors. 
As this validation is pretty generic and applicable to any algorithms that accept a spectra list would it be worth putting somewhere separate rather the implementing it within each algorithm? If so, where would be the best place to put it?

DavidFair:
I'm going to close this PR as the branch requires some additional work. Talking with Martyn it makes sense to see if we can tie these changes in with the IndexInfo development.

PR #19496, opened 2017-05-05
Fix ISIS Powder Instrument Settings test

DavidFair:
Description of work.
The unit test for ISIS Powder instrument settings previously checked that all valid enum values are printed when a user forget to set the property. This relies on the fact the `__dict__` object in Python 2 was ordered. Currently this causes the Python 3 tests to fail randomly on master. 

Instead we should test for only one of the values being printed which no longer rely on the order being consistent. 

**To test:**
Run `ISISPowderInstrumentSettingsTest` multiple times using a Python 3 build. If it does not fail say after 10 times the random order of the dict will have been triggered at least once.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


martyngigg (approved):
Changes make sense. It now doesn't rely on a particular order of entries in the dictionary.

:shipit:

martyngigg:
@mantidproject/gatekeepers Can we merge this ASAP? It fixes the Python 3 tests that are frequently failing on pull requests.

PR #19510, opened 2017-05-09
MPI support: Consolidation and first examples

### LINKED ISSUE #19491: Refine MPI mechanism and implement first examples
As a folow up to https://github.com/mantidproject/mantid/pull/19477:

- Implement MPI support for `CreateWorkspace` and a few other simple algorithms such that the concept can be tested.
- Show how it can be used from Python.
- Cleanup and refine MPI mechanism based on first real developing experience.

SimonHeybrock:
This is a follow-up to #19280. Some of the concepts are cleaned up and extended slightly.

- Added detailed documentation page for developers.
- Added support for MPI in two algorithms (`CreateWorkspace` and `Rebin`).
- Cleanup-up mechanism for testing MPI support in algorithms.

**To test:**

Code review.

Fixes #19491.

No release notes, MPI should not yet be part of our publicly supported features.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/API/inc/MantidAPI/Workspace.h
```diff
@@ -117,7 +117,7 @@ class MANTID_API_DLL Workspace : public Kernel::DataItem {
   /// The history of the workspace, algorithm and environment
   std::unique_ptr<WorkspaceHistory> m_history;
   /// Storage mode of the Workspace (used for MPI runs)
-  Parallel::StorageMode m_storageMode{Parallel::StorageMode::Cloned};
+  Parallel::StorageMode m_storageMode{Parallel::StorageMode::MasterOnly};
```
SimonHeybrock: I changed this such that workspaces created in `execNonMaster` get the right storage mode by default. For properly initialized workspaces (unlike those on non-master ranks if execution mode is `MasterOnly`), the storage mode is set via `IndexInfo` that is passed to `create<MatrixWorkspace>`.

Open problem/question:
- How should the storage mode of non-`MatrixWorkspaces` be set? I do not want to provide a public `setStorageMode`, since it cannot change after construction. Probably using a factory function or method?

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -642,6 +657,10 @@ double MatrixWorkspace::getXMax() const {
 }
 
 void MatrixWorkspace::getXMinMax(double &xmin, double &xmax) const {
+  if (m_indexInfo->communicator().size() > 1)
+    throw std::runtime_error(
+        "MatrixWorkspace: Parallel support for XMin and XMax not implemented.");
```
SimonHeybrock: Would not be hard to do this, but for the time being I am not doing this. The main issue here is that this cannot be supported unless a workspace always has a `Parallel::Communicator`. Currently `MatrixWorkspace` does, via `IndexInfo`, so this would be fine but maybe this requires more thought?

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -2016,6 +2035,12 @@ void MatrixWorkspace::updateCachedDetectorGrouping(const size_t index) const {
 void MatrixWorkspace::buildDefaultSpectrumDefinitions() {
   const auto &detInfo = detectorInfo();
   size_t numberOfDetectors{detInfo.size()};
+  if (numberOfDetectors == 0) {
+    // Default to empty spectrum definitions if there is no instrument.
+    m_indexInfo->setSpectrumDefinitions(
+        std::vector<SpectrumDefinition>(m_indexInfo->size()));
+    return;
+  }
```
SimonHeybrock: This is not related to MPI, it simply makes creating workspaces with `IndexInfo` simpler. Previously we were forced to manually set a vector of empty `SpectrumDefinition`, otherwise the workspaces would reject the `IndexInfo`.

My new policity here is: We build defaul spectrum definitions *whenever there is a unique way to do this*, i.e., 1:1 mapping, or all empty. Can anyone think of a reason why this would be bad?

Framework/API/test/AlgorithmMPITest.h
```diff
@@ -533,51 +533,41 @@ class AlgorithmMPITest : public CxxTest::TestSuite {
   AlgorithmMPITest() { AnalysisDataService::Instance(); }
 
   void testNoParallelism() {
-    runNoParallelism(Parallel::Communicator{});
```
SimonHeybrock: See comment in `ParallelRunner`.

Framework/Algorithms/src/CreateWorkspace.cpp
```diff
@@ -61,6 +69,17 @@ void CreateWorkspace::init() {
...
+#ifdef MPI_EXPERIMENTAL
+  std::vector<std::string> propOptions{
+      Parallel::toString(Parallel::StorageMode::Cloned),
+      Parallel::toString(Parallel::StorageMode::Distributed),
+      Parallel::toString(Parallel::StorageMode::MasterOnly)};
+  declareProperty(
+      "ParallelStorageMode",
+      Parallel::toString(Parallel::StorageMode::Distributed),
+      boost::make_shared<StringListValidator>(propOptions),
+      "The parallel storage mode of the output workspace for MPI builds");
```
SimonHeybrock: Probably there is a nicer way to do this, but for now this seems sufficient?

Framework/Algorithms/src/CreateWorkspace.cpp
```diff
@@ -144,16 +161,15 @@ void CreateWorkspace::exec() {
   }
 
   // Create the OutputWorkspace
+  auto storageMode = Parallel::fromString(getProperty("ParallelStorageMode"));
+  IndexInfo indexInfo(nSpec, storageMode, communicator());
```
SimonHeybrock: I really dislike the `communicator()` argument: This should always be the communicator held by the algorithm, so passing it is just repetitive boild plate, however I could not think of a good way to eliminate this. One option might be to make `Algorithm` a factory for `IndexInfo`, but I think that violates some other design principles.

Ideas or comments?

Framework/Algorithms/src/CreateWorkspace.cpp
```diff
@@ -232,9 +254,6 @@ void CreateWorkspace::exec() {
     }
   }
 
-  // Set distribution flag
-  outputWS->setDistribution(getProperty("Distribution"));
-
   // Set Y Unit label
```
SimonHeybrock: The (unchaged) code for handling the vertical axis below is probably not correct in the MPI case, unless it is simply a spectrum axis. I think this requires some more investigation amongst more algorithms.

Framework/Algorithms/src/Rebin.cpp
```diff
@@ -355,5 +355,10 @@ void Rebin::propagateMasks(API::MatrixWorkspace_const_sptr inputWS,
   }
 }
 
+Parallel::ExecutionMode Rebin::getParallelExecutionMode(
+    const std::map<std::string, Parallel::StorageMode> &storageModes) const {
+  return Parallel::getCorrespondingExecutionMode(storageModes.begin()->second);
```
SimonHeybrock: This is what I had previously done in a new base class `TriviallyParallelAlgorithm`. I probably still want to add a class like this, but for the time being I am postponing it until we have a cleared understanding or some of the issues we will be facing.

Framework/Algorithms/test/CreateWorkspaceTest.h
```diff
@@ -3,10 +3,53 @@
...
+  alg.initialize();
+  std::vector<double> dataEYX(2000);
+  int nspec = 1000;
+  alg.setProperty<int>("NSpec", nspec);
+  alg.setProperty<std::vector<double>>("DataX", dataEYX);
+  alg.setProperty<std::vector<double>>("DataY", dataEYX);
+  alg.setProperty<std::vector<double>>("DataE", dataEYX);
+  // In a true MPI run we could simply use "out", but in the threaded
+  // fake-runner we have only one ADS, so we have to avoid clashes.
+  std::string outName("out" + std::to_string(comm.rank()));
```
SimonHeybrock: Code like this now shows up in every unit test of parallel execution. I would really like to avoid this. I have considered running the algorithms as a child algorithm since then I do not have to set a name myself. However, we then need a parent algorithm.

Solutions?
- Create a test helper `ParallelAlgorithmFactory`, which holds a parent algorithm and can spawn the testee as a child.
- other?

Framework/TestHelpers/inc/MantidTestHelpers/ParallelRunner.h
```diff
@@ -57,6 +57,9 @@ void ParallelRunner::run(Function &&f, Args &&... args) {
     Mantid::Parallel::Communicator comm;
     f(comm, std::forward<Args>(args)...);
   } else {
+    // 1. Run serial
+    f(Mantid::Parallel::Communicator{}, std::forward<Args>(args)...);
```
SimonHeybrock: Two reasons to add this:
- Running code that is suposed to run in parallel should also work serially.
- Some static property type map in `Kernel` is constructed in a non-thread-safe manner, leading to segfaults unless something was run without threads first. If we do not do this here, *every* unit test must ensure to run something serially first.

SimonHeybrock (commented):
I have commented on some of the issues and questions I have encountered during this first set of algorithms with MPI support. I would like to thake things slowly in the beginning so we can make sure we have our concepts and structure right:

Framework/API/inc/MantidAPI/Workspace.h
LamarMoore: Constructor argument?

Framework/Algorithms/src/CreateWorkspace.cpp
```diff
@@ -144,16 +161,15 @@ void CreateWorkspace::exec() {
   }
 
   // Create the OutputWorkspace
+  auto storageMode = Parallel::fromString(getProperty("ParallelStorageMode"));
```
LamarMoore: Shouldn't this be wrapped in `#ifdef MPIExperimental` as well?

Framework/Algorithms/src/CreateWorkspace.cpp
LamarMoore: I guess an alternative is going through the property mechanism, however I think you run into the same design problems. For example, if you overloaded `declareProperty` within `Algorithm` to accept handle the special case of a WorkspaceProperty and a storage mode. Then you could create the indexInfo behind the scenes. 

Alternatively, you can specify a storage mode in the WorkspaceProperty itself.

Framework/API/inc/MantidAPI/Workspace.h
SimonHeybrock: Maybe, though we have a wide variety of workspaces, with various other constructor arguments, and some are created by factories.

Framework/Algorithms/src/CreateWorkspace.cpp
SimonHeybrock: Hmm, that is an interesting idea!

```cpp
alg.setProperty("OutputWorkspace", "ws", StorageMode::Distributed);
```

I think I kind of like that, but we need to think about it more. Implementing this might mean some significant effort. Also, it is not clear how the communicator would be injected? How do you actually create the `IndexInfo`? You simply query the output workspace property for the storage mode? But then how is this better then having a separate property?

Framework/Algorithms/src/CreateWorkspace.cpp
SimonHeybrock: Actually, I probably want to remove the `#ifdef` on the property -- otherwise I cannot run the new unit tests in a non-MPI build

Framework/DataObjects/test/WorkspaceCreationTest.h
```diff
@@ -387,6 +408,31 @@ class WorkspaceCreationTest : public CxxTest::TestSuite {
     check_zeroed_data(*ws);
   }
 
+  void test_default_StorageMode_is_Cloned() {
+    IndexInfo indices(2);
+    indices.setSpectrumDefinitions(std::vector<SpectrumDefinition>(2));
+    TS_ASSERT_EQUALS(
+        create<Workspace2D>(indices, BinEdges{1, 2, 4})->storageMode(),
```
OwenArnold: Good. I think this is important to check, particularly if someone starts changing what the defaults are on the Workspace subtypes.

docs/source/development/AlgorithmMPISupport.rst
```diff
@@ -0,0 +1,464 @@
...
+  ws = Rebin(dataWS, "1,1,7");
+
+  print("Histograms: " + str(ws.getNumberHistograms()))
+  for i in range(ws.getNumberHistograms()):
+      print("(Local) workspace index: " + str(i))
+      print(ws.readX(i))
+      print(ws.readY(i))
+
+
+Run Python with ``mpirun`` and the desired number of MPI ranks:
```
OwenArnold: This could be reworded I think. You first say that it can be run with python given the commands and then go on to say that it's not possible to run the commands. Maybe we need to explain what else is required.

docs/source/development/AlgorithmMPISupport.rst
```diff
@@ -0,0 +1,464 @@
...
+This index is represented by ``Indexing::GlobalSpectrumIndex``. [#spectrum-index]_
+
+The consequences are as follows:
+
+- Workspace indices should not be logged or written into output of other types such as tables. Instead spectrum numbers (``Indexing::SpectrumNumber``) or global spectrum indices (``Indexing::GlobalSpectrumIndex``) must be used.
+- The number of histograms in a workspace obtained from ``MatrixWorkspace::getNumberHistograms()`` may only be used for processing all spectra.
+  It should not be logged, written as output, or used for branching execution paths since it is meaningless.
+  If the total number of spectra in a workspace is required it can be accessed via ``MatrixWorkspace::indexInfo()::globalSize()``.
+- User input providing indices or spectrum numbers in way or another must be translated into local indices by ``IndexInfo``.
+  The most common cases will be covered by a workspace property that also accepts indices (under development).
```
OwenArnold: @LamarMoore you should create a placeholder in your work to ensure that this gets updated once you have completed the new property.

docs/source/development/AlgorithmMPISupport.rst
```diff
@@ -0,0 +1,464 @@
...
+
+If a workspace is distributed, i.e., has storage mode ``StorageMode::Distributed`` workspaces indices lose their meaning.
+In particular, ``MatrixWorkspace::getNumberHistograms()`` will return the local number of spectra and not the global size of the workspace.
+For purposes of interaction with the user interface and for internal consistency a global equivalent of the 'workspace index' concept has been introduced.
+This index is represented by ``Indexing::GlobalSpectrumIndex``. [#spectrum-index]_
+
+The consequences are as follows:
+
+- Workspace indices should not be logged or written into output of other types such as tables. Instead spectrum numbers (``Indexing::SpectrumNumber``) or global spectrum indices (``Indexing::GlobalSpectrumIndex``) must be used.
+- The number of histograms in a workspace obtained from ``MatrixWorkspace::getNumberHistograms()`` may only be used for processing all spectra.
```
OwenArnold: "may only be used for processing all spectra.".. don't we need to qualify this by saying "may only be used for processing all spectra on the local rank"

docs/source/development/AlgorithmMPISupport.rst
```diff
@@ -0,0 +1,464 @@
...
+
+.. rubric:: Footnotes
+
+.. [#split-instrument] The complexity and overhead of splitting the instrument, in particular given the overhead ensuing from handling all cases exemplified above, led to the decision split only the neutron data based on spectra, but not detectors.
+
+.. [#spectrum-index] Some will argue that this should be ``GlobalWorkspaceIndex``.
+  However it is not an index of a workspace so the term ``GlobalSpectrumIndex`` has been chosen for clarity.
+  On the user interface side this will still be named 'workspace index', dropping the 'global' since the distinction between global and local indices is irrelevant for users.
+
+.. categories:: Development
```
OwenArnold: This document is a nice addition. We need to keep this up to date with future MPI work. Not sure what the best strategy for the reminder will be.

Framework/TestHelpers/inc/MantidTestHelpers/ParallelAlgorithmCreation.h
```diff
@@ -0,0 +1,45 @@
...
+template <class T>
+std::unique_ptr<T> create(const Mantid::Parallel::Communicator &comm) {
+  using namespace Mantid;
+  auto alg = Kernel::make_unique<T>();
+  alg->setChild(true);
+  alg->setCommunicator(comm);
+  alg->initialize();
+  const auto &props = alg->getProperties();
+  for (auto prop : props) {
+    auto wsProp = dynamic_cast<API::IWorkspaceProperty *>(prop);
```
OwenArnold: You could eliminate the named `wsProp` variable here. But I think it does help readability.

Framework/Algorithms/src/CreateWorkspace.cpp
```diff
@@ -154,39 +172,51 @@ void CreateWorkspace::exec() {
...
+    const auto localIndices =
+        indexInfo.makeIndexSet({static_cast<GlobalSpectrumIndex>(i)});
+    if (localIndices.empty())
+      continue;
+
     const std::vector<double>::difference_type xStart = i * xSize;
     const std::vector<double>::difference_type xEnd = xStart + xSize;
     const std::vector<double>::difference_type yStart = i * ySize;
     const std::vector<double>::difference_type yEnd = yStart + ySize;
+    auto local_i = localIndices[0];
```
OwenArnold: Shouldn't we be enumerating over this rather than just using the first one?

OwenArnold (changes requested):

Framework/API/src/Algorithm.cpp
```diff
@@ -1545,6 +1546,34 @@ void Algorithm::cancel() {
...
+ *
+ * Example value=1 will turn warning into notice
+ * Example value=-1 will turn notice into warning
+ */
+void Algorithm::setLoggingOffset(const int value) {
+  if (m_communicator->rank() == 0)
+    g_log.setLevelOffset(value);
+  else {
+    int offset{1};
+    ConfigService::Instance().getValue("mpi.loggingOffset", offset);
```
SimonHeybrock: This is (currently?) used only internally here so I do not see a reason to extend the interface of `ConfigService`.

Framework/API/src/Algorithm.cpp
```diff
@@ -1678,7 +1710,16 @@ void Algorithm::execNonMaster() {
         // This is the reverse cast of what is done in
         // cacheWorkspaceProperties(), so it should never fail.
         const Property &prop = dynamic_cast<Property &>(*wsProp);
-        setProperty(prop.name(), ws->cloneEmpty());
+        auto clone = ws->cloneEmpty();
+        // Currently we have not implemented a proper cloneEmpty() for all
+        // workspace types, in particular the abundance of Workspace2D subtypes,
+        // so we do a safety check here.
```
SimonHeybrock: The implementation is actually really trivial, I simply chose not to do it since as long as we have no algorithms with MPI support that use, say, `GroupingWorkspace` there is no reason to do so.

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -642,6 +656,10 @@ double MatrixWorkspace::getXMax() const {
 }
 
 void MatrixWorkspace::getXMinMax(double &xmin, double &xmax) const {
+  if (m_indexInfo->size() != m_indexInfo->globalSize())
```
SimonHeybrock: Yes for `Rebin`, but of course it is relevant for all code using this method.

Framework/API/test/MatrixWorkspaceTest.h
```diff
@@ -70,6 +71,29 @@ boost::shared_ptr<MatrixWorkspace> makeWorkspaceWithDetectors(size_t numSpectra,
...
+  for (const auto storageMode : {StorageMode::MasterOnly, StorageMode::Cloned,
+                                 StorageMode::Distributed}) {
+    WorkspaceTester ws;
+    if (comm.rank() == 0 || storageMode != StorageMode::MasterOnly) {
+      Indexing::IndexInfo indexInfo(1000, storageMode, comm);
+      ws.initialize(indexInfo,
+                    HistogramData::Histogram(HistogramData::Points(1)));
+    }
+    if (storageMode == StorageMode::Distributed && comm.size() > 1) {
+      TS_ASSERT_THROWS(ws.getSpectrum(0).setSpectrumNo(42), std::logic_error);
```
SimonHeybrock: Extended the assert to include the exception message, should be clearer now.

Framework/Algorithms/src/CreateWorkspace.cpp
SimonHeybrock: No, since there is exactly 1 global index `i`. Locally the result is thus either empty (and we hit `continue` above) or of size 1.

OwenArnold: 👍

OwenArnold (approved):
Pending builds, but changes look good.

Framework/Parallel/src/StorageMode.cpp
```diff
@@ -25,5 +25,18 @@ std::string toString(const std::map<std::string, StorageMode> &map) {
   return ret;
 }
 
+/// Returns a StorageMode for a human-readable string representation.
+StorageMode fromString(const std::string &mode) {
+  if (mode == "Parallel::StorageMode::Cloned")
```
LamarMoore: Is there a reason you use, for example, `Parallel::StorageMode::Cloned` instead of just `Cloned`?

Framework/TestHelpers/inc/MantidTestHelpers/ParallelAlgorithmCreation.h
LamarMoore: I think clang (or cppcheck) might complain about this.

LamarMoore (commented):
Looks good, just a question about the `Parallel` string conversions.

Framework/Parallel/src/StorageMode.cpp
SimonHeybrock: Not sure about the `Parallel::` part, but `StorageMode::` is included to distinguish it from `ExecutionMode` (similar to having `StorageMode` and `ExecutionMode` as strongly typed enums).

LamarMoore: 👍

LamarMoore:
@OwenArnold  Are you happy for me to merge this?

OwenArnold:
@LamarMoore good by me.

PR #19531, opened 2017-05-11
Simplify function Detector::getPhi()

quantumsteve:
Description of work.

This simplifies the expression for `Detector::getPhi()` to a single trigonometric function. It's the same change I suggested in PR #18962.

**To test:**

Review changes and check the build servers.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Geometry/src/Instrument/Detector.cpp
+++ b/Framework/Geometry/src/Instrument/Detector.cpp
@@ -95,10 +95,10 @@ double Detector::getSignedTwoTheta(const V3D &observer, const V3D &axis,
 /// Get the phi angle between the detector with reference to the origin
 ///@return The angle
 double Detector::getPhi() const {
-  double phi = 0.0, dummy;
-  this->getPos().getSpherical(dummy, dummy, phi);
-  return phi * M_PI / 180.0;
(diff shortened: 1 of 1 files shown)


peterfpeterson (approved):
`atan2(y, x)` is the same as the code that was being called before

peterfpeterson:
:shipit:

SimonHeybrock:
Note that this will not be supported in `Instrument-2.0`, due to its ambiguity. Do you want to fix the Doxygen here (personally, I would put in a warning to not use this method)?

peterfpeterson:
@SimonHeybrock Would changing the name to something like `getAzimuthal` work better for you?

SimonHeybrock:
@peterfpeterson Maybe slightly better, but nevertheless I will not add it to `Instrument-2.0`. Conversion to spherical coordinates has nothing to do with detectors of a neutron beamline. Plus, we do not have `getRadius()` and `getPolar()` either, so having only `getAzimuthal()` is inconsistent.

quantumsteve:
@SimonHeybrock I updated the doxygen to encourage developers to not use `Detector::getPhi()` in new code. Changing it further is beyond the scope of this PR.

PR #19536, opened 2017-05-11
FindSXPeaks - Support grouped detectors

### LINKED ISSUE #19503: FindSXPeaks now throws error for instruments with grouped detectors
`FindSXPeaks` algorithm no longer works for instruments such as WISH that have hardware grouped detectors. 

### Expected behavior
Algorithm should work and find peaks

### Actual behavior
During `spectrumInfo` refactoring in #18554 a check was added to the algorithm that requires the mapping between workspace indicies and detectors to be unique. 

Previously this would use averages of the group for values of phi, theta etc., which for WISH is fine as they spatial close together.

Try running `FindSXPeaks` on WISH data

### Platforms affected
All.

DavidFair:
Description of work.
This PR resolves an issue where FindSXPeaks could not supports instruments with multiple detectors per spectra. When we come across multiple detectors we now correctly cast to a `DetectorGroup` and call `getPhi` which get the average `phi` value of all detectors.

Additionally I have added a validator to make sure only TOF data is used as previously it performed no validation.

**To test:**
*You will need minimum 16GB of RAM*
- Load a WISH data file - such as the following
`LoadRaw(Filename='WISH00036238.raw', OutputWorkspace='36238', LoadLogFiles=False)`
- Run `FindSXPeaks` on the workspace
- Open the loaded file in the instrument view, and drag the table workspace onto the view
- Check the peak positions look sane by clicking on some, and the within the peak plot (bottom left) the peak with the highest intensity is labelled.

For the validator:
- Create a sample workspace with units not TOF
- Ensure that the algorithm does not accept this workspace unless it has TOF units. 

Fixes #19503 .

Added to release notes here: [here](https://github.com/mantidproject/mantid/blob/7ee78a6f2fa2585fa73f1e7b0be87162046e1a09/docs/source/release/v3.10.0/diffraction.rst)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Crystal/src/FindSXPeaks.cpp
+++ b/Framework/Crystal/src/FindSXPeaks.cpp
@@ -18,12 +18,12 @@ using namespace Mantid::DataObjects;
 namespace {
 // Anonymous namespace
 using namespace Mantid;
-using wsIndexToDetIds = std::unordered_map<size_t, std::vector<detid_t>>;
+using WsIndexToDetIds = std::unordered_map<size_t, std::vector<detid_t>>;
 
-wsIndexToDetIds mapDetectorsToWsIndexes(const API::DetectorInfo &detectorInfo,
+WsIndexToDetIds mapDetectorsToWsIndexes(const API::DetectorInfo &detectorInfo,
                                         const detid2index_map &mapping) {
   const auto &detectorIds = detectorInfo.detectorIDs();
-  wsIndexToDetIds indexToDetMapping;
+  WsIndexToDetIds indexToDetMapping;
 
   indexToDetMapping.reserve(detectorIds.size());
   for (const auto detectorID : detectorIds) {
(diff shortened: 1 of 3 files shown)


samueljackson92 (dismissed):
Algorithm now works as expected. A couple of minor points from code review.

Framework/Crystal/src/FindSXPeaks.cpp
```diff
@@ -152,26 +206,27 @@ void FindSXPeaks::exec() {
...
-    // If no detector found, skip onto the next spectrum
-    if (!spectrumInfo.hasDetectors(static_cast<size_t>(i))) {
-      continue;
-    }
-    if (!spectrumInfo.hasUniqueDetector(i)) {
-      std::ostringstream sout;
-      sout << "Spectrum at workspace index " << i
-           << " has unsupported number of detectors.";
-      throw std::runtime_error(sout.str());
+    double phi = -999;
```
DavidFair: It was for testing. Good catch - I think setting it to `inf` would make a good canary value as not many calculations result in infinity (vs `NaN`)

KarlPalmen:
I found a suitable WISH file (I think it was 19610). Load took nearly 5 minutes and FindSXPeaks took about a minute and a half. Peaks were shown on instrument view and after you showed me, were OK.
When checking units validation, I found that the workspace without TOF units was not listed in those available to the algorithm. Memory was freed when Mantid was closed down. :shipit:

PR #19541, opened 2017-05-12
Do not plot curves on a log scale where all(y)<=0

### LINKED ISSUE #19529: Plotting 0 on log axis shows 0.1
If there is a spectrum that is entirely made of 0 values, plotting the spectrum on a log scale will give a show that log(0) is at 0.1
For example:
**Regular plot**
image
**log(y)**
image

martyngigg:
Lower the value considered as the minimum positive value for a data curve where `all(y) <= 0` is true.

**To test:**

Play with 1D plotting as much as possible. Check:

- plots where all Y data is positive for both linear & log scales 
- plots with a mix of Y data both positive & negative both linear & log scales 
- plots where all Y data is 0 on linear & log scales
- plots where all Y data is negative on linear & log scales

Also check embedded plots on instrument view & the slice viewer still work.

Fixes #19529.

[Release Notes](https://github.com/mantidproject/mantid/pull/19541/commits/c708aa33c96513a2a757b8c024d391e0c88cbe53
)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


mantid-roman:
Maybe in this case we shouldn't output anything? Just leave the plot empty?

mantid-roman:
I think if `QwtData::size()` returns 0 nothing is plotted.

martyngigg:
Thanks, I'll investigate that and see how it seems.

martyngigg:
@mantid-roman I've made some updates not to plot curves in this situation. I'd appreciate a review if you have some time.

@louisemccann Could you check a couple of the Indirect interfaces to see if the changes make sense there?

MantidQt/API/src/MantidQwtWorkspaceData.cpp
```diff
@@ -27,53 +38,64 @@ operator=(const MantidQwtWorkspaceData &data) {
...
-  const double maxDouble = std::numeric_limits<double>::max();
-  double curMin = maxDouble;
-  double curMinPos = maxDouble;
-  double curMax = -maxDouble;
+  // Set this to true to get the "real" data size
+  // It's correct value is then recalculated below. This is not
+  // too nice but a big refactor is not worth it given the new
+  // workbench/plotting developments.
+  m_plottable = DataStatus::Plottable;
+  m_minY = m_maxY = m_minPositive = 0.0;
```
mantid-roman: Is this for initialisation only? `m_minPositive` cannot be 0 for plottable data.

martyngigg: Yes, it's just so that they are all initialized. It should be set via all code paths below the `for` loop.

mantid-roman:
Works for me

louisemccann:
Works fine for me generally, if plotting with error bars if the curve is negative/zero the line disappears but the error bars remain. Example ws:
[error_bars.zip](https://github.com/mantidproject/mantid/files/1005066/error_bars.zip)

martyngigg:
@louisemccann Nice catch. I've fixed that now, if you could double check it's okay that would be good.

louisemccann:
Error bars are no longer plotted. Works for me

louisemccann (approved):

martyngigg:
Thanks. Could someone stick the squirrel on it?

mantid-roman:
:shipit:

PR #19556, opened 2017-05-16
New unit for TOF powder diffraction

### LINKED ISSUE #19532: Create new unit for 2D powder diffraction data reduction
For 2D reduction and analysis of TOF powder diffraction data from coming POWTEX@MLZ and DREAM@ESS instruments, a new unit
```latex
d_{\perp} = \sqrt{\lambda^2 - 2\log\cos\vartheta}
```
invented in [J. Appl. Cryst. (2015) 48, 1627--1636], is needed.

Labels: Framework, Powder diffraction
Please, assign me, I will implement it.

mganeva:
Implemented new unit for the TOF powder diffraction. 

**To test:**

Can be tested on sample workspaces, for example:
```python
ws = CreateSampleWorkspace(WorkspaceType="Histogram", Function="Powder Diffraction", NumBanks=1, XUnit="TOF")
ws_dp = ConvertUnits(ws, Target="dSpacingOrth", EMode="Direct")
```

The calculations are tested with Unittests, but can also be tested with a Python script, for example:
```python
from scipy.constants import m_n, h
import numpy as np

ssd = 10.0   # source-sample distance, default for CreateSampleWorkspace
sdd = 5.0     # source-detector distance

def d_perp (t, theta):
    tmp = t*h*1e+10/(m_n*(ssd + sdd)*1e+6)
    return np.sqrt(tmp*tmp - 2.0*np.log(np.cos(theta)))
    
ws = CreateSampleWorkspace(WorkspaceType="Histogram", Function="Powder Diffraction", NumBanks=1, XUnit="TOF")
ws_dp = ConvertUnits(ws, Target="dSpacingOrth", EMode="Direct")

ws_index = 0
x_tof = ws.readX(ws_index)
x_dp = ws_dp.readX(ws_index)
det = ws.getDetector(ws_index)
two_theta = ws.detectorTwoTheta(det)

x_dp_test = d_perp(x_tof, 0.5*two_theta)
print np.allclose(x_dp, x_dp_test, atol=1e-07)
```

Fixes #19532

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


louisemccann:
The examples above work fine. I tried converting back to TOF:
`ws_tof = ConvertUnits(ws_dp, Target="TOF", EMode='Direct')`
and the zero values are a bit off. Everything else is fine
image

mganeva:
@louisemccann Very nice point! Thank you for reviewing. I think, I've found the reason, I will fix it today.

louisemccann:
looks good to me :shipit:

Framework/Kernel/src/Unit.cpp
```diff
@@ -583,6 +583,54 @@ double dSpacing::conversionTOFMax() const { return DBL_MAX / factorTo; }
 
 Unit *dSpacing::clone() const { return new dSpacing(*this); }
 
+// ==================================================================================================
+/* D-SPACING Perpendicular
+ * ==================================================================================================
+ *
+ * Conversion uses equation: d^2 = lambda^2 - 2[Angstrom^2]*ln(sin(theta))
```
AndreiSavici: This comment is incompatible with the rest of the ticket. Should there be `cos` instead?

AndreiSavici (changes requested):
I asked several instrument scientists and programmers for powder diffraction and no one knows what `dSpacingOrth` is. We need some documentation why we need this unit, and where it is used. Also, the name is not nice. Should it be `perpendiculardSpacing` or `dSpacingPerpendicular` instead?

Framework/Kernel/src/Unit.cpp
mganeva: you are right, it should be cos. The implementation is correct, error is in the comment.

mganeva: I will change the name to dSpacingPerpendicular and write some additional documentation about unit. Where to put it?

AndreiSavici: Just add it to `UnitFactory.rst` concept page

mganeva: For me it is also unusual. The idea was to create a unit which is orthogonal to dSpacing(theta, lambda). This is needed for 2D Rietveld refinement of TOF powder diffraction data from POWTEX@MLZ and DREAM@ESS due to specific detector construction. We plan to test the approach at POWGEN soon.

AndreiSavici: Just add a detailed description (with formula) or a paper reference, to show what this unit is all about. What I mean is that I want to be able to check that the formula that you implement makes sense or not

mganeva: Some more information may be found here, in the talk of A. Houben: https://indico.esss.lu.se/event/785/session/5/material/0/3.pdf

mganeva: I will do it by Monday.

AndreiSavici (approved):

PR #19565, opened 2017-05-17
Scanning workspace construction

### LINKED ISSUE #19198: Scanning Workspace Construction Helper
Time indexing is now available in Mantid such that scanning workspaces can be constructed - see https://github.com/mantidproject/mantid/pull/19119. Example of the construction can be seen in the `MatrixWorkspaceTest.h`.

This PR will add a helper class to construct scanning workspaces in the typical ILL case - where every detector has the same set of time indexes. The class will take:
* A `start_time` and a vector of durations or a vector of start and end times
* A vector of positions and rotations
* The corresponding spectra for every detector and monitor for each scan index
* A way to create sample logs with the same start and end times

The class will then be able to return a workspace in a consistent state with the correct number of scan intervals. Any issues with mismatched sizes etc. should throw a suitable error.

ianbush:
A new class, `ScanningWorkspaceBuilder` has been added. This is meant to allow for easier construction of scanning workspaces.

`CreateSampleWorkspace` now supports building a sample scanning workspace using the builder.

A real loader using the `ScanningWorkspaceBuilder` is in progress, currently on the branch `d2b_loading`.

**To test:**

Code review. Check documentation is clear and complete.

Try creating a  scanning workspace with `CreateSampleWorkspace`. Check the dimensions, positions and angles are as expected.

Fixes #19198.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Algorithms/src/CreateSampleWorkspace.cpp
```diff
@@ -159,6 +160,11 @@ void CreateSampleWorkspace::init() {
                   boost::make_shared<BoundedValidator<double>>(0, 1000, true),
                   "The distance along the beam direction from the source to "
                   "the sample in M (default:10.0)");
+  declareProperty("NumScanPoints", 1,
+                  boost::make_shared<BoundedValidator<int>>(0, 360, true),
```
OwenArnold: Where does the 360 bounds come from?

ianbush: The detectors are rotated around the sample in 1 degree increments, hence the limit. Would ti be helpful to clarify this? It is only mentioned explicitly in the documentation.

OwenArnold: It probably would be better to be clear on this "1 degree increments" aspect of the functionality

Framework/DataObjects/src/ScanningWorkspaceBuilder.cpp
```diff
@@ -0,0 +1,323 @@
...
+        newStartTime, endTime));
+  }
+
+  setTimeRanges(timeRanges);
+}
+
+/**
+ * Supply a vector of vectors which contain positions. The inner vectors should
+ *contain the position for each time index, the outer vector the vector for each
+ *detector.
```
OwenArnold: So should we not also compare things like the outer vector position size with outer rotation vector size? There may be some other examples for this. I didn't see that as part of the validation.

Framework/DataObjects/inc/MantidDataObjects/ScanningWorkspaceBuilder.h
```diff
@@ -0,0 +1,112 @@
...
+      std::pair<Kernel::DateAndTime, Kernel::DateAndTime>> &timeRanges);
+  void setTimeRanges(const Kernel::DateAndTime &startTime,
+                     const std::vector<double> &durations);
+  void setPositions(const std::vector<std::vector<Kernel::V3D>> &positions);
+  void setRotations(const std::vector<std::vector<Kernel::Quat>> &rotations);
+  void setInstrumentAngles(const std::vector<double> &instrumentAngles);
+
+  void setIndexingType(const IndexingType indexingType);
+
+  API::MatrixWorkspace_sptr buildWorkspace();
```
OwenArnold: Shouldn't this be a `const` method. I'm not clear why it wouldn't be.

Framework/DataObjects/src/ScanningWorkspaceBuilder.cpp
```diff
@@ -0,0 +1,323 @@
...
+      const auto rotation =
+          Kernel::Quat(m_instrumentAngles[j], Kernel::V3D(0, 1, 0));
+      rotation.rotate(position);
+      outputDetectorInfo.setPosition({i, j}, position);
+      outputDetectorInfo.setRotation({i, j}, rotation);
+    }
+  }
+}
+
+void ScanningWorkspaceBuilder::createTimeOrientedIndexInfo(
```
OwenArnold: Should be `const` method

Framework/DataObjects/src/ScanningWorkspaceBuilder.cpp
```diff
@@ -0,0 +1,323 @@
...
+      spectrumDefinitions.access()[detIndex * m_nTimeIndexes + timeIndex].add(
+          detIndex, timeIndex);
+    }
+  }
+
+  indexInfo.setSpectrumDefinitions(spectrumDefinitions);
+  ws.setIndexInfo(indexInfo);
+}
+
+void ScanningWorkspaceBuilder::createDetectorOrientedIndexInfo(
```
OwenArnold: Should be `const` method

Framework/DataObjects/src/ScanningWorkspaceBuilder.cpp
OwenArnold: Also, since all these such functions are private, consider making these free functions and keeping the function declarations out of the header file. These can just live in an anonymous namespace.

Framework/Algorithms/src/CreateSampleWorkspace.cpp
```diff
@@ -309,6 +320,36 @@ MatrixWorkspace_sptr CreateSampleWorkspace::createHistogramWorkspace(
...
+  builder.setInstrument(inst);
+
+  auto angles = std::vector<double>();
+  auto timeRanges = std::vector<double>();
+  for (int i = 0; i < numScanPoints; ++i) {
+    angles.push_back(double(i));
+    timeRanges.push_back(double(i + 1));
+  }
+
+  builder.setTimeRanges(Kernel::DateAndTime(0), timeRanges);
```
OwenArnold: I think it would be good to make the setters take r-value references. That way we can just move these temporaries rather than copying them.

Framework/DataObjects/test/ScanningWorkspaceBuilderTest.h
```diff
@@ -0,0 +1,509 @@
...
+
+  void checkTimeRanges(const DetectorInfo &detectorInfo) {
+    for (size_t i = 0; i < nDetectors; ++i) {
+      for (size_t j = 0; j < nTimeIndexes; ++j) {
+        TS_ASSERT_EQUALS(detectorInfo.scanInterval({i, j}), timeRanges[j]);
+      }
+    }
+  }
+};
+
```
OwenArnold: No Performance test. I think tracking performance would probably be a good idea.

Framework/DataObjects/inc/MantidDataObjects/ScanningWorkspaceBuilder.h
```diff
@@ -0,0 +1,112 @@
...
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+class MANTID_DATAOBJECTS_DLL ScanningWorkspaceBuilder {
+public:
+  enum class IndexingType { DEFAULT, TIME_ORIENTED, DETECTOR_ORIENTED };
```
OwenArnold: What does `DEFAULT` mean?

Framework/DataObjects/src/ScanningWorkspaceBuilder.cpp
ianbush: The outer vector size must always be the number of detectors, `m_nDetectors`. So they are ensured to be the same.

Framework/DataObjects/inc/MantidDataObjects/ScanningWorkspaceBuilder.h
ianbush: Here `DEFAULT` means that `buildDefaultSpectrumDefinitions();` is called in `MatrixWorkspace::setIndexInfo()`, and no spectrum definitions are supplied.

I discussed this a little with @SimonHeybrock a while back, and it was decided not to promise any particular method for building the default spectrum definitions.

I wonder if my choice of names could be a little better here though...

Framework/DataObjects/src/ScanningWorkspaceBuilder.cpp
```diff
@@ -0,0 +1,323 @@
...
+      outputDetectorInfo.setPosition({i, j}, position);
+      outputDetectorInfo.setRotation({i, j}, rotation);
+    }
+  }
+}
+
+void ScanningWorkspaceBuilder::createTimeOrientedIndexInfo(
+    MatrixWorkspace &ws) {
+  auto indexInfo = ws.indexInfo();
+  auto spectrumDefinitions = Kernel::make_cow<std::vector<SpectrumDefinition>>(
```
SimonHeybrock: I would recommend simply using a vector here to reduce the hassle (move it when passing it to `IndexInfo`). The variant of `IndexInfo::setSpectrumDefinitions` that takes a `cow_ptr` is mainly for internal use by `MatrixWorkspace` and not required here.

Framework/DataObjects/src/ScanningWorkspaceBuilder.cpp
```diff
@@ -0,0 +1,323 @@
...
+  }
+
+  indexInfo.setSpectrumDefinitions(spectrumDefinitions);
+  ws.setIndexInfo(indexInfo);
+}
+
+void ScanningWorkspaceBuilder::createDetectorOrientedIndexInfo(
+    MatrixWorkspace &ws) {
+  auto indexInfo = ws.indexInfo();
+  auto spectrumDefinitions = Kernel::make_cow<std::vector<SpectrumDefinition>>(
```
SimonHeybrock: See above.

Framework/Indexing/inc/MantidIndexing/IndexInfo.h
```diff
@@ -10,7 +10,11 @@
 #include <vector>
 
 namespace Mantid {
+
+using specnum_t = int32_t;
+using detid_t = int32_t;
```
SimonHeybrock: The indexing module does not use those typedefs, there are type-safe types provided for this purpose, `SpectrumNumber` and `DetectorID`. Not sure why this was added to the header, maybe an oversight?

docs/source/release/v3.10.0/framework.rst
```diff
@@ -11,6 +11,7 @@ API
 
 - The default multiple file limit is now made facility dependent. It is 1000 for ILL, and 100 for all the others.
 - Frequency unit (GHz) included as an option to represent energy transfer.
+- Framework changes now mean scanning workspaces (workspaces with moving detectors) can be created. Currently this can be tested using a new option in :ref:`FilterEvents <algm-CreateSampleWorkspace-v1>`, by setting `NumScanPoints`.
```
SimonHeybrock: Maybe we should also mention the limitations, mainly that the `InstrumentView` should not work, and algorithms may either fail or do nonsense?

SimonHeybrock (commented):
Some comments below (have not yet looked through the tests):

Framework/DataObjects/inc/MantidDataObjects/ScanningWorkspaceBuilder.h
```diff
@@ -0,0 +1,112 @@
...
+  API::MatrixWorkspace_sptr buildWorkspace();
+
+private:
+  size_t m_nDetectors;
+  size_t m_nTimeIndexes;
+  size_t m_nBins;
+
+  HistogramData::Histogram m_histogram;
+
+  boost::shared_ptr<const Geometry::Instrument> m_instrument;
```
OwenArnold: Just an idea - for member variables that are mandatory, one way you might wish to avoid logical misuse errors is to wrap them as `boost::optional`. That way there is no danger of blindly accepting default initialised values if setters have not been called. You can always check that these have been initialised prior to use via `boost::optional::operator*()`

Framework/DataObjects/src/ScanningWorkspaceBuilder.cpp
```diff
@@ -0,0 +1,323 @@
...
+ *positions and rotations of the detectors.
+ *
+ * Here explicit assumptions are made - that the source is at (0, 0, 0), and the
+ *rotation is in the X-Z plane. This corresponds to the common case of moving
+ *detectors to increase angular coverage.
+ *
+ * @param instrumentAngles a vector of angles, the size matching the number of
+ *time indexes
+ */
+void ScanningWorkspaceBuilder::setInstrumentAngles(
```
ianbush: I am in two minds about this too. On balance I think this method is general enough to be here, as all real cases for scanning instruments at ILL and MLZ rotate around the sample position in such a way.

Framework/DataObjects/src/ScanningWorkspaceBuilder.cpp
```diff
@@ -0,0 +1,323 @@
...
+  if (!m_rotations.empty())
+    buildRotations(outputDetectorInfo);
+
+  if (!m_instrumentAngles.empty())
+    buildInstrumentAngles(outputDetectorInfo);
+
+  switch (m_indexingType) {
+  case IndexingType::DEFAULT:
+    outputWorkspace->setIndexInfo(
+        Indexing::IndexInfo(m_nDetectors * m_nTimeIndexes));
```
ianbush: Yes, you are right, this is a bit confusing at the moment. For now I think it makes sense to ensure the number of detectors in the instrument and `nDetectors` are the same.

I think moving the instrument into the constructor might be better, then `nDetectors` does not need to be given separately.

Framework/Indexing/inc/MantidIndexing/IndexInfo.h
ianbush: I am not sure how they got in there, this was indeed an oversight.

docs/source/release/v3.10.0/framework.rst
ianbush: Yes, we should, and that saving and loading will not give sensible results.

Framework/DataObjects/src/ScanningWorkspaceBuilder.cpp
SimonHeybrock: Would the code be any more complex if this was not a member?

Framework/DataObjects/src/ScanningWorkspaceBuilder.cpp
ianbush: All these methods make use of some member variables. I think this is possible only by passing in the extra information, so I am not sure if this will be so nice for code readability?

OwenArnold: Ok. I think there might be some that take only use a single internal parameter. But i agree these `build...` ones might not make sense as non-members.

Framework/DataObjects/inc/MantidDataObjects/ScanningWorkspaceBuilder.h
```diff
@@ -54,50 +54,50 @@ namespace DataObjects {
...
   void setTimeRanges(const std::vector<
-      std::pair<Kernel::DateAndTime, Kernel::DateAndTime>> &timeRanges);
+      std::pair<Kernel::DateAndTime, Kernel::DateAndTime>> timeRanges);
   void setTimeRanges(const Kernel::DateAndTime &startTime,
                      const std::vector<double> &durations);
-  void setPositions(const std::vector<std::vector<Kernel::V3D>> &positions);
-  void setRotations(const std::vector<std::vector<Kernel::Quat>> &rotations);
-  void setInstrumentAngles(const std::vector<double> &instrumentAngles);
+  void setPositions(const std::vector<std::vector<Kernel::V3D>> positions);
+  void setRotations(const std::vector<std::vector<Kernel::Quat>> rotations);
```
SimonHeybrock: If you want to move from these they should not be `const` (which is ok, since you are no passing by value).

Framework/DataObjects/src/ScanningWorkspaceBuilder.cpp
```diff
@@ -17,24 +17,14 @@ using namespace Mantid::Indexing;
...
-                                                   const size_t nBins)
-    : m_nDetectors(nDetectors), m_nTimeIndexes(nTimeIndexes), m_nBins(nBins),
-      m_histogram(BinEdges(nBins + 1, LinearGenerator(0.0, 1.0)),
+ScanningWorkspaceBuilder::ScanningWorkspaceBuilder(
+    const boost::shared_ptr<const Geometry::Instrument> &instrument,
+    const size_t nTimeIndexes, const size_t nBins)
+    : m_nDetectors(instrument->getNumberDetectors()),
+      m_nTimeIndexes(nTimeIndexes), m_nBins(nBins), m_instrument(instrument),
+      m_histogram(BinEdges(nBins + 1, LinearGenerator(1.0, 1.0)),
                   Counts(std::vector<double>(nBins, 0.0))),
```
SimonHeybrock: You can simply pass `nBins, 0.0`, `Counts` has the same constructors as `std::vector`.

Framework/DataObjects/src/ScanningWorkspaceBuilder.cpp
```diff
@@ -44,13 +34,13 @@ void ScanningWorkspaceBuilder::setInstrument(
  * @param histogram A histogram with bin edges defined
  */
 void ScanningWorkspaceBuilder::setHistogram(
-    const HistogramData::Histogram &histogram) {
+    const HistogramData::Histogram histogram) {
   if ((histogram.counts().size() != m_nBins) ||
       (histogram.binEdges().size() != m_nBins + 1))
```
SimonHeybrock: Use `Histogram::size()`. `Histogram` guarantees that the bin edges are always one more than the number of bins, no need to checking.

Framework/DataObjects/src/ScanningWorkspaceBuilder.cpp
```diff
@@ -243,30 +235,31 @@ void ScanningWorkspaceBuilder::buildPositions(
   }
 }
 
-void ScanningWorkspaceBuilder::buildInstrumentAngles(
+void ScanningWorkspaceBuilder::buildRelativeRotationsForScans(
     DetectorInfo &outputDetectorInfo) const {
   for (size_t i = 0; i < outputDetectorInfo.size(); ++i) {
     for (size_t j = 0; j < outputDetectorInfo.scanCount(i); ++j) {
```
SimonHeybrock: This is what I referred to on Slack (a bit cryptically maybe): If you would apply these rotations before merging the `DetectorInfos`, you could simply rotate a parent of the detectors via `DetectorInfo::setRoation(IComponent *, ...)`. That should also be much faster, since it uses `Eigen` directly internally. In principle rotating the full instrument would work, but that would also rotate monitors etc., so you'd have to figure out how to handle those.

Framework/DataObjects/src/ScanningWorkspaceBuilder.cpp
```diff
@@ -0,0 +1,312 @@
...
+    }
+  }
+}
+
+void ScanningWorkspaceBuilder::buildRelativeRotationsForScans(
+    DetectorInfo &outputDetectorInfo) const {
+  for (size_t i = 0; i < outputDetectorInfo.size(); ++i) {
+    for (size_t j = 0; j < outputDetectorInfo.scanCount(i); ++j) {
+      auto position = outputDetectorInfo.position({i, j});
+      const auto rotation = Kernel::Quat(m_instrumentAngles[j], m_rotationAxis);
```
SimonHeybrock: Can you double check what this does? `Eigen::Quaternion` requires the axis to be normalized, otherwise you get garbage (a quaternion that is not representing a rotation), not sure about `Kernel::Quat`.

Framework/DataObjects/src/ScanningWorkspaceBuilder.cpp
```diff
@@ -0,0 +1,312 @@
...
+      position -= m_rotationPosition;
+      rotation.rotate(position);
+      position += m_rotationPosition;
+      outputDetectorInfo.setPosition({i, j}, position);
+      outputDetectorInfo.setRotation({i, j}, rotation);
```
SimonHeybrock: You are dropping the previous rotation here, i.e, rotation is treated as an absolute value rather than an offset.

Framework/DataObjects/src/ScanningWorkspaceBuilder.cpp
ianbush: The rotation axis is normalised for you, see - https://github.com/mantidproject/mantid/blob/master/Framework/Kernel/src/Quat.cpp#L117.

Framework/DataObjects/test/ScanningWorkspaceBuilderTest.h
```diff
@@ -320,6 +328,58 @@ class ScanningWorkspaceBuilderTest : public CxxTest::TestSuite {
...
 
   void
+  test_creating_workspace_with_relative_rotations_on_previously_rotated_detectors() {
+
+    const auto &instWS =
+        WorkspaceCreationHelper::create2DWorkspaceWithFullInstrument(
+            int(nDetectors), int(nBins));
+    auto &instDetInfo = instWS->mutableDetectorInfo();
+
+    Quat rotation = Quat(10.0, V3D(0, 1, 0));
```
SimonHeybrock: If both the original rotation and the updated one are around the same axis you are not catching a wrong rotation order (remember, rotation matrices to not commute!).

SimonHeybrock:
I am happy with the code now, tests have been added, and I think despite scanning being a new feature there is not really any risk of breaking anything with this since the changes do not affect existing code/workflows (in case anyone is concerned about doing this just before a release).

:shipit:

martyngigg:
@ianbush There seems to be a failing test on Windows. Do you have access to a Windows machine?

ianbush:
@martyngigg Yes, I do, I will take a look at the failing test this morning. It is a slightly worrying difference in results...

martyngigg:
@ianbush The final failure is a real failure in the doc test [example code](/mantidproject/mantid/blame/master/docs/source/algorithms/RecordPythonScript-v1.rst#L69). The new parameter on the algorithm has caused the output comparison to fail.

ianbush:
Please see new PR #19606.

PR #19604, opened 2017-05-22
Copy defaults from FindPeaks

peterfpeterson:
Just copies default property values from `FindPeaks`. 

**To test:**

See what `FindPeaks` has for these properties.

*Does not need to be in the release notes* because it is such a minor change.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/src/PDCalibration.cpp
+++ b/Framework/Algorithms/src/PDCalibration.cpp
@@ -221,10 +221,10 @@ void PDCalibration::init() {
   declareProperty("HighBackground", true,
                   "Relatively weak peak in high background");
   declareProperty(
-      "MinGuessedPeakWidth", 4, min,
+      "MinGuessedPeakWidth", 2, min,
       "Minimum guessed peak width for fit. It is in unit of number of pixels.");
   declareProperty(
-      "MaxGuessedPeakWidth", 4, min,
+      "MaxGuessedPeakWidth", 10, min,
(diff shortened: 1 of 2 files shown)


KarlPalmen:
It indeed does use the same defaults as FindPeaks. I also noticed it uses `min` instead of `mustBePostive` even though it is exactly the same thing. Also unlike `min`, `mustBePositive` tells you it is a validator.

rosswhitfield:
These values were copied from what [GetDetOffsetsMultiPeaks was using](https://github.com/mantidproject/mantid/blob/master/Framework/Algorithms/src/GetDetOffsetsMultiPeaks.cpp#L907-L908)

KarlPalmen:
Is this issue vital for release 3.10?

KarlPalmen:
Also I doubt that changing of any default value is a minor change.

NickDraper:
Actually in this case the min validator is for ints, while the mustbepositive is for double values, they may not be the best variable names, but have not been set in the PR so are a bit out of scope for this PR.


If all the unit and system tests pass it whether means that this change is fine, or that the algorithm is not used in the, but this was done before the code freezer and  should be allowed, the changed defaults are more consistent with the other algorithm which makes more sense.

The only thing that I see is missing is a line in the release notes, which I'l add myself in a minute.

NickDraper:
Final change is release notes only, no need to wait for the build servers

:shipit:

PR #19672, opened 2017-05-24
Fix monitors information in detectors table

### LINKED ISSUE #18904: Monitor information does not show correctly in detectors table
Some cells for monitors do not show up correctly in the detectors table:

detector_table

In the image above, *R* for the first monitor should be positive, and the last should show 'yes' to all monitors.

soininen:
This PR fixes an issue in MantidPlot's detectors table where monitor information did not show up correctly:
- The *R* column would have an negative distance if monitor's *theta* was 180 degrees
- While calculating *q*, DetectorInfo would throw an uncaught exception preventing the remaining columns to be filled.

**To test:**

Load a workspace with monitors into MantidPlot. View the detector table by right-clicking the workspace and selecting 'Show Detectors'. Before these changes are applied, monitors before the sample have a negative *R*, and the *Monitor* column shows zero instead of 'yes'.

Fixes #18904.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


AnthonyLim23:
In master, if I load LOQ49886 (from training data)then the Monitor says yes instead of 1 as described above. If R=0 then the Monitor says 0

AnthonyLim23:
These changes are already in master.

soininen:
@AnthonyLim23 Could you try testing again with `<mantid-build-directory>/ExternalData/Testing/Data/UnitTest/ILL/IN4/084446.nxs`? The LOQ49886 file doesn't trigger the calculation of the *Q* column in the detectors table and thus the Monitor column is correct. Sorry for the confusion, I should have given more detailed testing instructions in the first place.

Anyway, the *R* column for the LOQ49886 might still show negative values even when these changes are applied. I'll look into it.

AnthonyLim23:
I have loaded the file you suggested. Am I right in saying that it is only the last entry on the table that is affected?

AnthonyLim23:
The Monitor column now says yes for the last entry in the table (only one with a negative R).

soininen:
Phi is the off-plane angle and for monitors which are on the beam axis, it is zero.

The calculation of the Q value for monitors would fail before these changes causing the rest of the monitor row (the *Monitor* column) to show zeros. This is due to `DetectorInfo::twoTheta()` throwing `std::logic_error` for monitors. Since `DetectorInfo` says that the scattering angle is not defined for monitors, I decided to set Q to not-a-number for monitors.

AnthonyLim23:
Ok, so it should do that. Once the merge conflict and test have passed :shipit:

PR #19699, opened 2017-05-25
WorkspacePropertyWithIndex Prototype Implementation

### LINKED ISSUE #19130: WorkspacePropertyWithIndex property
This issue is with regards to the new indexing features being implemented in Mantid.

This property will allow indexing translation and validation to occur as part of the declaration of the property. Users would declare the property using something like:
```cpp
declareProperty(make_unique<WorkspacePropertyWithIndex>("InputWorkspace", "", 
                          Direction::Input, UseSpectrumNumber|UseDetectorID|UseWorkspaceIndex), 
                          "Input workspace and selected indices loaded simultaneously.");
```
Users would then be able to retrieve indexes in the following manner:

```cpp
//c++11
const MatrixWorkspace_const_sptr inputWS;
  const IndexSet indexSet;
  std::tie(inputWS, indexSet) = getProperty("InputWorkspaceWithIndex");

//c++17 
auto [inputWS, indexSet] = getProperty("InputWorkspaceWithIndex");
```
```python
# python
inputWs, indexSet = getProperty(InputWorkspaceWithIndex);
```
N.B This issue relates to Mantid 4.0 although it could be added to before that.

LamarMoore:
**DO NOT MERGE**

This is a prototype for the implementation of the `WorkspacePropertyWithIndex`. See the [design document](https://github.com/mantidproject/documents/pull/42) for the full details. Please remember this is a prototype, code quality will be addressed after the mechanism is approved. 

What is currently implemented?
-----------------------------------
- Templated C++ WorkspacePropertyWithIndex (and unit tests)
- Python Exports

What is left?
-------------
- [x] Exporting all relevant types to python
- [x] Solving Algorithm History
- [ ] GUI Handling

Discussion Points
-------------------
- What unit tests are missing? (especially in python)
- Interaction with IPropertyManager
- Access to the ArrayProperty and IndexTypeProperty child properties
- simpleapi modifications for special WorkspacePropertyWithIndex handling

**To test:**

For now just a thorough code review

Fixes #19130.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/API/src/IndexTypeProperty.cpp
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  return types;
+}
+
+std::vector<std::string> IndexTypeProperty::allowedValues() const {
+  return m_allowedValues;
+}
+
+bool IndexTypeProperty::isMultipleSelectionAllowed() { return true; }
```
SimonHeybrock: What is the effect of this? I am not sure why this is set to `true`.

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -9,13 +9,16 @@
...
-#include "MantidKernel/TimeSeriesProperty.h"
+#include "MantidGeometry/MDGeometry/MDFrame.h"
+#include "MantidIndexing/IndexInfo.h"
 #include "MantidKernel/MDUnit.h"
 #include "MantidKernel/Strings.h"
+#include "MantidKernel/TimeSeriesProperty.h"
 #include "MantidKernel/make_unique.h"
-#include "MantidIndexing/IndexInfo.h"
+#include <MantidAPI/WorkspacePropertyWithIndex.h>
+#include <MantidAPI/WorkspacePropertyWithIndex.tcc>
```
SimonHeybrock: .

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -2047,6 +2055,80 @@ IPropertyManager::getValue<Mantid::API::MatrixWorkspace_const_sptr>(
   }
 }
 
+template <>
```
SimonHeybrock: Do we have to add bits of code like the one below for all workspace types (I saw it above also for `IEventWorkspace`)? Can that be avoided?

Framework/API/test/IndexTypePropertyTest.h
```diff
@@ -0,0 +1,83 @@
...
+    IndexTypeProperty prop(IndexType::SpectrumNumber |
+                           IndexType::WorkspaceIndex);
+
+    prop = "SpectrumNumber";
+
+    TS_ASSERT(prop.selectedType() & IndexType::SpectrumNumber);
+
+    prop = "WorkspaceIndex";
+
+    TS_ASSERT(prop.selectedType() & IndexType::WorkspaceIndex);
```
SimonHeybrock: Why check bit wise? Shouldn't it always be one-hot bit?

Framework/API/test/IndexTypePropertyTest.h
```diff
@@ -0,0 +1,83 @@
...
+  void testCorrectTypeReturnedWhenSetWithIndexType() {
+    IndexTypeProperty prop(IndexType::SpectrumNumber);
+
+    prop = IndexType::SpectrumNumber;
+
+    TS_ASSERT(prop.selectedType() & IndexType::SpectrumNumber);
+
+    prop = IndexType::WorkspaceIndex;
+
+    TS_ASSERT(prop.selectedType() & IndexType::WorkspaceIndex);
```
SimonHeybrock: Why check bit wise?

Framework/API/test/WorkspacePropertyWithIndexHelperTest.h
```diff
@@ -0,0 +1,34 @@
...
+  // This means the constructor isn't called when running other tests
+  static WorkspaceWithIndexHelperTest *createSuite() {
+    return new WorkspaceWithIndexHelperTest();
+  }
+  static void destroySuite(WorkspaceWithIndexHelperTest *suite) {
+    delete suite;
+  }
+
+  void testCreateWithSpectrumNumbers() {
+    TS_ASSERT_THROWS_NOTHING(createWithSpectrumNumbers<MatrixWorkspace>());
```
SimonHeybrock: Maybe we can come up with a better naming? It is not clear that this is creating a property and not a workspace?

Framework/API/test/WorkspacePropertyWithIndexTest.h
```diff
@@ -0,0 +1,319 @@
...
+    delete suite;
+  }
+
+  void testConstructor() {
+    TS_ASSERT_THROWS_NOTHING((MatrixWorkspaceIndexProp()));
+  }
+
+  void testConstructorFailsWithInvalidIndexType() {
+    TS_ASSERT_THROWS(MatrixWorkspaceIndexProp("InputWorkspaceWithIndex", 0),
+                     std::invalid_argument);
```
SimonHeybrock: I think this test becomes redundant if you use a strongly typed enum as commented above?

Framework/Algorithms/src/ChangePulsetime.cpp
```diff
@@ -19,16 +20,13 @@ using std::size_t;
 /** Initialize the algorithm's properties.
  */
 void ChangePulsetime::init() {
-  declareProperty(make_unique<WorkspaceProperty<EventWorkspace>>(
-                      "InputWorkspace", "", Direction::Input),
+  declareProperty(createWithWorkspaceIndices<EventWorkspace>("InputWorkspace"),
```
SimonHeybrock: This is still a bit cumbersome to type. Would it possibly make sense to merge the helper function with the `declareProperty`, such that we can do something like this:

```cpp
declareWorkspacePropertyWithIndex<EventWorkspace>("InputWorkspace");
// or
declarePropertyWithIndex<EventWorkspace>("InputWorkspace");
```
?

Framework/Algorithms/src/ChangePulsetime.cpp
```diff
@@ -38,41 +36,27 @@ void ChangePulsetime::init() {
 /** Execute the algorithm.
  */
 void ChangePulsetime::exec() {
-  EventWorkspace_const_sptr in_ws = getProperty("InputWorkspace");
+  EventWorkspace_sptr in_ws;
```
SimonHeybrock: Why remove the const?

Framework/Algorithms/src/ChangePulsetime.cpp
```diff
@@ -38,41 +36,27 @@ void ChangePulsetime::init() {
...
+  SpectrumIndexSet indexSet(0);
+
+  std::tie(in_ws, indexSet) = std::tuple<EventWorkspace_sptr, SpectrumIndexSet>(
```
SimonHeybrock: Is the cast necessary?

Framework/Algorithms/src/ChangePulsetime.cpp
```diff
@@ -38,41 +36,27 @@ void ChangePulsetime::init() {
...
-  if (workspaceIndices.empty()) {
-    doAll = true;
-    num_to_do = in_ws->getNumberHistograms();
-  }
-
   double timeOffset = getProperty("TimeOffset");
+  Progress prog(this, 0.0, 1.0, indexSet.size());
 
-  Progress prog(this, 0.0, 1.0, num_to_do);
-  PARALLEL_FOR_NO_WSP_CHECK()
```
SimonHeybrock: Parallel for removed on purpose?

Framework/API/inc/MantidAPI/WorkspacePropertyWithIndex.tcc
```diff
@@ -0,0 +1,190 @@
...
+      boost::dynamic_pointer_cast<API::MatrixWorkspace>(getWorkspace())
+          ->indexInfo();
+
+  // If no indices provided, then assume all
+  if (list.size() == 0)
+    return indexInfo.makeIndexSet();
+
+  switch (m_indexTypeProp->selectedType()) {
+  case IndexType::SpectrumNumber:
+    return indexInfo.makeIndexSet(std::vector<Mantid::Indexing::SpectrumNumber>(
```
SimonHeybrock: These methods may throw, where is the validation happening?

I think as it is, the way I implemented failures in `IndexInfo` are not very user friendly, i.e., the error messages are not informative (might not include bad indices, etc.), so if these become exposed to users we would probably need to improve those (either directly in `IndexInfo`, or by catching at an intermediate level to add better information).

SimonHeybrock (commented):
This looks like a very good start. I have some comments below, but I am not addressing all of your questions from the issue, I will need to have another look another time.

Framework/API/inc/MantidAPI/WorkspacePropertyWithIndex.h
```diff
@@ -0,0 +1,122 @@
...
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+template <typename TYPE = MatrixWorkspace>
+class WorkspacePropertyWithIndex : public WorkspaceProperty<TYPE>,
```
LamarMoore: @SimonHeybrock it would. We essentially need to re-implement a `PropertyWithValue<T>` which is not worth it.

Framework/API/src/MatrixWorkspace.cpp
LamarMoore: We do, and it cannot currently be avoided with the was `TypedValue` works.

Framework/API/test/WorkspacePropertyWithIndexHelperTest.h
LamarMoore: Yes this needs to be removed altogether. The builder functions aren't necessary.

Framework/Algorithms/src/ChangePulsetime.cpp
LamarMoore: Yes, `TypedValue` requires a cast.

SimonHeybrock: Is that because `std::tie` is s tuple of references? The way it is not is too cumbersome I think, we should try to find a better solution.

LamarMoore: The only way I can think of to improve the situation is with explicit specialisation which increases the maintenance cost. (It might be necessary to specialise for each workspace type which is ugly).

Framework/API/inc/MantidAPI/WorkspacePropertyWithIndex.tcc
```diff
@@ -0,0 +1,190 @@
...
+        list.begin(), list.end()));
+    break;
+  case IndexType::WorkspaceIndex:
+    return indexInfo.makeIndexSet(
+        std::vector<Mantid::Indexing::GlobalSpectrumIndex>(list.begin(),
+                                                           list.end()));
+    break;
+  }
```
LamarMoore: @SimonHeybrock  The list property (see ArrayProperty) converts the range to a vector by default. I could check the min and max values and see if the length of the list is the same as `max - min`.  If it is a pure range, then I would just use min and max, otherwise I will create the vector.

Framework/API/test/WorkspacePropertyWithIndexTest.h
LamarMoore: @SimonHeybrock it is difficult to use strongly typed enums without significantly reducing readability and increasing the amount of typing needed in the constructor. We would need to write something like `WorkspacePropertyWithIndex<MatrixWorkspace>("InputWorkspace", static_cast<int>(API::IndexType::WorkspaceIndex)|static_cast<int>(API::IndexType::SpectrumNumber))` as an example where we want to allow the user to enter both index types.

Framework/API/inc/MantidAPI/WorkspacePropertyWithIndex.tcc
SimonHeybrock: Converting it to a vector will then probably cause another issue if there are for example gaps in the spectrum numbers a workspace contains: If the property provides a contiguous vector of indices based on a user-provided min+max, the index translation will fail. That is, using min/max would be impossible if there are gaps in the spectrum numbers.

Framework/API/test/WorkspacePropertyWithIndexTest.h
SimonHeybrock: Why do you need to cast in the client code, instead of within `WorkspacePropertyWithIndex`, before passing the values on to the internal `ArrayProperty`?

Framework/API/src/MatrixWorkspace.cpp
SimonHeybrock: Hmm, looks tricky, basically we would need to convert a `std::tuple<T1 &, T2&>` to `std::tuple<T1, T2>`. I will think some more about it, probably it could be managable with some template magic.

Framework/API/test/WorkspacePropertyWithIndexTest.h
LamarMoore: Because the client code needs to specify which index types should be available at the GUI level.

SimonHeybrock: But why don't you cast where you are initializing the `IndexTypeProperty`?

Framework/Algorithms/src/ChangePulsetime.cpp
LamarMoore: I need to change the loop I think. OPENMP requires for in the canonical form unless I do something like [this](https://stackoverflow.com/questions/17848521/using-openmp-with-c11-range-based-for-loops)

Framework/API/inc/MantidAPI/WorkspacePropertyWithIndex.tcc
LamarMoore: The validation already occurs in `isValid`. Do you think this should also catch and ignore/rethrow any exceptions?

SimonHeybrock: As I said, either that, or look into making the exceptions in `IndexInfo` more informative.

SimonHeybrock: Obviously not ignore, you want the exception message in the validator!

Framework/API/test/WorkspacePropertyWithIndexTest.h
LamarMoore: The constructor has to be able to accept `WorkspaceIndex|SpectrumNumber`. The only way to do this is as an integral type. If I accept `API::IndexType` in the constructor, I cannot accept multiple index types. This has to be specified in the constructor.

SimonHeybrock: Ok, as clarified on slack: I had not realized that we need to use bitwise operations here to pass multiple options, so :+1: .

Framework/Algorithms/src/ChangePulsetime.cpp
```diff
@@ -50,8 +54,13 @@ void ChangePulsetime::exec() {
   double timeOffset = getProperty("TimeOffset");
   Progress prog(this, 0.0, 1.0, indexSet.size());
 
-  for (auto i : indexSet) {
-    out_ws->getSpectrum(i).addPulsetime(timeOffset);
+  // This loop is ugly but represents the best way
+  // of parallelising the Workspace access using OpenMP.
+  // In future we may be able to use a range-based loop
+  // with openMP.
```
SimonHeybrock: I'd remove this comment, it is not really useful.

Framework/API/inc/MantidAPI/WorkspacePropertyWithIndex.tcc
```diff
@@ -170,21 +173,39 @@ const SpectrumIndexSet WorkspacePropertyWithIndex<TYPE>::getIndices() const {
...
-  case IndexType::WorkspaceIndex:
-    return indexInfo.makeIndexSet(
-        std::vector<Mantid::Indexing::GlobalSpectrumIndex>(list.begin(),
-                                                           list.end()));
-    break;
+  // Determine if input is a pure range or just a list of numbers
+  auto res = std::minmax_element(list.begin(), list.end());
+  auto minIndex = res.first - list.begin();
+  auto maxIndex = res.second - list.begin();
+  bool isPureRange = (list[maxIndex] - list[minIndex] + 1) == list.size();
```
SimonHeybrock: Two comments:
- If a user enters "1,2,3,4,5" this will give `isPureRange == true`, even though it isn't? If spectrum 3 does not exist this would work, even though it should not. Maybe this is a corner case that can be ignored, but I think it should be discussed with input from others.
- This is still inefficient for long ranges. Should we maybe consider adding also separate input fields for min and max, such that conversion to a vector can be avoided?

Framework/Algorithms/src/ChangePulsetime.cpp
SimonHeybrock: There does not seem to be a nice solution to this. The main issue is that we would need to convert to a tuple of references. See also here https://stackoverflow.com/questions/26276943/make-custom-type-tie-able-compatible-with-stdtie.

Given that, it seem to be a bit pointless to explicitly convert something to a tuple, just to use `tie` to unpack it right away. I think we should consider just doing to separate calls, i.e.,

```cpp
EventWorkspace_const_sptr in_ws = getProperty("InputWorkspace");
SpectrumIndexSet indexSet = getProperty("InputWorkspace");
```

Or maybe someone has a better solution.

In any case, I think you should bring this issue back into the design document, so other people can have a look at this.

Framework/Algorithms/src/ChangePulsetime.cpp
LamarMoore: Someone may try to change it to a range-based loop in future. I think the comment lets other developers understand why it is written in canonical form.

Framework/API/inc/MantidAPI/WorkspacePropertyWithIndex.tcc
LamarMoore: 1. If spectrum 3 does not exist, an exception will be thrown which will be caught in the validator. The user would have entered invalid input. Not sure what the problem is.
2. The problem I think is that ArrayProperty may be inefficient for long ranges. Might be an idea to augment ArrayProperty to handle pure ranges more effectively? Also, how should we handle something like `1:33,45:55,76,78` which I think is perfectly valid input.

Framework/API/test/WorkspacePropertyWithIndexTest.h
```diff
@@ -0,0 +1,351 @@
...
+  // This pair of boilerplate methods prevent the suite being created statically
+  // This means the constructor isn't called when running other tests
+  static WorkspacePropertyWithIndexTest *createSuite() {
+    return new WorkspacePropertyWithIndexTest();
+  }
+  static void destroySuite(WorkspacePropertyWithIndexTest *suite) {
+    delete suite;
+  }
+
+  void testConstructor() {
```
OwenArnold: I think there are some simpler tests that aren't implemented here around object behaviour. For example, `operator==` is overloaded, but I don't see an explicit test for that. Would be good to test that `operator!=` does what you expect too.

MantidQt/API/src/WorkspaceIndexPropertyWidget.cpp
```diff
@@ -0,0 +1,73 @@
...
+          SLOT(userEditedProperty()));
+}
+
+QString WorkspaceIndexPropertyWidget::getValue() const {
+  return m_workspaceWidget->getValue() + ";" + m_indexTypeWidget->getValue() +
+         ";" + m_indexListWidget->getValue();
+}
+
+void WorkspaceIndexPropertyWidget::setValueImpl(const QString &value) {
+  auto values = value.split(";");
```
OwenArnold: Maybe you should make this ";" some kind of static const named token. You use it in a lot of places.

MantidQt/API/src/WorkspaceIndexPropertyWidget.cpp
```diff
@@ -0,0 +1,73 @@
...
+namespace API {
+
+WorkspaceIndexPropertyWidget::WorkspaceIndexPropertyWidget(
+    Mantid::Kernel::Property *prop, QWidget *parent, QGridLayout *layout,
+    int row)
+    : PropertyWidget(prop, parent, layout, row) {
+  m_groupBox = new QGroupBox(parent);
+  layout->addWidget(m_groupBox, row, 0, 1, 2);
+
+  auto wProp = dynamic_cast<Mantid::API::IWorkspacePropertyWithIndex *>(prop);
```
OwenArnold: Do we need to be concerned that this down cast could fail?

MantidQt/API/src/WorkspaceIndexPropertyWidget.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+  if (values.isEmpty()) {
+    m_workspaceWidget->setValue("");
+    m_indexListWidget->setValue("");
+    m_indexTypeWidget->setValue("");
+  } else if (values.size() == 3) {
+    m_workspaceWidget->setValue(values[0]);
+    m_indexListWidget->setValue(values[1]);
+    m_indexTypeWidget->setValue(values[2]);
+  }
```
OwenArnold: Presumably 0 or 3 (split items) are the only values we expect? Is there a good guarantee around this?

OwenArnold (commented):
Minor changes needed. Also had some questions in there.

SimonHeybrock:
https://github.com/mantidproject/mantid/pull/19940 replaced this PR, closing.

PR #19728, opened 2017-05-29
GroupDetectors v2 TimeIndex support

### LINKED ISSUE #19679: Support time indexes in GroupDetectors v2
There are tickets open that have recently added support for scanning workspaces, in #19198 and #19380. `GroupDetectors2` support is useful for testing these work-spaces further, so this ticket addresses adding the required support to the algorithm.

ianbush:
GroupDetectors v2 will now preserver time indexes correctly when merging.

**To test:**

Make use of the changes in PR #19398. Use `CreateSampleWorkspace` to create a workspace with time indexes. Try out different methods of grouping together detectors, and check time indexes are set correctly.

Fixes #19679.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


SimonHeybrock:
@ianbush Did you have a look if https://github.com/mantidproject/mantid/blob/master/Framework/Indexing/src/Group.cpp can dothe job here? It may not cover some of the special handling done here, but maybe we canmake it fit here.

ianbush:
@SimonHeybrock I have just pushed code to make use of `Group.cpp` - it seems to work nicely and the code looks much cleaner now.

docs/source/release/v3.11.0/framework.rst
```diff
@@ -15,6 +15,7 @@ New
 Improved
 ########
 
+- :ref:`GroupDetectors <algm-GroupDetectors-v2>` now supports workspaces with time indexes.
```
SimonHeybrock: Maybe the release notes should refer to scanning/moving detectors, since "time indices" is not meaningful for users (or even other developers).

Framework/DataHandling/inc/MantidDataHandling/GroupDetectors2.h
```diff
@@ -205,7 +205,8 @@ class DLLExport GroupDetectors2 : public API::Algorithm {
   /// Copy the and combine the histograms that the user requested from the input
   /// into the output workspace
   size_t formGroups(API::MatrixWorkspace_const_sptr inputWS,
-                    API::MatrixWorkspace_sptr outputWS, const double prog4Copy);
+                    API::MatrixWorkspace_sptr outputWS, const double prog4Copy,
+                    const bool keepAll, const std::set<int64_t> &unGroupedSet);
   /// Copy the and combine the event lists that the user requested from the
   /// input into the output workspace
   size_t formGroupsEvent(DataObjects::EventWorkspace_const_sptr inputWS,
```
SimonHeybrock: Should the same changes also be done for `formGroupsEvent`?

Edit: Actually that might be tricky, since `EventList::operator+=` handles merging of detector ID groups. So maybe it is ok to not handle that case for now. Should we at least throw an exception?

Framework/DataHandling/src/GroupDetectors2.cpp
```diff
@@ -307,7 +310,8 @@ void GroupDetectors2::exec() {
            : 1.);
 
   // Build a new map
-  const size_t outIndex = formGroups(inputWS, outputWS, prog4Copy);
+  const size_t outIndex =
+      formGroups(inputWS, outputWS, prog4Copy, keepAll, unGroupedSet);
 
   // If we're keeping ungrouped spectra
   if (keepAll) {
```
SimonHeybrock: Can you check if the `moveOthers` call is overwriting the grouping defined by `IndexInfo`? I think it might.

SimonHeybrock (changes requested):
A couple of comments:

Framework/DataHandling/inc/MantidDataHandling/GroupDetectors2.h
ianbush: I agree we should thrown an exception for that case.

Framework/DataHandling/src/GroupDetectors2.cpp
ianbush: `moveOthers` under the hood sets the output spectra to the input spectra:
```
outputWS.getSpectrum(outIndex) = inputWS.getSpectrum(sourceIndex);
```

This I guess would overwrite the grouping, but should not change the spectrum numbers you end up with.

I have move the setting of `IndexInfo` to after `moveOthers`, as I think that should make more sense.

Framework/DataHandling/src/GroupDetectors2.cpp
```diff
@@ -1155,6 +1158,10 @@ size_t
 GroupDetectors2::formGroupsEvent(DataObjects::EventWorkspace_const_sptr inputWS,
                                  DataObjects::EventWorkspace_sptr outputWS,
                                  const double prog4Copy) {
+  if (inputWS->detectorInfo().isScanning())
+    throw std::runtime_error("GroupDetectors does not currently support "
+                             "workspaces with detector scans.");
```
SimonHeybrock: Could you add a unit test for this?

Also, maybe explicitly mention `EventWorkspace` in the error message?

ianbush: As discussed on Slack this is now added, using `create<EventWorkspace>()` worked perfectly.

SimonHeybrock (approved):

PR #19730, opened 2017-05-29
Primitive of StretchedExpFT

### LINKED ISSUE #19473: Primitive of StretchedExpFT
Fit function StretchedExpFT may have a sharp maximum at `E=0` with a fast decay. In some cases, the decay can be significant within the bin width:

In the plot above, we have evaluated StretchedExpFT in the [-5, 5] range with 11 points (black dots) and 1000 points (red curve). The area under the red curve can be well approximated by the area of the rectangles (the integral using the black dots) except for the central rectangle. There, the area under the rectangle largely overestimates the area under the red curve.

As a result, the [Rienmnan sum](https://en.wikipedia.org/wiki/Riemann_sum) for the convolution of the resolution function and a peaky StretchedExpFT will overestimate the elastic line. In such cases, it is more accurate to use the primitive of the StretchedExpFT evaluated at the bin boundaries. See Appendix B of [Wuttke 2012](http://www.mdpi.com/1999-4893/5/4/604) for a detailed discussion.

jmborr:
**To test:**
Download file [helper.zip](https://github.com/mantidproject/mantid/files/1036213/helper.zip)  and unload to `/tmp`, then run the following script in MantidPlot:

```
import numpy as np

LoadNexus('/tmp/helper.nxs', OutputWorkspace='helper')

def getW(maxE, deltaE, name, func='StretchedExpFT'):
    '''
    Creates a workspace by evaluating function StretchedExpFT or PrimStretchedExpFT
    '''
    x = np.arange(-maxE, maxE+deltaE, deltaE)
    N = len(x)
    y = np.ones(len(x)-1)
    e = y*0.1
    print x
    CreateWorkspace(x, y, e, NSpec=1, UnitX='DeltaE', OutputWorkspace=name)
    function='name={},Height=1,Tau=2000.0,Beta=0.5,Centre=0'.format(func)
    EvaluateFunction(Function=function, InputWorkspace=name, EvaluationType='CentrePoint', OutputWorkspace='_junk')
    mtd[name].dataY(0)[:] = deltaE*mtd['_junk'].dataY(1)

# We evaluate function StretchedExpFT at the center of each bin
center_evaluated = getW(0.0034, 0.0004, 'center_evaluated')
# We evaluate function PrimStretchedExpFT at the center of each bin, which
# amounts to integrating StretchedExpFT within each bin
bin_integrated = getW(0.0034, 0.0004, 'bin_integrated', func='PrimStretchedExpFT')
# Calculate the integral of StretchedExpFT in the central bin [-0.0002,0.0002] with
# numerical integration of StretchedExpFT evaluated in a very fine binning (0.00001meV)
fine_binning = getW(0.0034, 0.00001, 'fine_binning')
Integration(InputWorkspace='fine_binning', RangeLower=-0.0002, RangeUpper=0.0002,
    OutputWorkspace='central_bin_integrated')
```

If you plot workspaces `center_evaluated`, `bin_integrated`, and `central_bin_integrated`, <b>you should see something like this</b>:
<center></center>

Explanation of the picture:
​In this picture there is an evaluation of function StretcheExpFT  with the typical binning of BASIS (0.4ueV), workspace `center_evaluated`. One should always compare the integral of ​the model within each bin against experimental QENS data, because the data is a histogram. However, we usually compare the value that the model takes at the center of each bin, instead of the integral. This approximation is OK if the model does not change much within each bin. This is specially true at high energies when StretcheExpFT is very smooth.
Sometimes the center-of-the-bin assumption is wrong, like in the central bin [-0.2, 0.2]ueV. The integral of StretcheExpFT in this bin is the red dot, and the value of the model at the center of the bin is the black dot at the top. Very different.
Finally, the blue line is the evaluation of PrimStretcheExpFT at each bin center, which coincides with the fat red dot for the central bin. Also you can notice it agrees with StretcheExpFT at high energies when the StretcheExpFT is smooth and the approximation of the model at the center-of-the-bin is valid.

Fixes #19473

[RELEASE NOTES](https://github.com/mantidproject/mantid/blob/8f9aad00a704b956f0428898cdfef775fe609bc5/docs/source/release/v3.11.0/framework.rst#id1)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/PythonInterface/plugins/functions/PrimStretchedExpFT.py
```diff
@@ -0,0 +1,181 @@
+# pylint: disable=invalid-name, anomalous-backslash-in-string, attribute-defined-outside-init
+
+'''
+@author Jose Borreguero, NScD
+@date October 06, 2013
```
AntonPiccardoSelg: Why is the date 2013?

docs/source/fitfunctions/PrimStretchedExpFT.rst
```diff
@@ -0,0 +1,123 @@
...
+within each bin. This is specially true at high energies when StretcheExpFT is very smooth.
+Sometimes this center-of-the-bin evaluation is wrong, like in the central bin [-0.2, 0.2]ueV.
+The integral of :math:`StretchedExpFT` in this bin is the red dot, and the value of
+:math:`StretchedExpFT` at the center of the bin is the black dot at the top. Very different!
+This situation is likely to arise for samples with a relaxation rate :math:`h/Tau \ll \delta E`.
+Finally, the blue line is the evaluation of :math:`PrimStretchedExpFT` at each bin center, which
+by design amounts to the integral of :math:`StretchedExpFT` within each bin. This curve
+coincides with the red dot for the central bin, as expected. Also you it agrees with the
+center-of-bin evaluation of :math:`StretchedExpFT` at high energies, when the center-of-the-bin
+approximation is acceptable.
```
AntonPiccardoSelg: Good documentation :+1:

AntonPiccardoSelg (changes requested):
Most of the logic here seems to be a duplicate of `StretchedExpFT`, therefore we should try to share its code. A template method based approach should be suitable here.

We should add unit tests for `PrimStretchedExpFT` too.

Framework/PythonInterface/plugins/functions/PrimStretchedExpFT.py
```diff
@@ -0,0 +1,181 @@
...
+            return np.zeros(len(xvals), dtype=float)
+        # override with optparms (used for the numerical derivative)
+        if optparms:
+            for name in optparms.keys():
+                p[name] = optparms[name]
+
+        ne = len(xvals)
+        # energy spacing. Assumed xvals is a single-segment grid
+        # of increasing energy values
+        refine_factor = 16.0
```
jmborr: refine_factor effectively partitions each energy bin, so that we can calculate the integrand over a series of points within each energy bin. These values will later be summed up to calculate the integral. The bigger refine_factor, the more accurate the integral and the slower the evaluation of the whole fit function. The current value is a good compromise.

jmborr:
@AntonPiccardoSelg You're right both StretchedExpFT and this function share much of the code. I think a more pythonic solution would be to use a mixin class to host the common code, instead of the method template.

Framework/PythonInterface/plugins/functions/PrimStretchedExpFT.py
```diff
@@ -0,0 +1,181 @@
...
+        self._parmList = ['Height', 'Tau', 'Beta', 'Centre']
+
+    def validateParams(self):
+        """Check parameters are positive"""
+        height = self.getParameterValue('Height')
+        tau = self.getParameterValue('Tau')
+        beta = self.getParameterValue('Beta')
+        Centre = self.getParameterValue('Centre')
+
+        for _, value in {'Height': height, 'Tau': tau, 'Beta': beta}.items():
```
jmborr: I'll be dammed, why did I use a dictionary, indeed :dizzy_face: ?

Framework/PythonInterface/plugins/functions/PrimStretchedExpFT.py
jmborr: Incompetence, I'd have to say :bowtie:

jmborr:
@AntonPiccardoSelg I think I addressed all your concerns. Can you please verify?

AntonPiccardoSelg (changes requested):
Changes look good. The refactoring is nice, we save a lot of duplication. :+1:

The output is as described: 
image

One minor thing which we could clean up: There are two unused imports in one of the test modules.

Framework/PythonInterface/test/python/plugins/functions/StretchedExpFTTestHelper.py
```diff
@@ -0,0 +1,116 @@
+from __future__ import (absolute_import, division, print_function)
+
+import numpy as np
+from scipy import constants
+from scipy.special import erfc
+
+from mantid.simpleapi import Fit, CreateWorkspace, SaveNexus, SaveAscii, EvaluateFunction
+from mantid.api import mtd, FunctionFactory
+from pdb import set_trace as tr
```
jmborr: good catches. I forgot to remove these

AntonPiccardoSelg (approved):

PR #19731, opened 2017-05-30
Reduce synchronization overhead in LoadNexusProcessed

### LINKED ISSUE #19727: Synchronization overhead in LoadNexusProcessed
### Expected behavior

Execution time in parallel loops should decrease with the number of core. 

### Actual behavior

Lots of time is spent waiting on a mutex in `EventWorksapceMRU::deleteIndex`. Is there anything more we can do about this?

image

I found this in #19726 by running LoadNexus on CORELLI data. @rosswhitfield @AndreiSavici Do we have a similar file in SystemTestData or StandardTestData?

### Platforms affected

All

quantumsteve:
Description of work.

This uses Poco::RWLock to avoid excess synchronization overhead in LoadNexusProcessed and possibly other algorithms using EventWorkspace. My goal is to use the RWLock to give exclusive access only when the number of buffers is being modified. A separate mutex in each MRUList synchronizes access to each array.

I tried improving the thread-safety of `Kernel::cow_ptr<T>`. An alternative might be to added another member function to `MRUList::find` that copies the `cow_ptr` while the mutex is locked.

The implementation of `boost::shared_mutex` is "suboptimal"  (see [stackoverflow](https://stackoverflow.com/questions/37303342/boostshared-mutex-slower-than-a-coarse-stdmutex-on-linux) and [trac](https://svn.boost.org/trac/boost/ticket/11798)). I tried using it and the results were slower than `std::mutex`. Until C++14 or later is available everywhere, Poco is the best available option.

running `LoadNexus` with /SNS/CORELLI/IPTS-17467/shared/autoreduce/CORELLI_34682_elastic.nxs

master:
LoadNexus successful, Duration 6.70 seconds
LoadNexus successful, Duration 4.60 seconds
LoadNexus successful, Duration 4.50 seconds
LoadNexus successful, Duration 4.40 seconds
LoadNexus successful, Duration 4.31 seconds

this branch:
LoadNexus-[Notice] LoadNexus successful, Duration 4.57 seconds
LoadNexus-[Notice] LoadNexus successful, Duration 1.86 seconds
LoadNexus-[Notice] LoadNexus successful, Duration 2.02 seconds
LoadNexus-[Notice] LoadNexus successful, Duration 1.85 seconds
LoadNexus-[Notice] LoadNexus successful, Duration 2.02 seconds
LoadNexus-[Notice] LoadNexus successful, Duration 1.93 seconds

**To test:**

Fixes #19727.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/DataObjects/src/EventWorkspaceMRU.cpp
+++ b/Framework/DataObjects/src/EventWorkspaceMRU.cpp
@@ -6,13 +6,16 @@ namespace DataObjects {
 
 EventWorkspaceMRU::~EventWorkspaceMRU() {
   // Make sure you free up the memory in the MRUs
-  for (auto &data : m_bufferedDataY) {
-    if (data) {
-      data->clear();
-      delete data;
+  {
+    Poco::ScopedWriteRWLock _lock(m_changeMruListsMutexY);
+    for (auto &data : m_bufferedDataY) {
+      if (data) {
+        data->clear();
+        delete data;
+      }
     }
   }
-
+  Poco::ScopedWriteRWLock _lock2(m_changeMruListsMutexE);
   for (auto &data : m_bufferedDataE) {
     if (data) {
       data->clear();
@@ -27,7 +30,7 @@ EventWorkspaceMRU::~EventWorkspaceMRU() {
  * @param thread_num :: thread number that wants a MRU buffer
  */
 void EventWorkspaceMRU::ensureEnoughBuffersE(size_t thread_num) const {
-  std::lock_guard<std::mutex> _lock(m_changeMruListsMutexE);
+  Poco::ScopedWriteRWLock _lock(m_changeMruListsMutexE);
   if (m_bufferedDataE.size() <= thread_num) {
     m_bufferedDataE.resize(thread_num + 1, nullptr);
     for (auto &data : m_bufferedDataE) {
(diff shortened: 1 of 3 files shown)


AndreiSavici (approved):
Tried with `/SNS/SEQ/IPTS-18590/shared/autoreduce/SEQ_130169_autoreduced.nxs` Initial was 5.5/3.5/4.1 seconds. After the change 4.8/5.1/4.7 seconds. For this particular file it seems that it slowed down. But I think that is over the sshfs. Saving a local copy reduces load time to 1.5 seconds, but that is no different from master loading the local file

Framework/DataObjects/src/EventWorkspaceMRU.cpp
```diff
@@ -139,18 +139,22 @@ void EventWorkspaceMRU::insertE(size_t thread_num, EType data,
  * @param index :: index to delete.
  */
 void EventWorkspaceMRU::deleteIndex(const EventList *index) {
-  {
+  // Avoid locks when the buffers are empty
+  // Can significantly speed up EventList::mutableHistogramRef()
+  if (!m_bufferedDataE.empty()) {
```
SimonHeybrock: Another thread might be writing to `m_bufferedDataE` at the same time, so I am not sure why we can do this outside the lock?

NickDraper:
Given Simon's comments, and PR came in after the code freeze, I'm removing the Leave me in this release label and moving it out to R3.11.

If there is an urgent need (rather than opportunity) to get it into this release is should be discussed and agreed on the TSC slack channel.

I think this is good work, and gives a real benefit, but it carries some risk and to some of our core functionality.

quantumsteve:
Let's leave this in the 3.11 milestone where we have more time to review and test.

SimonHeybrock:
Should we go forward with merging this? The changes look reasonable to me.

NickDraper:
Yes, now is a good time to merge this in.

NickDraper:
The reviews have passed, and the improvement is significant.

PR #19736, opened 2017-05-30
Resolve crash with double clicking table in Muon GUI

### LINKED ISSUE #19701: Muon crashes in  grouping tab
### Expected behavior
In muon analysis under the groupings tab, Mantid should handle unexpected data gracefully. 
### Actual behavior
Mantid crashes under certain circumstances. 
In the grouping tab of muon analysis enter a name into the next available row (directly under the filled ones). Then double click (it does not crash for a single click) the detectors column of the row below the one you have just named. This will generate a crash 
### Platforms affected
At least windows.

DavidFair:
Description of work.
This PR resolves a crash where Mantid would crash within the Muon GUI. Previously when `GroupTableChanged` was called it implicitly updates the selected bank and replotted this. 
If the user was to click the table again (such as a double click in another box) Qt would crash if it hadn't finished running the plotting code.  

To resolve this I've separated the code to update the index and plot for clarity. I've also added a flag to indicate if we want the selected group (on the home page) to change which will implicitly plot that group instead. 

@AnthonyLim23 - It's worth noting with these changes the currently selected bank on the home page will **not** automatically update to the most recent bank. This might initially catch users out who relied on this implicit behaviour.

**To test:**
- Open the Muon GUI
- Load a Muon run (any will do such as one from HIFI)
- Go to the grouping window
- Enter a new name 
- Click of the box so the new name is entered
- Go back to the new name, delete the text
- Without clicking anything else double click (must be double) the detector ID box
- If it no longer crashes this test passed.

Fixes #19701.

*Does not need to be in the release notes. - Has only been noticed in unscripted testing*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Muon/MuonAnalysis.h
+++ b/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Muon/MuonAnalysis.h
@@ -87,12 +87,12 @@ public:
   /// Destructor
   ~MuonAnalysis();
 
-  /// Sets index of group or pair to plot
-  /// and causes a replot
-  void setGroupOrPairAndReplot(int index);
-
   /// Gets current index of group or pair to plot
   int getGroupOrPairToPlot() const;
+
+  void setGroupOrPairIndexToPlot(int index);
+
+  void plotCurrentGroupAndPairs();
 
 signals:
   /// Request to hide/show Mantid toolbars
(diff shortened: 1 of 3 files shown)


AnthonyLim23:
I think we should put this in the release notes as it is different from previous versions. @DavidFair  would you like to put something in or should I?

DavidFair:
@AnthonyLim23 - If you could as you'll have a better grasp on the appropriate language for your users.

AnthonyLim23:
I will add something later, after testing it. I will run it by you too to make sure it covers everything

AnthonyLim23:
once it has passed the tests, :shipit:

PR #19740, opened 2017-05-30
Saved instrument view shapes crash mantid.

### LINKED ISSUE #19725: Instrument view saves shapes into a workspace that may crash mantid.
### Expected behavior

Uncertain but definitely shouldn't crash.

### Actual behavior

The table workspace created by the [Save Shapes to Table] button on the Draw tab doesn't contain the information about the shapes. It is unclear what to do with the workspace but dragging and dropping it to the Instrument view crashes mantid.

* Open instrument view.
* Draw some shapes.
* Save them to a table workspace.
* Double click on the created workspace to open a table view.
* Drag and drop the table workspace to the instrument view.
* Crash

### Platforms affected

All

mantid-roman:
An open table view truncate the parameter string for a shape. This leads to a null pointer returned by a factory method.

Added a check for a null pointer.

**To test:**

Instructions are in the issue description.

Fixes #19725.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidQt/MantidWidgets/src/InstrumentView/Shape2DCollection.cpp
+++ b/MantidQt/MantidWidgets/src/InstrumentView/Shape2DCollection.cpp
@@ -685,7 +685,9 @@ void Shape2DCollection::loadFromTableWorkspace(
   for (size_t i = 0; i < ws->rowCount(); ++i) {
     const auto params = col[i];
     auto shape = Shape2D::loadFromProject(params);
-    m_shapes.append(shape);
+    if (shape != nullptr) {
+      m_shapes.append(shape);
+    }
   }
   emit shapeCreated();
 }


MantidQt/MantidWidgets/src/InstrumentView/Shape2DCollection.cpp
```diff
@@ -685,7 +685,9 @@ void Shape2DCollection::loadFromTableWorkspace(
   for (size_t i = 0; i < ws->rowCount(); ++i) {
     const auto params = col[i];
     auto shape = Shape2D::loadFromProject(params);
-    m_shapes.append(shape);
+    if (shape != nullptr) {
```
DavidFair: This can be done as
`if (shape)`

mantid-roman: My thinking for doing it was:
1. According to mantid C++ coding standards: *Integers should not be cast to Booleans.* Raw pointers can be considered integers.
2. `auto shape` hides the actual type of variable `shape`. Comparing it with `nullptr` shows its type clearly, no need to investigate further what `if (shape)` means.

DavidFair: Ahh from the `auto` type I assumed this would be a pointer type. 
If it could be integer as implied with point 1. I agree this check should be against a `nullptr` type.

DavidFair:
This PR fixes the crash described in the issue. :shipit:

PR #19741, opened 2017-05-30
Swap order of statements to avoid race condition in Slice Viewer

martyngigg:
**Description of work**

Swaps the order of adding observers to an algorithm in the slice viewer to avoid a race condition when the algorithm starts asynchronously. This fixes the unreliable tests observed on the build servers.

**To test:**

I could only get this to fail originally on a RHEL 7 or OS X machine by running

`ctest -j8  --schedule-random --output-on-failure --repeat-until-fail 1000 -R SliceViewerMantidPlotTest_`

where `8` is the number of cores on the machine. Before this fix the test would fall over within the first 50 attempts. After this fix the test ran for ~2 hours without failure before I killed it to free up the server time. 

*Clears up random failure that users never saw. Does not need to be in the release notes.*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidQt/API/src/AlgorithmRunner.cpp
+++ b/MantidQt/API/src/AlgorithmRunner.cpp
@@ -72,14 +72,13 @@ void AlgorithmRunner::startAlgorithm(Mantid::API::IAlgorithm_sptr alg) {
 
   cancelRunningAlgorithm();
 
-  // Start asynchronous execution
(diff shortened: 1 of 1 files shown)


DavidFair:
I've ran this for 3 hours on RHEL7 and haven't seen any failures either. :shipit:

PR #19746, opened 2017-05-30
ComponentInfo positions and rotations version 2

### LINKED ISSUE #19278: ComponentInfo with readable positions and roatations

OwenArnold:
Fixes #19278. Puts positions and rotations into the `Beamline::ComponentInfo`. detector indices forwarded to the `Beamline::DetectorInfo`. 

Supersedes #19279

**To test:**

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/API/inc/MantidAPI/ComponentInfo.h
```diff
@@ -2,6 +2,8 @@
 #define MANTID_API_COMPONENTINFO_H_
 
 #include "MantidAPI/DllConfig.h"
+#include "MantidKernel/V3D.h"
+#include "MantidKernel/Quat.h"
```
SimonHeybrock: Maybe just forward declare?

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -256,10 +257,12 @@ makeDetectorInfo(const Instrument &oldInstr, const Instrument &newInstr) {
...
  * via registerContents.
  */
-void ExperimentInfo::makeAPIComponentInfo(const InfoComponentVisitor &visitor) {
+void ExperimentInfo::makeAPIComponentInfo(
+    const InfoComponentVisitor &visitor,
+    boost::shared_ptr<Beamline::DetectorInfo> detectorInfo) {
 
   // Internal Beamline ComponentInfo
-  m_componentInfo = visitor.componentInfo();
+  m_componentInfo = visitor.componentInfo(detectorInfo);
```
SimonHeybrock: I think as a follow up cleanup we can then move `DetectorInfo` into the visitor?

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -270,19 +273,24 @@ void ExperimentInfo::makeAPIComponentInfo(const InfoComponentVisitor &visitor) {
 /**
  * Take the visitor from the instrument if it's already there or make from
  * scratch.
- * @param instrument : Instrument which might carry a visitor.
+ * @param parInstrument : Parameterised instrument.
+ * @param newInstrument : unparameterised new instrument
```
SimonHeybrock: Isn't it in general parameterised?

Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
```diff
@@ -39,19 +41,51 @@ class MANTID_BEAMLINE_DLL ComponentInfo {
...
+                boost::shared_ptr<const std::vector<size_t>>
+                    assemblySortedComponentIndices,
+                boost::shared_ptr<const std::vector<std::pair<size_t, size_t>>>
+                    componentRanges,
+                boost::shared_ptr<std::vector<Eigen::Vector3d>> positions,
+                boost::shared_ptr<std::vector<Eigen::Quaterniond>> rotations,
+                boost::shared_ptr<DetectorInfo> detectorInfo);
+
   std::vector<size_t> detectorIndices(const size_t componentIndex) const;
+  std::vector<size_t> componentIndices(const size_t componentIndex) const;
```
SimonHeybrock: Is this name clear enough? Maybe consider if there are better options? `subcomponentIndices()` (not sure I like it) or `componentsInSubtree()`?

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -14,38 +15,179 @@ ComponentInfo::ComponentInfo()
...
-      m_size(m_assemblySortedDetectorIndices->size() + m_ranges->size()) {}
+      m_assemblySortedComponentIndices(
+          std::move(assemblySortedComponentIndices)),
+      m_detectorRanges(std::move(detectorRanges)),
+      m_componentRanges(std::move(componentRanges)),
+      m_positions(std::move(positions)), m_rotations(std::move(rotations)),
+      m_size(m_assemblySortedDetectorIndices->size() +
+             m_detectorRanges->size()),
+      m_detectorInfo(detectorInfo) {
+  if (m_rotations->size() != m_positions->size()) {
```
SimonHeybrock: If `ComponentInfo` is always created by the visitor, maybe we could consider removing all those checks?

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -14,38 +15,179 @@ ComponentInfo::ComponentInfo()
...
+
+  // Extract as a block
+  return std::vector<size_t>(
+      m_assemblySortedComponentIndices->begin() + range.first,
+      m_assemblySortedComponentIndices->begin() + range.second);
+}
+
 size_t ComponentInfo::size() const { return m_size; }
 
 bool ComponentInfo::operator==(const ComponentInfo &other) const {
```
SimonHeybrock: I think these comparison operators should be removed now that we store positions/rotations.

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -14,38 +15,179 @@ ComponentInfo::ComponentInfo()
...
+  const Eigen::Vector3d offset = newPosition - position(componentIndex);
+  const auto indices = this->componentIndices(
+      componentIndex); // Includes requested component index
+  for (auto &index : indices) {
+    if (isDetector(index)) {
+      m_detectorInfo->setPosition(index,
+                                  m_detectorInfo->position(index) + offset);
+    } else {
+      m_positions.access()[compOffsetIndex(index)] += offset;
+      // TODO. This could change L1/L2. Is that being detected?
```
SimonHeybrock: Now that we are storing the position of source and sample in `ComponentInfo` the positions access is fast and thus the L1/L2 caching mechanism should be removed from `DetectorInfo`.

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -14,38 +15,179 @@ ComponentInfo::ComponentInfo()
...
+                                const Eigen::Quaterniond &newRotation) {
+
+  using namespace Eigen;
+  const Eigen::Vector3d compPos = position(componentIndex);
+  const Eigen::Quaterniond currentRotInv = rotation(componentIndex).inverse();
+  const Eigen::Quaterniond rotDelta =
+      (newRotation * currentRotInv).normalized();
+
+  const Affine3d transform =
+      Translation3d(compPos) * rotDelta * Translation3d(-compPos);
```
SimonHeybrock: I think I had tried using `Affine3d` but in the end it was faster to use a manual position shift with a `Matrix3d` (see `API::DetectorInfo`). This is performance critical for `SCDCalibratePanels`, but maybe you can double check.

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -14,38 +15,179 @@ ComponentInfo::ComponentInfo()
...
+    if (isDetector(index)) {
+      m_detectorInfo->setPosition(index, newPos);
+      m_detectorInfo->setRotation(index, newRot);
+
+    } else {
+      const size_t childCompIndexOffset = compOffsetIndex(index);
+      m_positions.access()[childCompIndexOffset] = newPos;
+      m_rotations.access()[childCompIndexOffset] = newRot;
+
+      // TODO. This could change L1/L2. Is that being detected?
```
SimonHeybrock: See above.

Framework/Geometry/inc/MantidGeometry/Instrument/CacheComponentVisitor.h
```diff
@@ -0,0 +1,54 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+class MANTID_GEOMETRY_DLL CacheComponentVisitor
```
SimonHeybrock: Is this unused right now?

SimonHeybrock (commented):
Looks good so far, some small comments below.

Regarding your question on Slack regarding finishing this PR: I do not really see a way to split this. Now that you extra positions from the parameter map when setting the instrument you must also do the other bits, namely extract positions in `ExperimentInfo::populateInstrumentParameters()` and `ExperimentInfo::readParameterMap()`, and reinsert them in `Instrument::makeLegacyParameterMap()`. Also double check my PR for `DetectorInfo` where I did those bits, there might be one or two other small changes required.

You also need to change `DetectorInfo::setPosition(IComponent *)` to use `ComponentInfo` (or rather, remove the method from `DetectorInfo` and use `ComponentInfo` directly).

Framework/API/src/ExperimentInfo.cpp
OwenArnold: Yes, this is absolutely the plan. That would give the visitor just two relevant methods `componentInfo()` and `detectorInfo()`. But that's too large of a change to incorporate into this, so as you say, follow up.

Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
OwenArnold: Interesting. I did consider `subTree` in the name, but it also includes the original `componentIndex` argument in the full returns. This is because the client-code often needs to loop over everything including the current `componentIndex`.  Maybe `subTree` is still a better alternative to what I have?

Framework/Beamline/src/ComponentInfo.cpp
OwenArnold: If we do that, I would like to make the constructor `private` and possibly make declare `InfoComponentVisitor` as a `friend`. For the time being, I'd be happier keeping this in as it provides a useful sanity check on the inputs.

Framework/Beamline/src/ComponentInfo.cpp
OwenArnold: Agreed.

Framework/API/src/ExperimentInfo.cpp
OwenArnold: Probably. But ExperimentInfo::setInstrument also supports arguments where the Instrument is not parameterised. `InfoComponentVisitor` needs a ParameterMap, which it purges of information as it parses through the instrument tree.

Framework/API/inc/MantidAPI/ComponentInfo.h
OwenArnold: position() and rotation() return by value (as they do in `DetectorInfo`) so unfortunately we cannot forward declare.

Framework/API/src/ExperimentInfo.cpp
OwenArnold: Yes. We would definitely want to do this.

Framework/Geometry/inc/MantidGeometry/Instrument/CacheComponentVisitor.h
OwenArnold: Yes. Currently unused. I don't think I need it, and I'll definitely strip it out if it serves no purpose for the ParameterMap/ComponentInfo interactions.

Framework/Beamline/src/ComponentInfo.cpp
OwenArnold: 👍

Framework/API/inc/MantidAPI/ComponentInfo.h
SimonHeybrock: I thought you could?

Framework/API/src/ExperimentInfo.cpp
SimonHeybrock: Ok, I was just confused by the doxygen.

Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
SimonHeybrock: I think I would expect having the original index included -- if it was not included the result would be a bush, not a tree.

For consistency, should then also `detectorIndices` be renamed?

```cpp
detectorsInSubtree();
componentsInSubtree();
```

I am not 100% happy with this name either, but I feel it is slightly better. Maybe the main decision is whether we want to keep referring to "index" and "indices" all the time? Basically, is it obvious what code like this does:

```cpp
for (const auto &det : componentInfo.detectorsInSubtree(comp)) {
  detectorInfo.setPosition(det, Vector3d(0,0,0));
}
```

Woult it be less or more readable like this:

```cpp
for (const auto &detIndex : componentInfo.detectorsIndices(compIndex)) {
  detectorInfo.setPosition(detIndex, Vector3d(0,0,0));
}
```

To keep this consistent, we might then also consider renaming the `indexOf` methods?

Framework/API/inc/MantidAPI/ComponentInfo.h
```diff
@@ -63,12 +63,10 @@ class MANTID_API_DLL ComponentInfo {
           componentIds,
       boost::shared_ptr<const std::unordered_map<
           Geometry::IComponent *, size_t>> componentIdToIndexMap);
-  std::vector<size_t> detectorIndices(size_t componentIndex) const;
-  std::vector<size_t> componentIndices(size_t componentIndex) const;
+  std::vector<size_t> detectorsInSubTree(size_t componentIndex) const;
+  std::vector<size_t> componentsInSubTree(size_t componentIndex) const;
```
SimonHeybrock: This is nitpicking, but maybe consider "subtree" as one word, i.e., use `detectorInSubtree`?

Framework/API/inc/MantidAPI/ComponentInfo.h
OwenArnold: I stand corrected. I think you're actually right about this.

Framework/API/inc/MantidAPI/ComponentInfo.h
OwenArnold: Agreed. I'll change that.

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -446,58 +446,33 @@ T getParam(const std::string &paramType, const std::string &paramValue) {
...
-  if (auto parent = parComp->getParent()) {
-    // Note the unusual order. This is what Component::getRotation does.
-    rotation = parent->getRotation() * newRelRot;
-  }
-  detectorInfo.setRotation(*parComp, rotation);
+void updatePosition(ComponentInfo &componentInfo, const IComponent *component,
+                    const V3D &newRelPos) {
+  const auto compIndex = componentInfo.indexOf(component->getComponentID());
+  componentInfo.setPosition(compIndex,
+                            componentInfo.position(compIndex) + newRelPos);
```
SimonHeybrock: I am not sure why you changed the old code? `newRelPos` is the position relative to the parent, in the parent coordinate system, i.e., what is stored in the parameter map and *not* the relative position in the *global* coordinate system.

Framework/API/src/ResizeRectangularDetectorHelper.cpp
```diff
@@ -1,37 +1,35 @@
...
   transformation.pretranslate(origin);
 
-  const auto idstart = detector.idstart();
-  const auto idstep = detector.idstep();
-  const auto count = detector.xpixels() * detector.ypixels();
-  for (detid_t id = idstart; id < idstart + idstep * count; ++id) {
-    const auto index = detectorInfo.indexOf(id);
-    const auto pos = Kernel::toVector3d(detectorInfo.position(index));
-    detectorInfo.setPosition(index, Kernel::toV3D(transformation * pos));
+  for (auto index : componentInfo.componentsInSubtree(componentIndex)) {
```
SimonHeybrock: Just use `detectorsInSubtree` and drop the `if`?

SimonHeybrock (commented):
Start looks good, just two comments:

Framework/API/src/ResizeRectangularDetectorHelper.cpp
OwenArnold: Yes. I had thought about that but wasn't 100% sure as the old code was just moving detectors. I think you're right though. I'll make the change.

Framework/API/src/ExperimentInfo.cpp
OwenArnold: Yes. That was a mistake.

OwenArnold: I'm going to add to `ComponentInfo`/ `InfoComponentVisitor` data allowing us to go from a componentIndex to the componentIndex of the parent. I need this in a few other places too.

Framework/Geometry/inc/MantidGeometry/Instrument.h
```diff
@@ -269,10 +280,22 @@ class MANTID_GEOMETRY_DLL Instrument : public CompAssembly {
   void appendPlottable(const CompAssembly &ca,
                        std::vector<IObjComponent_const_sptr> &lst) const;
 
+  /// Make or create a component id vector ordered in a bottom up specific way.
```
OwenArnold: Must have been tired when I wrote this function name :)

Framework/Geometry/src/Instrument/Component.cpp
```diff
@@ -393,6 +387,152 @@ Quat Component::getRotation() const {
...
+    root = parent;
+  auto instrument = dynamic_cast<const Instrument *>(root);
+  return m_map->hasComponentInfo(instrument);
+}
+
+/// Return the relative position to the parent Component
+Kernel::V3D Component::getRelativePos() const {
+  if (m_map) {
+
+    if (hasComponentInfo() && !m_map->componentInfo().isDetector(index())) {
```
SimonHeybrock: Why excluding detectors?

SimonHeybrock (commented):
I had a short look through this, not very thorough so I may have missed some bits:

Framework/Geometry/src/Instrument/Component.cpp
OwenArnold: Because this method will get called during the creation of `DetectorInfo`. i.e. `Detector::getPos()` But at that point the `DetectorInfo` won't exist and can't be used by the `ComponentInfo` so we have to do something different for `Detectors`. This is part of the clean-up that will be needed. i.e. making the `InfoComponentVisitor` create both `ComponentInfo` and `DetectorInfo` in one hit.

SimonHeybrock: But isn't the `DetectorInfo` build *before* the `ComponentInfo`?

SimonHeybrock: Oh, it used to be, but now you reverse the order. What was the reason for that?

OwenArnold: I initially thought that all old parameter map code in `Component::getPos` and `DetectorInfo::getPos` etc could be made redundant if we were to lean on `ComponentInfo` methods such as `ComponentInfo::getRelativeRotation`, which means establishing the `ComponentInfo` first. In reality, it doesn't matter. As we know that the `InfoComponentVisitor` can and will be used to parse and flush the whole instrument tree in one go (so we effectively get the `DetectorInfo` and `ComponentInfo` in one hit). I suggest that it doesn't matter at this stage as long as we can preserve working functionality in the intermediate builds.

Framework/Geometry/src/Instrument.cpp
```diff
@@ -1299,54 +1372,80 @@ boost::shared_ptr<ParameterMap> Instrument::makeLegacyParameterMap() const {
...
   Eigen::Affine3d transformation;
-  for (size_t i = 0; i < m_detectorInfo->size(); ++i) {
-    const auto &det = std::get<1>(baseInstr.m_detectorCache[i]);
-    if (m_detectorInfo->isMasked(i)) {
-      pmap->forceUnsafeSetMasked(det.get(), true);
-    }
 
-    // Obtain parent position/rotation/scale.
-    const auto parent = det->getParent();
-    if (parent.get() != oldParent) {
```
OwenArnold: Yes. I think this same functionality could be reintroduced just by taking the parent index and comparing it to the last old one.

Framework/Geometry/src/Instrument.cpp
```diff
@@ -1299,54 +1372,80 @@ boost::shared_ptr<ParameterMap> Instrument::makeLegacyParameterMap() const {
...
+        pmap->addV3D(m_componentCache[i], ParameterMap::pos(),
+                     Kernel::toV3D(relPos));
+      }
+      if ((relRot * toQuaterniond(baseComponent->getRelativeRot()).conjugate())
+              .vec()
+              .norm() >= imag_norm_max) {
+        pmap->addQuat(m_componentCache[i], ParameterMap::rot(),
+                      Kernel::toQuat(relRot));
+      }
+    }
```
OwenArnold: Yes. I just haven't refactored it to that yet.

Framework/Geometry/src/Instrument/Component.cpp
SimonHeybrock: Well, it makes it much harder to review. Have you considered doing everything in the visitor right now instead of waiting for a follow up PR?

Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
```diff
@@ -39,19 +41,66 @@ class MANTID_BEAMLINE_DLL ComponentInfo {
...
+
+  std::vector<size_t> detectorsInSubtree(const size_t componentIndex) const;
+  std::vector<size_t> componentsInSubtree(const size_t componentIndex) const;
   size_t size() const;
-  bool operator==(const ComponentInfo &other) const;
-  bool operator!=(const ComponentInfo &other) const;
+  inline bool isDetector(const size_t componentIndex) const {
+    return componentIndex < m_assemblySortedDetectorIndices->size();
+  }
+  inline size_t compOffsetIndex(const size_t componentIndex) const {
```
OwenArnold: Yes. You're right that's unnecessary.

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -10,42 +11,254 @@ namespace Beamline {
...
+  if (isDetector(componentIndex)) {
+    checkDetectorInfo();
+    return m_detectorInfo->rotation(componentIndex);
+  }
+  const auto rangesIndex = compOffsetIndex(componentIndex);
+  return (*m_rotations)[rangesIndex];
+}
+
+Eigen::Vector3d
+ComponentInfo::relativePosition(const size_t componentIndex) const {
```
OwenArnold: 👍 Yes it needs to work like it did/does for Component

Framework/API/inc/MantidAPI/ComponentInfo.h
```diff
@@ -81,6 +81,8 @@ class MANTID_API_DLL ComponentInfo {
   bool hasParent(const size_t componentIndex) const;
   Kernel::V3D sourcePosition() const;
   Kernel::V3D samplePosition() const;
+  size_t sourceIndex() const;
+  size_t sampleIndex() const;
```
SimonHeybrock: Maybe drop the `Index` part of the name, as we did for all the other methods like `parent()`?

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -111,16 +111,45 @@ Eigen::Quaterniond ComponentInfo::rotation(const size_t componentIndex) const {
...
+ * see Instrument::makeLegacyParmeterMap for correct handling of those cases.
+ *
+ * BEWARE of peformance on repeated calls to this method as the transformation
+ *has to
+ * be established every time, independent of whether the parent remains the
+ *same.
+ *
+ * @param componentIndex
+ * @return
+ */
```
SimonHeybrock: With so many pitfalls, maybe we want to remove this method (now or later), and definitely not provide Python exports?

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -111,16 +111,45 @@ Eigen::Quaterniond ComponentInfo::rotation(const size_t componentIndex) const {
...
   if (parentIndex == componentIndex) {
     return position(componentIndex);
   } else {
-    return position(componentIndex) - position(parentIndex);
+    const auto parentPos = position(parentIndex);
+    Eigen::Affine3d transformation;
+    transformation =
+        rotation(parentIndex).conjugate(); // Inverse parent rotation
+    transformation.translate(-parentPos);
+    return transformation * position(componentIndex);
```
SimonHeybrock: I have found that using `Affine3d` is slower than doing an offset + rotation. If you use this in performance critical code consider simply writing `invrot *(position - parentPos)`.

Framework/DataHandling/src/LoadIsawDetCal.cpp
```diff
@@ -431,20 +431,23 @@ void LoadIsawDetCal::doRotation(V3D rX, V3D rY, DetectorInfo &detectorInfo,
...
-    rot0.inverse();
-    Rot *= rot0;
-  }
-  const auto grandparent = parent->getParent();
-  if (grandparent) {
-    Quat rot0 = grandparent->getRelativeRot();
+  const auto componentIndex = componentInfo.indexOf(comp->getComponentID());
+
+  if (componentInfo.hasParent(componentIndex)) {
+    const auto parentIndex = componentInfo.parent(componentIndex);
```
SimonHeybrock: Maybe code like this gets more readable if we drop the `Index` everywhere?

```cpp
const auto parent = componentInfo.parent(component);
```

Unless we mix it with legacy-type pointers to `Component` it feels clear and natural to me. What do you think?

Framework/Geometry/src/Instrument/ParameterMap.cpp
```diff
@@ -1161,24 +1164,54 @@ bool ParameterMap::hasDetectorInfo(const Instrument *instrument) const {
...
+  return *m_componentInfo;
+}
+
 /// Only for use by Detector. Returns a detector index for a detector ID.
 size_t ParameterMap::detectorIndex(const detid_t detID) const {
   return m_instrument->detectorIndex(detID);
 }
 
+size_t ParameterMap::componentIndex(const ComponentID componentId) const {
+  return m_instrument->componentIndex(componentId);
```
SimonHeybrock: @OwenArnold This is what I meant on slack: Replace this line by
```cpp
return m_componentInfo->indexOf(componentId);
```
With that change a fair number of unit tests seems to pass (still many of them are segfaulting though).

All you need to do is move `API::ComponentInfo` into `Geometry`, I think (I just fudged it for my test by making `API::ComponentInfo.h` header only so I could include it in `Geometry` files).

Framework/Algorithms/src/ResizeRectangularDetector.cpp
```diff
@@ -97,19 +97,19 @@ void ResizeRectangularDetector::exec() {
     throw std::runtime_error("Component with name " + ComponentName +
                              " was not found.");
 
-  auto &componentInfo = inputW->mutableComponentInfo();
-  const auto componentIndex = componentInfo.indexOf(comp->getComponentID());
-  if (!componentInfo.isDetector(componentIndex))
+  RectangularDetector_const_sptr det =
+      boost::dynamic_pointer_cast<const RectangularDetector>(comp);
```
OwenArnold: Given backwards compatibility for cases like this, and some others in `makeLegacyParameterMap` it would be nice to have some way of `ComponentInfo::isRectangularDetectorPixel`. That way we could avoid the dynamic cast and we are visiting each of these RectangularDetectorPixels anyway in the construction. We would only need a `std::vector<bool>` to store that information.

OwenArnold: I wouldn't expect that to be done here though. It might be an additional refactoring step.

SimonHeybrock: Note that this is not about the pixel -- `RectangularDetector` is not a detector, it is a component assembly and the grandparent of the `RectangularDetectorPixels`.

OwenArnold: Yes agree. Maybe just recording the component indexes of the `RectangularDetectors` would be enough. That way a method could be provided that utilised something like `isDetector` and `parent(parent(index))` with the flag for the RectangularDetectors.

Framework/Geometry/src/Instrument/InfoComponentVisitor.cpp
```diff
@@ -27,73 +29,138 @@ makeDetIdToIndexMap(const std::vector<detid_t> &detIds) {
...
+  const size_t componentIndex = m_componentIds->size();
+  (*m_componentIdToIndexMap)[component.getComponentID()] = componentIndex;
   m_componentIds->emplace_back(component.getComponentID());
+  m_positions->emplace_back(Kernel::toVector3d(component.getPos()));
+  m_rotations->emplace_back(Kernel::toQuaterniond(component.getRotation()));
+  const size_t componentStart = m_assemblySortedComponentIndices->size();
+  m_componentRanges->emplace_back(
+      std::make_pair(componentStart, componentStart + 1));
+  m_assemblySortedComponentIndices->push_back(componentIndex);
+  m_parentComponentIndices->push_back(componentIndex);
```
OwenArnold: Yes you need to extend this vector, but as you are assigning the wrong index here (it is not known yet) maybe a comment would help.

Framework/Geometry/src/Instrument/InfoComponentVisitor.cpp
```diff
@@ -27,73 +29,138 @@ makeDetIdToIndexMap(const std::vector<detid_t> &detIds) {
...
+          std::move(orderedDetectorIds))),
+      m_positions(boost::make_shared<std::vector<Eigen::Vector3d>>()),
+      m_rotations(boost::make_shared<std::vector<Eigen::Quaterniond>>()),
+      m_pmap(pmap), m_sourceId(sourceId), m_sampleId(sampleId) {
   const auto nDetectors = m_orderedDetectorIds->size();
-  m_assemblySortedDetectorIndices->reserve(nDetectors);
-  m_componentIdToIndexMap->reserve(nDetectors);
+  m_assemblySortedDetectorIndices->reserve(nDetectors);  // Exact
+  m_assemblySortedComponentIndices->reserve(nDetectors); // Approximation
+  m_parentComponentIndices->reserve(nDetectors);         // Approximation
```
OwenArnold: This reserve is not just an optimisation now.

SimonHeybrock:
Several improvements that yield (I think) 10-20% performance improvement for `ComponentInfo::setRotation`:

- avoid allocations by `componentsInSubtree`
- split handling of detectors and other components to avoid branch in inner loop
- avoid `cow_ptr::access` in inner loop

```
diff --git a/Framework/Beamline/src/ComponentInfo.cpp b/Framework/Beamline/src/ComponentInfo.cpp
index a95b020..5358963 100644
--- a/Framework/Beamline/src/ComponentInfo.cpp
+++ b/Framework/Beamline/src/ComponentInfo.cpp
@@ -49,11 +49,11 @@ ComponentInfo::ComponentInfo(
                                 "and rotations as assembly sorted component "
                                 "ranges");
   }
-  if (m_assemblySortedComponentIndices->size() != m_size) {
-    throw std::invalid_argument("ComponentInfo must have component indices "
-                                "input of same size as the sum of "
-                                "non-detector and detector components");
-  }
+  //if (m_assemblySortedComponentIndices->size() != m_size) {
+  //  throw std::invalid_argument("ComponentInfo must have component indices "
+  //                              "input of same size as the sum of "
+  //                              "non-detector and detector components");
+  //}
 }
 
 std::vector<size_t>
@@ -95,7 +95,7 @@ size_t ComponentInfo::size() const { return m_size; }
 
 Eigen::Vector3d ComponentInfo::position(const size_t componentIndex) const {
   if (isDetector(componentIndex)) {
-    checkDetectorInfo();
+    //checkDetectorInfo();
     return m_detectorInfo->position(componentIndex);
   }
   const auto rangesIndex = compOffsetIndex(componentIndex);
@@ -104,7 +104,7 @@ Eigen::Vector3d ComponentInfo::position(const size_t componentIndex) const {
 
 Eigen::Quaterniond ComponentInfo::rotation(const size_t componentIndex) const {
   if (isDetector(componentIndex)) {
-    checkDetectorInfo();
+    //checkDetectorInfo();
     return m_detectorInfo->rotation(componentIndex);
   }
   const auto rangesIndex = compOffsetIndex(componentIndex);
@@ -176,13 +176,38 @@ void ComponentInfo::setPosition(const size_t componentIndex,
 
   scanningCheck(componentIndex);
   const Eigen::Vector3d offset = newPosition - position(componentIndex);
+  {
+  const auto range = (*m_detectorRanges)[rangesIndex];
+  auto it = m_assemblySortedDetectorIndices->begin() + range.first;
+  const auto end = m_assemblySortedDetectorIndices->begin() + range.second;
+  //for (const auto index : detectorsInSubtree(componentIndex)) {
+  for (; it != end; ++it) {
+    const auto &index = *it;
+    m_detectorInfo->setPosition(index,
+                                m_detectorInfo->position(index) + offset);
+  }
+  }
+  {
+  auto &positions = m_positions.access();
+  const auto range = (*m_componentRanges)[rangesIndex];
+  auto it = m_assemblySortedComponentIndices->begin() + range.first;
+  const auto end = m_assemblySortedComponentIndices->begin() + range.second;
+  //for (const auto index : componentsInSubtree(componentIndex)) {
+  for (; it != end; ++it) {
+    const auto &index = *it;
+    size_t offsetIndex = compOffsetIndex(index);
+    positions[offsetIndex] += offset;
+  }
+  }
+  return;
+
   const auto indices = this->componentsInSubtree(
       componentIndex); // Includes requested component index
   for (auto &index : indices) {
     if (isDetector(index)) {
-      checkDetectorInfo();
-      m_detectorInfo->setPosition(index,
-                                  m_detectorInfo->position(index) + offset);
+      //checkDetectorInfo();
     } else {
       size_t offsetIndex = compOffsetIndex(index);
       m_positions.access()[offsetIndex] += offset;
@@ -215,11 +240,51 @@ void ComponentInfo::setRotation(const size_t componentIndex,
 
   auto transform = Eigen::Matrix3d(rotDelta);
 
+  {
+  const auto range = (*m_detectorRanges)[rangesIndex];
+  auto it = m_assemblySortedDetectorIndices->begin() + range.first;
+  const auto end = m_assemblySortedDetectorIndices->begin() + range.second;
+  //for (const auto index : detectorsInSubtree(componentIndex)) {
+  for (; it != end; ++it) {
+    const auto &index = *it;
+      auto newPos =
+          transform * (m_detectorInfo->position(index) - compPos) + compPos;
+      auto newRot = rotDelta * m_detectorInfo->rotation(index);
+      m_detectorInfo->setPosition(index, newPos);
+      m_detectorInfo->setRotation(index, newRot);
+  }
+  }
+  {
+  auto &positions = m_positions.access();
+  auto &rotations = m_rotations.access();
+  const auto range = (*m_componentRanges)[rangesIndex];
+  auto it = m_assemblySortedComponentIndices->begin() + range.first;
+  const auto end = m_assemblySortedComponentIndices->begin() + range.second;
+  //for (const auto index : componentsInSubtree(componentIndex)) {
+  for (; it != end; ++it) {
+    const auto &index = *it;
+      auto newPos = transform * (position(index) - compPos) + compPos;
+      auto newRot = rotDelta * rotation(index);
+      const size_t childCompIndexOffset = compOffsetIndex(index);
+      positions[childCompIndexOffset] = newPos;
+      rotations[childCompIndexOffset] = newRot;
+  }
+  }
+  return;
+
+  //const auto rangesIndex = compOffsetIndex(componentIndex);
+  //const auto range = (*m_componentRanges)[rangesIndex];
+  //auto it = m_assemblySortedComponentIndices->begin() + range.first;
+  //const auto end = m_assemblySortedComponentIndices->begin() + range.second;
   const auto indices = this->componentsInSubtree(
       componentIndex); // Includes requested component index
   for (auto &index : indices) {
+  //for (; it != end; ++it) {
+    //const auto &index = *it;
     if (isDetector(index)) {
-      checkDetectorInfo();
+      //checkDetectorInfo();
       auto newPos =
           transform * (m_detectorInfo->position(index) - compPos) + compPos;
       auto newRot = rotDelta * m_detectorInfo->rotation(index);
diff --git a/Framework/Geometry/src/Instrument/InfoComponentVisitor.cpp b/Framework/Geometry/src/Instrument/InfoComponentVisitor.cpp
index 67a6166..7607c27 100644
--- a/Framework/Geometry/src/Instrument/InfoComponentVisitor.cpp
+++ b/Framework/Geometry/src/Instrument/InfoComponentVisitor.cpp
@@ -68,7 +68,7 @@ InfoComponentVisitor::InfoComponentVisitor(
       m_pmap(pmap), m_sourceId(sourceId), m_sampleId(sampleId) {
   const auto nDetectors = m_orderedDetectorIds->size();
   m_assemblySortedDetectorIndices->reserve(nDetectors);  // Exact
-  m_assemblySortedComponentIndices->reserve(nDetectors); // Approximation
+  //m_assemblySortedComponentIndices->reserve(nDetectors); // Approximation
   // m_componentIdToIndexMap->reserve(nDetectors);          // Approximation
 }
 
@@ -190,7 +190,7 @@ size_t InfoComponentVisitor::registerDetector(const IDetector &detector) {
     (*m_componentIdToIndexMap)[detector.getComponentID()] = detectorIndex;
     (*m_componentIds)[detectorIndex] = detector.getComponentID();
     m_assemblySortedDetectorIndices->push_back(detectorIndex);
-    m_assemblySortedComponentIndices->push_back(detectorIndex);
+    //m_assemblySortedComponentIndices->push_back(detectorIndex);
   }
   /* Note that positions and rotations for detectors are currently
   NOT stored! These go into DetectorInfo at present. push_back works for other
```

LamarMoore (commented):
Changes look good just a few queries. Will you be replacing all of the areas where the component name is used to retrieve the component? I see that you then have to use `ComponentInfo::indexOf` to get the index, and then feed this into `ComponentInfo::setRotation/Position` which I am presuming is slow.

Framework/DataHandling/src/LoadIsawDetCal.cpp
```diff
@@ -431,20 +431,23 @@ void LoadIsawDetCal::doRotation(V3D rX, V3D rY, DetectorInfo &detectorInfo,
...
+    Quat rot0 = componentInfo.relativeRotation(parentIndex);
     rot0.inverse();
     Rot *= rot0;
+
+    if (componentInfo.hasParent(parentIndex)) {
+      const auto grandParentIndex = componentInfo.parent(parentIndex);
+      Quat rot0 = componentInfo.relativeRotation(grandParentIndex);
```
LamarMoore: auto?

PR #19758, opened 2017-05-31
Stitch1DMany refactoring

### LINKED ISSUE #18378: Refactor `Stitch1DMany`
The code of `Stitch1DMany` is getting too complicated, partly because of the new property `ScaleFactorFromPeriod`. This issue is to have a look at the code and see if it can be simplified. Specifically, if we could replace property `ManualScaleFactor`, which is currently a number, with a vector, that would make things a lot easier. I need to check with scientists that this change is OK. In addition, we should validate the properties as soon as possible and throw better error messages.

PranavBahuguna:
Description of work.

The main aim of this refactoring was to enable the `ManualScaleFactors` property to take a list of multiple values instead of just one. During stitching, each of these scale factors can be applied separately to each input workspace. In addition a significant amount of other refactoring was done to simplify the code as it was getting very complicated.

Edit: Another new thing this refactor is forcibly ensure that `InputWorkspaces` can only take a list of MatrixWorkspaces or WorkspaceGroups containing MatrixWorkspaces only.

**To test:**

- Mostly code review
- Ensure unit tests pass and the newly added ones make sense
- Ensure the relevant documentation is updated and makes sense

Fixes #18378.

*Release notes will be updated later.*

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Algorithms/src/Stitch1DMany.cpp
```diff
@@ -192,79 +201,95 @@ void Stitch1DMany::validateCommonInputs(
...
+      alg->setProperty("ManualScaleFactor", manualScaleFactors[i - 1]);
+    alg->execute();
+
+    lhsWS = alg->getProperty("OutputWorkspace");
+    double outScaleFactor = alg->getProperty("OutScaleFactor");
+    outScaleFactors.push_back(outScaleFactor);
+
+    if (!isChild()) {
+      // Copy each input workspace's history into our output workspace's history
+      for (auto &inputWS : toStitch) {
```
DavidFair: `const auto`

Framework/Algorithms/test/Stitch1DManyTest.h
```diff
@@ -447,8 +459,66 @@ class Stitch1DManyTest : public CxxTest::TestSuite {
...
+    alg.setProperty("StartOverlaps", "0.8, 1.6");
+    alg.setProperty("EndOverlaps", "1.1, 1.8");
+    alg.setPropertyValue("OutputWorkspace", "outws");
+    alg.setProperty("UseManualScaleFactors", "1");
+    alg.setProperty("ManualScaleFactors", "0.5, 0.7");
+    alg.execute();
+
+    // Test output ws
+    Workspace_sptr outws = alg.getProperty("OutputWorkspace");
+    auto stitched = boost::dynamic_pointer_cast<MatrixWorkspace>(outws);
```
DavidFair: `TS_ASSERT(stitched)`

Framework/Algorithms/test/Stitch1DManyTest.h
```diff
@@ -742,6 +812,120 @@ class Stitch1DManyTest : public CxxTest::TestSuite {
...
+    alg.setProperty("StartOverlaps", "0.8, 1.6");
+    alg.setProperty("EndOverlaps", "1.1, 1.9");
+    alg.setProperty("UseManualScaleFactors", "1");
+    alg.setProperty("ManualScaleFactors", "0.5, 0.7");
+    alg.setPropertyValue("OutputWorkspace", "outws");
+    alg.execute();
+
+    // Test output ws
+    Workspace_sptr outws = alg.getProperty("OutputWorkspace");
+    auto group = boost::dynamic_pointer_cast<WorkspaceGroup>(outws);
```
DavidFair: `TS_ASSERT(group)`

Framework/Algorithms/test/Stitch1DManyTest.h
```diff
@@ -742,6 +812,120 @@ class Stitch1DManyTest : public CxxTest::TestSuite {
...
+    TS_ASSERT_EQUALS(group->getNumberOfEntries(), 2);
+
+    // First item in the output group
+    auto stitched =
```
DavidFair: ^

Framework/Algorithms/test/Stitch1DManyTest.h
```diff
@@ -993,8 +1175,11 @@ class Stitch1DManyTest : public CxxTest::TestSuite {
     }
 
     TS_ASSERT_EQUALS(histNames[0], "CreateWorkspace");
```
DavidFair: Could we have
```
const std::string createWsName = "CreateWorkspace";
const std::string groupWsName = "GroupWorkspaces";

TS_ASSERT_EQUALS(histNames[1], createWsName); ...etc.
```
This way if the names change we only have to change in in one place instead of 4

DavidFair (changes requested):

Framework/Algorithms/inc/MantidAlgorithms/Stitch1DMany.h
```diff
@@ -94,13 +93,13 @@ class DLLExport Stitch1DMany : public API::Algorithm {
   std::vector<double> m_endOverlaps;
   std::vector<double> m_params;
   std::vector<double> m_scaleFactors;
+  std::vector<double> m_manualScaleFactors;
```
PranavBahuguna: Some calls to the above functions will pass their own values in instead of member variables (e.g. doStitch1D() at line 398 passes a variable `periodScaleFactors` instead of the `m_manualScaleFactors`). While I suppose its possible to remove some parameters, it feels more consistent to leave them in.

DavidFair: Could we tweak the parameter name to `scaleFactors` so other developers don't fall into the same trap?

PranavBahuguna: Hmm, the purpose of the `doStitch1D()` and `doStitch1DMany()` was to provide functions for running the `Stitch1D` and `Stitch1DMany` algorithms respectively. The parameters for those functions were named after the same properties they supply. It's not exactly easy to explain, but I don't think its a good idea to change it as it makes less sense semantically to me.

DavidFair (approved):

DavidFair:
Tests described pass
The validation works correctly according to the doc description.

I'm happy to :shipit:

PR #19765, opened 2017-06-01
AsynchronousTest.testCancelGroupWS randomly fails

### LINKED ISSUE #19764: AsynchronousTest.testCancelGroupWS randomly fails.
### Expected behavior

Test always passes

### Actual behavior

Test randomly fails.

### Platforms affected

Probably all

mantid-roman:
Added synchronisation using condition variables.

Fixes #19764.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/test/AsynchronousTest.h
+++ b/Framework/API/test/AsynchronousTest.h
@@ -8,10 +8,11 @@
 #include "MantidKernel/make_unique.h"
 #include "MantidTestHelpers/FakeObjects.h"
 #include <Poco/ActiveResult.h>
-#include "Poco/Condition.h"
-#include "Poco/Mutex.h"
 #include <Poco/NObserver.h>
 #include <Poco/Thread.h>
+
+#include <condition_variable>
+#include <mutex>
 
 using namespace Mantid::Kernel;
 using namespace Mantid::API;
@@ -19,9 +20,9 @@ using namespace std;
 
 namespace {
 constexpr int NO_OF_LOOPS = 10;
-Poco::Condition condition1;
-Poco::Condition condition2;
-Poco::Mutex mtx;
+std::condition_variable condition1;
+std::condition_variable condition2;
+std::mutex mtx;
 }
 
 class AsyncAlgorithm : public Algorithm {
@@ -48,12 +49,11 @@ public:
 
   void exec() override {
     if (synchronise) {
-      mtx.lock();
-      condition1.wait(mtx);
-      mtx.unlock();
+      std::unique_lock<std::mutex> lock(mtx);
+      condition1.wait(lock);
     }
     Poco::Thread *thr = Poco::Thread::current();
-    condition2.broadcast();
+    condition2.notify_all();
     for (int i = 0; i < NO_OF_LOOPS; i++) {
       result = i;
       if (thr)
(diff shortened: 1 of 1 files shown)


martyngigg (commented):
Any reason we can't use `std::condition_variable` & `std::mutex` here?

mantid-roman:
Will it work with Poco threading?

martyngigg:
Yep. The Poco thread classes are just wrappers around the native threading libraries the same as the `std` variants - I don't think it should be a problem.

mantid-roman:
OK. I'll change.

martyngigg:
Thanks. I think it's good to use `std` stuff where we can. We should probably think about switching the main algorithm threading to `std::thread` too but that's a different question.

mantid-roman:
@martyngigg I was wrong, the times are the same.

AntonPiccardoSelg:
Does this not lead to a potential condition variable deadlock? 
If the test thread acquires the mutex first, it can go all the way down to the point where it waits on cond2. Then the AsyncAlgorithm thread can get the mutex and waits on cond1. But since the test thread is already past the point where it notifies the threads waiting on cond1, AsyncAlgorithm will keep waiting on cond1, which in turn will leave the test thread waiting on cond2. Am I missing something?

Framework/API/test/AsynchronousTest.h
```diff
@@ -143,12 +156,18 @@ class AsynchronousTest : public CxxTest::TestSuite {
   void testCancelGroupWS() {
     WorkspaceGroup_sptr groupWS = makeGroupWorkspace();
     AsyncAlgorithm alg;
+    alg.synchronise = true;
```
martyngigg: Is this correct? `synchronise` is set to true on the parent object but not on the child algorithms so I don't think the `if` statement above will ever fire?

martyngigg (changes requested):
I agree with Anton here. If the threads were slow to start up then the main thread will block indefinitely on the second condition.

Framework/API/test/AsynchronousTest.h
```diff
@@ -11,12 +11,21 @@
 #include <Poco/NObserver.h>
 #include <Poco/Thread.h>
 
+#include <condition_variable>
+#include <mutex>
+
 using namespace Mantid::Kernel;
 using namespace Mantid::API;
 using namespace std;
```
martyngigg: Can remove `using namespace std`? We shouldn't encourage its use in headers, even in tests.

martyngigg (commented):
This looks fine to me now from a functional perspective.

I'm not too fond of the global variables though, even though they are in an an anonymous namespace. Having them here could encourage others to use condition variables in this manner. Can we move them to private variables of the test along with the whole class definition of  `AsyncAlgorithm`?

Framework/API/test/AsynchronousTest.h
mantid-roman: But the variables will be static members?

martyngigg: Do they need to be static?

mantid-roman: They have to be passed to the algorithms

martyngigg: Oh yes, of course. :man_facepalming:

That's okay. At least they are taken out of the global scope and won't clash with other variables in other test headers (The final test binary includes all of the headers in a single file).

mantid-roman: We cannot use `DECLARE_ALGORITHM` with a nested class

martyngigg: Why do we need to use `DECLARE_ALGORITHM` at all if we are instantiating the object directly?

mantid-roman: `AlgorithmFactory` is used to clone algorithms to process groups.

martyngigg: Okay. Let's leave it as it is but can we capitalise the names to follow the standard for static variables and still remove the `using namespace std`.

martyngigg (approved):
I'm happy with these changes now.

martyngigg:
:shipit:

PR #19774, opened 2017-06-01
Fix OptimizeCrystalPlacement intermittent failure

martyngigg:
Description of work.

Fixes an index error when computing the Jacobian of `PeakHKLErrors`.

**To test:**

I could reliably get this to fail with

```
ctest -j 8 --repeat-until-fail 30 -R CrystalTest_OptimizeCrystalPlacement
```

before this changes. Afterwards it seems to keep running as expected.

No issue

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Crystal/src/PeakHKLErrors.cpp
+++ b/Framework/Crystal/src/PeakHKLErrors.cpp
@@ -479,7 +479,7 @@ void PeakHKLErrors::functionDeriv1D(Jacobian *out, const double *xValues,
     for (int kk = 0; kk < static_cast<int>(nParams()); kk++) {
       out->set(i, kk, 0.0);
       out->set(i + 1, kk, 0.0);
-      out->set(i + 1, kk, 0.0);
+      out->set(i + 2, kk, 0.0);
     }
 
     double chi, phi, omega;


NickDraper:
All looks good in the code.  Simple change, no risk of side effects outside of this unit test. :shipit:

SimonHeybrock:
This does not seem to be part of a unit test, so why can there be no side effects?

NickDraper:
Good point @SimonHeybrock, thanks for the catch, I'll give it a more thorough test

NickDraper:
Code is initialising the Jacobian to 0, this is done in blocks of 3 as defined by the outer for loop.
The previous repeated lines, clearly are not correct:
```
       out->set(i + 1, kk, 0.0);
       out->set(i + 1, kk, 0.0);
```
The new change appears to correctly zero the jacobian, it is not clear why this causes a problem as later on all branches appear to set the jacobian to other values, but it is possible that not all of the values are set as some of the index values are not easy to follow.
If an unitialised values entered the optimisation then that could certainly through off the optimizer leading to different results intermittently.

Thanks to Anton for being a second pair of eyes on this one

NickDraper:
After the fix I can't get the unit test to fail using the script above (2 runs of 30 executions), but I couldn't get the unit test to fail on my PC before the change either (2 runs of 30 executions).

I've asked @mantid-roman to review the changes as well for a second opinion.

mantid-roman:
Tested on Windows (GSL 1) and Ubuntu (GSL 2). The difference between GSL 1 and 2 that matters here is that GSL 2 doesn't initialise  matrices with 0s as GSL 1 does. 
This should explain why the tests on windows always pass and on ubuntu they fail before and pass after the fix.

mantid-roman (approved):

NickDraper:
:shipit:

PR #19777, opened 2017-06-01
Reduce flake8 warnings

### LINKED ISSUE #19770: Reduce flake8 warning count
Work on reducing flake8 warning count: http://builds.mantidproject.org/job/master_flake8/

DavidFair:
Description of work.
This PR reduces the number of flake8 warnings on master. 

I've intentionally left all warnings about a method being too complex or `__init__` import unused attributes as they need to be completed on a case-by-case basis.

There is also a small change where in the iPython `__init__` file we were checking the notebook version against a hardcoded string. Talking to the original developer this check does not make sense any more and can be removed. 

- Removed or fixed up unused variables
- Removed unused imports
- Fixed various variable name issues
- Brought missing import(s) into file which required them
- Converted code to Python 3 where applicable as part of the process
- Reformatted several files to fix multiple warnings in one hit

**To test:**
Code review
Ensure tests pass

Fixes #19770.

*Does not need to be in the release notes. - Internal changes only*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


peterfpeterson:
I think @mdoucet should comment since it removes a bunch of things from `scripts/reduction_workflow/instruments/sans/sns_command_interface.py`.

DavidFair:
+1 - The reason I took them out of `sns_command_interface` was the comment said they were for the system tests so I moved them into the respective tests.

peterfpeterson:
I *think* users`import sns_command_interface` in ipython and do everything from there. It is a convenience to them. @mdoucet should clarify.

mdoucet:
Pete is correct. This is what users import to set up their reduction. Your changes to sns_command_interface.py will definitely break everything.

peterfpeterson:
I looks like line 3 of `scripts/reduction_workflow/instruments/sans/sns_command_interface.py` should be changed to say something about the package is the main import for users working from the command line and shouldn't have "unused" things removed.

DavidFair:
Ok i'll add words to that effect and revert the changes to `sns_command_interface`

DavidFair:
I've reverted the change and added a comment warning future developers they should not remove the imports despite being unused.
I've also marked the imports as ignored so flake shouldn't pick them up any more - there is no point us getting warnings about something we cannot change (easily).

peterfpeterson:
Please change `hfir_command_interface.py` and `sns_command_interface.py` so the diff is much smaller (just `flake8` suppressions and additional comments) to make what you did more obvious.

DavidFair:
I've attempted to reduce the diff down for `hfir_command_interface` and `sns_command_interface` as much as possible, however because of the `#noqa` lines it will still be quite large.

peterfpeterson (approved):
Looks good now

PR #19811, opened 2017-06-06
Enable 2D r wavlength-based range reductions for ISIS SANS

AntonPiccardoSelg:
Rob reported that wavelength-based range reductions don't work for 2D reductions and that they revert to 1D reductions. Digging into the code I found that this was added by [design](https://github.com/mantidproject/mantid/blob/master/scripts/SANS/ISISCommandInterface.py#L734:L738 ). 

I asked the scientists if this feature should be available for 2D and Richard responded:
>I can’t think of any reason to restrict the list of wavelength bands to only 1d reductions, 2d should be fine,

This is a fairly low risk change, since this kind of reduction is used rather seldomly. Also it has been confirmed by scientists that this is (now) the correct behaviour. 

**To test:**

Please find the test data here: smb://olympic/babylon5/Scratch/Anton/BenchMarkingSANS2D_D20

1. Add the test data to the Mantid path
2. Open the ISIS SANS GUI
3. Select SANS2DTUBES as the instrument
4. Load the user file: USER_SANS2D_154E_2p4_4m_M3_Xpress_8mm_SampleChanger_REAR.txt
5. For the sample scatter add: SANS2D00034484
6. Press the Load button
7. Go to the Reduction Settings tab and on the wavelength row set the wavelength step type to "Ranges Lin"
8. Add "4,5,6" into the wavelength field. See below
image
9. Go to the Run numbers tab and press "2D Reduce"
10. After the reduction has finished confirm that the ADS has the workspaces:
  * 34484rear_2D_4.0_5.0
  *  34484rear_2D_4.0_6.0
  * 34484rear_2D_5.0_6.0
11. Confirm that these workspaces have more than one spectrum/histogram

**Release notes **
Can be found [here](https://github.com/mantidproject/mantid/pull/19810/commits/f037be6b47d470db91a54f7feeae727e12a7de78)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/docs/source/release/v3.10.0/sans.rst
+++ b/docs/source/release/v3.10.0/sans.rst
@@ -28,5 +28,6 @@ Bug Fixes
 - Fixed Batch mode bug where merged reductions set in the GUI were not respected.
 - Fixed display of current IDF, which was not updated when operating in batch mode.
 - Fixed a bug where the user could try to save 1D data in the CanSAS format and caused a crash.
+- Enabled wavelength-range based reductions in 2D.
 
 `Full list of changes on github <http://github.com/mantidproject/mantid/pulls?q=is%3Apr+milestone%3A%22Release+3.10%22+is%3Amerged+label%3A%22Component%3A+SANS%22>`__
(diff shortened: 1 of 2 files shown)


AnthonyLim23:
instead of step I have dW/W as the label.

AntonPiccardoSelg:
@AnthonyLim23 Go to the line edit box after `dW/W` and change it to `Ranges Lin`. It will change the label as well.

AnthonyLim23:
ran the test and it produced the workspaces. Each workspace had 150 spectra>1, so it passed the test. :shipit:

PR #19815, opened 2017-06-06
Clean up of MuonFitPropertyBrowser 

### LINKED ISSUE #19713: MuonFitPropertyBrowser clean up
There are several strings that are repeated throughout the code. These should be replaced by const variables (to prevent typos).

AnthonyLim23:
There were a set of strings that where repeated several times throughout the file. These have been replaced by const function calls to prevent typos. 

**To test:**

Load muon analysis with multiple period data (HIFI is good). Make sure multiple fitting is selected in the settings tab. Then in the data analysis tab check that changing the Groups/Periods selected causes the appropriate selection of data (custom creates a pop up box). 

Fixes #19713 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/MuonFitPropertyBrowser.h
+++ b/MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/MuonFitPropertyBrowser.h
@@ -72,12 +72,6 @@ public:
   std::string outputName() const override;
   /// Prefix for simultaneous fit results
   static const std::string SIMULTANEOUS_PREFIX;
-  /// custom label
-  static const std::string CUSTOM_LABEL;
-  static const std::string ALL_GROUPS_LABEL;
-  static const std::string ALL_PAIRS_LABEL;
-  static const std::string ALL_PERIODS_LABEL;
-
   /// Set label for simultaneous fit results
   void setSimultaneousLabel(const std::string &label) override {
     m_simultaneousLabel = label;
--- a/MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/MuonFitPropertyBrowser.h
+++ b/MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/MuonFitPropertyBrowser.h
@@ -72,6 +72,12 @@ public:
   std::string outputName() const override;
   /// Prefix for simultaneous fit results
   static const std::string SIMULTANEOUS_PREFIX;
+  /// custom label
+  static const std::string CUSTOM_LABEL;
+  static const std::string ALL_GROUPS_LABEL;
+  static const std::string ALL_PAIRS_LABEL;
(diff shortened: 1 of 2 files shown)


MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/MuonFitPropertyBrowser.h
```diff
@@ -72,6 +72,12 @@ class EXPORT_OPT_MANTIDQT_MANTIDWIDGETS MuonFitPropertyBrowser
   std::string outputName() const override;
   /// Prefix for simultaneous fit results
   static const std::string SIMULTANEOUS_PREFIX;
+  /// custom label
+  static const std::string CUSTOM_LABEL;
```
DavidFair: If this is the only file these are used in can we just move them into an anonymous namespace in the .cpp file instead of making them visible at a larger namespace

AnthonyLim23: fixed it

DavidFair (approved):
I'm happy with these changes and its nice to see the use of constants instead of copy + paste.

MantidQt/MantidWidgets/src/MuonFitPropertyBrowser.cpp
```diff
@@ -59,6 +59,10 @@
 
 namespace {
 Mantid::Kernel::Logger g_log("MuonFitPropertyBrowser");
+const std::string CUSTOM_LABEL{"Custom"};
```
martyngigg: Why not construct them directly as `QString`s to save the overhead in always converting them below?

martyngigg (changes requested):

DavidFair:
@martyngigg The comparisons in the `if` are against `std::string` - I'm unsure whether its worth storing both `std::string` and `QStrings`?

martyngigg (commented):
@DavidFair That's a fair point but I think before all of the comparisons happen the strings are actually converted to `std::string`, e.g [here]( /mantidproject/mantid/pull/19815/files#diff-60e7d509d34fcca38434a49cc5c76e94R344) aren't they?

DavidFair:
Ahh good catch - I fully agree we should switch to using `QString` throughout then

AnthonyLim23:
That would leave a comparison between a QString and a string, as the comparisons are between the "option" (a converted string) with the original label

martyngigg (approved):
Thanks. The changes look good now.

I'll merge it as soon the builds pass.

martyngigg:
@AnthonyLim23 It looks like those last changes need a clang-format

martyngigg:
Test failure is one of the random MantidPlotSliceViewer failures and is unrelated to this.

The system tests are also no necessary for these changes.

PR #19826, opened 2017-06-07
Add caching to multiple file property

peterfpeterson:
Much like `FileProperty`, add caching of found files to `MultipleFileProperty`.

**To test:**

Review the code and make sure that `APITest_MultipleFilePropertyTest` isn't slower than it was before merging this change. Also test `Load` from the GUI with archive search turned on. It should have less lag when pressing the `Run` button after filling in the filename when multiple files are accepted.

*There was no associated issue.*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/src/MultipleFileProperty.cpp
+++ b/Framework/API/src/MultipleFileProperty.cpp
@@ -180,6 +180,13 @@ std::string MultipleFileProperty::getDefault() const {
  */
 std::string
 MultipleFileProperty::setValueAsSingleFile(const std::string &propValue) {
+  // if value is unchanged use the cached version
+  if ((propValue == m_oldPropValue) && (!m_oldFoundValue.empty())) {
+    PropertyWithValue<std::vector<std::vector<std::string>>>::operator=(
+        m_oldFoundValue);
+    return "";
+  }
+
   // Use a slave FileProperty to do the job for us.
   FileProperty slaveFileProp("Slave", "", FileProperty::Load, m_exts,
                              Direction::Input);
@@ -190,15 +197,21 @@ MultipleFileProperty::setValueAsSingleFile(const std::string &propValue) {
     return error;
 
   // Store.
+  std::vector<std::vector<std::string>> foundFiles;
   try {
-    std::vector<std::vector<std::string>> result;
-    toValue(slaveFileProp(), result, "", "");
-    PropertyWithValue<std::vector<std::vector<std::string>>>::operator=(result);
+    toValue(slaveFileProp(), foundFiles, "", "");
+    PropertyWithValue<std::vector<std::vector<std::string>>>::operator=(
+        foundFiles);
   } catch (std::invalid_argument &except) {
     g_log.debug() << "Could not set property " << name() << ": "
                   << except.what();
     return except.what();
   }
+
+  // cache the new version of things
+  m_oldPropValue = propValue;
+  m_oldFoundValue = foundFiles;
+
   return "";
 }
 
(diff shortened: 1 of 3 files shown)


Framework/API/src/MultipleFileProperty.cpp
```diff
@@ -216,6 +229,12 @@ MultipleFileProperty::setValueAsSingleFile(const std::string &propValue) {
  */
 std::string
 MultipleFileProperty::setValueAsMultipleFiles(const std::string &propValue) {
+  // if value is unchanged use the cached version
+  if ((propValue == m_oldPropValue) && (!m_oldFoundValue.empty())) {
+    return PropertyWithValue<std::vector<std::vector<std::string>>>::setValue(
+        toString(m_oldFoundValue));
```
quantumsteve: Would it make more sense to use  `PropertyWithValue<TYPE>::operator=(const TYPE &)` and avoid additional conversions between strings and vectors?

Framework/API/src/MultipleFileProperty.cpp
```diff
@@ -395,6 +414,10 @@ MultipleFileProperty::setValueAsMultipleFiles(const std::string &propValue) {
     allFullFileNames.push_back(fullFileNames);
   }
 
+  // cache the new version of things
+  m_oldPropValue = propValue;
+  m_oldFoundValue = allFullFileNames;
```
quantumsteve: Adjust the order and you may be able to use `m_oldFoundValue = std::move(allFullFileNames)`

Framework/API/src/MultipleFileProperty.cpp
```diff
@@ -395,6 +414,10 @@ MultipleFileProperty::setValueAsMultipleFiles(const std::string &propValue) {
...
+
   // Now re-set the value using the full paths found.
   return PropertyWithValue<std::vector<std::vector<std::string>>>::setValue(
       toString(allFullFileNames));
```
quantumsteve: Would it make more sense to use `PropertyWithValue<TYPE>::operator=(const TYPE &)` and avoid additional conversions between strings and vectors?

quantumsteve (commented):
Found a few places where the code could (should?) be improved. Give that these change don't change the functionality and the tight release timeline, I'll defer whether or not to merge as is to the gatekeepers.

AnthonyLim23:
Using 10 runs of the test and taking the mean, I got:

master: 0.61 pm 0.08 seconds 
this branch:

AnthonyLim23:
Behaves the same as previous release, however loading via the GUI does feel slower in the new branch. Also the new branch does produce the error:

```
Error in execution of algorithm LoadMuonLog:
Property with given name already exists search object TEMP_MAXPOWER
Unable to successfully run LoadMuonLog Child Algorithm
Unable to successfully run LoadMuonLog Child Algorithm
```

AnthonyLim23:
:shipit:

martyngigg:
I have opened #19831 to deal with code improvements as I think there are others that are worth addressing too.

Framework/API/src/MultipleFileProperty.cpp
```diff
@@ -190,15 +197,21 @@ MultipleFileProperty::setValueAsSingleFile(const std::string &propValue) {
...
+        foundFiles);
   } catch (std::invalid_argument &except) {
     g_log.debug() << "Could not set property " << name() << ": "
                   << except.what();
     return except.what();
   }
+
+  // cache the new version of things
+  m_oldPropValue = propValue;
+  m_oldFoundValue = foundFiles;
```
peterfpeterson: This causes a segfault

Framework/API/src/MultipleFileProperty.cpp
peterfpeterson: segfault here too

PR #19838, opened 2017-06-09
Include-What-You-Use Kernel header tidy up

DavidFair:
Description of work.
This PR contains some tidy up work in the Kernel includes from running IWYU. 

There are so many changes that I only changed the files where the tool recommended removing headers. I then fixed up any files which were broken by the implicit import being removed.

**To test:**
Ensure all builds pass

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


peterfpeterson:
:shipit: once the builds go through

Framework/Geometry/src/Instrument/FitParameter.cpp
```diff
@@ -4,8 +4,10 @@
 #include "MantidGeometry/Instrument/FitParameter.h"
 #include "MantidGeometry/Instrument/Parameter.h"
 #include "MantidGeometry/Instrument/ParameterFactory.h"
-#include <MantidKernel/StringTokenizer.h>
 #include "MantidGeometry/muParser_Silent.h"
+#include "MantidKernel/StringTokenizer.h"
+
+#include <boost/lexical_cast.hpp>
```
quantumsteve: Do we want quotes or angle brackets for third-party libraries? My preference has been for the former (leaving angle brackets for just the standard library), but the opposite isn't wrong and ultimately what matters is consistency.

PR #19842, opened 2017-06-09
Add input instrument approximation to MonteCarloAbsorption

### LINKED ISSUE #19757: Performance of MonteCarloAbsorption with large number of detectors
The [`MonteCarloAbsorption`](http://docs.mantidproject.org/nightly/algorithms/MonteCarloAbsorption-v1.html) algorithm is quite slow to run with large instruments, like ILL's IN5. The simulation might be more efficient if it was performed only for a selected set of detectors and the results interpolated over the rest.

soininen:
This PR extends [`MonteCarloAbsorption`](http://docs.mantidproject.org/nightly/algorithms/MonteCarloAbsorption-v1.html) by adding an option to approximate the instrument of the input workspace by a 'sparse' instrument. The sparse instrument consists of a grid of detectors spawning the opening angles of the full instrument thus enabling simulation with smaller number of detectors than in the full input instrument. The simulated absorption corrections are then interpolated to the full instrument in a two step process.

This PR shouldn't introduce changes to existing functionality.

**To test:**

Run the algorithm by enabling and disabling *SparseInstrument* to see if the results more-or-less match. If possible, check that when *SparseInstrument* is disabled, the results are unchanged compared to master.

Fixes #19757.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Algorithms/src/InterpolationOption.cpp
```diff
@@ -81,8 +73,32 @@ std::string InterpolationOption::propertyDoc() const {
...
+  case Value::CSpline:
+    interpolateCSplineInplace(inOut, stepSize);
+    return;
+  default:
+    throw std::runtime_error("InterpolationOption::applyInplace() - "
+                             "Unimplemented interpolation method.");
+  }
+}
+
+void InterpolationOption::applyInPlace(const HistogramData::Histogram &in,
```
gvardany: missing doxygen

Framework/Algorithms/src/SampleCorrections/DetectorGridDefinition.cpp
```diff
@@ -0,0 +1,98 @@
...
+  if (std::abs(m_minLatitude - m_maxLatitude) < tiny) {
+    m_minLatitude -= smallShift;
+    m_maxLatitude += smallShift;
+  }
+  if (std::abs(m_minLongitude - m_maxLongitude) < tiny) {
+    m_minLongitude -= smallShift;
+    m_maxLongitude += smallShift;
+  }
+  m_latitudeStep = (m_maxLatitude - m_minLatitude) /
+                   static_cast<double>(m_latitudePoints - 1);
```
gvardany: what if `m_latitudePoints = 1`?

Framework/HistogramData/src/Interpolate.cpp
```diff
@@ -60,9 +60,52 @@ void sanityCheck(const Histogram &input, const size_t stepSize,
   }
 }
 
+void sanityCheck(const Histogram &input, const Histogram &output) {
```
gvardany: missing doxygen

Framework/HistogramData/src/Interpolate.cpp
```diff
@@ -60,9 +60,52 @@ void sanityCheck(const Histogram &input, const size_t stepSize,
...
+  }
+  if (!std::is_sorted(inPoints.cbegin(), inPoints.cend())) {
+    throw std::runtime_error(
+        "interpolate - input X data must be sorted in ascending order.");
+  }
+}
+
+enum class InterpolationType { LINEAR, CSPLINE };
+
+template <typename XData, typename YData, typename XInterp, typename YInterp>
```
gvardany: same here

Framework/HistogramData/src/Interpolate.cpp
```diff
@@ -178,6 +208,15 @@ void interpolateLinearInplace(Histogram &inOut, const size_t stepSize) {
   interpolateYLinearInplace(inOut, stepSize, inOut.mutableY());
 }
 
+void interpolateLinearInplace(const Histogram &input, Histogram &output) {
```
gvardany: and here

Framework/HistogramData/src/Interpolate.cpp
```diff
@@ -215,5 +254,14 @@ void interpolateCSplineInplace(Histogram &inOut, const size_t stepSize) {
   interpolateYCSplineInplace(inOut, stepSize, inOut.mutableY());
 }
 
+void interpolateCSplineInplace(const Histogram &input, Histogram &output) {
```
gvardany: here

Framework/Algorithms/src/MonteCarloAbsorption.cpp
```diff
@@ -108,6 +448,26 @@ void MonteCarloAbsorption::init() {
 
   InterpolationOption interpolateOpt;
   declareProperty(interpolateOpt.property(), interpolateOpt.propertyDoc());
+  declareProperty("SparseInstrument", false, "Enable simulation on special "
+                                             "instrument with a sparse grid of "
+                                             "detectors interpolating the "
+                                             "results to the real instrument.");
+  auto twoOrMore = boost::make_shared<Kernel::BoundedValidator<int>>();
+  twoOrMore->setLower(2);
```
gvardany: As we have tested once, I think it is better to set an odd default value (so 3) for the latitude points. Supposedly most of the instruments will be more or less symmetric around the equatorial, where with 2 points only one would simulate the upper and lower bands, and get a systematic shift after interpolating to the equator.

Framework/Algorithms/src/MonteCarloAbsorption.cpp
```diff
@@ -67,6 +81,332 @@ struct EFixedProvider {
...
+  algebraElement->setAttribute("val", "cube");
+  typeElement->appendChild(algebraElement);
+  ShapeFactory shapeFactory;
+  return shapeFactory.createShape(typeElement);
+}
+
+/** Create a workspace whose instrument approximates that of modelWS.
+ *  @param modelWS A workspace to model.
+ *  @param grid An object defining the detector grid of the output workspace.
+ *  @param wavelengthPoints Number of points in the output workspace.
```
gvardany: missing @throw

gvardany (changes requested):
The new option is very useful indeed for speeding up the simulation for large instruments. Changes look fine in general, except few minor oversights inline. Probably the major thing is to move the new helper methods in `MonteCarloAbsorption.cpp` file to a `SparseWorkspaceCreation.cpp` or something, since they are too many. Also like this they can be tested in isolation. I will proceed to functional testing shortly.

Framework/Algorithms/src/MonteCarloAbsorption.cpp
```diff
@@ -67,6 +81,332 @@ struct EFixedProvider {
...
+MatrixWorkspace_uptr
+createSparseWS(const MatrixWorkspace &modelWS,
+               const Mantid::Algorithms::DetectorGridDefinition &grid,
+               const size_t wavelengthPoints) {
+  // Build a quite standard and somewhat complete instrument.
+  auto instrument = boost::make_shared<Instrument>("MC_simulation_instrument");
+  instrument->setReferenceFrame(
+      boost::make_shared<ReferenceFrame>(Y, Z, Right, ""));
+  // Add sample to origin.
+  const V3D samplePos{0.0, 0.0, 0.0};
```
gvardany: One more comment, here you set sample at the centre, and then the source along the negative ray of z. 
I wonder if these should also be copied from the actual instrument?

gvardany:
Results look good as well. Obtained identical results with master in case of no approximation.

Framework/Algorithms/src/SampleCorrections/DetectorGridDefinition.cpp
soininen: In principle, the validators in `MonteCarloAbsorption::init()` prevent this, but I guess it doesn't hurt to be extra careful here.

Framework/Algorithms/src/MonteCarloAbsorption.cpp
soininen: Good point. The bound can be changed later to 2 if needed without breaking the API. I'll change the default to an odd number as well.

Framework/Algorithms/src/MonteCarloAbsorption.cpp
soininen: The original reference frame can be (and should be) taken into account, but I'd rather move the sparse instrument such that the sample position is in the origin as the beam profile stuff kind of expects it.

gvardany (approved):

gvardany:
Changes look good, the separated tests as well.

martyngigg:
Agreed. The images for the docs are very nice too.

PR #19843, opened 2017-06-12
New algorithm ConjoinXRuns

### LINKED ISSUE #19721: New algorithm ConjoinXRuns
A new algorithm is needed to join the workspaces horizontally, i.e. by appending the columns of the input workspaces.

gvardany:
This adds a new algorithm `ConjoinXRuns` that concatenates workspaces horizontally by appending their bins.
Sample log entry can be specified to serve as the x-axis of the output.
Sample log merging is handled similarly as in MergeRuns.
Some common parts with MergeRuns are refactored out to a common helper.

**To test:**
Read the documentation and play around with the examples therein.

Fixes #19721 .

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


SimonHeybrock:
I do not have time to review this, but just a short comment here:

Having both `JoinRuns` and `MergeRuns` is confusing, since neither of the names really says what they actually do. What is the difference that prevents having this as a single algorithm? Is it just that the former merges data into the same spectrum?

gvardany:
`MergeRuns` actually sums the counts, while `JoinRuns` concatenates the bins in the spectra. I agree though either of the names reflect this clearly. I was thinking about having one algorithm with an option Merge/Join, or (sum/concatenate), but at the end it feels easier to maintain them separate. Merge runs is already quite heavy and does specific merging for the event workspaces, and some rebinning logic which are not relevant for joining.

Framework/Algorithms/test/JoinRunsTest.h
```diff
@@ -0,0 +1,250 @@
...
+    logAdder.setProperty("Workspace", "ws1");
+    logAdder.setProperty("LogText", "0.7");
+    logAdder.execute();
+
+    m_testee.setProperty("InputWorkspaces", m_testWS);
+
+    m_testee.setProperty("SampleLogAsXAxis", "TestNumLog");
+
+    // blocksize must be one in case of scalar log, fail
+    TS_ASSERT_THROWS(m_testee.execute(), std::runtime_error);
```
ianbush: It is nice to assert on the error message too here, to ensure that a good error message is generated and things are failing for the expected reason.

```cpp
TS_ASSERT_THROWS_EQUALS(
    builder.setHistogram(Histogram(x, y)), const std::logic_error &e,
    std::string(e.what()),
    "Histogram supplied does not have the correct size.")
```

Framework/Algorithms/test/JoinRunsTest.h
```diff
@@ -0,0 +1,250 @@
...
+    logAdder.execute();
+
+    m_testee.setProperty("InputWorkspaces", m_testWS);
+    m_testee.setProperty("SampleLogAsXAxis", "TestStrLog");
+
+    // string log not supported, fail
+    TS_ASSERT_THROWS(m_testee.execute(), std::runtime_error);
+  }
+
+  void testWithNumSeriesLog() {
```
ianbush: I think this test should be split into 2, :smile: case and :cry: case.

Framework/Algorithms/src/JoinRuns.cpp
```diff
@@ -0,0 +1,401 @@
...
+      m_outWS->getAxis(0)->unit() = UnitFactory::Instance().create(unit);
+    } catch (Exception::NotFoundError &) {
+      m_outWS->getAxis(0)->unit() = UnitFactory::Instance().create("Empty");
+    }
+  }
+
+  setProperty("OutputWorkspace", m_outWS);
+}
+} // namespace Algorithms
+} // namespace Mantid
```
ianbush: I think there needs to be some tests that at least touch the sample log behaviour to check they are wired up correctly. It does not need as many as in `MergeRuns`, but at least something that touches the input options.

docs/source/algorithms/JoinRuns-v1.rst
```diff
@@ -0,0 +1,124 @@
...
+
+.. testoutput:: JoinRunsTSLogExample
+
+    out has 6 bins with x-axis as: 0.0, 0.25, 10.0, 10.25, 20.0, 20.25
+
+
+.. categories::
+
+.. sourcelink::
+
```
ianbush: I think it is worth adding one example which shows the sample log behaviour. Then as the docs link to `MergeRuns` users can find more information in the docs there if they need it.

ianbush (changes requested):
Overall this looks good, things work as expected to join the runs.

There are some comments I have added, but none of them should require any major changes.

Other comments:
 * If I run try and use a time series log with too few entries I get the error: `Some invalid Properties found`. Is it possible to add validation for this in `validateInputs()`?
 * I would also make it clearer in the docs that this is for workspaces containing point data only.

Framework/Algorithms/CMakeLists.txt
```diff
@@ -160,6 +160,7 @@ set ( SRC_FILES
 	src/InterpolatingRebin.cpp
 	src/InterpolationOption.cpp
 	src/InvertMask.cpp
+        src/JoinRuns.cpp
```
ianbush: Now line 162 has spaces too. In `CMakeLists.txt` everything should have tabs.

ianbush (approved):

martyngigg:
Sorry for being a bit late to the party. I'm also not so keen on the `JoinRuns` name, nor the `MergeRuns` name to be honest, I don't know why we had the `Runs` part in it, that bit seems irrelevant to me.

I agree with @gvardany that maintaining the code separately makes sense as the operations are doing quite different things. We can always keep this separation at the code level but have a single algorithm? Is it worth having a new `Concatenate` (similar to https://docs.scipy.org/doc/numpy/reference/generated/numpy.concatenate.html) that has a parameter to control the axis/type of merge and then forwards to the appropriate internal implementation? `MergeRuns` the algorithm could be then deprecated in favour of `Concatenate` ?

gvardany:
I can see your point. I don't have a strong opinion here. However one could argue that summing the counts as `MergeRuns` is conceptually very different from making distributions up by appending their bins as done in `JoinRuns`. The latter changes the dimensions as opposed to the former, so combining them into single algorithm might as well lead to some confusion.  Regarding the term `Run` I think it is not so irrelevant anymore, since the sample log merging is actually all about the run object in API.   In the contrary, `ConjoinWorkspaces` also concatenates but in the vertical (spectrum) axis, being ignorant about sample logs (at least for the moment), hence it does not have `Run` in it's name. Also from the users perspective, both joining and merging are expected to be commonly used after loading raw files that are interchangeably called as runs. I agree that there is a room for harmonising all these operations in terms of naming scheme and sample log awareness, but I would not do this in this PR yet. In addition, there are plans to support merging of scanned workspaces, which is yet another type of merging, so we have to see how that fits in. And maybe then we can have a clearer overview on how to make more consistent workspace combination tools. Nonetheless, `Join>Concatenate` I am open to rename if people prefer it more.

martyngigg:
@gvardany I see your point about the sample logs and differences between `Merge` and `Conjoin`. I also  hadn't appreciated the future problem of merging scanning workspaces. I think I would feel happier with something like `ConjoinXRuns` to be clearer that there is an expansion in the X dimensionality but it also deals with information about the sample etc. What do you think?

Any other thoughts @mantidproject/gatekeepers ?

gvardany:
Yes, I like the idea of putting the axis label in the name, this moderates the confusion. Whether to put `X` or `XAxis`, in the middle or at the end, is a matter of taste.

OwenArnold:
@gvardany. I just took a quick look. 1. Could you update the PR description/title to match the algorithm name. 2. Could you update the docs for `MergeRuns` and cross reference with the new algorithm documentation. It should be made very clear to our users when they would use `MergeRuns` and when they would use the new algorithm.

martyngigg:
@OwenArnold Do @gvardany's latest updates cover what you were thinking?

If so then I'll merge it.

OwenArnold:
@martyngigg @gvardany. I think these changes could go further. Here are my suggestions.

1. Maybe include a schematic like this http://docs.mantidproject.org/nightly/algorithms/Divide-v1.html#compatible-sizes in the `ConjoinXRuns` to show exactly what we mean by the operations involve.
2. Maybe do the same as (1) for `MergeRuns`
3. The `MergeRuns` documentation is too light. Could you also change that documentation to point to this algorithm, and better explain what that algorithm is doing.

I know that this sounds like a pain, but we want to have really good documentation so that our users are happy with the total Mantid experience.

@gvardany. I also notice that you haven't added any performance tests. Does that mean that performance is too unimportant to be tracked for this algorithm? Or was that an oversight?

gvardany:
@OwenArnold I will improve the documentation with illustrations and add a performance test. For `MergeRuns` though I would prefer someone else to improve the documentation, since for me it is a bit obscure what it does with events, and which kind of rebinning it does if the bins in the input do not match.

ianbush:
@OwenArnold I have some changes in the pipeline for `MergeRuns` to handle detector scans. I will do some updates to the `MergeRuns` docs then.

OwenArnold:
Thanks all. @ianbush: Good idea, could you ensure that the documentation changes are ready for the next release 3.11?

ianbush:
Yes, the changes I have made should be ready for then. Though as @gvardany says I am not sure how well I can document how it handles events, as that is not something we have had need for yet.

martyngigg:
Good to go now?

PR #19846, opened 2017-06-12
Indirect ILL : fix issue with crop dead monitor channels

### LINKED ISSUE #19844: Indirect ILL Reduction, issue with zero monitor channels
For QENS data with mirror sense, a problem might occur during the unmirroring when the number of 0 monitor channels is different in left and right wings. This should be handled better.

gvardany:
This PR fixes 2 minor issues in indirect ILL reduction encountered in a pathological cases as explained below:

**To test:**
Set ILL/IN16B as facility/instrument.

1) IN16B can record the data in mirror mode, when the raw data is organised in two adjacent equal wings, as shown, where each wing (left or right half of the matrix) has e.g. 6 bins.

screen shot 2017-06-12 at 13 56 57

In the reduction there is an option to crop out the bins, where the monitor (the first row) has 0 counts.
In the common case such bins are the same for the left and the right wings and there is no trouble.

However in a pathological case, as shown in the figure, 0 monitor bins are not exactly the same (0,1,6 for the left, 0,5,6 for the right) , so after cropping one is left with different axes, which would be later impossible to sum. The solution is to shift the x-axis after cropping to start from zero, which is done in https://github.com/mantidproject/mantid/pull/19846/files#diff-6cdb203e68b9cc4f584faeece4167d87R338

In a yet other imaginary case, it can be that the number of bins with 0 count is different in left and right. In this case nothing can be done rather than raising an error with a comprehensive message (https://github.com/mantidproject/mantid/pull/19846/files#diff-ee33a12732aafd5eceda9c75ff15a864R445) to prevent Plus from failing later.

2) Another issue has been fixed in `IndirectILLReductionFWS` regarding the calibration scale when interpolation option is requested.

**To test**

**1)** 
```qens = IndirectILLReductionQENS(Run='140721-140722', CropDeadMonitorChannels=True) ``` 
Should run without errors. A new test is added for this problematic case.

**2)** 
```
fws_sum = IndirectILLReductionFWS(Run='141195:141200', CalibrationRun='140645:140650', CalibrationOption='Sum')
fws_int = IndirectILLReductionFWS(Run='141195:141200', CalibrationRun='140645:140650', CalibrationOption='Interpolate')
```
Plot the spectrum 0 of the 2 outputs, they should be parallel curves roughly within 0.01 distance.
The files for this check are attached. Fine for probing the fix, they are not perfect as a reference to build a system test upon, since in this example it is actually extrapolation that is performed. Hence another test will be added in a separate PR with better reference data.

[test-files.zip](https://github.com/mantidproject/mantid/files/1102516/test-files.zip)

Fixes #19844 .

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### CODE REVIEW

File: Testing/Data/UnitTest/ILL/IN16B/140722.nxs.md5 [hunk 1]
```diff
@@ -0,0 +1 @@
+ad26a21b1fbdfeb85d6ac92517976e8e
```
ianbush: Why was this file added in unit tests?

ianbush: Why was this file added in unit tests?

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/IndirectILLReductionQENS.py
```diff
@@ -435,6 +439,14 @@ def _perform_unmirror(self, ws, run):
             mask_min = 0
             mask_max = mtd[left].blocksize()
 
+            if (self._common_args['CropDeadMonitorChannels'] and
```
ianbush: :+1:

docs/source/release/v3.11.0/indirect_inelastic.rst
```diff
@@ -24,4 +24,7 @@ Improvements
 Bugfixes
 --------
 
+- An issue has been fixed in :ref:`algm-IndirectILLEnergyTransfer` when handling the data with mirror sense, that have shifted 0 monitor counts in the left and right wings.
+- An issue has been fixed in :ref:`algm-IndirectILLReductionFWS` when requesting the interpolation option for the detector calibration.
```
ianbush: With the release notes you mention when the issue occurs, but can you briefly clarify the what the issue is? E.g. in the first case an issue with different x-axes, the second an issue with the scale.

ianbush (changes requested):
This looks good to me, the changes are an improvement to the workflows. Just 2 minor comments.

ianbush (approved):

PR #19847, opened 2017-06-12
Extend SetUncertainties with custom mode

### LINKED ISSUE #15326: Extend SetUncertainties to allow user to set uncertainty to specified value
Extend SetUncertainties to allow user to set uncertainty to specified value.

http://docs.mantidproject.org/nightly/algorithms/SetUncertainties-v1.html

ThomasLohnert:
Fixes #15326.

Extended SetUncertainties algorithm with "custom" mode, allowing users to replace an arbitrary error value with another arbitrary value. Users can also set the precision of the comparison used for finding values to replace.

**To test:**
1. Create a new sample workspace with random values. 
1. Run SetUncertainties with "custom" mode. Set IfEqualTo to some value between 0 - 1. Pick any arbitrary value for SetErrorTo. 
1. Open the output work space and confirm (only) errors within tolerance to the chosen value (as defined by precision) have been replaced.
1.  Try different values for precision to verify it works as described in the ticket (the precision value is the amount of decimal places to check).
1. All other modes should be unchanged.

[Release notes here](https://github.com/mantidproject/mantid/commit/d75edf29496a8661732fb9b89c43d555e09831fb)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


peterfpeterson (approved):

SimonHeybrock (dismissed):
May I ask what this is supposed to be used for? I feel that giving users easy means to fake data is not a good idea, since it is basically facilitating scientific fraud.

peterfpeterson:
Since many refinement packages use the uncertainties as weights, this allows for exploring the fitting landscape.



Sent with Good (www.good.com)
________________________________

SimonHeybrock:
@peterfpeterson Isn't that precisely my point? Why should we support editing weights if that changes fitting results?

@mantidproject/gatekeepers Could we get another opinion on this?

martyngigg:
While of course in our standard algorithms we want to make it hard to do the wrong thing with regards to histograms, points, errors etc I think we will always need to support editing the raw data itself from a user side as we have been repeatedly ask to expose this type of thing over the years. 

Ultimately I see that user have to be responsible as well for what they do with the data. After all they could just save stuff out to other programs and edit data afterwards anyway...

docs/source/release/v3.11.0/framework.rst
```diff
@@ -11,10 +11,9 @@ Algorithms
 New
 ###
 
-
 Improved
 ########
-
+- :ref:`SetUncertainties <algm-SetUncertainties-v1>` now provides a "custom" mode, which lets the user arbitrarily replace any error value x with y.
```
OwenArnold: I don't think this comment reads very well. I'm not sure our users are going to make sense of this when they read the release notes. 

Also, as far as I can tell, the algorithm documentation has not been updated, but should contain some important specification updates such as how truncation (as opposed to rounding) is used in the comparison. I would suggest updating the usage documentation to include new examples in addition to text updates.

OwenArnold:
One advantage of this being an algorithm is that there is some coded traceability, via the workspace history, that the data has been modified. However, I'm not sure I understand why users need to modify uncertainties on data they have already loaded, instead or reloading/importing the corrected data.

AndreiSavici (changes requested):
Algorithm documentation needs to be updated

AndreiSavici (approved):

PR #19851, opened 2017-06-14
Remove unused MATLAB bindings

quantumsteve:
Description of work.

The MATLAB bindings are unused, untested and most likely don't work. It was decided that these should be removed.

**To test:**

Check that the builds still pass.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


AndreiSavici (approved):
The only matlab mentions left are in documentation/comments. Ready to merge if builds pass

rosswhitfield (approved):

martyngigg:
@abuts I presume you don't use these either?

abuts:
They were broken so I can not use it. 

From other side, I will be happy to use some parts of Mantid in the incoming project, but not sure how many efforts it would be to use them vs. running export ASCII couple of times.

martyngigg:
I think it would require significant effort to get this working properly for production use. We can always resurrect this from history if we need to.

abuts:
fair enough. As only myself seems is interested in this, and nobody else have asked for this in 3 years time, I think we can kill it.

PR #19852, opened 2017-06-14
Test cotire for automatic support of pre-compiled headers for faster builds

SimonHeybrock:
**Do not merge, this is currently just for review/testing**

Pre-compiled headers can give substantial build speedups. [cotire](https://github.com/sakra/cotire) automatically generates this for cmake.

The example in this PR adds cotire (which is just a single file cmake module) and enables it for some of our modules. It gave me a speedup from `4min46` to `4min16` for a clean build of `Algorithms` (including all modules it depends on).

**To test:**

If we were to enable this globally, it would need to be tested for all our systems. I have done so only for gcc so far.

**If you are testing this on Windows**, disable the hand-written pre-compiled header by setting `ENABLE_PRECOMPILED_HEADERS=OFF`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,8 @@ if ( PACKAGE_WORKBENCH AND NOT ENABLE_WORKBENCH )
   message ( FATAL_ERROR "Packaging workbench requested but workbench build is disabled!" )
 endif ()
 
+include(cotire)
+
 set ( CPACK_INSTALL_CMAKE_PROJECTS "${CMAKE_BINARY_DIR}" "Mantid" "ALL" "/" )
 
 ###########################################################################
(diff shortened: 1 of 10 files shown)


gvardany:
I get an error running cmake on OS X:
```
CMake Error in Framework/API/CMakeLists.txt:
  Evaluation file to be written multiple times for different configurations
  or languages with different content:

    /Users/vardanyan/build-mantid-debug/Framework/API/API_CXX_Release_cotire.cmake
```
And similarly for all the targets.

SimonHeybrock:
@gvardany Indeed, I also get this with gcc now, it seems some change to our `cmake` files in the past half year broke something. Can you run the build anyway? It seems to work for me (but I am not sure it is doing the right thing).

gvardany:
I got this on make:
`error: C++ operator name keywords was enabled in PCH file but is currently disabled`

NickDraper:
@SimonHeybrock This PR is stale and the builds have been failing since Oct 2018.  I'm closing it for now, you can reopen it when the branch is ready.

PR #19854, opened 2017-06-14
Rename Concepts names to be more user friendly

dtasev:
Description of work.

Filenames in the `docs/source/concepts` folder have been changed to look friendlier to users.

How it looks before the changes: http://docs.mantidproject.org/nightly/concepts/index.html
How it looks after the changes:
image

**To test:**
Look over the names. Add suggestions for improvements

Fixes #xxxx.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


DTasev:
`MultipleFileProperty` seems to be put into the `.. categories:: Concepts`, however it is not in the `docs/source/concepts` folder. Should it be moved to concepts folder, or removed from the category?

martyngigg:
Let's leave it where it is for now while some of the other changes are being made.

martyngigg (approved):

SimonHeybrock:
I would really dislike having spaces in actual class names. This makes this part of the documentation much harder to navigate for actual developers, and arguably many of the pages documenting classes are probably used mostly by (newer) developers...?

DTasev:
I am not sure how to proceed, I both agree that class names should be the same, but also I don't think it will hinder anyone from finding the information. 

Especially when the same name is written with a space in the title. For example `EventWorkspace` is written as `Event Workspace` in the title of its page. And searching for `EventWorkspace` doesn't bring the actual concept page as a result, before or after the change.

DTasev:
Adding in progress as further feedback is required

martyngigg:
This has gotten old and now we're moving some things to the `dev-docs` it's likely this won't get resurrected.

Issue #19866, opened 2017-06-16
Dev 2017 Speed up slow unit tests


NickDraper:
Tackle a few slow unit tests

Slack Channel
dev2017_faster_builds

Slow tests spreadsheet
https://docs.google.com/spreadsheets/d/1uzAhnFM9Q8jMmrOzaZ32oBx_SDIdojKUgps9-EXpslo/edit#gid=0
Use the Unit tests tab

PR #19874, opened 2017-06-16
Speed up VelocityCrossCorrelations

AndreiSavici:
Moved  for loop to numpy array math

Re #19866. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


NickDraper (approved):
Code review looks good, while the previous code was easier to read, you cannot dispute that this way is much faster.  The only other alternative would have been to rewrite in C++.  This change is a nice compromise.

The unit test is minimal but passes and should be sufficient to check for regression in this case.

PR #19877, opened 2017-06-16
Replace topaz peaks file

NickDraper:
Replace slow Isaw Peaks file with a nexus translation of the same peaks workspace.  Many of the algorithms that use this have been altered to use this instead.  Loading the same peaks workspace from a processed nexus file is 2x faster.

**To test:**
1. Unit tests must pass on all environments

re #19866. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Crystal/test/FindUBUsingLatticeParametersTest.h
```diff
@@ -171,14 +172,14 @@ class FindUBUsingLatticeParametersTest : public CxxTest::TestSuite {
...
-    TS_ASSERT_DELTA(latt.c(), 8.6566, 5e-4);
-    TS_ASSERT_DELTA(latt.alpha(), 92.6267, 5e-4);
-    TS_ASSERT_DELTA(latt.beta(), 103.7440, 5e-4);
-    TS_ASSERT_DELTA(latt.gamma(), 90.0272, 5e-4);
+    TS_ASSERT_DELTA(latt.a(), 13.9520, 1e-2);
+    TS_ASSERT_DELTA(latt.b(), 19.5145, 1e-2);
+    TS_ASSERT_DELTA(latt.c(), 8.6566, 1e-2);
+    TS_ASSERT_DELTA(latt.alpha(), 92.6267, 1e-2);
+    TS_ASSERT_DELTA(latt.beta(), 103.7440, 1e-2);
+    TS_ASSERT_DELTA(latt.gamma(), 90.0272, 1e-2);
```
SimonHeybrock: What is the reason for the adapted tolerances?

SimonHeybrock (changes requested):
Looks good, just one question:

Framework/Crystal/test/FindUBUsingLatticeParametersTest.h
NickDraper: The altered file lead to slightly different values in the test, It is not completely clear why, but the algorithm is using optimisation so it is likely that a slight difference in precision of one of the starting parameters lead to a slightly different minima.  The code of the algorithm itself was not changed at all, so I'm confident we have not introduced any errors.

SimonHeybrock: Ok, that makes sense. I hope it does not point to a deeper underlying issue.

SimonHeybrock (approved):

PR #19887, opened 2017-06-19
Improved error checking in WorkspaceOpOverloads

### LINKED ISSUE #14442: Improve error checking in WorkspaceCreationHelper helpers and WorkspaceOpOverloads
See #14427, #14429. If one uses `Create2DWorkspace123()` or similar methods and 
- forgets to set `isHist=true`
- uses `Mantid::API::WorkspaceHelpers::makeDistribution()` on the workspace created

then `makeDistribution()` will overrun buffers because in `WorkspaceCreationHelper`, `Create2DWorkspaceWithValues()` initializes the workspace with `nBins` instead of `nBins + 1`. It seems to me that `makeDistribution()` should check and raise an exception, as this can lead to subtle bugs. Sibling workspace create methods should probably be checked if they need similar sanity checks as well.

ThomasLohnert:
Fixes #14442.

Added a check in makeDistribution to throw an error if the input workspace uses point data when it expects bin edges. 

**To test:**
1. Create a 2D workspace with the isHist parameter set to false
1. Run the workspace through makeDistribution
1. Confirm exception is thrown
1. Create another 2D workspace with the isHist parameter set to true
1. Run the second workspace through makeDistribution
1. Confirm no exception is thrown

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/src/WorkspaceOpOverloads.cpp
+++ b/Framework/API/src/WorkspaceOpOverloads.cpp
@@ -539,8 +539,9 @@ void WorkspaceHelpers::makeDistribution(MatrixWorkspace_sptr workspace,
 
   const size_t numberOfSpectra = workspace->getNumberHistograms();
   // check X vs Y data size
-  if (workspace->x(0).size() == workspace->y(0).size()){
-	  throw std::runtime_error("Workspace is using point data for x (should be bin edges).");
+  if (workspace->x(0).size() == workspace->y(0).size()) {
+    throw std::runtime_error(
+        "Workspace is using point data for x (should be bin edges).");
   }
   for (size_t i = 0; i < numberOfSpectra; ++i) {
     if (forwards) {
--- a/Framework/API/src/WorkspaceOpOverloads.cpp
+++ b/Framework/API/src/WorkspaceOpOverloads.cpp
@@ -538,8 +538,7 @@ void WorkspaceHelpers::makeDistribution(MatrixWorkspace_sptr workspace,
                              "into distributions.");
 
   const size_t numberOfSpectra = workspace->getNumberHistograms();
-  // check X vs Y data size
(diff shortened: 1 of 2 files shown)


Framework/API/src/WorkspaceOpOverloads.cpp
```diff
@@ -538,6 +538,11 @@ void WorkspaceHelpers::makeDistribution(MatrixWorkspace_sptr workspace,
                              "into distributions.");
 
   const size_t numberOfSpectra = workspace->getNumberHistograms();
+  // check X vs Y data size
+  if (workspace->x(0).size() == workspace->y(0).size()) {
```
SimonHeybrock: You can use `workspace->histogram(0).xMode() == Histogram::XMode::Points` instead of the size check. That is directly stating the intent of the check and you can avoid the code comment above.

Framework/API/test/WorkspaceOpOverloadsTest.h
```diff
@@ -204,6 +204,14 @@ class WorkspaceOpOverloadsTest : public CxxTest::TestSuite {
     TS_ASSERT_EQUALS(ws->readY(1)[0], 1.0)
     TS_ASSERT_EQUALS(ws->readE(0)[0], 1.0)
     TS_ASSERT_EQUALS(ws->readE(1)[0], 1.0)
+
+    // MakeDistribution throws exception if input WS uses point data
```
SimonHeybrock: Having smaller separate test should usually be preferred, e.g., `test_makeDistribution_fails_for_point_data()` (again, avoiding a comment by giving meaningful names).

SimonHeybrock (changes requested):
Looks good, just two small comments for improvements:

SimonHeybrock (approved):

PR #19875, opened 2017-06-16
Algorithm headers

peterfpeterson:
Remove some inner classes from `Algorithm.h` to help speed up builds and shrink binary size.

**To test:**

If the builds pass, this is fine.

*There is no associated issue*

*Does not need to be in the release notes* because it will not be seen by anybody.

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)?
- [ ] Are the unit tests small and test the class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] Do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant (user and developer) documentation been added/updated?

Does everything look good? Mark the review as **Approve**. A member of `@mantidproject/gatekeepers` will take care of it.

PR #19902, opened 2017-06-20
Algorithm headers

peterfpeterson:
Remove some inner classes from `Algorithm.h` to help speed up builds and shrink binary size. The original attempt is in #19875 which creates strange issues with both windows and doxygen.

**To test:**

If the builds pass, this is fine.

*Does not need to be in the release notes* because it will not be seen by anybody.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/API/src/AlgorithmNotification.cpp
```diff
@@ -0,0 +1,86 @@
+#include "MantidAPI/Algorithm.h"
```
quantumsteve: In general we ask the [All .cpp files must have a corresponding .h file](http://www.mantidproject.org/C%2B%2B_Coding_Standards#CPP_file_Headers). Putting this in a separate file works, but would it be clearer to move this into `Algorithm.cpp`?

quantumsteve (changes requested):
Changes look good, don't change functionality and may improve build times. Before merging, however, we should resolve the proper location for the implementation.

quantumsteve (approved):

PR #19915, opened 2017-06-22
Small include cleanup based on session during developer workshop

SimonHeybrock:
What the title says. The main improvement is that we now avoid including `NexusFile` *everywhere* (previously `V3D` and `MatrixWorkspace` required it, solved by moving code from `MatrixWorkspace` into `SaveNexusProcessed`).

**To test:**

Should be fine if it compiles. Just code review.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/API/src/FileBackedExperimentInfo.cpp
```diff
@@ -3,8 +3,8 @@
 
 #include <sstream>
 
-#include <nexus/NeXusException.hpp>
 #include <nexus/NeXusFile.hpp>
+#include <nexus/NeXusException.hpp>
```
soininen: Was there a reason to break the alphabetic ordering here? If this is a mistake, please have a look at `LoadMcStasNexus.cpp`, ``.

soininen (changes requested):
Looks good to me, except in my alphabetic ordering, 'Exception' comes before 'File'.

Framework/API/src/FileBackedExperimentInfo.cpp
SimonHeybrock: Yes it does not compile if it is the otherway round (I don't know why). I believe in the cases you mentioned the `NeXusFile.hpp` must get included by some other header before.

soininen: :man_facepalming: for non-self-contained headers.

soininen (approved):

PR #19917, opened 2017-06-22
Fix 4 conversion warnings from GCC 6.3

### LINKED ISSUE #17593: Reducing GCC Compiler Warnings
Work on reducing GCC compiler warnings from [Maintenance](https://github.com/mantidproject/documents/blob/master/Project-Management/TechnicalSteeringCommittee/reports/MaintenanceTasks.md) document:
- Fix GCC 6 compiler warnings

dtasev:
Description of work.
A few explicit casts added to remove warnings from `-Wconversion` flag.

No other actual warnings were found in the clean-fedora24 build, the rest are wrong warnings because of macros. I am not sure if there is a way to mute them.

**To test:**
Code review

Fixes #17593. 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/MDAlgorithms/src/SmoothMD.cpp
```diff
@@ -126,8 +126,8 @@ KernelVector gaussianKernel(const double fwhm) {
 KernelVector renormaliseKernel(KernelVector kernel,
                                std::vector<bool> validity) {
 
-  if (std::accumulate(validity.cbegin(), validity.cend(), 0.0) <
-      kernel.size()) {
+  if (std::accumulate(validity.cbegin(), validity.cend(),
+                      static_cast<size_t>(0)) < kernel.size()) {
```
quantumsteve: Can this be replaced by checking that `kernel.size() == validity.size()` and `std::find(validity.cbegin(), validity.cend(), [](decltype(kernel)::reference b){ return b == false;}) != validity.cend()`?

MantidQt/RefDetectorViewer/src/RefMatrixWSImageView.cpp
```diff
@@ -48,7 +48,7 @@ RefMatrixWSImageView::RefMatrixWSImageView(QString wpsName, int peakMin,
 
   std::vector<float> data(static_cast<size_t>(totalYMax) * sz);
 
-  for (size_t px = 0; px < totalYMax; px++) {
+  for (size_t px = 0; px < static_cast<size_t>(totalYMax); px++) {
```
quantumsteve: I think you are better off making `totalYMax` type `size_t` and then casting it when constructing `source`. 

Two other tiny things:
1. change `px++` to `++px`
2. use `boost::make_shared` to construct `source`.

Framework/MDAlgorithms/src/SmoothMD.cpp
quantumsteve: I think this function could be refactored to make fewer copies of `kernel` and `validity`. Perhaps this is best done as a separate issue?

Framework/DataObjects/inc/MantidDataObjects/TableColumn.h
```diff
@@ -1,31 +1,17 @@
...
-namespace Mantid {
+#include <sstream>
+#include <vector>
 
-//----------------------------------------------------------------------
-// Forward declarations
-//----------------------------------------------------------------------
+#include "MantidAPI/Column.h"
 
+namespace Mantid {
```
quantumsteve: Are these changed intended for a different PR?

Framework/Algorithms/src/ConvertTableToMatrixWorkspace.cpp
```diff
@@ -10,11 +10,6 @@
 #include "MantidKernel/Unit.h"
 #include "MantidKernel/UnitFactory.h"
 
-#include <boost/mpl/if.hpp>
-#include <boost/type_traits.hpp>
-
-#include <sstream>
-
```
quantumsteve: Are these changes intended for a different PR?

Framework/DataObjects/test/MDGridBoxTest.h
```diff
@@ -792,8 +792,8 @@ class MDGridBoxTest : public CxxTest::TestSuite {
           MDEventsTestHelper::makeRecursiveMDGridBox<2>(numSplit,
                                                         recurseLevels);
 
-      for (double x = 0; x < numSplit; x += spacing)
-        for (double y = 0; y < numSplit; y += spacing) {
+      for (double x = 0; x < static_cast<double>(numSplit); x += spacing)
+        for (double y = 0; y < static_cast<double>(numSplit); y += spacing) {
```
quantumsteve: In general I don't like [floating-point variables as loop counters](https://www.securecoding.cert.org/confluence/display/c/FLP30-C.+Do+not+use+floating-point+variables+as+loop+counters). That said, I don't see any benefit to making larger changes in a disabled test. 👍

quantumsteve (changes requested):
Changes reduce several warnings introduced in GCC 6. See other comments for places where I think the warnings can be better addressed.

Framework/MDAlgorithms/src/SmoothMD.cpp
DTasev: Do you mean something like passing by `const &` and create a `KernelVector` inside the `renormaliseKernel`, which we would then return by value? I am not sure if we could modify `gaussian_kernels[dimension_number]` in place directly.

Framework/DataObjects/test/MDGridBoxTest.h
DTasev: The link was a good read, thank you, I hadn't considered that. But I agree, it's a disabled test and I do not understand enough about what `spacing` is/does to easily make the change to another type.

Framework/MDAlgorithms/src/SmoothMD.cpp
quantumsteve: I *think* one can avoid copying `kernel` again in `normaliseKernel` by passing a reference. I haven't looked carefully enough to see it one could also do the same with `renormaliseKernel` and not sure if enough time is spent in these functions to justify the work.

Framework/DataObjects/inc/MantidDataObjects/TableColumn.h
DTasev: The changes were indeed for another PR https://github.com/mantidproject/mantid/pull/19928, but because that has been merged the files no longer show changes

MantidQt/RefDetectorViewer/src/RefMatrixWSImageView.cpp
```diff
@@ -69,5 +69,4 @@ RefIVConnections *RefMatrixWSImageView::getConnections() {
 
 RefMatrixWSImageView::~RefMatrixWSImageView() {
   // nothing to do here, since m_imageView is deleted when the window closes
-  delete m_imageView;
```
DTasev: This was an accidental commit, the code wasn't previously there and was added by accident by the parent commit e5bde51. 

However looking at the constructor code, I don't think `m_imageView` is ever deleted, because it has no parent, so this could be a possible memory leak?

DTasev:
I seem to be getting an error when building with Fedora24, and it also seems to be crashing on the build server: 
```
/home/dtasev/projects/mantid/Framework/Algorithms/src/EQSANSResolution.cpp: In member function ‘virtual double Mantid::Algorithms::EQSANSResolution::getTOFResolution(double)’:
/home/dtasev/projects/mantid/Framework/Algorithms/src/EQSANSResolution.cpp:19:30: error: ‘pow’ was not declared in this scope
     return 392.31 * pow(wl, 6) - 3169.3 * pow(wl, 5) + 10445 * pow(wl, 4) -
```

DTasev:
The Fedora 24/25 builds seem to fail because some of the files have functions `pow` , `fabs`, but the `<cmath>` header is not included. I also saw something about adding a `-lm` flag to `gcc` to link the math library, but that didn't seem to work, I had to manually add the `#include <cmath>` in the file for it to compile.

 @quantumsteve Maybe the work related to fixing those errors (i.e. adding the `<cmath>` header in the files that don't have it) should be done in another PR?

quantumsteve (approved):

PR #19931, opened 2017-06-27
Improve KafkaEventStreamDecoder test

martyngigg:
**Description of work.**

Uses condition variables instead of waiting on a fixed time period for the decoder to acquire enough information for the test. This should fix the unreliable nature of the test when run on the builders: [OSX](http://builds.mantidproject.org/job/pull_requests-osx/14290/testReport/), [Win7](http://builds.mantidproject.org/job/pull_requests-win7/15079/testReport/junit/LiveDataTest/KafkaEventStreamDecoderTest/test_Multiple_Period_Event_Stream/)

It required introducing the concept of a callback that could be registered and is called by the decoder at certain points. The tests then know where they are rather than assuming a given amount of time would be sufficient. The design is slightly more instrusive than I would like but the callbacks could prove useful outside of testing. 

@LamarMoore @matthew-d-jones What do you think of the approach?

**To test:**

It was difficult to get this to fail locally but having the system under load and then running the `KafkaEventStreamDecoderTest` seemed to make fail occasionally. It was more noticeable on the builders. 

*No issue*

**Release Notes** 
*Internal change. Does not need to be in the release notes.*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/LiveData/inc/MantidLiveData/Kafka/KafkaEventStreamDecoder.h
+++ b/Framework/LiveData/inc/MantidLiveData/Kafka/KafkaEventStreamDecoder.h
@@ -44,6 +44,9 @@ namespace LiveData {
 */
 class DLLExport KafkaEventStreamDecoder {
 public:
+  using CallbackFn = std::function<void()>;
+
+public:
   KafkaEventStreamDecoder(std::shared_ptr<IKafkaBroker> broker,
                           const std::string &eventTopic,
                           const std::string &runInfoTopic,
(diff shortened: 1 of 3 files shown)


matthew-d-jones (approved):
This looks like a more satisfactory solution.
I've streamed some data in as a quick functional test, all is well :+1:

martyngigg:
Thanks, @matthew-d-jones.

LamarMoore:
Looks good to me. If @matthew-d-jones  is happy I can merge this.

PR #19932, opened 2017-06-27
Fix unreliable NonOrthogonal unit test

martyngigg:
**Description of work.**

The [NonOrthogonalTest](http://builds.mantidproject.org/job/master_incremental/3054/label=win7/testReport/junit/MantidQtAPITest/NonOrthogonalTest/test_transformLookpointToWorkspaceCoordWithZeroZeroCoords/) was failing intermittently. It turned out that the test workspace was 2D but other parts of the code assumed it was >=3D (see [here](https://github.com/mantidproject/mantid/blob/77b6d91d074702660515b1f1a19bee105485f9ac/MantidQt/API/test/NonOrthogonalTest.h#L309)). Occasionally the out of bounds array access must have given back contents of adjacent memory locations that were not correct. 

The fix was to simply use a 3D data set.

While fixing this I decided to refactor this portion of the code to reduce the usage of bare C-style arrays whose size is known at compile time. Please decide if I have gone to far here for this fix. 

**To test:**

I could get this to fail most of the time on windows using `ctest -C Release -R MantidQtAPI -j5 --repeat-until-fail 20` but it's not guaranteed. This of course should definitely work now.

A code review is required and a functional test of the slice viewer, both orthogonal & non-orthogonal modes, is wise. 

No issue.

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


AntonPiccardoSelg:
I started with the functional testing. The orthogonal functionality seems to be broken.

Mantid v3.10 in non-orthogonal mode

image

PR in non-orthogonal mode

image

martyngigg:
Hmm. There are also tests failing that I didn't spot.

Perhaps it's safer if I just fix the issue with that 1 test method. What do you think?

martyngigg:
If we are talking about looking at `matplotlib` as a replacement in the future maybe there's not much point in putting too much work into this?

martyngigg:
@AntonPiccardoSelg Thanks for catching that. The most recent commit should have fixed that problem.

AntonPiccardoSelg:
The SliceViewer shows a skewed images again as is expected for the data set.

Windows and OSX still seem to have issues with the unit tests

MantidQt/SliceViewer/src/CoordinateTransform.cpp
```diff
@@ -23,34 +23,37 @@ void NullTransform::checkDimensionsForHKL(Mantid::API::IMDWorkspace_sptr ws,
...
 
 NonOrthogonalTransform::NonOrthogonalTransform(
-    Mantid::API::IMDWorkspace_sptr ws, size_t dimX, size_t dimY)
-    : m_dimensionsHKL(true) {
+    const Mantid::API::IMDWorkspace_const_sptr &workspace, size_t dimX,
+    size_t dimY)
+    : m_dimensionsHKL(true), m_skewMatrix({0}) {
   // Set the skewMatrix for the non-orthogonal data
-  auto numberOfDimensions = ws->getNumDims();
+  auto numberOfDimensions = workspace->getNumDims();
```
quantumsteve: Should `workspace` be passed as a `const Mantid::API::IMDWorkspace &`  since it cannot be `NULL`?

martyngigg: @quantumsteve Ideally, yes. My changes were already propagating further than I would have ideally liked thought and this then ends up requiring many other functions to change. Do you think this is worth doing?

quantumsteve: A reference would better describe ownership, but I also understand the desire to keep pull requests small. Perhaps it's best to make a note of it in a separate issue?

martyngigg: Agreed. I'll make a separate issue.

MantidQt/API/src/NonOrthogonal.cpp
```diff
@@ -417,10 +419,9 @@ K = M24X + M22Y + M23Z
 L = M31X + M32Y + M33Z
 
 */
-void transformLookpointToWorkspaceCoord(Mantid::coord_t *lookPoint,
-                                        const Mantid::coord_t skewMatrix[9],
-                                        const size_t &dimX, const size_t &dimY,
-                                        const size_t &dimSlice) {
+void transformLookpointToWorkspaceCoord(
+    Mantid::coord_t *lookPoint, const std::array<Mantid::coord_t, 9> skewMatrix,
```
AntonPiccardoSelg: Should this not be `const std::array<Mantid::coord_t, 9>& skewMatrix`?

MantidQt/API/src/QwtRasterDataMDNonOrthogonal.cpp
```diff
@@ -79,10 +66,7 @@ double QwtRasterDataMDNonOrthogonal::value(double x, double y) const {
 void QwtRasterDataMDNonOrthogonal::setWorkspace(IMDWorkspace_const_sptr ws) {
   QwtRasterDataMD::setWorkspace(ws);
   // Create a lookpoint
-  if (m_lookPoint) {
-    delete[] m_lookPoint;
-  }
-  m_lookPoint = new coord_t[m_nd];
+  m_lookPoint.resize(m_nd);
```
AntonPiccardoSelg: :+1:

MantidQt/API/test/NonOrthogonalTest.h
```diff
@@ -224,124 +79,215 @@ class NonOrthogonalTest : public CxxTest::TestSuite {
...
+    skewMatrix[1][1] = 1;
+    skewMatrix[1][2] = 0;
+    skewMatrix[2][0] = 0;
+    skewMatrix[2][1] = 0;
+    skewMatrix[2][2] = 1.1547;
+    return skewMatrix;
+  }
+
+  template <typename T> bool skewWithinTolerance(T coord, T target) {
+    if ((coord >= target - 0.000005) && (coord <= target + 0.000005)) {
```
AntonPiccardoSelg: could we use ` return (coord >= target - 0.000005) && (coord <= target + 0.000005)`

MantidQt/API/test/NonOrthogonalTest.h
```diff
@@ -224,124 +79,215 @@ class NonOrthogonalTest : public CxxTest::TestSuite {
...
+      coordinates = {{1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0}};
+    } else {
+      coordinates = {{1.0, 0.0, static_cast<Mantid::coord_t>(-0.57735), 0.0,
+                      1.0, 0.0, 0.0, 0.0,
+                      static_cast<Mantid::coord_t>(1.1547)}};
+    }
+    return coordinates;
+  }
+
+  enum DimensionIndex { DimX = 0, DimY = 1, SliceDim = 2 };
```
AntonPiccardoSelg: It seems that `DimensionIndex` is not used anywhere. We can remove it if so.

AntonPiccardoSelg (changes requested):
The changes are a good improvement of the code, especially around the test.

The comments I have are minor fixes.

MantidQt/API/src/NonOrthogonal.cpp
martyngigg: You're right, it should.

MantidQt/API/test/NonOrthogonalTest.h
martyngigg: `DimensionIndex` itself is not but `DimX` etc are used throughout the test. This is a plain old `enum` so that you can avoid `static_casts` everywhere as we just want to use it as an `unsigned`. This means no qualification with the `enum` name is necessary.

MantidQt/API/src/NonOrthogonal.cpp
martyngigg: That's weird. The header has it defined as a reference. I'm surprised this linked...

MantidQt/API/test/NonOrthogonalTest.h
martyngigg: :+1:

MantidQt/API/test/NonOrthogonalTest.h
AntonPiccardoSelg: :man_facepalming: yep, must have been sleeping

martyngigg:
@AntonPiccardoSelg Thanks for the comments. I've cleaned up those bits.

@quantumsteve I also cleaned up a few of those shared_ptr->const reference issues as a couple of them were actually easier than they first looked.

AntonPiccardoSelg (approved):

AntonPiccardoSelg:
Code changes look good. Functional testing works as expected. Please :shipit: once the tests have passed.

PR #19937, opened 2017-06-27
Add math.h to list of includes in some .cpp files

peterfpeterson:
`master` wasn't compiling so this adds `#import <math.h>` in three `.cpp` files that needed it.

**To test:**

If the builds pass, this can be merged.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/src/EQSANSResolution.cpp
+++ b/Framework/Algorithms/src/EQSANSResolution.cpp
@@ -2,6 +2,7 @@
 // Includes
 //----------------------------------------------------------------------
 #include "MantidAlgorithms/EQSANSResolution.h"
+#include <math.h>
 
 namespace Mantid {
 namespace Algorithms {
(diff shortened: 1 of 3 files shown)


Framework/Algorithms/src/EQSANSResolution.cpp
```diff
@@ -2,6 +2,7 @@
 // Includes
 //----------------------------------------------------------------------
 #include "MantidAlgorithms/EQSANSResolution.h"
+#include <math.h>
```
AndreiSavici: This should be `#include <cmath>`

AndreiSavici: See for example https://stackoverflow.com/questions/10694255/c-cmath-vs-math-h-and-similar-c-prefixed-vs-h-extension-headers
Then all math functions should be defined in the `std` namespace. So replace `pow` with `std::pow`

rosswhitfield (approved):
master compiles again with this change :+1:

PR #19940, opened 2017-06-28
IndexProperty Implementation Prototype

### LINKED ISSUE #19130: WorkspacePropertyWithIndex property

LamarMoore:
**DO NOT MERGE**
**NOT READY FOR REVIEW**
This is an alternative implementation to the implementation of an IndexProperty see [here](https://github.com/mantidproject/mantid/pull/19699) for the first attempt. This [design](https://github.com/mantidproject/documents/pull/42) outlines the rational behind developing this property type. 

**To test:**
Code review. Ensure the property displays correctly in the GUI and can be accessed from python correctly. The `ChangePulsetime2` algorithm was implemented specifically to demonstrate the use of this property.

Fixes #19130. 

**Release Notes** 
Release notes need to be updated in due course and the concepts pages.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/API/inc/MantidAPI/Algorithm.h
```diff
@@ -189,6 +193,163 @@ class MANTID_API_DLL Algorithm : public IAlgorithm,
   /// Override if the algorithm is not part of the Mantid distribution.
   const std::string helpURL() const override { return ""; }
 
+  template <typename T> void checkWorkspaceType() const {
```
SimonHeybrock: - Why public?
- Maybe a better name?
- Why member?
- Do we actually need a function for this? Is the compiler error from `std::is_convertible` hard to read?

Framework/API/inc/MantidAPI/Algorithm.h
```diff
@@ -189,6 +193,163 @@ class MANTID_API_DLL Algorithm : public IAlgorithm,
...
+    static_assert(std::is_convertible<T *, MatrixWorkspace *>::value,
+                  "Workspace type needs to be convertible to MatrixWorkspace.");
+  }
+
+  inline bool isReserved(const std::vector<std::string> &list,
```
SimonHeybrock: - Why member?
- Maybe `isCompoundProperty` or a better name?

Framework/API/inc/MantidAPI/Algorithm.h
```diff
@@ -189,6 +193,163 @@ class MANTID_API_DLL Algorithm : public IAlgorithm,
...
+
+  using IPropertyManager::declareProperty;
+  using PropertyManagerOwner::declareProperty;
+  void declareProperty(std::unique_ptr<Kernel::Property> p,
+                       const std::string &doc = "") override;
+
+  Kernel::IPropertyManager::TypedValue
+  getProperty(const std::string &name) const override;
+
+  template <typename Workspace>
```
SimonHeybrock: This looks dangerous (or at least confusing), given that we also have a type named `Workspace`.

Framework/API/inc/MantidAPI/Algorithm.h
```diff
@@ -189,6 +193,163 @@ class MANTID_API_DLL Algorithm : public IAlgorithm,
...
+
+    *indexTypeProp = type;
+
+    *indexProp = list;
+  }
+
+  template <typename Workspace>
+  void setIndexProperty(const std::string &name,
+                        const boost::shared_ptr<Workspace> &wksp,
+                        IndexType type, const std::string &list) {
```
SimonHeybrock: Can you avoid some code duplication here by templating the type of `list`?

Framework/API/inc/MantidAPI/Algorithm.h
```diff
@@ -189,6 +193,163 @@ class MANTID_API_DLL Algorithm : public IAlgorithm,
...
+
+    *wsProp = wksp;
+
+    *indexTypeProp = type;
+
+    indexProp->setValue(list);
+  }
+
+  template <typename Workspace>
+  void setIndexProperty(const std::string &name, const std::string &wsName,
```
SimonHeybrock: Same here - template for the type of `wksp` (above) and `wsName`?

Framework/API/inc/MantidAPI/IndexProperty.h
```diff
@@ -0,0 +1,63 @@
...
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+class MANTID_API_DLL IndexProperty : public Kernel::ArrayProperty<int> {
+public:
+  explicit IndexProperty(
```
SimonHeybrock: Why `explicit` if there is more than one argument?

Framework/API/src/Algorithm.cpp
```diff
@@ -244,18 +244,20 @@ const std::string Algorithm::workspaceMethodInputProperty() const { return ""; }
...
+ *  @throw runtime_error Thrown if algorithm or Child Algorithm cannot be
+ *initialised
+ *
+ */
 void Algorithm::initialize() {
   // Bypass the initialization if the algorithm has already been initialized.
   if (m_isInitialized)
     return;
 
+  m_reservedList.clear();
```
SimonHeybrock: Why?

Framework/API/src/Algorithm.cpp
```diff
@@ -1622,6 +1628,26 @@ bool Algorithm::getAlgStartupLogging() const {
   return m_isAlgStartupLoggingEnabled;
 }
 
+void Algorithm::declareProperty(std::unique_ptr<Kernel::Property> p,
+                                const std::string &doc) {
+  if (isReserved(m_reservedList, p->name()))
```
SimonHeybrock: I think this is not sufficient, you also need to check the `WorkspaceIndexType` and `WorkspaceIndexSet` names?

Framework/API/src/IndexProperty.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+  try {
+    getIndices();
+  } catch (std::runtime_error &e) {
+    error = e.what();
+  } catch (std::out_of_range &) {
+    error = "Indices provided to IndexProperty are out of range.";
+  } catch (std::logic_error &) {
+    error = "Duplicate indices supplied to IndexProperty.";
+  }
```
SimonHeybrock: Probably this is sufficient for the first iteration but I foresee complaints about not providing the offending indices (this would need to change the error handling in `IndexInfo`/`SpectrumNumberTranslator`.

Framework/API/test/AlgorithmTest.h
```diff
@@ -159,6 +160,24 @@ const std::string FailingAlgorithm::FAIL_MSG("Algorithm failed as requested");
 
 DECLARE_ALGORITHM(FailingAlgorithm)
 
+class IndexingAlgorithm : public Algorithm {
+public:
+  IndexingAlgorithm() : Algorithm() {}
+  ~IndexingAlgorithm() override {}
```
SimonHeybrock: I don't think you need constructor and destructor?

Framework/API/test/AlgorithmTest.h
```diff
@@ -787,6 +806,86 @@ class AlgorithmTest : public CxxTest::TestSuite {
     TS_ASSERT_EQUALS(algCastConst, algCastNonConst);
   }
 
+  void testIndexingAlgorithm_declareIndexPropertyMethod() {
+    IndexingAlgorithm indexAlg;
+    TS_ASSERT_THROWS_NOTHING(
+        indexAlg.declareIndexProperty<MatrixWorkspace>("InputWorkspace"));
```
SimonHeybrock: I was surprised here. Is `declareIndexProperty` public? Is `declareProperty` public??

Framework/API/test/AlgorithmTest.h
```diff
@@ -787,6 +806,86 @@ class AlgorithmTest : public CxxTest::TestSuite {
...
+    AnalysisDataService::Instance().add("wksp", wksp);
+    indexAlg.setIndexProperty<MatrixWorkspace>(
+        "InputWorkspace", "wksp", IndexType::WorkspaceIndex, "1:5");
+
+    MatrixWorkspace_sptr wsTest;
+    Indexing::SpectrumIndexSet indexSet(0);
+
+    TS_ASSERT_THROWS_NOTHING(
+        std::tie(wsTest, indexSet) =
+            indexAlg.getIndexProperty<MatrixWorkspace>("InputWorkspace"));
```
SimonHeybrock: Do I understand correctly that this would now also work with C++17 structured bindings, without extra casting?
```cpp
auto [ws, indexSet] = indexAlg.getIndexProperty<MatrixWorkspace>("InputWorkspace"));
```

Framework/Algorithms/inc/MantidAlgorithms/ChangePulsetime2.h
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Mantid {
+namespace Algorithms {
+
+/** ChangePulsetime : TODO: DESCRIPTION
+ *
+ * @author
+ * @date 2011-03-31 09:31:55.674594
+ */
+class DLLExport ChangePulsetime2 : public API::Algorithm {
```
SimonHeybrock: I suppose this is only a temporary addition and will be removed before merging? Or should we have this as version 2 of the algorithm?

Framework/Algorithms/src/ChangePulsetime2.cpp
```diff
@@ -0,0 +1,64 @@
...
+                      "OutputWorkspace", "", Direction::Output),
+                  "An output event workspace.");
+}
+
+//----------------------------------------------------------------------------------------------
+/** Execute the algorithm.
+ */
+void ChangePulsetime2::exec() {
+  EventWorkspace_const_sptr in_ws;
+  Indexing::SpectrumIndexSet indexSet(0);
```
SimonHeybrock: Is there no default constructor on this class? Maybe it should be added.

SimonHeybrock (changes requested):
Looks really nice overall, just some comments on details below.

I have not been exceptionally thorough, will do another pass once you have it closer to something ready for merging.

Framework/API/inc/MantidAPI/Algorithm.h
LamarMoore: Hey sorry about that, this wasn't meant for review yet. There are a few things like that which need be changed (copied and pasted from other code).

Framework/API/inc/MantidAPI/Algorithm.h
LamarMoore: Doesn't work because of the shared_ptr business.

Framework/API/test/AlgorithmTest.h
LamarMoore: Yes because it explicitly returns a tuple.

Framework/API/inc/MantidAPI/Algorithm.h
SimonHeybrock: Can you elaborate?

Framework/API/src/Algorithm.cpp
LamarMoore: No, they are all created at the same time and only added to this reserved list in `declareIndexProperty`

SimonHeybrock: My point is:
```cpp
declareIndexProperty("Input");
declareProperty("InputIndexType");
```
Should you forbid the latter? Are you?

Framework/API/test/AlgorithmTest.h
LamarMoore: `IPropertyManager::declareProperty(std::unique_ptr<Property>, const std::string &)` is public. Other variations within IPropertyManager are protected.

Framework/API/src/Algorithm.cpp
LamarMoore: So that when the algorithm tries to declare the property in the init method, it doesn't fail due to the check on whether or not the property name is already being used.

SimonHeybrock: - What code outside (&before) `init()` would declare index properties?
- If it did, why would you want to clear that vector??

Framework/API/src/Algorithm.cpp
LamarMoore: Ah I see your point now.

LamarMoore:
I have compared the performance test results for `ChangePulsetime` versions 1 and 2(which uses IndexProperty) and there is no change in performance.

Framework/API/inc/MantidAPI/Algorithm.tcc
```diff
@@ -0,0 +1,133 @@
...
+                             "with properties declared using "
+                             "declareIndexProperty.");
+  auto *wsProp =
+      dynamic_cast<WorkspaceProperty<T1> *>(getPointerToProperty(name));
+  auto *indexTypeProp = dynamic_cast<IndexTypeProperty *>(
+      getPointerToProperty(name + "IndexType"));
+  auto *indexProp =
+      dynamic_cast<IndexProperty *>(getPointerToProperty(name + "IndexSet"));
+
+  wsProp->setValue(wsName);
```
SimonHeybrock: As mentioned before: This seems to be the only line different between the the variants of `setIndexProperty`, so why not template it and avoid code duplication?

Framework/Kernel/src/ArrayProperty.cpp
```diff
@@ -1,103 +1,9 @@
 #include "MantidKernel/ArrayProperty.h"
 
-// PropertyWithValue implementation
-#include "MantidKernel/PropertyWithValue.tcc"
-
+// ArrayProperty Definition
+#include "MantidKernel/ArrayProperty.tcc"
```
SimonHeybrock: Why was this necessary?

Framework/Algorithms/src/ChangePulsetime2.cpp
```diff
@@ -4,6 +4,7 @@
 #include "MantidKernel/ArrayProperty.h"
 #include "MantidKernel/System.h"
 
+#include "MantidAPI/Algorithm.tcc"
```
SimonHeybrock: As mentioned above... can avoid this by using explicit instatiations.

Framework/API/src/Algorithm.cpp
```diff
@@ -32,6 +32,9 @@
 
 #include <map>
 
+// Index property handling template definitions
+#include "MantidAPI/Algorithm.tcc"
```
SimonHeybrock: Unless you put explicit instantiations somewhere you will be forced to include this in all algorithms using the new property.

SimonHeybrock (commented):
Small comments:

Framework/API/inc/MantidAPI/Algorithm.tcc
LamarMoore: I need to add `WorkspaceProperty::operator=(const std::string &)`.

Framework/Kernel/src/ArrayProperty.cpp
LamarMoore: I get a whole bunch of compiler warnings otherwise about not being able to find the definition of `ArrayProperty<int>`

Framework/API/src/Algorithm.cpp
LamarMoore: There are explicit instantiations in ArrayProperty.cpp but for some reason I keep getting the warnings. The code compiles, but with loads of warnings.

LamarMoore: That being said, I would be grateful if anyone could give me some advice on how to get around this problem.

Framework/API/inc/MantidAPI/Algorithm.tcc
SimonHeybrock: Can't you just put this single line in a templated function to distinguish betwen the `std::string` and `shared_ptr` cases? Instead of duplicating the whole big method.

Framework/API/src/Algorithm.cpp
SimonHeybrock: Add explicit instantiations for the methods you define in `Algorithm.tcc`?

LamarMoore: But I would need to do this for every workspace type I intend to use...?

SimonHeybrock: Yes, similar to what is done here: https://github.com/mantidproject/mantid/blob/master/Framework/DataObjects/src/WorkspaceProperty.cpp.

Framework/API/inc/MantidAPI/Algorithm.tcc
```diff
@@ -2,7 +2,9 @@
 #include "MantidAPI/IndexProperty.h"
 #include "MantidAPI/WorkspaceProperty.h"
 
+#ifdef _WIN32
 #include "MantidKernel/ArrayProperty.tcc"
```
LamarMoore: I have done something ugly here but this is currently the only way I can get the build to work on both windows and linux. Another pair of eyes on how I have defined `IndexProperty` and the methods in `Algorithm` for managing this property would be helpful. N.B. adding the definitions of `ArrayProperty` methods to the header opens a whole other can of worms and causes the build to fail.

LamarMoore:
I have done something ugly in [Algorithm.tcc](https://github.com/mantidproject/mantid/pull/19940/files/2736e9ec2bb52715e62a0e187e61540feee378c4..6db7d4e4b52c1b55b1af8cbc19f4a6f687864cbd#diff-5c6bf53f4b50005f9771ca9738b6ebc1) but this is currently the only way I can get the build to work on both windows and linux. Another pair of eyes on how I have defined IndexProperty and the methods in Algorithm for managing this property would be helpful. N.B. adding the definitions of ArrayProperty methods to the header opens a whole other can of worms and causes the build to fail.

Framework/API/inc/MantidAPI/IndexProperty.h
```diff
@@ -0,0 +1,67 @@
...
+#include "MantidAPI/DllConfig.h"
+#include "MantidAPI/IWorkspaceProperty.h"
+#include "MantidAPI/IndexTypeProperty.h"
+#include "MantidIndexing/SpectrumIndexSet.h"
+#include "MantidKernel/ArrayProperty.h"
+
+namespace Mantid {
+namespace API {
+
+/** IndexProperty : TODO: DESCRIPTION
```
OwenArnold: Missing description

Framework/API/inc/MantidAPI/IndexTypeProperty.h
```diff
@@ -0,0 +1,61 @@
...
+
+#include "MantidAPI/DllConfig.h"
+#include <MantidKernel/PropertyWithValue.h>
+
+namespace Mantid {
+namespace API {
+
+enum IndexType { SpectrumNum = 1, WorkspaceIndex = 2 };
+
+/** IndexTypeProperty : TODO : Add support for DetectorIDs
```
OwenArnold: Missing TODO. 

Another minor point. I think the name `IndexTypeProperty` next to the `IndexProperty` are a little confusing. I'm not sure that `IndexTypeProperty` is a great name (but I can't think of anything better for now). One problem I have with it is that `SpectrumNumber` , for example, is not an `IndexType`

OwenArnold (changes requested):
A good start. I like the fact that you've added lots of tests as you've been developing this.

Framework/API/inc/MantidAPI/Algorithm.tcc
LamarMoore: This is the issue was speaking about before. If I remove the include, the build fails in Windows. If I remove the platform guard, this fails in Ubuntu. I tried other solutions to ensure Windows could find the template definitions but this is the only thing that works (see my previous comment on the PR)

Framework/API/inc/MantidAPI/Algorithm.tcc
```diff
@@ -0,0 +1,142 @@
...
+* This method can only be used if T1 is convertible to a MatrixWorkspace and
+* T2 is either std::string or std::vector<int>
+
+@param name Property name
+@param wsName Workspace as a pointer
+@param type Index type IndexType::WorkspaceIndex or IndexType::SpectrumNum
+@param list List of indices to be used.
+*/
+template <typename T1, typename T2, typename, typename>
+void Algorithm::setIndexProperty(const std::string &name,
```
LamarMoore: Because `WorkspaceProperty` supports this as users, especially in the context of python scripts, will expect to be able to set the workspace using its name.

Framework/API/inc/MantidAPI/IndexProperty.h
```diff
@@ -33,7 +33,7 @@ namespace API {
   File change history is stored at: <https://github.com/mantidproject/mantid>
   Code Documentation is available at: <http://doxygen.mantidproject.org>
 */
-class MANTID_API_DLL IndexProperty : public Kernel::ArrayProperty<int> {
+class DLLExport IndexProperty : public Kernel::ArrayProperty<int> {
```
LamarMoore: That has been reverted. I did this because I noticed that other classes which inherited `PropertyValue` with some specific type only used the `DLLExport` macro instead of the general import/export macro.

Framework/API/inc/MantidAPI/Algorithm.h
```diff
@@ -189,6 +191,40 @@ class MANTID_API_DLL Algorithm : public IAlgorithm,
...
 
+  using IPropertyManager::declareProperty;
+
+  using PropertyManagerOwner::declareProperty;
+
+  void declareProperty(std::unique_ptr<Kernel::Property> p,
+                       const std::string &doc = "") override;
+
+  Kernel::IPropertyManager::TypedValue
+  getProperty(const std::string &name) const override;
```
LamarMoore: Yes, see `m_reservedList`. If a user attempts to access a property which is compound, an exception will be thrown. This is why it was necessary to override getProperty/declareProperty/setProperty.

Framework/API/src/Algorithm.cpp
```diff
@@ -1651,6 +1656,31 @@ bool Algorithm::getAlgStartupLogging() const {
...
+  if (isCompoundProperty(name))
+    throw std::runtime_error("Algorithm::getIndexProperty must be used with "
+                             "properties declared using "
+                             "Algorithm::declareIndexProperty.");
+
+  return PropertyManagerOwner::getProperty(name);
+}
+
+bool Algorithm::isCompoundProperty(const std::string &name) const {
+  return std::find(m_reservedList.cbegin(), m_reservedList.cend(), name) !=
```
LamarMoore: Yes, but at the moment I am not sure if there is anything preventing two properties with the same name in general. However, I cannot think of another way of hiding the creation of the three properties and each property must have a name in order to be accessible.

Framework/API/inc/MantidAPI/IndexTypeProperty.h
LamarMoore: The DetectorID support will not be added now because we need to develop a method for translating from DetectorID to SpectrumIndexSet. Simon has pushed this down the priority list since there are not many cases where this is used compared to spectrum numbers and workspace indices.

Maybe `IndexPropertyConversionType`? (Seems like a lot to type). I guess what we want to convey is that the user will receive a `SpectrumIndexSet` after converting from this selected type. We also want to establish a relationship with the `IndexProperty`.

OwenArnold: As I say I don't have a better name. Maybe just leave it as it is for now.

Framework/API/src/Algorithm.cpp
OwenArnold: I don't think there's anything you can/should programatically do. My concern is that for ease of development, it's going to be better practice for new algorithms to make the `declareIndexProperty` calls first, because that's where the reserved list gets populated. For existing algorithms, this could cause some issues because the order of declaration defines the order of properties in the GUI dialogs (AlgorithmDialog) and dictates the layout for python functions generated in the `simpleapi`. I think this is something we might want to consider as we get closer to a rollout.

Framework/API/inc/MantidAPI/IndexTypeProperty.h
LamarMoore: TODO removed and [this issue](https://github.com/mantidproject/mantid/issues/20013) has been created for the future implementation of the DetectorID->SpectrumIndex conversion

LamarMoore:
This is the dialog created using this mechanism for the `ChangePulsetime2` algorithm which was created as part of this PR to demonstrate usage.

image

The python code generated by using this algorithm is shown below:
```python
Load(Filename=r'C:\mantid\build\ExternalData\Testing\Data\UnitTest\BSS_11841_event.nxs', OutputWorkspace='BSS_11841_event')
ChangePulsetime2(InputWorkspace='BSS_11841_event', InputWorkspaceIndexType='WorkspaceIndex', InputWorkspaceIndexSet='1:50', TimeOffset=30, OutputWorkspace='WS')
```

Framework/API/src/IndexProperty.cpp
```diff
@@ -0,0 +1,163 @@
...
+}
+
+Indexing::SpectrumIndexSet IndexProperty::getIndices() const {
+  if (m_indicesExtracted)
+    return m_indices;
+
+  auto wksp = boost::dynamic_pointer_cast<MatrixWorkspace>(
+      m_workspaceProp.getWorkspace());
+
+  if (wksp == nullptr)
```
OwenArnold: Optional: You could combine the declaration, assignment and null check into one line here.

Framework/API/src/IndexProperty.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+std::string IndexProperty::value() const {
+  if (m_min != m_max)
+    return std::to_string(m_min) + ":" + std::to_string(m_max);
+  else
+    return ArrayProperty<int>::value();
+}
+
+Indexing::SpectrumIndexSet IndexProperty::getIndices() const {
+  if (m_indicesExtracted)
```
OwenArnold: For what reason is this caching optimisation in place? If there is a good reason, could you add a comment to explain why this is done?

Framework/API/inc/MantidAPI/IndexProperty.h
```diff
@@ -0,0 +1,74 @@
...
+  std::vector<int> &operator=(const std::vector<int> &rhs) override;
+  std::string operator=(const std::string &rhs);
+  std::string setValue(const std::string &value) override;
+  Indexing::SpectrumIndexSet getIndices() const;
+  std::string value() const override;
+
+private:
+  const IWorkspaceProperty &m_workspaceProp;
+  const IndexTypeProperty &m_indexTypeProp;
+  int m_min;
```
OwenArnold: Could you be a bit more descriptive with names and comments for `m_min` and `m_max`? Presumably these are treated as inclusive?

OwenArnold: is `int` the right type to use here? Is there any case for having a signed integer, < 0?

Framework/Algorithms/src/ChangePulsetime2.cpp
```diff
@@ -0,0 +1,63 @@
...
+using namespace Mantid::Kernel;
+using namespace Mantid::API;
+using namespace Mantid::DataObjects;
+using std::size_t;
+
+//----------------------------------------------------------------------------------------------
+/** Initialize the algorithm's properties.
+ */
+void ChangePulsetime2::init() {
+  declareIndexProperty<EventWorkspace>("InputWorkspace");
```
OwenArnold: I have discussed this with @LamarMoore, but would like a wider discussion about having implicit defaults for the index type in the declaration @martyngigg. I think I would prefer that the index type had to be explicitly declared (thus thought about). Workspace Indexes are great for internal use (development), but can lead to flakey scripts.

Framework/Algorithms/src/ChangePulsetime2.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include "MantidAPI/WorkspaceFactory.h"
+#include "MantidDataObjects/EventWorkspace.h"
+#include "MantidKernel/ArrayProperty.h"
+#include "MantidKernel/System.h"
+
+namespace Mantid {
+namespace Algorithms {
+
+// Register the algorithm into the AlgorithmFactory
+DECLARE_ALGORITHM(ChangePulsetime2)
```
OwenArnold: It would be nice to make a proper v2 of this algorithm and distribute it as part of these changes so that it's clear for future developers how these features are to be used. It does however yield a breaking change to the property names. In it's current form, I think this algorithm must be omitted from our public/user algorithm namespaces to prevent real use and dependency on.

OwenArnold (changes requested):
This is starting to look very good. I've made some suggestions for improvements. I tried the feature out and it seemed to work fine. I'm not 100% sure I like the `simple.api` side of the changes, but I don't have any good alternatives at this point.

Framework/API/inc/MantidAPI/Algorithm.tcc
```diff
@@ -0,0 +1,144 @@
+#include "MantidAPI/Algorithm.h"
+#include "MantidAPI/IndexProperty.h"
+#include "MantidAPI/WorkspaceProperty.h"
+
+/* Currently required for Windows. Otherwise warning C4661 is issued. The
+ * ArrayProperty exlicit instantiations are not found.*/
+#ifdef _WIN32
```
OwenArnold: I think we need to somehow get to the bottom of this. If this is a compiler issue, we need to have a bug reference. Otherwise we will never know to be able to remove the preprocessor check.

Framework/API/src/Algorithm.cpp
martyngigg: `declareProperty` currently throws if you try and declare something that already has that name so we are safe there.

Framework/API/src/IndexProperty.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+  return error;
+}
+
+std::string IndexProperty::operator=(const std::string &rhs) {
+  return setValue(rhs);
+}
+
+Indexing::SpectrumIndexSet IndexProperty::getIndices() const {
+  MatrixWorkspace_sptr wksp;
```
martyngigg: Why not do
```c++
MatrixWorkspace_sptr wksp = boost::dynamic_pointer_cast<MatrixWorkspace>(m_workspaceProp.getWorkspace());
if(!wksp) throw ...
```
?

Framework/Algorithms/src/ChangePulsetime2.cpp
martyngigg: :+1:

Framework/Algorithms/src/ChangePulsetime2.cpp
martyngigg: I think I agree here that it's nicer to only have to look in 1 place to see what you are declaring. I guess it also depends if there is an obvious choice for a default, I'm not sure that is the case here...

On the face of it here you have no idea that something else has been declared for you with some other type. As mentioned previously I think it would also be good to indicate that this declares more that 1 property.

Framework/Algorithms/src/ChangePulsetime2.cpp
martyngigg: :+1:

martyngigg (commented):
My initial set of comments.

Mainly I'm slightly uneasy about the number of dynamic casts we have to keep making. It always feels to me that there's a more fundamental issue somewhere when we keeping having to ask what the types are.

Without some more thought I'm not sure I could offer an better solution at present though.

Framework/API/inc/MantidAPI/Algorithm.tcc
```diff
@@ -0,0 +1,138 @@
...
+                                     const int allowedIndexTypes,
+                                     PropertyMode::Type optional,
+                                     LockMode::Type lock,
+                                     const std::string &doc) {
+  declareProperty(
+      Kernel::make_unique<WorkspaceProperty<T>>(
+          propertyName, "", Kernel::Direction::Input, optional, lock),
+      doc);
+
+  declareProperty(Kernel::make_unique<IndexTypeProperty>(
```
LamarMoore: @martyngigg Would it be a compromise to use methods to build the names from the property name?

martyngigg: Yes I think a central method for doing that would be good. Perhaps a static on each class type?

LamarMoore: Why do the class types need to be aware of that? I though you would have methods in an anonymous  namespace with the methods in Algorithm.tcc.

Framework/API/src/Algorithm.cpp
SimonHeybrock: Giving this another review pass for all the changes while I was on vacation. You did not manage to resolve this issue as suggested?

Framework/API/src/Algorithm.cpp
SimonHeybrock: Also not resolved yet?

Framework/Algorithms/src/ChangePulsetime2.cpp
SimonHeybrock: 👍

Framework/Kernel/inc/MantidKernel/ArrayProperty.h
```diff
@@ -66,6 +66,98 @@ class ArrayProperty : public PropertyWithValue<std::vector<T>> {
...
+ *  @param value :: The values to assign to the property, given as a
+ * comma-separated list
+ *  @return True if the assignment was successful
+ */
+template <typename T>
+std::string ArrayProperty<T>::setValue(const std::string &value) {
+  // Implemented this method for documentation reasons. Just calls base class
+  // method.
+  return PropertyWithValue<std::vector<T>>::setValue(value);
+}
```
SimonHeybrock: Why was this moved from the `cpp`?

LamarMoore: Compiler warning on Windows. See conversation history.

Framework/Algorithms/src/ChangePulsetime2.cpp
```diff
@@ -7,8 +7,8 @@
 namespace Mantid {
 namespace Algorithms {
 
-// Register the algorithm into the AlgorithmFactory
-DECLARE_ALGORITHM(ChangePulsetime2)
+// Algorithm not registered as this will be used for testing MPI
+// DECLARE_ALGORITHM(ChangePulsetime2)
```
LamarMoore: This can be uncommented to test the ChangePulsetime2 algorithm with the GUI (workspace history, dialog box generation etc.)

Framework/Algorithms/src/ChangePulsetime2.cpp
```diff
@@ -0,0 +1,63 @@
+#include "MantidAlgorithms/ChangePulsetime2.h"
+#include "MantidAPI/WorkspaceFactory.h"
+#include "MantidDataObjects/EventWorkspace.h"
+#include "MantidKernel/ArrayProperty.h"
```
SimonHeybrock: Not necessary?

SimonHeybrock (approved):

OwenArnold (approved):

martyngigg (approved):

PR #19941, opened 2017-06-28
Update ParaView to v5.4.0

quantumsteve:
Description of work.

This updates Mantid to use [ParaView 5.4.0](https://blog.kitware.com/paraview-5-4-0-release-notes/). Users will notice the updated color legend that chooses nicer increments that can be represented with fewer digits. For the VSI, we remove our custom `SaveScreenshotReaction` in favor of the refactored `pqSaveScreenshotReaction` class. Since "AllViews" is no longer selected by default, our custom changes are no longer necessary.

I wasn't able to remove the `-Wpedantic` warning from `vtktypemacro`. Given the potential dangers from extra semicolons, I removed this flag within the Vates subdirectory.

I created a [separate issue](https://github.com/mantidproject/mantid/issues/19946) to address complications with ParaView 5.4 and older system versions of jsoncpp.

**To test:**

1. View the updated [build instructions](https://www.mantidproject.org/Building_VATES).
2. Build Mantid with -DMAKE_VATES=ON and -DParaView_DIR=\<your ParaView build directory\>.
3. Try viewing a MDHistoWorkspace and a MDEventWorkspace in the VSI.
4. Save a screenshot or two.

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


quantumsteve:
It's now building successfully on all platforms :tada:

AndreiSavici (approved):
It seems to work on build servers. Code review is OK. I have problems building it on Ubuntu 16.10 (tbb library). Time for me to upgrade

AntonPiccardoSelg (approved):
Changes look fine. MDHisto and MDEvent operations seem to be fine. Taking a screenshot works. Tested on Ubuntu `16.04`

martyngigg:
@quantumsteve I built ParaView 5.4.0 with the buildscript on the `v5.4.0` branch and built mantid against this. I was able to visualise an `MDHistoWorkspace` but I then loaded `SEQ_11499_md_enp.nxs` from the system test data and I now get a crash when visualising this. The console contains some errors messages:

```
ERROR: In /media/data1/source/paraview/ParaView-v5.4.0-source/VTK/Filters/Core/vtkMaskPoints.cxx, line 320
vtkMaskPoints (0x7edd180): No points to mask
```

I tried this with the current nightly and I actually get this same error but in the usual error window and Mantid does not crash, it just doesn't display anything.

quantumsteve:
@martyngigg I was able to reproduce the issue and _think_ it's an internal ParaView issue. Specifically, it looks like `vtkPVProgressHandler::GetLastMessage()` may return `NULL` and passing `NULL` to the `QString` constructor in `pqProgressManager::onMessage` is causing the segfault.

Does this match what you are seeing?

quantumsteve:
Commenting out everything in `pqProgressManager::onMessage` so that it is a noop appears to "fix" the segfault. Otherwise I seem to get some sort of infinite recursion and eventually a corrupt pointer.

martyngigg:
It looks to me as if `pqProgressManager::onMessage` calls `vtkOutputWindow::DisplayErrorText` which then calls `this->InvokeEvent(vtkCommand::MessageEvent, (void*)txt);` again that causes it to end up in `pqProgressManager::onMessage` again.

It's not entirely clear what starts this whole chain though...

martyngigg (changes requested):
I'm just marking this as not approved so that it doesn't get merged accidentally.

quantumsteve:
The last commit forces at least one box to be selected. This avoids the segfault @martyngigg found, though doesn't fix the lower-level issue.

quantumsteve:
I suspect it is related to this commit and turning on `VTK_LEGACY_REMOVE`
https://github.com/Kitware/ParaView/commit/2e16d4a2c27bd350b6337d8ef5e1a2e5c9b9bb77

quantumsteve:
Here is the `pqOutputWidget` that replaces `pqOutputWindow`. 
https://github.com/Kitware/ParaView/commit/bb4df3609eaa8dad9aa0a16310d4ffd5b90de02b

I also tried turning off `VTK_LEGACY_REMOVE` and get a segfault in `MdViewerWidget::renderWorkspace`. It might be easier to migrate to the new interface.

martyngigg:
@quantumsteve Migrating sounds like a better long term plan.

quantumsteve:
@martyngigg The last commit should be enough to match the behavior from v5.3.0. Switching to QMainWindow to take advantage of QDockWidget looks like a significant refactoring that should be done as a separate pull request.

Vates/VatesSimpleGui/ViewWidgets/inc/MantidVatesSimpleGuiViewWidgets/MdViewerWidget.h
```diff
@@ -1,6 +1,9 @@
 #ifndef MDVIEWERWIDGET_H_
 #define MDVIEWERWIDGET_H_
 
+#ifdef ERROR
+#undef ERROR
+#endif
```
quantumsteve: @martyngigg Any idea where the `ERROR` macro is coming from? Without this, windows won't build.

martyngigg:
@quantumsteve It seems to be coming from the `wingdi.h` header. It contains a `#define ERROR 0` line - 
 see https://github.com/baysao/google-glog/issues/33 for a similar error.

`wingdi.h` is included by the main `Windows.h` header which I assume is being included by Qt somewhere along the stack. The issue linked suggests using a project-wide `NOGDI` define could help. I assume this would be okay as Qt is the one doing to graphics handling and not us. I'm just compiling it on Windows to check.

martyngigg:
I tried compiling on Windows with `NOGDI` defined but it didn't help. I think we may have to live with this "workaround for now".

I can also confirm that the VSI no longer crashes when the "No points to mask" is encountered. However, now the output widget seems to be stuck where it is created. See screenshot. I can interact with it with the exception of not being able to drag the window around.

quantumsteve:
@martyngigg I tried removing `QDockWindow` and having it mostly working on OSX. The output window is shown, but is in the background. 

Please let me know if this addresses the larger issues you encountered on Ubuntu.

martyngigg:
@quantumsteve This seems to address the issue perfectly on Ubuntu. The window also appears on top of the VSI as I would expect.

quantumsteve:
@martyngigg Having the floating output window in the background on OSX is a minor issue and probably related to our use of setModal(true) on OS X. 

Is this good enough to merge?

rosswhitfield (approved):
Works on Arch Linux :+1:

martyngigg (approved):

PR #19957, opened 2017-07-04
Shorten arrays with integer lists

### LINKED ISSUE #15164: Property logging for algorithms with arrays
When algorithms properties are displayed in the log, if they contain long arrays (this has been noticed in BayesQuasi due to the requirement of fixed length FORTRAN arrays) then the whole array is printed to the log. This can have an effect on the speed of an algorithm as it takes a long time to print the whole array.
We should look into a better way of improving the way algorithm properties, such as these, are printed.

NickDraper:
Improves list handling for long lists in properties, integer separated lists such as 1,2,3 will collapse into 1-3.

**To test:**

1. Check unit and system tests pass, there are new unit tests for this.
1. Open Mantidplot
1. Set the log windows to display information logs (right click the window)
1. Run this script (it's easier in chunks)
1. Look at the values in the Workspace History GUI as well.
1. Check that generated scripts still work as expected (they may include list shortening, but that should still work).
``` python 
dataX = range(100)
dataY = range(100)
dataE = range(100)

# The workspace will be named "dataWS"
dataWS = CreateWorkspace(DataX=dataX, DataY=dataY, DataE=dataE, NSpec=1,UnitX="Wavelength")

'''Log entries like this (truncation, but not list shortening as these are double properties)
  Name: DataX, Value: 0,1,2,3,4,5,6,7,8 ... 94,95,96,97,98,99, Default?: No, Direction: Input
  Name: DataY, Value: 0,1,2,3,4,5,6,7,8 ... 94,95,96,97,98,99, Default?: No, Direction: Input
  Name: DataE, Value: 0,1,2,3,4,5,6,7,8 ... 94,95,96,97,98,99, Default?: No, Direction: Input
  '''


speclist = []
speclist += range(1,100)
CreateSampleWorkspace(OutputWorkspace='base', Function='Powder Diffraction')
MaskDetectors(Workspace='base', SpectraList=speclist)
'''Log entry like this (list shortening, no truncation)
  Name: SpectraList, Value: 1-99, Default?: No, Direction: Input
  '''

speclist = []
speclist += range(1,3)
speclist += range(4,50,2)
speclist += range(60,100)
CreateSampleWorkspace(OutputWorkspace='base', Function='Powder Diffraction')
MaskDetectors(Workspace='base', SpectraList=speclist)

'''Log entry like this (list shortening, and truncation)
  Name: SpectraList, Value: 1-2,4,6,8,10,12,1 ... 42,44,46,48,60-99, Default?: No, Direction: Input
  '''
```

**Python interface test**
``` python
speclist = []
speclist += range(1,3)
speclist += range(4,50,2)
speclist += range(60,100)
base = CreateSampleWorkspace(OutputWorkspace='base', Function='Powder Diffraction')
MaskDetectors(Workspace='base', SpectraList=speclist)


alg = base.getHistory().lastAlgorithm()
p= alg.getProperty("SpectraList")


print "List collapsing, no length limit",p.valueAsPrettyStr()
print "No list collapsing, no length limit",p.valueAsPrettyStr(0,False)
print "List collapsing, length limit 40",p.valueAsPrettyStr(40,True)
print "No list collapsing, length limit 40",p.valueAsPrettyStr(40,False)
```

closes #15164

**Release Notes** 

docs/source/release/ui.rst

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Kernel/inc/MantidKernel/Strings.h
```diff
@@ -71,6 +71,68 @@ DLLExport std::string join(ITERATOR_TYPE begin, ITERATOR_TYPE end,
...
+* Does not add the separator after the LAST item.
+*
+* For example, join a vector of strings with commas with:
+*  out = join(v.begin(), v.end(), ", ");
+*
+* @param begin :: iterator at the start
+* @param end :: iterator at the end
+* @param separator :: string to append between items.
+* @param listSeparator :: string to append between list items.
+* @return
```
soininen: So, what does this function return?

Framework/Kernel/inc/MantidKernel/Strings.h
```diff
@@ -71,6 +71,68 @@ DLLExport std::string join(ITERATOR_TYPE begin, ITERATOR_TYPE end,
...
+                                   const std::string &listSeparator = "-") {
+
+  std::stringstream result;
+
+  ITERATOR_TYPE i;
+  ITERATOR_TYPE previousValue;
+  std::string currentSeparator = separator;
+  for (i = begin; i != end;) {
+    // was this one higher than the last value
+    if (i == begin) {
```
soininen: How about just
```c++
ITERATOR_TYPE previousValue = begin;
result << *begin;
++begin;
```
before this `for` loop and be done with this check?

Framework/Kernel/inc/MantidKernel/Strings.h
```diff
@@ -71,6 +71,68 @@ DLLExport std::string join(ITERATOR_TYPE begin, ITERATOR_TYPE end,
...
+        }
+        // add the current value
+        result << currentSeparator;
+        result << *i;
+      }
+    }
+    previousValue = i;
+    i++;
+    // if we have got to the end and part of a list output the last value
+    if ((i == end) && (currentSeparator == listSeparator)) {
```
soininen: `i == end` never here, right?

Framework/Kernel/inc/MantidKernel/Strings.h
```diff
@@ -71,6 +71,68 @@ DLLExport std::string join(ITERATOR_TYPE begin, ITERATOR_TYPE end,
...
+      result << currentSeparator;
+      result << *previousValue;
+    }
+  }
+  return result.str();
+}
+/// Converts long strings into "start ... end"
+MANTID_KERNEL_DLL std::string shorten(const std::string &input,
+                                      const size_t &max_length);
```
soininen: Is the reference on `max_length` necessary?

Framework/Kernel/test/ArrayPropertyTest.h
```diff
@@ -301,6 +301,65 @@ class ArrayPropertyTest : public CxxTest::TestSuite {
                       static_cast<Property *>(0))
   }
 
+  void testPrettyPrinting() {
+    std::vector<std::string> inputList{
+        "1,2,3", "-1,0,1", "356,366,367,368,370,371,372,375", "7,6,5,6,7,8,10",
+        "1-9998, 9999, 2000, 20002-29999"};
+    std::vector<std::string> resultList{"1-3", "-1-1",
```
soininen: Does it make sense to make these vectors `const`? Similarly to other input and result lists below.

Framework/Kernel/test/ArrayPropertyTest.h
```diff
@@ -301,6 +301,65 @@ class ArrayPropertyTest : public CxxTest::TestSuite {
...
+  bool listShorteningwithType(const std::vector<std::string> &inputList,
+                              const std::vector<std::string> &resultList) {
+
+    bool success = true;
+    for (size_t i = 0; i < inputList.size(); i++) {
+      ArrayProperty<T> listProperty("i", inputList[i]);
+      std::string response = listProperty.valueAsPrettyStr(0, true);
+      TS_ASSERT_EQUALS(response, resultList[i]);
+      if (response != resultList[i]) {
+        success = false;
```
soininen: I think one could just return `false` here and be done with it. Not that performance matters in the unit tests, especially in the failure cases. Which never happen.

Framework/Kernel/test/PropertyHistoryTest.h
```diff
@@ -28,6 +28,22 @@ class PropertyHistoryTest : public CxxTest::TestSuite {
     TS_ASSERT_EQUALS(output.str(), correctOutput);
   }
 
+  void testOutputWithShortenedValue() {
+    std::string correctOutput = "Name: arg1_param, ";
+    correctOutput = correctOutput + "Value: 1234567 ... 4567890, ";
```
soininen: How about using `+=` operator here?

docs/source/release/v3.11.0/ui.rst
```diff
@@ -28,8 +28,15 @@ Plotting Improvements
...
-Scripting Window
-################
+Workspace History Window
+########################
+
+.. figure:: ../../images/WorkspaceHistoryShortenedValues.png
+   :class: screenshot
+   :align: right
+
+- Algorithm proprties with long values are now shown in a shortened format both in this display and the log.  The full hisory is still retained, and genereated python scripts will still work as before.
```
soininen: hisory -> history

Framework/Kernel/inc/MantidKernel/Property.h
```diff
@@ -141,6 +141,9 @@ class MANTID_KERNEL_DLL Property {
   }
   /// Returns the value of the property as a string
   virtual std::string value() const = 0;
+  /// Returns the value of the property as a pretty printed string
+  virtual std::string valueAsPrettyStr(size_t maxLength = 0,
+                                       bool collapseLists = true) const;
```
soininen: Could both arguments be declared `const`?

soininen (changes requested):

soininen:
While performing tests with the provided scripts, I came across two issues:
1. The changes here expect the indices to be in ascending order. How about if lists like [3, 2, 1] are given? Shouldn't these be shortened to 1-3 as well?
2. The string shortening cuts the strings sometimes middle-number which might be confusing. From the output of the second Python testing script:
```
List collapsing, length limit 40 1-2,4,6,8,10,12,1 ... 42,44,46,48,60-99
```
There is this lone `1` before the ellipsis. Might provoke confusion among the users. I don't think this is a big issue, though, so if it proves too time-consuming to fix we can drop it.

NickDraper:
Antti,
1. I though the same timing when I first implemented this, in fact version 1 I sorted the array first to get the most out of the list shortening.  However it then became apparent that for some algorithms the order of the vales entered into the property is important and meaningful, so we cannot sort them.  The way it is implemented at the moment you can take the outputted string if it is just gone through list shortening and feed it into another property and it will create an array that is an exact match.  This is important as this is the cause when people script to python and run it again.
2. This is possible, but probably not worth the effort, the list shortening and centre truncation take place at separate times, and at the point of centre truncation the input is already a string and could be any string, not necessarily a list, so the shortening by center truncation is a bit simplistic by design.

NickDraper:
I'll take a look at the code comments next week

Framework/Kernel/inc/MantidKernel/Strings.h
NickDraper: Yes it can because we advance the iterator two lines above, and the loop check has not happened yet.

However we can move this out of the loop and make it easier to read.

Framework/Kernel/test/ArrayPropertyTest.h
NickDraper: Actually here I would rather it ran through all of the test cases, the TS_ASSERT_EQUALS will output useful feedback for each test case that fails.

soininen:
One more question: how about negative numbers in lists?

NickDraper:
This will work for negative numbers just fine, one of the tests in StringsTest.h does this.  Anything that is 1 higher than the last value will be shortend.  So -1,0,1 becomes -1-1

NickDraper:
One thing it cannot handle is reverse lists so 123-10, but I'm not sure that the Property code to translate list syntax back into arrays can handle that either.  And that is pretty rare.

NickDraper:
@soininen  - ready for a re-review

soininen (approved):

soininen:
Nice to see an old issue get fixed. I run through the given test scripts and tried out a few different parameter combinations. The shortened output and the code changes look good to me.

PR #19971, opened 2017-07-06
Improving fitfunctions in python

### LINKED ISSUE #19010: Improved way of working with fit function in Python
By implementing https://github.com/mantidproject/documents/blob/master/Design/Fit%20functions%20in%20python.md 

Function evaluation and plotting are excluded from this issue.

KarlPalmen:
Description of work.

To provide python objects for fit functions rather than strings as of present as described in
https://github.com/mantidproject/documents/blob/master/Design/Fit%20functions%20in%20python.md
Substantial work has been done.

**To test:**

Check that the unit tests for FunctionWrapper and simpleapi work. Read the FunctionWrapper unit test to see what work has been done. Try it out in your own scripts.

Fixes #19010. 

**Release Notes** 

*To do.*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/PythonInterface/mantid/FunctionWrapper.py
```diff
@@ -0,0 +1,258 @@
...
+    # Untie all parameters with the given local name.
+    # Every member function must have a parameter of this name.
+       for i in range(0, self.__len__()):
+          self[i].untie(name)
+
+    def flatten (self):
+    # Return composite function, equal to self, but with 
+    # every composite function within replaced by
+    # its list of functions, so having a pure list of functions.
+    # This would enable sensible __add__ and __mult__ operations.
```
mantid-roman: Are you sure that this method will work for `Convolution`, `ProductFunction` and `MultidomainFunction` and their combinations? Are there tests for that?

Framework/PythonInterface/mantid/FunctionWrapper.py
```diff
@@ -0,0 +1,258 @@
...
+    
+    # If there are no composite functions, do nothing
+       needToFlatten = False
+       for i in range(0, self.__len__()):
+          if( not needToFlatten and isinstance(self[i],CompositeFunctionWrapper)):
+             needToFlatten = True
```
mantid-roman: There is no need to continue the loop after the first `True`

Framework/PythonInterface/test/python/mantid/FunctionWrapperTest.py
```diff
@@ -0,0 +1,376 @@
...
+        self.assertEqual(fc3_str.count("Sigma=1.3"),1)
+        
+    def test_add(self):
+        g0 = FunctionWrapper( "Gaussian", Height=7.5, Sigma=1.2, PeakCentre=10)  
+        g1 = FunctionWrapper( "Gaussian", Height=8.5, Sigma=1.25, PeakCentre=12)  
+        lb = FunctionWrapper("LinearBackground")
+        
+        c = lb + g0 + g1
+        
+        c_str = c.__str__()
```
mantid-roman: Use `str(c)` instead of `c.__str__()`.

Framework/PythonInterface/test/python/mantid/FunctionWrapperTest.py
```diff
@@ -0,0 +1,376 @@
...
+        self.assertEqual(p0_str, lb_str)
+           
+        g0_str = g0.__str__()
+        p1_str = p[1].__str__()
+        self.assertEqual(p1_str, g0_str)
+        
+        g1_str = g1.__str__()
+        p2_str = p[2].__str__()
+        self.assertEqual(p2_str, g1_str)   
+
```
mantid-roman: More tests are needed, for example:
```
a + (b + c)
a*(b + c)
a*b + c
...
```

Framework/PythonInterface/test/python/mantid/FunctionWrapperTest.py
```diff
@@ -0,0 +1,376 @@
...
+        testhelpers.assertRaisesNothing(self, ProductFunctionWrapper, g0, g1)
+
+    def test_mul(self):
+        g0 = FunctionWrapper( "Gaussian", Height=7.5, Sigma=1.2, PeakCentre=10)  
+        g1 = FunctionWrapper( "Gaussian", Height=8.5, Sigma=1.25, PeakCentre=12)  
+        lb = FunctionWrapper("LinearBackground")
+        
+        p = lb * g0 * g1
+        
+        p_str = p.__str__()
```
mantid-roman: There is no assertion that `p` is a `ProductFunction`

Framework/PythonInterface/test/python/mantid/FunctionWrapperTest.py
```diff
@@ -0,0 +1,376 @@
...
+    def test_multidomainfunction_creation(self):
+        g0 = FunctionWrapper( "Gaussian", Height=7.5, Sigma=1.2, PeakCentre=10)
+        g1 = FunctionWrapper( "Gaussian", Height=8.5, Sigma=1.2, PeakCentre=11)
+        testhelpers.assertRaisesNothing(self, MultiDomainFunctionWrapper, g0, g1)
+        m = MultiDomainFunctionWrapper( g0, g1)
+        self.assertEqual( m.nDomains(), 2)
+        
+    def test_prefinedfunction(self):
+        testhelpers.assertRaisesNothing(self, Gaussian, Height=7.5, Sigma=1.2, PeakCentre=10)
+
```
mantid-roman: Test that the wrappers produce correct results when used in `EvaluateFunction` algorithm.

Framework/PythonInterface/test/python/mantid/FunctionWrapperTest.py
```diff
@@ -0,0 +1,376 @@
...
+        g1 = FunctionWrapper( "Gaussian", Height=8.5, Sigma=1.2, PeakCentre=11)
+        testhelpers.assertRaisesNothing(self, MultiDomainFunctionWrapper, g0, g1)
+        m = MultiDomainFunctionWrapper( g0, g1)
+        self.assertEqual( m.nDomains(), 2)
+        
+    def test_prefinedfunction(self):
+        testhelpers.assertRaisesNothing(self, Gaussian, Height=7.5, Sigma=1.2, PeakCentre=10)
+       
+if __name__ == '__main__':
+    unittest.main()
```
mantid-roman: There are no tests for the generated functions, for example:

```
s = Gaussian() + FlatBackground()
```

mantid-roman (changes requested):

mantid-roman:
@KarlPalmen 
The latest version of the design document is this: https://github.com/mantidproject/documents/blob/master/Design/Fit%20functions%20in%20python.md

You probably used an outdated one.

Framework/PythonInterface/mantid/FunctionWrapper.py
```diff
@@ -0,0 +1,258 @@
...
+  def __init__ (self, name, **kwargs):
+    if not isinstance(name, str):
+       self.fun = name
+    else:
+       self.fun = FunctionFactory.createFunction(name)
+    
+       # Deal with attributes first
+#       for key in kwargs:
+#          if(self.fun.hasAttribute(key)):
+#              self.fun.storeAttributeValue(key, kwargs[key])
```
KarlPalmen: This is because you said a bug prevented attributes from working. If fixed, the code could be restored.

Framework/PythonInterface/mantid/FunctionWrapper.py
```diff
@@ -0,0 +1,258 @@
...
+      
+  def __setitem__ (self, name, value):
+      self.fun.setParameter(name, value)
+      
+  def __str__ (self):
+      return self.fun.__str__()
+                 
+  def __add__ (self, other):
+      if(isinstance(self,CompositeFunctionWrapper)):
+        sum = self
```
KarlPalmen: Because I thought it might prevent undesired modification of self, but I'm not sure of this.

Framework/PythonInterface/mantid/FunctionWrapper.py
```diff
@@ -0,0 +1,258 @@
...
+        sum = CompositeFunctionWrapper(self,other)
+        return sum.flatten()
+        
+  def __mul__ (self, other):
+      if(isinstance(self,ProductFunctionWrapper)):
+        sum = self
+        sum.fun.add(other.fun)
+        return sum.flatten()
+      else:
+        sum = ProductFunctionWrapper(self,other)
```
KarlPalmen: Name changed.

Framework/PythonInterface/mantid/FunctionWrapper.py
mantid-roman: Instance variables are pointers: changing `sum` will change `self` as well

Framework/PythonInterface/mantid/FunctionWrapper.py
```diff
@@ -0,0 +1,258 @@
...
+        return sum.flatten()
+        
+  def tie (self, *args, **kwargs):
+    for a in args:
+        if isinstance(a, dict):
+          for key in a:
+            self.fun.tie(key, str(a[key]), False)
```
KarlPalmen: Removal of the "false" argument caused an argument mismatch error. What should I do here?

mantid-roman: I'll have a look

Framework/PythonInterface/mantid/FunctionWrapper.py
```diff
@@ -0,0 +1,258 @@
...
+       return self
+       
+    def __delitem__ (self, index):
+       self.fun.__delitem__(index)
+       
+    def __len__ (self):
+       return self.fun.__len__()
+       
+    def __iter__ (self):
+       return self
```
KarlPalmen: I don't remember how and why I put it in.

KarlPalmen:
I've put updated design document at the top of this pull request and the original issue.

mantid-roman:
@KarlPalmen 
You need to merge origin/master into your branch and resolve the conflicts.

Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp
```diff
@@ -199,6 +213,23 @@ void export_IFunction() {
            "Split this function (if needed) into a list of "
            "independent functions")
 
+      .def("tie", &IFunction::tie,
+           (arg("self"), arg("parName"), arg("expr"), arg("isDefault")),
+           "Tie a parameter to an expression")
+
```
jmborr: `isDefault` is an optional argument of `IFunction::tie`. Shouldn't you use `BOOST_PYTHON_MEMBER_FUNCTION_OVERLOADS` in this case?

Framework/PythonInterface/test/python/mantid/FunctionWrapperTest.py
jmborr: The last assertion requires that string `7<Height` precedes `Sigma<2`, but this requirement is irrelevant to the test purpose.

Framework/PythonInterface/mantid/FunctionWrapper.py
KarlPalmen: I've tested this with evaluation you recently showed me and find it although it works with just addition and multiplication, it does not give the correct values for multiplication combined with addition.

KarlPalmen: I've added a test for this and as expected it fails.

mantid-roman:
@KarlPalmen The doc strings need parameter descriptions.

Framework/PythonInterface/mantid/FunctionWrapper.py
```diff
@@ -0,0 +1,392 @@
...
+
+    # ------------------------------------------------------------------------------------------------
+    wrapper_function.__name__ = name
+    # _replace_signature(fake_function, ("", ""))
+    globals()[name] = wrapper_function
+
+fnames = FunctionFactory.getFunctionNames()
+for i, val in enumerate(fnames):    
+   _create_wrapper_function(val)
+
```
mantid-roman: Some of the specifications have not been implemented.
Run the test scripts below to see what is missing.

1.
```
p = Polynomial(n=3, A0=1, A1=2, A2=4, A3=3)
print p
p.n = 4
p.A0 = 1.1
print p
print p['n']
print p.n
print p.A0
```
2.
```
g1 = Gaussian(Sigma=.5,Height=3)
g2 = Gaussian(PeakCentre=3, Sigma=.7, Height=4)
b = LinearBackground(A0=1, A1=0.2)
f = b + g1 + g2
print len(f) # must be 3
print f[0]
print f[1]

f = b * g1 * g2
print len(f) # must be 3
print f[0]
print f[1]

```
3.
```
g = Gaussian(Sigma=.5,Height=3)
b = LinearBackground(A0=1, A1=0.2)
f = b + g
print f
print f.f0
print f.f0.A0
```
4.
```
g = Gaussian(Sigma=.5,Height=3)
b = LinearBackground(A0=1, A1=0.2)
f = b + g + g

print f.LinearBackground
print f.Gaussian0
print f.Gaussian1
```
5.
```
g = Gaussian(Sigma=.5,Height=3)
b = LinearBackground(A0=1, A1=0.2)
f = b + g
f[1] = Lorentzian(PeakCentre=2)
```
6.
```
p = Polynomial(n=3, A0=1, A1=2, A2=4, A3=3)
p.tie(A0=2)
p.tie(A1='A2+A3')
p.tie({'A2': '2*A3'})
print p
f = (Gaussian() + Gaussian() + Gaussian()).flatten()
f.tie({'f1.Sigma': 'f0.Sigma', 'f2.Sigma': 'f0.Sigma'})
print f
f = (Gaussian() + Gaussian() + Gaussian()).flatten()
f.tieAll('Sigma')
print f
f.untie('f2.Sigma')
print f
f.untie('f0.Sigma')
print f
f.untie('f1.Sigma')
print f
f.tieAll('Sigma')
print f
f.untie('f1.Sigma', 'f2.Sigma')
print f
f.tieAll('Sigma')
f.untieAll()
print f
```
7.
```
p = Polynomial(n=3, A0=1, A1=2, A2=4, A3=3)
p.fix('A0')
p.fix('A0', 'A2')
print p
p.unfix('A0')
p.unfix('A0', 'A2')
print p
p.unfixAll()
print p
p.fixAll()
```
8.
```
g = Gaussian()
g.constrain('-1 < PeakCentre < 1', 'Sigma > 0')
f = (g + g + g).flatten()
f.constrainAll('Sigma>0')
f.constrainAll('Height > 1', 'Sigma > 0')
print f
```

mantid-roman (changes requested):

Framework/PythonInterface/mantid/FunctionWrapper.py
KarlPalmen: Thank you for the review.

KarlPalmen:
I've attempted to implement the dot method of accessing members such as
```
print p.A0
p.A1 = 4
```
and it has turned out to be very complicated and all it achieves is an alternative way of expressing existing code
```
print p['A0']
p['A1'] = 4
```
There we agreed not to implement it.

KarlPalmen:
Example 1. can be written as
```
p = Polynomial(n=3, A0=1, A1=2, A2=4, A3=3)
print p
p['n'] = 4
p['A0'] = 1.1
print p
print p['n']
print p['A0']
```
and example 3 can be written as
```
g = Gaussian(Sigma=.5,Height=3)
b = LinearBackground(A0=1, A1=0.2)
f = b + g
print f
print f[0]
print f[0]['A0']
```

KarlPalmen:
Example 7 can be written
```
p = Polynomial(n=3, A0=1, A1=2, A2=4, A3=3)
p.fix('A0')
p.fix('A2')
print p
p.untie('A0')
p.untie('A2')
print p
p.untieAllParameters()
print p
p.fixAllParameters()
```

KarlPalmen:
Example 8 can be rewritten
```
g = Gaussian()
g.constrain('-1 < PeakCentre < 1')
g.constrain('Sigma > 0')
f = (g + g + g).flatten()
f.constrainAll('Sigma>0')
f.constrainAll('Height > 1')
f.constrainAll('Sigma > 0')
print f
```

KarlPalmen:
After most recent change, example 6 can be written
```
p = Polynomial(n=3, A0=1, A1=2, A2=4, A3=3)
p.tie(A0=2)
p.tie(A1='A2+A3')
p.tie({'A2': '2*A3'})
print p

f = (Gaussian() + Gaussian() + Gaussian()).flatten()
f.tie({'f1.Sigma': 'f0.Sigma', 'f2.Sigma': 'f0.Sigma'})
print f
f = (Gaussian() + Gaussian() + Gaussian()).flatten()
f.tieAll('Sigma')
print f
f.untie('f2.Sigma')
print f
f.untie('f0.Sigma')
print f
f.untie('f1.Sigma')
print f
f.tieAll('Sigma')
print f
f.untie('f1.Sigma')
f.untie('f2.Sigma')
print f
f.tieAll('Sigma')
f.untieAllParameters()
print f
```

mantid-roman:
@KarlPalmen `f = b + g1 + g2` must produce a flat composite function with 3 elements.

KarlPalmen:
Thank you for making this clear.

KarlPalmen:
Outstanding issues in examples 4 & 5.

KarlPalmen:
Example 4 works if rewritten thus:
```
g = Gaussian(Sigma=.5,Height=3)
b = LinearBackground(A0=1, A1=0.2)
f = b + g + g

print f.f("LinearBackground")
print f.f("Gaussian 0")
print f.f("Gaussian 1")
```

jmborr:
@KarlPalmen I don't know how (my screw up?) but one of the 14 changed files in this pull request is `BASIS_Definition_20140101-.xml`. Can you remove these changes?

Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp
```diff
@@ -35,6 +36,32 @@ PyObject *getCategories(IFunction &self) {
...
+  void operator()(void *) {}
+  /// Does nothing
+  void operator()(const void *) {}
+};
+
+//------------------------------------------------------------------------------------------------------
+/**
+* Something that makes IFunction return to python a composite function
+* for Product function, Convolution or
+* any similar superclass of composite function.
```
jmborr: you mean subclass, right?

Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp
```diff
@@ -35,6 +36,32 @@ PyObject *getCategories(IFunction &self) {
...
+* @param self :: Enables it to be called as a member function on the
+* IFunction class
+*/
+Mantid::API::CompositeFunction_sptr castToCompositeFunction(IFunction &self) {
```
jmborr: Can you please explain why do we must return a shared pointer as opposed to a raw pointer or a reference?

mantid-roman: Good point. Maybe it should be a pointer

jmborr: My feeble understanding is that a raw pointer will not prevent the python garbage collector to delete the pointee once the associated python object reference count reaches zero. If we use a raw pointer, will `self.fun` in Line 253 of `FunctionWrapper.py` be deleted once variable `comp` goes out of scope?

mantid-roman: I am not sure what happens in that case but probably we have a problem

jmborr:
Regarding use of the dot operator, why not overload `__getattr__` and `__setattr__` to directly access attributes of `FunctionWrapper.fun` ? The required source in `FunctionWrapper.py`are:
```
  def __getattr__(self, item):
      if hasattr(self.fun, item):
          return getattr(self.fun, item)
      else:
          return self.fun[item]

  def __setattr__(self, key, value):
      if hasattr(self.fun, key):
          setattr(self.fun, key, value)
      else:
          self[key] = value
```
And also in `FunctionWrapper._init__`, substitute:
```
    if not isinstance(name, str):
       self.fun = name  # avoids recursive call because of __setattr__
    else:
       self.fun = FunctionFactory.createFunction(name)
```
for
```
    if not isinstance(name, str):
       self.__dict__['fun'] = name  # avoids recursive call because of __setattr__
    else:
       self.__dict__['fun'] = FunctionFactory.createFunction(name)
```
As a result, the last two lines in this script become equivalent:
```
from mantid import simpleapi as msapi
g0 = msapi.FunctionWrapper( "Gaussian", Height=7.5, Sigma=1.2, PeakCentre=10)
g0.fun['Height'] = 8.0
g0.Height = 8.0
```
For a user, the last line is much easier to understand.

Framework/PythonInterface/mantid/FunctionWrapper.py
```diff
@@ -0,0 +1,480 @@
+from mantid.simpleapi import FunctionFactory
+
+class FunctionWrapper:
```
jmborr: Should be a New Style class: `class FuncionWrapper(object):`

Framework/PythonInterface/mantid/FunctionWrapper.py
```diff
@@ -0,0 +1,480 @@
...
+      self.fun.removeConstraint(name)
+      
+  def getParameterName(self, index):
+      """ Get the name of the parameter of given index
+      
+          :param index: index of parameter
+      """
+      return self.fun.getParamName(index)
+      
+  def getFunction(self):
```
jmborr: This method does not  make sense because one can access the underlying function by accessing attribute `fun` directly from the `FunctionWrapper` instance. If you intent is to protect attribute `fun` from user tampering, then I suggest to change attribute `fun` to `_fun` to mark it as protected and then write only a getter property:
```
@property
def function(self):
    return self._fun
```
The code below shows that assignment to FunctionWrapper.function is unsuccessful:
```
g = msapi.FunctionWrapper( "Gaussian", Height=7.5, Sigma=1.2, PeakCentre=10)
print(g.function)
g2 = msapi.FunctionWrapper( "Gaussian", Height=8.5, Sigma=1.2, PeakCentre=11)
g.function = g2.fun  # unsuccessful assignment 
print(g.function)
g.fun = g2.fun  # successful assignment
print(g.function)
```
Results in:
```
name=Gaussian,Height=7.5,PeakCentre=10,Sigma=1.2
name=Gaussian,Height=7.5,PeakCentre=10,Sigma=1.2
name=Gaussian,Height=8.5,PeakCentre=11,Sigma=1.2
```

Framework/PythonInterface/mantid/FunctionWrapper.py
```diff
@@ -0,0 +1,480 @@
...
+       if 'Global' in kwargs:
+          list = kwargs['Global']
+          for name in list:
+             self.tieAll(name)
+       
+       # Set domain indices: 1 to 1       
+       for i in range(0, len(self)):
+          self.fun.setDomainIndex(i, i)
+          
+    def nDomains (self):
```
jmborr: This should be made a `property`

Framework/PythonInterface/mantid/FunctionWrapper.py
```diff
@@ -0,0 +1,480 @@
...
+     
+        
+def _create_wrapper_function(name):
+    """Create fake functions for the given name
+       It should not be called directly
+                  
+       :param name: name of fake function
+    """
+    # ------------------------------------------------------------------------------------------------
+    def wrapper_function(*args, **kwargs):
```
jmborr: I would substitute the `if` chain with a registry:
```
name_to_constructor = {
    'CompositeFunction': CompositeFunctionWrapper,
    'ProductFunction': ProductFunctionWrapper,
    'Convolution': ConvolutionWrapper,
    'MultiDomainFunction': MultiDomainFunctionWrapper,
    }
    # constuctor is FunctionWrapper if the name is not in the registry
    constructor = name_to_constructor.get(name, FunctionWrapper)  
    return  constructor(name, *args, **kwargs)
````
In fact, I would place `name_to_constructor` at the scope of the module and then have each derived class of `FunctionWrapper` subscribe to this registry .

**Comment by jmborr** [comment]
I don't know if there's a naming convention for the mantid python API, but I think now would be the best time for the interface to follow the standard [PEP8](https://www.python.org/dev/peps/pep-0008/) style guide.

File: Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp (hunk 15 above)
Comment (KarlPalmen) [inline Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp]: I'm not sure, but that would explain why the code is duplicated.

**Comment by KarlPalmen** [comment]
@jmborr  The addition of `` BASIS_Definition_20140101-.xml `` was to restore a file lost from the previous merge from master, because it was too large for a normal commit. Therefore this change should be regarded as part of that merge from master.

**Comment by KarlPalmen** [comment]
@jmborr I've attempted to implement ``comp.name`` as you suggested, exposing ``IFunction.hasParameter`` and find that the unit test fails, because it cannot find ``__coerce__`` and ``__iter__``  parameters, when attempting ``item = comp[nameorindex]``.

File: Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp (hunk 16 above)
Comment (mantid-roman) [inline Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp]: @KarlPalmen I think now that `castToCompositeFunction` is a dangerous method. Could you please have a think if it can be removed and instead `IFunction` would have 'composite' methods such as `__len__()`, `getFunction` which return 0 or throw exceptions if the function is not a `CompositeFunction`?

Comment (KarlPalmen) [inline Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp]: @mantid-roman I thought about what ``__len__()`` should return for a plain FunctionWrapper and found that the following script works:
```
g = Gaussian(Sigma=.5,Height=3)
print g
print g[0], g[1], g[2]
g[1] = 7
print g
```
and therefore ``__len__()`` should return the number of parameters and this is very easy to implement.  I'm still to think of other 'composite' methods.

**Comment by jmborr** [comment]
@KarlPalmen Can you be more specific? What is `comp`? Which unit test is failing?

File: Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp (hunk 16 above)
Comment (KarlPalmen) [inline Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp]: I found 13 methods in ``CompositeFunctionWrapper`` not in ``FunctionWrapper`` besides the ``__len__()`` already mentioned. Each would need considering. Some such as ``tieAll``  would do nothing. Others such as ``fixAll`` would just ``fix".

**Comment by KarlPalmen** [comment]
@jmborr  It fails in numerous tests and ``comp`` is a CompositeFunctionWrapper object as in the code for such. One example, is the ``+``  fails for want of ``__coerce__``.

File: Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp (hunk 16 above)
Comment (jmborr) [inline Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp]: One remark: never a good idea for `len()` to mean different concepts for subclasses of CompositeFunctionWrapper. Otherwise you will break code using `len(some_object)` that is agnostic of the type of `some_object`. If `len()` is to return the number of parameters for `FunctionWrapper`, it should return the number of parameters for any derived class such as `CompositeFunctionWrapper`.

Comment (mantid-roman) [inline Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp]: I agree with @jmborr .

Comment (KarlPalmen) [inline Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp]: But we would have the `.` operator meaning different things for a subclass of ``FunctionWrapper``, ``CompositeFunctionWrapper`` and may be that's why it's not working.

Comment (KarlPalmen) [inline Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp]: Nevertheless, if we don't support the numeric index functionality that happens to occur for ``FunctionWrapper,`` we can have ``__len__()`` return 0 or an exception.

Comment (jmborr) [inline Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp]: @KarlPalmen If `__getattr__` and `__setattr__` are defined only in parent `FunctionWrapper`, how can  `.` have a different meaning in subclass `CompositeFunctionWrapper`? Can you provide an example?

Comment (KarlPalmen) [inline Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp]: @jmborr Look at scripts 1, 3 & 4 provided by @mantid-roman above and one can see different functionality for the dot operator. However the functionality for ``CompositeFunctionWrapper`` could be seen as an extension of the functionality of ``FunctionWrapper``.

**Comment by jmborr** [comment]
@KarlPalmen unit tests in `FunctionWrapperTest` run OK in my local version of this branch where I defined `__getattr__`, `__setattr__`, and properties `function` and `name`. I have uploaded my modified `FunctionWrapper` to [dropbox](https://www.dropbox.com/s/y4raruzgr2vcaij/FunctionWrapper.py?dl=0). Also the [IFunction exports](https://www.dropbox.com/s/naa81qni8s1esuo/IFunction.cpp?dl=0).

File: Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp (hunk 16 above)
Comment (jmborr) [inline Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp]: I see what you mean. For a `CompositeFunctionWrapper` object, accessing its component functions and their associated parameters through operator `.` is not possible since most of the identifier strings (e.g. "Gaussian 0", "f0.f1", or "f0.f1.Height) are not valid python identifiers. Then, the only pythonic access interface is through `__getitem__` and `__settitem__`. I guess the choice is to either use  `.` for parameters and attributes owned by the function, or dispose of this functionality.

Comment (jmborr) [inline Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp]: For composite functions, I would access the component functions _only_ by their composite index. For instance:
```
c1 = f1 + f2
c2 = f3 + f4
cc = c1 + c2
```
then `cc[1][0] == f3`. I think the adoption of prefixes like "f" or numerical suffixes like in "Gaussian 0" are somewhat arbitrary. The composite index [1][0] is a more natural representation of the hierarchical nature of the composite function. This even allows for fancy slicing. For non-composite functions, `f[0]` would return itself. In this convention, `__len__` would make sense as the number of functions, not the number of parameters.
Also, I would not use numerical indexes to access fitting parameter. I think they should be accessed only by their names.

File: Framework/PythonInterface/mantid/FunctionWrapper.py (hunk 18 above)
Comment (KarlPalmen) [inline Framework/PythonInterface/mantid/FunctionWrapper.py]: This was a method a I added in early days, not sure whether I'd eventually need it. I may remove it.

Thanks for letting me know about ``_fun``. There is no such method ``function``. I think you meant ``getFunction`` and if so, the ``get`` would explain why the assignment did not work.

**Comment by KarlPalmen** [comment]
@jmborr  I've tried your code and it works for parameters of non-composite functions. I've added a unit test for this and committed. I also notice that the indentation has been tidied. Thanks.

File: Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp (hunk 16 above)
Comment (KarlPalmen) [inline Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp]: Also strings such as "f1.Sigma" are also not valid python identifiers. 

I've now got the dot operator working for non-composite functions, thanks to @jmborr .

Comment (KarlPalmen) [inline Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp]: Going back to the original purpose of this thread, @mantid-roman suggested the cast function, that we'd like to remove, to solve a serious problem with the ``*`` operator for composite functions.
We want
```
p = f1 * f2 *f3 
p = ProductFunctionWrapper( f1, f2, f3)
```
to have the same effect and in particular the component functions would be referred to by
`` p[0], p[1], p[2]``.
This does not occur from repeated use of the ``*`` operator and so the result needs every composite function within it replaced by its list of functions (which I refer to flattening).
When doing this within a ProductFunction, a composite function within it is presented as a IFunction. 

I think if we were to remove the cast and implement integer indexing and length as suggested, we'd at best get only the first function in the composite function and lose the others from the product.

File: Framework/PythonInterface/mantid/FunctionWrapper.py (hunk 17 above)
Comment (KarlPalmen) [inline Framework/PythonInterface/mantid/FunctionWrapper.py]: I see ``class FunctionWrapper(object)`` in the code your provided for the dot operator. May be that's why it works. I'm new to new style classes. https://www.python.org/doc/newstyle/

File: Framework/PythonInterface/mantid/FunctionWrapper.py (hunk 19 above)
Comment (KarlPalmen) [inline Framework/PythonInterface/mantid/FunctionWrapper.py]: I'm not yet familiar with ``property`` in Python.

Comment (jmborr) [inline Framework/PythonInterface/mantid/FunctionWrapper.py]: @KarlPalment I meant to use the `@property` decorator so one can type `f.nDomains` (Python-style) instead of `f.nDomains()` (C-style)

File: Framework/PythonInterface/mantid/FunctionWrapper.py (hunk 18 above)
Comment (jmborr) [inline Framework/PythonInterface/mantid/FunctionWrapper.py]: @KarlPalmen `function` is the property I wrote above in this thread in order to replace method `getFunction()`. Getters and setters in a Python API are replaced with [property methods](https://stackoverflow.com/questions/17330160/how-does-the-property-decorator-work) because it reads more pythonic:

File: Framework/PythonInterface/mantid/FunctionWrapper.py (hunk 19 above)
Comment (KarlPalmen) [inline Framework/PythonInterface/mantid/FunctionWrapper.py]: I see. 

I'm finding http://www.python-course.eu/python3_properties.php a good introduction to properties.

File: Framework/PythonInterface/mantid/FunctionWrapper.py (hunk 20 above)
Comment (KarlPalmen) [inline Framework/PythonInterface/mantid/FunctionWrapper.py]: I've tried this and it works, but I don't understand what you mean by `registry` here and hence `subscribe to this registry` here.

Comment (jmborr) [inline Framework/PythonInterface/mantid/FunctionWrapper.py]: @KarlPalmen registry was a fancy way to say dictionary. Subscription means entering a pair `key:value` in the dictionary whenever a new wrapper class is coded. An automatic way of doing this would be desirable, something like the `DECLARE_FUNCTION` macro.

Comment (KarlPalmen) [inline Framework/PythonInterface/mantid/FunctionWrapper.py]: Thank you @jmborr  for explaining. I'd leave it till the next new wrapper class is encoded.

File: Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp (hunk 16 above)
Comment (jmborr) [inline Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp]: @KarlPalmen I think the issue is to enforce the associative property of the product operator:
```
( f * g ) * h == f * (g * h) == f * g * h
```
To me the most appropriate venue is to enforce this in the `ProductFunction` C++ class. It  requires overriding `addFunction()` and `replaceFunction()` of parent class `CompositeFunction`. For instance, if we are adding a `ProductFunction`, then we loop `addFunction()` over its components functions. 
Once this is done, writing any of the three previous expressions will result in the same object.

Comment (KarlPalmen) [inline Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp]: Yes that is a possible solution, but one must make sure that none of ``f, g or h`` is a composite function other than ProductFunction, as such a situation has both addition and multiplication and so is no longer associative. In such a case, the current code makes no attempt to flatten and so the function cast is not an issue there.

Comment (KarlPalmen) [inline Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp]: The ``+`` operator has no such issue, because unlike ProductFunction, CompositeFunction receives its component composite functions as CompositeFunction. It is this peculiarity of the python binding that is the issue.

Comment (KarlPalmen) [inline Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp]: I'm been working on this issue today with the help of @martyngigg , trying to find a way of doing without the ``castToCompositeFunction`` function.

**Comment by KarlPalmen** [comment]
I've spent some time with the help of @martyngigg attempting to remove the need for the ``castToComposite`` function and eventually, Martyn decided that ``ProductFunction``, ``Convolution`` need properly exporting to Python as ``CompositeFunction`` is (hence no problem with addition). He has thought of a way of doing this and will produce an issue for this.

File: Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp (hunk 16 above)
Comment (KarlPalmen) [inline Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp]: Issue #20279 has been created by @martyngigg  to enable ``castToComposite`` to be removed. It does so by tackling the root cause of the problem arising from the fact that ``ProductFunction`` etc. are not really exported to python.

**Comment by KarlPalmen** [comment]
I think all issues in @mantid-roman review have been addressed including the cast to composite.

File: Framework/PythonInterface/mantid/FunctionWrapper.py [hunk 21]
```diff
@@ -0,0 +1,519 @@
...
+             flatSelf.fun.add(self[i].fun)
+              
+       return flatSelf          
+  
+class ProductFunctionWrapper(CompositeFunctionWrapper):
+    """ Wrapper class for Product Fitting Function
+    """
+    def __init__ (self, *args):
+       """ Called when creating an instance
+           It should not be called directly.
```
mantid-roman: Please document the arguments.

Framework/PythonInterface/mantid/FunctionWrapper.py
```diff
@@ -0,0 +1,519 @@
...
+       """
+       self.pureAddition = False
+       self.pureMultiplication = True
+       return self.initByName("ProductFunction", *args)
+        
+class ConvolutionWrapper(CompositeFunctionWrapper):
+    """ Wrapper class for Convolution Fitting Function
+    """
+    def __init__ (self, *args):
+       """ Called when creating an instance
```
mantid-roman: Please document the arguments.

Framework/PythonInterface/mantid/FunctionWrapper.py
```diff
@@ -0,0 +1,519 @@
...
+       self.pureAddition = False
+       self.pureMultiplication = False
+       return self.initByName("Convolution", *args)
+        
+class MultiDomainFunctionWrapper(CompositeFunctionWrapper):
+    """ Wrapper class for Product Fitting Function
+    """
+    def __init__ (self, *args, **kwargs):
+       """ Called when creating an instance
+           It should not be called directly
```
mantid-roman: Please document the arguments.

Framework/PythonInterface/mantid/FunctionWrapper.py
```diff
@@ -0,0 +1,519 @@
...
+       """
+       if len(args) == 1 and  not isinstance(args[0], FunctionWrapper):
+          # We have a composite function to wrap
+          self.fun = args[0]
+       else:
+          if name == "CompositeFunction":
+             the_function = FunctionFactory.createFunction(name)
+             self.fun = the_function
+          else:
+             self.fun = FunctionFactory.createCompositeFunction(name)
```
mantid-roman: Is this still necessary after exporting classes  from CurveFitting to python?

Framework/PythonInterface/mantid/FunctionWrapper.py
```diff
@@ -0,0 +1,519 @@
...
+             self.tieAll(name)
+        
+       # Set domain indices: 1 to 1       
+       for i in range(0, len(self)):
+          self.fun.setDomainIndex(i, i)
+     
+    @property     
+    def nDomains (self):
+       """ Return number of domains 
+           Used for testing
```
mantid-roman: It is not just for testing.

Framework/PythonInterface/mantid/api/src/Exports/MultiDomainFunction.cpp
```diff
@@ -0,0 +1,26 @@
...
+      .def("getFunction", &MultiDomainFunction::getFunction,
+           (arg("self"), arg("i")), "Get the i-th function.")
+      .def("__getitem__", &MultiDomainFunction::getFunction,
+           (arg("self"), arg("i")), "Get the i-th function.")
+      .def("add", &MultiDomainFunction::addFunction,
+           (arg("self"), arg("function")), "Add a member function.")
+      .def("setDomainIndex", &MultiDomainFunction::setDomainIndex,
+           (arg("self"), arg("funIndex"), arg("domainIndex")),
+           "Associate a function and a domain.")
+      .def("nDomains", &MultiDomainFunction::getNumberDomains, arg("self"),
```
mantid-roman: This method should be exported in the `IFunction` class.

docs/source/release/v3.11.0/framework.rst
```diff
@@ -102,6 +102,7 @@ Python Fit Functions
 ####################
 
 - A bug that makes it difficult to define and use attributes in python fit functions has been fixed.
+- The usability of the fit functions has been improved.
```
mantid-roman: I think we need to be more specific here.

mantid-roman (changes requested):

mantid-roman:
@KarlPalmen We need an rst user doc file describing how to use the new functionality.

martyngigg:
I have one of my usual pedantic comments on the `FunctionWrapper.py` file - its casing is inconsistent with everything else. Should we just call the file `fitfunctions.py`?

docs/source/concepts/FitFunctionsInPython.rst
```diff
@@ -0,0 +1,170 @@
...
+=======================
+
+Introduction
+------------
+
+Mantid enables Fit function objects to be produced in python. For example
+
+.. code:: python
+
+  g = gaussian()
```
mantid-roman: It should be `Gaussian` with capital `G`

mantid-roman (approved):

Framework/PythonInterface/mantid/fitfunctions.py
```diff
@@ -0,0 +1,524 @@
+from mantid.simpleapi import FunctionFactory
+from pdb import set_trace as tr
```
martyngigg: This line looks like it was left over from debugging?

Framework/PythonInterface/mantid/fitfunctions.py
```diff
@@ -0,0 +1,524 @@
+from mantid.simpleapi import FunctionFactory
```
martyngigg: As the `simpleapi` imports this and this then imports the `simpleapi` it would be better if this line just read `from mantid.api import FunctionFactory`

martyngigg (changes requested):
A couple of tidy ups that are worth doing I think.

Also the BASIS IDF seems to have been changed here? I don't think that was intended?

KarlPalmen:
@martyngigg  The change to the BASIS IDF was deletion because to was too big to allow a merge to be committed and then its restoration. You were helping me when it was restored and committed.

martyngigg:
@KarlPalmen I wouldn't expect to see it in the list of changes though as it should be the same as master.

Can you open a git bash prompt in your code directory and type:

```
git checkout origin/master -- instrument/BASIS_Definition_20140101-.xml 
```

you then may need to run

```
git add instrument/BASIS_Definition_20140101-.xml
git commit --no-verify -m"Revert changes to BASIS definition Refs #19010"
```

docs/source/release/v3.11.0/framework.rst
```diff
@@ -102,6 +102,7 @@ Python Fit Functions
 ####################
 
 - A bug that makes it difficult to define and use attributes in python fit functions has been fixed.
+- The usability of the fit functions has been improved, enabling users to construct and modify the functions as objects rather than strings.
```
martyngigg: I think we should link to the new documentation here or users won't know how to get more detail on this.

KarlPalmen: How does one do this?
The new documentation is in the concepts folder.

mantid-roman: You can create a link to the concept page like this: 
```
:ref:`Some text <FitFunctionsInPython>`
```

KarlPalmen: Thank you @mantid-roman . It works.

martyngigg (approved):

PR #19081, opened 2017-03-07
Set MantidPlot DPI awareness to unaware in Windows

martyngigg:
Forces `dpiAware=False` within MantidPlot by using an explicit manifest file.

Setting this flag ensures that the system will scale MantidPlot icons and bitmaps by the amount selected by the user in the system preferences. 

**To test:**

This needs to be tested on Windows with a machine connected to a high-resolution display, e.g. 4K.

The icons in the MantidPlot toolbar should be very small prior to this change and after applying this fix and rebuilding then you should see the icons at a suitable size.

Another check is to use the Process Explorer tool to verify that the MantidPlot executable is DPI unaware as described [here](https://technet.microsoft.com/en-gb/library/dn528847.aspx#Anchor_0) 

Fixes #19078 

[Release notes](/mantidproject/mantid/commit/42aad0f74fd207930418ba4001fe4de96f31c1ab) have been updated.

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)? Is it well structured with small focussed classes/methods/functions?
- [ ] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant documentation been added/updated?
- [ ] Is user-facing documentation written in a user-friendly manner?
- [ ] Has developer documentation been updated if required?

- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

PR #19974, opened 2017-07-06
OS X support for high DPI screens

gvardany:
MantidPlot showed up blur on 4K screens with OS X. This fixes the scaling.

**To test:**
You have to be on 4K screen with OS X.

Install OS X packages built with and without this change. 
Open MantidPlot and compare the graphics and especially the texts.
Optionally go through several interfaces/viewers and see if there is nothing dramatically off.

No issue.

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/docs/source/release/v3.11.0/ui.rst
+++ b/docs/source/release/v3.11.0/ui.rst
@@ -14,6 +14,8 @@ Windows
 OS X
 ####
 
+- MantidPlot will now support smart scaling on high DPI (4K) screens on OS X.
+
 User Interface
 --------------
 
--- a/installers/MacInstaller/Info.plist.in
+++ b/installers/MacInstaller/Info.plist.in
@@ -32,5 +32,9 @@
 	<true/>
 	<key>NSHumanReadableCopyright</key>
 	<string>Copyright 2011 ISIS Rutherford Appleton Laboratory and NScD Oak Ridge National Laboratory</string>
+	<key>NSPrincipalClass</key>
+    <string>NSApplication</string>
+    <key>NSHighResolutionCapable</key>
+    <string>True</string>
 </dict>
 </plist>
--- a/installers/MacInstaller/Info.plist.in
+++ b/installers/MacInstaller/Info.plist.in
@@ -33,8 +33,8 @@
 	<key>NSHumanReadableCopyright</key>
 	<string>Copyright 2011 ISIS Rutherford Appleton Laboratory and NScD Oak Ridge National Laboratory</string>
 	<key>NSPrincipalClass</key>
-    <string>NSApplication</string>
-    <key>NSHighResolutionCapable</key>
-    <string>True</string>
+	<string>NSApplication</string>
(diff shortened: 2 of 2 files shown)


gvardany:
@quantumsteve Did you manage to have a look into this?

quantumsteve (changes requested):
Last year @martyngigg and I looked into enabling this on OS X and Windows. Just enabling the flags generates very small icons in menus (see below). [Qt 5](http://doc.qt.io/qt-5/highdpi.html) has more options for High DPI displays, but modifying MantidPlot is a much larger task than shown here. 
image

I'd recommend we instead enable high DPI display support in the new workbench on day 1 and focus our efforts there.

gvardany:
True that the icons are tiny, but I think especially mac users would generally prefer small and sharp over large and fuzzy. Moreover, without the flag, also the texts and the plots are fuzzy, which are lot more important to get sharp. So somehow I thought getting just the icons small is a good compromise, and more preferable until we do it properly. I would agree to put this off for the moment but aiming to enable it for the new workbench from the very beginning should be considered.

quantumsteve:
@martyngigg pointed out that Mantid is explicitly set to be DPI unaware on Windows (see https://github.com/mantidproject/mantid/pull/19081), which is identical to the current behavior on MacOS. I'm sympathetic to your desire for Retina support, but this half-solution will disappoint many users, such as those with poorer eye sight that need larger icons. Doing so properly requires at least Qt5, new icons, and more exhaustive testing and refinement. Let's leave Retina support as a benefit of the new workbench.

Does this PR change the *saved* figures or just the MantidPlot window? 

The `Info.plist` may be edited after Mantid is installed and I don't see harm if you want to document this change for motivated users.

gvardany:
I have probably missed #19081. I am not actually sure about the saved figures, I have not tested it.
If it is possible to patch the file after installation, I agree that could be a good compromise.
So where do you suggest to document it? in wiki?

quantumsteve:
The wiki is fine, though not terribly visible and hard to keep updated. I think you'd be better off putting it in the sphinx documentation and adding a link in the release notes.

gvardany:
I just tried to amend the .plist after installation and it did not work. This should be done before installation, but the issue there is that the .dmg is mounted read-only, so it did not allow me to modify the package content before installation. Any idea one could make and checkbox in the installation step to ask for the preference or something like that?

gvardany:
Also it is as blurry in the saved plot, as it is inside MantidPlot.

quantumsteve:
I'd suggest doublechecking your changes to the Info.plist. I tested this PR by updating it after installing nightly and made the figure above.

I haven't tried making an install option like you describe above and suspect it won't be trivial. Given the small number of people we expect will make this change, I'd recommend focusing your efforts elsewhere.

gvardany:
You are absolutely right, this problem affects probably <1% of the users. I see what I can do for my client setup, and we can put this off until it is more relevant.

PR #19978, opened 2017-07-07
Inconsistent use of number density

### LINKED ISSUE #19962: Inconsistent use of number density
In a number of places within Mantid number density is used inconsistently, between number of atoms per cubic Angstrom and number of formula units (molecules) per cubic Angstrom. This issue proposes to standardise everywhere to atoms per cubic Angstrom.

Changes required:
 * The documentation for the parameter `SampleNumberDensity` needs correcting in `SetSampleMaterial`.
* `SetSampleMaterial` needs a comment to clarifiy the two different `rho` calculations https://github.com/mantidproject/mantid/pull/12758/files#diff-7e6895aa1590493fe92220f55c3d630bR305.
 * `Material.cpp` needs to be clear in the doxygen that it is atoms per cubic Angstrom.
 * `MaterialBuilder` needs to set the atomic number density in `Material`. Doxygen should be fixed so that everything refers to atoms per cubic Angstrom.
 * Tests that will break:
   * `KernelTest.MaterialBuilderTest.test_Number_Density_Set_By_Formula_MassDensity`
   * `KernelTest.MaterialXMLParserTest.test_MassDensity_Attribute`
   * `CalculateSampleTransmissionTest.CalculateSampleTransmissionTest.test_mass_density`
   * `IndirectTransmissionTest.IndirectTransmissionTest.test_mass_density`
   * `SampleTest.SampleTest.test_material`

Related - the `Zparameter` is sometimes inconsistently defined between atoms in a unit cell and formula units in a unit cell. This should be standardised to formula units per unit cell.
 * `ZParameter` doxygen in `MaterialBuilder`

We need to look for anywhere else these issues might be lurking too.

ianbush:
The main change is to ensure `MaterialBuilder` uses atoms per cubic Angstrom for the number density, instead of formula units per cubic Angstrom.

There are some documentation changes to ensure that it is clear that `atoms / Angstrom^3` are used for number density and not anything else.

**To test:**

Check unit tests pass. Is there anywhere else the documentation should be clarified? Is there any knock-on effect on user scripts we should worry about, or is this just a bug fix?

Fixes #19962

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


AndreiSavici:
Did we decide atoms/volume or formula units/volume? I don't recall a final answer on this

ianbush:
I went for atoms / volume as it was the original intention in `Material`. Switching to formula units / volume I think would be a bigger breaking change.

Framework/Kernel/src/MaterialBuilder.cpp
```diff
@@ -257,11 +257,13 @@ double MaterialBuilder::getOrCalculateRho(
   } else if (m_zParam && m_cellVol) {
     return totalNumAtoms * m_zParam.get() / m_cellVol.get();
   } else if (m_massDensity) {
+    // g / cc -> atoms / Angstrom^3
     double rmm = 0.;
     for (const auto &formulaUnit : formula) {
       rmm += formulaUnit.atom->mass * formulaUnit.multiplicity;
     }
-    return (m_massDensity.get() / rmm) * PhysicalConstants::N_A * 1e-24;
+    return (m_massDensity.get() * totalNumAtoms / rmm) *
```
martyngigg: :+1:

Framework/PythonInterface/test/python/plugins/algorithms/IndirectTransmissionTest.py
```diff
@@ -123,12 +123,12 @@ def test_mass_density(self):
         ws = IndirectTransmission(Instrument=instrument, Analyser=analyser, Reflection=reflection,
                                   ChemicalFormula=formula, DensityType='Mass Density', Density=density, Thickness=thickness)
 
-        ref_result = [6.65800, 0.82122, 2.58186, 53.50693, 56.08880, 0.03342, 0.1, 0.82676, 0.17096]
+        ref_result = [6.65800, 0.82122, 2.58186, 53.50693, 56.08880, 0.100272, 0.1, 0.56512, 0.43021]
         mass_values = ws.column(1)
         np.testing.assert_array_almost_equal(mass_values, ref_result, decimal=4)
 
         # mass density 1.0 = number density 0.033424 for water
```
martyngigg: The comment and code here are now inconsistent. I would suggest getting rid of the comment, it doesn't seem that useful.

martyngigg:
Code-wise I think this is fine now. `MassDensity` was introduced much more recently so we may be saved that most scripts are still using `NumberDensity` as input and shouldn't need to change.

On the documentation side is it worth updating `SetSampleMaterial` with an explicit calculation of how the number density is derived from the mass density?

ianbush:
@martyngigg A good point, I have added the mass density calculation to `SetSampleMaterial`.

SpencerHowells:
Let's get this straight! I have been calculating absorption for 45 years! The equation contains the product of number density & cross-section so these 2 variables MUST be consistent. In the amorphous/liquid community we have used number density in atoms/volume so the cross-sections are average per atom. I don't mind which we use provided SetSampleMatrial is consistent.

SpencerHowells:
Regarding the Indirect routines I decided to give users the option of inputting Mass Density so that they don't need to calculate the number density and risk having it in the wrong units. This has happened often in the past.

ianbush:
I have pushed some extra changes to hopefully improve the validation in `MaterialBuilder`, meaning a number density should always be set, or an error thrown.

ianbush:
@martyngigg @peterfpeterson This should be ready for review now.

Framework/PythonInterface/plugins/algorithms/CylinderPaalmanPingsCorrection2.py
```diff
@@ -405,10 +405,10 @@ def _set_material(self, ws_name, chemical_formula, density_type, density):
             mat = builder.setFormula(chemical_formula).setMassDensity(density).build()
```
martyngigg: I think we can simplify this a bit now can't we? The `builder` variable is no longer needed as the number density will be calculated on the fly? So basically the `if` just needs 1 line each for whether the property is `SampleMassDensity` or `SampleNumberDensity`?

Framework/PythonInterface/plugins/algorithms/IndirectTransmission.py
```diff
@@ -114,7 +114,7 @@ def PyExec(self):
 
         logger.notice('Analyser : ' + analyser + reflection + ' with energy = ' + str(efixed))
 
-        SetSampleMaterial(InputWorkspace=workspace, ChemicalFormula=formula)
+        SetSampleMaterial(InputWorkspace=workspace, ChemicalFormula=formula, SampleNumberDensity=density)
```
martyngigg: Should we do something similar here and get rid of the `MaterialBuilder` in here too and just switch on the property name? This also avoids doing the calculation twice?

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/FlatPlatePaalmanPingsCorrection.py
```diff
@@ -332,10 +332,10 @@ def _set_material(self, ws_name, chemical_formula, density_type, density):
             mat = builder.setFormula(chemical_formula).setMassDensity(density).build()
             number_density = mat.numberDensity
         else:
+            set_material_alg.setProperty('SampleNumberDensity', density)
```
martyngigg: Same comment as for `CylinderPaalmanPingsCorrection` above.

Framework/PythonInterface/test/python/plugins/algorithms/CalculateSampleTransmissionTest.py
```diff
@@ -13,16 +13,16 @@ def test_sample_transmission_calculation(self):
 
         # Using water sample
         formula = "H2-O"
-        density = 0.033424
+        density = 0.100272
```
martyngigg: What's the reason for changing the value here? Should this test case not stay the same?

Framework/PythonInterface/test/python/plugins/algorithms/IndirectTransmissionTest.py
```diff
@@ -123,12 +123,11 @@ def test_mass_density(self):
...
                                   ChemicalFormula=formula, DensityType='Mass Density', Density=density, Thickness=thickness)
 
-        ref_result = [6.65800, 0.82122, 2.58186, 53.50693, 56.08880, 0.03342, 0.1, 0.82676, 0.17096]
+        ref_result = [6.65800, 0.82122, 2.58186, 53.50693, 56.08880, 0.100272, 0.1, 0.56512, 0.43021]
         mass_values = ws.column(1)
         np.testing.assert_array_almost_equal(mass_values, ref_result, decimal=4)
 
-        # mass density 1.0 = number density 0.033424 for water
-        density = 0.033424
+        density = 0.100272
```
martyngigg: Same question here. I don't quite understand if we are still using the `Number Density` below why that needs to change as the `MaterialBuilder` was only corrected for the mass density setting wasn't it?

martyngigg (changes requested):
I think this is nearly there - I just have a few comments.

Framework/PythonInterface/plugins/algorithms/CylinderPaalmanPingsCorrection2.py
ianbush: I agree `builder` is no longer required, but I think the extra lines are still needed as the algo returns the number density.

Framework/PythonInterface/test/python/plugins/algorithms/CalculateSampleTransmissionTest.py
ianbush: The docs for the algorithm say that number density is in `Atoms / Angstrom^3`, so `1 g /cc` (standard water density) in `test_mass_density` should correspond to `0.100272 Atoms / Angstrom^3` (`3 x 0.033424 molecules / Angstrom^3`) here.

I think the point of this test was to show that mass density and number density, at standard density, give the same results. Hence why I thought changing this made more sense than changing the mass density.

Framework/PythonInterface/test/python/plugins/algorithms/IndirectTransmissionTest.py
ianbush: Same answer as above.

Framework/PythonInterface/test/python/plugins/algorithms/CalculateSampleTransmissionTest.py
ianbush: As an additional point when Miguel was testing this these were the numbers he used. I assume this is be cause `0.1 mm` of water at `1 g/cc` is a number most instrument scientists will know.

Framework/PythonInterface/plugins/algorithms/CylinderPaalmanPingsCorrection2.py
ianbush: Ah, perhaps it is better to do `density = ws.experimentInfo().sample().getMaterial().numberDensity`? This should ensure the calculation is done once.

martyngigg: Good point. Yes I think your option is the best way forward now.

Framework/PythonInterface/test/python/plugins/algorithms/CalculateSampleTransmissionTest.py
martyngigg: Fair point, that does make sense.

martyngigg (approved):

martyngigg:
I'm happy with these changes now. @peterfpeterson @louisemccann Please provide any comments if you have any.

Framework/Kernel/src/MaterialBuilder.cpp
```diff
@@ -257,15 +257,19 @@ double MaterialBuilder::getOrCalculateRho(
...
+           PhysicalConstants::N_A * 1e-24;
   } else if (m_formula && m_formula->size() == 1) {
     return m_formula->at(0).atom->number_density;
   } else {
-    return EMPTY_DBL();
+    throw std::runtime_error("The number density could be determined. Please "
+                             "provide the number density, ZParameter and unit "
+                             "cell volume or mass density.");
```
louisemccann: Should this be `The number density couldn't be determined`?

louisemccann (approved):
See comment about the error message. Aside from that looks good to me

Framework/Kernel/src/MaterialBuilder.cpp
ianbush: :+1: Whoops, good spot!

peterfpeterson (approved):
I think everything has been sorted out and is consistent.

Issue #19893, opened 2017-06-20
Resolve Valgrind issues in Kernel/KernelTests


DavidFair:
Resolve various issues picked up by valgrind including
- Of by one errors
- Leaked memory
- Bare `new` -> `unique_ptr` where possible 
- Uninitialised reads
...etc

PR #19980, opened 2017-07-10
Fixed off-by-one access and lost memory in TimeSeriesProperty

DavidFair:
Description of work.
As part of the issue fixing various issues from Valgrind in Kernel I picked up on a off-by-one error in TimeSeriesProperty.

If the `nextTspIndex` fell outside of the bounds of the vector (e.g. we are parsing right up to the last entry) we would read the value still. This went unnoticed as we would then go on to discard the incorrect value instead of using it.
By rearranging the loop slightly we can avoid the off by one access whilst still retaining existing behaviour.

There is also a fix for a memory leak in the unit test. Unfortunately because of the interface it is non-trivial to switch to using `unique_ptr` instead of a bare new / delete

**To test:**
Code review
Ensure unit tests pass

Part of #19893 . (Other PR is here: #19981 )

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Kernel/src/TimeSeriesProperty.cpp
+++ b/Framework/Kernel/src/TimeSeriesProperty.cpp
@@ -614,43 +614,49 @@ void TimeSeriesProperty<TYPE>::splitByTimeVector(
                   << "Time index = " << index_tsp_time << "\n\n";
 
     bool continue_add = true;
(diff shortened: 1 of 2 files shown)


DavidFair:
Closed by #19981

PR #19981, opened 2017-07-10
Fix various memory leaks in Kernel::Matrix

### LINKED ISSUE #19893: Resolve Valgrind issues in Kernel/KernelTests

DavidFair:
Description of work.
This PR resolves a memory leak which could occur when the move constructor is invoked by replacing the bare new / delete with `std::unique_ptr`. This correctly deletes the memory reserved for the array.

It also resolves various other potential memory leaks which were pointed out by @martyngigg if we exit early from the `GuassJordan` method by using a `std::vector` instead.

I've also tidied up the member variable names to following the coding convention. The biggest changes being `v` -> `m_rawData`, `nx` -> `m_numRowsX`, `ny` -> `m_numColumnsY`.

-----

On a slightly related note a fix was merged upstream for a memory leak in the `neXus` library we were having. This has been merged [upsteam](https://github.com/nexusformat/code/pull/450) but I'm not sure how long it will take for package maintainers to pick up the changes (or if they will). 

-----

**To test:**
Code review
Ensure unit tests pass - pay attention to the Python Goniometer tests. This is very sensitive to any changes in Kernel::Matrix because of the Python Interface layer and will usually highlight any problems immediately. 

Fixes #19893 

*Does not need to be in the release notes. - Internal changes only*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Kernel/src/Matrix.cpp
```diff
@@ -156,63 +160,61 @@ Matrix<T>::Matrix(const std::vector<T> &data, const size_t nrow,
...
-            jR++;
-          }
+  setMem(m_numRowsX, m_numColumnsY);
+  size_t iR(0);
+  for (size_t i = 0; i <= m_numRowsX; i++) {
+    if (i != nrow) {
+      size_t jR(0);
+      for (size_t j = 0; j <= m_numColumnsY; j++) {
+        if (j != ncol) {
+
```
martyngigg: Really picky I know but as there are other changes can we remove this whitespace?

Framework/Kernel/src/Matrix.cpp
```diff
@@ -975,36 +970,37 @@ T Matrix<T>::Invert()
...
-    if (V[0][0] != static_cast<T>(0.))
-      V[0][0] = static_cast<T>(1.) / V[0][0];
+  if (m_numRowsX == 1) {
+    T det = m_rawData[0][0];
+    if (m_rawData[0][0] != static_cast<T>(0.))
+      m_rawData[0][0] = static_cast<T>(1.) / m_rawData[0][0];
     return det;
   }
-  auto indx = new int[nx]; // Set in lubcmp
+  std::vector<int> indx;
```
martyngigg: We could do the sizing on 1 line with `std::vector<int> indx(m_numRowsX)`?

Framework/Kernel/src/Matrix.cpp
```diff
@@ -1033,18 +1030,20 @@ T Matrix<T>::factor()
    @return the factored matrix
 */
 {
-  if (nx != ny || nx < 1)
-    throw std::runtime_error("Matrix::factor Matrix is not NxN");
+  if (m_numRowsX != m_numColumnsY || m_numRowsX < 1)
+    throw std::runtime_error("Matrix::factor Matrix is not m_numRowsXN");
```
martyngigg: Replacement error on the string. The error message no longer makes sense.

martyngigg (changes requested):
Good set of improvements.

Some minor comments below.

Framework/Kernel/src/Matrix.cpp
```diff
@@ -1,9 +1,13 @@
-#include "MantidKernel/Exception.h"
 #include "MantidKernel/Matrix.h"
+
+#include "MantidKernel/Exception.h"
+#include "MantidKernel/make_unique.h"
 #include "MantidKernel/MersenneTwister.h"
 #include "MantidKernel/TimeSeriesProperty.h"
 #include "MantidKernel/V3D.h"
 
+#include <algorithm>
```
DavidFair: I pulled in `std::swap` for the transpose method

martyngigg: Ah yes. :+1:

Framework/Kernel/src/Matrix.cpp
DavidFair: Bit of a brain fart when I was writing that - I completely forgot you can set the size on construction, there is another place in this file I'll also change that too.

Framework/Kernel/src/Matrix.cpp
```diff
@@ -110,32 +114,32 @@ Matrix<T>::Matrix(const std::vector<T> &A, const std::vector<T> &B)
...
 }
 //
 template <typename T>
 Matrix<T>::Matrix(const std::vector<T> &data)
-    : nx(0), ny(0), V(nullptr) {
+    : m_numRowsX(0), m_numColumnsY(0) {
   size_t numel = data.size();
-  size_t nxt = static_cast<size_t>(sqrt(double(numel)));
-  size_t test = nxt * nxt;
+  size_t m_numRowsXt = static_cast<size_t>(sqrt(double(numel)));
```
DavidFair: Unfortunately we don't know `m_numRows` at this point as the constructor is testing if the matrix has valid dimensions at this point.

martyngigg: But we throw if it's not valid so if it's not square there won't be any object anyway...

Wouldn't something like:

```cpp
size_t numElements= data.size();
m_numRows = static_cast<size_t>(sqrt(double(numElements)));
if (numElements != m_numRows*m_numRows ) {
  throw(std::invalid_argument("...."
}
setMem(m_numRows , m_numRows);
...
```

work?

DavidFair: I would agree however we set the number of rows and columns in `setmem` so this ends up doing redundant work in this case. 

Looking through at the constructors in all but this case we have the details to set `m_numRows` and `m_numCols` as an initialiser list. I can also tidy up a lot of the code in `setmem` too by asserting that they are both set and just handling memory allocations in the function.

martyngigg: It's a fair point about the confusion over where `m_numRows` is set. Let's not change too much more structurally. My main thing here actually was just the variable name. Let's leave it as it currently is but have

```
size_t numRowsSquare = static_cast<size_t>(sqrt(double(numElements)));
```

as the variable name.

DavidFair: Oh woops - I've changed it so that `setmem` was a private method which takes no params to tidy it up.

To avoid this PR becoming too large I'll rename the variable names as suggested and hold onto the diff for a future PR?

martyngigg: Sounds good. `setMem` could better be called `resize` as a public member so I see no harm in leaving it as so.

DavidFair: But its only called internally and purely allocates a new block of memory hence I ended up refactoring it to do what it was documented. Anyway I'll open a separate low priority issue to apply the diff in the future.

Framework/TestHelpers/inc/MantidTestHelpers/FakeObjects.h
```diff
@@ -177,9 +177,9 @@ class WorkspaceTester : public AxeslessWorkspaceTester {
     AxeslessWorkspaceTester::init(numspec, j, k);
 
     // Put an 'empty' axis in to test the getAxis method
-    m_axes.resize(2);
-    m_axes[0] = new Mantid::API::RefAxis(j, this);
-    m_axes[1] = new Mantid::API::SpectraAxis(this);
+    m_axes.resize(0);
```
martyngigg: What's the reason for choosing `push_back` over the original implementation? I would still keep the `.resize(2)` anyway to avoid additional allocations.

martyngigg: @DavidFair Could you address this final comment? I think this is ready to go otherwise.

martyngigg (approved):
Good to go from me now.

PR #19992, opened 2017-07-12
Python fit function attribute error

### LINKED ISSUE #19989: Python fit function attribute error.
### Expected behavior

A fit function defined in python can override the default `setAttributeValue` method to, for example, store an attribute value in a variable. A call to `getAttributeValue` should return the new value.

### Actual behavior

If `setAttributeValue` is defined by following instructions from the online tutorial mantid crashes. Removing the call to the super class's `setAttributeValue` method stops crashing but then `getAttributeValue` doesn't return the new value.

Run the script. It will print out: `Attribute wasn't set properly: 0.05 != 1.1`

```
class fitefunc(IFunction1D):

    def init(self):
        self.declareAttribute("BinWidth", 0.05)

    def setAttributeValue(self, name, value):
        if name == "BinWidth":
            self._binW = value

        #super(fitefunc, self).setAttributeValue(name, value)

    # Define the fit function
    def function1D(self, xvals):
        binW = self.getAttributeValue('BinWidth')
        if binW != self._binW:
            print("Attribute wasn't set propertly: %s != %s" % (binW, self._binW))
        return np.ones(len(xvals))
    
FunctionFactory.subscribe(fitefunc)

f = FunctionFactory.createFunction('fitefunc')
f.setAttributeValue('BinWidth', 1.1)
f.function1D([1, 2, 3])
```

Uncomment the call to `super(...` to see the crash.

### Platforms affected

All

mantid-roman:
Store the new attribute value after calling user defined python method 'setAttributeValue`. 

**To test:**

Use the script from the issue description.

Fixes #19989.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/PythonInterface/mantid/api/src/FitFunctions/IFunctionAdapter.cpp
+++ b/Framework/PythonInterface/mantid/api/src/FitFunctions/IFunctionAdapter.cpp
@@ -167,6 +167,7 @@ void IFunctionAdapter::setAttribute(const std::string &attName,
     object value = object(handle<>(getAttributeValue(*this, attr)));
     callMethod<void, std::string, object>(getSelf(), "setAttributeValue",
                                           attName, value);
+    storeAttributeValue(attName, attr);
   } catch (UndefinedAttributeError &) {
     IFunction::setAttribute(attName, attr);
   }
--- a/Framework/PythonInterface/test/python/plugins/functions/AttributeTest.py
+++ b/Framework/PythonInterface/test/python/plugins/functions/AttributeTest.py
@@ -54,6 +54,12 @@ class AttributeTest(unittest.TestCase):
         self.assertAlmostEqual(res[1],  0.29960535, 7)
         self.assertAlmostEqual(res[2],  0.07493335, 7)
         self.assertAlmostEqual(res[3], -0.09645745, 7)
+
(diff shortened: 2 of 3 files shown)


Framework/PythonInterface/mantid/api/src/FitFunctions/IFunctionAdapter.cpp
```diff
@@ -167,6 +167,7 @@ void IFunctionAdapter::setAttribute(const std::string &attName,
     object value = object(handle<>(getAttributeValue(*this, attr)));
     callMethod<void, std::string, object>(getSelf(), "setAttributeValue",
                                           attName, value);
+    storeAttributeValue(attName, attr);
```
jmborr: Line 172 below `IFunction::setAttribute` is an alias for ``IFunction::storeAttributeValue`, so what is the rationale for the duplicate?

mantid-roman: `IFunction::setAttribute` is virtual with the default implementation that just stores the new value. Subclasses may choose to override it to, for example, change the number of parameters (see `Polynomial').

jmborr (approved):
EvaluateFunction, Fit, and the Fit wizard work as expected.

:shipit:

PR #20006, opened 2017-07-14
Added validation for event workspaces in ConjoinWorkspaces

### LINKED ISSUE #5564: ConjoinWorkspaces: validateInputs is not called for event workspaces
This issue was originally [TRAC 4717](http://trac.mantidproject.org/mantid/ticket/4717)

Original Reporter: Russell Taylor

Meaning that the check that both workspaces come from the same instrument, and have the same units is done for Workspace2Ds but not for EventWorkspaces.

josephframsay:
Validate EventWorkspaces according to their bins

**To test:**

```python
ws = CreateSampleWorkspace("Event")
ws2 = CreateSampleWorkspace("Event")

ws_cropped = CropWorkspace(ws, StartWorkspaceIndex=0, EndWorkspaceIndex=2)
ws2_cropped = CropWorkspace(ws2, StartWorkspaceIndex=3, EndWorkspaceIndex=5)

ws2_cropped = Rebin(ws2_cropped, Params="100,200,700")

ConjoinWorkspaces(ws_cropped, ws2_cropped)
```

Running the above script in Mantid should produce a ValueError, where previously there was an unhandled exception.

Fixes #5564 . 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/inc/MantidAlgorithms/WorkspaceJoiners.h
+++ b/Framework/Algorithms/inc/MantidAlgorithms/WorkspaceJoiners.h
@@ -56,8 +56,7 @@ protected:
 
   using Mantid::API::Algorithm::validateInputs;
   void validateInputs(const API::MatrixWorkspace &ws1,
-	  const API::MatrixWorkspace &ws2,
-	  const bool checkBinning);
+                      const API::MatrixWorkspace &ws2, const bool checkBinning);
   void validateInputs(const API::MatrixWorkspace &ws1,
                       const API::MatrixWorkspace &ws2);
   void getMinMax(const API::MatrixWorkspace &ws, specnum_t &min,
--- a/Framework/Algorithms/inc/MantidAlgorithms/WorkspaceJoiners.h
+++ b/Framework/Algorithms/inc/MantidAlgorithms/WorkspaceJoiners.h
@@ -56,7 +56,8 @@ protected:
 
   using Mantid::API::Algorithm::validateInputs;
   void validateInputs(const API::MatrixWorkspace &ws1,
-                      const API::MatrixWorkspace &ws2, const bool checkBinning = true);
+                      const API::MatrixWorkspace &ws2,
+                      const bool checkBinning = true);
   void getMinMax(const API::MatrixWorkspace &ws, specnum_t &min,
                  specnum_t &max);
 
--- a/Framework/Algorithms/inc/MantidAlgorithms/WorkspaceJoiners.h
+++ b/Framework/Algorithms/inc/MantidAlgorithms/WorkspaceJoiners.h
@@ -56,6 +56,9 @@ protected:
 
   using Mantid::API::Algorithm::validateInputs;
   void validateInputs(const API::MatrixWorkspace &ws1,
+	  const API::MatrixWorkspace &ws2,
+	  const bool checkBinning);
+  void validateInputs(const API::MatrixWorkspace &ws1,
                       const API::MatrixWorkspace &ws2);
   void getMinMax(const API::MatrixWorkspace &ws, specnum_t &min,
                  specnum_t &max);
--- a/Framework/Algorithms/inc/MantidAlgorithms/WorkspaceJoiners.h
+++ b/Framework/Algorithms/inc/MantidAlgorithms/WorkspaceJoiners.h
@@ -56,9 +56,7 @@ protected:
 
   using Mantid::API::Algorithm::validateInputs;
   void validateInputs(const API::MatrixWorkspace &ws1,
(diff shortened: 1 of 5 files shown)


Framework/Algorithms/src/ConjoinWorkspaces.cpp
```diff
@@ -55,6 +55,8 @@ void ConjoinWorkspaces::exec() {
   }
 
   if (event_ws1 && event_ws2) {
+    this->validateInputs(*event_ws1, *event_ws2);
+
     // We do not need to check that binning is compatible, just that there is no
     // overlap
     // make sure we should bother checking
```
KarlPalmen: Is this comment still correct?

Framework/Algorithms/src/ConjoinWorkspaces.cpp
```diff
@@ -71,25 +73,24 @@ void ConjoinWorkspaces::exec() {
...
-
-  MatrixWorkspace_sptr output = execWS2D(*ws1, *ws2);
-  // Copy the history from the original workspace
-  output->history().addHistory(ws1->getHistory());
-
-  // Delete the second input workspace from the ADS
-  AnalysisDataService::Instance().remove(getPropertyValue("InputWorkspace2"));
-  // Set the result workspace to the first input
-  setProperty("InputWorkspace1", output);
 }
```
KarlPalmen: Have you done anything besides replace (if { code, return } more_code) with if {code} else {more_code} ?

KarlPalmen: I see you added this->validateInputs(*event_ws1, *event_ws2);

KarlPalmen (commented):
I just want answers to my comments. The python script works as expected.

KarlPalmen (commented):
I haven't put names to all the new faces yet. So if you want to talk to me about this, please do so.

KarlPalmen (approved):
I'm happy with this.

martyngigg:
The RHEL7 system test has picked up a failure in `LETReduction`. It is now hitting this validation.

martyngigg:
It was explicitly stated in the code that the binning doesn't matter for events so I think we need to skip matching those and just test everything else.

Framework/Algorithms/src/WorkspaceJoiners.cpp
```diff
@@ -172,13 +172,16 @@ MatrixWorkspace_sptr WorkspaceJoiners::execEvent() {
...
 void WorkspaceJoiners::validateInputs(const MatrixWorkspace &ws1,
                                       const MatrixWorkspace &ws2) {
-  // This is the full check for common binning
-  if (!WorkspaceHelpers::commonBoundaries(ws1) ||
-      !WorkspaceHelpers::commonBoundaries(ws2)) {
-    g_log.error(
-        "Both input workspaces must have common binning for all their spectra");
-    throw std::invalid_argument(
-        "Both input workspaces must have common binning for all their spectra");
+  // Workspaces with point data are allowed to have different binning
```
martyngigg: I'm not this is exactly what we want here. Point data and event workspaces are different things. `EventWorkspace` is a special type of `MatrixWorkspace` that stores collections of signal/timestamp pairs per spectrum. Histograms of these collections are then generated on the fly from a current set of bin boundaries. Here the check is unimportant because of this on-the-fly generation.

For the `MatrixWorkspace` case, regarless of point or histogram status, the algorithm appends the spectra from the rhs to the lfs workspace but uses the X values exclusively from the lhs. If they are not the same in both cases then the numbers at the end could be misleading.

Perhaps the easiest way to solve this is to add a flag to `validateInputs` to disable the bin checking and then set this to `true` on line 58 of `ConjoinWorkspaces.cpp` as we already have checked for event vs non-event here?

Framework/Algorithms/inc/MantidAlgorithms/WorkspaceJoiners.h
```diff
@@ -56,6 +56,8 @@ class DLLExport WorkspaceJoiners : public API::Algorithm {
 
   using Mantid::API::Algorithm::validateInputs;
   void validateInputs(const API::MatrixWorkspace &ws1,
```
martyngigg: We can collapse these into a single method with a default parameter:
```
void validateInputs(const API::MatrixWorkspace &ws1,  const API::MatrixWorkspace &ws2, const bool checkBinning=true);
```

rather than needing two methods.

martyngigg (changes requested):
Good, we're almost there. I think there is just one small tidy up remaining.

martyngigg (approved):
Thanks, @josephframsay. This looks good now.

PR #20007, opened 2017-07-14
Added expanduser to Load in simpleapi

### LINKED ISSUE #19068: Script Mantid Python Window doesn't expand user variables in Linux
Quick sample code that will fail to find the workspace, in shell:
`touch ~/test.nxs` or use an existing file with a `~` character at the start of the path. 

Then in Mantid
`ws = Load(Filename=~/test.nxs)` this should find the file then fail. Currently it just fails as the file is not found.

The issue could be in the algorithms, because they are the ones responsible for expanding the path (I think)

josephframsay:
Description of work.

Added expansion of user's home directory (denoted by `~`) to `FileProperty`, so now files with `~` in the path can be loaded with standard behaviour, whereas previously they were not found.

**To test:**

Place a .nxs file in your home directory (~), or some subdirectory. From the script interpreter in MantidPlot, run `Load(Filename="~/<path-to-file>", OutputWorkspace="ws")`. The file should be loaded as expected.

Fixes #19068 . 

**Release Notes** 
[here](https://github.com/mantidproject/mantid/pull/20007/files#diff-d43ebaec89d7bc36364c8bf4c9f97673)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


dtasev (changes requested):
- Only works for `Load`, do you think it would be possible to add for all files that need to be found? See Martyn's suggestion: https://github.com/mantidproject/mantid/issues/19068

- If wrong filename type is passed, e.g. tuple or list, the original error is obscured:
- Original: 
```
ValueError: Problem when setting Filename. This is the detailed error description: When converting parameter "Filename": TypeRegistry::retrieve(): No PropertyValueHandler registered for type 'std::vector<std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >, std::allocator<std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > > >'
If the file has been found but you got this error, you might not have read permissions or the file might be corrupted.
If the file has not been found, you might have forgotten to add its location in the data search directories.
```
  - New:

```
/usr/lib/python2.7/posixpath.pyc in expanduser(path)
    252     """Expand ~ and ~user constructions.  If user or $HOME is unknown,
    253     do nothing."""
--> 254     if not path.startswith('~'):
    255         return path
    256     i = path.find('/', 1)

AttributeError: 'list' object has no attribute 'startswith'
```

josephframsay:
These changes fix both of the above problems, though some testing on machines other than mine might be a good idea

Framework/API/src/FileProperty.cpp
```diff
@@ -396,5 +401,55 @@ std::string FileProperty::convertExtension(const std::string &filepath) const {
...
+  // Position of the first slash after the variable
+  auto nextSlash = find_if(start, filepath.end(), &isSlash);
+
+  // UNIX-style home variable (ie ~/blah)
+  if (std::distance(start, nextSlash) == 1) {
+    char *home;
+
+    if (!(home = std::getenv("HOME")))
+      if (!(home = std::getenv("USERPROFILE")))
+        if (!(home = std::getenv("HOMEPATH")))
```
DTasev: `HOMEPATH` doesn't seem to add the drive letter, and will fail if it reaches that point, it would probably be better to remove it.

Also as it currently is, even if the variable exists, but is empty, e.g. `HOME="";USERPROFILE="/home/username` the condition will still go through on the first check for `HOME` and return the empty string, which causes a failure. I am not sure if we want to account for that.

@martyngigg Do you know if there is a chance for having a declared `HOME` and/or `USERPROFILE` but with an empty value?

dtasev (changes requested):

Framework/API/src/FileProperty.cpp
martyngigg: It's always possible that a user could set those to empty values. In this case I would just expect us to fail gracefully with a reasonable error message. I would use the values as they are given and if the path is not correct then it should give the normal message that it couldn't find the file.

Does https://github.com/python/cpython/blob/2.7/Lib/ntpath.py#L285 help? This is how python does it. Similarly the `posixpath` module defines the rules for Unix.

Can I suggest we `ifdef` this and use different rules for Windows and Unix?

DTasev: It seems for Windows we'd want both the `~` only syntax, and the `~username` syntax that I didn't know existed.

I checked and the syntax `~user/blah` does _not_ work on my Windows 10. @josephframsay Can you check if it works on Windows 7? 
Maybe we should remove that second check, and just leave the `~`? Then the only difference we have is:
- for Linux we check `HOME`
- for Windows we check `HOME` and `USERPROFILE`

As I said in the previous comment, `HOMEPATH` did not seem to include the drive letter

josephframsay: I think it works on my Windows 7. However I'm inclined to agree that we should just check for `~` - copying the behaviour from `ntpath` and `posixpath` has the potential to give us some real headaches. The library that makes expanding `~user/blah` feasible on UNIX doesn't exist on Windows, so I'm currently running into some really nasty (possibly really error-prone) workarounds with `ifdef`. Does anyone actually use the `~user/blah` syntax?

DTasev: In this case, I'd say keep only the `~` syntax, and remove the other one

dtasev (approved):

dtasev (changes requested):
Changes look good, only need to resolve merge conflict in release notes

Framework/API/src/FileProperty.cpp
```diff
@@ -396,5 +398,44 @@ std::string FileProperty::convertExtension(const std::string &filepath) const {
...
+ *  Note: this function does not support the "~user/blah" format for a named
+ *  user's home directory - if this is encountered, the filepath is returned
+ *  unchanged.
+ *  @param filepath The path to expand
+ *  @return The expanded path
+ */
+std::string FileProperty::expandUser(const std::string &filepath) const {
+  auto start = filepath.begin();
+
+  if (*start != '~') // No user variables in filepath
```
martyngigg: Is `filepath` guaranteed to be non-zero length? If not we should check if `start = filepath.end()` first.

Framework/API/src/FileProperty.cpp
```diff
@@ -396,5 +398,44 @@ std::string FileProperty::convertExtension(const std::string &filepath) const {
...
+    return filepath;
+
+  // Position of the first slash after the variable
+  auto nextSlash = find_if(start, filepath.end(), &isSlash);
```
martyngigg: `isSlash` could be a lambda as it's only used here.

Framework/API/src/FileProperty.cpp
```diff
@@ -396,5 +398,44 @@ std::string FileProperty::convertExtension(const std::string &filepath) const {
...
+
+  // ~user/blah format - no support for this as yet
+  if (std::distance(start, nextSlash) != 1)
+    return filepath;
+
+  char *home;
```
martyngigg: I'm finding this a little hard to parse. I'm also thinking that, for a given process, the value of `HOME`/`USERPROFILE` could probably be cached once and then reused.

If we put a function at anonymous namespace scope, something like:
```cpp
const std::string & getHomePath() {
    static std::string homePath;
    static bool initialized(false);
    if(initialized) return homePath;
    initialized = true;
    homePath = std::getenv("HOME");
    if(!homePath.empty()) return homePath;
    homePath = std::getenv("USERPROFILE");
    return homePath;
}
```
then the environment lookup will happen once and it'll be a little clearer.

martyngigg (changes requested):
I've made a few suggestions for code improvements.

It also looks like there are no unit tests for this new behaviour. Can we add some, please?

martyngigg:
@josephframsay Unfortunately the unit test has failed on Windows.

Framework/API/src/FileProperty.cpp
```diff
@@ -78,7 +78,7 @@ const std::string &getHomePath() {
 
   char *userProfile = std::getenv("USERPROFILE"); // Not usually set on UNIX
   // Return even if it's an empty string, as we can do no better
-  homePath = std::string(userProfile);
```
DTasev: Isn't this also an issue for `homePath = std::string(home);` above?

Framework/API/src/FileProperty.cpp
```diff
@@ -56,21 +55,105 @@ void addExtension(const std::string &extension,
...
+
+  char *home = std::getenv("HOME"); // Usually set on Windows and UNIX
+  if (home) {
+    homePath = std::string(home);
+    return homePath;
+  }
+
+  char *userProfile = std::getenv("USERPROFILE"); // Not usually set on UNIX
+  // Return even if it's an empty string, as we can do no better
+  homePath = userProfile ? std::string(userProfile) : "";
```
DTasev: Isn't the nullptr access also an issue for `homePath = std::string(home);` on the line above?

martyngigg (approved):
I'm happy with this now.

Framework/API/src/FileProperty.cpp
josephframsay: No, as there's a check for it on line 74

Framework/API/src/FileProperty.cpp
```diff
@@ -4,16 +4,15 @@
...
 #include "MantidKernel/Strings.h"
 
 #include <Poco/File.h>
 #include <Poco/Path.h>
 
 #include <boost/make_shared.hpp>
 
-#include <algorithm>
-#include <cctype>
+#include <cstdlib>
```
DTasev: Unused header

Framework/API/test/FilePropertyTest.h
```diff
@@ -184,6 +185,25 @@ class FilePropertyTest : public CxxTest::TestSuite {
...
+  void testExpandUserVariables() {
+    FileProperty fp("Dir", "", FileProperty::Directory);
+    std::string msg = fp.setValue("~");
+
+    char *homepath = std::getenv("HOME");
+    if (!homepath) {
+      homepath = std::getenv("USERPROFILE");
+    }
+
+    if (homepath && Poco::File(homepath).exists()) {
```
DTasev: Wouldn't this segfault is `USERPROFILE` and `HOME` are not set?

dtasev (changes requested):
Just minor comments

Framework/API/test/FilePropertyTest.h
josephframsay: I don't think so, as I think `getenv` returns a null pointer if the variable is not set, so the first condition of the `&&` will fail

dtasev (approved):

DTasev:
Happy to merge it now

PR #20009, opened 2017-07-14
ManageUserDirectories add missing slashes

### LINKED ISSUE #6098: Directory handling is not consistent in the Manage User Directories Dialog
This issue was originally [TRAC 5252](http://trac.mantidproject.org/mantid/ticket/5252)

Spencer has reported:

```
There is a 'feature' which sometimes throws out my routines.
In the Data Search Directories list the location ends with / or \.
Default Save Directory when you Browse & Select doesn't have the that last / or \.
Then when I do os.path.join the location is incorrect.
If I remember I add the /  !
```

We need to have a much more consistent handling of directory paths throught Mantid, i.e properties files, GUI etc

---

Keywords: NewStarter

josephframsay:
Slashes are now added to paths in the ManageUserDirectories dialog if they are not already there. This means that paths are treated as paths to directories, so can be loaded and joined correctly.

**To test:**

Open the 'Manage User Directories' dialog, add one of the following paths. 

- Path with no slash at the end
- Path of empty spaces
- Path with spaces at one (or both) ends
- Empty path

On reopening the dialog, the path should appear with a slash appended where appropriate.

Fixes #6098 . 

**Release Notes** 
*Does not need to be in the release notes.* - This is a minor change which Mantid should have previously handled

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidQt/API/src/ManageUserDirectories.cpp
+++ b/MantidQt/API/src/ManageUserDirectories.cpp
@@ -133,8 +133,9 @@ void ManageUserDirectories::saveProperties() {
 void ManageUserDirectories::appendSlashIfNone(QString &path) const {
   path = path.trimmed();
   if (!(path.endsWith("/") || path.endsWith("\\") || path.isEmpty())) {
-	  // Don't need to add to a \\, as it would just get changed to a / immediately after
-	  path.append("/");
+    // Don't need to add to a \\, as it would just get changed to a /
+    // immediately after
+    path.append("/");
   }
 }
 
--- a/MantidQt/API/src/ManageUserDirectories.cpp
+++ b/MantidQt/API/src/ManageUserDirectories.cpp
@@ -132,10 +132,9 @@ void ManageUserDirectories::saveProperties() {
 */
 void ManageUserDirectories::appendSlashIfNone(QString &path) const {
   path = path.trimmed();
-  if (path.endsWith("/") || path.endsWith("\\") || path.isEmpty()) {
-    return;
-  } else {
-    path.append("/");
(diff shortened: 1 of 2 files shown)


MantidQt/API/src/ManageUserDirectories.cpp
```diff
@@ -119,6 +124,21 @@ void ManageUserDirectories::saveProperties() {
...
 
+/**
+ * Appends a forward slash to the end of a path if there is no slash (forward or
+ * back) there already, and strip whitespace from the path.
+ *
+ * @param path :: A reference to the path
+*/
+void ManageUserDirectories::appendSlashIfNone(QString &path) const {
+  path = path.trimmed();
+  if (path.endsWith("/") || path.endsWith("\\") || path.isEmpty()) {
```
brandonhewer: Is it possibly better to use:
```c++
if(!(path.endsWith("/") || path.endsWith("\\") || path.isEmpty())) { 
  path = path.append("/");
}
```
Rather than if-else with a return statement (i.e. using a not operator).

brandonhewer (changes requested):

brandonhewer (approved):

martyngigg:
@brandonhewer Can I just check that you did some functional testing here and not just a code review?

brandonhewer:
@martyngigg Yes I tried each case given in the description (no slash at end, empty spaces, path with spaces at both ends, empty path). I also attempted adding extra slashes and whitespace into the file path. Resulting behaviour was as expected.

martyngigg:
@brandonhewer Great. Thanks.

Just for future reference when you submit the review could you give a brief description of what you tried. It helps for future reference.

PR #20010, opened 2017-07-14
InfoComponentVisitor makes DetectorInfo and ComponentInfo

### LINKED ISSUE #20002: Move DetectorInfo into InfoComponentVisitor
This is a grouping or related items taken from the umbrella Item listed here: #19964

* `InfoComponentVisitor` to create `DetectorInfo` as well as `ComponentInfo` (1 from umbrella list)
* `Beamline::DetectorInfo` and `Beamline::ComponentInfo` should be created together. Construction should be as safe as possible as both have a runtime dependency on each other (2 from umbrella list)
* Remove work-around code in ComponentInfo::getPos etc, that tries to deal with current staged creation of the `..Info` objects. (7 from umbrella list)
* Possibly link these items via some other structure to ensure that lifetime rules (internal pointer ownership of each) make better sense. This is is a design aspect that falls out of the above. I am proposing a grouping called `Beamline` (agreed name for Instrument 2.0), that holds and provides well established semantics for `ComponentInfo` and `DetectorInfo`.

OwenArnold:
Fixes #20002. Make `ComponentInfo` and `DetectorInfo` via visitor double dispatch mechanism.

WIP

Check key performance areas such as `CalibrateSXDPanels` and common Load/Save algorithms. Should be at least as good as base branch `instrument_2_0_1` 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Beamline/inc/MantidBeamline/Beamline.h
+++ b/Framework/Beamline/inc/MantidBeamline/Beamline.h
@@ -47,6 +47,8 @@ public:
   Beamline();
   Beamline(const Beamline &other);
   Beamline &operator=(const Beamline &other);
+  Beamline(Beamline &&) = default;
+  Beamline &operator=(Beamline &&) = default;
   Beamline(ComponentInfo &&componentInfo, DetectorInfo &&detectorInfo);
   const ComponentInfo &componentInfo() const;
   const DetectorInfo &detectorInfo() const;
--- a/Framework/Beamline/inc/MantidBeamline/Beamline.h
+++ b/Framework/Beamline/inc/MantidBeamline/Beamline.h
@@ -2,7 +2,8 @@
 #define MANTID_BEAMLINE_BEAMLINE_H_
 
 #include "MantidBeamline/DllConfig.h"
-#include <memory>
+#include <boost/shared_ptr.hpp>
+#include <boost/proto/functional/range/empty.hpp>
 
 namespace Mantid {
 namespace Beamline {
@@ -18,6 +19,8 @@ class DetectorInfo;
   this type ensures
   that client code is not exposed to creation or setup internals. Beamline owns
   both ComponentInfo and DetectorInfo.
+
+  Beamline is deliberately trivial to copy.
 
   Copyright &copy; 2017 ISIS Rutherford Appleton Laboratory, NScD Oak Ridge
   National Laboratory & European Spallation Source
@@ -42,16 +45,20 @@ class DetectorInfo;
 */
 class MANTID_BEAMLINE_DLL Beamline {
 public:
-  Beamline(std::unique_ptr<ComponentInfo> &&componentInfo,
-           std::unique_ptr<DetectorInfo> &&detectorInfo);
+  Beamline() = default;
+  Beamline(const Beamline &other);
+  Beamline &operator=(const Beamline &other);
+  Beamline(ComponentInfo &&componentInfo, DetectorInfo &&detectorInfo);
   const ComponentInfo &componentInfo() const;
   const DetectorInfo &detectorInfo() const;
   ComponentInfo &mutableComponentInfo();
   DetectorInfo &mutableDetectorInfo();
+  bool empty() const;
 
 private:
-  std::unique_ptr<ComponentInfo> m_componentInfo;
-  std::unique_ptr<DetectorInfo> m_detectorInfo;
+  bool m_empty = true;
+  boost::shared_ptr<ComponentInfo> m_componentInfo{nullptr};
+  boost::shared_ptr<DetectorInfo> m_detectorInfo{nullptr};
 };
 
 } // namespace Beamline
--- a/Framework/Beamline/inc/MantidBeamline/Beamline.h
+++ b/Framework/Beamline/inc/MantidBeamline/Beamline.h
@@ -2,14 +2,13 @@
 #define MANTID_BEAMLINE_BEAMLINE_H_
 
 #include "MantidBeamline/DllConfig.h"
-#include <boost/shared_ptr.hpp>
-#include <boost/proto/functional/range/empty.hpp>
+#include "MantidBeamline/ComponentInfo.h"
+#include "MantidBeamline/DetectorInfo.h"
+#include <memory>
 
 namespace Mantid {
 namespace Beamline {
 
-class ComponentInfo;
-class DetectorInfo;
 /** Beamline : Also known as Instrument 2.0
 
   This is the top-level object for accessing ComponentInfo, DetectorInfo and
@@ -45,7 +44,7 @@ class DetectorInfo;
 */
 class MANTID_BEAMLINE_DLL Beamline {
 public:
-  Beamline() = default;
+  Beamline();
   Beamline(const Beamline &other);
   Beamline &operator=(const Beamline &other);
   Beamline(ComponentInfo &&componentInfo, DetectorInfo &&detectorInfo);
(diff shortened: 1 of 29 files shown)


Framework/Beamline/inc/MantidBeamline/Beamline.h
```diff
@@ -0,0 +1,72 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+class MANTID_BEAMLINE_DLL Beamline {
```
SimonHeybrock: I think we will be running into trouble with this name -- there will be clashes with the namespace of the same name (I had this issue with `Histogram`, this is why the module is names `HistogramData`.

Framework/Beamline/inc/MantidBeamline/Beamline.h
```diff
@@ -0,0 +1,72 @@
...
+  Beamline alias();
+  bool empty() const;
+
+private:
+  // For internal use only.
+  Beamline(boost::shared_ptr<ComponentInfo> &componentInfo,
+           boost::shared_ptr<DetectorInfo> &detectorInfo);
+  bool m_empty;
+  boost::shared_ptr<ComponentInfo> m_componentInfo;
+  boost::shared_ptr<DetectorInfo> m_detectorInfo;
```
SimonHeybrock: Why do we need to store these as `shared_ptr`? The data inside `ComponentInfo` and `DetectorInfo` is shared, so it is not clear why we would want another level of sharing.

Framework/Geometry/inc/MantidGeometry/Instrument.h
```diff
@@ -6,25 +6,22 @@
 #include "MantidGeometry/IDetector.h"
 #include "MantidGeometry/Instrument/CompAssembly.h"
 #include "MantidGeometry/Instrument/ObjComponent.h"
-
 #include "MantidKernel/DateAndTime.h"
+#include "MantidBeamline/Beamline.h"
```
SimonHeybrock: Consider avoiding this include and instead use a forward declaration (and store `Beamline` as `unique_ptr`).

Framework/Geometry/inc/MantidGeometry/Instrument/ParameterMap.h
```diff
@@ -6,6 +6,7 @@
 #include "MantidGeometry/IDetector.h"
 #include "MantidGeometry/IDTypes.h" //For specnum_t
 #include "MantidGeometry/Instrument/Parameter.h"
+#include "MantidBeamline/Beamline.h"
```
SimonHeybrock: Maybe also avoid this include?

Framework/Geometry/src/Instrument.cpp
```diff
@@ -1317,14 +1286,15 @@ boost::shared_ptr<ParameterMap> Instrument::makeLegacyParameterMap() const {
...
-  pmap->setComponentInfo(nullptr);
+  pmap->clearBeamline();
   // Instrument is only needed for DetectorInfo access so it is not needed.
   pmap->setInstrument(nullptr);
 
-  if (!hasDetectorInfo() && !hasComponentInfo())
+  if (!hasInstrumentVisitor())
     return pmap;
 
+  auto componentCache = m_instrVisitor->componentIds();
```
SimonHeybrock: `auto &`?

SimonHeybrock (changes requested):
Looks pretty good already. My main comment is regarding the copy/ownership mechanism in `Beamline::Beamline` that you already mentioned on Skype. I do not really understand the mechanism. It seems to be made for handling cases where one of `ComponentInfo` and `DetectorInfo` can be `nullptr`? What is that used for?

Framework/Beamline/inc/MantidBeamline/Beamline.h
OwenArnold: There was a previous chat on the general slack thread, and I think Duc suggested "Beamline" a name. Instrument 2.0 doesn't seem to cut, it and `Instrument` is too easily confused with existing `Instrument 1.0` IMHO. I do agree that `Beamline` itself might not be descriptive enough. Open for suggestions... maybe

* `Geometry`  - as in `Beamline::Geometry`, too easily confused with namespace `Geometry`?
* `BeamlineInstrument`?

Framework/Beamline/inc/MantidBeamline/Beamline.h
OwenArnold: I had to change it from a `unique_ptr` to this. I think i'm going to write down the various object/ownership issues I encountered. I did add a regression tests for each of these as I came across and fixed these. I think some of your questions will relate to these so I'll write it properly once.

Framework/Geometry/inc/MantidGeometry/Instrument.h
OwenArnold: Considered that, and sat of the fence for a while - eventually decided that ownership and copying semantics issues were complex enough that I wanted to shield from the client. will define further below.

Framework/Beamline/inc/MantidBeamline/Beamline.h
SimonHeybrock: The discussion in slack was when I looked for a module name, and indeed I then used Duc's suggestion `Beamline`.

- `Geometry` has nothing to do with neutron instruments.
- `BeamlineInstrument` does not make a lot of sense?
- `Instrument` does not seem too bad for me. The existing `Geometry::Instrument` is being removed after all.

Personally I would, I think, also prefer `Beamline::Beamline`, but I think that implies that we can never use `using namespace Beamline`? Is that an issue?

Framework/Geometry/src/Instrument.cpp
OwenArnold: 👍

OwenArnold:
1. **Deep Copy**

Ensuring correct COW of held requires invoking the copy constructor of DetectorInfo and ComponentInfo. For example, when an `ExperimentInfo` is copied, if the COW ptr holding the vector of positions is not reference count incremented, then modifying positions of the first `Workspace`/`ExperiementInfo` will result in positions on the second incorrectly being changed.

```cpp
ExperimentInfo a;
...
ExperimentInfo b = a; // Deep copy in beamline required here!
a.beamline().detectorInfo().setPosition(V3D{1,1,1}); // Otherwise sharing in b not broken.
```

Current solution: Copy constructors of held pointers (`DetectorInfo` and `ComponentInfo`) in `Beamline` must be invoked on a `Beamline` copy

2. **Shallow Copy**

The original design had `ExperimentInfo` as the owner of `Beamline`. References loaned out to `ParameterMap` as part of `ExperimentInfo::setInstrument`. However, th there are cases where the Instrument is used after the ExperimentInfo has goine out of scope. While we do NOT want a deep copy, we need to ensure that the ParameterMap can have shared ownership of the beamline. See https://github.com/mantidproject/mantid/blob/8c85f9a168ef352e6dc9cbebdb03cc58fee69b15/Framework/API/test/ExperimentInfoTest.h#L974-L999

Current solution: `Beamline` provides an `alias` member function which calls a private constructor, which copies the held pointers but elides their copy constructors. Ownership of member pointers is shared so that use cases such as: `ParameterMap` can remain valid after creating `ExperimentInfo` goes out of scope.

3. **Move construction and assignment**

For convenience we might want to move an existing named beamline object, or reuse a temporary. 

Current solution: A default move constructor and move assignment operator are declared on `Beamline`.

SimonHeybrock:
I think this has been dealt with in another PR, closing?

PR #20040, opened 2017-07-19
Indirect - Elastic Window Multiple - Normalizing by lowest temperature

### LINKED ISSUE #20039: Indirect - Elastic Window - Normalise by lowest temperature
Add code for normalizing list of temperatures by the lowest temperature in elastic window algorithm.

brandonhewer:
Temperatures from the ElasticWindowMultiple algorithm are currently not being normalised by lowest temperature.

Changed retrieving the first temperature in the list of temperatures, to finding the minimum.

**To test:**
1. Open Mantid release 3.10.
2. Navigate to Interfaces > Indirect > Data Analysis and open the Elwin tab.
3. In the 'Input File' field, provide *_red.nxs files which were reduced with a temperature log (files provided below).
4. Ensure 'Normalize to Lowest Temp' is checked.
5. Run the algorithm.
6. Save the resultant workspace.
7. Open Mantid from this branch.
8. Repeat steps 2-5, ensuring the same workspace is used.
9. Load the previously saved file into the ADS.
10. View both workspace - if the lowest temperature was not the first y-value in the original *_red.nxs file, these workspaces should be different - else they should be the same.

Test files:
[Elwin Files.zip](https://github.com/mantidproject/mantid/files/1176302/Elwin.Files.zip)

Fixes #20039 

**Release Notes** 

Added fix to the release notes.
---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


jmborr:
@brandonhewer please propose one or more tests, and also write release notes

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/ElasticWindowMultiple.py
```diff
@@ -318,8 +320,9 @@ def _get_sample_units(self, ws_name):
         if self._sample_log_name in run:
             # Look for sample unit in logs in workspace
             tmp = run[self._sample_log_name].value
-            value_action = {'last_value': lambda x: x[len(x) - 1],
-                            'average': lambda x: x.mean()}
+            value_action = {'last_value': lambda x: x[-1],
+                            'average': lambda x: x.mean(),
+                            'lowest_value': lambda x: np.amin(x)}
```
jmborr: Lines 307 to 309 are the same as lines 323 to 325. They should be placed at the method's scope (line 302)

brandonhewer: We're just waiting to see whether this pull request is relevant - there was potentially miscommunication from Spencer.

jmborr: OK

SpencerHowells:
@brandonhewer Just noticed that getInstrRun is imported from IndirectCommon. This is probably better off within the algo.

jmborr:
[Line 245](https://github.com/mantidproject/mantid/blob/7eabd6e69dbd2413faad94d7690db6d4f27279e7/Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/ElasticWindowMultiple.py#L245) is the condition to instantiate `self._elt_workspace`, so shouldn't the line be:

 `if self._elt_workspace == '':`

jmborr:
In method `PyInit()`, the description of properties `OutputInQ`, `OutputInQSquared`, `OutputELF`, and `OutputELT` should state the physical meaning of the data contained, not tautologies like "Output workspace ELT" :smirk:

brandonhewer:
@jmborr Line 245 is actually quite strange - If this line is supposed to be:
```python
if self._elt_workspace == '':
```
This would imply Line 240:
```python
if self._elt_workspace != '':
```
is incorrect.

brandonhewer:
I believe the comment here on what the code is doing is incorrect.

jmborr:
@brandonhewer I think the confusion originates in the reuse of `self._elt_workspace`. In Line 240 `self._elt_workspace` contains the name of the output workspace, i.e, a _string_. Notice how the contents of `self._elt_workspace` are compared against the empty string. Then in Line 258 `self._elt_workspace` is assigned a workspace _instance_. That the same attribute means different things at different parts of the code is bound to generate confusion when someone other than the creator of the algorithm reads the code. I would have one variable for the name, and one variable for the object itself.

brandonhewer:
@jmborr A good observation - I'll change this in order to avoid confusion!

jmborr:
@brandonhewer so I tried to do the test but the data files are not useful, unless I'm missing something. The file with the lowest run number (`osi104367_graphite002_red.nxs`) is also the file with the lowest sample temperature assigned. Thus, the function `_normalize_to_lowest_temp` in Mantid v.3.10 works fine and both v.3.10 and this branch give the same results.

SpencerHowells:
@jmborr The original version assumed that the first run is the lowest with temperature increasing. I first thought of checking the first & last run as to which is the lowest to take into account the case of starting high & decreasing. Then I thought that a user might go up & down and this change would take it into account. We don't have data doing that!

jmborr:
@SpencerHowells I tried loading the files from highest run number to lowest, but lo and behold, statement `self.getProperty('InputWorkspaces').value.getNames()` actually sorts the file names so I end up with a list from lowest to highest run number :sob:
In the end I did a silly hack, I renamed one of the files with intermediate run numbers with a run number which is now the lowest of the set. Now the normalization does not select this file because it is not the one with the lowest temperature, as it should be :smile:

jmborr (approved):
Works as expected
:shipit:

quantumsteve:
@jmborr We're trying to kill off the :squirrel:

PR #19977, opened 2017-07-07
ISIS Reflectometry: Reprocess rows with changed data

PranavBahuguna:
This PR adds several new features to the ISIS Reflectometry interface:
- While reduction is paused, selecting new rows of data will cause them to be processed after resuming while deselected ones and already processed data items will be ignored. This is as opposed to previously where the interface simply reprocessed all selected rows, regardless of whether they had been processed before.
- Manually altering data in a previously processed row while paused will cause that row and its containing group to be set unprocessed.
- Adding/removing rows from a group while paused will set that group unprocessed.

**To test:**
- Code review.
- For manual testing:
  - Open the interface from Interfaces->Reflectometry->ISIS Reflectometry
  - Change instrument to `POLREF` and enter investigation id `1520249` (you may need to login to the ICAT catalog). Transfer all runs to the table.
  - Select several groups and process them. Pause midway through and select some new additional groups before processing again. Fully processed groups that were selected should be ignored and the new groups should be processed too.
  - Select several groups, process them and as before, pause midway through. In a previous fully-processed group, change the data of some of the rows. When you resume again, these rows (as well as their containing groups) should be set unprocessed.
- Repeating same steps as before, add/remove some rows from groups while paused. The affected groups should be set unprocessed,

Fixes #19970.

**Release Notes** 

*Release notes have been updated.*

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [x] Is the code of an acceptable quality?
- [x] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)?
- [x] Are the unit tests small and test the class in isolation?
- [x] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [x] Do changes function as described? Add comments below that describe the tests performed?
- [ ] Do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant (user and developer) documentation been added/updated?

Does everything look good? Mark the review as **Approve**. A member of `@mantidproject/gatekeepers` will take care of it.

Issue #19951, opened 2017-07-03
ISIS Reflectometry: Prevent external table modification while processing


PranavBahuguna:
While the current interface greys out the `process` buttons and prevents window closure while processing runs, modification to the table of data is not prevented and can cause errors when done. These include:

- Adding/deleting rows/groups
- Grouping rows
- Cutting/pasting rows
- Editing row data
- Transferring rows

In addition, the whole `Reflectometry` menu (except for `Export .TBL` and `Slit Calculator`) should be disabled (#21215). Furthermore, one should not be able to select any run in the table while processing occurs.

Edit: Another bug I want to address in this is that currently, pausing reduction will immediately make the `process` button immediately available (before reduction itself has actually paused). If `process` is clicked before reduction has finished, a critical error would be thrown. `Process` should be re-enabled only when reduction has finished, not just when the pause button is clicked. This has been noted as part of #21167.

In summary we need to:
- [x] Disable editing rows in the data processor widget
- [x] Disable relevant items in the menus #21215

gemmaguest:
I'm closing this issue as the main fixes have been added. There is one related issue which is not a priority at the moment so has been moved to #21167

PR #20041, opened 2017-07-19
Reflectometry interface: Introduce autoreduction button

### LINKED ISSUE #19453: Reflectometry Interface: Autoreduce a set of runs given an investigation ID
A button to autoreduce data should be added to the interface. When clicked, this button will do the following:
- Check if an investigation ID has been entered (and request user to log into catalog and enter an ID if not).
- Obtain runs from the investigation ID.
- Transfer runs to the table.
- Reduce all rows in the table.
- Be able to pause and restart autoreduction

Part of umbrella #17029.

PranavBahuguna:
As described in the attached issue, this PR adds a button to perform autoreduction. When supplied an investigation id and the button is clicked, the following occurs:

- Obtain runs that are part of the investigation the id was supplied for.
- Transfer all runs to the processing table.
- Select all runs and process them.

When autoreduction is running, the button is disabled. Pausing reduction will re-enable the button. If the instrument, investigation id or transfer strategy is changed while paused before clicking 'autoreduce' again, a new autoreduction (including searching for runs and transfer) will be started. Otherwise it simply resumes reducing runs just like `Process` would.

~_Wait for PR #19977 to be merged first before merging this._~

**To test:**
- Code review.
- Ensure documentation is updated and sensible.
- Manual testing steps:
  - Login into the ICAT catalog.
  - The interface can be accessed from Interfaces->Reflectometry->ISIS Reflectometry.
  - Select `POLREF`, enter investigation id `1520249` and click on the `autoreduce` button to begin autoreduction.
  - Ensure that changing id, instrument or transfer strategy causes a new autoreduction to be started.
  - Currently it is possible to cause critical errors by clicking on buttons such as `delete row` while reduction is in progress. These will be dealt with in #19951.

Fixes #19453.

**Release Notes** 

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


gemmaguest:
Some initial observations:

### Highlighting
The green highlighting of completed rows is a nice feature but I'm noticing a few inconsistencies with it on Windows 10:
- If a parent row only has a single child row, the child row gets highlighted but the parent row doesn't. (If it has more than one child row, then the parent is highlighted when all child rows are complete, as you'd expect.)
- The highlighting is sometimes not done until I force a redraw e.g. by moving the mouse over the rows.
- When you click on a green row, the selected row colour is only marginally different to the non-selected row colour, so it's hard to see which rows are selected.

### Transmission run
The reduced runs include the transmission run, which gives an error because the angle is 0.0. This should probably be excluded, and maybe the 'Alignment check' run too - we should ask Max about how we can find out which runs to exclude.

### Progress bar
- When you pause the autoreduction, the progress bar and %complete number stop where they were, which is great, but the bar still animates. Would it be better if it stopped animating to be clear that nothing is currently happening?
- I wasn't sure how to re-start autoreduction - I assume you hit the `Autoreduce` button again? It might be nice to have a resume button next to the pause button to make this clearer, even if it just does the same thing. It could use the same Play icon as the Autoreduce button.
- The Pause icon looks more like a Stop icon - is there a better Pause icon available?
- When resuming autoreduction, the progress bar and %complete number start from 0 again. Would it make more sense for them to continue from where they were? I guess this might not make sense if changes have been made, though?

PranavBahuguna:
@gemmaguest Thanks for reviewing my PR. Regarding your observations:

###Highlighting

- 1 - This is actually deliberate - the parent highlighting is supposed to be done only if post-processing is done with all child rows. With one child, no post-processing is done, so the parent is un-highlighted. We can change this of course, but I felt that doing it this way made more sense.

- 2 - There is nothing I can do about this unfortunately.

- 3 - Probably something to discuss with scientists, that is if they want selected rows to be highlighted. Will handle this in a separate issue.

###Transmission run

- Agreed, we should bring this up in the meeting with them

###Progress bar

- 1 - I'll see what I can do about this, probably in a separate issue

- 2 - Agreed. I think it might be nicer to somehow to have a single autoreduction button whose icon changes from a 'play' to a 'pause' icon.

- 3 - I had a design for one, probably won't be hard to implement.

- 4 - While it could be possible to not reset it to zero, this could potentially get very messy quickly it requires storing the progress and accounting for adding/deleting rows, changing data in processed rows and or changing the selection. Meanwhile, simply resetting progress isn't too bad an alternative as it can still show the progress of reduction of the remaining unprocessed rows, which is still helpful. Trying to change it would probably be far more effort than its worth, considering all the other possible things that can be done with the interface.

gemmaguest:
:+1: That all sounds good. Nothing here's a showstopper so it's fine to deal with them in separate issues.

gemmaguest (approved):
The code changes look good and behave as described :+1:

A very minor thing which could be fixed later: if you are not logged in to ICat and you enter incorrect login details, you get the message "Error Logging in: Please press 'Search' to try again.". This is slightly misleading now this is not just specific to the `Search` button.

PranavBahuguna:
@gemmaguest Thanks, I'll just throw those extra things into some other PR.

PR #20042, opened 2017-07-19
Periods crash in muon analysis

### LINKED ISSUE #19975: Periods in muon analysis causing a crash.
The following instructions caused the error 'Fit Algorithm failed. Unknown property search object InputWorkspace_1' 

1. Change facility to ISIS, default instrument to MUSR
2. Open the Muon Analysis interface
3. Click Manage Directiories and add Mantid training course data to Data Search Directories
4. Browse... and open MUSR00015189.nxs from the training course data
5. Switch between the three options in Group/Group Pair combo box.
6. Go to Settings tab, make sure multiple fitting is disabled and TF Asymmetry enabled
7. Go to Data Analysis tab
8. There is thee options in Data/Workspace combo box. Switch between them and the number of options increases...
9. When you have all six options in the Data/Workspace combo box, add GausOsc fitting function within the CompositeFunction
10. Fit. An error dialog with the aforementioned message appears

There are two bugs here. The first is that the periods are not loaded into data analysis correctly in the first place. The second is that the fit is trying to fit to multiple periods (the bugs do not occur if there is only a single period).

AnthonyLim23:
Muon analysis was crashing when multiple period data was loaded into the single fit. 

**To test:**

The following instructions caused the error 'Fit Algorithm failed. Unknown property search object InputWorkspace_1'

1. Open the Muon Analysis interface
2. Load MUSR00015189.nxs 
3. Switch between the options in Group/Group Pair combo box on the home tab (two is enough).
4. Go to Settings tab, make sure multiple fitting is disabled
5. Go to Data Analysis tab
6. Add a function (e.g. flatBackground) and do a fit. It should work.
7. Change the workspace and do another fit. These should also (now) work. 

Ignore the message "cannot reach the specified tolerance in F", this just means that the fit was poor. 

Fixes #19975. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


MantidQt/MantidWidgets/src/MuonFitPropertyBrowser.cpp
```diff
@@ -380,6 +380,27 @@ void MuonFitPropertyBrowser::enumChanged(QtProperty *prop) {
...
+      }
+      end = option.find_first_of(";");
+      QString selectedPeriod = QString::fromStdString(option.substr(0, end));
+      // turn on only the relevant box
+      for (auto iter = m_periodBoxes.constBegin();
+           iter != m_periodBoxes.constEnd(); ++iter) {
+        if (selectedPeriod == iter.key()) {
+          m_boolManager->setValue(iter.value(), true);
+        } else {
+          m_boolManager->setValue(iter.value(), false);
```
PranavBahuguna: Why not just do something like this?

```
someBool = selectedPeriod == iter.key()
m_boolManager->setValue(iter.value(), true);
```

Or even just:

```
m_boolManager->setValue(iter.value(), selectedPeriod == iter.key());
```

PranavBahuguna (changes requested):
Changes work as expected and the bug no longer appears when following the testing instructions. Code also looks good, just address my one comment and its probably good to go.

MantidQt/MantidWidgets/src/MuonFitPropertyBrowser.cpp
AnthonyLim23: do you mean

```
bool state = (selectedPeriod == iter.key()) ? true : false;
m_boolManager->setValue(iter.value(),,state);
```

PranavBahuguna: I don't think the conditional is necessary, it should compile just fine without it.

AnthonyLim23: I will try it after lunch.

PranavBahuguna (approved):
All looks good, :shipit: once checks pass

PR #20045, opened 2017-07-19
Interactive Plots in Performance Test Reports

### LINKED ISSUE #20044: Interactive Performance Test Plots
Currently, performance test graphs are [static images](http://builds.mantidproject.org/job/master_performancetests2/Master_branch_performance_tests/) which give an indication of how performance changes with changes to the code base.  Monitoring performance test has become increasingly tedious since the number of performance tests is ever increasing. One proposal which came out of the developer workshop 2017 was the introduction of interactive graphs where each data point is a clickable link which navigates to the corresponding commit which resulted in the build of the performance tests. The [plotly](https://plot.ly/python/) plotting library was identified as the tool which would facilitate this interactivity.

LamarMoore:
The detailed html reports now contained interactive graphs which allow navigation to commit history directly by clicking data points.

**To test:**

Run performance the script `make_report.py` which is in `Testing\Performance Tests`. You will need the database file and a pip install of plotly.

Fixes #20044 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Testing/PerformanceTests/analysis.py
+++ b/Testing/PerformanceTests/analysis.py
@@ -17,6 +17,7 @@ import plotly.graph_objs as go
 # This is the date string format as returned by the database
 DATE_STR_FORMAT_MICRO = "%Y-%m-%d %H:%M:%S.%f"
 DATE_STR_FORMAT_NO_MICRO = "%Y-%m-%d %H:%M:%S"
+mantid_addr = "https://github.com/mantidproject/mantid"
 
 
 # ============================================================================================
@@ -135,7 +136,7 @@ def plot_runtime(annotate, saveImage, path, **kwargs):
 
     last_num = kwargs.get('last_num', -1)
 
-    if annotate and saveImage == False:
+    if annotate and not saveImage:
         # retrieve commitids for annotation on the plotly graph
         results = get_results(name, type='', where_clause='')
         sorted = [(res["revision"], res["variables"], res["date"], res) for res in results]
@@ -143,17 +144,17 @@ def plot_runtime(annotate, saveImage, path, **kwargs):
 
         commitids = []
         for (rev, variable, date, res) in sorted:
-            commitids.append("""<a href="https://github.com/mantidproject/mantid/commit/%s">  </a>""" % res["commitid"])
+            commitids.append("""<a href="%s/commit/%s">  </a>""" % (mantid_addr, res["commitid"]))
 
         if last_num > 0:
             commitids = commitids[-last_num:]
 
-        for j in range(0, len(xData), 1):
+        for x, y, text in zip(xData, yData, commitids):
             annotations.append(
                 dict(
-                    x=xData[j],
-                    y=yData[j],
-                    text=commitids[j],
+                    x=x,
+                    y=y,
+                    text=text,
                     showarrow=False,
                     font=dict(family='sans serif', size=10),
                     xanchor='center',
@@ -190,7 +191,7 @@ def plot_runtime(annotate, saveImage, path, **kwargs):
         plt.ylabel(yAxisTitle)
         plt.plot(xData, yData, "-b.")
         plt.ylim(ymin=0)
-        plt.savefig(path + "/" + im_filename)
+        plt.savefig(os.path.join(path, im_filename))
         plt.close()
         return """<img src="%s"/>""" % im_filename
     else:
@@ -201,6 +202,9 @@ def plot_runtime(annotate, saveImage, path, **kwargs):
 default_html_header = """<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd"><html>
 <head><LINK href="report.css" rel="stylesheet" type="text/css"></head>
 """
+default_plotly_header = """<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd"><html>
+    <head><LINK href="report.css" rel="stylesheet" type="text/css"><script src="https://cdn.plot.ly/plotly-latest.min.js"></script></head>
+    """
 
 default_html_footer = """</body></html>"""
 
@@ -257,9 +261,7 @@ def make_environment_html(res):
 # ============================================================================================
 def make_detailed_html_file(basedir, name, fig1, fig2, last_num):
     """ Create a detailed HTML report for the named test """
-    html = """<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd"><html>
-    <head><LINK href="report.css" rel="stylesheet" type="text/css"><script src="https://cdn.plot.ly/plotly-latest.min.js"></script></head>
-    """
+    html = default_plotly_header
     html += """<h1>Detailed report for %s</h1><br>""" % (name)
     html += fig1 + "\n"
     html += fig2 + "\n"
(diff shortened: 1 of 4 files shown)


Testing/PerformanceTests/analysis.py
```diff
@@ -115,147 +119,93 @@ def get_results_matching(results, field, value):
...
+        sorted.sort(reverse=False)
+
+        commitids = []
+        for (rev, variable, date, res) in sorted:
+            commitids.append("""<a href="https://github.com/mantidproject/mantid/commit/%s">  </a>""" % res["commitid"])
+
+        if last_num > 0:
+            commitids = commitids[-last_num:]
+
+        for j in range(0, len(xData), 1):
```
martyngigg: We could replace the range with a `zip`: 
```python
for x, y, text in zip(xData, yData, commitids):
```

Testing/PerformanceTests/analysis.py
```diff
@@ -302,13 +253,16 @@ def make_environment_html(res):
     """ % (res['host'], res['environment'], res['runner'])
     return html
 
-#============================================================================================
+
+# ============================================================================================
 def make_detailed_html_file(basedir, name, fig1, fig2, last_num):
     """ Create a detailed HTML report for the named test """
-    html = default_html_header
+    html = """<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd"><html>
```
martyngigg: We seemed to have defined this header string twice. Can we define this once at the head of the file?

Testing/PerformanceTests/analysis.py
```diff
@@ -384,39 +340,39 @@ def make_detailed_html_file(basedir, name, fig1, fig2, last_num):
     table_rows.reverse()
     for row in table_rows:
         html += row
-#        # Add the row header every 30 entries
-#        count += 1
-#        if count % 30 == 0: html += table_row_header
+    # # Add the row header every 30 entries
+    #        count += 1
```
martyngigg: Can we remove the dead code?

Testing/PerformanceTests/analysis.py
```diff
@@ -384,39 +340,39 @@ def make_detailed_html_file(basedir, name, fig1, fig2, last_num):
...
+    #        if count % 30 == 0: html += table_row_header
 
     # And one more at the end for good measure
     html += table_row_header
     html += "</table>"
 
-    if len(results)> 0:
+    if len(results) > 0:
```
martyngigg: Empty lists evaluate to false so it's more pythonic to use:
```python
if results:
```

Testing/PerformanceTests/analysis.py
```diff
@@ -384,39 +340,39 @@ def make_detailed_html_file(basedir, name, fig1, fig2, last_num):
...
         html += """<h3>Environment</h3>
         %s""" % make_environment_html(results[0])
 
     html += default_html_footer
 
-#    last_date = sorted[-1][1]["date"]
-#    results = get_results(name, type='', get_log=False, where_clause=" date = '%s'" % last_date)
-#    if len(results)>0:
-#        html +=
+    #    last_date = sorted[-1][1]["date"]
```
martyngigg: More dead code?

Testing/PerformanceTests/analysis.py
```diff
@@ -633,7 +577,7 @@ def generate_html_report(path, last_num, x_field='revision'):
     html += get_html_summary_table(test_names)
 
     # -------- Overall success history graphs ------------
-    #if dofigs:
+    # if dofigs:
```
martyngigg: More dead code?

martyngigg (changes requested):
This is really good. :+1: It's a far better experience flicking through the plots and identifying the problem commits.

My comments are mainly style related and quite minor.

Testing/PerformanceTests/analysis.py
```diff
@@ -7,17 +7,20 @@
...
 import numpy as np
 import datetime
 import random
+import plotly.offline as offline
+import plotly.graph_objs as go
 
 # This is the date string format as returned by the database
 DATE_STR_FORMAT_MICRO = "%Y-%m-%d %H:%M:%S.%f"
 DATE_STR_FORMAT_NO_MICRO = "%Y-%m-%d %H:%M:%S"
+mantid_addr = "https://github.com/mantidproject/mantid"
```
martyngigg: Upper case, please :)

Testing/PerformanceTests/analysis.py
```diff
@@ -115,147 +120,96 @@ def get_results_matching(results, field, value):
...
+        plt.ylim(ymin=0)
+        plt.savefig(os.path.join(path, im_filename))
+        plt.close()
+        return """<img src="%s"/>""" % im_filename
+    else:
+        return offline.plot(fig, output_type='div', show_link=False, auto_open=False, include_plotlyjs=False)
 
 
 # The default HTML header
 default_html_header = """<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd"><html>
```
martyngigg: As these are constants can we move them to the top of the module with the others (and uppercase the variable names). This is the more standard way of using these type of "constants".

martyngigg (changes requested):
Looks good, just a couple of more minor style points and it's ready to go I think.

martyngigg (approved):

LamarMoore:
**DO NOT MERGE** System test machines need to be updated with a plotly installation before this can go ahead.

Testing/PerformanceTests/analysis.py
```diff
@@ -115,147 +129,85 @@ def get_results_matching(results, field, value):
...
+
+    annotations = []
+
+    last_num = kwargs.get('last_num', -1)
+
+    if annotate and not saveImage:
+        # retrieve commitids for annotation on the plotly graph
+        results = get_results(name, type='', where_clause='')
+        sorted = [(res["revision"], res["variables"], res["date"], res) for res in results]
+        sorted.sort(reverse=False)
```
MikeHart85: `sorted` is a built-in function for sorting (very similar to `[].sort` but a bit more general) and it would be better to avoid using it as a variable name.  Also, it looks like you're only using `revision`, `variables` and `date` for sorting by, and after that are only interested in `commitid`.

The sorting is probably better left to SQL, and `commitids` could then be generated using list comprehension, which could be implemented like this:

```python
    if annotate and not saveImage:
        # retrieve commitids for annotation on the plotly graph
        results = get_results(name, orderby_clause='ORDER BY revision, variables, date ASC')
        commitids = ["""<a href="%s/commit/%s">  </a>"""
                     % (MANTID_ADDRESS, res["commitid"])
                     for res in results]
```

It would be even better if `get_results` had an additional `field_names` parameter, so you could make it do `SELECT commitid FROM TestRuns WHERE ... ORDER BY ...` instead of `SELECT * FROM ...`, similar to how `get_all_field_values` does it. Just so that there's less data to retrieve and we get only exactly what we need. But that's probably not worthwhile here, the performance difference would be negligible I think.

Testing/PerformanceTests/analysis.py
```diff
@@ -115,147 +129,85 @@ def get_results_matching(results, field, value):
...
+                        mode='lines+markers',
+                        marker=dict(
+                            size='5',
+                            color="blue"
+                        )
+                        )
+
+    annotations = []
+
+    last_num = kwargs.get('last_num', -1)
```
MikeHart85: `None` is generally a better "invalid or absent value" placeholder in Python, since we're not restricted by type.

The check a bit further down would then become:
```python
if last_num is not None:
    commitids = commitids[-last_num:]
```
This would then also support `last_num == 0` for getting all results in spite of providing a last_num, and negative `last_num` values to mean "skip the first (-)X items" instead of "give me the last X items". 

Another alternative with the same effect would be to forgo having a `last_num` local variable and just do:
```python
if 'last_num' in kwargs:
    commitids = commitids[-kwargs['last_num']:]  # or kwargs.get('last_num')
```
Although I find this a bit less nice to read.

MikeHart85 (commented):
A few small comments and suggestions. Otherwise good to go!

martyngigg (commented):
Recent changes have improved things again. Still good from me.

MikeHart85 (approved):
All looks good to me.

Issue #19985, opened 2017-07-11
Corrections - Updating to Monte Carlo Absorption Corrections


louisemccann:
This is an umbrella ticket for all the changes required to update the current Indirect Absorption Corrections workflow to use Monte Carlo absorption.

Completed for Mantid 3.11:
- [x] update IndirectAbsorption algorithms - #18282
- [x] update documentation - #18723
- [x] update Indirect GUI - #18813
- [x] Split into [Shape]MonteCarlo - #19907 
- [x] Create[shape]Corrections - #20184
- [x] Unit tests, documentation etc

To-do in Mantid 3.12
- [x] Merge apply monte carlo with appy paalman pings - #21333
- [ ] Deprecate IndirectAbsorption algorithms
- [x] Split GUI into CalcualtePP, CalculateMC, and Apply - #20471, #21333

brandonhewer:
All necessary components of this umbrella issue have now been complete.

PR #20050, opened 2017-07-19
Absorption Corrections - Add SimpleShapeMonteCarloAbsorption

### LINKED ISSUE #19907: Split AbsorptionCorrections into Create and Apply
Split the Indirect Absorption Corrections algorithms into Calculate and Apply

louisemccann:
part of #19985

Adds a new algorithm that calls MonteCarloAbsorption.

The MonteCarloAbsorption algorithm has several inputs which must be set before the algorithm can be run. This includes setting the beam geometry, sample geometry, and sample material. This algorithm sets these inputs for 3 different sample geometries:
Flat Plate
Cylinder
Annulus
The algorithm has been written in such a way that it should be easy to add additional shapes.
The algorithm can only run with data that has its x-axis units in wavelength.

**To test:**

Create a workspace eg
``` python
ws=CreateSampleWorkspace(Function="Quasielastic",
                         XUnit="Wavelength",
                         XMin=-0.5,
                         XMax=0.5,
                         BinWidth=0.01)
```
Run `SimpleShapeMonteCarloAbsorption`
Try with differnent shapes and options.
Use Chemical Formals: `H2-O` and mass density `1.0`

Fixes #19907. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/SimpleShapeMonteCarloAbsorption.py
```diff
@@ -0,0 +1,311 @@
...
+
+        flatPlateCondition = VisibleWhenProperty('Shape', PropertyCriterion.IsEqualTo, 'FlatPlate')
+        cylinderCondition = VisibleWhenProperty('Shape', PropertyCriterion.IsEqualTo, 'Cylinder')
+        annulusCondition = VisibleWhenProperty('Shape', PropertyCriterion.IsEqualTo, 'Annulus')
+
+        # height is common to all options
+
+        self.declareProperty(name='Height', defaultValue=0.0,
+                             validator=FloatBoundedValidator(0.0),
+                             doc='Height of the sample environment')
```
gvardany: in `cm`?

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/SimpleShapeMonteCarloAbsorption.py
```diff
@@ -0,0 +1,311 @@
...
+
+        # flat plate options
+
+        self.declareProperty(name='Width', defaultValue=0.0,
+                             validator=FloatBoundedValidator(0.0),
+                             doc='Width of the FlatPlate sample environment')
```
gvardany: `cm`

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/SimpleShapeMonteCarloAbsorption.py
```diff
@@ -0,0 +1,311 @@
...
+        self.setPropertySettings('Width', flatPlateCondition)
+
+        self.declareProperty(name='Thickness', defaultValue=0.0,
+                             validator=FloatBoundedValidator(),
+                             doc='Thickness of the FlatPlate sample environment')
```
gvardany: `cm`

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/SimpleShapeMonteCarloAbsorption.py
```diff
@@ -0,0 +1,311 @@
...
+        self.setPropertySettings('Thickness', flatPlateCondition)
+
+        self.declareProperty(name='Center', defaultValue=0.0,
+                             doc='Center of the FlatPlate sample environment')
+        self.setPropertySettings('Center', flatPlateCondition)
```
gvardany: `cm`?

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/SimpleShapeMonteCarloAbsorption.py
```diff
@@ -0,0 +1,311 @@
...
+
+        self.declareProperty(name='Angle', defaultValue=0.0,
+                             doc='Angle of the FlatPlate sample environment')
```
gvardany: degree? Also would be nice to be more specific about the definition of the angle. Is it with respect to the beam or beam normal? Clockwise or anti-clockwise? `-PI` to `PI`, or `0` to `2PI` ?

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/SimpleShapeMonteCarloAbsorption.py
```diff
@@ -0,0 +1,311 @@
...
+        self.setPropertySettings('Angle', flatPlateCondition)
+
+        # cylinder options
+
+        self.declareProperty(name='Radius', defaultValue=0.0,
+                             validator=FloatBoundedValidator(0.0),
+                             doc='Radius of the Cylinder sample environment')
```
gvardany: cm

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/SimpleShapeMonteCarloAbsorption.py
```diff
@@ -0,0 +1,311 @@
...
+        self.setPropertySettings('Radius', cylinderCondition)
+
+        # annulus options
+
+        self.declareProperty(name='OuterRadius', defaultValue=0.0,
+                             validator=FloatBoundedValidator(0.0),
+                             doc='Outer radius of the Annulus sample environment')
```
gvardany: cm

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/SimpleShapeMonteCarloAbsorption.py
```diff
@@ -0,0 +1,311 @@
...
+        self.setPropertySettings('OuterRadius', annulusCondition)
+
+        self.declareProperty(name='InnerRadius', defaultValue=0.0,
+                             validator=FloatBoundedValidator(0.0),
+                             doc='Inner radius of the Annulus sample environment')
```
gvardany: cm

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/SimpleShapeMonteCarloAbsorption.py
```diff
@@ -0,0 +1,311 @@
+from __future__ import (absolute_import, division, print_function)
+
+from mantid.simpleapi import *
+from mantid.api import *
```
gvardany: you don't need the full api and kernel, I would prefer to be specific here

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/SimpleShapeMonteCarloAbsorption.py
gvardany: One could also have a float bounded validator for the angle

Framework/PythonInterface/test/python/plugins/algorithms/WorkflowAlgorithms/SimpleShapeMonteCarloAbsorptionTest.py
```diff
@@ -0,0 +1,186 @@
...
+        self._corrected_flat_plate = corrected
+
+    def _test_corrections_workspace(self, corr_ws):
+        x_unit = corr_ws.getAxis(0).getUnit().unitID()
+        self.assertEquals(x_unit, 'Wavelength')
+
+        y_unit = corr_ws.YUnitLabel()
+        self.assertEquals(y_unit, 'Attenuation factor')
+
+        num_hists = corr_ws.getNumberHistograms()
```
gvardany: I would assert on the block size as well

Framework/PythonInterface/test/python/plugins/algorithms/WorkflowAlgorithms/SimpleShapeMonteCarloAbsorptionTest.py
```diff
@@ -0,0 +1,186 @@
...
+
+import unittest
+
+
+class SimpleShapeMonteCarloAbsorptionTest(unittest.TestCase):
+    def setUp(self):
+        """
+
+        """
+        red_ws = Load('irs26176_graphite002_red.nxs')
```
gvardany: Should we have one more test with ILL reduced data? Just to make sure everything is and stays consistent.

gvardany (changes requested):
The new wrapper does what it is supposed to. Only minor comments about clarifications of the units of the inputs and the tests (see inline).

gvardany:
One more thing I forgot to mention, the `MonteCarloAbsorption` now has an option to approximate the instrument. I was wondering if these properties should also be propagated? This might be needed for some of our direct geometry instruments, but from the other hand the new algorithm you have added is targeted only on indirect ones.

docs/source/algorithms/SimpleShapeMonteCarloAbsorption-v1.rst
```diff
@@ -0,0 +1,69 @@
...
+-----------
+
+Sets up a sample shape, along with the required material properties, and runs
+the :ref:`MonteCarloAbsorption <algm-MonteCarloAbsorption-v1>` algorithm. This algorithm merely
+serves as a simpler interface to define the shape & material of the sample without having
+to resort to the more complex :ref:`CreateSampleShape <algm-CreateSampleShape-v1>` & :ref:`SetSampleMaterial <algm-SetSampleMaterial-v1>`
+algorithms. The computational part is all taken care of by :ref:`MonteCarloAbsorption <algm-MonteCarloAbsorption-v1>`. Please see that
+documentation for more details.
+Currently the shape geometries supported are:
+* Flat Plate
```
gvardany: This did not look right, maybe use `-` instead of `*`

SpencerHowells:
@gvardany we are now doing a Calculate.... algo to do sample & container from non-wavelength input ws. It converts to wavelength for calculating the corrections & then converts back to the input x-unit. I also have a scheme for Elastic S(Q,0) input. The output ws is in the Paalman-Pings format & will be used in the Apply... algo.

louisemccann:
@gvardany It's meant to be a general algorithm rather than a specific indirect one, so long is as the workspace is in units of wavelength this algorithm should work. If it will definitely be used for direct I will put in the instrument interpolation options

gvardany:
Regarding the instrument approximation in MC, I am not sure if they will be definitely needed. So I think  we can ship this as is, and then later if the need arises, we can add the 3 pass-through properties as well.

gvardany (approved):

PR #20055, opened 2017-07-20
Change use of removePeak according to erase-remove idiom

### LINKED ISSUE #19472: Use erase-remove idiom instead of PeaksWorkspace::removePeak(...)
### Expected behavior

Use the [erase-remove idiom](https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom) to remove large numbers of elements from a std::vector. 

### Actual behavior

In several places,  PeaksWorkspace::removePeak(int) is being called within a loop. This is considerably slower (see #17307).

[CentroidPeaks.cpp#L179](https://github.com/mantidproject/mantid/blob/2d0a1d20fd18f13dfef6a657316d9eedd21dac0a/Framework/Crystal/src/CentroidPeaks.cpp#L179)

[CentroidPeaks.cpp#L309](https://github.com/mantidproject/mantid/blob/2d0a1d20fd18f13dfef6a657316d9eedd21dac0a/Framework/Crystal/src/CentroidPeaks.cpp#L309)

[DeleteTableRows.cpp#L46](https://github.com/mantidproject/mantid/blob/2d0a1d20fd18f13dfef6a657316d9eedd21dac0a/Framework/DataHandling/src/DeleteTableRows.cpp#L46)

[DiffPeaksWorkspaces.cpp#L91](https://github.com/mantidproject/mantid/blob/2d0a1d20fd18f13dfef6a657316d9eedd21dac0a/Framework/Crystal/src/DiffPeaksWorkspaces.cpp#L91)

[OptimizeLatticeForCellType.cpp#84](https://github.com/mantidproject/mantid/blob/2d0a1d20fd18f13dfef6a657316d9eedd21dac0a/Framework/Crystal/src/OptimizeLatticeForCellType.cpp#84)

[PeakIntegration.cpp#L112](https://github.com/mantidproject/mantid/blob/546b2eccbaf236fd217ccbd21573f88aa0272103/Framework/Crystal/src/PeakIntegration.cpp#L112)

[SaveHKL.cpp#L600](https://github.com/mantidproject/mantid/blob/546b2eccbaf236fd217ccbd21573f88aa0272103/Framework/Crystal/src/SaveHKL.cpp#L600)

### Platforms affected

all

MatthewBowles:
removePeak was being used inside loops to repeatedly erase peaks from a vector. Implemented a removePeaks method to remove multiple peaks at once and minimise calls to erase, according to the [erase-remove idiom.](https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom)

**To test:**

Code review and check build servers

Fixes #19472

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Crystal/inc/MantidCrystal/CentroidPeaks.h
+++ b/Framework/Crystal/inc/MantidCrystal/CentroidPeaks.h
@@ -38,8 +38,7 @@ private:
   void integrate();
   void integrateEvent();
   int findPixelID(std::string bankName, int col, int row);
-  void removeEdgePeaks(Mantid::DataObjects::PeaksWorkspace_sptr &peakWS,
-                       const int &edge);
+  void removeEdgePeaks(Mantid::DataObjects::PeaksWorkspace *peakWS);
   Geometry::Instrument_const_sptr inst;
 
   /// Input 2D Workspace
--- a/Framework/Crystal/inc/MantidCrystal/CentroidPeaks.h
+++ b/Framework/Crystal/inc/MantidCrystal/CentroidPeaks.h
@@ -38,6 +38,7 @@ private:
   void integrate();
   void integrateEvent();
   int findPixelID(std::string bankName, int col, int row);
+  void removeEdgePeaks(Mantid::DataObjects::PeaksWorkspace_sptr& peakWS, const int& edge);
   Geometry::Instrument_const_sptr inst;
 
   /// Input 2D Workspace
--- a/Framework/Crystal/inc/MantidCrystal/CentroidPeaks.h
+++ b/Framework/Crystal/inc/MantidCrystal/CentroidPeaks.h
@@ -38,7 +38,7 @@ private:
   void integrate();
   void integrateEvent();
   int findPixelID(std::string bankName, int col, int row);
-  void removeEdgePeaks(Mantid::DataObjects::PeaksWorkspace *peakWS);
+  void removeEdgePeaks(Mantid::DataObjects::PeaksWorkspace &peakWS);
   Geometry::Instrument_const_sptr inst;
 
   /// Input 2D Workspace
--- a/Framework/Crystal/inc/MantidCrystal/CentroidPeaks.h
+++ b/Framework/Crystal/inc/MantidCrystal/CentroidPeaks.h
@@ -38,7 +38,8 @@ private:
   void integrate();
   void integrateEvent();
   int findPixelID(std::string bankName, int col, int row);
-  void removeEdgePeaks(Mantid::DataObjects::PeaksWorkspace_sptr& peakWS, const int& edge);
+  void removeEdgePeaks(Mantid::DataObjects::PeaksWorkspace_sptr &peakWS,
+                       const int &edge);
   Geometry::Instrument_const_sptr inst;
 
   /// Input 2D Workspace
--- a/Framework/Crystal/src/CentroidPeaks.cpp
+++ b/Framework/Crystal/src/CentroidPeaks.cpp
@@ -169,17 +169,7 @@ void CentroidPeaks::integrate() {
   }
   PARALLEL_CHECK_INTERUPT_REGION
 
-  for (int i = int(peakWS->getNumberPeaks()) - 1; i >= 0; --i) {
-    // Get a direct ref to that peak.
-    auto &peak = peakWS->getPeak(i);
-    int col = peak.getCol();
-    int row = peak.getRow();
-    std::string bankName = peak.getBankName();
-
-    if (edgePixel(inst, bankName, col, row, Edge)) {
-      peakWS->removePeak(i);
-    }
-  }
+  removeEdgePeaks(peakWS, Edge);
 
   // Save the output
   setProperty("OutPeaksWorkspace", peakWS);
@@ -299,17 +289,8 @@ void CentroidPeaks::integrateEvent() {
   }
   PARALLEL_CHECK_INTERUPT_REGION
 
-  for (int i = int(peakWS->getNumberPeaks()) - 1; i >= 0; --i) {
-    // Get a direct ref to that peak.
-    auto &peak = peakWS->getPeak(i);
-    int col = peak.getCol();
-    int row = peak.getRow();
-    std::string bankName = peak.getBankName();
+  removeEdgePeaks(peakWS, Edge);
 
-    if (edgePixel(inst, bankName, col, row, Edge)) {
-      peakWS->removePeak(i);
-    }
-  }
   // Save the output
   setProperty("OutPeaksWorkspace", peakWS);
 }
@@ -355,6 +336,22 @@ int CentroidPeaks::findPixelID(std::string bankName, int col, int row) {
         boost::dynamic_pointer_cast<const Detector>(component);
     return pixel->getID();
   }
+}
+
+void CentroidPeaks::removeEdgePeaks(Mantid::DataObjects::PeaksWorkspace_sptr& peakWS, const int& edge) {
+    std::vector<int> badPeaks;
+    for (int i = int(peakWS->getNumberPeaks()) - 1; i >= 0; --i) {
+        // Get a direct ref to that peak.
+        auto &peak = peakWS->getPeak(i);
+        int col = peak.getCol();
+        int row = peak.getRow();
+        std::string bankName = peak.getBankName();
+
+        if (edgePixel(inst, bankName, col, row, edge)) {
+            badPeaks.push_back(i);
+        }
+    }
+    peakWS->removePeaks(badPeaks);
 }
 
 } // namespace Mantid
--- a/Framework/Crystal/src/CentroidPeaks.cpp
+++ b/Framework/Crystal/src/CentroidPeaks.cpp
@@ -341,7 +341,7 @@ void CentroidPeaks::removeEdgePeaks(
   int Edge = getProperty("EdgePixels");
   std::vector<int> badPeaks;
   size_t numPeaks = peakWS->getNumberPeaks() - 1;
-  for (size_t i = 0; i < numPeaks;  i++) {
+  for (size_t i = 0; i < numPeaks; i++) {
     // Get a direct ref to that peak.
     const auto &peak = peakWS->getPeak(i);
     int col = peak.getCol();
--- a/Framework/Crystal/src/CentroidPeaks.cpp
+++ b/Framework/Crystal/src/CentroidPeaks.cpp
@@ -168,7 +168,7 @@ void CentroidPeaks::integrate() {
     PARALLEL_END_INTERUPT_REGION
   }
   PARALLEL_CHECK_INTERUPT_REGION
-  removeEdgePeaks(peakWS.get());
+  removeEdgePeaks(*peakWS);
 
   // Save the output
   setProperty("OutPeaksWorkspace", peakWS);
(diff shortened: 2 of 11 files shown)


Framework/Crystal/inc/MantidCrystal/CentroidPeaks.h
```diff
@@ -38,6 +38,8 @@ class DLLExport CentroidPeaks : public API::Algorithm {
   void integrate();
   void integrateEvent();
   int findPixelID(std::string bankName, int col, int row);
+  void removeEdgePeaks(Mantid::DataObjects::PeaksWorkspace_sptr &peakWS,
```
DavidFair: Instead of passing the address of the shared pointer could we just do:
`void removeEdgePeaks(DataObjects::PeaksWorkspace &peaksWS, ...)`
Then the caller just does
`removeEdgePeaks(ws.get());`
Which removes another layer of indirection too

Framework/DataObjects/src/PeaksWorkspace.cpp
```diff
@@ -149,6 +150,28 @@ void PeaksWorkspace::removePeak(const int peakNum) {
   peaks.erase(peaks.begin() + peakNum);
 }
 
+/** Removes multiple peaks
+* @param badPeaks peaks to be removed
+*/
+void PeaksWorkspace::removePeaks(const std::vector<int> &badPeaks) {
+  if (badPeaks.size() == 0)
+    return;
+  int i = 0;
```
DavidFair: Can we move `i` into the for loop as it is only used there
`for (int i = 0; first < end....`

Framework/DataObjects/src/PeaksWorkspace.cpp
```diff
@@ -149,6 +150,28 @@ void PeaksWorkspace::removePeak(const int peakNum) {
...
+  auto first = peaks.begin();
+  auto end = peaks.end();
+  auto result = first;
+  for (; first < end; ++first, ++i) {
+    // if index of peak is not in badPeaks
+    if (std::find(badPeaks.begin(), badPeaks.end(), i) == badPeaks.end()) {
```
DavidFair: As these are integers underneath and the container is arbitrary size could we instead sort then use [std::binary_search](http://en.cppreference.com/w/cpp/algorithm/binary_search)?

This way we do `O(n) + m * O(log N)` operations instead of `m * O(n)` (where m is the number of bad peaks)

Framework/DataObjects/test/PeaksWorkspaceTest.h
```diff
@@ -485,6 +485,28 @@ class PeaksWorkspaceTest : public CxxTest::TestSuite {
...
+    // add peaks
+    Peak p(inst, 1, 3.0);
+    Peak p2(inst, 2, 6.0);
+    Peak p3(inst, 3, 9.0);
+    pw->addPeak(p);
+    pw->addPeak(p2);
+    pw->addPeak(p3);
+
+    // number of peaks = 4, now remove 3
+    std::vector<int> badPeaks;
```
DavidFair: Use `std::vector<int> badPeaks {0, 2, 3};` so we don't have to do 3 push backs vs one op.

DavidFair (changes requested):

Framework/Crystal/src/CentroidPeaks.cpp
```diff
@@ -357,5 +336,23 @@ int CentroidPeaks::findPixelID(std::string bankName, int col, int row) {
   }
 }
 
+void CentroidPeaks::removeEdgePeaks(
+    Mantid::DataObjects::PeaksWorkspace *peakWS) {
+  int Edge = getProperty("EdgePixels");
+  std::vector<int> badPeaks;
+  for (int i = int(peakWS->getNumberPeaks()) - 1; i >= 0; --i) {
```
DavidFair: Use `size_t` instead of int like so:
`for (size_t i = peakWS->getNumberPeaks() - 1; i >= 0; --i){`

Also why does `i` decrement instead of increment? Usually for loops increment hence me missing this the first time.

Finally a point of note - never cast using c style casts in C++:
`int foo = int(bar)` is incorrect as if the types are not safely cast-able the compiler will just "fix it" leading to very subtle bugs. Instead use
`int foo = static_cast<int>(bar);` - the compiler will check if this safe to do at compile time and if not (e.g. trying to cast a int to a class for example) emit an error or warning whereas the other will silently compile.

See here for more details: http://cppblogs.blogspot.co.uk/2013/08/c-style-casts.html

Framework/Crystal/src/CentroidPeaks.cpp
```diff
@@ -357,5 +336,23 @@ int CentroidPeaks::findPixelID(std::string bankName, int col, int row) {
...
+    // Get a direct ref to that peak.
+    auto &peak = peakWS->getPeak(i);
```
DavidFair: Make this a 
`const auto &peak` so that we know its impossible to accidentally change the value as we take it by ref

Framework/Crystal/src/CentroidPeaks.cpp
```diff
@@ -357,5 +336,23 @@ int CentroidPeaks::findPixelID(std::string bankName, int col, int row) {
...
+    int col = peak.getCol();
+    int row = peak.getRow();
+    std::string bankName = peak.getBankName();
```
DavidFair: Take this by const ref instead as the compiler will make a copy of this:
`const std::string &bankName`

Framework/DataObjects/src/PeaksWorkspace.cpp
```diff
@@ -149,6 +150,28 @@ void PeaksWorkspace::removePeak(const int peakNum) {
...
+/** Removes multiple peaks
+* @param badPeaks peaks to be removed
+*/
+void PeaksWorkspace::removePeaks(std::vector<int> &badPeaks) {
+  if (badPeaks.size() == 0)
+    return;
+  auto first = peaks.begin();
+  auto end = peaks.end();
+  auto result = first;
+  std::sort(badPeaks.begin(), badPeaks.end());
```
DavidFair: Sort then get your iterators - whilst the standard guarantees that not mutating the container's objects will preserve iterators it still looks fishy and can throw developers off if that makes sense?

Framework/DataObjects/src/PeaksWorkspace.cpp
```diff
@@ -149,6 +150,28 @@ void PeaksWorkspace::removePeak(const int peakNum) {
   peaks.erase(peaks.begin() + peakNum);
 }
 
+/** Removes multiple peaks
+* @param badPeaks peaks to be removed
```
DavidFair: I would update the documentation to note that we will take ownership of the container as we manipulate it (plus the caller probably doesn't care about it after we finish? )

Framework/DataObjects/src/PeaksWorkspace.cpp
```diff
@@ -149,6 +150,28 @@ void PeaksWorkspace::removePeak(const int peakNum) {
...
+*/
+void PeaksWorkspace::removePeaks(std::vector<int> &badPeaks) {
```
DavidFair: And change this to 
`(std::vector<int> &&badPeaks)`

And the callers to
`removePeaks(std::move(myVec));`

This way we take ownership of the container without making a copy.

Framework/DataObjects/src/PeaksWorkspace.cpp
```diff
@@ -149,6 +150,28 @@ void PeaksWorkspace::removePeak(const int peakNum) {
...
+    return;
+  auto first = peaks.begin();
+  auto end = peaks.end();
+  auto result = first;
+  std::sort(badPeaks.begin(), badPeaks.end());
+  for (int i = 0; first < end; ++first, ++i) {
+    // if index of peak is not in badPeaks
+    if (!std::binary_search(badPeaks.begin(), badPeaks.end(), i)) {
+      // include in result
+      *result = std::move(*first);
```
DavidFair: Could we rename `first` - I keep looking at it thinking we are prepending the first element to the result iterator. Maybe something like `index`?

DavidFair (changes requested):
Some more changes around this PR which I previously did not notice

Framework/Crystal/src/CentroidPeaks.cpp
```diff
@@ -357,5 +336,24 @@ int CentroidPeaks::findPixelID(std::string bankName, int col, int row) {
   }
 }
 
+void CentroidPeaks::removeEdgePeaks(
+    Mantid::DataObjects::PeaksWorkspace *peakWS) {
+  int Edge = getProperty("EdgePixels");
+  std::vector<int> badPeaks;
+  size_t numPeaks = peakWS->getNumberPeaks() - 1;
+  for (size_t i = 0; i < numPeaks; i++) {
```
DavidFair: Changing this to:
```
for (int i = 0; i < static_cast<int>(numPeaks); i++){
```
Will fix the warnings you are seeing on GCC / MSVC

DavidFair (approved):

Framework/DataObjects/src/PeaksWorkspace.cpp
martyngigg: I would slightly disagree here. `removePeaks(std::vector<int> &&badPeaks)` forces every caller to use a `move`, even when that might not be what they want to do - forcing them to copy the badPeaks list first if they want to preserve it for some reason.

How about `removePeaks(std::vector<int> badPeaks)` taking the vector by value? If you keep your call as `removePeaks(std::move(badPeaks))` then it will still just move the original vector but in the case where the list wants to be preserved the caller doesn't move and the compiler copies the vector (as the user would have had to do anyway).

Framework/Crystal/inc/MantidCrystal/CentroidPeaks.h
```diff
@@ -38,6 +38,7 @@ class DLLExport CentroidPeaks : public API::Algorithm {
   void integrate();
   void integrateEvent();
   int findPixelID(std::string bankName, int col, int row);
+  void removeEdgePeaks(Mantid::DataObjects::PeaksWorkspace *peakWS);
```
martyngigg: I would use `removeEdgePeaks(Mantid::DataObjects::PeaksWorkspace &peakWS);` instead and call it as `removeEdgePeaks(*ws)`. That way `removeEdgePeaks` doesn't need to check for a null pointer.

Framework/DataObjects/src/PeaksWorkspace.cpp
```diff
@@ -149,6 +150,28 @@ void PeaksWorkspace::removePeak(const int peakNum) {
   peaks.erase(peaks.begin() + peakNum);
 }
 
+/** Removes multiple peaks
+* @param badPeaks peaks to be removed. Takes ownership of container.
+*/
+void PeaksWorkspace::removePeaks(std::vector<int> &&badPeaks) {
+  if (badPeaks.size() == 0)
```
martyngigg: It is generally better practice and slightly more efficient to use `badPeaks.empty()` rather than checking `size() == 0`.

martyngigg (changes requested):

martyngigg (approved):
Looks good now, thanks.

VickieLynch:
SNS user had this error after running CentroidPeaks with EdgePixels=10.  If I try to open a PeaksWorkspace in MantidPlot after running any of the algorithms that use removePeaks, I get this error and MantidPlot terminates.  I will try to fix in new branch.

PR #20082, opened 2017-07-25
Re #20073: Modified GenericDataProcessor to pass message as QString rather than std::exception.

### LINKED ISSUE #20073: No error message when exception thown from ISIS Reflectometry GUI
### Expected behavior
The appropriate error message would be displayed.

### Actual behavior
In the space where the error message would normally appear there is instead the text `std::exception`.

1. Open the Reflectometry GUI.
2. Press the process button
3. Confirm the action.

### Platforms affected
- Fedora 26
- Ubuntu

edwardb96:
- The reductionError signal has been modified to send a `QString` containing the message from the caught exception rather than a `std::exception` by value which caused object slicing.
- The `GenericDataProcessor` widget (and as a result also the `ReflDataProcessorWidget`) has been modified to use `QString` rather than `std::string`.

**To test:**
1. Open the Reflectometry GUI.
2. Press the process button
3. Confirm the action.
4. Ensure an appropriate error message is shown.

1. Compile and run the unit test targets `CustomInterfacesTest` and `MantidWidgetsTest`.

Fixes #20073

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


MantidQt/MantidWidgets/src/DataProcessorUI/GenericDataProcessorPresenter.cpp
```diff
@@ -472,9 +472,8 @@ void GenericDataProcessorPresenter::endReduction() {
 /**
 Handle reduction error
 */
-void GenericDataProcessorPresenter::reductionError(std::exception ex) {
-
-  m_view->giveUserCritical(ex.what(), "Error");
+void GenericDataProcessorPresenter::reductionError(QString ex) {
+  m_view->giveUserCritical(ex.toStdString(), "Error");
```
martyngigg: I think we should update the *view* class so that `giveUserCritical` can accept a QString so that we avoid the conversion to `std::string`

martyngigg (changes requested):

MantidQt/MantidWidgets/src/DataProcessorUI/QDataProcessorWidget.cpp
```diff
@@ -538,14 +538,25 @@ void QDataProcessorWidget::accept(DataProcessorMainPresenter *mainPresenter) {
...
                                             std::string title) {
 
   QMessageBox::critical(this, QString::fromStdString(title),
-                        QString::fromStdString(prompt), QMessageBox::Ok,
+                        prompt, QMessageBox::Ok,
                         QMessageBox::Ok);
 }
 
+
+/** Shows a critical error dialog
```
martyngigg: We shouldn't leave dead code around. Can we get rid of this?

MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/DataProcessorUI/QDataProcessorWidget.h
```diff
@@ -86,7 +86,7 @@ class EXPORT_OPT_MANTIDQT_MANTIDWIDGETS QDataProcessorWidget
                             const std::string &defaultValue) override;
   bool askUserYesNo(std::string prompt, std::string title) override;
   void giveUserWarning(std::string prompt, std::string title) override;
-  void giveUserCritical(std::string prompt, std::string title) override;
+  void giveUserCritical(QString prompt, std::string title) override;
```
martyngigg: It seems a little inconsistent to me that we accept QString for one argument but the other is a `std::string`.

I think we should take the opportunity to make this consistent across this class. As I assume most of the time these interact with Qt widgets then it feels like we should use `QString` across the class. @AntonPiccardoSelg Any thoughts here?

AntonPiccardoSelg: From what I understand of the DataProcessorUI, the idea is to be only dependent on `Qt` where it is necessary. Hence the interfaces such as `DataProcessorView` (which is the interface for `QDataProcessorWidget`) do not depend on `Qt`. This was also true until recently for the `GenericDataProcessorWidget` (however it gained a `Qt` dependency through the async work; essentially it needs signals and slots).

Previously, the `QDataProcessorWidget` used `std::string` exclusively which was changed in parts to `QString` when Raquel implemented the Python exports, since `SIP` requires `QString`. 

We have two choices:
1. Give up the flexibility of easily using other UI frameworks, but make it use `Qt` structures consistently.
2. Remain with the flexibility, but be terribly inconsistent and make use of `QString::fromStdString` in a lot of places.

I would opt for `1.` since the core element of the `DataProcessorUI` is already tied up with `Qt` and having consistency would be a big improvement in terms of maintainability.

martyngigg: I agree. I think the idea of using other toolkits is nice but it is not really that practical.

Edward, can you make the changes?

edwardb96: Yes.

MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/DataProcessorUI/QDataProcessorWidget.h
```diff
@@ -27,7 +27,7 @@ Copyright &copy; 2016 ISIS Rutherford Appleton Laboratory, NScD Oak Ridge
 National Laboratory & European Spallation Source
 
 This file is part of Mantid.
-
+.what()
```
AntonPiccardoSelg: ?

MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/DataProcessorUI/QDataProcessorWidget.h
```diff
@@ -157,18 +155,19 @@ class EXPORT_OPT_MANTIDQT_MANTIDWIDGETS QDataProcessorWidget
 
 signals:
   void comboProcessInstrument_currentIndexChanged(int index);
-  void runPythonAlgorithm(const QString &pythonCode);
+  void ranPythonAlgorithm(const QString &pythonCode);
 
 public slots:
   void on_comboProcessInstrument_currentIndexChanged(int index);
+  void setModel(QString name) override;
```
AntonPiccardoSelg: Is it worth having this as `const&`?

MantidQt/MantidWidgets/src/DataProcessorUI/DataProcessorGenerateNotebook.cpp
```diff
@@ -49,23 +47,22 @@ specified via the corresponding hinting line edit in the view
 @returns ipython notebook string
 */
 DataProcessorGenerateNotebook::DataProcessorGenerateNotebook(
-    std::string name, const std::string instrument,
+    QString name, const QString instrument,
```
AntonPiccardoSelg: Do you know if there is a reason why we don't just have `const QString&` for all cases?

MantidQt/MantidWidgets/src/DataProcessorUI/DataProcessorGenerateNotebook.cpp
```diff
@@ -189,51 +183,50 @@ std::string plotsString(const std::vector<std::string> &output_ws,
...
   // We can do that from the processing algorithm
   size_t nGroups = processor.numberOfOutputProperties();
 
   // A vector to store the list of groups
-  std::vector<std::string> workspaceList;
+  QStringList workspaceList;
 
   // Now we iterate through groups to get the relevant workspace
-  for (size_t group = 0; group < nGroups; group++) {
+  for (auto group = 0u; group < nGroups; group++) {
```
AntonPiccardoSelg: :+1:, you don't see that very often yet

MantidQt/MantidWidgets/src/DataProcessorUI/DataProcessorTwoLevelTreeManager.cpp
```diff
@@ -333,17 +331,16 @@ void DataProcessorTwoLevelTreeManager::pasteSelected(const std::string &text) {
...
     // Use group where rows in clipboard belong and paste new rows to it
     // Add as many new rows as required
-    for (size_t i = 0; i < lines.size(); ++i) {
-      std::vector<std::string> values;
-      boost::split(values, lines[i], boost::is_any_of("\t"));
+    for (auto i = 0; i < lines.size(); ++i) {
+      auto values = lines[i].split("\t");
 
-      int groupId = boost::lexical_cast<int>(values.front());
+      int groupId = boost::lexical_cast<int>(values.front().toStdString());
```
AntonPiccardoSelg: Would [this](http://doc.qt.io/qt-5/qstring.html#toInt) work here?

MantidQt/MantidWidgets/src/DataProcessorUI/GenericDataProcessorPresenter.cpp
```diff
@@ -36,54 +36,75 @@ using namespace Mantid::Kernel;
...
+                          QString const &value);
+
+void setAlgorithmProperty(IAlgorithm *const alg, std::string const &name,
+                          std::string const &value);
+
+void setAlgorithmProperty(IAlgorithm *const alg, QString const &name,
+                          std::string const &value);
+
+void setAlgorithmProperty(IAlgorithm *const alg, QString const &name,
+                          QString const &value);
```
AntonPiccardoSelg: You could save the declaration with a little reordering below.

MantidQt/MantidWidgets/src/DataProcessorUI/GenericDataProcessorPresenter.cpp
```diff
@@ -538,42 +563,44 @@ void GenericDataProcessorPresenter::postProcessGroup(
...
-  const std::string outputWSName =
+  const QString outputWSName =
       getPostprocessedWorkspaceName(groupData, m_postprocessor.prefix());
 
   // Go through each row and get the input ws names
   for (const auto &row : groupData) {
 
     // The name of the reduced workspace for this row
-    const std::string inputWSName =
+    const QString inputWSName =
```
AntonPiccardoSelg: `auto`

AntonPiccardoSelg (changes requested):
Most of the comments are minor. There are quite a few places (and not all are marked) where plain `auto` instead of QString could be used. Also there are some places where `const QString&` could be used as a function argument. 

The changes here are getting quite large. As discussed earlier, it might have been worth making incremental changes and using QString::toStdString at interface boundaries.

Also be aware that `mantidqtpython` might be affected by your changes.

MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Reflectometry/ReflDataProcessorPresenter.h
```diff
@@ -4,6 +4,7 @@
 #include "MantidQtMantidWidgets/DataProcessorUI/GenericDataProcessorPresenter.h"
 #include "MantidQtMantidWidgets/DataProcessorUI/DataProcessorMainPresenter.h"
 #include "MantidQtMantidWidgets/DataProcessorUI/DataProcessorTreeManager.h"
+#include "MantidAPI/IEventWorkspace_fwd.h"
```
edwardb96: It contains the alias `IEventWorkspace_sptr` which is used as the return value of `retrieveWorkspaceByName`. Which is a chunk of code I extracted from `parseUniform` to make it more readable.

MantidQt/MantidWidgets/src/DataProcessorUI/DataProcessorGenerateNotebook.cpp
edwardb96: In the original code `std::string` was sometimes passed by value and sometimes by `const&` when making the changes I tried not to change which technique was used as I too wasn't sure if there was a reason for it.

It seems to be a contentious issue [here](https://stackoverflow.com/questions/2104328/whats-the-best-way-to-send-qstrings-in-a-function-call) since QString uses copy-on-write internally.

AntonPiccardoSelg: I agree that it might be going to far in this issue. If it was purely a read of the QString, then `const &` would still be better, since you copy the `cow` handler  itself.

MantidQt/MantidWidgets/src/DataProcessorUI/GenericDataProcessorPresenter.cpp
```diff
@@ -1409,20 +1442,18 @@ void GenericDataProcessorPresenter::plotGroup() {
...
 
   // Set of workspaces to plot
-  std::set<std::string> workspaces;
+  std::set<QString> workspaces;
   // Set of workspaces not found in the ADS
-  std::set<std::string> notFound;
+  QSet<QString> notFound;
 
-  const auto items = m_manager->selectedData();
+  auto const items = m_manager->selectedData();
```
edwardb96: That's a good idea. The codebase doesn't seem to be all that consistent with regard to the position of the const. 

I tend to prefer `const` on the right as to me it seems more consistent and easy to remember which part is constant, especially in situations like `int const *` where if you're used to `const` applying to the thing on the right you may end up thinking it's the pointer which is constant rather than the pointee. Additionally methods which are `const` can only have `const` on the right.

> const applies to the thing on the left

is easier to remember than

> const applies to the thing on the left except when it doesn't or can't in which case it applies to the thing on the right.

Regardless it would be good to adopt one style or the other.

MantidQt/MantidWidgets/src/DataProcessorUI/GenericDataProcessorPresenter.cpp
```diff
@@ -1277,26 +1312,28 @@ void GenericDataProcessorPresenter::renameHandle(const std::string &oldName,
...
+    m_mainPresenter->notifyADSChanged(m_workspaceList);
+  }
 }
 
 /**
 Handle ADS replace events
 */
 void GenericDataProcessorPresenter::afterReplaceHandle(
     const std::string &name, Mantid::API::Workspace_sptr workspace) {
+  auto Qname = QString::fromStdString(name);
```
edwardb96: Resolved in 8746c1e

MantidQt/MantidWidgets/src/DataProcessorUI/GenericDataProcessorPresenter.cpp
```diff
@@ -1277,26 +1312,28 @@ void GenericDataProcessorPresenter::renameHandle(const std::string &oldName,
...
   // if a workspace with oldName exists then replace it for the same workspace
   // with newName
-  if (!m_workspaceList.contains(QString::fromStdString(oldName)))
-    return;
-
-  m_workspaceList.remove(QString::fromStdString(oldName));
-  m_workspaceList.insert(QString::fromStdString(newName));
-  m_view->setTableList(m_workspaceList);
-  m_mainPresenter->notifyADSChanged(m_workspaceList);
+  auto QoldName = QString::fromStdString(oldName);
```
edwardb96: Resolved in 8746c1e

MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/DataProcessorUI/QDataProcessorWidget.h
edwardb96: Resolved in 67375c8

MantidQt/MantidWidgets/inc/MantidQtMantidWidgets/DataProcessorUI/QDataProcessorWidget.h
edwardb96: Removed in d9e458d

MantidQt/CustomInterfaces/src/Reflectometry/ReflDataProcessorPresenter.cpp
```diff
@@ -413,18 +430,18 @@ void ReflDataProcessorPresenter::parseLogValue(const std::string &inputStr,
...
+  QString prefix = "TOF_";
+  QString instrument = m_view->getProcessInstrument();
 
   outName = findRunInADS(runNo, prefix, runFound);
   if (!runFound ||
-      AnalysisDataService::Instance().doesExist(outName + "_monitors") ==
+      AnalysisDataService::Instance().doesExist((outName + "_monitors").toStdString()) ==
           false ||
-      AnalysisDataService::Instance().retrieveWS<IEventWorkspace>(outName) ==
+      AnalysisDataService::Instance().retrieveWS<IEventWorkspace>(outName.toStdString()) ==
```
edwardb96: Fixed in f4cef21de0

MantidQt/MantidWidgets/src/DataProcessorUI/GenericDataProcessorPresenter.cpp
edwardb96:
```
$ grep -rn 'auto const' | wc -l
98
$ grep -rn 'const auto' | wc -l
4605
$ grep -rn 'const int' | wc -l
2654
$ grep -rn 'int const' | wc -l
39
```


Running grep on the codebase reveals that although const-on-the-right is used, it is used two orders of magnitude less frequently than const-on-the-left. If we were to adopt an on-the-right style it is likely that we would need tool support of some kind (in which case clang's libTooling might be an option).

MantidQt/CustomInterfaces/src/Reflectometry/ReflDataProcessorPresenter.cpp
```diff
@@ -361,16 +380,14 @@ void ReflDataProcessorPresenter::parseUniform(const std::string &timeSlicing,
...
                                              std::vector<double> &startTimes,
                                              std::vector<double> &stopTimes) {
 
-  std::vector<std::string> timesStr;
-  boost::split(timesStr, timeSlicing, boost::is_any_of(","));
-
+  auto timeStr = timeSlicing.split(",");
   std::vector<double> times;
-  std::transform(timesStr.begin(), timesStr.end(), std::back_inserter(times),
-                 [](const std::string &astr) { return std::stod(astr); });
```
edwardb96: Resolved in 82465ede870fd6d21e276aeac7d35bb90238befd

AntonPiccardoSelg:
@edwardb96 There are some merge conflicts

edwardb96:
@AntonPiccardoSelg I believe they're now fixed.

MantidQt/CustomInterfaces/src/Reflectometry/ReflRunsTabPresenter.cpp
```diff
@@ -278,12 +290,10 @@ void ReflRunsTabPresenter::transfer() {
     return;
   }
 
-  for (auto rowIt = selectedRows.begin(); rowIt != selectedRows.end();
-       ++rowIt) {
-    const int row = *rowIt;
-    const std::string run = m_searchModel->data(m_searchModel->index(row, 0))
-                                .toString()
-                                .toStdString();
+  for (auto &&row : selectedRows) {
```
edwardb96: Interesting reads [here](https://edmundv.home.xs4all.nl/blog/2014/01/28/use-auto-and-and-for-range-based-for-loops/) and [here](https://blog.petrzemek.net/2016/08/17/auto-type-deduction-in-range-based-for-loops/).

To summarise `auto&&` is a good default for generic code since it works all the time. But for non generic code `const auto&` can be considered better since it can be unnecessarily confusing.

AntonPiccardoSelg (approved):
* Check reduction of POLREF with investigation id: 1520249: Looks fine
*  When not providing any input an exception is throw which gives meaningful information

Looks good to me. Can be merged when all tests have passed again

martyngigg (approved):

PR #20091, opened 2017-07-25
Fixed crash when mantid/instrument does not exist

### LINKED ISSUE #19961: Mantid crash on startup after clearing the instruments
MantidPlot should roll back to the default settings if some of the user settings are corrupted.
Particularly, MantidPlot crashes on startup if the instruments download folder is cleared.

### Expected behaviour
Start normally with the default configurations
### Actual behaviour
Segmentation fault
1. Start MantidPlot, set some default instrument
2. Close it
3. Go to your home directory and remove `.mantid/instrument/` folder
4. Start MantidPlot again -> 💣 
(in fact if you close again, and reopen it will suddenly work)
### Platforms affected
All

dtasev:
Description of work.

Fixes the crash when trying to start Mantid and the mantid/instrument folder does not exist by making sure we have created all of the expected directories (if they did not exist).

The reason for the crash previously was because the mantid/instrument folder didn't exist, then inside the call to [`updateFacilities`](https://github.com/mantidproject/mantid/blob/468f7bef5844b3d6a18ac15f4594988ea29a011f/Framework/Kernel/src/ConfigService.cpp#L1806) in [`ConfigServiceImpl`](https://github.com/mantidproject/mantid/blob/468f7bef5844b3d6a18ac15f4594988ea29a011f/Framework/Kernel/src/ConfigService.cpp#L165) when trying to get the facility filename through `getFacilityFilename`, it would throw and be captured inside the `updateFacilities` function, thus silently failing and never reading in the facilities.

The reason for the throw in `getFacilityFilename` is that `getInstrumentDirectories` will only have a single member if the mantid/instrument folder does not exist. However [this `if/else`](https://github.com/mantidproject/mantid/pull/20091/files#diff-040f08e39f367ab14a0acc3afeac2d00R1782) increments it and it goes to the `.end()` iterator, which consequently throws and gets captured in the `updateFacilities` function, failing silently.

The actual crash happens because we never add any facilities into the `m_facilities` vector inside the `updateFacilities` function, so when Mantid tries to create a drop down menu for the user to select a facility, we access an empty vector and crash with a segfault.

---

For Windows the path is:
- `%APPDATA%/mantidproject/mantid/instrument`

For Linux the path is:
- `~/.mantid/instrument`

**To test:**
- Build Mantid, go to the mantid cache folder (from the paths above) and delete the instrument folder. Mantid should start successfully.
- Discuss whether we might want to remove the catch inside [`updateFacilities`](https://github.com/mantidproject/mantid/blob/468f7bef5844b3d6a18ac15f4594988ea29a011f/Framework/Kernel/src/ConfigService.cpp#L1844) so that it doesn't silently capture all errors, even when they cause a segfault afterwards. It might be useful to add more sanity checks (like checking that the facilities are added/the proper files/directories are found) to fail with clearer errors for the developers

Fixes #19961 . 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


DavidFair (changes requested):
Talking to Dimitar about this:
We need to remove the try - catch - log and return logic. Instead make it fail because if we cannot find instruments.xml user assumptions will no longer hold true.

Also we need to refactor the iterator advancement which is bug prone and hard to reason with.

DavidFair (approved):
This resolves the issue with deleting the `mantid` folder in the `AppDir` (or home dir) causing Mantid to crash. :shipit:

martyngigg (approved):

PR #20109, opened 2017-07-27
Improved error message in Rebin when sentinel values not set

### LINKED ISSUE #20000: Rebin fails on Exercise 4 of Introduction to Python in Mantid (Extending Python With Mantid)
The current automated solution - that I've written. Rebin seems to fail when monitors are included - succeeds by excluding monitors. Tested manually and achieved same outcomes.

``` python
from mantid.kernel import *
from mantid.api import *
from mantid.simpleapi import Load, ConvertUnits, Rebin
from numpy import *

class ConvertToEnergy(PythonAlgorithm):

    def PyInit(self):
        self.declareProperty(FileProperty(name="Filename",
                                          defaultValue="",
                                          action=FileAction.Load))
        self.declareProperty(name="Ei", defaultValue=0.0,
                             validator=FloatMandatoryValidator())
        self.declareProperty(FloatArrayProperty(name="BinParams",
                                                validator=FloatArrayLengthValidator(3),
                                                direction=Direction.Input))
        self.declareProperty(WorkspaceProperty(name="OutputWorkspace",
                                               defaultValue="",
                                               direction=Direction.Output))

    def PyExec(self):
        _tmpws = Load(Filename=self.getPropertyValue("Filename")) # Excluding Monitors fixes this
        _tmpws = ConvertUnits(InputWorkspace=_tmpws, Target="DeltaE",
                              EMode="Direct", EFixed=self.getPropertyValue("Ei"))
        _tmpws = Rebin(InputWorkspace=_tmpws, Params=self.getPropertyValue("BinParams"))
        output = WorkspaceFactory.create(_tmpws, NVectors=1)
        output.setX(0, _tmpws.dataX(0))
        spectraSum = np.zeros(_tmpws.getNumberBins())

        for i in range(_tmpws.getNumberBins()):
            sum = 0

            for j in range(_tmpws.getNumberHistograms()):
                sum += _tmpws.dataY(j)[i]

            spectraSum[i] = sum

        _tmpws.delete()
        output.setY(0, spectraSum)
        self.setProperty("OutputWorkspace", OutputWorkspace)

AlgorithmFactory.subscribe(ConvertToEnergy)
```

josephframsay:
**Description of work.**

It was found that, when rebinning a workspace containing monitors after running ConvertUnits on that workspace, an InvalidBinEdgesError was thrown. This is due to cells in the monitor being assigned to `DBL_MIN` as no sensible value could be produced. No more elegant fix could be found, but a more meaningful error message is now given when the above occurs.

**To test:**
```
from mantid.kernel import *
from mantid.api import *
from mantid.simpleapi import Load, ConvertUnits, Rebin
from numpy import *

_tmpws = LoadRaw(Filename="MAR11015.raw", LoadMonitors="Include") # Excluding Monitors fixes this
_tmpws = ConvertUnits(InputWorkspace=_tmpws, Target="DeltaE",
                      EMode="Direct", EFixed="12.9729")
_tmpws = Rebin(InputWorkspace=_tmpws, Params="-11,0.01,11")
```

Running this script used to produce `Negative or zero bin widths not allowed.` - not very meaningful for the user in this case. It should now output `One or more x-values was unusually low (below -1e100). This usually occurs when a monitor spectrum has not been masked after ConvertUnits has been run on the workspace`.

Fixes #20000  

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/HistogramData/src/Rebin.cpp
+++ b/Framework/HistogramData/src/Rebin.cpp
@@ -40,7 +40,11 @@ Histogram rebinCounts(const Histogram &input, const BinEdges &binEdges) {
     auto owidth = xo_high - xo_low;
     auto nwidth = xn_high - xn_low;
 
-    if (owidth <= 0.0 || nwidth <= 0.0)
+	if (xo_high < -1e100 || xo_low < -1e100) {
+		throw InvalidBinEdgesError("One or more x-values was unusually low (below -1e100). This usually occurs when a monitor spectrum has not been masked after ConvertUnits has been run on the workspace");
+	}
+
+	if (owidth <= 0.0 || nwidth <= 0.0)
       throw InvalidBinEdgesError("Negative or zero bin widths not allowed.");
 
     if (xn_high <= xo_low)
--- a/Framework/HistogramData/src/Rebin.cpp
+++ b/Framework/HistogramData/src/Rebin.cpp
@@ -40,7 +40,7 @@ Histogram rebinCounts(const Histogram &input, const BinEdges &binEdges) {
     auto owidth = xo_high - xo_low;
     auto nwidth = xn_high - xn_low;
 
-	if (xo_high < -1e100 || xo_low < -1e100) {
+	if (xo_high == -DBL_MAX || xo_low == -DBL_MAX) {
 		throw InvalidBinEdgesError("One or more x-values was unusually low (below -1e100). This usually occurs when a monitor spectrum has not been masked after ConvertUnits has been run on the workspace");
 	}
 
--- a/Framework/HistogramData/src/Rebin.cpp
+++ b/Framework/HistogramData/src/Rebin.cpp
@@ -41,7 +41,7 @@ Histogram rebinCounts(const Histogram &input, const BinEdges &binEdges) {
     auto nwidth = xn_high - xn_low;
 
     if (owidth <= 0.0 || nwidth <= 0.0){
-      if (xo_high == -DBL_MAX || xo_low == -DBL_MAX) {
+      if (xo_high == -DBL_MAX && xo_low == -DBL_MAX) {
 	throw InvalidBinEdgesError("One or more x-values was unusually low "
 				   "(below -1e100). This usually occurs when a "
 				   "monitor spectrum has not been masked after "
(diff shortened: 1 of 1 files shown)


Framework/HistogramData/src/Rebin.cpp
```diff
@@ -40,6 +40,13 @@ Histogram rebinCounts(const Histogram &input, const BinEdges &binEdges) {
     auto owidth = xo_high - xo_low;
     auto nwidth = xn_high - xn_low;
 
+    if (xo_high < -1e100 || xo_low < -1e100) {
```
AntonPiccardoSelg: In this case is either `xo_high` or `xo_low` equal `DBL_MIN`? If so is there are reason not to `if (xo_high == DBL_MIN || xo_low < DBL_MIN)`?

AntonPiccardoSelg (changes requested):

Framework/HistogramData/src/Rebin.cpp
josephframsay: Neither would equal DBL_MIN - they come out as -DBL_MAX if the error I described occurs. The problem originates in ConvertUnits::getDetectorValues where, for want of a better value, efixed is set to DBL_MIN. The error then propagates through to here. I've made the check more explicit.

AntonPiccardoSelg (approved):
Checked the error message when the test script is run and it looks good to me.

SimonHeybrock (changes requested):
You are adding code to the inner loop in a low level function.
- Have you checked the performance tests? This might be influencing all users of `Rebin`.
- I am not sure that this extra check will not make valid use fail. In particular `xo_low == - DBL_MAX` looks like something that may occur?
- I am not happy with referencing higher-level concepts such as monitors and the `ConvertUnits` algorithm in a low level library that is not aware of those concepts. Is there maybe another solution?

josephframsay:
There's no noticeable effect on performance that we could find. Could a compromise be to make this a special case of the previously existing error check, ie to look for it only once we know that `owidth == 0`? This would have even less effect on performance, and would only fail in the case of real failures. Though I see what you mean about referencing higher-level concepts in a low-level library, the problem only bubbles to the surface at this point, as it's hidden in the internals of other classes until then.

SimonHeybrock:
Yes that sounds better.

SimonHeybrock (approved):

Issue #19433, opened 2017-04-24
Update gSOAP by using the system package or making it an external project


quantumsteve:
Updating gSOAP to the latest version (which supports OpenSSL 1.1) would be better than [backporting changes to the copy in our repository](https://github.com/mantidproject/mantid/pull/19432). This would be a good time to move this code out of our repository. gSOAP is available through dnf, apt and homebrew and could be added to our Windows third-party repo. Alternatively, it could be changed to an external project that downloads and builds the latest version from sourceforge.

PR #20124, opened 2017-07-30
Suppress gSoap warnings

quantumsteve:
Description of work.

This suppresses GCC 7 warnings from gSOAP. Instead of further modifying our local copy, the better long-term solution is to use gSOAP as an external library or external project. See issue #19433

**To test:**

If the builds pass, :squirrel:

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


peterfpeterson (approved):
This doesn't look like a crazy solution. Part of the problem with gSoap is that it is generated code that you have to hand edit stubs from. My understanding is that we didn't include the whole thing, just a subset, but that doesn't make it much better.

Issue #19964, opened 2017-07-05
Cleanup related to recent Instrument-2.0 changes


SimonHeybrock:
1. move `DetectorInfo` init into `InfoComponentVisitor`
2. safe init proceedure such that `DetectorInfo` will be guaranteed to have a valid `ComponentInfo` and vice versa. Could be done by making `InfoComponentVisitor` a factory for both (and forbidding other ways of instantiation -- would make testing hard?)
3. remove `relativePosition` and `relativeRotation`?
4. avoid dynamic casts by `ParComponentFactory` while `InfoComponentVisitor` is walking the tree.
5. avoid some dynamic casts related to `RectangularDetectorPixel`.
6. remove `Detector` caching mechanism from `API::SpectrumInfo` and `API::DetectorInfo`. This was a performance optimization that was relevant when all access such as `l2` and `position` needed the detector, now it is used only be rare direct access via `SpectrumInfo::detector`.
7. Clean up `Component::getPos`, `Component::getRot`, and `Detector::getPos/Rot` as follow up to 1.) and 2.)
8. Run `InfoComponentVisitor` including creation of `ComponentInfo` and `DetectorInfo` in `LoadInstrument` (on *base* instrument) before adding it to the IDS. This avoids duplicated tree parsing if same instrument loaded more than once. This will work since there is no `ParameterMap` and once the instrument is added to the IDS it is guarenteed to be unchanged.
9. Clean up init proceedure in `ExperimentInfo::setInstrument()`. Who should create the visitor? What should its lifetime be? Does `Instrument` need to hold the visitor?

SimonHeybrock:
Regarding 6.):
We cannot remove this directly since `SpectrumInfo::detector()` and `DetectorInfo::detector()` return a reference, i.e., we must store the `shared_ptr<Detector>` somewhere. Probably the way to solve this would be to return a pointer instead. Currently the `detector()` method is still used a lot (mainly to get the detector ID), so we should first reduce the number of uses.

SimonHeybrock:
@OwenArnold Most items look outdated, but some are maybe still relevant...? Should I keep this open, or close?

OwenArnold:
@SimonHeybrock yes, some items done, but would like to keep this open for future reference and in case we get Bhuvan back next year (that kind of thing).

PR #20132, opened 2017-08-01
Move DetectorInfo from API to Geometry

SimonHeybrock:
This is a preparation for other cleanup in #19964. `DetectorInfo` now no longer depends on anything in `API` so this move was possible.

This is being merged into the feature branch for instrument 2.0 and not into `master`.

Notes:
- I kept `DetectorInfoTest` in `API` since it depends strongly on workspaces, which are not available in `GeometryTest`.
- I did not move the Python export to the `Geometry` module. I think this is currently irrelevant and later we might anyway directly export the `Beamline` version of this class.

**To test:**

This is a trivial test and succesful compilation should indicate that the changes are correct, i.e., a short code review should be sufficient.
No issue, no release notes.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


OwenArnold (approved):
Looking at the move of `DetectorInfo` declarations and definitions, if the builds pass then this is fine to merge. Otherwise, no issues introduced to worry about.

PR #20136, opened 2017-08-01
Add hidden column

### LINKED ISSUE #20133: Add hidden options column to the DataProcessorWidget
The DataProcessorWidget works well for a variety of reduction workflows. Some workflows however need to pass information to the the reduction algorihtm which is not intended for the user. Currently this information can be set globally for all rows. On a row-by-row basis the developer can make use of the options column, but this is visible to the user. 

We want to add a `Hidden Options` column to the DataProcessorWidget, which allows the developer to set row-specific settings which should not be presented to the user.

AntonPiccardoSelg:
This PR adds a hidden column as the last column of the DataProcessorWidget. The column allows the developer to add non-use-editable data on a row-by-row basis. Currently the developer can only add data for all rows. 

Fixes #20133 

**To test:**

This is again hard to test for a GUI which does not make use of the this feature. ISIS Reflectometry is currently the only GUI which makes use of this. In principle this should not have any effect on the Reflectometry reduction. Things to check are:

* Does the GUI look the same
* Manual testing can be done like this:
  - Login to the ICAT catalog
  - Open the ISIS Reflectometry interface change instrument to `POLREF`
  - Enter RB number `1520249` and transfer all runs
  - Run the reduction.
  - Compare some of the resulting reduced workspaces with the equivalient output from a version of Mantid which does not contain these changes.

**Release Notes** 
---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


gemmaguest (approved):
Looks good to me :+1: I can't see any differences in the interface. I've checked several workspaces from the given example and they look identical to the results before these changes.

PR #20142, opened 2017-08-01
Fix IMAT IDFs from 2015 and 2016 causing issues due to overlapping IDs

### LINKED ISSUE #19552: IMAT contains duplicate detector IDs
The IMAT [IDF](https://github.com/mantidproject/mantid/blob/master/instrument/IMAT_Definition_2015.xml) tries to reuse detector ids...this is not doing what the author expected. This affects **IMAT_Definition_2015.xml AND IMAT_Definition_2016.xml**. Both should be fixed as part of this.

For example, ID 1 is used twice because it appears once [here](https://github.com/mantidproject/mantid/blob/master/instrument/IMAT_Definition_2015.xml#L103), and then again in the range [here](https://github.com/mantidproject/mantid/blob/master/instrument/IMAT_Definition_2015.xml#L45). The second attempt leaves an parasitic detector - it is part of the full in-memory instrument but cannot actually be referenced since it's ID is invalid. See (`Instrument::markAsDetector` ) if you're interested in seeing how the detector ends up orphaned.

After fixing, much of this can be sanity checked by eye. 

1. ID Lists should contain unique detector id ranges.
1. ID Lists should be used only once within a given IDF.

This is blocking the rollout of Instrument 2.0 features.

dtasev:
Description of work.

The camera has been removed from the definition files. This should hopefully fix any ID collisions.

**To test:**

Fixes #19552. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


OwenArnold:
Before I check this, can you verify that putting an (temporary) else in here https://github.com/mantidproject/mantid/blob/f4143857254a0e82ca2fc92731a6a7cdfaa66d1b/Framework/Geometry/src/Instrument.cpp#L719
like 
```cpp
else {
  throw std::runtime_error("Duplicate detector dropped");
}
```
Then building and running `LoadEmptyInstrument` for all the IDFs changed does not result in an error, and also that the number of detectors reported in the Detector table matches the expectations from the IDF?

DTasev:
@OwenArnold Just a note that I couldn't reproduce the error before. I tried it and all of the IDFs loaded in with the change and I did not get Mantid to throw

OwenArnold:
@DTasev That could be because of the rhs of the `||` unique object check. Delete that out and see if it throws.

OwenArnold (approved):
Duplicates removed. 
Instrument Scientists consulted about changes
Test Data verified with Instrument Scientist.

PR #20145, opened 2017-08-02
Add draft multiple scattering concept page and diagram

### LINKED ISSUE #20072: More detailed description of multiple scat in: MayersSampleCorrection and MultipleScatteringCylinderAbsorption
Explain in http://docs.mantidproject.org/nightly/algorithms/MayersSampleCorrection-v1.html and
http://docs.mantidproject.org/nightly/algorithms/MultipleScatteringCylinderAbsorption-v1.html in more detail how multiple scattering is calculated and how these algorithms do this (subtly) different.

Done as part #11945

jamesquantum:
Multiple scattering concept page (originally mentioned in #11945) added which briefly explains theory as well as tying together current Mantid algorithms aimed at correcting multiple scattering.

Would be helpful also to get a review from another scientists point of view for example Richard Heenan or Spencer Howells.

Hopefully some usage examples will be added eventually as well as a more detailed description of the MayersSampleCorrection and MultipleScatteringCylinderAbsorption.

Fixes #20072


SpencerHowells:
The liquid & amorphous community have been doing this for 40 years. 
The first manual describing it is A K Soper, W S Howells and A C Hannon, 	
ATLAS - Analysis of Time-of-Flight Diffraction Data from Liquid and Amorphous Samples, RAL-89-046. 1989.
A more recent version is A K Soper, GudrunN.
Unfortunately these routines are in Fortran!

martyngigg:
Thanks, Spencer.

@jamesphysics We need to, at minimum, reference the material that Spencer suggests.

jamesphysics:
@peterfpeterson

docs/source/concepts/MultipleScattering.rst
```diff
@@ -0,0 +1,77 @@
...
+
+Mantid solutions
+~~~~~~~~~~~~~~~~
+
+Mayers Sample Correction
+------------------------
+Information about the underlying Mantid algorithm can be found in the 
+:ref:`MayersSampleCorrection <algm-MayersSampleCorrection>` documentation.
+
+Multiple Scattering Cylinder Absorption
```
peterfpeterson: This is a Carpenter style correction

docs/source/concepts/MultipleScattering.rst
```diff
@@ -0,0 +1,77 @@
...
+---------------------------------------
+Documentation for this algorithm can found within the 
+:ref:`MultipleScatteringCylinderAbsorption <algm-MultipleScatteringCylinderAbsorption>` page.
+
+References
```
peterfpeterson: Shouldn't the references be in the individual algorithm pages? While they are all trying to solve the same problem, they do it in a different way. They could be in both places. Either way, see the references at the bottom of [MultipleScatteringCylinderAbsorption](https://github.com/mantidproject/mantid/blob/master/docs/source/algorithms/MultipleScatteringCylinderAbsorption-v1.rst) for an example of how to format references. For ATLAS, I'd make the `RAL-89-046` a link to the actual document.

Anders-Markvardsen:
I have asked Richard Heenan to have a quick read also

RichardHeenan:
In some areas, such as small angle scattering, there may be useful approximations that can be applied that are not present for the more general wide angle scattering case. Again matters may become complicated, as for example small angle scatter followed by incoherent scatter from hydrogen can be more significant in blurring sharp features than double small angle scatter. Also the strongest multiple SANS or SAXS is not necessarily at the smallest Q, though may be most apparent there. Early considerations of multiple small angle scatter were given by for example  J.Schelten & W.Schmatz, J.Appl.Cryst. 13(1980)385-390 and J.R.D.Copley J.Appl.Cryst 21(1988)639-644.

Anders-Markvardsen:
Hi James note your refs looks like this after build to .html:

capture

and 

capture

in both crome and firefox

Anders-Markvardsen (approved):
I think this is very good first draft multiple scat concept page

PR #13961, opened 2015-10-13
Improve performance on LoadIsawPeaks algorithm

MikeHart85:
Fixes #13632.

The [release notes](http://www.mantidproject.org/ReleaseNotes_3_6_Framework_Changes) have been updated.

Identified CompAssembly::getComponentByName() as the cause for slow performance on some files. 

When a file has many peaks that use detector banks with IDs >= 10, an issue with getComponentByName() will cause considerably slow loading times.

Issue mainly addressed by keeping a local cache of pointers to banks to minimize the number of times getComponentByName() needs to be called.

PR #20151, opened 2017-08-02
Precompute Trig functions in Indexing

### LINKED ISSUE #20150: Indexing Utils Performance
A quick run of the profiler while running `FindPeaksUsingLatticeParameterTest` shows that the majority of the time is spent calculating `cos` and `sin` functions. These can be pre-computed as they are based on fixed lattice angles.

`IndexingUtils` is still a mess. It could do with some attention and porting to `Eigen`. This issue will not resolve those major issues.

OwenArnold:
Unnecessary re-computation in UB scanning

**Tester** 
* All tests should pass.
* This should be faster than what we originally had.

fixes #20150


VickieLynch (approved):
Same answer, but faster with TOPAZ data.

Framework/Geometry/inc/MantidGeometry/Crystal/IndexingUtils.h
```diff
@@ -136,9 +136,10 @@ class MANTID_GEOMETRY_DLL IndexingUtils {
                           double req_tolerance, double min_vol);
 
   /// Get the vector in the direction of "c" given other unit cell information
-  static Kernel::V3D Make_c_dir(const Kernel::V3D &a_dir,
-                                const Kernel::V3D &b_dir, double c,
-                                double alpha, double beta, double gamma);
+  static Kernel::V3D makeCDir(const Kernel::V3D &a_dir,
+                              const Kernel::V3D &b_dir, const double &c,
+                              const double &cosAlpha, const double &cosBeta,
+                              const double &cosGamma, const double &sinGamma);
```
OwenArnold: Absolutely none. Yes, i'll make that change.

OwenArnold:
@VickieLynch I wanted to speed up the total system unit test as it's the longest running in the whole framework (FindUBUsingLatticeParametersTest). The majority of the time is now spent loading the TOPAZ peaks file. It would be really nice to replace this with something faster, or some other example, ideally made-in-memory data set. Do you have any ideas on how this could be done?

VickieLynch:
@OwenArnold LoadIsawPeaks was made to run very fast by someone in the UK in the past, but recently it is very slow (because of instrument 2.0 changes?) even for small files.  I would prefer it be improved rather that the test made smaller.

VickieLynch:
@OwenArnold Mike Hart Improve performance on LoadIsawPeaks algorithm #13961

OwenArnold:
Thanks @VickieLynch. I'll take a look to see what's making this slower.

PR #17153, opened 2016-08-05
Rewrite of SCDCalibratePanels

VickieLynch:
Description of work.
SCDCalibratePanels   has been rewritten to calibrate the position and rotations of each panel independently in parallel.  There are options to calibrate the panel size and the L1 for all the panels.
 Only the U of the UB matrix is refined.  There is a script SCDCalibratePanelsResults.py that takes the output of  this algorithm and plots the theoretical vs calculated position of each peak for each panel.  The RMSD in mm is  calculated and printed.

**To test:**
https://www.dropbox.com/s/5eidrhs9ehnfkge/Si2mm_Cubic_F.integrate?dl=0

``` python
LoadIsawPeaks(Filename='Si2mm_Cubic_F.integrate', OutputWorkspace='peaks_ws')
SCDCalibratePanels(PeakWorkspace='peaks_ws', a=5.431, b=5.431, c=5.431, alpha=90, beta=90, gamma=90, DetCalFilename='SCDCalibrate.DetCal', XmlFilename='SCDCalibrate.xml',EdgePixels=20)
```

Fixes #16199 

---
#### Reviewer

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):
##### Code Review
- [x] Is the code of an acceptable quality?
- [x] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)? Is it well structured with small focussed classes/methods/functions?
- [x] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [x] If there are changes in the release notes then do they describe the changes appropriately?
##### Functional Tests
- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
- [x] Has the relevant documentation been added/updated?
- [x] Is user-facing documentation written in a user-friendly manner?
- [ ] Has developer documentation been updated if required?
- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

PR #20161, opened 2017-08-03
Fix performance regressions in ComponentInfo::setRotation and setPosition.

SimonHeybrock:
This is mainly a problem for SCDCalibratePanels. Performance should be
back to normal.

The code is a bit ugly now and could be made nicer if we had spans as in GSL or `boost:hana`. Tried looking into stealing from there, but it is not so simple.

**To test:**

- Code review.
- Maybe compare perormance to master, but probably it is sufficient to do that just once before we merge this feature branch into master.

No issue, no release notes.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Beamline/src/ComponentInfo.cpp
+++ b/Framework/Beamline/src/ComponentInfo.cpp
@@ -184,28 +184,18 @@ void ComponentInfo::setPosition(const size_t componentIndex,
 
   // Optimization: Not using detectorsInSubtree and componentsInSubtree to avoid
   // memory allocations.
-  const auto rangesIndex = compOffsetIndex(componentIndex);
-
   // Optimization: Split loop over detectors and other components.
-  const auto detRange = (*m_detectorRanges)[rangesIndex];
-  auto it = m_assemblySortedDetectorIndices->begin() + detRange.first;
-  auto end = m_assemblySortedDetectorIndices->begin() + detRange.second;
-  if (it != end)
+  const auto range = detectorRangeInSubtree(componentIndex);
+  if (!range.empty())
     failIfScanning();
-  for (; it != end; ++it) {
-    const auto &index = *it;
-    // Optimzation: After failIfScanning() we know that the time index is always
-    // 0 so we use slightly faster access with index pair instead of only
-    // detector index.
+  // Optimization: After failIfScanning() we know that the time index is always
+  // 0 so we use faster access with index pair instead of only detector index.
+  for (const auto &index : range) {
     m_detectorInfo->setPosition({index, 0},
                                 m_detectorInfo->position({index, 0}) + offset);
   }
 
-  const auto compRange = (*m_componentRanges)[rangesIndex];
-  it = m_assemblySortedComponentIndices->begin() + compRange.first;
-  end = m_assemblySortedComponentIndices->begin() + compRange.second;
-  for (; it != end; ++it) {
-    const auto &index = *it;
+  for (const auto &index : componentRangeInSubtree(componentIndex)) {
     size_t offsetIndex = compOffsetIndex(index);
     m_positions.access()[offsetIndex] += offset;
   }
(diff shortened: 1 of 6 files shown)


SimonHeybrock:
Performance of `SCDCalibratePanels` for this example: https://github.com/mantidproject/mantid/pull/17153:

`master`: around 50s (if we include a fix for performance regression introduced with scanning).
`instrument_2_0_1`: 80-90s.
this branch: 45s.

Note that a significant part of this is due to normalization of the quaternions in `ComponentInfo` (as done in `DetectorInfo`). Probably this is affecting the iteration count and thus yields a 20s improvement.

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -83,19 +85,23 @@ ComponentInfo::componentsInSubtree(const size_t componentIndex) const {
...
   const auto rangesIndex = compOffsetIndex(componentIndex);
-  const auto range = (*m_componentRanges)[rangesIndex];
+  const auto detRange = (*m_detectorRanges)[rangesIndex];
+  const auto compRange = (*m_componentRanges)[rangesIndex];
 
   // Extract as a block
-  return std::vector<size_t>(
-      m_assemblySortedComponentIndices->begin() + range.first,
-      m_assemblySortedComponentIndices->begin() + range.second);
+  std::vector<size_t> indices(
```
OwenArnold: This looks more complex. Not sure how this would be quicker either since the `m_assemblySortedComponentIndices` has ALL component indices already?

Framework/Geometry/src/Instrument/InstrumentVisitor.cpp
```diff
@@ -217,7 +216,6 @@ size_t InstrumentVisitor::registerDetector(const IDetector &detector) {
     (*m_componentIdToIndexMap)[detector.getComponentID()] = detectorIndex;
     (*m_componentIds)[detectorIndex] = detector.getComponentID();
     m_assemblySortedDetectorIndices->push_back(detectorIndex);
-    m_assemblySortedComponentIndices->push_back(detectorIndex);
```
OwenArnold: I see (from previous comment). I guess this would reduce memory overhead.

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -170,22 +175,39 @@ ComponentInfo::relativeRotation(const size_t componentIndex) const {
...
-      m_positions.access()[offsetIndex] += offset;
-    }
+
+  // Optimization: Not using detectorsInSubtree and componentsInSubtree to avoid
+  // memory allocations.
+  const auto rangesIndex = compOffsetIndex(componentIndex);
+
+  // Optimization: Split loop over detectors and other components.
+  const auto detRange = (*m_detectorRanges)[rangesIndex];
+  auto it = m_assemblySortedDetectorIndices->begin() + detRange.first;
```
OwenArnold: I can see how this is faster. My main issue here is what it's done to readability. For both `setPosition` and `setRotation` can you add more comments to each block section. 

We also discussed extracting methods that could return the begin/end iterators, that might improve things slightly.

OwenArnold (changes requested):
Good work improving on `master` performance! Some minor tidy-up requested.

Framework/Beamline/src/ComponentInfo.cpp
OwenArnold: Ignore this.... I didn't realise you modified the `InstrumentVisitor` when I was reading this.

Framework/Geometry/src/Instrument/InstrumentVisitor.cpp
OwenArnold: There should probably be a comment somewhere what the meaning of this is now. Maybe the doxygen in the header would be ok.

SimonHeybrock:
@OwenArnold I think this should be adressing your comments.

OwenArnold (approved):

OwenArnold:
@SimonHeybrock. These changes are ideal. Will merge if builds pass.

SimonHeybrock:
@OwenArnold I was considering replacing `(detector|componentsInSubtree` by the variants returning a `Range`, but maybe we leave it like this for now (keep the new variant `private`)?

OwenArnold:
@SimonHeybrock  sounds fine to me. If we find we have a use case for it we can change the access.

PR #20170, opened 2017-08-04
More detailed description of Mayers algorithm

jamesquantum:
Hopefully now gives a better description of what the algorithm does and where certain terms are coming from

#20072


Anders-Markvardsen (approved):
Read it again and it now reads clear to me (including 'fixed number of second order scattering events' which is clear now looked more closely at the input parameters)

martyngigg:
@jamesphysics This is a great improvement to this page. My only minor comment is that the `[1]__` doesn't appear as a link when I build it. I think the same is true in the `MultipleScatteringCylinderAbsortption` too.

Could you look at that?

Anders-Markvardsen:
thanks @martyngigg just noticed this all well for #20145

jamesphysics:
thanks for noticing this, should be fixed now on all three documents

martyngigg:
Looks good. Thanks.

martyngigg (approved):

martyngigg:
Windows failure is unrelated.

PR #20171, opened 2017-08-04
Instrument 2.0.1 - Instrument 2 with ComponentInfo

OwenArnold:
The features here have been developed against a long-running branch `instrument_2_0_1`. They have been incrementally tested reviewed and merged in the form of feature branches into `instrument_2_0_1`. The branch is now ready for merging into master.

Fixes:

#19278 
#20002 
#20163

Main features addressed here are:

* Expand `ComponentInfo` for storing rotations and positions
* Replaced Mantid's component rotation and translation functionality to use `ComponentInfo`
* Simplified `Detector` and `Component` regarding parameter map usage
* Eliminated `DetectorInfo`'s compatibility modes around `Instrument 1.0` for setting positions and rotations for non-detectors and navigating the instrument tree.
* Used `InstrumentVisitor` mechanism to parse `Instrument 1.0` for both `ComponentInfo` and `DetectorInfo`
* Optimised Performance using `SCDCalibratePanels` as a known benchmark.

Authored by Owen Arnold and @SimonHeybrock 

Fixes a number of features that have been developed against the `instrument_2_0_1` branch.

**Release Notes** 

Most changes internal. We will modify the functional/performance section of the release notes when we have a better indication of performance improvements across the whole framework.

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)?
- [ ] Are the unit tests small and test the class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####
As an internal change, unit system-tests provide a good indication that this is functionally correct

##### Performance Tests #####
Please pay attention to performance tests against current master. Any facility specific edge-cases should be tested, especially where load performance, and read-write performance is known to be critical.

Does everything look good? Mark the review as **Approve**. A member of `@mantidproject/gatekeepers` will take care of it.

PR #20176, opened 2017-08-04
Geometry::ComponentInfo shapes

### LINKED ISSUE #20175: Geometry::ComponentInfo should have Object/Shape
Required for visualisation and solid angle operations. As an intermediate step, Objects will be taken from `Instrument1` and  held as a vector in `Geometry::ComponentInfo`.

OwenArnold:
**WIP DO NOT MERGE THIS** -- The PR should go against master or another long-term dev branch such as `instrument_2_0_2`. 

blocked by #20171

fixes #20175


Framework/Geometry/inc/MantidGeometry/Instrument/ComponentInfo.h
```diff
@@ -54,19 +54,24 @@ class MANTID_GEOMETRY_DLL ComponentInfo {
...
-  boost::shared_ptr<const std::vector<Mantid::Geometry::IComponent *>>
-      m_componentIds;
+  boost::shared_ptr<const std::vector<Geometry::IComponent *>> m_componentIds;
   /// Map of component ids to indexes
   boost::shared_ptr<const std::unordered_map<Geometry::IComponent *, size_t>>
       m_compIDToIndex;
 
+  /// Shapes for each component
+  boost::shared_ptr<std::vector<boost::shared_ptr<const Geometry::Object>>>
+      m_shapes;
```
SimonHeybrock: This looks like it will be pointing to the shapes in the old instrument tree? Maybe it would make sense to store all unique shapes in `ComponentInfo` and have raw pointers or indices into a vector of shapes here to avoid the `shared_ptr` overhead?

But I suppose having such a vector of unique shapes is not so trivial/cheap to set up?

SimonHeybrock (commented):
Looks ok so far. As far as I understand `ObjComponent::shape()` simply returns a pointer, so extracting the shapes is cheap?

Framework/Geometry/inc/MantidGeometry/Instrument/ComponentInfo.h
OwenArnold: @SimonHeybrock Yes, I did think about storing the unique shapes separately and having raw pointers to those unique shapes. But I I don't know if the memory overhead, would really be worth the additional code + lifetime management issues. I think this could be an optimisation to consider when or if needed.

LamarMoore: @OwenArnold If you have instruments composed of repeating pixels, for example those defined using rectangular pixels, only one shape needs to be stored. Every other pixel is a copy of the original in another position. One only needs to know extents/relative rotations. Could that be a more compact/efficient way for some detector types. (Most instruments I have seen use repetition in the definition).

PR #20181, opened 2017-08-07
Make background culling optional

### LINKED ISSUE #20174: Make one percent culling optional for peak integratation
When `IntegratePeaksMD` is run it will calculate the integrated signal in a peak region and in a background region. It will then subtract the background region from the peak region. This has resulted in poor performance for `WISH`. The reason is not the general approach but the culling of the background data which has been introduced serveral months ago. This culling is not appropriate for `WISH`, hence we should make it optional.

AntonPiccardoSelg:
Fixes #20174

The algorithm `IntegratePeaksMD` sums the counts around a peak and separately the counts in a shell-like volume in the background around the peak. It then subtracts the integrated background from the integrated peak. Since the peak can spill into the background region, the top 1% of `MDBox`es in the background region are discarded. This has been requested by `TOPAZ`. For `WISH` this has become a problem since discarding `1%` of the background skews the SNR to values that make us believe we are dealing with a peak when we are actually dealing with background. Therefore the `1%` culling is being made optional in this PR. 

**To test:**

Ensure that the code makes sense. To see the differences you would have to test this on NoMachine. This might be not easily possible. The `WISH` scientists have had a look at this functionality on NoMachine though. 

**Release Notes** 

See [here](https://github.com/mantidproject/mantid/pull/20181/files#diff-d43ebaec89d7bc36364c8bf4c9f97673)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/inc/MantidAPI/IMDNode.h
+++ b/Framework/API/inc/MantidAPI/IMDNode.h
@@ -204,12 +204,14 @@ public:
   * @param signal [out] :: set to the integrated signal
   * @param errorSquared [out] :: set to the integrated squared error.
   * @param innerRadiusSquared :: radius^2 of inner background
+  * @param useOnePercentBackgroundCorrection :: if one percent correction should be applied to background.
    */
   virtual void
   integrateSphere(Mantid::API::CoordTransform &radiusTransform,
                   const coord_t radiusSquared, signal_t &signal,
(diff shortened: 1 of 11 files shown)


VickieLynch (approved):
Like that the default is what we have currently.  Seems to work well with true/false

PR #20186, opened 2017-08-07
Multiple scattering cylinder absorption documentation update

jamesquantum:
Added some more description to the documentation page and a couple of comments to the code.

The documentation page now has a brief explanation of how the algorithm calculates absorption but still lacking an explanation of multiple scattering. https://github.com/mantidproject/mantid/blob/master/Framework/Algorithms/src/MultipleScatteringCylinderAbsorption.cpp#L292 Seems to correspond to the calculation of the multiple scattering factor described in Mildner et al 1973 equation (1) but the exact origin of the constants COEFF4 and COEFF5 seem to have been lost in translation from  Java to c++. My guess is that it is a second order power series in (muR) but it is unclear how the precise coefficients were arrived at or what assumptions they are based on

#20072


peterfpeterson:
Please add a comment in `MultipleScatteringCylinderAbsorption.cpp` that `static const double C[]` is Chebyshev polynomials and copied directly from Carpenter 1969 table 1. The name of the constant could be changed to `CHEBYSHEV` if you want to help improve the code.

peterfpeterson (approved):
LGTM

martyngigg:
:+1:

PR #20197, opened 2017-08-08
Make one percent culling in IntegrateEllipsoids and IntegrateEllipsoidsTwoStep optional

### LINKED ISSUE #20195: Make one percent culling optional for peak integratation
When IntegreateEllipsoids is run it will calculate the integrated signal in a peak region and in a background region. It will then subtract the background region from the peak region. This has resulted in poor performance for WISH. This culling is not appropriate for WISH, hence we should make it optional.

AntonPiccardoSelg:
Fixes #20195 

The algorithms `IntegrateEllipsoids` and `IntegrateEllipsoidsTwoStep` sum the counts around a ellipsoidal peak and separately the counts in a ellipsoidal-shell-like volume in the background around the peak. They  then subtracts the integrated background from the integrated peak. Since the peak can spill into the background region, the top 1% of `MDBox`es in the background region are discarded. T For `WISH` this has become a problem since discarding `1%` of the background skews the SNR to values that make us believe we are dealing with a peak when we are actually dealing with background. Therefore the `1%` culling is being made optional in this PR. 

**To test:**

Ensure that the code makes sense. To see the differences you would have to test this on NoMachine. This might be not easily possible. 

**Release Notes** 

See [here]()

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


VickieLynch (approved):
Default for new property gives same results as before.  Setting to false results in approximately the same intensities.  Works well

PR #20215, opened 2017-08-09
Fix MDEventInserter use of run_number instead of run_index

rosswhitfield:
@OwenArnold It looks like you wrote MDEventInserter so tell me if I'm wrong.

An MDEvent has the run index (index into the vector of ExperimentInfo of the workspace) not the run number in the object, see [here](https://github.com/mantidproject/mantid/blob/master/Framework/DataObjects/inc/MantidDataObjects/MDEvent.h#L19-L20)

`insertMDEvent` incorrectly described this property as the run number which has resulted in some incorrect usage of MDEvents. This fixes `insertMDEvent`.

After this is merged the remaining incorrect usages should be fixed, I will create issues for the following, from what I see they are
```
ConvertCWSDExpToMomentum
ConvertSpiceDataToRealSpace
ConvertCWSDMDtoHKL
IntegratePeaksCWSD
ImportMDEventWorkspace
```

**To test:**
 * Code review should suffice

Does not need to be in the release notes, internal changes only

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/DataObjects/inc/MantidDataObjects/MDEventInserter.h
+++ b/Framework/DataObjects/inc/MantidDataObjects/MDEventInserter.h
@@ -63,15 +63,15 @@ public:
   determined internally using type information on the MDEventType.
   @param signal : intensity
   @param errorSQ : squared value of the error
-  @param runno : run number
+  @param runindex : run index (index into the vector of ExperimentInfo)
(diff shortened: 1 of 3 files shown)


NickDraper (approved):
Changes simple and small

PR #20221, opened 2017-08-10
Fix wrap_pyui.py when building for Python3

### LINKED ISSUE #20220: wrap_pyui.py fails when compiling for Python3
`buildconfig/wrap_pyui.py` script fails when the cmake variable `PYTHON_EXECUTABLE=/.../python3`, i.e. one is building for Python3. This is due incompatibility between non-ascii characters (e.g. Å) in some of the python files, and the default encoding in Python2's  `open()`.

The issues popped up on Ubuntu 14.04. Might affect other platforms as well.

soininen:
The call to `open()` in `buildconfig/wrap_pyui.py` script failed with Python3 as some of the opened files would in fact be in UTF-8 but `open()` would treat them as 7bit ASCII or whatever. This PR fixes building with Python3.

**To test:**

You need to have a GNU/Linux system with Python3 installed.
1. Check that the builds pass. This should take care of Python2 part and serve as a reference.
2. `rm -fr *` your build directory. 
3. Your build directory is now gone.
4. Run cmake with `-DPYTHON_EXECUTABLE=/usr/bin/python3`.
5. Build all. There shouldn't be any problems with Python.

A code review would also be a plus.

Fixes #20220. 

**Release Notes** 

This was an internal change, affecting only the rare case of experimental Python3 builds. *Does not need to be in release notes.*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


martyngigg (approved):
Looks good. The files are generated with Python 3 and contain the expected `#pylint` line

PR #20222, opened 2017-08-10
Pre-processing for FFT

### LINKED ISSUE #20216: FFT padding and apodization
The muon group would like to improve FFT data. The two most common ways are to use padding (add zeros at the end of the time domain function prior to the transform) and apodization (see http://mathworld.wolfram.com/ApodizationFunction.html). This should improve the quality of FFT data.

AnthonyLim23:
The muon group wanted to be able to do padding (adding zeros after the data) and apply apodization functions to the data. 

To test:
Load some data and then:

1. Try adding padding (values greater than 0)
2. Try the using the apodization functions (note some will cause the data to become larger at large time scales).
3. Try a combination of the two above. 

Fixes #20216. 

**Release Notes** 

Added to release notes.
---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Algorithms/src/FFTPreProcessing.cpp
```diff
@@ -0,0 +1,228 @@
...
+    spectra = std::vector<int>(numSpectra);
+    std::iota(spectra.begin(), spectra.end(), 0);
+  }
+
+  Progress prog(this, 0.0, 1.0, numSpectra + spectra.size());
+  if (inputWS != outputWS) {
+
+    // Copy all the Y and E data
+    PARALLEL_FOR_IF(Kernel::threadSafe(*inputWS, *outputWS))
+    for (int64_t i = 0; i < int64_t(numSpectra); ++i) {
```
mantid-roman: Does it make sense to keep unprocessed spectra? 
* If spectra are padded they get larger while the unprocessed ones have their old size. This is wrong in a `MatrixWorkspace`
* If there is no padding but the spectra are histograms. The processed ones become point data but the rest are histograms.

Framework/Algorithms/src/FFTPreProcessing.cpp
```diff
@@ -0,0 +1,228 @@
...
+  int specLength = static_cast<int>(spectra.size());
+  std::vector<double> norm(specLength, 0.0);
+  PARALLEL_FOR_IF(Kernel::threadSafe(*inputWS, *outputWS))
+  for (int i = 0; i < specLength; ++i) {
+    PARALLEL_START_INTERUPT_REGION
+    const auto specNum = static_cast<size_t>(spectra[i]);
+
+    if (spectra[i] > static_cast<int>(numSpectra)) {
+      g_log.error("The spectral index " + std::to_string(spectra[i]) +
+                  " is greater than the number of spectra!");
```
mantid-roman: No need for logging. It's done by the base class.

Framework/Algorithms/src/FFTPreProcessing.cpp
```diff
@@ -0,0 +1,228 @@
...
+  } else if (method == "Connes") {
+    return connes;
+  } else if (method == "Cosine") {
+    return cosine;
+  } else if (method == "Gaussian") {
+    return gaussian;
+  } else if (method == "Welch") {
+    return welch;
+  }
+  return none;
```
mantid-roman: An exception is preferable here instead of silently returning none.

Framework/Algorithms/src/FFTPreProcessing.cpp
```diff
@@ -0,0 +1,228 @@
...
+* Applies the appodization function to the data.
+* @param histogram :: [input] Input histogram.
+* @param function :: [input] the apodization function 
+* @returns :: Histogram of the apodized data
+*/
+HistogramData::Histogram FFTPreProcessing::applyApodizationFunction(
+    const HistogramData::Histogram &histogram, const double decayConstant,
+    fptr function) {
+  HistogramData::Histogram result(histogram);
+  result.points();
```
mantid-roman: Is it necessary to convert to points? Can we keep the original format?

Framework/Algorithms/src/FFTPreProcessing.cpp
```diff
@@ -0,0 +1,228 @@
...
+HistogramData::Histogram
+FFTPreProcessing::addPadding(const HistogramData::Histogram &histogram,
+                             const int padding) {
+
+  HistogramData::Histogram result(histogram);
+  // make sure point data
+  result.points();
+  auto &xData = result.mutableX();
+  auto &yData = result.mutableY();
+  auto &eData = result.mutableE();
```
mantid-roman: Do they need to be mutable?

Framework/Algorithms/src/FFTPreProcessing.cpp
```diff
@@ -0,0 +1,228 @@
...
+  for (int j = 0; j < padding; j++) {
+
+    for (size_t i = 0; i < yData.size(); ++i) {
+      newXData.push_back(xValue);
+      newYData.push_back(0.0);
+      newEData.push_back(0.0);
+      xValue += dx;
+    }
+  }
+
```
mantid-roman: Can the `std` algorithms be used to move the data?

docs/source/algorithms/FFTPreProcessing-v1.rst
```diff
@@ -0,0 +1,90 @@
...
+
+
+Description
+-----------
+
+This algorithm can prepare data for 
+  :ref:`FFT <algm-FFT>` by applying padding and/or applying an apodization function. 
+
+Padding is when the input data is extended by adding extra measurments of zero at regular intervals after the ned of the input data. 
+
```
mantid-roman: `ned` -> `end`? Could you explain here what padding is for?

docs/source/algorithms/FFTPreProcessing-v1.rst
```diff
@@ -0,0 +1,90 @@
...
+.. alias::
+
+.. properties::
+
+
+Description
+-----------
+
+This algorithm can prepare data for 
+  :ref:`FFT <algm-FFT>` by applying padding and/or applying an apodization function.
```
mantid-roman: This padding doesn't work for FFT. RealFFT only.

docs/source/algorithms/FFTPreProcessing-v1.rst
```diff
@@ -0,0 +1,90 @@
...
+found at large time scales. The current implementation includes the following functions:
+
+- None.
+- Bartlett.
+- Lorentz.
+- Connes.
+- Cosine.
+- Gaussian.
+- Welch.
+
```
mantid-roman: I don't think many people know what they are. Math expressions could be used here.
Also it would be helpful if the meaning of `DecayConstant` is explained for each function.

mantid-roman (changes requested):

Framework/Algorithms/src/FFTPreProcessing.cpp
AnthonyLim23: good point. I will get rid of the unprocessed spectra

Framework/Algorithms/src/FFTPreProcessing.cpp
AnthonyLim23: so could this just be `error(....)`?

Framework/Algorithms/src/FFTPreProcessing.cpp
AnthonyLim23: ok

Framework/Algorithms/src/FFTPreProcessing.cpp
AnthonyLim23: I did this to ensure that the number of x and y data points where the same. I could convert it back at the end. I think it may work if I code it with the move for the original data.

Framework/Algorithms/src/FFTPreProcessing.cpp
AnthonyLim23: no, you are right

Framework/Algorithms/src/FFTPreProcessing.cpp
mantid-roman: When you throw from `exec` `Algorithm::execute` catches it and logs. Simply delete the lines.

Framework/Algorithms/src/FFTPreProcessing.cpp
mantid-roman: This is what other algorithms do. But if you convert to points this should be mentioned in the docs.

AnthonyLim23: I will convert it back at the end

Framework/Algorithms/src/FFTPreProcessing.cpp
```diff
@@ -0,0 +1,250 @@
...
+  auto &eData = result.e();
+  // assume approx evenly spaced
+  double dx = xData[1] - xData[0];
+  std::vector<double> newXData;
+  std::vector<double> newYData;
+  std::vector<double> newEData;
+
+  if (getProperty("NegativePadding")) {
+	  double xValue = xData[0] - dx*yData.size()*padding;
+	  for (int j = 0; j < padding; j++) {
```
mantid-roman: Sorry, forgot to mention that when padding on both ends each end gets half of the one-sided padding. So when `padding == 1` the size of the data doubles in both cases.

docs/source/algorithms/FFTPreProcessing-v1.rst
```diff
@@ -0,0 +1,93 @@
...
+This algorithm can prepare data for 
+  :ref:`FFT <algm-FFT>` by applying padding and/or applying an apodization function. 
+
+Padding is when the input data is extended by adding extra measurments of zero at regular intervals. For real data this is only done after the end of the input data. However, for complex data the padding should be shared between the start and end of the input data.
+
+`Apodization functions <http://mathworld.wolfram.com/ApodizationFunction.html>`_ can be used to remove data with large errors. These are usualy
+found at large time scales. These take a decay constant 
+(:math:`\tau` ) that determines the rate at which the data goes to zero. 
+The time the function is evaluated at is denoted by :math:`R`.
```
mantid-roman: `R`?

AnthonyLim23: oops.

docs/source/algorithms/FFTPreProcessing-v1.rst
```diff
@@ -0,0 +1,93 @@
...
+The current implementation includes the following functions:
+
+- None.
+- Bartlett :math:`1-\frac{\abs({t}}{\tau}`.
+- Lorentz :math:`\exp\left(-\frac{t}{\tau}\right)`.
+- Connes :math:`\left(1-\frac{t^2}{\tau^2}\right)^2`.
+- Cosine :math:`\cos\left(\frac{\pi t}{2 tau}\right)`.
+- Gaussian :math:`\exp\left(-\frac{t^2}{2\tau^2}\right)`.
+- Welch :math:`1-\frac{t^2}{\tau^2}`.
+
```
mantid-roman: I just checked the wolfram link. The decay constant must only be specified for Lorentz and Gaussian. For the other functions it is equal to the x-range.

Framework/Algorithms/src/FFTPreProcessing.cpp
```diff
@@ -195,40 +195,20 @@ FFTPreProcessing::addPadding(const HistogramData::Histogram &histogram,
   auto &eData = result.e();
   // assume approx evenly spaced
   double dx = xData[1] - xData[0];
-  std::vector<double> newXData;
-  std::vector<double> newYData;
-  std::vector<double> newEData;
-  size_t amountOfPadding = yData.size();
+  std::vector<double> newXData(yData.size()*(1+padding),0.0);
+  std::vector<double> newYData(yData.size()*(1+padding),0.0);
+  std::vector<double> newEData(yData.size()*(1+padding),0.0);
```
mantid-roman: Can the size of the new vectors be calculated once and stored in a const?

Framework/Algorithms/src/FFTPreProcessing.cpp
```diff
@@ -195,40 +195,20 @@ FFTPreProcessing::addPadding(const HistogramData::Histogram &histogram,
...
+  double x = xData[xData.size()-1];
```
mantid-roman: `xData.back()` would be clearer.

Framework/Algorithms/src/FFTPreProcessing.cpp
```diff
@@ -195,40 +195,20 @@ FFTPreProcessing::addPadding(const HistogramData::Histogram &histogram,
...
+  int offset = 0;
```
mantid-roman: `size_t` type is preferable here.

Framework/Algorithms/src/FFTPreProcessing.cpp
```diff
@@ -195,40 +195,20 @@ FFTPreProcessing::addPadding(const HistogramData::Histogram &histogram,
...
-
-    for (size_t i = 0; i < amountOfPadding; ++i) {
-      newXData.push_back(xValue);
-      newYData.push_back(0.0);
-      newEData.push_back(0.0);
-      xValue += dx;
-    }
-  }
-
+          offset=int(std::floor(0.5*double(yData.size())));
```
mantid-roman: Can we use integer arithmetic here?

Could you add a comment explaining briefly what is achieved by a non-zero offset?

Framework/Algorithms/src/FFTPreProcessing.cpp
```diff
@@ -195,40 +195,20 @@ FFTPreProcessing::addPadding(const HistogramData::Histogram &histogram,
...
+          x = xData[0] - dx * (1.+double(offset));
+  } 
+  std::generate(newXData.begin(),newXData.end(),[&x,&dx]{return x+=dx;});
+  std::copy(xData.begin(),xData.end(),newXData.begin()+offset);
+  std::copy_n(yData.begin(),yData.size(),newYData.begin()+offset);
```
mantid-roman: I don't see a reason to use different versions of `copy` here.

mantid-roman (changes requested):

Framework/Algorithms/src/FFTPreProcessing.cpp
AnthonyLim23: I have avoided integer arithmetic for when yData.size is odd. The I have more control over the rounding.

mantid-roman: I believe you get the same result as in `yData.size() / 2`

Framework/Algorithms/src/FFTPreProcessing.cpp
```diff
@@ -0,0 +1,233 @@
...
+  result.points();
+  auto &xData = result.mutableX();
+  auto &yData = result.mutableY();
+  auto &eData = result.mutableE();
+
+  for (size_t i = 0; i < yData.size(); ++i) {
+    yData[i] *= function(xData[i], decayConstant);
+    eData[i] *= function(xData[i], decayConstant);
+  }
+  result.binEdges();
```
mantid-roman: Convert to histogram only when the input is a histogram.

Framework/Algorithms/src/FFTPreProcessing.cpp
```diff
@@ -0,0 +1,233 @@
...
+    std::copy(eData.begin(), eData.end(), newEData.begin() + offset);
+    result = HistogramData::Histogram(
+        HistogramData::Points(newXData), HistogramData::Counts(newYData),
+        HistogramData::CountStandardDeviations(newEData));
+
+  } else {
+    result = HistogramData::Histogram(HistogramData::Points(newXData),
+                                      HistogramData::Counts(newYData));
+  }
+  result.binEdges();
```
mantid-roman: Only for histograms.

mantid-roman (changes requested):

mantid-roman:
@AnthonyLim23 It was easier to change the code directly than to explain what I wanted.

mantid-roman (approved):

Framework/Algorithms/inc/MantidAlgorithms/FFTPreProcessing.h
```diff
@@ -0,0 +1,82 @@
...
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+File change history is stored at: <https://github.com/mantidproject/mantid>
+Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+class DLLExport FFTPreProcessing : public API::Algorithm {
```
martyngigg: I agree with Roman here, pre-processing is ambiguous, is the name requested by the Muon scientists?

Framework/Algorithms/inc/MantidAlgorithms/ApodizationFunctionHelper.h
```diff
@@ -0,0 +1,40 @@
...
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+File change history is stored at: <https://github.com/mantidproject/mantid>
+Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+double lorentz(double time, double decayConstant);
```
martyngigg: We shouldn't introduce functions at global scope like this. At minimum they should be in a namespace.

I can't see that they are used outside of `FFTPreProcessing.cpp`. Can they in fact not just be in an anonymous namespace in that file?

martyngigg (changes requested):

Framework/Algorithms/inc/MantidAlgorithms/ApodizationFunctions.h
```diff
@@ -0,0 +1,45 @@
...
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+File change history is stored at: <https://github.com/mantidproject/mantid>
+Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+
+namespace Mantid {
+namespace Algorithms {
+double lorentz(double time, double decayConstant);
```
martyngigg: Looking at this again I think this still might be a little too general with the names in this namespace.

I think it might be a good idea to add a class called `ApodizationFunction` and make this static members of that class. That way the quite general names can't clash with others.

martyngigg (changes requested):

martyngigg (approved):

PR #20226, opened 2017-08-10
Geometry::ComponentInfo shapes

### LINKED ISSUE #20175: Geometry::ComponentInfo should have Object/Shape

OwenArnold:
"Shape" and "Scale Factor" are done through `ComponentInfo` such that one one ObjComponent's responsibilities, Solid Angle, can now be done directly in `ComponentInfo`

As this is an internal change, a code review should be sufficient. No release notes needed.

supersedes PR #20176

fixes #20175

### DIFF
--- a/Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
+++ b/Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
@@ -68,6 +68,7 @@ private:
   boost::shared_ptr<const std::vector<size_t>> m_parentIndices;
   Mantid::Kernel::cow_ptr<std::vector<Eigen::Vector3d>> m_positions;
   Mantid::Kernel::cow_ptr<std::vector<Eigen::Quaterniond>> m_rotations;
+  Mantid::Kernel::cow_ptr<std::vector<Eigen::Vector3d>> m_scaleFactors;
   const size_t m_size = 0;
   const int64_t m_sourceIndex = -1;
   const int64_t m_sampleIndex = -1;
@@ -88,6 +89,7 @@ public:
                 boost::shared_ptr<const std::vector<size_t>> parentIndices,
                 boost::shared_ptr<std::vector<Eigen::Vector3d>> positions,
                 boost::shared_ptr<std::vector<Eigen::Quaterniond>> rotations,
+                boost::shared_ptr<std::vector<Eigen::Vector3d>> scaleFactors,
                 int64_t sourceIndex, int64_t sampleIndex);
 
   std::vector<size_t> detectorsInSubtree(const size_t componentIndex) const;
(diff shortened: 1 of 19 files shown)


OwenArnold:
@SimonHeybrock I don't want to expand this PR beyond ::solidAngle additions. I'll do boundingBox as part of another PR. I'm not quite finished yet. For example, still need to replace all "sca" write usage.

LamarMoore (commented):
Looks really good.

Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
```diff
@@ -121,6 +123,9 @@ class MANTID_BEAMLINE_DLL ComponentInfo {
   size_t sample() const;
   size_t root() const;
   double l1() const;
+  Eigen::Vector3d scaleFactor(const size_t componentIndex) const;
+  void setScaleFactor(const size_t componentIndex,
+                      const Eigen::Vector3d &scaleFactor);
```
OwenArnold: Not that I can think of currently. The only place that seems to be directly setting this parameter is in the algorithm I have already modified.

LamarMoore (approved):

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -23,6 +23,7 @@ ComponentInfo::ComponentInfo(
     boost::shared_ptr<const std::vector<size_t>> parentIndices,
     boost::shared_ptr<std::vector<Eigen::Vector3d>> positions,
     boost::shared_ptr<std::vector<Eigen::Quaterniond>> rotations,
+    boost::shared_ptr<std::vector<Eigen::Vector3d>> scaleFactors,
```
martyngigg: Very few things use the scale factors. Doesn't this mean that `ComponentInfo` will require a set of scale factors to be constructed?

OwenArnold: It's actually quite well imbedded in may of the calculations of `ObjComponent`. But you're right. Only one algorithm to my knowledge actually writes these scale factors.

OwenArnold: @martyngigg basically this is the problem where `ParameterMaps` have been used to store generic information that is used by the Framework. We need to purge that information at the start from the `ParameterMap` so we have to assume it could exist.

martyngigg: @OwenArnold I see. Thanks for the explanation.

martyngigg (approved):
This looks good to me and seems to be consistent with how the other information in Beamline is stored.

PR #20227, opened 2017-08-10
Remove automatic vanadium creation from ISIS Powder

### LINKED ISSUE #20225: ISIS_Powder - Remove creating automatic vanadiums from focus script
Originally we added aspects to the focus script which would create a vanadium automatically if it wasn't present. This is not longer suitable because it causes confusion for users and can silently break when the parameters used change slightly (which it currently is).

Instead we will retain the default behaviour which is to print a useful error message describing to the user they must create the Vanadium spline manually.

DavidFair:
Description of work.
This PR removes a small step of ISIS Powder to create a Vanadium spline if one was not found whilst focusing. This has been a source of bugs and confusion. Currently it is also broken hence this is the best time to remove it

**To test:**
- Delete any existing Vanadium spline from the Calibration folder
- Run Focus with Polaris data set 
- Ensure a sane error message is thrown

Fixes #20225 

**Release Notes** 
*Does not need to be in the release notes. - This only was used on one instrument who didn't know it existed*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/scripts/Diffraction/isis_powder/abstract_inst.py
+++ b/scripts/Diffraction/isis_powder/abstract_inst.py
@@ -107,16 +107,6 @@ class AbstractInst(object):
         # the instrument override the optional behaviour
         raise NotImplementedError("spline_vanadium_ws must be implemented per instrument")
 
-    # Optional overrides
-    @staticmethod
-    def _can_auto_gen_vanadium_cal():
-        """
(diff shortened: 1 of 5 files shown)


josephframsay (approved):
Changes look good, and testing works as expected

PR #20229, opened 2017-08-11
ISIS SANS GUI experimental

### LINKED ISSUE #19955: Create a new barebones ISIS SANS GUI
Create a barebones ISIS SANS GUI based on the DataProcessorWidget which can handle simple data reductions and allows the user to manipulate some input properties of the reduction similar to the old GUI. 

The GUI should include
* DataProcessorWidget
* User file input
* Batch file input
* Run button
*  ...

Note this does not include feature such as Beam Centre Finder etc.

AntonPiccardoSelg:
**Do not merge before ~~#20336~~ has been merged**

Fixes #19955 

This PR establishes an MVP for the new ISIS SANS GUI. The changes are experimental and minimal and require user feedback.

~~I will try to segment this PR into finer grained PRs. So **don't test this or merge this**~~
I had a look a the code and there is no easy way of partitioning it in its current form. 

## To test:

Please find the relevant data here: \\olympic\Babylon5\Scratch\Anton\DATA_FOR_GUI_TEST\GUI_TEST

##### Standard reduction

1. Open the `ISIS SANS v2 experimental` GUI
2. Load the user file `USER_SANS2D_154E_2p4_4m_M3_Xpress_8mm_SampleChanger_FRONT.txt`
   *  Go `Settings` and confirm that some of the entries in the different tabs have been filled
4. Load the batch file `batch_file.csv`
   * Confirm that there are two rows. Confirm that only `SampleScatter` and `OutputName` has been populated
5. In the `Save Options` select `Both` (this will output the workspace onto the ADS and save the data to the different file types after the reduction. Select `CanSas` and `NXcanSAS` as the save file format.
6. Press `Process` and press `Ok` when asked if you want to reduce all rows.
7. Confirm that after a while you end up with two reduced workspaces with the name `reduction_line_1` and `reduction_line_2`. Confirm that this has been saved out int the appropriate formats in your save location. (Delete the outputs to not confuse this with tests below)

##### Standard reduction with the user changing something in the settings tab

1. Open the `ISIS SANS v2 experimental` GUI
2. Load the user file `USER_SANS2D_154E_2p4_4m_M3_Xpress_8mm_SampleChanger_FRONT.txt`
3. Load the batch file `batch_file.csv`
4. In the `Save Options` select `Memory` (this will only output to the ADS only)
5. Go to the `Settings` tab. You can change several things if you want. For a first test I would suggest to setting `General->ReductionDimensionality` to 2D, since this effect is clearly visible in the output workspace.
6. Press `Process` and press `Ok` when asked if you want to reduce all rows.
7. Confirm that after a while you end up with two reduced workspaces with the name `reduction_line_1` and `reduction_line_2`. Confirm that nothing has been saved out. Ensure that the output workspaces have 150 spectra, this shows that the reduction was 2D.

##### Making changes to the settings

Changes in the settings can be investigated without having to run a full reduction.  The way to check if your changes in the Setting have had an effect is to make use of the `Settings->State Diagnostic` tab. This tab will show the full state information which is passed to the reduction algorithms. This diagnostic tab is currently not hidden since the SANS scientists normally have a large aptitude for knowing the internals of the reduction. This will help everyone to track potential bugs and helps the instrument scientists and the developer reason about the huge amount of parameters that can be set. We can disable(make it invisible) when the the GUI loses its `experimental` status.

Clarification what we see in the `Settings->State Diagnostic` tab: It displays the state information for the row which is currently selected. This state information which contain settings for individual reduction steps, for example "wavelength conversion" or "scaling". The tree view lets the user explore the nested nature of the state. For example, if you want to know if the mask files have been picked up, then you got to `mask->mask_files` and you will see that the list of mask files have been picked up. If this is not clear enough, please ask me.

1. Open the `ISIS SANS v2 experimental` GUI
2. Load the user file `USER_SANS2D_154E_2p4_4m_M3_Xpress_8mm_SampleChanger_FRONT.txt`
3. Load the batch file `batch_file.csv`
4. Go to  `Settings->State Diagnostic` tab, make sure that the selected row index is 0.
    * Go to `save->user_specified_output_name` and confirm that it says `reduction_line_1`
5. Select row index 1
    * Go to `save->user_specified_output_name` and confirm that it says `reduction_line_2`
6. Confirm that `adjustment->normalize_to_monitor->incident_monitor` is 2. Go to the `Adjustment->General->Incident Monitor` line edit and set it to 1. Then go back to the `Settings->State Diagnostic` tab, press `Update row selection` (you have to do this since we are not listening to change events, that would be overkill for this diagnostic tool). 
   * Confirm that `adjustment->normalize_to_monitor->incident_monitor` is 1.
7. Now we test the options tab. Select row index 0. Confirm that `wavelength->wavelength_low` is 1.75. Confirm this also for row index 1.
8. Go to the `Runs` tab and for the first row (index 0) type "WavelengthMin=2" into the `Options` column. 
9. Go back to the `Settings->State Diagnostic` tab and  press `Update row selection` 
   * Confirm that `wavelength->wavelength_low` is 2 for row index 0 and 1.75 for row index 1. 

## Release notes

See [here](https://github.com/mantidproject/mantid/pull/20229/commits/7cd20c18131b9b82504f9d9d9c1807244b8189d4#diff-1e24107d27f6865f223226403f9f3521)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


martyngigg:
@AntonPiccardoSelg There are some merge conflicts after #20082 was merged to master. I thought I could resolve it easily but after resolving the conflict in the header the code doesn't build.

scripts/Interface/ui/sans_isis/sans_data_processor_gui.py
```diff
@@ -0,0 +1,1542 @@
...
+        if file_names:
+            line_edit.setText(file_names[0])
+
+
+# ----------------------------------------------------------------------------------------------------------------------
+# Gui Classes
+# ----------------------------------------------------------------------------------------------------------------------
+class SANSDataProcessorGui(QtGui.QMainWindow, ui_sans_data_processor_window.Ui_SansDataProcessorWindow):
+    data_processor_table = None
+    main_presenter = None
```
martyngigg: I don't think this is ever set is it? The `__init__` method creates a new private `_main_presenter` variable, was this intended to be that?

scripts/Interface/ui/sans_isis/sans_data_processor_gui.py
```diff
@@ -0,0 +1,1542 @@
...
+import os
+try:
+    from mantidplot import *
+    canMantidPlot = True
+except ImportError:
+    canMantidPlot = False
+
+from . import ui_sans_data_processor_window as ui_sans_data_processor_window
+from PyQt4 import QtGui, QtCore
+from mantid.simpleapi import *
```
martyngigg: Is this used? I would assume that mantid stuff would have been called by the presenter?

scripts/Interface/ui/sans_isis/sans_data_processor_gui.py
```diff
@@ -0,0 +1,1542 @@
...
+from mantidqtpython import MantidQt
+from sans.common.enums import (ReductionDimensionality, OutputMode, SaveType, SANSInstrument,
+                               RangeStepType, SampleShape, ReductionMode, FitType)
+from sans.gui_logic.gui_common import (get_reduction_mode_from_gui_selection,
+                                       get_string_for_gui_from_reduction_mode)
+
+# ----------------------------------------------------------------------------------------------------------------------
+# Globals
+# ----------------------------------------------------------------------------------------------------------------------
+gui_logger = Logger("SANS GUI LOGGER")
```
martyngigg: Can we have uppercase for module-level variables?

scripts/Interface/ui/sans_isis/sans_data_processor_gui.py
```diff
@@ -0,0 +1,1542 @@
+from __future__ import (absolute_import)
```
martyngigg: Moving forward I would like to define some sense of order for module imports. As this is new code it would be good to reduce the amount to "cleanup". The proposal is to follow IPython's  [style guide](https://raw.githubusercontent.com/wiki/ipython/ipython/template.py). For the imports this would be as it says:

```
# List all imports, sorted within each section (stdlib/third-party (including mantid)/local sans imports).
# For 'import foo', use one import per line.  For 'from foo.bar import a, b, c'
# it's OK to import multiple items, use the parenthesized syntax 'from foo
# import (a, b, ...)' if the list needs multiple lines.
# Separate stdlib, third-party, and IPython imports by a blank line.
```

scripts/SANS/sans/gui_logic/presenter/property_manager_service.py
```diff
@@ -0,0 +1,107 @@
+from mantid.api import Algorithm
+from mantid.kernel import (PropertyManagerProperty, PropertyManagerDataService)
+from sans.state.state_base import create_deserialized_sans_state_from_property_manager
+
+
+class PropertyManagerService(object):
+    sans_property_manager_prefix = "SANS_PROPERTY_MANAGER_THIS_NEEDS_TO_BE_UNIQUE_"
+
+    class ConverterAlgorithm(Algorithm):
```
martyngigg: What's the reason we need an algorithm for this?

scripts/SANS/sans/gui_logic/presenter/run_tab_presenter.py
```diff
@@ -0,0 +1,880 @@
...
+        super(RunTabPresenter, self).__init__()
+        self._facility = facility
+
+        # Presenter needs to have a handle on the view since it delegates it
+        self._view = None
+        self.set_view(view)
+
+        # Models that are being used by the presenter
+        self._state_model = None
+        self._state_model = None
```
martyngigg: This line is duplicated from the previous one.

scripts/SANS/sans/gui_logic/presenter/run_tab_presenter.py
```diff
@@ -0,0 +1,880 @@
...
+
+    def _set_dummy_workspace(self):
+        number_of_rows = self._view.get_number_of_rows()
+        for row in range(number_of_rows):
+            self._add_to_hidden_options(row, "InputWorkspace", SANS_DUMMY_INPUT_ALGORITHM_PROPERTY_NAME)
+
+    @staticmethod
+    def _create_dummy_input_workspace():
+        if not AnalysisDataService.doesExist(SANS_DUMMY_INPUT_ALGORITHM_PROPERTY_NAME):
+            create_alg = AlgorithmManager.create("CreateSampleWorkspace")
```
martyngigg: If you just need a workspace with "nothing" in it could we just create a 1 spectra/1 bin workspace using the workspace factory and add it to the ADS manually? It would save a lot of work here.

martyngigg (changes requested):
These are some initial comments from looking over the code.

Generally the structure looks good. The code is well separated using MVP and there are good tests. 

One comment from a quick startup of the GUI is that it's possible to push buttons before anything is "ready". For example, I can press process even though the table is empty. Can we disable the buttons until we a ready?

scripts/SANS/sans/gui_logic/presenter/property_manager_service.py
AntonPiccardoSelg: The reason for this is convoluted. The `DataProcessorWidget` can only take basic, string-based inputs. The state object which contains all the information required for the reduction cannot be passed into the `DataProcessorWidget`. It was just never intended for more complex reduction setups. We circumvent this limitation by converting the state object into a `PropertyManager` which is then placed into the the `PropertyManagerDataService`. The algorithm which performs the reduction then grabs the state from the `PropertyManagerDataService`. The state object can be serialized into a Python `dict`, but not directly into a `PropertyManager`. The `ConverterAlgorithm` is used to perform this conversion.

AntonPiccardoSelg:
@martyngigg I agree that we should disable buttons. The main issue is that the data table cannot be current disabled at the moment. In fact Ed is currently working on this.

scripts/SANS/sans/gui_logic/presenter/property_manager_service.py
martyngigg: I see. Could we not just use the `PropertyManagerProperty` directly rather than needing to go through an algorithm?

AntonPiccardoSelg: How do we set the value on a `PropertyManagerProperty` in Python?

martyngigg: @AntonPiccardoSelg You are correct, we can only assign via a string to the property directly so this is the cleanest way to do that at the moment.

martyngigg:
I've finally gotten around to doing to functional testing. It looks very nice. We discusses a few minor issues with the possible confusion around the instrument selection but these are minor and can be cleaned up after further testing with the scientists.

My only real issue was that when I was browsing for the user and batch files the directory dialog didn't remember where I had been. Would it be easy to change it so that either browse button opened in the last used directory?

martyngigg:
**Note that this should not be merged before #20336**.

AntonPiccardoSelg:
@martyngigg 
* unified the qsettings location for all `Browse` buttons in the GUI.
* made the side bar of the SANS GUI wider.

martyngigg:
@AntonPiccardoSelg This looks good now. There's just the minor issue of that warning on startup:

`QLayout: Attempting to add QLayout "" to MaskingTable "masking_table", which already has a layout`

I don't know if this could cause issues down the line. A [stackoverflow]( https://stackoverflow.com/questions/25450598/qlayout-attempting-to-add-qlayout-to-qwidget-which-already-has-a-layout) post suggests that the issue is with how the vertical layout is created for the mask tab. If the table widget is passed in via the constructor then it tries to reset the layout.

I think we might need to change how the layout is attached in the designer and use the same approach as on the rhs of the mask tab and create a `QVBoxLayout` manually rather than through the context menu.

AntonPiccardoSelg:
@martyngigg I had forgotten about that issue. Thanks for reminding me. I will fix that :+1:

AntonPiccardoSelg:
@martyngigg That warning should be gone now.

martyngigg:
@AntonPiccardoSelg Looks good, thanks.

martyngigg (approved):

PR #20237, opened 2017-08-11
Add sphere support to SetSample

### LINKED ISSUE #20207: Add Sphere support to SetSample
Pearl uses a Vanadium sphere which SetSample currently does not support.
Add support so we can carry on diagnosing the issue with tail end noise.

DavidFair:
Description of work.
This PR adds support for Spheres to set sample. This is used by PEARL currently but it also allows other people to use a sphere with relative ease.

Additionally I've made the shape field case insensitive as it was slightly confusing why a `sphere` wouldn't work but a `Sphere` is OK from a user perspective. 

**To test:**
Run the following
```
ws = CreateSampleWorkspace()

SetSample(InputWorkspace=ws, Geometry={'Shape' : 'Sphere', 'Center': [0., 0., 0.], 'Radius' : 1.0}, Material={'ChemicalFormula': 'V'})
```

On the workspace view the "Sample Details". If this has entries (for a Vanadium) SetSample completed successfully. 

Fixes #20207 . 

**Release Notes** 
Added to the release notes [here](https://github.com/mantidproject/mantid/pull/20237/files#diff-d43ebaec89d7bc36364c8bf4c9f97673)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/DataHandling/src/SetSample.cpp
```diff
@@ -508,6 +506,38 @@ SetSample::createCylinderLikeXML(const Kernel::PropertyManager &args,
...
 
+std::string
+SetSample::createSphereXML(const Kernel::PropertyManager &args,
+                           const Geometry::ReferenceFrame &refFrame) const {
+
+  // We have to do a cast to double to help the compiler select the correct
+  // overload for the return type.
+  // Convert to meters when getting the radius
+  const double radius =
+      static_cast<double>(args.getProperty(ShapeArgs::RADIUS)) * 0.01;
```
AnthonyLim23: add a comment somewhere saying that it expects an input of untis (cm?)

Framework/DataHandling/test/SetSampleTest.h
```diff
@@ -286,6 +286,44 @@ class SetSampleTest : public CxxTest::TestSuite {
     TS_ASSERT_EQUALS(false, sampleShape.isValid(V3D(0, 0.041, -0.001)));
   }
 
+  void test_Setting_Geometry_As_Sphere() {
+    // This is the test using correct expected case
+    run_Geometry_As_Sphere_Test("Sphere");
+  }
+
+  void test_Setting_Uppercase_Geometry_As_Sphere() {
```
AnthonyLim23: could we add a test for mixed case (e.g. sPhERe)?

AnthonyLim23 (commented):
Need to test it but looks good so far. Just a couple of minor things

AnthonyLim23 (approved):
worked as described, except the command was sample material.

Framework/DataHandling/src/SetSample.cpp
```diff
@@ -508,6 +506,38 @@ SetSample::createCylinderLikeXML(const Kernel::PropertyManager &args,
...
+  // XML needs center position of bottom base but user specifies center of
+  // sphere
+  const auto axisIdx = refFrame.pointingUp();
+
+  // Multiple radius by 2 to convert into height (diameter)
+  const V3D baseCentre = getShapeBaseCoordinates(centre, (radius * 2), axisIdx);
+  const std::string tag{"sphere"};
+
+  std::ostringstream xmlShapeStream;
+  xmlShapeStream << "<" << tag << " id=\"sample-shape\"> "
```
martyngigg: This doesn't make sense to me. This is not the correct XML to define a sphere: http://docs.mantidproject.org/nightly/concepts/HowToDefineGeometricShape.html#sphere. A sphere doesn't have an axis and only requires a centre and a radius to configure it. I am surprised the test works.

martyngigg (changes requested):

PR #20246, opened 2017-08-15
Formatted chebyshev table and added comments

jamesquantum:
Added newlines as well as commented headings of the rows and columns around the CHEBYSHEV array so that it visually corresponds better to the Carpenter (1969) Table 1 in order that the code is more readable when trying to understand the behaviour of the algorithm.
re #20235


jamesphysics:
I didn't anticipate clang-format disliking the changes but none the less with the comments it should now be slightly easier to read with the aid of the original paper

Framework/Algorithms/src/MultipleScatteringCylinderAbsorption.cpp
```diff
@@ -34,22 +34,28 @@ using namespace Geometry;
...
+    0.730284, -0.249987, 0.019448, -0.000006, 0.000249,  -0.000004, // (1,1)
+    0.848859, -0.452690, 0.056557, -0.000009, 0.000000,  -0.000006, // (1,2)
+    1.133129, -0.749962, 0.118245, -0.000018, -0.001345, -0.000012, // (1,3)
+    1.641112, -1.241639, 0.226247, -0.000045, -0.004821, -0.000030, // (1,4)
+    0.848859, -0.452690, 0.056557, -0.000009, 0.000000,  -0.000006, // (2,1)
+    1.000006, -0.821100, 0.166645, -0.012096, 0.000008,  -0.000126, // (2,2)
+    1.358113, -1.358076, 0.348199, -0.038817, 0.000022,  -0.000021, // (2,3)
+    0.0,      0.0,       0.0,      0.0,       0.0,       0.0,       // (2,4)
+    1.133129, -0.749962, 0.118245, -0.000018, -0.001345, -0.000012, // (3,1)
+    1.358113, -1.358076, 0.348199, -0.038817, 0.000022,  -0.000021, // (3,2)
```
peterfpeterson: Assuming `(m,n) = (2,3) = (3,2)`. The other Chebyshev polynomials have this feature, but I wasn't sure and the paper isn't explicit about it.

peterfpeterson (approved):
Changes look good. It is much easier to compare directly to the table.

PR #20251, opened 2017-08-15
Enforce the associative property for ProductFunction

### LINKED ISSUE #20250: Enforce the associative property for ProductFunction
Given functions `f`, `g`, and `h`, enforce that the three expressions `(f * g) * h`,  `f * (g * h)`, and `f * g * h` result in the same `ProductFunction` object.

jmborr:
Base class for composite functions that are associative, and implementation for `ProductFunction`.

Fixes #20250

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/CMakeLists.txt
+++ b/Framework/API/CMakeLists.txt
@@ -10,6 +10,7 @@ set ( SRC_FILES
 	src/AlgorithmProxy.cpp
 	src/AnalysisDataService.cpp
 	src/ArchiveSearchFactory.cpp
+	src/AssociativeCompositeFunction.cpp
 	src/Axis.cpp
 	src/BinEdgeAxis.cpp
 	src/BoxController.cpp
@@ -169,6 +170,7 @@ set ( INC_FILES
 	inc/MantidAPI/AlgorithmProxy.h
 	inc/MantidAPI/AnalysisDataService.h
 	inc/MantidAPI/ArchiveSearchFactory.h
+	inc/MantidAPI/AssociativeCompositeFunction.h
 	inc/MantidAPI/Axis.h
 	inc/MantidAPI/BinEdgeAxis.h
 	inc/MantidAPI/BoxController.h
@@ -358,6 +360,7 @@ set ( TEST_FILES
 	AlgorithmMPITest.h
 	AlgorithmTest.h
 	AnalysisDataServiceTest.h
+	AssociativeCompositeFunctionTest.h
 	AsynchronousTest.h
 	BinEdgeAxisTest.h
 	BoxControllerTest.h
--- a/Framework/API/inc/MantidAPI/AssociativeCompositeFunction.h
+++ b/Framework/API/inc/MantidAPI/AssociativeCompositeFunction.h
@@ -1,0 +1,57 @@
+#ifndef MANTID_API_ASSOCIATIVECOMPOSITEFUNCTION_H_
+#define MANTID_API_ASSOCIATIVECOMPOSITEFUNCTION_H_
+
+#include "MantidAPI/CompositeFunction.h"
+
+namespace Mantid {
+    namespace API {
+/** Subclass of CompositeFunction that enforces the associative property. If ASF
+   denotes the composition operation, then:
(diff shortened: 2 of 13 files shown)


jmborr:
@mantid-roman @martyngigg I have opened this issue to enforce the associative property in the `ProductFunction` when creating fit functions. For instance, instantiating from string
 ```
(composite=ProductFunction;
   name=ExpDecay;
   composite=ProductFunction;name=Gaussian;name=Lorentzian)
)
```
will result in a function whose string representation is
```
(composite=ProductFunction;name=ExpDecay;name=Gaussian;name=Lorentzian)
```
So functions `f * (g * h)` and `(f * g) * h` will be represented by their "flat" version `f * g * h`.
Please let me know your thoughts about it.

**Comment by KarlPalmen** [comment]
Issue #20279 has been created to enable ``castToComposite`` to be removed. It does so by tackling the root cause of the problem arising from the fact that ``ProductFunction`` etc. are not really exported to python.

This PR request may be worth merging if the above issue cannot be resolved. It might also be worth merging if it would eliminate the need for the ``flatten`` method from ``CompositeFunctionWrapper``, but while this may apply to ``ProductFunction``s, it can't apply to ``CompositeFunction``, because this is a superclass of ``AssociativeCompositeFunction`` and so ``flatten`` must remain.

**Comment by jmborr** [comment]
@KarlPalmen That's why I wrote[ Addition](https://github.com/mantidproject/mantid/blob/e378026b952632c7049cc95ffb356f6ec580e4af/Framework/CurveFitting/inc/MantidCurveFitting/Functions/AdditionFunction.h) funtion. With `Addition` and `ProductFunction`, there should be no need for flatten. :bowtie:

**Comment by jmborr** [comment]
@KarlPalmen  @mantid-roman As of now, the automatic flattening process works as long as the associative function to be added does not have a tie among its component functions.  One example of this case:
```
f = (composite=ProductFunction;name=Linear)
g = (composite=ProductFunction;name=Gauss;name=Cubic;ties=(f1.c0=f0.c))  # a cross-tie
```
Now I want to do `h = f * g`. The current fit framework will result in:
```
h = (composite=ProductFunction;name=Linear;(composite=ProductFunction;name=Gauss;name=Cubic;ties=(f1.c0=f0.c)))
```
But I want the flat version because we have the product of three simple functions:
```
h = (composite=ProductFunction;name=Linear;name=Gauss;name=Cubic;ties=(f2.c0=f1.c)) 
```
The original tie expression `f1.c0=f0.c` makes sense in the context of composite `g`. To promote this tie to the new composite context `h`, the only way I can think of is to construct a mapping between the old and the new function indexes (`g.f1` becomes `h.f2` and `g.f0` becomes `h.f1`) and then recreate all the ties with the new indexes.
Well, do you see other alternatives?

KarlPalmen:
I haven't addressed this issue in my works so far and there are no unit tests for it. 

I think the suggested idea is good and it may occur automatically. I expect this would be the case, if the ties inside the functions refer to parameters by their index within the composite function rather than the parameter name. Flattening does not change the order of the parameters and hence their index.

KarlPalmen:
This fixes a problem that has been fixed in an alternative way. This could be considered in maintenance as a better way to fix this problem. Hence I'm moving it to next release.

jmborr:
@KarlPalmen Thanks, Karl. I had no time as of late to finish this off.
By the way, what is the alternate way you mention? Where can I find the related code.

KarlPalmen:
The alternative way it has been fixed is by properly exposing the curve fitting functions to python as in issue #20279 and retaining the flattening when all add or all multiply.

KarlPalmen:
I'll be away for the next two weeks after today.

martyngigg:
Is this still valid with the Python functionality?

KarlPalmen:
This change is not necessary after the exposure of ``CurveFitting`` to python, but may be desirable.

KarlPalmen:
This issue solves a problem with the binding of fitting functions to python that has been solved by other means.

jmborr:
I'm closing this pull request since the issue was solved in the python API

PR #20274, opened 2017-08-17
API::ComponentInfo::BoundingBox needed

### LINKED ISSUE #20264: Geometry::ComponentInfo BoundingBox
`ComponentInfo` requires a `BoundingBox` implementation. Performance should be benchmarked and tuned as part of this as some effort seems to already have gone into caching of bounding boxes as part of the parametermaps.

Blocked by #20226

OwenArnold:
Implement cache free BoundingBox in Instrument 2.0. Without re-writing `Geometry::BoundingBox` this has to live in the wrapper layers.

**Tester**
* Code Review
* As far as I can tell, things are running faster than they did before `PredictFractionalPeaks` runs 2* faster, but please pay attention to any performance regressions you come across.

**This PR is for release 3.12 of Mantid**

Fixes #20264


Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -336,8 +341,10 @@ ComponentInfo::Range
 ComponentInfo::componentRangeInSubtree(const size_t index) const {
   const auto rangesIndex = compOffsetIndex(index);
   const auto range = (*m_componentRanges)[rangesIndex];
-  return {m_assemblySortedComponentIndices->begin() + range.first,
-          m_assemblySortedComponentIndices->begin() + range.second};
+  ComponentInfo::Range ret = {
+      m_assemblySortedComponentIndices->begin() + range.first,
+      m_assemblySortedComponentIndices->begin() + range.second};
+  return ret;
```
SimonHeybrock: Why this change? I cannot spot a difference?

Framework/Beamline/src/DetectorInfo.cpp
```diff
@@ -305,6 +305,18 @@ Eigen::Vector3d DetectorInfo::samplePosition() const {
   return m_componentInfo->samplePosition();
 }
 
+bool DetectorInfo::isRectangularDetectorPixel(const size_t index) const {
```
SimonHeybrock: I am wondering if we should make this a more general concept? `RectangularDetector` is a special case, but I suppose a lot of client code might be interested more generally in whether this is a regular/structured assembly?

Also, I think the naming is misleading, it seems to imply that the pixel is rectangular, the actual meaning (it is part of a rectangular/regular assembly of pixels) is only clear to people familiar with the old instrument code.

Framework/Geometry/src/Instrument/ComponentInfo.cpp
```diff
@@ -183,5 +197,136 @@ double ComponentInfo::solidAngle(const size_t componentIndex,
...
+  if (reference && !reference->isAxisAligned()) { // copy coordinate system
+
+    std::vector<Kernel::V3D> coordSystem;
+    coordSystem.assign(reference->getCoordSystem().begin(),
+                       reference->getCoordSystem().end());
+
+    // realign to reference coordinate system
+    absoluteBB.realign(&coordSystem);
+  }
+  return absoluteBB;
```
SimonHeybrock: I suppose this is just copied from the old method to compute the BB?

Framework/Geometry/src/Instrument/ComponentInfo.cpp
```diff
@@ -183,5 +197,136 @@ double ComponentInfo::solidAngle(const size_t componentIndex,
...
+
+      // Get bounding box for rectangular bank.
+      // Record detector ranges to skip
+      // Skip all sub components.
+      detExclusions.emplace(std::make_pair(corner1, corner2));
+      compIterator = innerRangeComp.rend();
+    } else {
+      absoluteBB.grow(componentBoundingBox(index, reference));
+      ++compIterator;
+    }
```
SimonHeybrock: Same here, I suppose this is just copied from the old method to compute the BB?

SimonHeybrock (commented):
This looks quite good. I have not yet looked at the bounding box math and the added tests in detail, so here are just some comments mainly on the API:

Framework/Beamline/src/DetectorInfo.cpp
OwenArnold: I'd be tempted to remove it. What do you think? We don't actually need it for this current work. I do however like `ComponentInfo::isRectangularBank`, that's required for this work, and generally something I think that would be useful.

Framework/Geometry/src/Instrument/ComponentInfo.cpp
OwenArnold: Essentially yet. At the point we wrote a `Beamline::BoundingBox` we could look at replaceing some of this. I don't see the value in changing any of this yet.

Framework/Beamline/src/DetectorInfo.cpp
SimonHeybrock: I think `isRectangularBank` is also a bit narrow. I do not strongly object having it now, but maybe we can think of this concept could/should be made more general to support, e.g., things like the structured detectors Lamar did for LOKI?

martyngigg: I agree it would be good if we we could generalise the concept of a structured assembly. In many ways our standard PSD tubes could fall into this type of category as well. We already have a special flag so that the instrument view can treat them differently and draw them as one. Perhaps this could feed into that?

Framework/Geometry/src/Instrument.cpp
```diff
@@ -816,7 +816,12 @@ std::vector<detid_t> Instrument::getMonitors() const {
  */
 void Instrument::getBoundingBox(BoundingBox &assemblyBox) const {
```
martyngigg: Would it be too much work to change the signature of this method to  `BoundingBox Instrument::getBoundingBox() const` ? That feels like a better signature for this.

Framework/Geometry/src/Instrument/ComponentInfo.cpp
martyngigg: I checked with the old version of `ObjComponent::getBoundingBox` and this looks to be doing the same thing.

Framework/Geometry/src/Instrument/ComponentInfo.cpp
martyngigg: The `RectangularDetector` class used to have its own `getBoundingBox` method that filled in the details for itself. If we have other detector types this `if/else if` is going to get a bit long isn't it? Could the visitor pattern work here too?

Framework/Geometry/src/Instrument.cpp
OwenArnold: I was trying not to change the public API of `Instrument`, though you will see for `API::ComponentInfo` I am returning by value via `BoundingBox API::ComponentInfo::boundingBox() const`. Since `Instrument` is being phased out, I wanted to minimise changes there. Is there a particular reason you want this changed?

Framework/Geometry/src/Instrument/ComponentInfo.cpp
OwenArnold: I'll look into it.

Framework/Geometry/src/Instrument/ComponentInfo.cpp
OwenArnold: @martyngigg. Yes, deliberately so. In this phase on `Instrument 2.0` we are attempting to get `ObjComponent` functionality into `ComponentInfo`.

Framework/Beamline/src/DetectorInfo.cpp
OwenArnold: I think it would be worth taking a look, and I would be happy to have more general concepts. What I am bearing in mind is that there are areas of the existing codebase that explicitly need to identify and work with `RectangularDetector`. The current codebase is peppered with methods like this [example](https://github.com/mantidproject/mantid/blob/master/Framework/Geometry/src/Instrument.cpp#L1298)

OwenArnold: Maybe one option is to store the `RectangularBank` flag to support the case I have linked to above, but also to have a flag that can be used to recognise "classes" of assemblies (`StructuredDetector` etc), for which the processing I have put into `ComponentInfo` would be applicable.

Framework/Geometry/src/Instrument.cpp
martyngigg: No particular reason. I must admit I'd forgotten that this class wasn't going to survive the phase out so it was purely a cleanup argument.

Framework/Beamline/src/ComponentInfo.cpp
OwenArnold: I  think this is a relic from some previous debugging I was doing, I'll change it back.

Framework/Geometry/src/Instrument/ComponentInfo.cpp
```diff
@@ -183,5 +197,136 @@ double ComponentInfo::solidAngle(const size_t componentIndex,
...
+      ++compIterator;
+    } else if (isStructuredBank(index)) {
+      auto innerRangeComp = m_componentInfo->componentRangeInSubtree(index);
+      // nSubComponents, subtract off self hence -1.
+      auto nSubComponents = innerRangeComp.end() - innerRangeComp.begin() - 1;
+      auto innerRangeDet = m_componentInfo->detectorRangeInSubtree(index);
+      auto nSubDetectors =
+          std::distance(innerRangeDet.begin(), innerRangeDet.end());
+      auto nY = nSubDetectors / nSubComponents;
+      size_t corner1 = *innerRangeDet.begin();
```
LamarMoore: Minor point maybe instead of using corner maybe `bottomLeft`, `bottomRight`, `topLeft`, `topRight`? Helps with visualising what you are doing. Also `nSubComponents`  could be `numHorzColumns`. Makes the calculation easier to read?

LamarMoore (approved):
Looks good to me.

Framework/Geometry/src/Instrument/ComponentInfo.cpp
```diff
@@ -183,5 +197,136 @@ double ComponentInfo::solidAngle(const size_t componentIndex,
...
+ * Compute the bounding box for the component with componentIndex taking into
+ *account
+ * all sub components.
+ *
+ * @param componentIndex : Component index to get the bounding box for
+ * @param reference : Optional reference for coordinate system for non-axis
+ *aligned bounding boxes
+ * @return Absolute bounding box
+ */
+BoundingBox ComponentInfo::boundingBox(const size_t componentIndex,
```
LamarMoore: Checked this for `StructuredDetector`. Since it is built in exactly the same way as `RectangularDetector` the optimization should be valid 👍

SimonHeybrock:
This also looks ready to me. @martyngigg Did you want to have a look or should I go ahead and merge?

martyngigg (approved):
Looks good to me too now.

PR #20282, opened 2017-08-18
Fix osx master build

mantid-roman:
Removed mentions of the VatesSimpleGui stand alone executable from build/install scripts.

### DIFF
--- a/buildconfig/CMake/CppCheckSetup.cmake
+++ b/buildconfig/CMake/CppCheckSetup.cmake
@@ -30,7 +30,6 @@ if ( CPPCHECK_EXECUTABLE )
         Framework/Kernel/inc
         Vates/VatesAPI/inc
         Vates/VatesSimpleGui/ViewWidgets/inc
-        Vates/VatesSimpleGui/StandAloneExec/inc
         Vates/VatesSimpleGui/QtWidgets/inc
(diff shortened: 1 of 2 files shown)


martyngigg:
@mantid-roman It looks like there are more changes required in the `make_package.rb` script relating to the library names.

martyngigg (approved):
@mantidproject/gatekeepers This fixes the OSX packaging that had apparently been broken in a few places. The PR on OSX has been rebuilt with `BUILD_PACKAGE=ON` and it now passes.

buildconfig/CMake/CppCheckSetup.cmake
```diff
@@ -30,7 +30,6 @@ if ( CPPCHECK_EXECUTABLE )
         Framework/Kernel/inc
         Vates/VatesAPI/inc
         Vates/VatesSimpleGui/ViewWidgets/inc
-        Vates/VatesSimpleGui/StandAloneExec/inc
```
quantumsteve: Should these other cppcheck include directories be updated?

martyngigg: We noticed that too and @mantid-roman fixed them in #20291, which is also ready to go.

PR #20309, opened 2017-08-22
Abins: dft loaders enhancements and new dmol3 loader

dymkowsk:
1) Scaling fixes: 
    a) Fix bug with scaling for LoadCASTEP 
    b) Fix bug with scaling in Instrument class 
    c) Re-define proportionality constants for quantum event 2 and 3.  
    d) Make sure intensities do not depend on number of points per broaden peak and used bin width
 
    Na2SiF6 (CASTEP): Bin width 1cm-1 and different number of points per broadened peak:    
na2sif6_1cm_diff_pkt

   Na2SiF6 (CASTEP):  Different bin widths and 200 pkt per point:
na2sif6__diff_cm_200_pkt

2) Refactor DFT loaders and corresponding tests to make code more compact. Create helper classes GeneralDFTParser and  GeneralLoadDFTTester which are used by  loaders and their tests.
3) Add support for DMOL3 DFT program: http://accelrys.com/products/datasheets/dmol3.pdf. Currently only Gamma point / molecular DFT output supported. Several benchmarks have been done to verify implementation. 

     AbINS calculation of S(Q, E) for  Na2SiF6 with data from CASTEP and DMOL3: 
screenshot from 2017-08-25 19-23-59

System test for LoadDMOL3 has been added. This test reproduces the result for Na2SiF6 from the figure above. 

Also benchmark for LTA_40_O2 has been performed:
screenshot from 2017-08-26 08-37-30

It can be compared with figures 6, 7 from:
T. Lemishko et al, J. Phys. Chem. C 2016, 120, 24904−24909.

**To test:**
System test for LoadDMOL3. Since it is a longer test it has  a flag  excludeInPullRequests=True and it has to be tested manually before it can be merge into master.

Fixes #xxxx. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


dymkowsk:
@brandonhewer while preparing pictures for my article I observed problems with my scalling so I made a commit to fix it. I will soon update my systemtests.

scripts/AbinsModules/Instruments/Instrument.py
```diff
@@ -91,7 +91,8 @@ def _gaussian(self, sigma=None, points=None, center=None):
         :return: numpy array with calculated Gaussian values
         """
         sigma_factor = 2.0 * sigma * sigma
-        return 1.0 / np.sqrt(sigma_factor * np.pi) * np.exp(-(points - center) ** 2 / sigma_factor)
+        norm = points.size / (6 * sigma)
```
dymkowsk: @brandonhewer  here is a bug: points.size is:
points.size = number of frequencies * number of points divided per broadened peak

instead of  points.size it should be  a number of points divided per broadened peak, e.g,
AbinsModules.AbinsParameters.pkt_per_peak.

I will do some additional tests and commit a fix for that later today. Apart from that one line I also have to double check my modifications to scaling parameters for higher quantum order events, regenerate my reference data and check if they are OK.)

scripts/AbinsModules/LoadCRYSTAL.py
```diff
@@ -120,18 +123,18 @@ def _determine_dispersion(self):
...
-        obj_file.readline()  # Line: X                    Y                    Z
+        self._parser.find_first(file_obj=file_obj, msg="DIRECT LATTICE VECTORS CARTESIAN COMPONENTS (ANGSTROM)")
+        file_obj.readline()  # Line: X                    Y                    Z
         dim = 3
         vectors = []
         for i in range(dim):
-            line = obj_file.readline()
+            line = file_obj.readline()
             line = line.split()
             vector = []
```
brandonhewer: A small point here - could vector be initialised with a list comprehension here `vector = [float(item) for item in line]`. This would be faster and more pythonic

scripts/AbinsModules/LoadCRYSTAL.py
```diff
@@ -335,41 +340,17 @@ def _parse_item(self, item=None, container=None, part=None):
...
-        allowed_keywords = [" X ", " Y ", " Z ", "-", "REAL", "IMAGINARY", "MODES", "DISPERSION"]
+        allowed_keywords = [b" X ", b" Y ", b" Z ", b"-", b"REAL", b"IMAGINARY", b"MODES", b"DISPERSION"]
+
         # remove empty lines:
         pos = None
-        while not self._file_end(file_obj=file_obj):
+        while not self._parser.file_end(file_obj=file_obj):
             pos = file_obj.tell()
             line = file_obj.readline()
             if line.strip():
```
brandonhewer: Line 364-367 could be rewritten with `return any(key in line for key in allowed_keywords)`

scripts/AbinsModules/LoadCRYSTAL.py
```diff
@@ -529,6 +485,7 @@ def create_kpoints_data_helper(self, atomic_displacements=None, atomic_coordinat
         :param column: number of atomic_displacements column to parse
         :param freq_num: number of mode (frequency)
         :param row_width: current width of row to parse
+        :return normalised atomic displacements
         """
         xdisp = atomic_displacements[0]
         ydisp = atomic_displacements[1]
```
brandonhewer: In the loop starting on Line 498, there is a lot recalculated in every iteration, including the number of atomic_coordinates

scripts/AbinsModules/LoadDMOL3.py
```diff
@@ -0,0 +1,238 @@
...
+        :param obj_file: file object from which we read
+        :param data: Python dictionary to which found lattice vectors should be added
+        """
+        self._parser.find_first(file_obj=obj_file, msg="$cell vectors")
+
+        dim = 3
+        vectors = []
+        for i in range(dim):
+            line = obj_file.readline().split()
+            vector = list(map(self._convert_to_angstroms, line))
```
brandonhewer: Would it be better to use a list comprehension here rather than `list(map(...`

scripts/AbinsModules/SPowderSemiEmpiricalCalculator.py
```diff
@@ -367,6 +376,7 @@ def _calculate_s_powder_one_atom(self, atom=None):
                 s = self._calculate_s_powder_one_atom_core(atom=atom)
```
brandonhewer: Line 321 could be rewritten with a list comprehension `result = [self._s_calculate_power_one_atom(atom=atom) for atom in atoms]`

brandonhewer (changes requested):
As a general note - it can be important to know that python function calls are rather expensive (python uses a hash table to get from the function signature to a pointer to the function). This shouldn't usually be a massive issue, however, for tight loops/loops called frequently, this can have a significant impact on performance. This impact can be reduced by initialising a local variable with a pointer to the function.
I only mention this as a lot of the Abins code includes calling functions from within loops, and often nested loops.

dymkowsk:
Thanks for suggestions. I agree that list comprehension is a good idea to speed up things.  I will make a new commit with more usage of list comprehension.

scripts/AbinsModules/LoadCRYSTAL.py
dymkowsk: Here atomic displacements are rescaled. I can rearrange code so it will be more compact.

brandonhewer:
System test for LoadDMOL3 works as expected, however, when attempting to use an invalid phonon file, I received the following error:
abinserror
Could this error be made more user friendly?

dymkowsk:
Yes it can be made more user friendly. I will fix error message content.

dymkowsk:
I fixed bug. The problem was that self._dft_program was None. Now every time consistency between extension of file and DFT program is checked local variable dft_program is used and it always corresponds to the current value of property DFTProgram. 

Below is how it looks now when user tries to set DMOL3 but loads file with CASTEP extension.
first_message

second_message

brandonhewer:
Great, I couldn't find any more issues. Seems to all be working.

brandonhewer (approved):

scripts/test/AbinsLoadDMOL3Test.py
```diff
@@ -0,0 +1,47 @@
+from __future__ import (absolute_import, division, print_function)
+import unittest
+from mantid.simpleapi import logger
+import AbinsModules
+
+
+def old_python():
```
quantumsteve: Now that Mantid only supports Python 2.7+ [this check](https://github.com/mantidproject/mantid/blob/81b9c3d53435f150ae0764d6475778ba24fd5a12/scripts/AbinsModules/AbinsTestHelpers.py#L42) is unnecessary and should be removed.

quantumsteve (changes requested):
Please remove checks for `old_python` in new work and either remove or write an issue to remove the `old_python()` method and fix anything that depends on it.

dymkowsk:
@quantumsteve I have removed check for old Python from my tests. Should I fix anything else?

quantumsteve (approved):

quantumsteve:
I think you may also remove `AbinsTestHelpers.is_numpy_valid` but that can be done as a separate issue.

PR #20329, opened 2017-08-23
Fix bug in determining data range

peterfpeterson:
The issue was when a single xmin/xmax was specified for a dataset
with multiple spectra

Description of work.

**To test:**

```python
LoadNexusMonitors(Filename='NOM_91603', OutputWorkspace='NOM_91603')
NormaliseByCurrent(InputWorkspace='NOM_91603', OutputWorkspace='NOM_91603')
ConvertUnits(InputWorkspace='NOM_91603', OutputWorkspace='NOM_91603',
               Target='Wavelength', EMode='Elastic')
ResampleX(InputWorkspace='NOM_91603', OutputWorkspace='NOM_91603',
         XMin=.1, XMax=2.9, NumberBins=6000, LogBinning=True)
```

NOMAD data isn't specifically required, just event data with more than one spectrum that you want to `ResampleX`.

**Release Notes** are in the diffraction section, because that is who uses this algorithm.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Algorithms/src/ResampleX.cpp
```diff
@@ -114,8 +114,23 @@ map<string, string> ResampleX::validateInputs() {
...
-  bool updateXMins = xmins.empty(); // they weren't set
-  bool updateXMaxs = xmaxs.empty(); // they weren't set
+  const size_t numSpectra = inputWS->getNumberHistograms();
+
+  // pad out the ranges by copying the first value to the rest that are needed
+  if (xmins.size() == 1) {
+    const double value = xmins.front();
+    while (xmins.size() < numSpectra)
+      xmins.push_back(value);
+  }
```
quantumsteve: Use std::vector::insert. The third form looks like what you've done here and below 

http://en.cppreference.com/w/cpp/container/vector/insert

Framework/Algorithms/src/ResampleX.cpp
```diff
@@ -114,8 +114,21 @@ map<string, string> ResampleX::validateInputs() {
...
+  // pad out the ranges by copying the first value to the rest that are needed
+  if (xmins.size() == 1 && numSpectra > xmins.size()) {
+    const double value = xmins.front();
+    xmins.insert(xmins.end(), numSpectra - xmins.size(), value);
+  }
+  if (xmaxs.size() == 1 && numSpectra > xmaxs.size()) {
+    const double value = xmaxs.front();
+    xmaxs.insert(xmaxs.end(), numSpectra - xmaxs.size(), value);
+  }
+
```
quantumsteve: Please update the documentation for each parameter.

quantumsteve (changes requested):
I tried the included script and it functions as described in the PR. The documentation should be updated to include this behaviour and/or a log message should be added stating this workaround.

Framework/Algorithms/src/ResampleX.cpp
peterfpeterson: The [algorithm documentation](http://docs.mantidproject.org/nightly/algorithms/ResampleX-v1.html) already states this. It was missing behavior that was only recently noticed.

quantumsteve: The fact is was only recently noticed suggests we need to make it more visible in the documentation, such as in the parameter description. At the same time I worry about getting too deeply into the realm of yak shaving.

quantumsteve (approved):

PR #20337, opened 2017-08-25
Add support for unicode string conversion in python2

peterfpeterson:
Description of work.

**To test:**

While the particular file doesn't matter, testing both unicode strings and native types together is important. The ARCS file in the instructions below is in the unit tests.

1. Build against python2
1. Run `wksp = Load(Filename=unicode('ARCS_2963.nxs'), MetaDataOnly=True)` and get `ValueError: When converting parameter...`. 
1. Merge this branch
1. Run `wksp = Load(Filename=unicode('ARCS_2963.nxs'), MetaDataOnly=True)` and see that it now works

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/PythonInterface/mantid/kernel/src/Exports/IPropertyManager.cpp
+++ b/Framework/PythonInterface/mantid/kernel/src/Exports/IPropertyManager.cpp
@@ -11,6 +11,7 @@
 #include <boost/python/list.hpp>
 #include <boost/python/register_ptr_to_python.hpp>
 #include <boost/python/return_internal_reference.hpp>
+#include <boost/python/str.hpp>
 
 using namespace Mantid::Kernel;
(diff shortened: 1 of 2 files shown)


martyngigg (changes requested):
This looks good to me.

I tried the suggestions and got the expected results. I also tried 

```python
ws = sapi.Rebin(ws, Params=unicode("500,10,10000"))
```

with and without unicode and I got the expected result.

I think it would be good to add a unit test to [`Framework/PythonInterface/test/python/mantid/api/AlgorithmTest.py`](https://github.com/mantidproject/mantid/blob/master/Framework/PythonInterface/test/python/mantid/api/AlgorithmTest.py) to make sure this keeps working.

martyngigg (approved):
Looks good.

Once the tests pass LGTM.

PR #20350, opened 2017-08-30
Remove default detectors IDs from Workspace2D

SimonHeybrock:
`Workspace2D` used to init its spectra with default detector IDs starting from 1. This does not make sense since there is not instrument, i.e., the IDs are just dangling references, and if an instrument is loaded IDs are in general not contiguous and starting from 1, yielding orphaned detectors and still dangling references.

Most actually code rebuilds the spectra mapping upon loading an instrument so this is mainly relevant for unit tests. Note that `EventWorkspace` does not set default IDs, so this PR brings `Workspace2D` in line with `EventWorkspace`.

As a side effect:
- Fixed bug introduced when refactoring `LoadQKK` (would not init histograms anymore).
- Fixed `UpdateInstrumentFromFileTest`, which did not set IDs for any but spectrum 0.

**To test:**

Code review.
No associated issue.
No release notes, this should be an internal change?

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Algorithms/src/AsymmetryCalc.cpp
```diff
@@ -129,6 +129,7 @@ void AsymmetryCalc::exec() {
   // Create a point data workspace with only one spectra for forward
   auto outputWS = DataObjects::create<API::HistoWorkspace>(
       *inputWS, 1, tmpWS->points(forward));
+  outputWS->getSpectrum(0).setDetectorID(static_cast<detid_t>(1));
```
OwenArnold: Isn't the modern way to do this via `setSpectrumDefinition` on `ExperimentInfo`?

Framework/Algorithms/src/CreateWorkspace.cpp
```diff
@@ -199,13 +199,6 @@ void CreateWorkspace::exec() {
     const std::vector<double>::difference_type yEnd = yStart + ySize;
     auto local_i = localIndices[0];
 
-    if (!parentWS) {
-      // There is no instrument so setting detector IDs makes no sense, but if
-      // we don't we break quite a few unit tests relying on this legacy
-      // behavior.
-      outputWS->getSpectrum(local_i).setDetectorID(i + 1);
```
OwenArnold: Good to see this gone. Could you modify the unit tests for `CreateWorkspace` to guard against its accidental/deliberate reintroduction

Framework/API/src/ExperimentInfo.cpp
```diff
@@ -491,16 +491,12 @@ void ExperimentInfo::setNumberOfDetectorGroups(const size_t count) const {
 void ExperimentInfo::setDetectorGrouping(
```
OwenArnold: This method is public, but as far as I can tell it is not used externally. It's called by `cacheDetectorGroupings`, which is only used by `ConvertToMD`, basically I'm wondering if the try-catch is needed anymore.

OwenArnold (commented):
My major concern here is what we do now that the behaviour of `CreateWorkspace` has changed. You have correctly fixed things in our code and executed scripts. What about external scripts? How do we announce these changes to the community.

Framework/Algorithms/src/AsymmetryCalc.cpp
SimonHeybrock: Well, I think that is exactly the problem: `setSpectrumDefinition` will not let you set anything that corresponds to detector IDs that do not exist.

Framework/API/src/ExperimentInfo.cpp
SimonHeybrock: It strips out any bad detector IDs and we still have many places that set bad IDs so this cannot be removed.

SimonHeybrock:
@OwenArnold Regarding your main concern: I agree that we should definitely announce this. Personally I see this as a bug fix since it removes potentially unwanted mappings. Anyone who actually needs sensible mapping to the instrument is like to run, e.g., `LoadInstrument` with `RewriteSpectrumMapping=True`, since the default IDs from `CreateWorkspace` are not sensible (or just match accidentally).

Would having this in the release notes be enough?

Framework/API/src/ExperimentInfo.cpp
OwenArnold: Ok. I think that is probably true. So lets leave this alone.

OwenArnold (approved):
Latest changes to documentation look good. Along the lines of our discussion this morning and sufficiently detailed too.

PR #20357, opened 2017-08-31
New version of SpecularReflectionCalculateTheta

### LINKED ISSUE #20332: Add new version of SRCT for detectors at twoTheta
The current version of `SpecularReflectionCalculateTheta` expects detectors to be at `theta`. However, most reflectometry algorithms now have a second version which expect detectors at `twoTheta` (see the parent issue #20327 for more detail).

We should also add a second version of `SpecularReflectionCalculateTheta` that works with detectors at `twoTheta`. This will avoid an extra [division by two](https://github.com/mantidproject/mantid/blob/master/Framework/Algorithms/src/ReflectometryReductionOneAuto2.cpp#L416) which is made in `ReflectometryReductionOneAuto2`, and will make more sense with the actual experimental set up:

setup

### Expected behavior
`SpecularReflectionCalculateTheta2` should return `twoTheta` when detectors are at `twoTheta`.

### Actual behavior
`SpecularReflectionCalculateTheta` returns `2*twoTheta` if detectors are at `twoTheta`.

Run this simple example of a point detector. The calculated `twoTheta` value should be the same as the `twoTheta` reported by the instrument.
```
import math

ws=Load(Filename='INTER43572')
twoTheta=SpecularReflectionCalculateTheta(InputWorkspace=ws,DetectorComponentName='point-detector')
print '  Calculated twoTheta = ' + str(twoTheta)

inst = ws.getInstrument()
det = inst.getDetector(4)
detTwoTheta = ws.detectorTwoTheta(det) * 180 / math.pi
print '  Detector 4 twoTheta =', detTwoTheta
```

gemmaguest:
This PR adds version 2 of `SpecularReflectionCalculateTheta`, which works with detectors at `twoTheta` (the correct position) instead of `theta` (where the reflectometry algorithms historically required detectors). See the parent issue #20327 for more context.

Version 2 is identical to version 1 apart from a factor of two difference, so the code has been refactored to do the calculation in the base class.

Fixes #20332.

**To test:**
Run the script in the linked issue.

**Release Notes** 
*Added to the release notes*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/src/SpecularReflectionAlgorithm.cpp
+++ b/Framework/Algorithms/src/SpecularReflectionAlgorithm.cpp
@@ -116,7 +116,7 @@ void SpecularReflectionAlgorithm::initCommonProperties() {
  */
 Mantid::Geometry::IComponent_const_sptr
 SpecularReflectionAlgorithm::getSurfaceSampleComponent(
-    Mantid::Geometry::Instrument_const_sptr inst) {
+    Mantid::Geometry::Instrument_const_sptr inst) const {
   std::string sampleComponent = "some-surface-holder";
   if (!isPropertyDefault("SampleComponentName")) {
     sampleComponent = this->getPropertyValue("SampleComponentName");
@@ -141,7 +141,7 @@ SpecularReflectionAlgorithm::getSurfaceSampleComponent(
  */
 boost::shared_ptr<const Mantid::Geometry::IComponent>
 SpecularReflectionAlgorithm::getDetectorComponent(
-    MatrixWorkspace_sptr workspace, const bool isPointDetector) {
+    MatrixWorkspace_sptr workspace, const bool isPointDetector) const {
   boost::shared_ptr<const IComponent> searchResult;
   if (!isPropertyDefault("SpectrumNumbersOfDetectors")) {
     const std::vector<int> spectrumNumbers =
@@ -197,6 +197,36 @@ bool SpecularReflectionAlgorithm::isPropertyDefault(
     const std::string &propertyName) const {
   Property *property = this->getProperty(propertyName);
   return property->isDefault();
+}
+
+/**
+ * Calculate the twoTheta angle from the detector and sample locations.
+ * @return: twoTheta
+ */
+double SpecularReflectionAlgorithm::calculateTwoTheta() const {
+  MatrixWorkspace_sptr inWS = this->getProperty("InputWorkspace");
+
+  const std::string analysisMode = this->getProperty("AnalysisMode");
+
+  Instrument_const_sptr instrument = inWS->getInstrument();
+
+  IComponent_const_sptr detector =
+      this->getDetectorComponent(inWS, analysisMode == pointDetectorAnalysis);
+
+  IComponent_const_sptr sample = this->getSurfaceSampleComponent(instrument);
+
+  const V3D detSample = detector->getPos() - sample->getPos();
+
+  boost::shared_ptr<const ReferenceFrame> refFrame =
+      instrument->getReferenceFrame();
+
+  const double upoffset = refFrame->vecPointingUp().scalar_prod(detSample);
+  const double beamoffset =
+      refFrame->vecPointingAlongBeam().scalar_prod(detSample);
+
+  const double twoTheta = std::atan(upoffset / beamoffset) * 180 / M_PI;
+
+  return twoTheta;
 }
 
 } // namespace Algorithms
--- a/Framework/Algorithms/src/SpecularReflectionAlgorithm.cpp
+++ b/Framework/Algorithms/src/SpecularReflectionAlgorithm.cpp
@@ -213,9 +213,8 @@ double SpecularReflectionAlgorithm::calculateTwoTheta() const {
   IComponent_const_sptr detector =
       this->getDetectorComponent(inWS, analysisMode == pointDetectorAnalysis);
 
-  IComponent_const_sptr sample = this->getSurfaceSampleComponent(instrument);
-
-  const V3D detSample = detector->getPos() - sample->getPos();
+  const V3D detSample =
+      detector->getPos() - inWS->spectrumInfo().samplePosition();
 
   boost::shared_ptr<const ReferenceFrame> refFrame =
       instrument->getReferenceFrame();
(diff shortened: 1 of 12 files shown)


gemmaguest:
I spoke to Max and he is happy with the new behaviour

Framework/Algorithms/src/SpecularReflectionAlgorithm.cpp
```diff
@@ -199,5 +199,35 @@ bool SpecularReflectionAlgorithm::isPropertyDefault(
...
+  const std::string analysisMode = this->getProperty("AnalysisMode");
+
+  Instrument_const_sptr instrument = inWS->getInstrument();
+
+  IComponent_const_sptr detector =
+      this->getDetectorComponent(inWS, analysisMode == pointDetectorAnalysis);
+
+  IComponent_const_sptr sample = this->getSurfaceSampleComponent(instrument);
+
+  const V3D detSample = detector->getPos() - sample->getPos();
```
soininen: Why not get the sample position simply by
```
inWS->spectrumInfo().samplePosition();
```

Framework/Algorithms/src/SpecularReflectionAlgorithm.cpp
```diff
@@ -199,5 +199,35 @@ bool SpecularReflectionAlgorithm::isPropertyDefault(
...
+
+  boost::shared_ptr<const ReferenceFrame> refFrame =
+      instrument->getReferenceFrame();
+
+  const double upoffset = refFrame->vecPointingUp().scalar_prod(detSample);
+  const double beamoffset =
+      refFrame->vecPointingAlongBeam().scalar_prod(detSample);
+
+  const double twoTheta = std::atan(upoffset / beamoffset) * 180 / M_PI;
```
soininen: Would it make sense to use `std::atan2(upoffset, beamoffset)` here?

Framework/Algorithms/src/SpecularReflectionCalculateTheta2.cpp
```diff
@@ -0,0 +1,72 @@
...
+/** Execute the algorithm.
+ */
+void SpecularReflectionCalculateTheta2::exec() {
+
+  const double twoTheta = calculateTwoTheta();
+
+  std::stringstream strstream;
+  strstream << "Recalculated two theta as: " << twoTheta;
+
+  this->g_log.information(strstream.str());
```
soininen: Instead of using `stringstream`, would it be simpler to use
```
g_log.information() << "Recalculated two theta as: " << twoTheta << '\n';
```

docs/source/algorithms/SpecularReflectionCalculateTheta-v2.rst
```diff
@@ -0,0 +1,62 @@
...
+   inst = ws.getInstrument()
+   ref_frame = inst.getReferenceFrame()
+   upoffset = ref_frame.vecPointingUp() 
+   det_position = {ref_frame.pointingUpAxis(): 1.0, ref_frame.pointingAlongBeamAxis(): 1.0, ref_frame.pointingHorizontalAxis():0}
+   MoveInstrumentComponent(ws, 'point-detector',RelativePosition=False, **det_position)
+   MoveInstrumentComponent(ws, 'some-surface-holder',RelativePosition=False,  X=0, Y= 0, Z=0)
+
+   # Calculate the two theta.
+   two_theta = SpecularReflectionCalculateTheta(InputWorkspace=ws, DetectorComponentName='point-detector', AnalysisMode='PointDetectorAnalysis')
+   print two_theta
```
soininen: To ease switching to python 3, `print(two_theta)`.

docs/source/algorithms/SpecularReflectionCalculateTheta-v2.rst
```diff
@@ -0,0 +1,62 @@
...
+
+Description
+-----------
+
+Uses the Specular reflection condition :math:`\theta_{In} \equiv \theta_{Out}`
+to calculate and return a corrected :math:`\theta_{In}`.
+
+.. math:: 
+
+   2\centerdot\theta = tan^{-1}\left(\frac{UpOffset}{BeamOffset}\right)
```
soininen: `tan` -> `\tan`

soininen (changes requested):
Generally, nice and welcome changes. The test script worked on a random INTER file I found on my HD.

Framework/Algorithms/src/SpecularReflectionCalculateTheta2.cpp
gemmaguest: I'll change this to use standard `string` concatenation as well because that's more efficient than `stringstream` operations

Framework/Algorithms/src/SpecularReflectionAlgorithm.cpp
```diff
@@ -213,9 +213,8 @@ double SpecularReflectionAlgorithm::calculateTwoTheta() const {
   IComponent_const_sptr detector =
       this->getDetectorComponent(inWS, analysisMode == pointDetectorAnalysis);
 
-  IComponent_const_sptr sample = this->getSurfaceSampleComponent(instrument);
-
-  const V3D detSample = detector->getPos() - sample->getPos();
+  const V3D detSample =
+      detector->getPos() - inWS->spectrumInfo().samplePosition();
```
soininen: I was a bit too quick to comment on this previously, sorry about that! If I understood correctly, there is a `SampleComponentName` property which enables a user to name a specific component as the sample. This property is read in `SpecularReflectionAlgorithm::getSurfaceSampleComponent()`, on the now-deleted line above. If `SampleComponentName` is not specified, `getSurfaceSampleComponent()` makes an educated guess that the instrument has a component called 'some-surface-holder'. I don't know how many instruments actually have a component with that name, but I do know that many instruments have a component specifically tagged as Sample in the IDF. So, what I should have proposed here in the first place, is to
- check if `SampleComponentName` is given
- if yes, find the component's position
- if no, use `SpectrumInfo::samplePosition()` because it is usually better guess than 'some-surface-holder'
In any case, this change should be reverted as it effectively disables `SampleComponentName`.

gemmaguest: Oops, I hadn't spotted this either! That's a good suggestion, and `SpecularReflectionPositionCorrect` should probably use the same method too. I'll create a separate issue for this and revert this change for now.

soininen (approved):

PR #20361, opened 2017-08-31
Small cleanup & speedup in init of MatrixWorkspace related to index handling

SimonHeybrock:
This is a follow up to ~~#20350~~, wait until it is merged.

This is a small cleanup and speedup of `MatrixWorkspace::initialize` and related code. A lot of this is related to the suboptimal duplicate index handling that I introduced a while ago (related to Instrument-2.0 and also MPI). The mechanism is still quite hard to understand, but this might be a small step in the right direction and fixes some performance bottlenecks.

Performance improvement example:

```python
    ws = LoadEmptyInstrument(InstrumentName='TOPAZ')
    ws = LoadEmptyInstrument(InstrumentName='TOPAZ')
    ws = LoadEmptyInstrument(InstrumentName='TOPAZ')
```

Each load speeds up by almost `0.4 s` on my system (around 30%, except for the first load which takes much longer since the full instrument is being loaded, so the relative change is much less). For smaller instruments this is less significant.

Creating workspaces in other places may also see a speedup, but it depends on how exactly it is being initialized.

**To test:**

Code review
No issue. Internal change -> no release notes.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/DataHandling/src/LoadEmptyInstrument.cpp
```diff
@@ -114,13 +114,15 @@ void LoadEmptyInstrument::exec() {
...
+    setProperty(
+        "OutputWorkspace",
+        create<EventWorkspace>(
+            *ws, indexInfo, BinEdges{0.0, std::numeric_limits<double>::min()}));
   } else {
     const double detector_value = getProperty("DetectorValue");
     const double monitor_value = getProperty("MonitorValue");
     auto ws2D = create<MatrixWorkspace>(
-        *ws, indexInfo, Histogram(BinEdges(2), Counts(1, detector_value),
+        *ws, indexInfo, Histogram(BinEdges{0.0, 1.0}, Counts(1, detector_value),
```
OwenArnold: And different bin edges here based on the 'MakeEventWorkspace' flag?

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -1984,20 +1986,24 @@ void MatrixWorkspace::rebuildDetectorIDGroupings() {
...
+        error = "MatrixWorkspace: SpectrumDefinition contains an out-of-range "
+                "time index for a detector, i.e., the spectrum definition does "
+                "not match the instrument in the workspace.";
+      } else {
+        detIDs.insert(allDetIDs[detIndex]);
+      }
     }
     spec.setDetectorIDs(std::move(detIDs));
   }
+  if (parallelException)
```
OwenArnold: Looks OK. code `if` and `else if` looks identical. Could probably be refactored out.

Framework/DataHandling/src/LoadEmptyInstrument.cpp
```diff
@@ -114,13 +114,15 @@ void LoadEmptyInstrument::exec() {
   Indexing::IndexInfo indexInfo(number_spectra);
   bool MakeEventWorkspace = getProperty("MakeEventWorkspace");
   if (MakeEventWorkspace) {
-    setProperty("OutputWorkspace",
-                create<EventWorkspace>(*ws, indexInfo, BinEdges(2)));
+    setProperty(
+        "OutputWorkspace",
+        create<EventWorkspace>(
+            *ws, indexInfo, BinEdges{0.0, std::numeric_limits<double>::min()}));
```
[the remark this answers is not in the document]
SimonHeybrock: These are the original parameters from `Workspace2D::init` and `EventWorkspace::init`.

Framework/DataHandling/src/LoadEmptyInstrument.cpp
SimonHeybrock: Yes, see above, default for `EventWorkspace` is different in `init`.

Framework/API/src/MatrixWorkspace.cpp
SimonHeybrock: The error message is actually different.

OwenArnold: Yep. Didn't spot that first time round.

OwenArnold (approved):

SimonHeybrock:
Note that this PR causes a performance regression of some performance tests such as `ChangePulsetimesTestPerformance`. The reason is `createEventWorkspaceWithStartTime` from `WorkspaceCreationHelper` which adds default detector IDs and thus cause slow failures via exceptions when updating the grouping. This should not influence production code.

PR #20367, opened 2017-09-01
Enable FindPeaks to work for MatrixWorkspace with various X length

### LINKED ISSUE #20365: FindPeaks cannot work with MatrixWorkspace with various number of bins in auto mode
FindPeaks cannot work with MatrixWorkspace if all of its spectra do not have same number of bins in the automatic peak finding mode.  

### Actual behavior

```
ws = LoadGSS(Filename='156005.gda')
FindPeaks(InputWorkspace=ws, PeaksList='output', WorkspaceIndex=2)
```
```
RuntimeError                              Traceback (most recent call last)
<ipython-input-2-287db394e362> in <module>()
----> 1 FindPeaks(InputWorkspace=ws, PeaksList='output', WorkspaceIndex=2)

/home/wzz/Mantid_Project/debug/bin/mantid/simpleapi.pyc in algorithm_wrapper(InputWorkspace,WorkspaceIndex=None,FWHM=None,Tolerance=None,PeakPositions=None,FitWindows=None,PeakFunction=None,BackgroundType=None,HighBackground=None,MinGuessedPeakWidth=None,MaxGuessedPeakWidth=None,GuessedPeakWidthStep=None,PeakPositionTolerance=None,RawPeakParameters=None,MinimumPeakHeight=None,MinimumPeakHeightObs=None,CostFunction=None,Minimizer=None,StartFromObservedPeakCentre=None, Version=1)
    987         set_properties(algm, *args, **final_keywords)
    988         try:
--> 989             algm.execute()
    990         except RuntimeError as e:
    991             if e.args[0] == 'Some invalid Properties found':

RuntimeError: Histogram::setSharedX: size mismatch
```

wdzhou:
Enable FindPeaks to work with MatrixWorkspace that has spectra with various lengths in auto peak finding mode.

**To test:**

Fixes #20365. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Algorithms/src/FindPeaks.cpp
```diff
@@ -462,8 +463,12 @@ void FindPeaks::findPeaksUsingMariscotti() {
...
 
   for (int k = start; k < end; ++k) {
-    const auto &S = smoothedData->y(k);
-    const auto &F = smoothedData->e(k);
+    int ws_i(k);
+    if (singleSpectrum)
+      ws_i = 0;
+
+    const auto &S = smoothedData->y(ws_i);
+    const auto &F = smoothedData->e(ws_i);
```
SimonHeybrock: I don't understand what is happening here, why was this changed?

Framework/Algorithms/src/FindPeaks.cpp
```diff
@@ -632,24 +637,52 @@ void FindPeaks::findPeaksUsingMariscotti() {
 API::MatrixWorkspace_sptr FindPeaks::calculateSecondDifference(
     const API::MatrixWorkspace_const_sptr &input) {
   // We need a new workspace the same size as the input ont
-  MatrixWorkspace_sptr diffed = WorkspaceFactory::Instance().create(input);
+  MatrixWorkspace_sptr diffed = 0;
```
SimonHeybrock: I think if you use the factory functions I provide in `DataObjects/WorkspaceCreation.h` like this:
```cpp
auto diffed = create<MatrixWorkspace>(input);
```
it will automatically copy all X arrays for you and you can completely remove all the code duplication you added below (unless I am overlooking something), and remove the line
```cpp
    // Copy over the X values
    diffed->setSharedX(i, input->sharedX(i));
```
below.

The only actual thing that needs to be done is adapt those factory methods to work in a way that permits variable X lengths, namely use `setHistogram` instead of `setSharedX` here: https://github.com/mantidproject/mantid/blob/master/Framework/DataObjects/inc/MantidDataObjects/WorkspaceCreation.h#L267 (detail to figure out: how to set the correct Y/E values, let me know if you have trouble with that).

SimonHeybrock (dismissed):
I think this code duplication is not the right way to go (and you would have the same problem in many other algorithms). But I think there is a better and shorter solution, see comment below:

Framework/Algorithms/src/FindPeaks.cpp
wdzhou: Simon, Thanks.  I will refactor the code according to your suggestion.

Framework/Algorithms/src/FindPeaks.cpp
wdzhou: The old way to find peaks in automatic way includes creating a MatrixWorkspace with same number of histograms and vectors of X to calculate derivatives.  If FindPeaks only works on 1 spectrum, it wastes resources to do such.  So instead, my solution is to create a 1-spectrum MatrixWorkspace.

Framework/Algorithms/src/FindPeaks.cpp
wdzhou: I had the same thoughts here too. As it wastes resources to produce a multi-spectrum MatrixWorkspace if there is only one spectrum to find peaks at, I create a single-spectrum MatrixWorkspace other than a multip-spectrum workspace.

wdzhou: I tried to use 
```auto diffed = create<MatrixWorkspace>(input)```
but I got compiling error as
```
/Users/wzz/Mantid/Framework/Algorithms/src/FindPeaks.cpp:659:14: error: no matching function for call to 'create'
    diffed = Mantid::DataObjects::create<API::MatrixWorkspace>(input);
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/wzz/Mantid/Framework/DataObjects/inc/MantidDataObjects/WorkspaceCreation.h:260:35: note: candidate template ignored: disabled by 'enable_if' [with T = Mantid::API::MatrixWorkspace, P = boost::shared_ptr<Mantid::API::MatrixWorkspace>]
          typename std::enable_if<std::is_base_of<API::MatrixWorkspace,
```

On the other hand, so far, in master, this is what I see for method create<>(parent):
```
template <class T, class P,
          typename std::enable_if<std::is_base_of<API::MatrixWorkspace,
                                                  P>::value>::type * = nullptr>
std::unique_ptr<T> create(const P &parent) {
  const auto numHistograms = parent.getNumberHistograms();
  auto ws =
      create<T>(parent, numHistograms, detail::stripData(parent.histogram(0)));
  for (size_t i = 0; i < numHistograms; ++i) {
    ws->setSharedX(i, parent.sharedX(i));
  }
  return ws;
}
```

**Comment by SimonHeybrock** [comment]
@wdzhou Reviews should NOT be dismissed if you replied to them.

File: Framework/Algorithms/src/FindPeaks.cpp (hunk 2 above)
Comment (SimonHeybrock) [reply Framework/Algorithms/src/FindPeaks.cpp]: @wdzhou Yes, the methods uses a reference to the parent, not a `shared_ptr`.

File: Framework/Algorithms/src/FindPeaks.cpp (hunk 1 above)
Comment (SimonHeybrock) [reply Framework/Algorithms/src/FindPeaks.cpp]: Ok that makes sense :+1:. However below you duplicated a lot of code below for this purpose (calculating the derivative), I think that can and should be avoided to avoid maintenance nightmares.

If you observe an actual speedup I think that is also worth mentioning in the release notes.

**Review by SimonHeybrock — CHANGES_REQUESTED** [review]
See first (accidentally dismissed) review for comments.

File: Framework/Algorithms/src/FindPeaks.cpp (hunk 1 above)
Comment (wdzhou) [reply Framework/Algorithms/src/FindPeaks.cpp]: I refactored all the duplicate codes (hopefully).  On the other hand, I don't think that the speedup can be significant unless the workspace to find peaks has huge number of spectra.

File: Framework/Algorithms/src/FindPeaks.cpp [hunk 3]
```diff
@@ -462,8 +464,12 @@ void FindPeaks::findPeaksUsingMariscotti() {
   const int blocksize = static_cast<int>(smoothedData->blocksize());
 
   for (int k = start; k < end; ++k) {
-    const auto &S = smoothedData->y(k);
-    const auto &F = smoothedData->e(k);
+    int ws_i(k);
+    if (singleSpectrum)
+      ws_i = 0;
```
SimonHeybrock: Why do you need these changes, aren't `start` and `end` above set up correctly to do this already?

SimonHeybrock: Why do you need this? You are just undoing what the computation of `start` and `end` above did. Can't you just get rid of `start` and `end` and simply loop over all spectra in the workspace here?

SimonHeybrock:
@wdzhou Overall the way you distinguish between a single spectrum and a full workspace is still adding a lot of extra code and makes things hard to read. I think it can be cleaned up nicely however, but a few small adaptations:

- Return `std::vector<Histogram>` from `calculateSecondDifference` instead of a workspace.
- Use `m_indexSet = m_dataWS->indexInfo().makeIndexSet(...)` to generate the indices you need to process (in the current implementation this is either the full workspace, obtained by `makeIndexSet()`, or a single spectrum, obtained, e.g., by makeIndexSet({m_wsIndex}).
- Once you have the set you do not need any code that distinguishes between the cases, for example:

```cpp
void FindPeaks::calculateStandardDeviation(
    const API::MatrixWorkspace_const_sptr &input,
    std::vector<Histogram> &smoothed, const int &w) {
  // Guard against anyone changing the value of z, which would mean different
  // phi values were needed (see Marriscotti p.312)
  static_assert(g_z == 5, "Value of z has changed!");
  // Have to adjust for fact that I normalise Si (unlike the paper)
  const int factor = static_cast<int>(std::pow(static_cast<double>(w), g_z));

  const double constant =
      sqrt(static_cast<double>(this->computePhi(w))) / factor;

  for (size_t i = 0; i < m_indexSet.size(); ++i) {
    smoothed[i].setSharedE(input->sharedE(m_indexSet[i]));
    std::transform(smoothed[i].e().cbegin(), smoothed[i].e().cend(),
                   smoothed[i].mutableE().begin(),
                   std::bind2nd(std::multiplies<double>(), constant));
  }
}
```

Note that this has much less (or no) added code than required if you handle this by hand without `IndexSet`.

Aside: There is an `operator*` for `HistogramE` and `double`, so you can probably get rid of the `std::transform` as well.

SimonHeybrock:
@wdzhou You seem to have removed code that deals with non-constant histogram sizes. Will you add this (as in my main comment of the first review), or has the purpose of this PR changed?

Framework/Algorithms/src/FindPeaks.cpp
wdzhou: I need to check!

wdzhou:
@SimonHeybrock FindPeaks did not work with workspaces with non-constant histogram sizes.  This is what I am trying to fix.  The issue was caused by calling setSharedX() and setSharedE().

wdzhou:
@SimonHeybrock FindPeaks did not work with workspaces with non-constant histogram sizes.  This is what I am trying to fix.  The issue was caused by calling setSharedX() and setSharedE().

wdzhou:
@SimonHeybrock FindPeaks calls child algorithm SmoothData.  I don't think that there is a way to use a vector of Histograms to replace the MatrixWorkspace object for smoothed second order derivatives.

On the other hand, I do accept your suggestion to use indexSet to simply the code.

SimonHeybrock:
@wdzhou I pushed some change here: https://github.com/mantidproject/mantid/pull/20413. I think it is doing the same thing as this PR, but removing 10 lines of code instead of adding 50. Could you have a look if this works? I do not have the file you are referring to in the bug report, so I could not test that part. Is it available somewhere?

wdzhou:
Issue #20365 is fixed by PR #20413. Thus this PR is closed without merging.

PR #20372, opened 2017-09-03
Fix recursive runPythonScript in async mode

martyngigg:
**Description of work**

Fixes a subtle race condition when using `runPythonScript(..., async=True)` inside an already executing python script. 

**To test:**

I tested this by saving this script:

```
import mantidplot

for i in range(10):
    print("before")
    mantidplot.runPythonScript("CreateSampleWorkspace(OutputWorkspace='test{}')".format(i), True)
    print("after")
```

to a file and then using `View->Manage Custom Menus` to add this script to a new menu.

Before this fix MantidPlot will hang on running the script.

Another test would be to add a similar line to the startup of a Python GUI.

*No issue*

**Release Notes** 

*Broken  Does not need to be in the release notes.*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


martyngigg:
@mdoucet I think this should fix the problem you are seeing with MantidPlot.

mdoucet (approved):
I tried it out and it now works fine.

PR #20373, opened 2017-09-04
Indirect - AbsorptionCorrections - Shape Specific Monte Carlo Calculation of Absorption Factors

### LINKED ISSUE #20184: Create shape specific monte carlo absorption corrections algorithm
Create an algorithm that calculates absorption factors from both the container and sample. Currently, supported shapes will be:

- Flat Plate
- Annulus
- Cylinder

Absorption factors for the sample and container will be calculated by SimpleShapeAbsorptionCorrections.

brandonhewer:
A monte carlo algorithm for calculating the absorption factors from the container and sample, given the shape of the container has been created: 'CalculateMonteCarloAbsorption'. 

SimpleShapeMonteCarloAbsorption has been modified to allow for taking a property, 'Elastic', which when true, will result in Elastic Absorption Corrections being calculated, using an elastic instrument definition file.

CalculateMonteCarloAbsorption can currently support the following shapes:
- Flat Plate
- Annulus
- Cylinder

**To test:**
1. Load the Sample and Container files into the ADS (supplied below).
2. Run the CalculateMonteCarloAbsorptionCorrections algorithm from the Algorithms window, on the created Sample and Container workspaces, for each shape. Example valid property values (NOTE: set the Shape property first, such that the correct properties are shown):
    - **FlatPlate**
        - _BeamHeight_: 3.5
        - _BeamWidth_: 4.0
        - _EventsPerPoint_: 200
        - _SampleChemicalFormula_: H2-0
        - _SampleDensityType_: Mass Density
        - _SampleDensity_: 1.0
        - _ContainerChemicalFormula_: Al
        - _ContainerDensityType_: Mass Density
        - _ContainerDensity_: 1.0
        - _Height_: 2.0
        - _SampleWidth_: 2.0
        - _SampleThickness_: 2.0
        - _ContainerFrontThickness_: 1.5
        - _ContainerBackThickness_: 1.5
    - **Cylinder**
        - _BeamHeight_: 3.5
        - _BeamWidth_: 4.0
        - _EventsPerPoint_: 200
        - _SampleChemicalFormula_: H2-0
        - _SampleDensityType_: Mass Density
        - _SampleDensity_: 1.0
        - _ContainerChemicalFormula_: Al
        - _ContainerDensityType_: Mass Density
        - _ContainerDensity_: 1.0
        - _Height_: 2.0
        - _SampleRadius_: 0.5
        - _ContainerInnerRadius_: 1.0
        - _ContainerOuterRadius_: 2.0
    - **Annulus**
        - _BeamHeight_: 3.5
        - _BeamWidth_: 4.0
        - _EventsPerPoint_: 200
        - _SampleChemicalFormula_: H2-0
        - _SampleDensityType_: Mass Density
        - _SampleDensity_: 1.0
        - _ContainerChemicalFormula_: Al
        - _ContainerDensityType_: Mass Density
        - _ContainerDensity_: 1.0
        - _Height_: 2.0
        - _SampleInnerRadius_: 1.2
        - _SampleOuterRadius_: 1.8
        - _ContainerInnerRadius_: 1.0
        - _ContainerOuterRadius_: 2.0
3. For each run of the algorithm, ensure that a \*\_ass workspace and a \*\_acc are contained within the Output Group Workspace (with user specified name - specified by property _CorrectionsWorkspace_).

**Test files:**
[test-files.zip](https://github.com/mantidproject/mantid/files/1305812/test-files.zip)

Fixes #20184. 

**Release Notes** 
Changes added to release notes. Full algorithm documentation and workflow diagram to be added in a separate PR.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


SpencerHowells:
@brandonhewer We have 2 forms of 'Elastic'. If Emode=elastic in the input ws then it should just ConvertUnit to wavelength. Similarly for Emode = indirect or direct. However If Emode=indirect there is an extra option of IndirectElastic which will use the extra code I sent Creating the wavelength ws.

brandonhewer:
@SpencerHowells Will make the relevant changes, thanks for the explanations :)

brandonhewer:
Would your code for converting to wavelength be better placed in CalculateMonteCarloAbsorption (this is where conversion to wavelength seems to currently be done)?

SpencerHowells:
@brandonhewer In my code if input ws has Emode=Indirect it automatically checks for  q-axis to set IndirectElastic so there's no need for it to be input.

brandonhewer:
@SpencerHowells Would this mean, in the case of an input workspace with Emode=Indirect, and whose X-Axis or Y-Axis is in units of Momentum Transfer, IndirectElastic should always be assumed?

SpencerHowells:
@brandonhewer yes for both.

SpencerHowells:
@brandonhewer Regarding second point. Just realised that an S(Q,w) ws would have the y-unit = MomentumTransfer. A better test for IndirectElastic would be x-unit not EnergyTransfer and then test which unit is Q.

SpencerHowells:
@brandonhewer For testing:
Normal mode: sample iris26176, H2O, density=1 and can iris26174, V, density=6
Elastic: just sample osi104367_elf, H20

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/CalculateMonteCarloAbsorption.py
```diff
@@ -0,0 +1,670 @@
...
+        waves = (0.01 * np.arange(-1, workspace.blocksize())) + wave
+        logger.information('Waves : ' + str(waves))
+        nhist = workspace.getNumberHistograms()
+        for idx in range(nhist):
+            workspace.setX(idx, waves)
+        self._change_angles(workspace, self._q_values, wave)
+
+        return workspace
+
+    def _change_angles(self, workspace, q_values, wave):
```
gvardany: missing doxygen? what does this function do exactly? Is this also something ISIS-specific?

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/CalculateMonteCarloAbsorption.py
```diff
@@ -0,0 +1,670 @@
...
+        of the input workspace must be in units of Q.
+
+        :param workspace:   The input workspace.
+        :return:            The output wavelength workspace.
+        """
+        self._q_values = workspace.getAxis(1).extractValues()
+
+        # ---------- Load Elastic Instrument Definition File ----------
+
+        idf_name = workspace.getInstrument().getName() + '_elastic_Definition.xml'
```
gvardany: This part seems ISIS-bound. We do not have any `elastic_Definition` for ILL instruments. But this algorithm should run as well. Can you explain what it is, and why do you need a special IDF for indirect elastic?

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/CalculateMonteCarloAbsorption.py
```diff
@@ -0,0 +1,670 @@
...
+            analyser_comp = inst.getComponentByName(inst.getStringParameter('analyser')[0])
+
+            if analyser_comp is not None and analyser_comp.hasParameter('Efixed'):
+                return analyser_comp.getNumberParameter('EFixed')[0]
+
+        raise ValueError('No Efixed parameter found')
+
+    # ------------------------------- Converting to/from wavelength -------------------------------
+
+    def _convert_to_wavelength(self, workspace):
```
gvardany: I agree, the name is self-explanatory here, but doxygens should also be added.

Framework/PythonInterface/test/python/plugins/algorithms/WorkflowAlgorithms/CalculateMonteCarloAbsorptionTest.py
```diff
@@ -0,0 +1,144 @@
...
+    def _run_correction_with_container_test(self, shape):
+
+        if shape == 'FlatPlate':
+            self._setup_flat_plate_container()
+        else:
+            self._setup_annulus_container()
+
+        self._run_correction_and_test(shape)
+
+    def _run_indirect_elastic_test(self, shape):
```
gvardany: Here you are overriding the member variables, so this tests can be run only after all the others. Moreover you are calling this 3 times now and loading the same file trice, but only the shapes are different. This is not desired. I would suggest to have separate member variables for this case, and set them up also in `setUp`, so the tests are more decoupled.

Framework/PythonInterface/test/python/plugins/algorithms/WorkflowAlgorithms/CalculateMonteCarloAbsorptionTest.py
```diff
@@ -0,0 +1,144 @@
...
+    def tearDown(self):
+        DeleteWorkspace(self._red_ws)
+
+        if self._container_ws is not None:
+            DeleteWorkspace(self._container_ws)
+        if self._indirect_elastic_ws is not None:
+            DeleteWorkspace(self._indirect_elastic_ws)
+
+    def _setup_container(self):
+        container_ws = Load('irs26173_graphite002_red.nxs')
```
gvardany: You call this twice, once for each shape, but you should load only once. I would move these to `setUp` just like yo do with the `red_ws`.

gvardany (changes requested):
Nice work, generally well written algorithm.
My main concern is about indirect elastic mode.
Tests could also be slightly improved, and I would like to have some reduced data from ILL also tested in the unit tests. I think there are some in the unit tests directory, I will point you, or I can send you the files.
In the separate PR for documentation, flowchart and usage examples, I would also add a system test that actually compares the values of corrections with the reference, possibly with very relaxed tolerances. Here we assert only on the dimensions and units.
I will proceed to functional tests shortly.

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/CalculateMonteCarloAbsorption.py
```diff
@@ -0,0 +1,670 @@
...
+        else:
+            convert_unit_alg = self.createChildAlgorithm("ConvertUnits", enableLogging=True)
+
+            if self._emode == 'Indirect':
+                x_unit = workspace.getAxis(0).getUnit().unitID()
+                y_unit = workspace.getAxis(1).getUnit().unitID()
+
+                # Check whether to create wavelength workspace for Indirect Elastic
+                if (x_unit == 'MomentumTransfer' and not y_unit == 'EnergyTransfer') \
+                        or (y_unit == 'MomentumTransfer' and not x_unit == 'EnergyTransfer'):
```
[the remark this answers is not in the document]
brandonhewer: It is possible that the workspace will be transposed - such that EnergyTransfer is in the y-axis. The algorithm should apply in both cases.

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/CalculateMonteCarloAbsorption.py
```diff
@@ -0,0 +1,670 @@
...
+
+            if self._shape == 'Annulus':
+                if self._container_inner_radius >= self._sample_inner_radius:
+                    issues['ContainerInnerRadius'] = 'Must be less than SampleInnerRadius'
+                if self._container_outer_radius <= self._sample_outer_radius:
+                    issues['ContainerOuterRadius'] = 'Must be greater than SampleOuterRadius'
+
+        return issues
+
+    def _get_efixed(self):
```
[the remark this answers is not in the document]
brandonhewer: This sounds like a very sensible suggestion, I'll see if I can find it's other uses and move it to IndirectCommon.

gvardany:
Great, thanks. I see the point about transposing, but ILL fixed window scan reduction produces a workspace with spectrum number as y-axis (similar to QENS), and an empty unit (e.g. temperature) as x-axis. So none of the axes is energy or momentum transfer, and the correction needs to be computed in elastic approximation for the wavelength derived from EFixed. If we could just make this algorithm run also for this case (see file attached) that would save us a lot of grief. 
[ill-efws.tar.zip](https://github.com/mantidproject/mantid/files/1310225/ill-efws.tar.zip)

gvardany:
So in my mind the unit branching goes like that:
```
if x_unit is 'Wavelength':
    // do nothing, proceed to calculation directly
else if x_unit is 'EnergyTransfer':
   // convert to wavelength, proceed to calculation
else if x_unit is 'Q' or 'Qˆ2':
   if y_unit is 'Wavelength':
       //transpose and proceed to calculation
   elif y_unit is 'EnergyTransfer':
      //transpose, convert to wavelength and proceed to calculation
else:
    // create single wavelength workspace from EFixed, and proceed to calculation
```

Will this fit in with what you need?

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/CalculateMonteCarloAbsorption.py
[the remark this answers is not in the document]
brandonhewer: CalculatePaalmanPings does reimplement getEFixed, however it is not a Python algorithm.

SpencerHowells:
@brandonhewer @gvardany Getting efixed. The code in IndirectCommon was written a long time ago. More recently it has been possible to use ws.efixed. However I found that it was inconsistent: sometimes it worked, sometimes it didn't. I don't know why - perhaps it is to do do with reading its value from the IPF. It would be nice to soret it out so that we can be consistent.

SpencerHowells:
@brandonhewer @gvardany Elastic mode. The MCabsorption algo requires the angles corresponding to the spectra. For ISIS the users wanted to create files in Q and this can be the x or y axis. Detector angles can also be summed so the angle corresponding to the Q value is not be that from the IDF. So I calculate the angles from Q and modify the detector table. However this still has the monitors in it and so does not do it correctly. I therefore created the *elastic* IPF by removing the monitors and it now works. There are new IPF for IRIS, OSIRIS & IN16b. We should certainly test it for IN16b.

SpencerHowells:
@gvardany When your elastic is in angles, do you sum detectors?

gvardany:
@SpencerHowells In all the cases we do sum the detectors by grouping the vertical pixels in the tubes, so they all have the same scattering angle in the scattering plane. If MC absorption actually takes into account the off-plane angle as well (I guess it does), then you are right this will not be so correct. But my guess would be that this has a tiny effect (for IN16B), since the height of the detector is not much, and it is placed twice farther in the analyser's position in mantid, so the off-plane angle is small which will translate to a small change in the neutron path in the sample. I will try to quantify this effect. For more precise corrections one should consider computing them before grouping the detectors, which is currently difficult to imagine in our workflows. So I think your workaround is justified.

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/CalculateMonteCarloAbsorption.py
```diff
@@ -0,0 +1,775 @@
...
+                                'NumberOfWavelengthPoints': self.getProperty('NumberOfWavelengthPoints').value,
+                                'EventsPerPoint': self.getProperty('EventsPerPoint').value,
+                                'Interpolation': self.getProperty('Interpolation').value}
+
+        self._sample_ws = self.getProperty("SampleWorkspace").value
+        self._container_ws = self.getProperty('ContainerWorkspace').value
+        self._shape = self.getProperty('Shape').value
+        self._height = self.getProperty('Height').value
+
+        self._sample_unit = self._sample_ws.getAxis(0).getUnit().unitID()
```
gvardany: `getAxis` throws an error for workspace groups, and the error message is not too nice to show to the user. Since this algorithm is not intended to run over workspace groups I would suggest have an explicit check for that in `validateInputs`, and raise a descriptive error message there. The same applies to container workspace, since L434 will complain for a workspace group.

brandonhewer: Would it be useful for the algorithm to be able to run sequentially across WorkspaceGroups? This could be done by skipping these setup checks when the Workspace is a WorkspaceGroup; the checks are done first for the WorkspaceGroup and then each Workspace in turn as they are sequentially used in the algorithm.

gvardany: Generally the algorithms should transparently run over groups, but in this case it is debatable, since the output itself is a group workspace. Also different workspaces are normally for different samples, so allowing to run the same absorption over a group might be error prone. So I think we should rather forbid it here, and also later in the interface tab.

brandonhewer: For the former point, multiple WorkspaceGroups would be created in the case of supplying a WorkspaceGroup as input.
As long as the other properties in setup are initialized per workspace in the workspace group, the latter shouldn't be an issue, as each run will be independent of the previous and subsequent runs.
The checks that appear important are whether the container and sample workspace types are matched and have the same length, when they are both supplied.

SpencerHowells:
@gvardany I agree that your workaround for off-plane should be ok. I was more concerned about adding angles, which we do, leading to a mismatch between the angles for the spectra and those in the detector table. I've had a look at the file you sent and it appears to be ok. I'm surprised that from a scientific point of view you do not convert to Q.

gvardany:
Eventually it is converted to Q, actually now we have the option directly in the reduction algorithm. So the output can be either. This means @brandonhewer one more else clause to the branching, that I forgot:

```
if x_unit is 'Wavelength':
    // do nothing, proceed to calculation directly
elif x_unit is 'EnergyTransfer':
   // convert to wavelength, proceed to calculation
elif x_unit is 'Q' or 'Qˆ2':
   if y_unit is 'Wavelength':
       //transpose and proceed to calculation
   elif y_unit is 'EnergyTransfer':
      //transpose, convert to wavelength and proceed to calculation
   else:
      // elastic, create single wavelength workspace
else:
    // create single wavelength workspace from EFixed, and proceed to calculation
``` 

Regarding the off-plane approximation, actually it seems to have a large effect. For the most off-plane pixel I get MC correction that is 3 times different from the in-plane tube centre. I am not sure however how much will be the difference between summed pixels corrected with in-plane factor, or each pixel  corrected properly and then summed. 
Anyway this flows beyond the scope of this PR.
For the moment we can just bracket the `_elasticDefinition` IDF stuff with `if instrument is (OS)IRIS` and do nothing special with the angles for ILL case.

SpencerHowells:
@gvardany What is the maximum off-plane angle? How does that give a factor of 3? I was thinking more of  low angles where the off-plane will change the Q values.

brandonhewer:
@gvardany Why would the check for whether the Y-Axis is in Wavelength or DeltaE only be performed if X is Q or Q^2. Surely this is simply a transposition of case 1 and 2 whereby the X-Axis is in Wavelength or DeltaE? So:

```
if x_unit is 'Wavelength':
    // do nothing, proceed to calculation directly
elif x_unit is 'EnergyTransfer':
   // convert to wavelength, proceed to calculation
if y_unit is 'Wavelength':
    // transpose and proceed to calculation
elif y_unit is 'EnergyTransfer':
   // transpose, convert to wavelength and proceed to calculation
elif emode is "Indirect":
  if x_unit is 'Q' or 'Qˆ2':
     // elastic, create single wavelength workspace
  elif y_unit is 'Q' or 'Q^2':
    // elastic, transpose and create single wavelength workspace
  else:
    // Indirect inelastic, create single wavelength workspace from EFixed
else:
   // create single wavelength workspace
```

SpencerHowells:
@brandonhewer @gvardany regarding the version with Q values, it should already have the correct angles so the updating angles option is not required. It just needs creating the x value in wavelength.

SpencerHowells:
@brandonhewer @gvardany we still need, I assume, to check that the y-axis is in Q.

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/CalculateMonteCarloAbsorption.py
```diff
@@ -0,0 +1,796 @@
...
+
+        if self._emode == "Indirect":
+
+            # Check for Indirect Elastic (units of Q)
+            if y_unit == 'MomentumTransfer':
+                return self._create_waves_indirect_elastic(workspace)
+            elif x_unit == 'MomentumTransfer':
+                self._transposed = True
+                return self._create_waves_indirect_elastic(self._tranpose_ws(workspace))
+            return self._convert_units(workspace, "Wavelength", self._emode, self._efixed)
```
gvardany: It's this line that I wanted also to be `return self._create_waves_indirect_elastic(workspace)`. I.e. if it is indirect, and none of the units is `MomentumTransfer`, fall back to indirect elastic.
The rest seems fine.

brandonhewer: If neither axis are in Q, is this case not an Indirect Inelastic case?
In this situation, does it still make sense to use _create_waves_indirect_elastic?

SpencerHowells: The IN16b elastic case can be x=Temperature & y=Spectrum

brandonhewer: How can elastic cases be distinguished from inelastic, or do they not need to be?
Indirect Inelastic cases will be picked up by create_waves_indirect_elastic by making this change.

SpencerHowells: Inelastic will have x=EnergyTransfer

brandonhewer: If x (or y in case of transposition) is EnergyTransfer, workspace is currently converted to Wavelength using the ConvertUnits algorithm with EFixed, so it will never reach this point in the code. 
Are we considering any case where x (or y in case of transposition) isn't EnergyTransfer and emode is Indirect, to be Indirect Elastic?

SpencerHowells: If emode is Indirect and x (or y) is not EnergyTransfer it is Indirect Elastic. ISIS then has Q and ILL Spectrum.

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/CalculateMonteCarloAbsorption.py
```diff
@@ -0,0 +1,794 @@
...
+
+        # The beam properties and monte carlo properties are simply passed straight on to the
+        # SimpleShapeMonteCarloAbsorptionCorrection algorithm so they are being put into
+        # a dictionary for simplicity
+        sample_workspace = self.getProperty("SampleWorkspace").value
+        container_workspace = self.getProperty("ContainerWorkspace").value
+        sample_is_group = isinstance(sample_workspace, WorkspaceGroup)
+        container_is_group = isinstance(container_workspace, WorkspaceGroup)
+
+        if container_workspace and sample_is_group != container_is_group:
```
gvardany: Not critical, but these condition blocks read hard, any way to streamline better?

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/CalculateMonteCarloAbsorption.py
```diff
@@ -0,0 +1,794 @@
...
+                                'NumberOfWavelengthPoints': self.getProperty('NumberOfWavelengthPoints').value,
+                                'EventsPerPoint': self.getProperty('EventsPerPoint').value,
+                                'Interpolation': self.getProperty('Interpolation').value}
+
+        self._sample_ws = sample_workspace
+        self._container_ws = container_workspace
+        self._shape = self.getProperty('Shape').value
+        self._height = self.getProperty('Height').value
+
+        self._sample_unit = self._sample_ws.getAxis(0).getUnit().unitID()
```
gvardany: Now when you tried to get it working for groups, this causes a problem, because `self._sample_ws` does not have getAxis for groups. I get `None type has no attribute getAxis()` at validation step.
I think this should be wrapped in a check, we are dealing with group workspaces, we should take the unit probably from the first workspace in the group.

gvardany (changes requested):
Updates are good, except:
- it does not work for group workspaces
- for indirect inelastic (the file that I send earlier) it modifies the input workspace itself (check the x-axis and the unit) - something is mishandled in conversions/transpositions.

After fixed, please add 3 more unit tests:
- group ws
- ILL elastic indirect (put the file that I sent in ILL/IN16B subfolder in unit tests)
- ILL indirect QENS, there is a file already ILL/IN16B/091515_red.nxs

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/CalculateMonteCarloAbsorption.py
brandonhewer: This step should never actually be reached for the group workspace itself -> it will later be done for each workspace in turn.

brandonhewer: Could you give me the steps you took to produce this issue?

gvardany:
The input workspace is intact now.
This tarball contains two files, one is a group, for which I get the error described when choosing the group as sample workspace. 
For the other one (elastic indirect) now I get different error during cropping.
[error-examples.tar.zip](https://github.com/mantidproject/mantid/files/1324491/error-examples.tar.zip)

NickDraper:
@brandonhewer This has conflict that need addressing, and I'm not sure if the  final review comments have been addressed.

brandonhewer:
@gvardany I've added the fix for allowing WorkspaceGroups. The other issue appears to be a difference in how the IN16B workspace and ISIS workspaces are formatted at this point.

brandonhewer:
@gvardany The X-Values are in reverse order in the IN16B workspace?

gvardany:
For a group workspace with QENS, now mantid crashes during the execution, here is the tip of the trace:
```
0   libMantidAPI.dylib            	0x0000000102927d5f Mantid::API::Algorithm::reportCompleted(double const&, bool) + 1855 (Algorithm.cpp:1617)
1   libMantidAPI.dylib            	0x0000000102923a5d Mantid::API::Algorithm::doCallProcessGroups(Mantid::Kernel::DateAndTime&) + 1901 (Algorithm.cpp:1253)
2   libMantidAPI.dylib            	0x00000001029214af Mantid::API::Algorithm::execute() + 6079 (Algorithm.cpp:539)
```
Could you have a look and add a unit test with a group workspace in this PR? Tests with the ILL files can be added in the follow up ticket.

brandonhewer:
@gvardany Did you receive this error using the group workspace included above? I've been unable to reproduce it.

gvardany:
@brandonhewer It seems to work fine now, I am not sure what happened then, probably incomplete/mixed build with another branch...

gvardany:
@brandonhewer I tried again, strangely enough, it works if there are two workspaces in the group, and crashes if there are more. Can you try with more than 2?

brandonhewer:
@gvardany I can reproduce this, not entirely sure where it's going wrong yet.

samueljackson92:
I think this odd behaviour might be coming from trying to insert the group output from the child algorithm into the parent group workspace. While this should logically work, an appears to work for 2 workspaces, I'm not 100% convinced that mantid handles this correctly. 

As we're quite close to release I've disabled group workspaces for now so we can get this in. We can take another look during the next release cycle about how to properly handle this with more time on our hands.

gvardany (approved):
Sounds good to me. Indeed the group workspace issue seemed a problem deeper in mantid.

Framework/API/src/WorkspaceGroup.cpp
```diff
@@ -224,6 +224,38 @@ void WorkspaceGroup::print() const {
   }
 }
 
+/*
+ * @return  A non-const iterator pointing to the first workspace in this
+ *          workspace group.
+ */
```
ianbush: For doxygen comments this needs to start with `/**` not `/*` . Could you fix this in this file and the Python exports?

```
/**
 * @return  A non-const iterator pointing to the first workspace in this
 *          workspace group.
 */

ianbush:
I think the failing test on Windows is unrelated.

PR #20374, opened 2017-09-04
Convert instrument view miniplot to a matplotlib-based widget

### LINKED ISSUE #20240: Convert instrument view miniplot to matplotlib
As part of the the preparation work for the new workbench we need to convert the miniplot within the instrument view to be a matplotlib-based widget in C++.

This will involve writing a small set of classes in C++ to wrap `matplotlib` and shield developers from the intricacies of dealing with the Python C api.

### LINKED ISSUE #20331: Create a C++ library to wrap matplotlib
Create a new library in the widgets layer that will provide a Matplotlib figure canvas as a C++ widget that hides all of the Python internals from the C++ developer.

martyngigg:
The Qwt dependence has been removed from the instrument view code.

**To test:**

Essentially all functionality of the miniplot needs to be tested here.

Also, an installer should be built and tested.

Fixes #20240 and fix #20331

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


mantid-roman (dismissed):
Comments on the miniplot:

1. It is slower than the Qwt. 
2. Font is different from the rest of the GUI.
3. Plot area is smaller because the labels on the y axis are printed horizontally.
4. Size of the miniplot changes depending on the plotted values.
5. When there is no data the axes change to defaults.
6. Labels on the x axis overlap. The old miniplot makes an attempt (not always working though) to find labels that don't overlap.
7. Tries to add a peak on each click even if the "add peak" tool isn't selected.
8. Doesn't remove peak labels.
9. The y axis in the log scale sometimes is drawn upside down.

The PR includes more than just the miniplot. The other stuff seems to be quite important and complex. Shouldn't they be reviewed separately?

martyngigg:
@mantid-roman On the final point. Do you mean the new library? I originally had a different PR for this but no one took it for review. Stand alone the library is hard to review - most of the issues will come up during use so I chose to keep them together. I can move it out if you prefer?

martyngigg:
On my machine the speed does not seem a problem. Could this be a windows issue? Is it a debug build?

mantid-roman:
It might have been a debug build. I'll need to double check.

martyngigg:
On pt 3., I didn't see any labels the Y axis in the original. How are do they show up?

mantid-roman:
They are printed vertically so the axis always has the same width (horizontal).

mantid-roman:
Maybe you are right about testing a library, especially one for the GUI. But what didn't look right to me is that the number of unit tests was rather small for the amount and complexity of the tested code. But maybe it will grow as when we start using it.

mantid-roman:
I rebuilt it in release. When you drag the mouse across detectors fast enough the new miniplot visibly skips some of them. I am not saying that it's necessarily a bad thing or needs to be fixed but it's different from the current behaviour.

soininen:
Testing this with some random IN4 workspace.
A few issues I've encountered this far:
* The plot stops updated after using the Time-of-flight control at the bottom of Instrument view; one has to exit the Pick tab to get the plot showing some data again. The old qwt based widget seems to have the same behavior, though.
* Magnifying the plot by dragging a rectangle opens an error dialog giving a cryptic message 'Cannot create a Peak object because of the error: Attemp to assign to property (DetectorID) of incorrect type'
* After demagnifying by clicking the plot with the second mouse button, the vertical axis doesn't get updated automatically anymore. One has to change the Y scale to log or linear to restore the behavior.
* Also the horizontal axis stops updating after demagnifying.

ianbush:
Just to note I am about to pick up the branch to work with.

martyngigg:
This has become a nightmare to try and patch up. I'm going to try and split the branch up into smaller chunks in other PRs.

PR #20386, opened 2017-09-06
`ConvertSpectrumAxis` option to leave the new axis unordered

### LINKED ISSUE #20375: ConvertSpectrumAxis needs an option to leave the resulting axis unordered.
`ConvertSpectrumAxis` currently automatically orders the resulting spectrum axis ascending. This is not always desired in the case of a scanning workspace, where e.g. 2theta ranges of adjacent pixels might overlap during the scan. Hence blindly ordering by 2theta will mix up the pixels, and one would need to do expensive detector id lookups to be able to perform anything meaningful, needless to say that the scanned detector features are not yet sufficiently exposed to python.
An option to not order the axis, but to preserve the correspondence with the original workspace will hence overcome the problem.

gvardany:
This adds a new boolean option (true by default) to the above mentioned algorithm, to be able to switch off the ordering, which is needed for scanning workspaces (see the issue). 

Also, in the default ordered axis case, the performance is improved (complexity from Nˆ2 down to N) by creating the new numeric axis from the whole vector once, instead of inserting each point individually in a loop, where insertion (`NumericAxis::setValue`) triggers another loop over the whole axis. This made it possible to execute the algorithm in finite time over scanning workspaces, that are normally very large.

**To test:**
1) Load the attached detector scan file (`tar -xzvf` first, you might also need to set facility to `ILL`).
[scanfile.zip](https://github.com/mantidproject/mantid/files/1280438/scanfile.zip)
2) Run ConvertSpectrumAxis to convert to`SignedTheta` without ordering.
3) The spectrum axis of the result should be
   
- 0 for spectra 0-570 (monitor)
- -34.24 >> -5.749 for spectra 571 - 1141 (pixel 1) 
- -34.19 >> -5.699 for spectra 1142 - 1712 (pixel 2)

and so on.

Fixes #20375 . 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/src/ConvertSpectrumAxis2.cpp
+++ b/Framework/Algorithms/src/ConvertSpectrumAxis2.cpp
@@ -260,7 +260,8 @@ MatrixWorkspace_sptr ConvertSpectrumAxis2::createOutputWorkspace(
       outputWorkspace->getSpectrum(currentIndex)
           .copyInfoFrom(inputWS->getSpectrum(it->second));
       ++currentIndex;
-      progress.report("Setting output spectrum #" + std::to_string(currentIndex));
+      progress.report("Setting output spectrum #" +
+                      std::to_string(currentIndex));
     }
   }
 
--- a/Framework/Algorithms/src/ConvertSpectrumAxis2.cpp
+++ b/Framework/Algorithms/src/ConvertSpectrumAxis2.cpp
@@ -67,7 +67,7 @@ void ConvertSpectrumAxis2::init() {
                   "(EMode=Indirect))");
 
   declareProperty("OrderAxis", true, "Whether or not to sort the resulting"
-                                     " spectum axis.");
+                                     " spectrum axis.");
 }
 
 void ConvertSpectrumAxis2::exec() {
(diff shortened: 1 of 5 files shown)


Framework/Algorithms/src/ConvertSpectrumAxis2.cpp
```diff
@@ -222,20 +247,24 @@ MatrixWorkspace_sptr ConvertSpectrumAxis2::createOutputWorkspace(
...
-    outputWorkspace->getSpectrum(currentIndex)
-        .copyInfoFrom(inputWS->getSpectrum(it->second));
-    ++currentIndex;
-
-    progress.report("Creating output workspace...");
+
+  // Note that this is needed only for ordered case
+  if (m_toOrder) {
+    std::multimap<double, size_t>::const_iterator it;
+    for (it = m_indexMap.begin(); it != m_indexMap.end(); ++it) {
```
NickDraper: Probably a minor point seeing as you have found a way to avoid this, but I think this loop could be multithreaded safely.

NickDraper (commented):
The options that are only different by case or underscores are a mess, but I understand that they were not introduced in this PR.  It would be nice to clean them up though.

I'd be in favour of removing the poor entries and dealing with any fallout (which I suspect would be minor at the most)

NickDraper:
On balance this is ok, but I'd really like to see the repeated Target options removed

NickDraper (approved):

PR #20389, opened 2017-09-06
Crystal Field Multisite Function

### LINKED ISSUE #19341: Crystal Field: better support for multi-site calculations.
During an experiment, it was found that the multi-site CrystalFieldMulti object cannot have / does not function with

 a.	Background
 b.	ResolutionModel
 c.	estimate_parameters

In case a) and b), whilst these can be set for the component sites, these properties are more properly part of the multi-site model, because only a single material is being measured. Else ties need to be established to ensure that only one background / resolution model is used (also even if this is done, the generated string is wrong, and gives an error when passed to Fit)

mantid-roman:
Added CrygstalFieldFunction which addresses the issues with calculations of multi-site crystal fields. 

**To test:**

See the unit tests and the doc page on the examples of usage.

Fixes #19341.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/CurveFitting/inc/MantidCurveFitting/Functions/CrystalFieldControl.h
```diff
@@ -0,0 +1,125 @@
...
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+class MANTID_CURVEFITTING_DLL CrystalFieldControl
+    : public API::CompositeFunction {
+public:
+  CrystalFieldControl();
+  /// Set a value to attribute attName
```
AnthonyLim23: Is attName a typo?

mantid-roman: Yes

Framework/CurveFitting/inc/MantidCurveFitting/Functions/CrystalFieldFunction.h
```diff
@@ -0,0 +1,247 @@
...
+
+  /// Get a reference to an attribute
+  std::pair<API::IFunction *, std::string>
+  getAttributeReference(const std::string &attName) const;
+  /// Build and cache the attribute names
+  void buildAttributeNames() const;
+
+  /// Make maps between parameter names and indices
+  void makeMaps() const;
+  void makeMapsSS() const;
```
AnthonyLim23: what are SS, SM, MS and MM?

Framework/CurveFitting/src/ComplexMatrix.cpp
```diff
@@ -301,8 +301,6 @@ void ComplexMatrix::eigenSystemHermitian(GSLVector &eigenValues,
   eigenValues.resize(n);
   eigenVectors.resize(n, n);
   auto workspace = gsl_eigen_hermv_alloc(n);
-  // RT: I think there is a bug in this function. It returns garbage in
-  // eigenvectors in a repeated call.
```
AnthonyLim23: has this bug been fixed?

mantid-roman: The bug wasn't in the gsl. And it has been fixed.

Framework/CurveFitting/src/Functions/CrystalFieldControl.cpp
```diff
@@ -0,0 +1,354 @@
...
+    if (name == "Temperatures") {
+      m_temperatures = attr.asVector();
+      buildControls();
+    } else if (name == "FWHMs") {
+      m_FWHMs = attr.asVector();
+      if (m_FWHMs.size() == 1 && m_FWHMs.size() != m_temperatures.size()) {
+        m_FWHMs.assign(m_temperatures.size(), m_FWHMs.front());
+      }
+    }
+    API::IFunction::setAttribute(name, attr);
```
AnthonyLim23: could we add an esle statement at the end that produces an error/warning?

Framework/CurveFitting/src/Functions/CrystalFieldFunction.cpp
```diff
@@ -0,0 +1,1320 @@
+#include "MantidCurveFitting/Functions/CrystalFieldFunction.h"
+#include "MantidCurveFitting/Functions/CrystalElectricField.h"
```
AnthonyLim23: swap lines 1 and 2 (alphabetical order)

Framework/CurveFitting/src/Functions/CrystalFieldFunction.cpp
```diff
@@ -0,0 +1,1320 @@
...
+}
+
+/// Get i-th parameter
+double CrystalFieldFunction::getParameter(size_t i) const {
+  checkSourceFunction();
+  checkTargetFunction();
+  if (i < m_nControlParams) {
+    return m_control.getParameter(i);
+  } else if (i < m_nControlSourceParams) {
+    return m_source->getParameter(i - m_nControlParams);
```
AnthonyLim23: should we also check if the index is valid?

Framework/CurveFitting/src/Functions/CrystalFieldFunction.cpp
```diff
@@ -0,0 +1,1320 @@
...
+  checkSourceFunction();
+  updateTargetFunction();
+  IFunction::setUpForFit();
+}
+
+/// Declare a new parameter
+void CrystalFieldFunction::declareParameter(const std::string &, double,
+                                            const std::string &) {
+  throw Kernel::Exception::NotImplementedError(
+      "CrystalFieldFunction cannot not have its own parameters.");
```
AnthonyLim23: the double negative makes this error message a bit tricky to understand

Framework/CurveFitting/src/Functions/CrystalFieldFunction.cpp
```diff
@@ -0,0 +1,1320 @@
...
+  for (size_t iSpec = 0; iSpec < m_control.nFunctions(); ++iSpec) {
+    std::string prefix(SPECTRUM_PREFIX);
+    prefix.append(std::to_string(iSpec)).append(".");
+    auto names = m_control.getFunction(iSpec)->getAttributeNames();
+    for (auto &name : names) {
+      name.insert(name.begin(), prefix.begin(), prefix.end());
+    }
+    m_attributeNames.insert(m_attributeNames.end(), names.begin(), names.end());
+  }
+  auto nSpec = nSpectra();
```
AnthonyLim23: this doesn't seem to be used anywhere other than the for loop below. Could we not just put 
`for (size_t iSpec = nSpectra(); iSpec < m_target->nFunctions(); ++iSpec) {`

Framework/CurveFitting/src/Functions/CrystalFieldFunction.cpp
mantid-roman: We do it already. `m_nControlSourceParams` is set to the total number of parameters in `m_control` and `m_source`. `m_source->getParameter` makes also its own check.

Framework/CurveFitting/src/Functions/CrystalFieldFunction.cpp
```diff
@@ -0,0 +1,1320 @@
...
+
+/// Build the target function in a single site - multi spectrum case.
+void CrystalFieldFunction::buildSingleSiteMultiSpectrum() const {
+  auto fun = new MultiDomainFunction;
+  m_target.reset(fun);
+
+  DoubleFortranVector en;
+  ComplexFortranMatrix wf;
+  ComplexFortranMatrix ham;
+  ComplexFortranMatrix hz;
```
AnthonyLim23: what is hz?

Framework/CurveFitting/src/Functions/CrystalFieldFunction.cpp
```diff
@@ -0,0 +1,1320 @@
...
+  CrystalFieldUtils::buildSpectrumFunction(*spectrum, peakShape, values, xVec,
+                                           yVec, fwhmVariation, fwhm,
+                                           nRequiredPeaks, fixAllPeaks);
+  return IFunction_sptr(spectrum);
+}
+
+API::IFunction_sptr CrystalFieldFunction::buildPhysprop(
+    int nre, const DoubleFortranVector &en, const ComplexFortranMatrix &wf,
+    const ComplexFortranMatrix &ham, const std::string &propName) const {
+
```
AnthonyLim23: could make a function containing the below if statements and remove the returns. Then the new function can be called below with a get propFun and then returning it. Then the same function can be used in https://github.com/mantidproject/mantid/pull/20389/files#diff-e713a6a7351e91d3ef66099a149c18dfR936

Framework/CurveFitting/src/Functions/CrystalFieldFunction.cpp
mducle: It's the Zeeman Hamiltonian term (the magnetic field dependent part) - `ham` is the crystal field Hamiltonian in zero external field. I guess it could be changed to `hzeeman` or something like that...

AnthonyLim23: I think thats a good idea. I thought it may have been the z component of the Hamiltonian.

AnthonyLim23:
a few of the test have an index out of range

Framework/CurveFitting/src/Functions/CrystalFieldFunction.cpp
mantid-roman: The reason is that this method creates functions for physical properties while the method below updates already created functions.

AnthonyLim23:
code works as expected. The new fit functions work with the fit function GUI

AnthonyLim23 (changes requested):
There are a few changes to make the code more readable. A few tests need fixing.

AnthonyLim23 (changes requested):
tests need fixing and there are couple bits that are not clear in the code

Framework/CurveFitting/src/Functions/CrystalFieldControl.cpp
mantid-roman: I don't think it is needed here. Temperature and FWHMs require additional actions before `IFunction::setAttribute` is called.

Framework/CurveFitting/src/Functions/CrystalFieldFunction.cpp
mantid-roman: :)

Framework/CurveFitting/src/Functions/CrystalFieldControl.cpp
AnthonyLim23: I just thought it would be good to make sure that the name is valid. So is that checked elsewhere?

mantid-roman: Yes, by `IFunction::setAttribute`.

Framework/CurveFitting/src/Functions/CrystalFieldFunction.cpp
AnthonyLim23: :)

AnthonyLim23 (approved):

martyngigg:
@mantid-roman The RHEL failure was just a problem with that machine.

martyngigg:
The Windows build failure is actually just a warning that is fixed in #20564

martyngigg:
@mantid-roman The CodeConventions system test is failing.

PR #20392, opened 2017-09-06
Re #20369 fixed loop and naming of workspaces

### LINKED ISSUE #20369: Allow multiple scattering data to be distinguished with LoadMcStas
LoadMcStas can read a mcstas output containing multiple event datasets, however it currently output only the sum of all event data workspaces. 

What is returned is named EventData_ws where ws is the name of the output workspace + all histogram workspaces in the Nexus named as specified in the Nexus file + _ws.

Here simple extend this to also have all individual event workspaces in the nexus (if different from one event dataset) outputted using the same naming convention as for the histogram workshapces.

jamesquantum:
Fix by @michaeljturner "There were two things wrong: first, the new workspace pointer was created outside a for loop, so each subsequent dataset just overwrote the previous one in memory, so I put that inside the for loop.  Then all the workspaces were given the same name, so they overwrote each other in workspace data service."

Fixes #20369

### DIFF
--- a/Framework/DataHandling/src/LoadMcStas.cpp
+++ b/Framework/DataHandling/src/LoadMcStas.cpp
@@ -60,6 +60,7 @@ void LoadMcStas::init() {
 
 //----------------------------------------------------------------------------------------------
 /** Execute the algorithm.
+ *
  */
 void LoadMcStas::exec() {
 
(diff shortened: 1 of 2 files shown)


Anders-Markvardsen:
@michaeljturner and @jamesphysics please note ccpcheck is complaining about 
capture

Anders-Markvardsen:
When I load mcstas_event_hist.h5 I get:

capture

that is EventData_ws has been replaced by ws_EventWS_0 and ws_EventWS_1. Where the content of EventData_ws is the sum of ws_EventWS_0 and ws_EventWS_1. 

Can this be changed so that the first workspace returned is still EventData_ws and this is the sum of all the event workspaces found in the nexus file. And if there is more than one event data in the nexus file then each of these are in addition outputted, e.g., as the last workspaces in the group workspace

michaeljturner:
Have updated the numbering of the event workspaces, not yet implemented adding together the workspaces

NickDraper:
@Anders-Markvardsen , can you nurse this through as James isn't around to do it anymore

Anders-Markvardsen:
@michaeljturner I will fix the remaining work for this issue from here

Anders-Markvardsen:
Reasons for choosing this PR. 1) it initially, accidentally, coded a solution slightly different from the intended and to untangle this may take longer than starting from fresh 2) regardless this branch before reason major Qt changes and therefore would like be difficult to merge now

Issue #20410, opened 2017-09-08
Make WorkspaceUnitValidator work for multiple units


samueljackson92:
Currently `WorkspaceUnitValidator` will only work for a single unit type. For #20409 it would be useful to check if the workspace has units of either TOF or d spacing.

PR #20431, opened 2017-09-11
Allow FindSXPeaks to accept multiple units

### LINKED ISSUE #20409: Convert FindSXPeaks to accept both TOF and d-spacing
It should be possible to make this algorithm work with data in units of d-spacing as well as TOF. We should try to do this before the code freeze so that the x axis resolution is not tied to the name "TofResolution".

To do:
 - [ ] Change name of parameter
 - [ ] Compute the TOF value from d-spacing value and pass to SXPeak

samueljackson92:
~~**Note this cannot be merged before #20410.**~~

This PR converts the FindSXPeaks algorithm to accept workspaces either with TOF or d-spacing as input.

**To test:**

Here's a script you can use for testing. Data and cal file can be found in `smb://olympic/babylon5/Scratch/Sam`

```python
raw = LoadRaw("WISH00033602")
raw = CropWorkspace(raw, XMin=6000, XMax=95000)
# For this run there is a noisy detector tube the produces a lot of rubbish
# peaks. Here I'm using a mask workspace to avoid finding peaks there.
tube_mask = LoadCalFile("~/data/tube_mask_33602.cal", InputWorkspace=raw, 
                     MakeGroupingWorkspace=False, MakeOffsetsWorkspace=False, WorkspaceName='tube')
                     
MaskDetectors(raw, MaskedWorkspace=tube_mask)
peaks1 = FindSXPeaks(raw, PeakFindingStrategy="AllPeaks", SignalBackground=100, Resolution=0.01)

d_space = ConvertUnits(raw, Target="dSpacing", EMode="Elastic")
peaks2 = FindSXPeaks(d_space, PeakFindingStrategy="AllPeaks", SignalBackground=10, Resolution=0.01)
CheckWorkspacesMatch(peaks1, peaks2)
```

Also:
 - [ ] Check you can also use the absolute resolution option & that the x units produce sensible results when using d-spacing.
 - [ ] Check tests pass
 - [ ] Code review

Overlaying both the resulting workspaces on top of the instrument view should show something like this:

screen shot 2017-09-11 at 16 02 00

Fixes #20409.

**Release Notes** 

[See here](https://github.com/mantidproject/mantid/compare/20409-findsxpeaks-multiple-units?expand=1#diff-f20ef25841b2c33b446d2e9ba7341f29R35)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


edwardb96 (changes requested):
This seems to produce sensible results.
This script: 

```python
raw = LoadRaw("WISH00033602")
raw = CropWorkspace(raw, XMin=6000, XMax=95000)
# For this run there is a noisy detector tube the produces a lot of rubbish
# peaks. Here I'm using a mask workspace to avoid finding peaks there.
tube_mask = LoadCalFile("~/Code/Work/Mantid/PR#20431/tube_mask_33602.cal", InputWorkspace=raw, 
                     MakeGroupingWorkspace=False, MakeOffsetsWorkspace=False, WorkspaceName='tube')
                     
MaskDetectors(raw, MaskedWorkspace=tube_mask)
peaks1 = FindSXPeaks(raw, PeakFindingStrategy="AllPeaks", SignalBackground=100, Resolution=0.01)

d_space = ConvertUnits(raw, Target="dSpacing", EMode="Elastic")
peaks2 = FindSXPeaks(d_space, 
	PeakFindingStrategy="AllPeaks",
	SignalBackground=10, 
	ResolutionStrategy='AbsoluteResolution',
	XUnitResolution=500,
	PhiResolution=1.5,
	TwoThetaResolution=1.5)
CheckWorkspacesMatch(peaks1, peaks2)
```

produces this graph, 

screenshot from 2017-09-18 13-35-32

which I'm sure could be made to match the original more exactly by further adjustment of the parameter values.

Just a few style points/suggestions...

Framework/Crystal/inc/MantidCrystal/FindSXPeaksHelper.h
```diff
@@ -66,6 +68,8 @@ struct DLLExport SXPeak {
 private:
   /// TOF
   double _t;
+  /// d-spacing
+  double _d_spacing;
   /// 2 * theta
   double _th2;
```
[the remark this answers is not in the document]
samueljackson92: You are right of course, but I was following the style of the code that was already there. Whether that is a valid excuse is up for questioning 😄 .

`d_spacing` is pretty descriptive as to what the value will contain give the context. I agree that `t` and `th2` are not great.

Framework/Crystal/inc/MantidCrystal/FindSXPeaksHelper.h
```diff
@@ -183,7 +191,8 @@ class DLLExport AllPeaksStrategy : public PeakFindingStrategy {
   AllPeaksStrategy(const BackgroundStrategy *backgroundStrategy,
                    const API::SpectrumInfo &spectrumInfo,
                    const double minValue = EMPTY_DBL(),
-                   const double maxValue = EMPTY_DBL());
+                   const double maxValue = EMPTY_DBL(),
+                   const bool tofUnits = true);
```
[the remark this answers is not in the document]
samueljackson92: I thought this as well when writing it. I was surprised that `Unit.h` did not contain such a set of types, and really it should be implemented there instead.

Framework/Crystal/test/FindSXPeaksTest.h
```diff
@@ -306,6 +309,87 @@ class FindSXPeaksTest : public CxxTest::TestSuite {
...
+    makeOnePeak(4, 60, 5, workspace);
+    makeOnePeak(8, 45, 8, workspace);
+
+    FindSXPeaks alg;
+    alg.initialize();
+    alg.setProperty("InputWorkspace", workspace);
+    alg.setProperty("OutputWorkspace", "found_peaks");
+    alg.execute();
+    TSM_ASSERT("FindSXPeak should have been executed.", alg.isExecuted());
+
```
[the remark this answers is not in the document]
samueljackson92: Agreed.

Framework/Kernel/inc/MantidKernel/CompositeValidator.h
```diff
@@ -65,12 +70,22 @@ class DLLExport CompositeValidator : public IValidator {
...
 
 private:
-  /// Private Copy constructor: NO DIRECT COPY ALLOWED
-  CompositeValidator(const CompositeValidator &);
   /// Verify the value with the child validators
   std::string check(const boost::any &value) const override;
+  /// Verify the value with the child validators with logical "and"
+  std::string checkLogicalAnd(const boost::any &value) const;
+  /// Verify the value with the child validators with logical "and"
+  std::string checkLogicalOr(const boost::any &value) const;
```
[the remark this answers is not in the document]
samueljackson92: Maybe that is clearer yes. The comment is also wrong here.

I originally implemented it as a separate class, but then we can just rename the old one as that would break backwards compatibility with the python api which exposes `CompositeValidator`.

Framework/Kernel/test/CompositeValidatorTest.h
```diff
@@ -66,6 +66,22 @@ class CompositeValidatorTest : public CxxTest::TestSuite {
...
+
+  void
+  test_Given_TwoValidators_When_CheckIsValid_That_ValidValuesReturnValid() {
+    // Arrange
+    BoundedValidator<int> *val1 = new BoundedValidator<int>(1, 50);
+    BoundedValidator<int> *val2 = new BoundedValidator<int>(60, 100);
+
+    CompositeValidator comp(CompositeRelation::OR);
+    comp.add(IValidator_sptr(val1));
+    comp.add(IValidator_sptr(val2));
```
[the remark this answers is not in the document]
samueljackson92: Copy & paste from above.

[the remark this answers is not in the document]
edwardb96: What would you need to break in order to use `boost::make_shared` here?

Framework/Crystal/inc/MantidCrystal/FindSXPeaksHelper.h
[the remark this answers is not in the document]
edwardb96: Looking at it I'd guess that `Unit.h` was written before strongly typed `enum`s were a thing in c++. I take it that adding them is something which would need to be done in a separate issue/PR.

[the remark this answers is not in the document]
samueljackson92: Yes, but I have added an enum here for now as it is nicer and is much easier to do.

Framework/Crystal/inc/MantidCrystal/FindSXPeaks.h
```diff
@@ -114,6 +102,11 @@ class DLLExport FindSXPeaks : public API::Algorithm {
 
   //
   void reducePeakList(const peakvector &, Mantid::API::Progress &progress);
+
+  // Check if this workspace has units of TOF
+  FindSXPeaksHelper::XAxisUnit
+  workspaceHasTOFUnits(Mantid::API::MatrixWorkspace_const_sptr workspace) const;
```
edwardb96: This method should probably have a different name now that it no longer returns a `bool`.

edwardb96 (approved):

edwardb96:
Ok to merge when the builds pass. :+1:

Framework/Crystal/src/FindSXPeaks.cpp
```diff
@@ -118,27 +128,29 @@ void FindSXPeaks::init() {
...
 
   declareProperty(
       "Resolution", 0.01, mustBePositiveDouble,
       "Tolerance needed to avoid peak duplication in number of pixels");
 
-  declareProperty("TofResolution", 5000., mustBePositiveDouble,
-                  "Absolute tolerance in time-of-flight needed to avoid peak "
-                  "duplication in number of pixels. The values are specified "
-                  "in microseconds.");
+  declareProperty(
```
martyngigg: This change will break any keyword argument usage of this call in Python. If we are okay changing this then I think `XResolution` would be slightly better.

We could deal with this by defining a specialised version of the function in Python: https://github.com/mantidproject/mantid/blob/master/Framework/PythonInterface/mantid/simpleapi.py#L43 and allowing the old `TofResolution` keyword but produce a deprecation warning at the same time?

martyngigg (changes requested):

Framework/Crystal/src/FindSXPeaks.cpp
samueljackson92: Could do. I thought that this would be ok because this parameter was only implemented in this release cycle. So while it is in the nightly, it is not in any actual release yet.

If that is not ok then I shall do the python api fix as well.

martyngigg: @samueljackson92 My apologies I didn't realise it was this release cycle. In that case changing it is okay.

Could we just rename it to `XResolution` (I think the unit part is slightly redundant)?

samueljackson92: @martyngigg That's ok, it was exactly obvious that was the case. Yes I shall rename the parameter.

martyngigg (approved):
Thanks. Good to go now.

Framework/Crystal/src/FindSXPeaks.cpp
```diff
@@ -128,26 +128,25 @@ void FindSXPeaks::init() {
...
+                  "enable the XResolution, PhiResolution, "
+                  "ThetaResolution.\n");
 
   declareProperty(
       "Resolution", 0.01, mustBePositiveDouble,
       "Tolerance needed to avoid peak duplication in number of pixels");
 
   declareProperty(
-      "XUnitResolution", 5000., mustBePositiveDouble,
+      "XResolution", 5000., mustBePositiveDouble,
```
martyngigg: I've also just thought now. Is 5000 a good default now?

Framework/Crystal/src/FindSXPeaks.cpp
```diff
@@ -128,26 +128,25 @@ void FindSXPeaks::init() {
...
       "Absolute tolerance in time-of-flight or d-spacing needed to avoid peak "
       "duplication in number of pixels. The values are specified "
```
martyngigg: I think the final sentence here is also not correct as the value can now be either microseconds or angstroms?

samueljackson92: True. Will change.

Framework/Crystal/src/FindSXPeaks.cpp
samueljackson92: We'll going the other way and using angstroms is also incorrect. But perhaps it is better because the resolution will be tighter by default.

martyngigg:
:+1:

PR #20434, opened 2017-09-12
Fix bug with TableWorkspace plot type & sorting

### LINKED ISSUE #19518: Sorting a PeaksWorkspace column causes the column type to reset
If you set the type of a column then sort a column the type is reset. This is kind of annoying.

### Expected behavior
The type should remain unchanged

### Actual behavior
The type is reset to it's default

 - Right click on a column -> set type -> something other than the default
 - Right click on a column -> sort by -> ascending
 - Column type disappears.

### Platforms affected
All

samueljackson92:
This fixes a bug where if you set the plot type on a table workspace column via a table view, the workspace is not correctly updated with the plot type.

**To test:**
 - Follow the steps to reproduce the bug in the issue report #19518.
 - Apply the fix, check this is now solved.
 - Code review.

Fixes #19518

**Release Notes** 
[See here](https://github.com/mantidproject/mantid/compare/19518-sort-peaks-ws-column?expand=1#diff-314160101dcfa40a50e85e320e4e750fR54)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidPlot/src/Mantid/MantidTable.cpp
+++ b/MantidPlot/src/Mantid/MantidTable.cpp
@@ -312,6 +312,12 @@ void MantidTable::cellEdited(int row, int col) {
   d_table->setText(row, col, QString::fromStdString(s.str()));
 }
 
+void MantidTable::setPlotDesignation(Table::PlotDesignation pd,
+                                     bool rightColumns) {
+  Table::setPlotDesignation(pd, rightColumns);
+  setPlotTypeForSelectedColumns(pd);
+}
+
 //------------------------------------------------------------------------------------------------
 /** Call an algorithm in order to delete table rows
  *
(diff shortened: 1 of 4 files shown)


MantidPlot/src/Mantid/MantidTable.cpp
```diff
@@ -423,6 +429,19 @@ void MantidTable::sortColumns(const QStringList &s, int type, int order,
   }
 }
 
+/** Set the plot type on the workspace for each selected column
+ *
+ * @param plotType :: the plot type to set the selected columns to.
+ */
+void MantidTable::setPlotTypeForSelectedColumns(int plotType) {
+  const QStringList list = selectedColumns();
+  for (int i = 0; i < static_cast<int>(list.count()); i++) {
```
edwardb96: I think this can be converted to a range-based for loop?

edwardb96 (changes requested):
This fix works fine on my PC (Fedora/Linux). 👍 
It's probably worth changing the for loop over though.

MantidPlot/src/Mantid/MantidTable.cpp
samueljackson92: Yes. You're right. 👍

edwardb96 (approved):

PR #20442, opened 2017-09-12
CalculatePolynomialBackground

### LINKED ISSUE #20404: Linear background fitting for ILL reflectometry
The ILL reflectometry reduction workflow needs more complicated ways of calculating background than what `CalculateFlatBackground` can provide, namely:
* Not flat but linear background option
* Multiple background fitting ranges per histogram

soininen:
This PR introduces a new `CalculatePolynomialBackground` algorithm. As the name states, the algorithm fits a polynomial of given degree to histograms in a workspace and returns the evaluated values.

As a side effect, `Mantid::CurveFitting::Jacobian` was moved to `Mantid::API` to provide access to a Jacobian implementation. The class was further renamed to `BasicJacobian` to avoid name clash with the abstract `Mantid::API::Jacobian`.

**To test:**

Try surprising it with confusing input. Use your favorite datasets or play with the following script as a starting point:
```python
import numpy

ws = CreateWorkspace(
    DataX=numpy.array([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]),
    DataY=numpy.array([1, 2, 1, 2, 1, 2, 1, 2, 1]),
)

polybkg = CalculatePolynomialBackground(ws)
```

Resolves #20404.

**Release Notes** 

Release notes were updated accordingly.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Algorithms/src/CalculatePolynomialBackground.cpp
```diff
@@ -0,0 +1,279 @@
...
+                  "Degree of the fitted polynomial.");
+  declareProperty(Kernel::make_unique<Kernel::ArrayProperty<double>>(
+                      Prop::XRANGES, std::vector<double>(), orderedPairs),
+                  "A list of fitting ranges given as pairs of X values.");
+}
+
+//----------------------------------------------------------------------------------------------
+/** Execute the algorithm.
+ */
+void CalculatePolynomialBackground::exec() {
```
samueljackson92: This `exec` method does too much. From a brief glance I can see that there could be separate methods for:
 - Finding the included range
 - Running the fit
 - Pulling the parameters & errors out for the fit workspace
 - Evaluating the background function and filling the Y/E values

Framework/Algorithms/test/CalculatePolynomialBackgroundTest.h
```diff
@@ -0,0 +1,152 @@
...
+private:
+  static boost::shared_ptr<Algorithms::CalculatePolynomialBackground>
+  makeAlgorithm() {
+    auto a = boost::make_shared<Algorithms::CalculatePolynomialBackground>();
+    a->initialize();
+    a->setChild(true);
+    a->setRethrows(true);
+    return a;
+  }
+};
```
samueljackson92: Could we also add a performance test?

Framework/CurveFitting/test/Functions/ProductQuadraticExpTest.h
```diff
@@ -199,7 +199,7 @@ to check that the results are equal.
                   LinearIncrementingAssignment(0, 0.1));
     FunctionDomain1DVector domain(xValues);
 
-    Mantid::CurveFitting::Jacobian jacobian(nResults, 5);
+    Mantid::API::BasicJacobian jacobian(nResults, 5);
```
samueljackson92: Kind of feels like this abstract base class should be changed to be called `IJacobian` or something instead and then call this just `Jacobian`, but perhaps that touches too much code to change?

samueljackson92 (changes requested):
I found that when I fit the following workspace I get a good fit, but the error bars seem to be huge? It looks like they're off by like a factor of `~1000` if you comment the last line of the script.

```python
import numpy as np

x = np.arange(1000.0, 5000.0, 10)
y = np.sin(x/1000.0)
y += np.random.normal(0, 0.1, size=y.shape) + 1000

ws = CreateWorkspace(
    DataX=x,
    DataY=y,
)

polybkg = CalculatePolynomialBackground(ws, degree=3)
#polybkg.setE(0, polybkg.dataE(0) / 1000.0) 
```
screen shot 2017-09-18 at 11 48 38

Framework/Algorithms/test/CalculatePolynomialBackgroundTest.h
soininen: Great idea!

soininen:
I've removed the error calculations from the algorithm for the moment and reverted the Jacobian move/rename. It seems to be easy to construct cases where the errors are way overestimated the way they are calculated now.

NickDraper:
@samueljackson92 Could you take another look at this one please

samueljackson92:
@soininen With the script I posted above I now get a very bad fit. This is much worse than the original fit with large errors.

screen shot 2017-09-26 at 16 35 11

soininen:
@samueljackson92 Thanks for noticing a bug! A classical off-by-one error had slipped in.

samueljackson92:
@soininen  This seems to be giving sensible results now. Looks good to me. 👍

samueljackson92 (approved):

PR #20449, opened 2017-09-12
Bring support for HRPD into ISISPowder scripts

### LINKED ISSUE #20152: Move HRPD to ISIS Powder
- [x] Create the instrument class
- [x] Supporting files (eg. `hrpd_algs`, parameter maps etc)
- [x] Identify HRPD-unique steps
- [x] Implement HRPD-specific algs
- [x] Patch changes into the main flow
- [x] Add unit tests for any functions
- [x] Add a system test
- [x] Documentation
- [ ] ???
- [ ] Profit

josephframsay:
A collection of routines and classes have been written to replace the CRI scripts previously used by HRPD in Mantid. Some refactoring also took place, including adding the option for `isis_powder.SampleDetails` to support geometries other than cylinders.

**To test:**

Scientific validation has taken place with HRPD, so as long as the unit and system tests for `ISIS_PowderHRPD` pass without issues, the changes should be good to go (plus code review)

Fixes #20152

**Release Notes** 
[here](https://github.com/mantidproject/mantid/commit/88890ea0d064688f8c400b4e769276260e785597)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


scripts/Diffraction/isis_powder/routines/sample_details.py
```diff
@@ -4,21 +4,28 @@
...
+                                                       exception_msg="Sample property \"shape\" was not passed to "
+                                                                     "SampleDetails. If you have been using this class "
+                                                                     "without worrying about \"shape\", you probably "
+                                                                     "want to set this to \"cylinder\"")
+        center = common.dictionary_key_helper(dictionary=kwargs, key="center",
+                                              exception_msg=property_err_string + "center")
+        SampleDetails._validate_center(center)
+        self._center = [float(i) for i in center]  # List of X, Y, Z position
+
+        if self.shape_type == "cylinder":
```
samueljackson92: `shape_type` and `shape` should be called `_shape_type` and `_shape` to match the [python conventions](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles).

scripts/Diffraction/isis_powder/routines/sample_details.py
```diff
@@ -4,21 +4,28 @@
...
+            self.shape = _Cylinder(kwargs)
+        elif self.shape_type == "slab":
+            self.shape = _Slab(kwargs)
+        else:
+            raise KeyError("Shape type \"" + self.shape_type + "\" not supported: current supported shape types are "
```
samueljackson92: Don't use string concatenation with `+`, use [.format](https://docs.python.org/2/library/string.html#formatstrings) instead.

samueljackson92 (changes requested):
Generally good. I have a couple of comments that should be addressed.

scripts/Diffraction/isis_powder/routines/sample_details.py
```diff
@@ -95,7 +80,60 @@ def _validate_sample_details_constructor_inputs(height, radius, center):
...
+        # Ensure we got double (or int) types and they are sane
+
+        # Attempt to convert them all to floating point relying on the fact on
+        # the way Python has aliases to an object
+        for key, value in iteritems(values_to_check):
+            _check_value_is_physical(property_name=key, value=value)
+            _check_can_convert_to_float(property_name=key, value=value)
+
+    def _print(self):
+        print("Sample Details")
```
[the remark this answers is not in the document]
josephframsay: I think David's thinking behind this is that (as far as I can tell) the logger doesn't show up in the script window, only in the results log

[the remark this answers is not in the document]
samueljackson92: Ah I see. This is probably better then.

scripts/Diffraction/isis_powder/routines/sample_details.py
```diff
@@ -3,22 +3,31 @@
...
-        self.shape_type = "cylinder"
-        SampleDetails._validate_sample_details_constructor_inputs(height=height, radius=radius, center=center)
-        self.height = float(height)
-        self.radius = float(radius)
-        self.center = [float(i) for i in center]  # List of X, Y, Z position
+        self._shape_type = common.dictionary_key_helper(dictionary=kwargs, key="shape", throws=False)
+        if self._shape_type is None:
+            self._shape_type = "cylinder"
+            warning = "Failed to supply parameter \"shape\" - defaulting to \"cylinder\""
+            logger.warning(warning)
```
samueljackson92: Could we make this so that it also makes a `print` call here as well? Just so you get a message both in the scrit window and the log window?

samueljackson92 (changes requested):

samueljackson92 (approved):

samueljackson92:
I think this is good to go now. 👍

PR #20451, opened 2017-09-12
Change calculation of `Workspace2D::blocksize`

peterfpeterson:
Since there is no strict requirement that all histograms in a `Workspace2D` have the same number of bins, change `size_t blocksize()` to reflect this. This also does some modernization of various code to reemphasize this point.

**To test:**

The main thing to check is the rewrite of `Workspace2D::blocksize()` ea9cd90951437ce8b3beeb3081230d8926d31f47. Everything else is moving code away from `blocksize` or caching the value.

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/API/src/WorkspaceOpOverloads.cpp
```diff
@@ -412,22 +412,26 @@ bool WorkspaceHelpers::commonBoundaries(const MatrixWorkspace &WS) {
...
   const size_t numHist = WS.getNumberHistograms();
   for (size_t j = 1; j < numHist; ++j) {
-    const double sum =
-        std::accumulate(WS.readX(j).begin(), WS.readX(j).end(), 0.);
+    const auto &x_j = WS.readX(j);
+    // they should all have the same number of x-values
+    if (x_0.size() != x_j.size())
+      return false;
+
+    const double sum = std::accumulate(x_j.begin(), x_j.end(), 0.);
```
SimonHeybrock: As probably mentioned in the past, what `commonBoundaries` does is conceptually broken. Would it be in the scope of this PR to fix this?

Framework/DataObjects/src/Workspace2D.cpp
```diff
@@ -138,9 +139,15 @@ size_t Workspace2D::size() const { return data.size() * blocksize(); }
...
-  return (!data.empty())
-             ? static_cast<ISpectrum const *>(data[0])->dataY().size()
-             : 0;
+  if (data.empty()) {
+    return 0;
+  } else {
+    size_t numBins = data[0]->size();
+    const auto numHist = this->getNumberHistograms();
+    for (size_t i = 1; i < numHist; ++i)
+      numBins = std::max(numBins, data[i]->size());
```
SimonHeybrock: Given that many algorithms that use `blocksize` might be assuming that all histograms have the same size, wouldn't it make more sense to throw an exception upon mismatch? That we we could catch cases where algorithm do not support variable sizes yet. If we simply return the maximum it seems like we could corrupt data easily...?

SimonHeybrock (changes requested):
So far I only had a look at the key point. Please see comments below.

Apart from that, shouldn't changes also be made for `EventWorkspace`?

Framework/DataObjects/src/Workspace2D.cpp
peterfpeterson: That is what all of the other changes were about. I looked over the various places that `blocksize` was used and changed to more appropriate methods.

Framework/API/src/WorkspaceOpOverloads.cpp
SimonHeybrock: As you mentioned on Slack: Out of scope. :+1:

Framework/DataObjects/src/Workspace2D.cpp
SimonHeybrock: Ok, if you are certain that you covered all cases. But at the very least I suppose the method should be renamed to `maximumHistogramSizeInWorkspace` or something like that, and the Doxygen be fixed.

SimonHeybrock:
For future reference (discussed on Slack): I believe changing the behavior of `blocksize()` is not the correct way of doing things given that it is exposed to Python and may thus break user scripts.

peterfpeterson:
The only thing that appears to be left is to re-write `XDataConverter` and its two sub-classes: `ConvertToPointData` and `ConvertToHistogram`. Since histogram knows everything, the implementation should be significantly smaller than what it is.

Framework/DataObjects/src/EventWorkspace.cpp
```diff
@@ -148,9 +148,11 @@ size_t EventWorkspace::blocksize() const {
                            "therefore cannot determine blocksize (# of bins).");
   } else {
     size_t numBins = 0;
-    for (const auto *iter : data) {
+    for (const auto *iter : data)
       numBins = std::max(numBins, iter->histogram_size());
-    }
+    if (numBins != data[0]->histogram_size())
```
SimonHeybrock: See `Workspace2D`.

SimonHeybrock:
You will also need to change some workspace-internal use of `blocksize()`, in particular in the computation of MatrixWorkspace::size()` which right now would always fail on you for variable/length workspaces.

SimonHeybrock:
@peterfpeterson Regarding `XDataConverter`, I think this may work (but may indicate another problem that we need to fix?):

1. Change https://github.com/mantidproject/mantid/blob/master/Framework/Algorithms/src/XDataConverter.cpp#L56 to `auto outputWS = inputWS->clone()'`.
2. Use `outputWS->setHistogram` (with the converted X) instead of `setSharedX`, because the latter enforces a size check whereas the former doesn't.

Now the problem is: As far as I can tell this exploits a loophole, namely the workspace will temporarily contain both histogram *and* point data. Probably this should be fixed?

In that case the solution to above problem would be to use basically what is there already but with dummy `numXValues`, `numYValues` so you do not need `blocksize()`, then again use `setHistogram` in the second step to avoid size checks in `Histogram`.

Framework/API/src/WorkspaceOpOverloads.cpp
```diff
@@ -412,22 +412,26 @@ bool WorkspaceHelpers::commonBoundaries(const MatrixWorkspace &WS) {
     return true;
 
   // But even if they're not they could still match...
-  const double commonSum =
-      std::accumulate(WS.readX(0).begin(), WS.readX(0).end(), 0.);
+  const auto &x_0 = WS.readX(0);
```
SimonHeybrock: `readX` -> `x`.

Framework/API/src/WorkspaceOpOverloads.cpp
```diff
@@ -412,22 +412,26 @@ bool WorkspaceHelpers::commonBoundaries(const MatrixWorkspace &WS) {
...
+  const double commonSum = std::accumulate(x_0.begin(), x_0.end(), 0.);
   // If this results in infinity or NaN, then we can't tell - return false
   if (!std::isfinite(commonSum))
     return false;
   const size_t numHist = WS.getNumberHistograms();
   for (size_t j = 1; j < numHist; ++j) {
-    const double sum =
-        std::accumulate(WS.readX(j).begin(), WS.readX(j).end(), 0.);
+    const auto &x_j = WS.readX(j);
```
SimonHeybrock: `readX` -> `x`.

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -376,7 +376,7 @@ void MaxEnt::exec() {
 * counts
 * @return : Spectrum 'spec' as a complex vector
 */
-std::vector<double> MaxEnt::toComplex(const API::MatrixWorkspace_sptr &inWS,
+std::vector<double> MaxEnt::toComplex(API::MatrixWorkspace_const_sptr &inWS,
```
SimonHeybrock: Since you already change things like this, consider going all the way and pass a reference to the workspace.

Framework/Algorithms/test/AppendSpectraTest.h
```diff
@@ -216,8 +216,10 @@ class AppendSpectraTest : public CxxTest::TestSuite {
     for (size_t wi = 0; wi < out->getNumberHistograms(); wi++) {
       TS_ASSERT_EQUALS(out->getSpectrum(wi).getSpectrumNo(), specnum_t(wi));
       TS_ASSERT(!out->getSpectrum(wi).getDetectorIDs().empty());
-      for (size_t x = 0; x < out->blocksize(); x++)
-        TS_ASSERT_DELTA(out->readY(wi)[x], 2.0, 1e-5);
+      const auto &y = out->readY(wi);
+      const size_t numBins = y.size();
+      for (size_t i = 0; i < numBins; i++)
```
SimonHeybrock: Use range-base for?

Framework/Algorithms/test/ConjoinWorkspacesTest.h
```diff
@@ -246,9 +246,11 @@ class ConjoinWorkspacesTest : public CxxTest::TestSuite {
     TS_ASSERT_EQUALS(out->getNumberHistograms(), 15);
     TS_ASSERT_EQUALS(out->blocksize(), numBins);
 
-    for (size_t wi = 0; wi < out->getNumberHistograms(); wi++)
-      for (size_t x = 0; x < out->blocksize(); x++)
-        TS_ASSERT_DELTA(out->readY(wi)[x], 2.0, 1e-5);
+    for (size_t wi = 0; wi < out->getNumberHistograms(); wi++) {
+      const auto &y = out->readY(wi);
+      for (size_t j = 0; j < y.size(); ++j)
```
SimonHeybrock: Use range-based for?

Framework/Algorithms/test/PlusMinusTest.in.h
```diff
@@ -844,14 +844,19 @@ class @PLUSMINUSTEST_CLASS@ : public CxxTest::TestSuite
       bool breakOut=false;
       for (size_t wi=0; wi < work_out1->getNumberHistograms(); wi++)
       {
-        for (size_t i=0; i<work_out1->blocksize(); i++)
+        const auto &xIn = work_in1->readX(wi);
+        const auto &xOut = work_out1->readX(wi);
+        const auto &yOut = work_out1->readY(wi);
+        const auto &eOut = work_out1->readE(wi);
```
SimonHeybrock: Use new access methods `x`, `y,` etc.

Framework/CurveFitting/test/MultiDomainCreatorTest.h
```diff
@@ -62,37 +62,37 @@ class MultiDomainCreatorTest : public CxxTest::TestSuite {
     ws1.reset(new WorkspaceTester);
     ws1->initialize(1, 10, 10);
     {
-      Mantid::MantidVec &x = ws1->dataX(0);
-      Mantid::MantidVec &y = ws1->dataY(0);
+      auto &x = ws1->dataX(0);
```
SimonHeybrock: `mutableX`

Framework/CurveFitting/test/MultiDomainCreatorTest.h
```diff
@@ -62,37 +62,37 @@ class MultiDomainCreatorTest : public CxxTest::TestSuite {
...
       // Mantid::MantidVec& e = ws1->dataE(0);
-      for (size_t i = 0; i < ws1->blocksize(); ++i) {
-        x[i] = 0.1 * double(i);
-        y[i] = 1.0;
+      for (size_t i = 0; i < x.size(); ++i) {
+        x[i] = 0.1 * static_cast<double>(i);
       }
+      auto &y = ws1->dataY(0);
+      std::fill(y.begin(), y.end(), 1.);
```
SimonHeybrock: `ws->mutableY(0) = 1.0;`

Framework/CurveFitting/test/MultiDomainCreatorTest.h
```diff
@@ -62,37 +62,37 @@ class MultiDomainCreatorTest : public CxxTest::TestSuite {
...
     }
 
     ws2.reset(new WorkspaceTester);
     ws2->initialize(1, 10, 10);
     {
-      Mantid::MantidVec &x = ws2->dataX(0);
-      Mantid::MantidVec &y = ws2->dataY(0);
+      auto &x = ws2->dataX(0);
```
SimonHeybrock: mutableX

Framework/CurveFitting/test/MultiDomainCreatorTest.h
```diff
@@ -62,37 +62,37 @@ class MultiDomainCreatorTest : public CxxTest::TestSuite {
...
       // Mantid::MantidVec& e = ws2->dataE(0);
-      for (size_t i = 0; i < ws2->blocksize(); ++i) {
-        x[i] = 1 + 0.1 * double(i);
-        y[i] = 2.0;
+      for (size_t i = 0; i < x.size(); ++i) {
+        x[i] = 1. + 0.1 * static_cast<double>(i);
       }
+      auto &y = ws2->dataY(0);
+      std::fill(y.begin(), y.end(), 2.);
```
SimonHeybrock: `mutableY = 2.0`

Framework/CurveFitting/test/MultiDomainCreatorTest.h
```diff
@@ -62,37 +62,37 @@ class MultiDomainCreatorTest : public CxxTest::TestSuite {
...
     }
 
     ws3.reset(new WorkspaceTester);
     ws3->initialize(1, 10, 10);
     {
-      Mantid::MantidVec &x = ws3->dataX(0);
-      Mantid::MantidVec &y = ws3->dataY(0);
+      auto &x = ws3->dataX(0);
```
SimonHeybrock: .

Framework/CurveFitting/test/MultiDomainCreatorTest.h
```diff
@@ -62,37 +62,37 @@ class MultiDomainCreatorTest : public CxxTest::TestSuite {
...
       // Mantid::MantidVec& e = ws3->dataE(0);
-      for (size_t i = 0; i < ws3->blocksize(); ++i) {
-        x[i] = 2 + 0.1 * double(i);
-        y[i] = 3.0;
+      for (size_t i = 0; i < x.size(); ++i) {
+        x[i] = 2. + 0.1 * static_cast<double>(i);
       }
+      auto &y = ws3->dataY(0);
+      std::fill(y.begin(), y.end(), 3.);
```
SimonHeybrock: = 3.0

Framework/DataObjects/src/FractionalRebinning.cpp
```diff
@@ -81,13 +81,13 @@ void normaliseOutput(MatrixWorkspace_sptr outputWS,
                      boost::shared_ptr<Progress> progress) {
   for (int64_t i = 0; i < static_cast<int64_t>(outputWS->getNumberHistograms());
        ++i) {
-    MantidVec &outputY = outputWS->dataY(i);
-    MantidVec &outputE = outputWS->dataE(i);
-    for (size_t j = 0; j < outputWS->blocksize(); ++j) {
+    const auto &outputX = outputWS->readX(i);
+    auto &outputY = outputWS->dataY(i);
+    auto &outputE = outputWS->dataE(i);
```
SimonHeybrock: Use `x`/`mutableY`, etc.

qt/scientific_interfaces/Indirect/ResNorm.cpp
```diff
@@ -295,8 +295,7 @@ void ResNorm::previewSpecChanged(int value) {
       fit->setSharedY(0, fitWs->sharedY(1));
       fit->setSharedE(0, fitWs->sharedE(1));
 
-      for (size_t i = 0; i < fit->blocksize(); i++)
-        fit->mutableY(0)[i] /= scaleFactors->cell<double>(m_previewSpec);
+      fit->mutableY(0) /= scaleFactors->cell<double>(m_previewSpec);
```
SimonHeybrock: :+1:

SimonHeybrock (changes requested):
Just mostly minor comments on the rest of the changes:

peterfpeterson:
You're solution for  `XDataConverter` and its two sub-classes (`ConvertToPointData` and `ConvertToHistogram`) didn't work out since the output needs to be converted to `Workspace2D` from `EventWorkspace`. `ws->clone()` isn't quite the right thing to do.

This is exercised by `AlgorithmsTest_UnwrapMonitorTest` which calls `Rebin` which calls `ConvertToHistogram`.

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -957,11 +963,23 @@ bool MatrixWorkspace::isCommonBins() const {
   if (!m_isCommonBinsFlagSet) {
     m_isCommonBinsFlag = true;
 
+    const size_t numHist = getNumberHistograms();
+    bool oneNonEmpty = false;
+    if (numHist > 1) {
+      !readY(0).empty();
```
SimonHeybrock: Please do not use `readY` etc., it is deprecated

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -957,11 +963,23 @@ bool MatrixWorkspace::isCommonBins() const {
...
+      for (size_t i = 1; i < numHist; ++i) {
+        if (!readY(0).empty()) {
+          oneNonEmpty = true;
```
SimonHeybrock: Maybe unit unit test should be added for the case of empty histograms?

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -1008,8 +1026,8 @@ void MatrixWorkspace::maskBin(const size_t &workspaceIndex,
         workspaceIndex, this->getNumberHistograms(),
         "MatrixWorkspace::maskBin,workspaceIndex");
   // Then check the bin index
-  if (binIndex >= this->blocksize())
-    throw Kernel::Exception::IndexError(binIndex, this->blocksize(),
+  if (binIndex >= readY(workspaceIndex).size())
```
SimonHeybrock: y

Framework/Algorithms/src/XDataConverter.cpp
```diff
@@ -79,6 +79,12 @@ void XDataConverter::exec() {
   setProperty("OutputWorkspace", outputWS);
 }
 
+std::size_t
+XDataConverter::getNewYSize(const API::MatrixWorkspace_sptr inputWS) {
+  // this is the old behavior of MatrixWorkspace::blocksize()
+  return inputWS->y(0).size();
```
SimonHeybrock: Won't this cause a size mismatch exception in line 69?

Framework/DataObjects/src/Workspace2D.cpp
```diff
@@ -134,13 +135,26 @@ size_t Workspace2D::getNumberHistograms() const {
...
-  return (!data.empty())
-             ? static_cast<ISpectrum const *>(data[0])->dataY().size()
-             : 0;
+  if (data.empty()) {
+    return 0;
+  } else {
+    size_t numBins = 0;
+    for (const auto *iter : data)
+      numBins = std::max(numBins, iter->size());
+    if (numBins != data[0]->size())
```
SimonHeybrock: Doesn't this still have the same issue as the previous version? What if `data[0]->size()` is the maximum? This will not catch it will it?

Maybe this?
```cpp
for (const auto *iter : data)
  if (iter->size() != data[0]->size())
    throw;
```

And most importantly: Add unit tests for this please!

SimonHeybrock (changes requested):
Hi Pete, some more comments, I think there are two actual issues that need fixing:

Framework/API/src/MatrixWorkspace.cpp
peterfpeterson: `ValidateCommonBoundaries` already has one

SimonHeybrock: @peterfpeterson Sorry, can't find a test covering this new special case, link?

peterfpeterson: It is CommonBinsValidatorTest L20. I'm adding a couple of tests in MatrixWorkspace

SimonHeybrock: Yes but that does not seem to cover the new code, namely cases where some of the histograms are empty while others are not?

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -957,18 +963,28 @@ bool MatrixWorkspace::isCommonBins() const {
...
+          break;
+        }
+      }
+    }
+
     // there being only one or zero histograms is accepted as not being an error
-    if (blocksize() || getNumberHistograms() > 1) {
+    if (oneNonEmpty || numHist > 1) {
       // otherwise will compare some of the data, to save time just check two
       // the first and the last
```
SimonHeybrock: I think this is now definitely not enough any more, is it? Maybe at least the length should be checked, and if it it all the same continue with the existing "test" via first and last?

Framework/API/src/WorkspaceOpOverloads.cpp
SimonHeybrock: Maybe also here a length check should be added? Otherwise having variable-length histograms introduces new failure modes to the method, e.g., if you have a workspace with two histograms with bin edges `{1,2,4}` and `{3,4}`.

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -171,8 +171,13 @@ const std::string MatrixWorkspace::toString() const {
...
   os << id() << "\n"
      << "Title: " << getTitle() << "\n"
-     << "Histograms: " << getNumberHistograms() << "\n"
-     << "Bins: " << blocksize() << "\n";
+     << "Histograms: " << getNumberHistograms() << "\n";
+
+  try {
+    os << "Bins: " << blocksize() << "\n";
+  } catch (std::length_error) {
+    os << "Bins: variable\n"; // TODO shouldn't use try/catch
```
OwenArnold: So is this try-catch still required?

OwenArnold (changes requested):
Generally looks good. There is no mention of these changes in any documentation or release notes.

Framework/API/src/MatrixWorkspace.cpp
peterfpeterson: Since there isn't currently a good way to check if `blocksize()` will work versus raise an exception, it is still needed.

peterfpeterson:
[MatrixWorkspaceTest](https://github.com/mantidproject/mantid/pull/20451/files#diff-7ea08eec89b97f0ffb9087b604241e7f) has checks for empty workspaces and the `blocksize` family of functions.

SimonHeybrock:
Remaining things to address before approval:

- Unit tests for new behavior of `Workspace2D::blocksize()` and `EventWorkspace::blocksize()` (throwing exceptions).
- Unit tests for of `MatrixWorkspace::isCommonBins()` (does it work for variable X length?).
- Update doxygen of `blocksize()`.

SimonHeybrock (approved):

Framework/API/src/MatrixWorkspace.cpp
OwenArnold: Ok. Also I think this should catch should be modified to catch by reference, but I don't want to hold this up.

OwenArnold (approved):

martyngigg:
Should this core of a change really go in at this point?

PR #20380, opened 2017-09-05
Re #20302: Replaces DataProcessor class name prefixes with a namespace

edwardb96:
Renames a number of classes/files in the DataProcessorUI directory, reducing the file/class name size to improve readability and uses a `namespace` instead.

**To test:**
CI checks and code review should be sufficient.

Fixes #20302 . 

**Release Notes** 

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)?
- [ ] Are the unit tests small and test the class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] Do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant (user and developer) documentation been added/updated?

Does everything look good? Mark the review as **Approve**. A member of `@mantidproject/gatekeepers` will take care of it.

Issue #20302, opened 2017-08-21
Replace DataProcessor class name prefix with a namespace


edwardb96:
Currently many of the data processor widget classes are prefixed with `DataProcessor` which makes the class names unnecessarily long and related code difficult to read.

I propose that these classes be moved into their own namespace with the prefixes removed from both the class and filename.

PR #20452, opened 2017-09-12
Revert "Re #20302: Replaces DataProcessor class name prefixes with a namespace"

peterfpeterson:
Reverts mantidproject/mantid#20380 because it broke [`master` builds](http://builds.mantidproject.org/job/master_incremental/3213/)


peterfpeterson:
@edwardb96 and @gemmaguest should know that the merge was reverted.

PR #20405, opened 2017-09-07
NumericAxis, refactored out `m_binEdges` and the static `indexOfValue`

gvardany:
This PR improves the design of `NumericAxis` by removing the bin edges member, and reducing the complexity for `setValue` usages from Nˆ2 to N (see the issue). It will compute the bin edges only if they are explicitly asked. `indexOfValue` will be found directly from bin centres. 
In the contrary, the `SpectrumAxis` and `BinEdgeAxis` override this behaviour to retrieve the lower bound from bin edges. 

The static `NumericAxis::indexOfValue`, previously used by `SpectrumAxis` and `BinEdgeAxis` is removed.

**To test:**
Code review and tests pass.

Fixes #20387 . 

**Release Notes** 

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)?
- [ ] Are the unit tests small and test the class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] Do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant (user and developer) documentation been added/updated?

Does everything look good? Mark the review as **Approve**. A member of `@mantidproject/gatekeepers` will take care of it.

PR #20455, opened 2017-09-13
Speed Up Color Fill Plot for Point Data

### LINKED ISSUE #20454: Color Fill Plot for Point Data Workspaces is Slow
Try performing a color fill plot in Mantid, on the two workspaces below. Plotting the histogram workspace is quick, but the point data one is incredibly slow.

```
ws = CreateSampleWorkspace(
    Function='Flat background',
    NumBanks=10,
    BankPixelWidth=100)

ws2 = ConvertToPointData(ws)
```

This issue should fix this problem while preserving the existing behaviour of plotting point data workspaces.

ianbush:
A fix to avoid re-calculating the `BinEdges` for every call to `MatrixWorkspace::getSignalAtCoord`. This is a fix for the issue, but longer term it would be preferable to have a nicer method in the appropriate `Axis` class to do this in a robust way.

More tests were added to check the intermediate cases.

**To test:**

Plot some point data workspaces using the color fill plot option. Check the performance is similar to that for plotting histogram workspaces. For example try the workspaces from the script below.
```
ws = CreateSampleWorkspace(
    Function='Flat background',
    NumBanks=10,
    BankPixelWidth=100)

ws2 = ConvertToPointData(ws)
```

Fixes #20454. 

*Does not need to be in the release notes.* - not a behavioural change or user reported bug.
---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


josephframsay:
Code quality of the changes looks good, but when doing a colour fill plot on the workspace in the test script, the plot crashes due to (I think) the value of `i` on line 1540 in `MatrixWorkspace.cpp` being 0

josephframsay (changes requested):
Code quality of the changes looks good, but when doing a colour fill plot on the workspace in the test script, the plot crashes due to (I think) the value of `i` on line 1540 in `MatrixWorkspace.cpp` being 0

ianbush:
@josephframsay I've tested this on Linux and Windows and was unable to reproduce this issue, using the script provided. Can you double check the line number of the crash, and let me know if you have any other settings in Mantid that might explain a difference in behavior?

josephframsay (approved):

josephframsay:
The above bug was fixed, and plotting performance is improved

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -1520,13 +1520,26 @@ signal_t MatrixWorkspace::getSignalAtCoord(
...
 
   if (wi < nhist) {
-    const auto &X = this->binEdges(wi);
-    auto it = std::lower_bound(X.cbegin(), X.cend(), x);
-    if (it == X.end()) {
+    const auto &xVals = x(wi);
+    auto it = std::lower_bound(xVals.cbegin(), xVals.cend(), xCoord);
+    if (it == xVals.end()) {
+      if (!isHistogramData()) {
+        auto deltaX = (xVals[xVals.size() - 1] - xVals[xVals.size() - 2]) / 2.0;
```
martyngigg: `vector.back()` would be useful here to avoid having to call `size()` all of the time. It's also quicker to multiply than to divide so it would be better to write `0.5*` than `/2.0`. How about something like:

```cpp
auto binEnd = xVals.back() + 0.5*(xVals.back() - *std::next(xVals.rbegin()));
if(xCoord < binEnd)
  return yVals.back()
```

Framework/API/test/MatrixWorkspaceTest.h
```diff
@@ -918,16 +918,22 @@ class MatrixWorkspaceTest : public CxxTest::TestSuite {
...
     // Create a test workspace
     const auto ws = createTestWorkspace(4, 5, 5);
+    auto normType = Mantid::API::NoNormalization;
 
     // Get signal at coordinates
-    std::vector<coord_t> coords = {0.0, 1.0};
-    TS_ASSERT_DELTA(
-        ws.getSignalAtCoord(coords.data(), Mantid::API::NoNormalization), 0.0,
-        1e-5);
+    std::vector<coord_t> coords = {-1.0, 1.0};
```
martyngigg: Good tests.

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -1520,13 +1520,26 @@ signal_t MatrixWorkspace::getSignalAtCoord(
...
+      }
       // Out of range
       return std::numeric_limits<double>::quiet_NaN();
     } else {
-      size_t i = (it - X.begin());
+      size_t i = (it - xVals.begin());
+      if (!isHistogramData()) {
+        // In this case we have point data. Find the closest index to our x
+        // coordinate, if the value is in range.
+        auto deltaX = (xVals[1] - xVals[0]) / 2.0;
```
martyngigg: Similarly here with `/2.0`.

Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -1520,13 +1520,27 @@ signal_t MatrixWorkspace::getSignalAtCoord(
...
 
   if (wi < nhist) {
-    const auto &X = this->binEdges(wi);
-    auto it = std::lower_bound(X.cbegin(), X.cend(), x);
-    if (it == X.end()) {
+    const auto &xVals = x(wi);
+    auto it = std::lower_bound(xVals.cbegin(), xVals.cend(), xCoord);
+    if (it == xVals.end()) {
+      if (!isHistogramData()) {
+        auto binEnd =
```
martyngigg: I just had another think about this. If the value is not found couldn't it also be in a "bin" before the first point?

ianbush: No, I don't think so. `std::lower_bound` returns the first point that is greater than or equal to `xCoord`.

martyngigg: Good point. My brain was misfiring for a minute there.

martyngigg:
I should have thought of this but is `https://github.com/mantidproject/mantid/blob/master/Framework/API/inc/MantidAPI/Axis.h#L92` not the correct method to use?

ianbush:
The answer to that is not clear to me at the moment, as changes in PR #20405 might affect the answer...

martyngigg:
Ah, good point. Should we be using on of [VectorHelper]( https://github.com/mantidproject/mantid/blob/master/Framework/Kernel/inc/MantidKernel/VectorHelper.h#L72) methods maybe?

ianbush:
I've just pushed some changes, these now use `Kernel::VectorHelper` methods. I've tested and I think the functionality remains the same.

martyngigg:
Unrelated test failure.

PR #20464, opened 2017-09-13
More robust fix for asynchronous python execution in MantidPlot

### LINKED ISSUE #20437: runPythonScript causing crash
In muon analysis the code crashes when runPythonScript is called. 

To reproduce the bug:
1. load some data into muon analysis
2. Change to the data analysis tab
3. Get an error message

I have had the crash on both windows and unix.

martyngigg:
Description of work.

Uses an internal Python method to check if the GIL has been
acquired by the current thread rather than relying on our own flag.

This only applies in the situations where an executing python script then calls the `runPythonScript` method itself to call more code. The check for the GIL state was not robust enough to cover all cases before.  

**To test:**

Try an interface:
* Open the Muon Analysis GUI
* Load some muon data (MUSR data is in the doctest data)
* click the Data Analysis tab and mantid would crash before this fix

Try a recursive python script:

* Create a new python file with the following contents:
```python
import mantidplot

for i in range(10):
    mantidplot.runPythonScript("CreateWorkspace(OutputWorkspace='test_{}')".format(i), True)
```
* click View->Manage Custom Menus in MantidPlot
* click add menu and enter a name
* click add script and find the file you just saved
* highlight both the script and menu in the tree and click the `->` arrow to attach the script to the menu.
* go to the new menu and run the script
* before the fix MantidPlot would crash.

Fixes #20437

**Release Notes** 

This was broken during this dev cycle so no need for release notes

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidPlot/src/ApplicationWindow.cpp
+++ b/MantidPlot/src/ApplicationWindow.cpp
@@ -15064,7 +15064,7 @@ bool ApplicationWindow::runPythonScript(const QString &code, bool async,
   }
   bool success(false);
   if (async) {
-    m_iface_script->recursiveAsyncSetup();
+    const bool locked = m_iface_script->recursiveAsyncSetup();
     auto job = m_iface_script->executeAsync(ScriptCode(code));
     // Start a local event loop to keep processing events
     // while we are running the script. Inspired by the IPython
@@ -15077,7 +15077,7 @@ bool ApplicationWindow::runPythonScript(const QString &code, bool async,
       eventLoop.exec();
       timer.stop();
     }
-    m_iface_script->recursiveAsyncTeardown();
+    m_iface_script->recursiveAsyncTeardown(locked);
     success = job.result();
   } else {
     success = m_iface_script->execute(ScriptCode(code));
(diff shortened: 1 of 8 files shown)


AnthonyLim23:
It works if I change the last line of the script to 

`mantidplot.runPythonScript("CreateWorkspace(DataX=[0],DataY=[0],OutputWorkspace='test_{}')".format(i), True)`

The code looks good.

AnthonyLim23 (approved):

peterfpeterson:
From [python's changelog](https://docs.python.org/3/whatsnew/changelog.html) it looks like changing
```c++
bool PythonGIL::locked() {
#if PY_VERSION_HEX < 0x03000000
  PyThreadState *ts = _PyThreadState_Current;
#else
  PyThreadState *ts = _PyThreadState_UncheckedGet();
#endif
  return (ts && ts == PyGILState_GetThisThreadState());
}
```
will fix the python3 builds. I only tried the second test in python3.

martyngigg:
Thanks Pete. I've pushed the fix.

PR #20465, opened 2017-09-13
Allow std::array to be used with ListValidator

### LINKED ISSUE #20101: Switch algorithms using std::vector to std::array for known length list validators
Currently within the code base we accept std::vector but not std::array for a `ListValidator`. 
This incurs us having to setup a dynamic allocator in debug mode and relies on the compiler to correctly deduce it should be optimised on Release mode.

By using std::array the compiler allocates the memory on the stack and the optimiser can better reason with the additional constraints imposed. 

- We should add a new constructor to support std::array in `ListValidator.h`
- Possibly remove the std::vector version if it does not make sense to change the potential options at runtime
- Switch several places with the following code pattern:
```c++
std::vector<std::string> options {foo, bar, baz};
// to
std::array<std::string, 3> options {foo, bar, baz};
```

josephframsay:
Added a constructor to `ListValidator` to accept `std::array`, and changed many occurences of this pattern:
```
std::vector<std::string> options {foo, bar, baz};
boost::make_shared<ListValidator<std::string>>(options)
```
to this pattern:
```
std::array<std::string, 3> options {foo, bar, baz};
boost::make_shared<ListValidator<std::string>>(options)
```

The compiler doesn't have to worry about dynamic allocation, so can allocate memory on the stack, saving some time.

Some minor refactoring also took place, where use `using namespace std` was removed

**To test:**

Code review, and make sure all the unit tests for Framework still pass. In particular, I think all the possible occurences of the first pattern have been replaced, but if you come across any that I've missed then shout

Fixes #20101 

**Release Notes** 
*Does not need to be in the release notes, as the change is internal*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Kernel/inc/MantidKernel/ListValidator.h
+++ b/Framework/Kernel/inc/MantidKernel/ListValidator.h
@@ -48,40 +48,12 @@ public:
   ListValidator() : TypedValidator<TYPE>(), m_allowMultiSelection(false) {}
 
   /** Constructor
-   *  @param values :: A set of values consisting of the valid values
-   *  @param allowMultiSelection :: True if the list allows multi selection
-   */
-  explicit ListValidator(const std::set<TYPE> &values,
-                         const bool allowMultiSelection = false)
-      : TypedValidator<TYPE>(), m_allowedValues(values.begin(), values.end()),
-        m_allowMultiSelection(allowMultiSelection) {
-    if (m_allowMultiSelection) {
-      throw Kernel::Exception::NotImplementedError(
-          "The List Validator does not support Multi selection yet");
-    }
-  }
-
-  /** Constructor
-   *  @param values :: An unordered set of values consisting of the valid values
-   *  @param allowMultiSelection :: True if the list allows multi selection
-   */
-  explicit ListValidator(const std::unordered_set<TYPE> &values,
-                         const bool allowMultiSelection = false)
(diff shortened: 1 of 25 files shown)


Framework/Kernel/inc/MantidKernel/ListValidator.h
```diff
@@ -76,10 +76,39 @@ template <typename TYPE> class ListValidator : public TypedValidator<TYPE> {
   }
 
   /** Constructor
-   *  @param values :: A vector of the valid values
+   *  @param values :: An array of the valid values
    *  @param aliases :: Optional aliases for the valid values.
    *  @param allowMultiSelection :: True if the list allows multi selection
    */
+  template <std::size_t SIZE>
+  explicit ListValidator(const std::array<TYPE, SIZE> &values,
```
martyngigg: @josephframsay This is a nice improvement. The only thing I'm not too keen on is the code duplication in the constructors.

We could swap the template here for `template<typename T>` and then delete the second `vector`-based constructor. As we're using iterators then this should work for all of the container types.

martyngigg (changes requested):

martyngigg (approved):
This looks good to me now. Even though it's marked maintenance I'm happy for this to be merged as soon as the builds pass.

PR #20494, opened 2017-09-15
Replace BOOST_FOREACH with range-based for loops

peterfpeterson:
[`BOOST_FOREACH`](https://theboostcpplibraries.com/boost.foreach) appears to be from a time when range-based for loops existed. This gets rid of them.

**To test:**

Code review and build servers passing is sufficient.

*Does not need to be in the release notes* because it is way too far into the weeds

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Algorithms/src/CalculateTransmission.cpp
```diff
@@ -187,7 +186,7 @@ void CalculateTransmission::exec() {
     beamMonitorIndex = getIndexFromDetectorID(*sampleWS, beamMonitorID);
     logIfNotMonitor(sampleWS, directWS, beamMonitorIndex);
 
-    BOOST_FOREACH (size_t transmissionIndex, transmissionIndices)
+    for (auto transmissionIndex : transmissionIndices)
```
quantumsteve: This could also be done with [std::find](http://en.cppreference.com/w/cpp/algorithm/find).

Framework/Algorithms/src/RadiusSum.cpp
```diff
@@ -227,9 +225,9 @@ void RadiusSum::inputValidationSanityCheck() {
     throw std::invalid_argument(s.str());
   }
 
-  std::vector<double> boundary_limits = getBoundariesOfInputWorkspace();
+  const std::vector<double> boundary_limits = getBoundariesOfInputWorkspace();
   std::stringstream s;
-  BOOST_FOREACH (auto &value, boundary_limits)
+  for (const auto &value : boundary_limits)
     s << value << " , ";
```
quantumsteve: I think this loop leaves an extra `,` at the end of the string. You may be better of doing something like 

```
std::copy(boundary_limits.begin(), std::prev(boundary_limits.end()), std::ostream_iterator<double>(s, ", "));
s << boundary_limits.back();
```

quantumsteve (changes requested):
Changes look good. It's always nice to migrate to a more modern feature.

I'd suggest fixing the formatting issue (extra comma) in this PR instead of a separate one. Switching to `std::find` is your choice. I'm happy to approve it either way.

quantumsteve (approved):

martyngigg:
I assumed the build failures were unrelated at first but I can reproduce the failure of the `docs-qthelp` target on my machine. Starting MantidPlot and then trying to run the `AlignAndFocusPowder` algorithm gives me an exception:

alignandfocuspowder-exception

The C++ exception type is unknown to Python so the console error message in the build is not very helpful.

qt/widgets/common/src/PropertyWidget.cpp
```diff
@@ -40,15 +38,8 @@ double stringToRoundedNumber(const std::string &s) {
...
-
-  std::istringstream i(s);
-  double roundedNumber;
-
-  if (!(i >> roundedNumber))
-    throw std::runtime_error("");
-
-  return roundedNumber;
+    throw std::runtime_error("string contains a comma");
+  return std::stod(s);
```
quantumsteve: I suspect `std::stod` is throwing a `std::out_of_range` exception, which doesn't inherit from `std::runtime_error` and could be causing the doctest failure.

martyngigg: You're right. Applying the following diff

```diff
diff --git a/qt/python/mantidqtpython/mantidqt.sip b/qt/python/mantidqtpython/mantidqt.sip
index 249af60..d119fdd 100644
--- a/qt/python/mantidqtpython/mantidqt.sip
+++ b/qt/python/mantidqtpython/mantidqt.sip
@@ -162,7 +162,7 @@ class InterfaceManager
 %End
 public:
   MantidQt::API::AlgorithmDialog* createDialogFromName(const QString&, const int = -1,
-                                                       QWidget* = 0, bool = false);
+                                                       QWidget* = 0, bool = false) throw(std::exception);
   MantidQt::API::UserSubWindow* createSubWindow(const QString& interface_name, QWidget* parent = 0);
 
   void showHelpPage(const QString & url=QString());
```
gives a better (still not great) error message in Python. It at least matches the C++ one.

martyngigg (approved):
I tested several interfaces after the most recent change and all seem to handle the `EMPTY_DBL` defaults as before.

It should be good to merge when the builds pass.

PR #20501, opened 2017-09-18
Fix WorkspaceFactory create when axis size different from parent

### LINKED ISSUE #17907: WorkspaceFactory error after creating a workspace from parent with 1 histogram and TextAxis
### Expected behavior
`WorkspaceFactory.create(<parent>..)` produces a usable workspace inheriting the parent's meta data

### Actual behavior
When using `WorkspaceFactory.create(...)` on a workspace with TextAxis and 1 histogram (I haven't tested if it breaks with multiple histograms) the produced workspace causes an unexpected crash in Mantid, and makes that workspace unusable.

This **DOES** work when the workspace is with SpectraNumbers
Run the following script
```
dataX = []
dataY = []
YVals = ["Text"]
for x in range(20):
    dataX.append(x)
    dataY.append(x)

print dataX
print dataY
print YVals

we = CreateWorkspace(dataX, dataY, UnitX="Wavelength", \
VerticalAxisUnit="Text", VerticalAxisValues=YVals, YUnitLabel="Counts", \
WorkspaceTitle="workspacetitle")

# inherit from parent workspace, and copy the histogram 5 times
third_ws = WorkspaceFactory.create(we, NVectors=5)
mtd['labelledworkspace']=third_ws
```

### Platforms affected
Tested only on Ubuntu 14.04, but probably affects all

DanNixon:
Corrects detection of differing sized axis so that newly created workspaces always have an axis of sufficient length.

Keeps existing values in `TextAxis` when it is cloned with a different length.

**To test:**

- See script in original ticket
- The `labelledworkspace` should now be able to be opened
- It should have a valid `TextAxis` where the first value is copied from the parent workspace and the remaining values are empty

Fixes #17907.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/src/WorkspaceFactory.cpp
+++ b/Framework/API/src/WorkspaceFactory.cpp
@@ -51,16 +51,15 @@ MatrixWorkspace_sptr
 WorkspaceFactoryImpl::create(const MatrixWorkspace_const_sptr &parent,
                              size_t NVectors, size_t XLength,
                              size_t YLength) const {
-  bool differentSize(true);
   // Use the parent sizes if new ones are not specified
   if (NVectors == size_t(-1))
     NVectors = parent->getNumberHistograms();
+
   if (XLength == size_t(-1))
     XLength = parent->dataX(0).size();
-  if (YLength == size_t(-1)) {
-    differentSize = false;
+
+  if (YLength == size_t(-1))
     YLength = parent->blocksize();
-  }
 
   // If the parent is an EventWorkspace, we want it to spawn a Workspace2D (or
   // managed variant) as a child
@@ -72,7 +71,7 @@ WorkspaceFactoryImpl::create(const MatrixWorkspace_const_sptr &parent,
   MatrixWorkspace_sptr ws = create(id, NVectors, XLength, YLength);
 
   // Copy over certain parent data members
-  initializeFromParent(*parent, *ws, differentSize);
+  initializeFromParent(*parent, *ws);
 
   return ws;
 }
@@ -86,11 +85,9 @@ WorkspaceFactoryImpl::create(const MatrixWorkspace_const_sptr &parent,
  * @brief WorkspaceFactoryImpl::initializeFromParentWithoutLogs
  * @param parent
  * @param child
- * @param differentSize
  */
 void WorkspaceFactoryImpl::initializeFromParentWithoutLogs(
-    const MatrixWorkspace &parent, MatrixWorkspace &child,
-    const bool differentSize) const {
+    const MatrixWorkspace &parent, MatrixWorkspace &child) const {
   child.setTitle(parent.getTitle());
   child.setComment(parent.getComment());
   child.setInstrument(parent.getInstrument()); // This call also copies the
@@ -106,7 +103,7 @@ void WorkspaceFactoryImpl::initializeFromParentWithoutLogs(
   child.setDistribution(parent.isDistribution());
 
   // Only copy the axes over if new sizes are not given
-  if (!differentSize) {
+  if (child.blocksize() == parent.blocksize()) {
     // Only copy mask map if same size for now. Later will need to check
     // continued validity.
     child.m_masks = parent.m_masks;
@@ -128,7 +125,7 @@ void WorkspaceFactoryImpl::initializeFromParentWithoutLogs(
     const size_t newAxisLength = child.getAxis(i)->length();
     const size_t oldAxisLength = parent.getAxis(i)->length();
 
-    if (!differentSize || newAxisLength == oldAxisLength) {
+    if (newAxisLength == oldAxisLength) {
       // Need to delete the existing axis created in init above
       delete child.m_axes[i];
       // Now set to a copy of the parent workspace's axis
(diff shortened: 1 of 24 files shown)


Framework/API/src/TextAxis.cpp
```diff
@@ -28,7 +28,6 @@ Axis *TextAxis::clone(const std::size_t length,
                       const MatrixWorkspace *const parentWorkspace) {
   UNUSED_ARG(parentWorkspace)
   auto newAxis = new TextAxis(*this);
-  newAxis->m_values.clear();
```
SimonHeybrock: This change would give inconsistent behavior of `Axis::clone()`, which basically seems to be meant as a polymorphic constructor (*not* copy constructor, i.e., the returned `Axis` is empty). I am not sure this can be changed without causing trouble somewhere. Could you check with someone knowing more about `Axis`, maybe @martyngigg?

Framework/API/src/WorkspaceFactory.cpp
```diff
@@ -51,16 +51,23 @@ MatrixWorkspace_sptr
 WorkspaceFactoryImpl::create(const MatrixWorkspace_const_sptr &parent,
                              size_t NVectors, size_t XLength,
                              size_t YLength) const {
-  bool differentSize(true);
+  bool differentSize(false);
```
SimonHeybrock: Now the flag can be `false` only if the numbers of histogram did not change, which will change other observable behavior of `create`. In particular bin masking will not be copied any more (see `initializeFromParent` below) if the number of histograms changed. I am not saying this is the wrong thing to do but I cannot really make a call. Maybe this also warrants a wider discussion?

SimonHeybrock (changes requested):
Changes look a priori good and tests were added :+1:.

However, I have two comments regarding points that probably need discussion:

Framework/API/src/TextAxis.cpp
DanNixon: Yes, this is not critical in fixing the original issue and does make the behavior of `TextAxis` differ from that of other axes. I'll re add this.

Framework/API/src/WorkspaceFactory.cpp
DanNixon: From what I can tell bin masking is the only other time `differentSize` is used. That being the case it makes sense to me to remove this parameter and use `parent->blocksize() != child->blocksize()` as the test for copying bin masking.

For axes the use of `differentSize` makes very little sense to me as if the X axis does not change then you end up copying all axes from the parent, which leads to invalid child workspaces.

Framework/API/src/TextAxis.cpp
martyngigg: My recollection is that these methods were designed to be similar to the workspace factory in that they gave an empty object so I think adding this back is the best option.

Framework/API/src/WorkspaceFactory.cpp
martyngigg: The issue around the bin masks is that the map being copied contains a lookup against the workspace index so if you are changing the number of histograms the map is really not valid either. I have to say I think I would be very nervous at changing this behaviour.

martyngigg:
Can we agree that it would be better if this went into the next release? I feel it's a little late to put in this type of a change if it's risky.

Framework/Algorithms/src/Integration.cpp
```diff
@@ -380,7 +380,7 @@ MatrixWorkspace_sptr Integration::getOutputWorkspace(MatrixWorkspace_sptr inWS,
   if (inWS->id() == "RebinnedOutput") {
     MatrixWorkspace_sptr outWS = API::WorkspaceFactory::Instance().create(
         "Workspace2D", maxSpec - minSpec + 1, 2, 1);
-    API::WorkspaceFactory::Instance().initializeFromParent(*inWS, *outWS, true);
+    API::WorkspaceFactory::Instance().initializeFromParent(*inWS, *outWS);
```
SimonHeybrock: If `inWs` has `blocksize() == 1` this will now copy bin masks, whereas the old code did not, I think?

Framework/Algorithms/src/WorkspaceJoiners.cpp
```diff
@@ -40,7 +40,7 @@ MatrixWorkspace_sptr WorkspaceJoiners::execWS2D(const MatrixWorkspace &ws1,
       "Workspace2D", totalHists, ws1.x(0).size(), ws1.y(0).size());
   // Copy over stuff from first input workspace. This will include the spectrum
   // masking
-  WorkspaceFactory::Instance().initializeFromParent(ws1, *output, true);
+  WorkspaceFactory::Instance().initializeFromParent(ws1, *output);
```
SimonHeybrock: Same here, this will now copy bin masks? There is some code handling that below.

Framework/Crystal/src/PeakIntegration.cpp
```diff
@@ -95,8 +95,7 @@ void PeakIntegration::exec() {
   outputW = API::WorkspaceFactory::Instance().create(
       inputW, peaksW->getNumberPeaks(), YLength + 1, YLength);
   // Copy geometry over.
-  API::WorkspaceFactory::Instance().initializeFromParent(*inputW, *outputW,
-                                                         true);
+  API::WorkspaceFactory::Instance().initializeFromParent(*inputW, *outputW);
```
SimonHeybrock: Same here.

Framework/DataHandling/src/EventWorkspaceCollection.cpp
```diff
@@ -227,7 +227,7 @@ size_t EventWorkspaceCollection::getNumberEvents() const {
 void EventWorkspaceCollection::resizeTo(const size_t size) {
   for (auto &ws : m_WsVec) {
     auto tmp = createWorkspace<DataObjects::EventWorkspace>(size, 2, 1);
-    WorkspaceFactory::Instance().initializeFromParent(*ws, *tmp, true);
+    WorkspaceFactory::Instance().initializeFromParent(*ws, *tmp);
```
SimonHeybrock: Here it does not matter, I think?

SimonHeybrock:
I did not go through all changes to `initializeFromParent`, but I think there would be issues with bin masking with this change?

PR #20507, opened 2017-09-18
Get rid of unused arg warnings in gcc7

peterfpeterson:
Exactly what the summary suggests.

**To test:**

Since you can only see the warnings on gcc7, you'll have to compile there to see any difference. Nominally, just make sure the builds pass.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/test/ExperimentInfoTest.h
+++ b/Framework/API/test/ExperimentInfoTest.h
@@ -467,7 +467,7 @@ public:
     mappings.emplace(1, std::set<Mantid::detid_t>{2});
     expt.cacheDetectorGroupings(mappings);
 
-    size_t index = 0;
+    size_t index{0};
     TS_ASSERT_THROWS_NOTHING(index = expt.groupOfDetectorID(1));
     TS_ASSERT_EQUALS(index, 0);
   }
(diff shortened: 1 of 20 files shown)


Framework/HistogramData/test/BinEdgesTest.h
```diff
@@ -23,12 +23,14 @@ class BinEdgesTest : public CxxTest::TestSuite {
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wunused-value"
 #endif
```
quantumsteve: Can you remove the clang pragmas?

Framework/API/test/ExperimentInfoTest.h
```diff
@@ -467,7 +467,7 @@ class ExperimentInfoTest : public CxxTest::TestSuite {
     mappings.emplace(1, std::set<Mantid::detid_t>{2});
     expt.cacheDetectorGroupings(mappings);
 
-    size_t index;
+    size_t index = 0;
```
quantumsteve: While this is correct and I *don't* see any need to change it, part of me wants to use uniform initialization here. The newer syntax is a bit more restrictive about potential narrowing casts.

quantumsteve (changes requested):
This dramatically reduces the number of warnings generated when building tests. It also removes some ugly code used to suppress similar warnings from clang.

Would you try removing the additional pragmas noted in this review?

quantumsteve (approved):

PR #20511, opened 2017-09-19
Refactor index handling in LoadEventNexus

### LINKED ISSUE #20500: MPI-compatible workspace setup in LoadEventNexus
Make setup of workspaces, in particular partitioning of indices/spectra compatible with MPI.
Part of #20344.

SimonHeybrock:
Refactor index handling in `LoadEventNexus` to use `IndexInfo`, making it ready for introducing MPI support.

**The following subtleties are important and should be addressed/confirmed by the reviewer**

*Fixed*
1. `EventWorkspaceCollection::padSpectra` used to set `ws->getSpectrum(i).setDetectorID(padding[i] - 1);` (only in the non-empty case), but as far as I understand the calling code passed detector IDs in `padding`, so the first ID is dropped, an unwanted one is added. This happens only for non-ISIS files. I think this is wrong and this PR should fix it.
2. The monitor range-check in `LoadEventNexus::createSpectraList` was based on spectrum numbers and effectively prevent loading the first couple of spectra (in the case of provides spectrum range/list). I think this is wrong and this PR should fix it.
3. The range check in `LoadEventNexus::createSpectraList` for `m_specMax` looks broken in the original, should be fixed.
4. In `LoadEventNexus::createSpectraList`, if both min/max *and* list of spectra is provided, any sanity check for the *list*-part seems to be skipped in the original, and the corresponding events are not loaded. However, spectra for the list of spectra are nevertheless created in the non-ISIS case. Should be fixed in this PR, validating and loading spectra from the list as well.
5. If the detector IDs in the instrument have a gap and min/max for spectra selection is used, empty spectra are created for all non-existing detector IDs falling in the range between min and max. This happens only for non-ISIS files. I think this is wrong and this PR should fix it.
6. If a bank is selected and `SingleBankPixelsOnly` is checked, min/max/range are not validated and ignored when creating spectra (i.e., spectra for all pixels in the bank are created). But then min and max (but not range!) are still used to select which events are actually being loaded from the file. This PR will now emit a warning if min/max/range are set in this case and ignore the limits, i.e., events for all pixels in the bank are loaded.
7. For non-ISIS files, spectrum numbers usually start from 1 and are contiguous. Exception: If min/max/range property is set, the spectrum numbers are the detector IDs, i.e., loading a range of spectra will create a workspace whose spectrum numbers do not match the corresponding spectra of the same file loaded fully. Fixed to be consistented (had to modify a couple of unit tests in `LoadEventNexus`).

*Currently not fixed*

1. If a bank is selected for loading the ISIS spectra mapping from the file is ignored, probably leading to a complete mess being loaded since for such files `event_id` is the spectrum number, whereas in "bank"-mode `event_id` is interpreted as detector ID.

**To test:**

Really careful code review.

Fixes #20500.

**Release Notes** 
Some bugs are fixed, changing the behavior, but when I previously reported the issues they were closed, so I do not know if this is relevant?

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/DataHandling/src/LoadEventNexus.cpp
+++ b/Framework/DataHandling/src/LoadEventNexus.cpp
@@ -1101,16 +1101,14 @@ void LoadEventNexus::createSpectraMapping(
                                  " Try unchecking SingleBankPixelsOnly.");
       allDets.insert(allDets.end(), dets.begin(), dets.end());
     }
-    if (!allDets.empty()) {
-      m_ws->resizeTo(allDets.size());
-      // Make an event list for each.
-      for (size_t wi = 0; wi < allDets.size(); wi++) {
-        const detid_t detID = allDets[wi]->getID();
-        m_ws->setDetectorIdsForAllPeriods(wi, detID);
-      }
-      spectramap = true;
-      g_log.debug() << "Populated spectra map for select banks\n";
+    m_ws->resizeTo(allDets.size());
+    // Make an event list for each.
+    for (size_t wi = 0; wi < allDets.size(); wi++) {
+      const detid_t detID = allDets[wi]->getID();
+      m_ws->setDetectorIdsForAllPeriods(wi, detID);
     }
+    spectramap = true;
+    g_log.debug() << "Populated spectra map for select banks\n";
 
   } else {
     spectramap =
--- a/Framework/DataHandling/src/LoadEventNexus.cpp
+++ b/Framework/DataHandling/src/LoadEventNexus.cpp
@@ -1380,37 +1380,40 @@ bool LoadEventNexus::loadISISVMSSpectraMapping(const std::string &filename,
   }
   // Monitor filtering/selection
   const std::vector<detid_t> monitors = m_ws->getInstrument()->getMonitors();
-  const size_t nmons(monitors.size());
+  const auto &detectorInfo = m_ws->getSingleHeldWorkspace()->detectorInfo();
   if (monitorsOnly) {
     g_log.debug() << "Loading only monitor spectra from " << filename << "\n";
+    std::vector<Indexing::SpectrumNumber> spectrumNumbers;
+    std::vector<SpectrumDefinition> spectrumDefinitions;
     // Find the det_ids in the udet array.
-    m_ws->resizeTo(nmons);
-    for (size_t i = 0; i < nmons; ++i) {
+    for (size_t i = 0; i < monitors.size(); ++i) {
       // Find the index in the udet array
       const detid_t &id = monitors[i];
       std::vector<int32_t>::const_iterator it =
           std::find(udet.begin(), udet.end(), id);
       if (it != udet.end()) {
         const specnum_t &specNo = spec[it - udet.begin()];
-        m_ws->setSpectrumNumberForAllPeriods(i, specNo);
-        m_ws->setDetectorIdsForAllPeriods(i, id);
+        spectrumNumbers.emplace_back(specNo);
+        spectrumDefinitions.emplace_back(detectorInfo.indexOf(id));
       }
     }
+    Indexing::IndexInfo indexInfo(spectrumNumbers);
+    indexInfo.setSpectrumDefinitions(std::move(spectrumDefinitions));
+    m_ws->setIndexInfo(indexInfo);
   } else {
     g_log.debug() << "Loading only detector spectra from " << filename << "\n";
 
     SpectrumDetectorMapping mapping(spec, udet, monitors);
     auto uniqueSpectra = mapping.getSpectrumNumbers();
-    Indexing::IndexInfo indexInfo(std::vector<Indexing::SpectrumNumber>(
-        uniqueSpectra.begin(), uniqueSpectra.end()));
     std::vector<SpectrumDefinition> spectrumDefinitions;
-    const auto &detectorInfo = m_ws->getSingleHeldWorkspace()->detectorInfo();
     for (const auto spec : uniqueSpectra) {
       spectrumDefinitions.emplace_back();
       for (const auto detID : mapping.getDetectorIDsForSpectrumNo(spec)) {
         spectrumDefinitions.back().add(detectorInfo.indexOf(detID));
       }
     }
+    Indexing::IndexInfo indexInfo(std::vector<Indexing::SpectrumNumber>(
+        uniqueSpectra.begin(), uniqueSpectra.end()));
     indexInfo.setSpectrumDefinitions(std::move(spectrumDefinitions));
 
     m_ws->setIndexInfo(createSpectraList(indexInfo));
--- a/Framework/DataHandling/src/LoadEventNexus.cpp
+++ b/Framework/DataHandling/src/LoadEventNexus.cpp
@@ -1087,6 +1087,8 @@ void LoadEventNexus::createSpectraMapping(
       getProperty("SpectrumMax"), getProperty("SpectrumList"));
   if (!monitorsOnly && !bankNames.empty()) {
     m_ws->setIndexInfo(indexSetup.makeIndexInfo(bankNames));
+    g_log.notice()
+        << "Selecting banks will ignore spectrum min/max/list selection\n";
     g_log.debug() << "Populated spectra map for select banks\n";
   } else if (auto mapping = loadISISVMSSpectraMapping(m_top_entry_name)) {
     if (monitorsOnly) {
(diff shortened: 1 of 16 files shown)


Framework/DataHandling/src/LoadEventNexus.cpp
```diff
@@ -1081,56 +1086,52 @@ void LoadEventNexus::deleteBanks(EventWorkspaceCollection_sptr workspace,
...
   if (!monitorsOnly && !bankNames.empty()) {
-    std::vector<IDetector_const_sptr> allDets;
-
+    const auto &componentInfo = m_ws->getSingleHeldWorkspace()->componentInfo();
+    std::vector<SpectrumDefinition> spectrumDefinitions;
     for (const auto &bankName : bankNames) {
-      // Only build the map for the single bank
-      std::vector<IDetector_const_sptr> dets;
-      m_ws->getInstrument()->getDetectorsInBank(dets, bankName);
+      const auto &bank = m_ws->getInstrument()->getComponentByName(bankName);
```
OwenArnold: I guess this is a candidate for us adding some kind of component name to `ComponentInfo` as a future step.

Framework/DataHandling/src/LoadEventNexus.cpp
```diff
@@ -1081,56 +1086,52 @@ void LoadEventNexus::deleteBanks(EventWorkspaceCollection_sptr workspace,
...
+      std::vector<size_t> dets;
+      if (bank) {
+        const auto bankIndex = componentInfo.indexOf(bank->getComponentID());
+        dets = componentInfo.detectorsInSubtree(bankIndex);
```
OwenArnold: It's possible you could avoid the allocation here and use the `ComponentInfo` `Range`s (see my BoundingBox pr). That would mean moving the `SpectrumDefinition`s generation internally into `ComponentInfo`. Maybe though that's a dependency that we don't want to introduce?

Framework/DataHandling/src/LoadEventNexus.cpp
```diff
@@ -1695,78 +1686,48 @@ void LoadEventNexus::loadSampleDataISIScompatibility(
...
-    auto nmonitors = m_ws->getInstrument()->getMonitors().size();
-
-    for (size_t i = 0; i < nmonitors; ++i) {
-      if (std::find(m_specList.begin(), m_specList.end(), i + 1) !=
-          m_specList.end()) {
-        throw std::invalid_argument("Inconsistent range property: some of the "
-                                    "selected spectra correspond to monitors.");
-      }
+    const auto &detectorInfo = m_ws->getSingleHeldWorkspace()->detectorInfo();
+    for (const auto &specDef : *filteredIndexInfo.spectrumDefinitions()) {
```
OwenArnold: Would be helpful to have this check for monitors extracted as a separate method.

Framework/DataHandling/src/LoadEventNexus.cpp
```diff
@@ -1695,78 +1686,48 @@ void LoadEventNexus::loadSampleDataISIScompatibility(
...
-
+    // Check if SpectrumList was supplied (or filled via min/max above)
+    const auto indices =
+        indexInfo.makeIndexSet(std::vector<Indexing::SpectrumNumber>(
+            m_specList.begin(), m_specList.end()));
+    m_specMin =
+        static_cast<specnum_t>(indexInfo.spectrumNumber(*indices.begin()));
+    m_specMax =
+        static_cast<specnum_t>(indexInfo.spectrumNumber(*(indices.end() - 1)));
+    const auto filteredIndexInfo = Indexing::extract(indexInfo, indices);
```
OwenArnold: A comment here explaining what is being filtered out to accompany the code would be useful. The fact that you are creating a complete slice as a new `IndexInfo` might not be so clear from the code alone.

Framework/DataHandling/src/LoadEventNexus.cpp
```diff
@@ -9,6 +9,8 @@
 #include "MantidAPI/Sample.h"
```
OwenArnold: This sounds a little basic, but I would check to see if `LoadEventNexus` expects/gurantees the Instrument to be present (otherwise you'll have null/empty `ComponentInfo`, `DetectorInfo` etc). Sounds stupid, but I did come across this exact problem assuming that `LoadNexusProcessed` required the the instrument information to be present in the file.

OwenArnold (changes requested):
A nice start. As you seem to be working on this, I'll re-review again once changes have been made.

Framework/DataHandling/src/LoadEventNexus.cpp
SimonHeybrock: I don't think this is an issue. It used to create a vector of shared pointers to `Detector`, which was much much more work. So the new code should be at least as efficient.

Framework/DataHandling/test/EventWorkspaceCollectionTest.h
```diff
@@ -143,6 +145,28 @@ class EventWorkspaceCollectionTest : public CxxTest::TestSuite {
...
+    const auto ws = boost::dynamic_pointer_cast<WorkspaceGroup>(
+        collection.combinedWorkspace());
+    for (size_t i = 0; i < periods; ++i) {
+      auto eventWS =
+          boost::dynamic_pointer_cast<EventWorkspace>(ws->getItem(i));
+      TS_ASSERT_EQUALS(eventWS->getSpectrum(0).getSpectrumNo(), 3);
+      TS_ASSERT_EQUALS(eventWS->getSpectrum(1).getSpectrumNo(), 1);
+      TS_ASSERT_EQUALS(eventWS->getSpectrum(2).getSpectrumNo(), 2);
+      TS_ASSERT_EQUALS(eventWS->sample().getThickness(), thickness);
+    }
```
OwenArnold: Looks good

OwenArnold:
@SimonHeybrock  regarding your point 1. In the comment above `padSpectra` was only called in one place in the codebase and that is `LoadEventNexus` where it is passed a list of spectrum numbers to load. I don't agree with the - 1 offset, as that supposes that the spectrum numbers start from 1 or greater (usually true at ISIS but not elsewhere).

peterfpeterson:
Standard TOPAZ reduction still works. Event filtering with `SNSPowderReduction` on NOMAD still works.

Framework/DataHandling/test/LoadEventNexusIndexSetupTest.h
```diff
@@ -0,0 +1,370 @@
...
+    // Note that spectrum numbers are simply starting at 1, i.e., are not the
+    // same as without bank-filtering. This is consistent with the old behavior.
+    TS_ASSERT_EQUALS(indexInfo.spectrumNumber(0), SpectrumNumber(1));
+    TS_ASSERT_EQUALS(indexInfo.spectrumNumber(1), SpectrumNumber(2));
+    const auto specDefs = indexInfo.spectrumDefinitions();
+    TS_ASSERT_EQUALS(specDefs->at(0), SpectrumDefinition(1));
+    TS_ASSERT_EQUALS(specDefs->at(1), SpectrumDefinition(3));
+  }
+
+  void test_makeIndexInfo_from_bank_filter_ignored() {
```
OwenArnold: Could you group these legacy/characterisation tests separate from the tests that exercise desired behaviour. I think that would be very useful for future refactoring work. Naming the tests as such and putting them in a block together would work I think.

Framework/DataHandling/src/LoadEventNexusIndexSetup.cpp
```diff
@@ -0,0 +1,158 @@
...
+                               "' as a component assembly in the instrument "
+                               "tree; or it did not contain any detectors. Try "
+                               "unchecking SingleBankPixelsOnly.");
+  }
+  Indexing::IndexInfo indexInfo(spectrumDefinitions.size());
+  indexInfo.setSpectrumDefinitions(std::move(spectrumDefinitions));
+  return indexInfo;
+}
+
+IndexInfo LoadEventNexusIndexSetup::makeIndexInfo(
```
OwenArnold: I can see you have tests for this, but is it worth adding comments to the effect of 2,3,4 under "Currently not fixed" in the PR description. Again this is only to aid future refactoring/fixing.

OwenArnold:
Looks good. This is quite a tricky PR to validate. I think it would be good to talk through 1-5 in the "Fixed" list of the PR over skype.

Framework/DataHandling/test/LoadEventNexusIndexSetupTest.h
SimonHeybrock: Is there any apart from this that does not also check desired behavior?

Framework/DataHandling/src/LoadEventNexusIndexSetup.cpp
SimonHeybrock: Yes, will add doxygen once we know if this is actually the correct thing to do, or if we maybe want to fix even more things.

OwenArnold (approved):

OwenArnold:
:shipit:

PR #20513, opened 2017-09-19
Re #20489 More reliable argument handling in StartLiveData

### LINKED ISSUE #20489: StartLiveData sometimes ignores Listener Properties
PR #17926 makes it possible to pass Listener Properties as arguments when calling `StartLiveData` from Python.

However, under some circumstances, those arguments may be ignored and the properties are treated as if they have their default values. This only affect properties "pulled in" from the specific Listener being used, not properties that are part of `StartLiveData` itself.

This happens because Listener Properties are reset / reloaded whenever the `Instrument`, `Listener` or `Connection` property of `StartLiveData` changes. `Instrument` is treated as a special case, so that never causes this problem. But if `Listener` or `Connection` has been passed in as an argument, it's down to luck / Python `dict` order whether Listener Properties are reset before (OK) or after they are set (which destroys the value).

MikeHart85:
Python wrapper for StartLiveData now ensures that the `Connection` and `Listener` properties are set immediately after the `Instrument` property and before all other properties. 

This is to ensure that properties that are dependent on them, such as the dynamically loaded Listener Properties, are loaded correctly and not overwritten after being set.

**To test:**

Testing might be difficult, because the issue appears to be triggered by the order of the Python `dict` that contains the arguments passed to `StartLiveData`, which is undefined. But you can try the following, and hopefully it will fail the way it did on my machine.

Run the following in script interpreter (**must be connected to the STFC network**):
```python
StartLiveData(Instrument='ISIS_Kafka_Event',
                        Connection='kafka_event',
                        UpdateEvery=5,
                        PreserveEvents=True,
                        AccumulationMethod='Replace',
                        RunTransitionBehavior='Stop',
                        OutputWorkspace='ws',
                        InstrumentName='oldschema')
```

Before this change, this will usually fail with:

> Failed to find topic '_events' on broker

If you have the log set to debug, you will additionally see:

> Name: InstrumentName, Value: , Default?: Yes, Direction: Input

After the change, the listener will start and you will instead see, as expected: 

> Name: InstrumentName, Value: oldschema, Default?: No, Direction: Input

The started algorithm will run indefinitely, so you'll need to cancel it in the algorithm progress window.

Fixes #20489 

**Release Notes** 

Should this be in release notes? It's a bugfix for a small edge case issue that was found internally, not reported.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


martyngigg:
@MikeHart85 There seems to be a conflict with the current version of the release branch. Could you look into that and then I'll test it.

martyngigg (commented):
The code changes look fine to me.

@NickDraper Would you have time to verify the functionality?

NickDraper (approved):
Functionally works fine,

Issue #20344, opened 2017-08-28
Multi-process loading of event data from Nexus files


SimonHeybrock:
Loading large event files is slow. Current implementation uses threading. Does not help much since HDF5 is serial internally. Use multiple processes (MPI) to load.

SimonHeybrock:
This should have been closed by one of the linked PRs.

PR #20526, opened 2017-09-20
EventParser Implementation

### LINKED ISSUE #20430: Event Parser for distributed nexus loader
Part of #20344 

This issue deals with the creation of an object which parses event data, prepares for sending on different mpi ranks, re-distributes data accross ranks and fills some output object (EventLists?).

LamarMoore:
Implementation of the event parser for parallel event loading/processing. See #20344. 

**To test:**

Core review.

Fixes #20430. 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Parallel/inc/MantidParallel/Collectives.h
+++ b/Framework/Parallel/inc/MantidParallel/Collectives.h
@@ -3,6 +3,7 @@
 
 #include "MantidParallel/Communicator.h"
 #include "MantidParallel/DllConfig.h"
+#include "MantidParallel/Nonblocking.h"
 
 #ifdef MPI_EXPERIMENTAL
 #include <boost/mpi/collectives.hpp>
@@ -68,6 +69,19 @@ void gather(const Communicator &comm, const T &in_value, int root) {
         "Parallel::gather on root rank without output argument.");
   }
 }
+
+template <typename T>
+void all_to_all(const Communicator &comm, const std::vector<T> &in_values,
+                std::vector<T> &out_values) {
+  int tag{0};
+  out_values.resize(comm.size());
+  std::vector<Request> requests;
+  for (int rank = 0; rank < comm.size(); ++rank)
+    requests.emplace_back(comm.irecv(rank, tag, out_values[rank]));
+  for (int rank = 0; rank < comm.size(); ++rank)
+    comm.send(rank, tag, in_values[rank]);
+  wait_all(requests.begin(), requests.end());
+}
 }
 
 template <typename... T> void gather(const Communicator &comm, T &&... args) {
@@ -76,6 +90,15 @@ template <typename... T> void gather(const Communicator &comm, T &&... args) {
     return boost::mpi::gather(comm, std::forward<T>(args)...);
 #endif
   detail::gather(comm, std::forward<T>(args)...);
+}
+
+template <typename... T>
+void all_to_all(const Communicator &comm, T &&... args) {
+#ifdef MPI_EXPERIMENTAL
+  if (!comm.hasBackend())
+    return boost::mpi::all_to_all(comm, std::forward<T>(args)...);
+#endif
+  detail::all_to_all(comm, std::forward<T>(args)...);
 }
 
 } // namespace Parallel
--- a/Framework/Parallel/inc/MantidParallel/IO/EventParser.h
+++ b/Framework/Parallel/inc/MantidParallel/IO/EventParser.h
@@ -1,37 +1,72 @@
 #ifndef MANTID_PARALLEL_IO_EVENT_PARSER_H
 #define MANTID_PARALLEL_IO_EVENT_PARSER_H
 
+#include "MantidParallel/DllConfig.h"
+#include "MantidTypes/TofEvent.h"
 #include <cstdint>
 #include <vector>
 
 namespace Mantid {
-namespace Types {
-class TofEvent;
-}
-
 namespace Parallel {
 namespace IO {
+// TODO: replace with Chunker::LoadRange
+struct DLLExport LoadRange {
+  size_t bankIndex;
+  size_t eventOffset;
+  size_t eventCount;
+};
+
+struct DLLExport Event {
+  int32_t index; // global spectrum index
+  Mantid::Types::TofEvent tofEvent;
+};
 
 template <class IndexType, class TimeZeroType, class TimeOffsetType>
-class EventParser { // TODO
+class MANTID_PARALLEL_DLL EventParser { // TODO
 public:
   EventParser(std::vector<std::vector<int>> rankGroups,
-              std::vector<int32_t> &&bankOffsets,
+              const std::vector<int32_t> &bankOffsets,
               std::vector<std::vector<Mantid::Types::TofEvent> *> &eventLists);
 
-  void setPulseInformation(std::vector<IndexType> event_index,
-                           std::vector<TimeZeroType> event_time_zero);
+  void
+  setPulseInformation(const std::vector<IndexType> &event_index,
+                      const std::vector<TimeZeroType> &event_time_zero) const;
 
-  void startParsing(int32_t *event_id_start, TimeOffsetType *event_time_offset_start,
-                    size_t count);
+  void startParsing(int32_t *event_id_start,
+                    TimeOffsetType *event_time_offset_start,
+                    const LoadRange &range);
+
+  void extractEventsForRanks(const std::vector<int32_t> &globalSpectrumIndex,
+                             const TimeOffsetType *eventTimeOffset,
+                             size_t offset);
+
+  void eventIdToGlobalSpectrumIndex(int32_t *event_id_start, size_t count,
+                                    size_t bankIndex) const;
+
+  void populateEventList(
+      std::vector<std::vector<Mantid::Types::TofEvent> *> &eventList,
+      const std::vector<Event> &events);
+
+  const std::vector<int32_t> &globalSpectrumIndex() const {
+    return m_globalSpectrumIndex;
+  }
+
+  const std::vector<std::vector<Event>> &rankData() const {
+    return m_allRankData;
+  }
+
   void wait();
   void finalize();
 
 private:
+  std::vector<std::vector<int>> m_rankGroups;
   std::vector<int32_t> m_bankOffsets;
   std::vector<std::vector<Mantid::Types::TofEvent> *> &m_eventLists;
-  std::vector<IndexType> m_event_index;
-  std::vector<TimeZeroType> m_event_time_zero;
+  mutable std::vector<IndexType> m_eventIndex;
+  mutable std::vector<TimeZeroType> m_eventTimeZero;
+  mutable std::vector<int32_t> m_globalSpectrumIndex;
+  std::size_t m_posInEventIndex;
+  std::vector<std::vector<Event>> m_allRankData;
 };
 
 } // namespace IO
--- a/Framework/Parallel/inc/MantidParallel/IO/EventParser.h
+++ b/Framework/Parallel/inc/MantidParallel/IO/EventParser.h
@@ -1,0 +1,39 @@
+#ifndef MANTID_PARALLEL_IO_EVENT_PARSER_H
+#define MANTID_PARALLEL_IO_EVENT_PARSER_H
+
+#include <cstdint>
+#include <vector>
+
+namespace Mantid {
+namespace Types {
+class TofEvent;
+}
+
+namespace Parallel {
+namespace IO {
+
+class EventParser { // TODO
+public:
+  EventParser(std::vector<std::vector<int>> rankGroups,
+              std::vector<int32_t> &&bankOffsets,
+              std::vector<std::vector<Mantid::Types::TofEvent> *> &eventLists);
+
+  void setPulseInformation(std::vector<int32_t> event_index,
+                           std::vector<int64_t> event_time_zero);
+
+  void startParsing(int32_t *event_id_start, int32_t *event_time_offset_start,
+                    size_t count);
+  void wait();
+  void finalize();
+
+private:
+  std::vector<int32_t> m_bankOffsets;
+  std::vector<std::vector<Mantid::Types::TofEvent> *> &m_eventLists;
+  std::vector<int32_t> m_event_index;
+  std::vector<int64_t> m_event_time_zero;
+};
+
+} // namespace IO
+} // namespace Parallel
+} // namespace Mantid
+#endif // MANTID_PARALLEL_IO_EVENT_PARSER_H
--- a/Framework/Parallel/inc/MantidParallel/IO/EventParser.h
+++ b/Framework/Parallel/inc/MantidParallel/IO/EventParser.h
@@ -18,16 +18,43 @@ struct DLLExport LoadRange {
   size_t eventCount;
 };
 
-struct DLLExport Event {
-  int32_t index; // global spectrum index
+struct DLLExport EventListEntry {
+  int32_t globalIndex; // global spectrum index
   Types::Event::TofEvent tofEvent;
 };
 
+/** Distrubuted (MPI) parsing of Nexus events from a data stream. Data is
+distributed accross MPI ranks for writing to a event lists.
+
+@author Lamar Moore
+@date 2017
+
+Copyright &copy; 2017 ISIS Rutherford Appleton Laboratory, NScD Oak Ridge
+National Laboratory & European Spallation Source
+
+This file is part of Mantid.
+
+Mantid is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation; either version 3 of the License, or
+(at your option) any later version.
+
+Mantid is distributed in the hope that it will be useful,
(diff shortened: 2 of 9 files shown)


Framework/Parallel/src/IO/EventParser.cpp
```diff
@@ -0,0 +1,198 @@
...
+ *@param event_index
+ *@param event_time_zero
+ *
+ */
+template <class IndexType, class TimeZeroType, class TimeOffsetType>
+void EventParser<IndexType, TimeZeroType, TimeOffsetType>::setPulseInformation(
+    const std::vector<IndexType> &event_index,
+    const std::vector<TimeZeroType> &event_time_zero) const {
+  m_eventIndex = event_index;
+  m_eventTimeZero = event_time_zero;
```
SimonHeybrock: Does it make sense to pass by value and move?

Framework/Parallel/inc/MantidParallel/IO/EventParser.h
```diff
@@ -0,0 +1,75 @@
...
+  void wait();
+  void finalize();
+
+private:
+  std::vector<std::vector<int>> m_rankGroups;
+  std::vector<int32_t> m_bankOffsets;
+  std::vector<std::vector<Mantid::Types::TofEvent> *> &m_eventLists;
+  mutable std::vector<IndexType> m_eventIndex;
+  mutable std::vector<TimeZeroType> m_eventTimeZero;
+  mutable std::vector<int32_t> m_globalSpectrumIndex;
```
SimonHeybrock: Why do these need to be mutable?

Framework/Parallel/inc/MantidParallel/IO/EventParser.h
```diff
@@ -0,0 +1,75 @@
...
+template <class IndexType, class TimeZeroType, class TimeOffsetType>
+class MANTID_PARALLEL_DLL EventParser { // TODO
+public:
+  EventParser(std::vector<std::vector<int>> rankGroups,
+              const std::vector<int32_t> &bankOffsets,
+              std::vector<std::vector<Mantid::Types::TofEvent> *> &eventLists);
+
+  void
+  setPulseInformation(const std::vector<IndexType> &event_index,
+                      const std::vector<TimeZeroType> &event_time_zero) const;
```
SimonHeybrock: Why const?

Framework/Parallel/src/IO/EventParser.cpp
```diff
@@ -0,0 +1,198 @@
...
+ * @param bankIndex Index into the list of bank offsets.
+ */
+template <class IndexType, class TimeZeroType, class TimeOffsetType>
+void EventParser<IndexType, TimeZeroType, TimeOffsetType>::
+    eventIdToGlobalSpectrumIndex(int32_t *event_id_start, size_t count,
+                                 size_t bankIndex) const {
+  m_globalSpectrumIndex.resize(count);
+
+  for (size_t i = 0; i < count; i++)
+    m_globalSpectrumIndex[i] = event_id_start[i] - m_bankOffsets[bankIndex];
```
SimonHeybrock: Could do this in-place in `event_id` to save memory bandwidth?

Framework/Parallel/src/IO/EventParser.cpp
```diff
@@ -0,0 +1,198 @@
...
+        offset;
+
+    for (size_t i = start; i < end; ++i) {
+      // TODO: select appropriate ranks round robin?
+      // int rank = global_spectrum_index[i] % nrank;
+      // rank_data[rank].push_back(...)
+      int rank = 0;
+      m_allRankData[rank].push_back(
+          Event{globalSpectrumIndex[i],
+                TofEvent{static_cast<double>(eventTimeOffset[i]),
```
SimonHeybrock: I think I would do the translation into `TofEvent` on the receiver side. `TofEvent` has 16 bytes (aligned to 8 byte), so that brings `Event` to 24 bytes. If you keep the original types it might be only 16 byte (if `event_time_offset` is 4 byte). Also it may be the case that we use not all ranks for loading, so as much work as possible should be done on the receiver side.

SimonHeybrock (commented):
Just a quick review pass before lunch, have not made it to the tests yet. Structure looks mostly good, I think:

Framework/Parallel/src/IO/EventParser.cpp
```diff
@@ -0,0 +1,199 @@
...
+EventParser<IndexType, TimeZeroType, TimeOffsetType>::EventParser(
+    std::vector<std::vector<int>> rankGroups,
+    const std::vector<int32_t> &bankOffsets,
+    std::vector<std::vector<TofEvent> *> &eventLists)
+    : m_rankGroups(rankGroups), m_bankOffsets(bankOffsets),
+      m_eventLists(eventLists), m_posInEventIndex(0) {
+  if (m_bankOffsets.empty())
+    throw std::invalid_argument("The bankOffsets vector cannot be empty.");
+  if (m_eventLists.empty())
+    throw std::invalid_argument("The evenLists vector cannot be empty.");
```
SimonHeybrock: I am not sure such a check makes sense. You will be receiving input data with arbitrary `event_id` (or `bankIndex`), and unless you check every single one you have no sensible way of checking the length of `eventLists` (or `bankOffsets`). Having these to checks here gives a false sense of security, and having length 0 is no worse than having the length off by 1.

Maybe it makes more sense to put verbose warning comments at the appropriate points, saying that there is no range check, i.e., you just have to rely on the client using this class.

LamarMoore:
@SimonHeybrock @OwenArnold This represents a first pass of an MPI implementation of event parsing. In this implementation, data is redistributed and then written to the EventLists. I will interleave this operation in the next iteration if necessary, as such, I have not had any use for the `RankGroups`.

Framework/Parallel/src/IO/EventParser.cpp
```diff
@@ -0,0 +1,284 @@
...
+                         m_eventTimeZero[pulse]}});
+    }
+  }
+}
+
+/** Uses MPI calls to redistribute chunks which must be processed on certain
+ *ranks.
+ *@param result output data which must be processed on current rank. (May
+ *be updated by other ranks)
+ *@param data Data on this rank which belongs to several other ranks.
```
OwenArnold: If this `data` param represents data on other ranks, presumably we need to indicate that it cannot be modified on the current rank. i.e. should be ```const std::vector<std::vector<Event>>&```?

Framework/Parallel/test/EventParserTest.h
```diff
@@ -0,0 +1,464 @@
...
+  LoadRange generateBasicRange(size_t bank) {
+    LoadRange range;
+    range.eventOffset = 0;
+    range.eventCount = m_event_ids[bank].size();
+    range.bankIndex = bank;
+
+    return range;
+  }
+
+  // TODO: Generate load ranges for different MPI ranks
```
OwenArnold: Commented out block?

Framework/Parallel/test/EventParserTest.h
```diff
@@ -0,0 +1,464 @@
...
+#include <numeric>
+
+using namespace Mantid;
+using namespace Parallel::IO;
+using Mantid::Types::Core::DateAndTime;
+using Mantid::Types::Event::TofEvent;
+
+namespace detail {
+template <typename IndexType, typename TimeZeroType, typename TimeOffsetType>
+class FakeParserDataGenerator {
```
OwenArnold: Naming is good. But would be nice to have brief comments somewhere on how this is generating the events for use by the parser. Looks like there is quite a bit of logic in this type.

Framework/Parallel/test/EventParserTest.h
```diff
@@ -0,0 +1,464 @@
...
+  }
+
+  // TODO: Generate load ranges for different MPI ranks
+  // std::vector<LoadRange> generateMPIRanges(size_t bank, int rank,
+  //                                         int numRanks) {
+  //  return std::vector<LoadRange>{};
+  //}
+
+  static EventParser<IndexType, TimeZeroType, TimeOffsetType>
+  createSimpleParser(std::vector<std::vector<TofEvent> *> &eventLists,
```
OwenArnold: Don't understand why this creational method is on the generator. Doesn't make 100% sense to me.

Framework/Parallel/test/EventParserTest.h
```diff
@@ -0,0 +1,464 @@
...
+                            rankData[0].cbegin() + end,
+                            [pulses, pulse](const Event &e) {
+                              return e.tofEvent.pulseTime() ==
+                                     static_cast<int64_t>(pulses[pulse]);
+                            }));
+    }
+  }
+};
+
+// TODO Add performance Tests
```
OwenArnold: Are you going to add performance tests as part of this PR?

OwenArnold (changes requested):
Starting to look good. I think we should have a conversation about some aspects of this PR.

Framework/Parallel/test/EventParserTest.h
LamarMoore: Yes, currently writing those.

Framework/Parallel/test/EventParserTest.h
LamarMoore: That should have been removed.

Framework/Parallel/inc/MantidParallel/IO/EventParser.h
```diff
@@ -0,0 +1,80 @@
...
+#include "MantidParallel/DllConfig.h"
+#include "MantidTypes/Event/TofEvent.h"
+#include <cstdint>
+#include <future>
+#include <vector>
+
+namespace Mantid {
+namespace Parallel {
+namespace IO {
+// TODO: replace with Chunker::LoadRange
```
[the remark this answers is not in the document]
LamarMoore: Yes. I need to rebase after another of Simon's PRs.

Framework/Parallel/inc/MantidParallel/IO/EventParser.h
```diff
@@ -0,0 +1,80 @@
...
+  size_t eventCount;
+};
+
+struct DLLExport Event {
+  int32_t index; // global spectrum index
+  Types::Event::TofEvent tofEvent;
+};
+
+template <class IndexType, class TimeZeroType, class TimeOffsetType>
+class MANTID_PARALLEL_DLL EventParser { // TODO
```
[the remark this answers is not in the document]
LamarMoore: 👍

Framework/Parallel/src/IO/EventParser.cpp
```diff
@@ -0,0 +1,228 @@
...
+template <class IndexType, class TimeZeroType, class TimeOffsetType>
+void EventParser<IndexType, TimeZeroType, TimeOffsetType>::setPulseInformation(
+    std::vector<IndexType> event_index,
+    std::vector<TimeZeroType> event_time_zero) {
+  m_eventIndex = std::move(event_index);
+  m_eventTimeZero = std::move(event_time_zero);
+  m_posInEventIndex = 0;
+}
+
+/** Used the detectorIDs supplied to calculate the corresponding global spectrum
```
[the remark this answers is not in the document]
LamarMoore: event_ids are detector ids. I will update the comment

Framework/Parallel/src/IO/EventParser.cpp
```diff
@@ -0,0 +1,284 @@
...
+    std::vector<TimeZeroType> event_time_zero) {
+  m_eventIndex = std::move(event_index);
+  m_eventTimeZero = std::move(event_time_zero);
+  m_posInEventIndex = 0;
+}
+
+/** Uses the detectorIDs supplied to calculate the corresponding global spectrum
+ * numbers using the bankOffsets stored at object creation. NB event_id_start is
+ * transformed to contain the spectrum indices.
+ * @param event_id_start Starting position of chunk of data containing detector
```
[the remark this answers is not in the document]
LamarMoore: The comment below shows **NB event_id_start is transformed to contain the spectrum indices** as well as the documentation for `event_id`. I will try to use better wording.

Framework/Parallel/test/EventParserTest.h
LamarMoore: I did this so I could access these methods from the performance test. My initial idea was to create a parser based on the data generated. This will be modified shortly either way.

Framework/Parallel/inc/MantidParallel/IO/EventParser.h
```diff
@@ -18,16 +18,43 @@ struct DLLExport LoadRange {
   size_t eventCount;
 };
 
-struct DLLExport Event {
-  int32_t index; // global spectrum index
+struct DLLExport EventListEntry {
+  int32_t globalIndex; // global spectrum index
```
OwenArnold: :thumbsup:

Framework/Parallel/src/IO/EventParser.cpp
```diff
@@ -0,0 +1,285 @@
...
+}
+
+template <class IndexType, class TimeZeroType, class TimeOffsetType>
+void EventParser<IndexType, TimeZeroType, TimeOffsetType>::wait() const {
+  if (m_future.valid())
+    m_future.wait();
+}
+
+template <class IndexType, class TimeZeroType, class TimeOffsetType>
+void EventParser<IndexType, TimeZeroType, TimeOffsetType>::finalize() const {
```
OwenArnold: Not sure we need `finalize` at present.

Framework/Parallel/src/IO/EventParser.cpp
```diff
@@ -0,0 +1,285 @@
...
+ * separates then according to MPI ranks and then appends them to the workspace
+ * event list.
+ * @param event_id_start Buffer containing detector IDs.
+ * @param event_time_offset_start Buffer containing TOD.
+ * @param range contains information on the detector bank which corresponds to
+ * the data in the buffers, the file index offset where data starts and the
+ * number of elements in the data array.
+ */
+template <class IndexType, class TimeZeroType, class TimeOffsetType>
+void EventParser<IndexType, TimeZeroType, TimeOffsetType>::startParsing(
```
OwenArnold: I think it would be useful to know that what is started is done in an asynchronous manner, as you need to know this for the other methods to make sense. Can you make this `startAsync`  or `executeAsync`?

OwenArnold (changes requested):

Framework/Parallel/test/EventParserTest.h
```diff
@@ -0,0 +1,501 @@
...
+        auto eventSize = getRandEventSize(1, maxEventsPerPixel / numPulses);
+        pulseEventSize += eventSize;
+        auto *list = m_event_lists[pixel];
+        auto prev_end = list->size();
+        std::generate_n(std::back_inserter(*list), eventSize, [this, pulse]() {
+          return TofEvent(getRandomTimeOffset(100000),
+                          m_event_time_zero[pulse]);
+        });
+        std::fill_n(std::back_inserter(m_event_ids[bank]), eventSize,
+                    static_cast<IndexType>(m_bank_offsets[bank] + pixel));
```
SimonHeybrock: If I understand correctly, this is creating events in `event_id` that are sorted by `event_id` (within the pulse and bank)? Sounds like a special case, but probably unlikely to cause issues (maybe distoring performance tests?).

SimonHeybrock (commented):
Some comments:

Framework/Parallel/inc/MantidParallel/Collectives.h
```diff
@@ -68,6 +69,19 @@ void gather(const Communicator &comm, const T &in_value, int root) {
         "Parallel::gather on root rank without output argument.");
   }
 }
+
+template <typename T>
+void all_to_all(const Communicator &comm, const std::vector<T> &in_values,
+                std::vector<T> &out_values) {
+  int tag{0};
```
OwenArnold: Is there any significance of this constant? I think boost mpi has an `any_tag`

OwenArnold (commented):
Looks good. I see you are still working on this though.

Framework/Parallel/inc/MantidParallel/Collectives.h
SimonHeybrock: It does (http://www.boost.org/doc/libs/1_54_0/doc/html/boost/mpi/any_tag.html), but "This constant may be used for the tag parameter of receive operations to indicate that a send with any tag will be matched by the receive." is not what we want, as far as I understand.

SimonHeybrock: Also this is code using the threading backend, not `boost::mpi`.

SimonHeybrock:
I have addressed some main points I commented on earlier. There are some remaining comments I had regarding testability and tests, but they can be addressed in a follow-up PR, such that we can move forward.

@OwenArnold Do you have more comments at this point or can we merge this into the feature branch?

OwenArnold (approved):

OwenArnold:
Do not merge until reviewed by @AntonPiccardoSelg

AntonPiccardoSelg (approved):

OwenArnold:
As this is being merged into a long-running feature branch, and has had several reviewers, we are bypassing the separate gatekeeper. Gatekeeper will need to verify again before we can merge to master.

PR #20481, opened 2017-09-15
Moved TofEvents from DataObjects to Types::Events

LamarMoore:
`TofEvents` (and `DateAndTime`) have been moved out of `Mantid::Kernel` into a lower level library `Mantid::Types` in preparation for use with the parallel nexus loader #20344.

**To test:**

Code review and ensure all unit/system tests pass.

Pay attention to:
 - `Framework/Kernel/DateAndTimeHelpers.h/cpp`
 -  `Framework/Types/Core/DateAndTime.h/cpp`
 - `Framework/Types/Events/TofEvent.h/cpp`
 - `buildconfig/CMake/GenerateMantidExportHeader.cmake`
 - `Framework/Types/CMakeList.txt`
Fixes #20436 . 

**Release Notes** 
*Internal Change. Does not need to be in the release notes.*

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)?
- [ ] Are the unit tests small and test the class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] Do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant (user and developer) documentation been added/updated?

Does everything look good? Mark the review as **Approve**. A member of `@mantidproject/gatekeepers` will take care of it.

PR #20597, opened 2017-09-22
Refactored DateAndTime to remove POCO and Logging Dependencies

### LINKED ISSUE #20523: Remove logging and POCO dependencies from DateAndTime
This issue is related to removing logger and poco dependencies from `Mantid::Kernel::DateAndTime` in preparation for moving the module out of Kernel into a lower level library for use elsewhere.

LamarMoore:
Logging and poco dependencies were removed from `DateAndTime` in preparation for moving this library to the lower level `Mantid::Types` library.

**To test:**

Code review and ensure unit tests pass. Would be good to have input on any edge cases I may have missed regarding file loading and date/time formats.

Fixes #20523. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Kernel/inc/MantidKernel/DateAndTime.h
+++ b/Framework/Kernel/inc/MantidKernel/DateAndTime.h
@@ -5,8 +5,8 @@
 #ifndef Q_MOC_RUN
 #include <boost/date_time/posix_time/posix_time.hpp>
 #endif
+#include <cstdint>
 #include <iosfwd>
-#include <stdint.h>
 #include <string>
 #include <time.h>
 #include <vector>
@@ -122,7 +122,7 @@ public:
   static const boost::posix_time::ptime GPS_EPOCH;
 
   /// Const of one second time duration
-  static const time_duration oneSecond;
+  static const time_duration ONE_SECOND;
 
   static time_t utc_mktime(struct tm *utctime);
 
--- a/Framework/Kernel/inc/MantidKernel/DateAndTime.h
+++ b/Framework/Kernel/inc/MantidKernel/DateAndTime.h
@@ -41,7 +41,7 @@ public:
   DateAndTime(const double seconds, const double nanoseconds);
   DateAndTime(const int32_t seconds, const int32_t nanoseconds);
   DateAndTime(const int64_t seconds, const int64_t nanoseconds);
-  DateAndTime(const std::string &ISO8601_string, bool displayLogs = true);
+  DateAndTime(const std::string &ISO8601_string);
   DateAndTime(const boost::posix_time::ptime _ptime);
 
   void set_from_ptime(boost::posix_time::ptime _ptime);
@@ -53,7 +53,7 @@ public:
   std::time_t to_localtime_t() const;
   std::tm to_tm() const;
 
-  void setFromISO8601(const std::string &str, bool displayLogs = true);
+  void setFromISO8601(const std::string &str);
   std::string toSimpleString() const;
   std::string
   toFormattedString(const std::string format = "%Y-%b-%d %H:%M:%S") const;
@@ -131,14 +131,6 @@ private:
   int64_t _nanoseconds;
 };
 #pragma pack(pop)
-
-/// The epoch for GPS times.
-const boost::posix_time::ptime
-    DateAndTime::GPS_EPOCH(boost::gregorian::date(1990, 1, 1));
-
-/// Const of one second time duration
-const time_duration DateAndTime::oneSecond =
-    boost::posix_time::time_duration(0, 0, 1, 0);
 
 /** Represents a time interval.
 
--- a/Framework/Kernel/inc/MantidKernel/DateAndTime.h
+++ b/Framework/Kernel/inc/MantidKernel/DateAndTime.h
@@ -8,7 +8,7 @@
 #include <cstdint>
 #include <iosfwd>
 #include <string>
-#include <time.h>
+#include <ctime>
 #include <vector>
 
 namespace Mantid {
@@ -41,7 +41,7 @@ public:
   DateAndTime(const double seconds, const double nanoseconds);
   DateAndTime(const int32_t seconds, const int32_t nanoseconds);
   DateAndTime(const int64_t seconds, const int64_t nanoseconds);
-  DateAndTime(const std::string &ISO8601_string);
+  DateAndTime(std::string ISO8601_string);
   DateAndTime(const boost::posix_time::ptime _ptime);
 
   void set_from_ptime(boost::posix_time::ptime _ptime);
@@ -53,7 +53,7 @@ public:
   std::time_t to_localtime_t() const;
   std::tm to_tm() const;
 
-  void setFromISO8601(const std::string &str);
+  void setFromISO8601(std::string str);
   std::string toSimpleString() const;
   std::string
   toFormattedString(const std::string format = "%Y-%b-%d %H:%M:%S") const;
--- a/Framework/Kernel/inc/MantidKernel/DateAndTime.h
+++ b/Framework/Kernel/inc/MantidKernel/DateAndTime.h
@@ -41,7 +41,7 @@ public:
   DateAndTime(const double seconds, const double nanoseconds);
   DateAndTime(const int32_t seconds, const int32_t nanoseconds);
   DateAndTime(const int64_t seconds, const int64_t nanoseconds);
-  DateAndTime(std::string ISO8601_string);
+  DateAndTime(const std::string &ISO8601_string);
   DateAndTime(const boost::posix_time::ptime _ptime);
 
   void set_from_ptime(boost::posix_time::ptime _ptime);
@@ -53,7 +53,7 @@ public:
   std::time_t to_localtime_t() const;
   std::tm to_tm() const;
 
-  void setFromISO8601(std::string str);
+  void setFromISO8601(const std::string &str);
   std::string toSimpleString() const;
   std::string
   toFormattedString(const std::string format = "%Y-%b-%d %H:%M:%S") const;
--- a/Framework/Kernel/inc/MantidKernel/DateAndTime.h
+++ b/Framework/Kernel/inc/MantidKernel/DateAndTime.h
@@ -5,10 +5,10 @@
 #ifndef Q_MOC_RUN
 #include <boost/date_time/posix_time/posix_time.hpp>
 #endif
-#include <time.h>
 #include <iosfwd>
 #include <stdint.h>
 #include <string>
+#include <time.h>
 #include <vector>
 
 namespace Mantid {
@@ -114,7 +114,17 @@ public:
   static void createVector(const DateAndTime start,
                            const std::vector<double> &seconds,
                            std::vector<DateAndTime> &out);
-  static bool stringIsISO8601(const std::string &str);
+
+  /// The difference in seconds between standard unix and gps epochs.
+  static const uint32_t EPOCH_DIFF = 631152000;
+
+  /// The epoch for GPS times.
+  static const boost::posix_time::ptime GPS_EPOCH;
+
+  /// Const of one second time duration
+  static const time_duration oneSecond;
+
+  static time_t utc_mktime(struct tm *utctime);
 
 private:
   /// A signed 64-bit int of the # of nanoseconds since Jan 1, 1990.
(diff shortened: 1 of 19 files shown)


SimonHeybrock:
@LamarMoore Could you avoid refactoring tons of Doxygen with every PR? It makes the diff hard to read. Maybe this is a setting issue in your IDE?

Could you undo the formatting changes here?

Framework/Kernel/src/DateAndTime.cpp
```diff
@@ -381,30 +375,9 @@ const DateAndTime &DateAndTime::defaultTime() {
  * @param str :: ISO8601 format string: "yyyy-mm-ddThh:mm:ss[Z+-]tz:tz"
  * @param displayLogs :: flag to indiciate if the logs should be displayed
  */
-void DateAndTime::setFromISO8601(const std::string &str, bool displayLogs) {
+void DateAndTime::setFromISO8601(const std::string &str) {
   // Make a copy
   std::string time = str;
-  // Some ARGUS files have an invalid date with a space instead of zero.
```
SimonHeybrock: Does is make sense to add a unit test, checking that this throws now?

LamarMoore:
@SimonHeybrock It is probably clang on my system which reformatting the doxygen. I will look into it.

SimonHeybrock (changes requested):
Looks good, but please undo the reformatting of all the doxygen to keep the diff reasonable and avoid merge conflicts.

Framework/Kernel/inc/MantidKernel/DateAndTime.h
```diff
@@ -114,30 +114,24 @@ class MANTID_KERNEL_DLL DateAndTime {
...
-  static bool stringIsISO8601(const std::string &str);
+
+  /// The difference in seconds between standard unix and gps epochs.
+  static const uint32_t EPOCH_DIFF = 631152000;
+
+  /// The epoch for GPS times.
+  static const boost::posix_time::ptime GPS_EPOCH;
+
+  /// Const of one second time duration
+  static const time_duration oneSecond;
```
martyngigg: Can we capitalize this for consistency?

Framework/Kernel/inc/MantidKernel/DateAndTime.h
```diff
@@ -114,30 +114,24 @@ class MANTID_KERNEL_DLL DateAndTime {
...
+
+  static time_t utc_mktime(struct tm *utctime);
```
martyngigg: Should this in fact be in the helpers namespace?

Framework/Kernel/src/DateAndTime.cpp
```diff
@@ -1,28 +1,28 @@
...
-#include <boost/lexical_cast.hpp>
-#include <boost/date_time/time.hpp>
 #include <boost/date_time/date.hpp>
+#include <boost/date_time/time.hpp>
+#include <boost/lexical_cast.hpp>
 
-#include <math.h>
 #include <exception>
 #include <limits>
+#include <math.h>
```
martyngigg: --> `<cmath>`

Framework/Kernel/inc/MantidKernel/DateAndTime.h
```diff
@@ -5,10 +5,10 @@
 #ifndef Q_MOC_RUN
 #include <boost/date_time/posix_time/posix_time.hpp>
 #endif
-#include <time.h>
 #include <iosfwd>
 #include <stdint.h>
```
martyngigg: Can we convert these to the C++ versions: `<cmath>` & `<cstdint>`?

Framework/Kernel/src/DateAndTime.cpp
martyngigg: Agreed.

Framework/Kernel/test/DateAndTimeTest.h
```diff
@@ -8,20 +8,20 @@
...
+#include <cxxtest/TestSuite.h>
+#include <sstream>
 #include <sys/stat.h>
 #include <time.h>
-#include <sstream>
 
 using namespace Mantid;
 using namespace Mantid::Kernel;
 
+using std::cout;
```
martyngigg: Do we need `cout` in the test?

martyngigg (changes requested):
Generally looks good. Just a few minor comments

Framework/Kernel/inc/MantidKernel/DateAndTime.h
LamarMoore: It is used internally (within `DateAndTime`) as well.

Framework/Kernel/inc/MantidKernel/DateAndTime.h
```diff
@@ -5,10 +5,10 @@
 #ifndef Q_MOC_RUN
 #include <boost/date_time/posix_time/posix_time.hpp>
 #endif
-#include <time.h>
+#include <cstdint>
 #include <iosfwd>
-#include <stdint.h>
 #include <string>
+#include <time.h>
```
martyngigg: Should this be `<ctime>`?

Framework/Kernel/inc/MantidKernel/DateAndTime.h
```diff
@@ -114,30 +114,24 @@ class MANTID_KERNEL_DLL DateAndTime {
   static void createVector(const DateAndTime start,
                            const std::vector<double> &seconds,
                            std::vector<DateAndTime> &out);
-  static bool stringIsISO8601(const std::string &str);
+
+  /// The difference in seconds between standard unix and gps epochs.
```
martyngigg: Is there a reason we have one defined in the header and the others in the `.cpp` file?

Framework/Kernel/src/DateAndTime.cpp
```diff
@@ -379,32 +371,10 @@ const DateAndTime &DateAndTime::defaultTime() {
 /** Sets the date and time using an ISO8601-formatted string
  *
  * @param str :: ISO8601 format string: "yyyy-mm-ddThh:mm:ss[Z+-]tz:tz"
- * @param displayLogs :: flag to indiciate if the logs should be displayed
  */
-void DateAndTime::setFromISO8601(const std::string &str, bool displayLogs) {
+void DateAndTime::setFromISO8601(const std::string &str) {
   // Make a copy
   std::string time = str;
```
martyngigg: We could change the signature of `setFromISO8601` to `setFromISO8601(std::string time)` and remove the manual copy? That way we let the compiler do a move if possible: https://stackoverflow.com/questions/46028709/why-is-passing-by-value-if-a-copy-is-needed-recommended-in-c11-if-a-const-re

Framework/Kernel/test/DateAndTimeTest.h
```diff
@@ -8,20 +8,19 @@
 #ifndef DATEANDTIMETEST_H_
 #define DATEANDTIMETEST_H_
 
-#include <cxxtest/TestSuite.h>
 #include "MantidKernel/DateAndTime.h"
 #include "MantidKernel/System.h"
+#include <cxxtest/TestSuite.h>
+#include <sstream>
 #include <sys/stat.h>
 #include <time.h>
```
martyngigg: `<ctime>`

Framework/Kernel/test/DateAndTimeTest.h
```diff
@@ -58,6 +57,10 @@ class DateAndTimeTest : public CxxTest::TestSuite {
     TS_ASSERT_EQUALS(d, expected);
   }
 
+  void test_constructor_fails_invalid_string() {
```
martyngigg: Could we add some tests with strings that look a bit like a date time, i.e the old strings that would have passed but will now fail?

martyngigg (changes requested):

Framework/Kernel/inc/MantidKernel/DateAndTime.h
LamarMoore: Yes, we cannot define non-integral types in the header. Builds fail.

Framework/Kernel/src/DateAndTime.cpp
LamarMoore: `str` is used within the exception handling (which is why it is preserved)

Framework/Kernel/inc/MantidKernel/DateAndTime.h
LamarMoore: @martyngigg I do not think we can initialise non-integral types in a header without some compilers complaining.

Framework/Kernel/src/DateAndTime.cpp
martyngigg: Oh yes, I missed that. In that case we should stick with the original `const &` version to avoid having 2 copies when an l-value input is used.

Either that or we remove the `str` from the exception and just throw the `std::invalid_argument` without including the string. The caller could then decide whether to include it in their error message if they want. It seems a shame that most of the time when the string is valid we are paying the price of a copy just for the error message.

Framework/Kernel/inc/MantidKernel/DateAndTime.h
martyngigg: Yes, static non-integral types have to be defined out of line. I was suggesting that we be consistent and put `EPOCH_DIFF` in `.cpp` too. I'm not overly concerned though.

martyngigg (approved):
I'm happy with this to be merged now.

SimonHeybrock:
@LamarMoore The diff is showing unrelated changes from our `LoadEventNexus` feature branch? If you remove those commits I could merge this.

SimonHeybrock:
@LamarMoore Removed unrelated commits as dicussed on slack, looks ok, and github show previous review comments. 👍

SimonHeybrock (approved):

Framework/Kernel/src/DateAndTimeHelpers.cpp
```diff
@@ -0,0 +1,86 @@
...
+bool stringIsISO8601(const std::string &date) {
+  Poco::DateTime dt;
+  int tz_diff;
+  return Poco::DateTimeParser::tryParse(Poco::DateTimeFormat::ISO8601_FORMAT,
+                                        date, dt, tz_diff);
+}
+
+/** Verifies whether or not a string conforms to ISO8601. Corrects the string
+ *if it does not and is of the ARGUS file date/time format
+ *
```
OwenArnold: For future readers, i think it would be useful to have something in the comments here, or for `isArgusDateTime` which gives and example of what that date-time string is like. and what the corresponding sanitized string ends up like. I wouldn't object if this was done as a separate PR, as I know that this PR is holding up others.

Framework/Kernel/test/DateAndTimeHelpersTest.h
```diff
@@ -0,0 +1,34 @@
...
+    TS_ASSERT(
+        DateAndTimeHelpers::stringIsISO8601("1990-01-02T03:04:02.000+05:30"));
+    TS_ASSERT(DateAndTimeHelpers::stringIsISO8601("1990-01-02 03:04"));
+    TS_ASSERT(DateAndTimeHelpers::stringIsISO8601("1990-01-02"));
+    TS_ASSERT(DateAndTimeHelpers::stringIsISO8601("1822-01-02"));
+
+    TS_ASSERT(!DateAndTimeHelpers::stringIsISO8601("January 1, 2345"));
+    TS_ASSERT(!DateAndTimeHelpers::stringIsISO8601("2010-31-56"));
+    TS_ASSERT(!DateAndTimeHelpers::stringIsISO8601("1990-01-02 45:92:22"));
+  }
```
OwenArnold: As a follow on from my comment above. is there any reason that the other methods in this namespace can't be tested? I think that would be useful.

Framework/Kernel/src/DateAndTimeHelpers.cpp
LamarMoore: This will be added in this PR as changes are made there for DateAndTime anyway https://github.com/mantidproject/mantid/pull/20481

Framework/Kernel/test/DateAndTimeHelpersTest.h
LamarMoore: Same as above. Will be added in this PR https://github.com/mantidproject/mantid/pull/20481

martyngigg:
@OwenArnold I'll merge this once the builds are happy

PR #20611, opened 2017-09-22
Remove hard-coded internal parameter in MonteCarloAbsorption

### LINKED ISSUE #20532: Expose internal point selection parameter in MonteCarloAbsorption to users
`MonteCarloAbsorption` must attempt to select a random point with the given sample/container/environment object. It does this by using the bounding box and then testing if this actually is within the object up until a maximum number of tries.

The maximum number of tries variable is hard-coded internally. In some cases it would be useful for users to be able to increase this value if the algorithm is having a hard time generating a point.

martyngigg:
**Description of work** 

Exposes the value for the number of tries that are used to select a random scatter point to users. It was previously hard coded and some shapes, most notably a thin annulus, would hit the maximum number of tries and stop.

The default value has also been increased so that hopefully most users will now not have to care about this.

**To test:**

A modified version of the usage example can be used to check this.

The following script fails with v<=3.10.1. It should pass with these changes.

```
# Default up axis is Y
SetSample(data, Geometry={'Shape': 'HollowCylinder', 'Height': 0.4, 'Center': [0.,0.,0.],
    'InnerRadius': 1., 'OuterRadius': 1.01},
     Material={'ChemicalFormula': '(Li7)2-C-H4-N-Cl6', 'SampleNumberDensity': 0.07})
# Simulating every data point can be slow. Use a smaller set and interpolate
abscor = MonteCarloAbsorption(data, NumberOfWavelengthPoints=20)
```

Try different shapes from the usage example and also try setting the new `MaxScatterPtAttempts` parameter back to 500 to see the old error message reappear.

Fixes #20532

[Release Notes](/mantidproject/mantid/pull/20611/commits/4ba197e6965d873e3dbf6004892afe3d0c56806b)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


mantid-roman (approved):
It works. Although setting `MaxScatterPtAttemts` to 500 didn't cause an error. Probably it's a good thing.

PR #20617, opened 2017-09-22
Ensure the correct generator is used for eigen & xmlrunner

martyngigg:
Description of work.

Passes the `-G ${CMAKE_GENERATOR}` argument to the `cmake` command that processes the external project configure commands for eigen and xmlrunner.

This is required on Windows where Visual Studio 2015 & 2017 are installed along with more cmake > 3.7. Without these changes the eigen & xmlrunner steps use the 2017 generator regardless of the generator chosen by the caller.

**To test:**

Code review.

*No issue*

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/buildconfig/CMake/Eigen.cmake
+++ b/buildconfig/CMake/Eigen.cmake
@@ -19,7 +19,7 @@ else()
   # for static libraries, object libraries, and executables without exports.
   cmake_policy(SET CMP0063 "OLD")
 
-  execute_process(COMMAND ${CMAKE_COMMAND} . WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/eigen-download )
+  execute_process(COMMAND ${CMAKE_COMMAND} -G "${CMAKE_GENERATOR}" . WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/eigen-download )
   execute_process(COMMAND ${CMAKE_COMMAND} --build . WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/eigen-download )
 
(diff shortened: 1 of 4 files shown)


ianbush:
Tested on Ubuntu 14.04 and it looks good to me. When this is merged you might want to warn people that clearing out the build directory is a good idea - I had an error when just checking out and building as `Eigen` was building with `Make` and Mantid with `ninja`.

martyngigg:
It looks like that was what the build error on Windows was too.

martyngigg:
We're going to have to be careful when merging this as I think we'll need to clean the build servers out too.

martyngigg:
I think I've figured out a check to put in the buildscript to avoid having to manually clean stuff.

martyngigg:
The builds seem to pass now so I think we're good. We'll still need to tell developers to clean their builds though. Maybe I should have put that check in the cmake directly?

ianbush:
I am not sure I understand the last commit - for me it seems like I still need clean the build directory out. Although you do say developers still need to clean their builds out?

peterfpeterson (approved):
The `cmake` changes are reasonable. I'm not sure that your "temporary" fix for build servers will help builds that happen *after* this one.

martyngigg:
@ianbush The `buildscript` changes check whether the generators match and clean the external projects if they don't. Developers will still need to do that manually.

The `buildscript` changes will get it through the incremental builds too but nothing after that.

ianbush (approved):
Thanks for the clarification, looks good to me.

PR #20628, opened 2017-09-25
Fix missing indirect reduction tabs

martyngigg:
Description of work.

The indirect reduction tabs were checking for the facility but were not checking whether the returned string was empty. This lead to missing tabs if the facility was not set correctly.

**To test:**

Before merging, check

* start MantidPlot
* in IPython type `config['default.facility] = ""`
* start `Interfaces->Indirect->Data Reduction`
* see that tabs are missing

after merging perform the same check and see that ISIS tabs are present.

*No issue*

**Release Notes** 

*Introduced in this cycle so it does not need to be in the release notes.*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


gvardany (approved):
Good catch, fixed :shipit:

PR #20630, opened 2017-09-25
Fix warnings.

OwenArnold:
Low risk change. Fixing warnings. Should not have been moving a temporary in this original code.

**To test:**

If the unit tests are passing, then this is working.


martyngigg (approved):
Good to go.

PR #20633, opened 2017-09-25
Fix use of std::move with const in ScanningWorkspaceBuilder

### LINKED ISSUE #20542: std::move not occurring in ScanningWorkspaceBuilder
### Expected behavior

Use `std::move` only when the object is actually moved.

### Actual behavior

There are at least 5 places in `ScanningWorkspaceBuilder` where the object is const and std::move is used, making the later have no effect.

Check the clang-tidy build
http://builds.mantidproject.org/view/Static%20Analysis/job/clang_tidy_framework/275/warnings14Result/category.749444337/file.454359356/

https://clang.llvm.org/extra/clang-tidy/checks/misc-move-const-arg.html

### Platforms affected

All

The ideal solution would have separate setters specialized for const l-value and r-value references. As long as you only call these setters once, it might be "good enough" to just remove const from the definition.

See slide 29 of https://github.com/CppCon/CppCon2014/blob/master/Presentations/Back%20to%20the%20Basics!%20Essentials%20of%20Modern%20C%2B%2B%20Style/Back%20to%20the%20Basics!%20Essentials%20of%20Modern%20C%2B%2B%20Style%20-%20Herb%20Sutter%20-%20CppCon%202014.pdf

Also, in the tests there are at least two cases where the moved-from object is used, which may cause undefined behavior.

https://github.com/mantidproject/mantid/blob/3f0a66642d163d8e32f8818b8636af5455f12ee6/Framework/DataObjects/test/ScanningWorkspaceBuilderTest.h#L396

https://github.com/mantidproject/mantid/blob/3f0a66642d163d8e32f8818b8636af5455f12ee6/Framework/DataObjects/test/ScanningWorkspaceBuilderTest.h#L409

This was introduced in PR #19565.

ianbush:
Fix `clang-tidy` issues - using `std::move` with `const` - see the issue for more details.

The values are only ever intended to be set once, so removing `const` should be sufficient.

**To test:**

No functional changes expected, though unexpected behaviour could be fixed. Check warnings are fixed.

Fixes #20542.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


SimonHeybrock (approved):
Code looks good.

martyngigg (commented):
Does the `ScanningWorkspaceBuilder` header not need to change here? I think that might be the problem with doxygen?

ianbush:
@martyngigg Yes, I think you are right. Strange this got triggered in this PR but not an earlier one, as the discrepancy was there before.

PR #20644, opened 2017-09-26
Indirect Data Analysis: Suppress warning on missing optional output 

### LINKED ISSUE #20610: Cannot save or plot result in Indirect Elwin interface
### Expected behavior
After loading and running a nexus file in the Elwin tab of the Indirect Data Analysis GUI, Plot Result would be expected to plot the output data, and Save Result would be expected to save it to an external file.

### Actual behavior
Neither operation is successful - in both cases we get a 'workspace not found' error.

Go to `Interfaces->Indirect->Indirect Data Analysis` and, inside the Elwin tab, load a focused IRIS file (such as [this](https://github.com/mantidproject/mantid/files/1324565/irs26176_graphite002_red.zip)). Click Run, and then try to do Plot Result and Save Result.

### Platforms affected
All

martyngigg:
Description of work.

**To test:**

See the test instructions in the original issue. The missing `_elt` workspace should now not produce a warning. To produce the `_elt` workspace first use the Indirect Data Reduction interface to reduce a newer file such as `IRIS00073585` and save the result. Load this into the data analysis interface, click "Normalise to Lowest Temperature" and click run. The `_elt` workspace should now appear and plotting should show all 4 plots.

Fixes #20610 

**Release Notes** 

*Feature introduced this cycle. Does not need to be in the release notes.*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/qt/scientific_interfaces/Indirect/IndirectTab.cpp
+++ b/qt/scientific_interfaces/Indirect/IndirectTab.cpp
@@ -620,20 +620,19 @@ QString IndirectTab::runPythonCode(QString code, bool no_output) {
  * @return False if no workspace found, True if workspace found
  */
 bool IndirectTab::checkADSForPlotSaveWorkspace(const std::string &workspaceName,
-                                               const bool &plotting) {
+                                               const bool plotting,
+                                               const bool warn) {
   const auto workspaceExists =
       AnalysisDataService::Instance().doesExist(workspaceName);
-  if (workspaceExists) {
-    return true;
-  } else {
+  if (warn && !workspaceExists) {
     const std::string plotSave = plotting ? "plotting" : "saving";
     const auto errorMessage = "Error while " + plotSave +
                               ":\nThe workspace \"" + workspaceName +
                               "\" could not be found.";
     const char *textMessage = errorMessage.c_str();
-    QMessageBox::warning(NULL, tr("Workspace not found"), tr(textMessage));
-    return false;
+    QMessageBox::warning(nullptr, tr("Indirect "), tr(textMessage));
   }
+  return workspaceExists;
 }
 
 } // namespace CustomInterfaces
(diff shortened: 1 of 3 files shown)


NickDraper:
@brandonhewer Could you fit in a review for this PR soon please.

qt/scientific_interfaces/Indirect/IndirectTab.cpp
```diff
@@ -620,20 +620,19 @@ QString IndirectTab::runPythonCode(QString code, bool no_output) {
...
-  } else {
+  if (warn && !workspaceExists) {
     const std::string plotSave = plotting ? "plotting" : "saving";
     const auto errorMessage = "Error while " + plotSave +
                               ":\nThe workspace \"" + workspaceName +
                               "\" could not be found.";
     const char *textMessage = errorMessage.c_str();
-    QMessageBox::warning(NULL, tr("Workspace not found"), tr(textMessage));
-    return false;
+    QMessageBox::warning(nullptr, tr("Indirect "), tr(textMessage));
```
samueljackson92: The bulk of this function is in a `if` statement. Consider flipping the condition and using an early exit?

samueljackson92 (changes requested):
Tested this but still got a warning when trying to save a result, but not when trying to plot

qt/scientific_interfaces/Indirect/IndirectTab.cpp
martyngigg: I'd need to go with:
```
  if(!warn) return workspaceExists;
  if(!workspaceExists) {
    const std::string plotSave = plotting ? "plotting" : "saving";
    const auto errorMessage = "Error while " + plotSave +
                              ":\nThe workspace \"" + workspaceName +
                              "\" could not be found.";
    const char *textMessage = errorMessage.c_str();
    QMessageBox::warning(nullptr, tr("Indirect "), tr(textMessage));
  }
  return workspaceExists;
```
which to me doesn't read as well. It's not performance critical code so the early exit isn't a huge advantage. What do you think?

samueljackson92: You're right. I don't think the condition can be easily flipped. Leave it as it is currently.

martyngigg:
@samueljackson92 Thanks for catching that. I've pushed a fix.

samueljackson92 (approved):

samueljackson92:
@martyngigg This now works for me. Wait for the builds to pass and this is good to go. 👍

PR #20668, opened 2017-09-27
SimpleAPI: an option to control storing on ADS

### LINKED ISSUE #20664: Option in simpleapi to not store the output on ADS for child algorithms
For historic reasons the child algorithms called through python simple api always store the output on ADS. Changing this would break the api, so another magic keyword `Child=True/False` should be implemented,  governing whether the output should be stored or not in ADS. This will help to get rid of multiline algorithm setups with `self.createChildAlgorithm` and replace them with function-like calls, coherent with other function-like calls inside the same algorithms.

gvardany:
This PR adds one more magic keyword to function-like simple api to get a handle over storing on ADS.
By default, the outputs will be stored on ADS as before, but now this can be disabled.
Disabling this will make it possible to eliminate `self.createChildAlgorithm` multiline setups by replacing them with neater function-like calls.
As a proof-of-concept, the necessary refactor has been done for `ExtractMonitors`.
The rest of the refactoring can be done in a separate ticket.

**To test:**
Code review.
Exercise with simpleapi calls with and without the new keyword, check they behave as supposed.
Pick an instance of `self.createChildAlgorithm` in python and modify it like it is done here for `ExtractMonitors`, see if it works.

Fixes #20664 . 

**Release Notes**

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


SimonHeybrock:
Question: Usually workspaces in Python are `std::weak_ptr` (ADS is the owner). As far as I can see this is still the case, and the owner is simply the output workspace property of the child algorithm? I am thus wondering if this causes issues with workspaces being deleted early, while you still have the weak pointer in Python?

gvardany:
That's a good point, I haven't thought about that. This might be a problem and has to be studied further. But the current test fails actually for a different reason, that is when an input workspace is not on ADS, the pointer is cleared [here](https://github.com/mantidproject/mantid/blob/master/Framework/API/inc/MantidAPI/WorkspaceProperty.tcc#L167)

gvardany:
Actually it was a confusion from the debugger. It turned out that simpleAPI was always setting the properties by name and not by object. Regarding your comment, it seems the property holds the [shared pointer](https://github.com/mantidproject/mantid/blob/master/Framework/API/inc/MantidAPI/WorkspaceProperty.h#L72), it just gets [cleared](https://github.com/mantidproject/mantid/blob/master/Framework/API/inc/MantidAPI/WorkspaceProperty.tcc#L346) right after storing on ADS (if so requested).

martyngigg:
@SimonHeybrock @gvardany Workspaces are only held as `weak_ptrs` when they come out of the ADS in python. Extracting from the algorithm should give you back the `shared_ptr` as expected.

martyngigg:
@gvardany There are some merge conflicts here now.

martyngigg:
@gvardany This looks really nice when using it from a Python algorithm. At the moment I cannot generally do `ws = CreateSampleWorkspace(StoreInADS=False)` from anywhere. Should we allow this too? This could be quite useful, particularly in vanilla python, where the ADS is not really useful. In the future maybe could then consider a flag to set the default storage behaviour.

If we don't pursue this then I think we should probably print a warning when this flag is passed but it doesn't apply.

gvardany:
Yes that would be indeed nice for the vanilla python case, and later one can have a global setting.

Framework/PythonInterface/mantid/simpleapi.py
```diff
@@ -878,19 +887,20 @@ def ignore_property(name_to_check, regex_to_ignore):
...
         prop = algm_obj.getProperty(name)
-        # Parent algorithms store their workspaces in the ADS
-        # Child algorithms should store their workspaces in the property
-        # but they don't at the moment while the issues with history recording Python Child Algs
-        # is resolved: ticket #5157
+
         if _is_workspace_property(prop):
             value_str = prop.valueAsStr
             try:
                 retvals[name] = _api.AnalysisDataService[value_str]
```
martyngigg: I think I've found a problem with trying the ADS first. It can be illustrated with the following script in the MantidPlot script window:

```python
# Run these 2 lines
test_ws = CreateSampleWorkspace(StoreInADS=True)
test_ws = CreateSampleWorkspace(StoreInADS=False)

# Now simulate deleting the workspace from the ADS using the delete button
DeleteWorkspace('test_ws')
Pause(0.5) # required to give the async delete time to happen.

# Now run this line. I would expect the variable to still be alive but it is not.
print (test_ws.getNumberHistograms())
```

I think the issue is that as `AnalysisDataService[value_str]` is first  `test_ws = CreateSampleWorkspace(StoreInADS=False)` then still goes through this line and a `weak_ptr` to the workspace is still returned the second time around (see https://github.com/mantidproject/mantid/blob/master/Framework/PythonInterface/inc/MantidPythonInterface/kernel/DataServiceExporter.h#L158)

Ideally I would like to remove the `weak_ptr` from the c++ side and move it to be controlled from Python but I think this is outside of the scope of this change.

martyngigg (changes requested):

Framework/PythonInterface/mantid/simpleapi.py
gvardany: Good catch, indeed, so do you think just swapping the order(value first, then check ADS) should be fine for the moment? Can you think of a counter-example to that? 
Btw, I've found out that the simple API specialisations for Load and RenameWorkspace do not respect the logging keyword, probably also the new keyword. It seems they set properties from a dictionary that is not the original kwargs, try:
```
ws=CreateSampleWorkspace()
RenameWorkspace('ws','ws2',EnableLogging=False)
```

martyngigg: @gvardany I *think* swapping to order should do the trick. In the case where `StoreInADS=True` the `getProperty.value` for a workspace should be `None` so falling back to the ADS should be the correct thing to do.

I hadn't noticed the `Load`/`RenameWorkspace` behaviour - I would say this is a bug as all of the functions are intended to have the same behaviour with regards to these things. I'd be happy if that were sorted out in a new issue if you think it's too much work for now.

gvardany: Actually I think it would be fit to fix them here as well, this shouldn't be too hard, I will have a look.

gvardany:
I have disabled the ADS keyword for `RenameWorkspaces`, since it does not make sense for anonymous ones. Regarding the logging for Load, it seems to set the logging of the child loader explicitly to [true](https://github.com/mantidproject/mantid/blob/master/Framework/DataHandling/inc/MantidDataHandling/Load.h#L88) (this argument is never actually passed, so the default is used). But this is the responsibility of Load, on python side the keyword seems to be delegated properly.

martyngigg (approved):
This looks good now. Happy to merge once builders are happy.

PR #20670, opened 2017-09-27
StartLiveData in Replace mode resets changes to Instrument 

### LINKED ISSUE #20517: StartLiveData in Replace mode resets changes to Instrument
When you run `StartLiveData` with `AccumulationMethod='Replace'`, the Instrument is also replaced every time a new chunk of data is processed.

This causes any changes made to the Instrument, such as `MoveInstrumentComponent`, to be reset back to IDF defaults.

This does not happen with `AccumulationMethod`s `Add` or `Append`.

### To reproduce

Execute `FakeISISEventDAE` with `NPeriods=5` and `NSpectra=10`.

In Script Interpreter, run:

```python
ConfigService.setFacility('TEST_LIVE')
StartLiveData(Instrument='ISIS_Event', OutputWorkspace='OutWS', UpdateEvery=2, AccumulationMethod='Replace', PeriodList=[1,3], SpectraList=[2,4,6])
```
Show Instrument, and move it a bit so you have an overview of all detectors.

In Script Interpreter, run:

```python
MoveInstrumentComponent('OutWS', 'det33', X=1.0)
```

You should see one of the detectors briefly disappear (because it moved), and then reappear with the next data update (because the instrument was reloaded).

If you repeat these steps with `AccumulationMethod='Add'` and `AccumulationMethod='Append'`, the moved detector will not reset, as expected.

OwenArnold:
Copy the instrument from old workspace to new, to prevent loosing geometry changes. Fixes #20517. See there for manual testing instructions.

**To test:**

* Code review
* Ensure new unit test passes
* Run manual tests according to issue


martyngigg (approved):
The code changes look sensible to me. I think users would expect that their changes to the instrument would be persisted.

I tested this in conjunction with the unscripted testing of live data and all seems well.

PR #20678, opened 2017-09-27
Fix clang-tidy modernize-header issues

peterfpeterson:
Also redo `Timer.cpp` and `XMLFormatter.h` to use c++11 constructs.

**To test:**

If the builds pass, it should be good.

*Does not need to be in the release notes* because it is just cleaning up implementation.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/DataHandling/src/LoadRaw/isisraw.h
+++ b/Framework/DataHandling/src/LoadRaw/isisraw.h
@@ -11,9 +11,6 @@ struct ISISCRPT_STRUCT;
 #include "item_struct.h"
 #include <cstdlib>
 #include <cstring>
-#ifdef _WIN32 /* _WIN32 */
-#include <ctime>
-#endif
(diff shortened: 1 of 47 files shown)


quantumsteve (changes requested):
The header changes look good. Nice to have these unwelcome headers updated.

I'm a bit concerned about the number of clang-format changes to `XmlFormatter.h`. Will this make it harder to change to the upstream codebase?

Please check with the build servers if `<ctime>` is still only required on windows.

peterfpeterson:
The build servers have spoken. `time.h` isn't needed anywhere, but I'll try another commit to remove `ifdef` around `<ctime>` in files. I'll push a separate commit to reduce the diff in `XmlFormatter.h`.

quantumsteve (approved):

PR #20682, opened 2017-09-27
Expose fitting penalties to python

### LINKED ISSUE #20634: Expose setConstraintPenalty to IFunction

VickieLynch:
Description of work.
Python fitting function needs higher penalty for violating constraints (Brendan's IC)

**To test:**
```python
# create a workspace with a gaussian peak sitting on top of a linear (here flat) background
ws = CreateSampleWorkspace(Function="User Defined", UserDefinedFunction="name=LinearBackground, \
   A0=0.3;name=Gaussian, PeakCentre=5, Height=10, Sigma=0.7", NumBanks=1, BankPixelWidth=1, XMin=0, XMax=10, BinWidth=0.1)

# Setup the data to fit:
workspaceIndex = 0  # the spectrum with which WorkspaceIndex to fit
startX = 1      # specify fitting region
endX = 9      #

# Setup the model, here a Gaussian, to fit to data
myFunc = Gaussian(Height=8, Sigma=1, PeakCentre=4)
# Centre is 5
myFunc.addConstraints("PeakCentre<3")
myFunc.setConstraintPenaltyFactor("PeakCentre",1.0e3)

# Do the fitting
fit_output = Fit(InputWorkspace='ws', WorkspaceIndex=0, StartX = startX, EndX=endX, Output='fit', Function=myFunc)
paramTable = fit_output.OutputParameters  # table containing the optimal fit parameters
fitWorkspace = fit_output.OutputWorkspace

print("The fit was: " + fit_output.OutputStatus)
print("chi-squared of fit is: %.2f" % fit_output.OutputChi2overDoF)
print("Fitted Height value is: %.2f" % paramTable.column(1)[0])
print("Fitted centre value is: %.2f" % paramTable.column(1)[1])
print("Fitted sigma value is: %.2f" % paramTable.column(1)[2])
# fitWorkspace contains the data, the calculated and the difference patterns
print("Number of spectra in fitWorkspace is: " +  str(fitWorkspace.getNumberHistograms()))
print("The 20th y-value of the calculated pattern: %.4f" % fitWorkspace.readY(1)[19])
```

Fixes #20634 

**Release Notes** 
Will put in 1.12 Framework release notes:
setConstraintPenaltyFactor is now exposed to Python to set the penalty for the constraints in the fitting function

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


jmborr:
@VickieLynch what object is `self` in the test?

jmborr:
@VickieLynch let me know when all the build servers pass and you're done. Also, add the `in progress` label.

VickieLynch:
@jmborr All builds have passed

jmborr:
@VickieLynch I ran the script without penalty by setting
`myFunc.setConstraintPenaltyFactor("PeakCentre",0.0)`
and still got `PeakCentre=3`. Do you know the reason?

VickieLynch:
@jmborr I agree the penalty=0 shows this is not working correctly, but I do not see the exception

jmborr:
@VickieLynch I ran the script in the scrip* window of mantidplot. The exception was shown in the **log** window, not in the script window

peterfpeterson (dismissed):

Framework/PythonInterface/mantid/api/src/Exports/IFunction.cpp
```diff
@@ -218,6 +218,10 @@ void export_IFunction() {
            (arg("self"), arg("name")),
            "Remove the constraint on the named parameter")
 
+      .def("setConstraintPenaltyFactor", &IFunction::setConstraintPenaltyFactor,
+           (arg("self"), arg("name"), arg("value")),
+           "Set the constraint penalty factor for named parameter")
+
```
jmborr: Can you travel back the test to find why the example is not working?

jmborr (changes requested):

NickDraper:
So after talking to @mantid-roman , this will currently not work.  The Function object in Python  (so the Gaussian in this case) is translated to a string and back as it passes into the Fit algorithm, and the methods for writing to the string and reading back do not do anything with the constraint penalty value.

The places you need to look are IFunction.cpp::writeToString and FunctionFactory.cpp

NickDraper (changes requested):
See my comment the penalty value is lost during string serialization.  
Also the unit tests only test setting the penalty to a high value (which is just like the default), you need a test where the penalty is reduced so a constraint can be shown to be ignored.

VickieLynch:
Explains why I could not see difference with test.  I will close since this did not work.

mantid-roman:
I can take a look at the issue.

jmborr:
@mantid-roman What if instead of [passing the string representation](https://github.com/mantidproject/mantid/blob/master/Framework/PythonInterface/mantid/simpleapi.py#L340), we pass the underlying unwrapped function object?
```
function = function.fun
```

mantid-roman:
@martyngigg We need your opinion on this. Passing the function means it will be modified by Fit.

martyngigg:
As discussed on Slack. Cloning could be expensive and cause a lot of unnecessary work so we'll accept the modification of the function by Fit and explain in the documentation how to preserve the old copy if required.

mantid-roman:
@VickieLynch How urgently do you need this fixed?

VickieLynch:
@mantid-roman We have a postdoc working on single crystal integration who requested this.  It is not urgent and he can probably find another way to get the fits he wants.

mantid-roman:
@VickieLynch Why does he feel a need to change the penalty factor? Will changing the default help? At the moment it is arbitrary set to 1000

VickieLynch:
@mantid-roman Postdoc, Brendan, says my branch with this change does work for him.  He sets to penalties to 1e10 here:
https://github.com/neutrons/fit_peak_intensities/blob/master/ICCFitTools.py  line 579 is where I set the penalty which calls line 25 in https://github.com/neutrons/fit_peak_intensities/blob/master/ICConvoluted.py

Issue #19553, opened 2017-05-15
Find and Manage elimination of duplicate detector ids


OwenArnold:
Once the issues in these IDFs have been fixed
~~#19549 BASIS~~
~~#19550 IN6~~
~~#19551 EXED~~
~~#19552 IMAT (x2)~~

After this we can remove the expensive try-catch-patch code in `InfoComponentVisitor` that handles the possibility of missing or duplicate detectorids, which I'm having to maintain as part of #19468. I am currently blocked by these IDFs being broken.

OwenArnold:
@DTasev @LamarMoore IMAT is the only instrument definition set for Mantid that remains broken. #19552

OwenArnold:
All IDFs fixed

PR #20686, opened 2017-09-28
Block Duplicate Detector IDs

### LINKED ISSUE #20460: Block Duplicate Detector IDs
Do not merge till after #20274 complete.

As a follow on from #19553 we should now look to apply proper restrictions around Mantid since duplicate detector ids are really not supported. See how `markAsDetector` silently swallows [errors](https://github.com/mantidproject/mantid/blob/master/Framework/Geometry/src/Instrument.cpp#L714)

This will also allow the removal of the try-catch block in `InstrumentVisitor::registerDetector`.

OwenArnold:
#fixes #20460

Mantid will now throw if you try to register multiple detectors with the same ID rather than silently carrying on. This will prevent fixes like [this](#19553) cohort being required in future.

Since Instrument 2.0 is being made via `InstrumentVisitor` from `Instrument` (Instrument 1.0), this change enables us to drop the try-catch hack we tried to previously mitigate this issue. This is good news, because the current method of recording the number of dropped detectors is flawed since detector indexes need to be a single contiguous range.

**Tester**

* A code review. 
* You could also try `LoadEmptyInstrument` with an old-style IDF containing duplicates.
* You could also then try loading a valid IDF using `LoadEmptyInstrument`, then save it via `SaveProcessNexus`. Then using HDFView, modify the xml section of the instrument to give the IDF duplicate detector IDS, then try loading via `LoadNexusProcessed`, you should get a message like that below.

screen shot 2017-10-04 at 08 23 56
* After performing the steps above, you should still have a workspace loaded (though it has NO instrument). You should be able to successfully attach an instrument by running `LoadInstrument` on the existing workspace


SimonHeybrock (dismissed):
Code changes look good, just one comment:

martyngigg (approved):
This looks good to me too and I think it makes much more sense that the old behaviour.

PR #20691, opened 2017-09-28
Errors for unsorted data in mean algorithm 

### LINKED ISSUE #20658: Mean with unsorted data
bug report from James Lord: 

> 
> We discovered a feature: If we have two point workspaces with the same X values in a different order, for example from successive scans up and down in field, Mean doesn't notice the mismatch of X and happily goes ahead and generates junk. It ought at least to warn that the values don't match. The fix in this case is to use SortXAxis.
> 
> These data were from ALC analysis but we'd have the same with a results table from a sequential fit in Muon Analysis and then Convert to Workspace.

AnthonyLim23:
The mean algorithm calculates the average y value for each x value in the workspaces. If the workspaces did not have the same order for the data then this would produce the wrong result. This is because it simply iterated over the x values.

**To test:**
With the script below you can test:
1. It works if the data is good
2. If throws an error if one workspace has more spectra than the other
3. It throws an error if the x values are different between the two spectra 
```

# Create two  2-spectrum workspaces with Y values 1->8 & 3->10
dataX = [0,1,2,3,4,
         0,1,2,3,4]
dataY = [1,2,3,4,
         5,6,7,8]
ws_1 = CreateWorkspace(dataX, dataY, NSpec=2)


dataY = [3,4,5,6,
         7,8,9,10]
dataX = [0,1,2,3,4,
         0,1,3,2,4]
ws_2 = CreateWorkspace(dataX, dataY, NSpec=2)

result = Mean("ws_1, ws_2") # note the comma-separate strings
print "Mean of y values in first spectrum:", result.readY(0)
print "Mean of y values in second spectrum:", result.readY(1)
```

Fixes #20658. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/PythonInterface/test/python/plugins/algorithms/MeanTest.py
+++ b/Framework/PythonInterface/test/python/plugins/algorithms/MeanTest.py
@@ -10,9 +10,10 @@ class MeanTest(unittest.TestCase):
         a = CreateWorkspace(DataX=[1,2,3],DataY=[1,2,3],DataE=[1,1,1],UnitX='TOF')
         try:
             c = Mean(Workspaces='a,b') # 'b' does not exist.
-            self.fail("Should not have got here. Should throw without both workspace names indexing real workspaces in IDF.")
         except RuntimeError:
             pass
+        else:
+            self.fail("Should not have got here. Should throw without both workspace names indexing real workspaces in IDF.")
         finally:
             DeleteWorkspace(a)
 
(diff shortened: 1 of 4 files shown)


AnthonyLim23:
need to add a test

Framework/PythonInterface/test/python/plugins/algorithms/MeanTest.py
```diff
@@ -33,9 +35,23 @@ def test_throws_if_workspace_axis1_unequal(self):
...
             self.fail("Should not have got here. Should throw as axis1 is unequal in size between a and b.")
+        finally:
+            DeleteWorkspace(a)
+            DeleteWorkspace(b)
+
+    def test_throws_if_workspace_unorded(self):
+        a = CreateWorkspace(DataX=[1,2,1,2],DataY=[1,2,3,4],DataE=[1,1,1,1],UnitX='TOF',NSpec=2)
+        b = CreateWorkspace(DataX=[1,2,2,1],DataY=[1,2,3,4],DataE=[1,1,1,1],UnitX='TOF',NSpec=2)
+        try:
+            c = Mean(Workspaces='a,b') # 'a' and 'b' are different sizes.
```
gemmaguest: I think the comment here is incorrect (copy-paste error?)

gemmaguest (changes requested):
The changes look good :+1:. I get the correct results if the data are ok, or sensible error messages if the workspaces need to be sorted or are not the same shape.

Just one minor point about a misleading comment.

gemmaguest (approved):

gemmaguest:
All looks good now :+1:. The tests probably don't need to be re-run because the last commit was just a comment change.

PR #20694, opened 2017-09-28
Muon crash

### LINKED ISSUE #20693: Muon Analysis crash
Muon Analysis will crash if the user deletes all of the workspaces while in the data analysis tab.

AnthonyLim23:
Just added an error message if the user deletes all of the files while using muon analysis. 

Fixes #20693. 

**Release Notes** 

---
Does not need to be in release notes -not reported by user. 

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/qt/widgets/common/src/MuonFitPropertyBrowser.cpp
+++ b/qt/widgets/common/src/MuonFitPropertyBrowser.cpp
@@ -348,7 +348,7 @@ void MuonFitPropertyBrowser::setWorkspaceName(const QString &wsName) {
 * @param prop :: A pointer to the function name property
 */
 void MuonFitPropertyBrowser::enumChanged(QtProperty *prop) {
-  if (m_workspaceNames.size() == 0) {
+  if (m_workspaceNames.empty() && this->isVisible()) {
     g_log.error("No Data available. Please load Some data.");
     return;
   }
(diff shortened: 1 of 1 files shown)


samueljackson92:
@AnthonyLim23 This stops the crash but now I get:
```
No Data available. Please load Some data.
```

Printed 5 times when I open the interface.

AnthonyLim23:
ah. That will be because it sets up the interface at start up and there will be no data. The message is relevant but if we want to remove it we will need to be a bit more clever.

qt/widgets/common/src/MuonFitPropertyBrowser.cpp
```diff
@@ -348,6 +348,10 @@ void MuonFitPropertyBrowser::setWorkspaceName(const QString &wsName) {
 * @param prop :: A pointer to the function name property
 */
 void MuonFitPropertyBrowser::enumChanged(QtProperty *prop) {
+  if (m_workspaceNames.empty() && this->isVisible()) {
+    g_log.error("No Data available. Please load Some data.");
+    return;
+  }
```
samueljackson92: This is still not quite right. I think we always want to return when `m_workspaceNames` is empty but we only want to print the error if the `this` is visible AND `m_workspaceNames` is empty. 

i.e.
```cpp
if(m_workspaceNames.empty()) {
    if (this->visible()))
        g_log.error("...");
    return;
}
```

samueljackson92 (changes requested):

samueljackson92 (approved):
This no longer crashes and now only shows an error on the tab that causes the crash.

PR #20696, opened 2017-09-28
Colour Fill plots of ragged workspaces.

### LINKED ISSUE #20685: Colour fill plot doesn't show ragged workspaces correctly
### Expected behavior

Colour fill plot should draw each spectrum using that spectrum's x axis.

### Actual behavior

It seems that only the first spectrum axis is being used.

Draw this workspace:

```
w=CreateWorkspace(DataX='1,2,3,4,1,3,5,9,1,4,7,11,1,2,3,4', DataY='1,2,3,4,1,2,3,4,1,2,3,4', NSpec=4)
```

### Platforms affected

All

mantid-roman:
Check x axes of all histograms before setting the x scale.

**To test:**

Create a workspace with this script:
```
w=CreateWorkspace(DataX='1,2,3,4, 2,3,5,9, 3,4,7,11, 1,2,3,4', DataY='1,5,2,1,2,3,1,2,3,1,2,3', NSpec=4)
```
Then open a Colour Fill plot. It should show the data correctly.

Fixes #20685

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidPlot/src/Spectrogram.cpp
+++ b/MantidPlot/src/Spectrogram.cpp
@@ -250,10 +250,10 @@ MantidQt::API::QwtRasterDataMD *Spectrogram::dataFromWorkspace(
          ++iHisto) {
       auto &x = matrixWorkspace->x(iHisto);
       if (x.front() < minX) {
-        minX = x.front();
+        minX = static_cast<Mantid::coord_t>(x.front());
       }
       if (x.back() > maxX) {
-        maxX = x.back();
+        maxX = static_cast<Mantid::coord_t>(x.back());
       }
     }
   }
--- a/MantidPlot/src/Spectrogram.cpp
+++ b/MantidPlot/src/Spectrogram.cpp
@@ -39,6 +39,7 @@
 #include "Mantid/MantidMatrix.h"
 #include "Mantid/MantidMatrixFunction.h"
 #include "MantidAPI/IMDIterator.h"
+#include "MantidAPI/MatrixWorkspace.h"
 #include "MantidKernel/Strings.h"
 #include "MantidKernel/make_unique.h"
 #include "MantidQtWidgets/Common/PlotAxis.h"
@@ -227,6 +228,7 @@ MantidQt::API::QwtRasterDataMD *Spectrogram::dataFromWorkspace(
   // colour range
   QwtDoubleInterval fullRange =
       MantidQt::API::SignalRange(*workspace).interval();
+
   if (range) {
     wsData->setRange(*range);
   } else {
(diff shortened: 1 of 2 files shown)


MantidPlot/src/Spectrogram.cpp
```diff
@@ -237,6 +239,25 @@ MantidQt::API::QwtRasterDataMD *Spectrogram::dataFromWorkspace(
...
+
+  // A MatrixWorkspace can be ragged. Make sure the x axis covers all
+  // histograms.
+  auto matrixWorkspace =
+      boost::dynamic_pointer_cast<const Mantid::API::MatrixWorkspace>(
+          workspace);
+  if (matrixWorkspace) {
+    for (size_t iHisto = 0; iHisto < matrixWorkspace->getNumberHistograms();
+         ++iHisto) {
+      auto &x = matrixWorkspace->x(iHisto);
```
samueljackson92: `const auto& x`?

MantidPlot/src/Spectrogram.cpp
```diff
@@ -237,6 +239,25 @@ MantidQt::API::QwtRasterDataMD *Spectrogram::dataFromWorkspace(
...
+      if (x.front() < minX) {
+        minX = static_cast<Mantid::coord_t>(x.front());
+      }
+      if (x.back() > maxX) {
+        maxX = static_cast<Mantid::coord_t>(x.back());
+      }
+    }
+  }
```
samueljackson92: This is good. But can we move it to a separate private method? This one already does quite a lot.

samueljackson92 (changes requested):
Colour fill plot now looks sensible 👍 . A couple of minor code comments.

MantidPlot/src/Spectrogram.cpp
martyngigg: Should we put this code in https://github.com/mantidproject/mantid/blob/master/Framework/API/src/MatrixWorkspace.cpp#L1357 instead and have `MWXDimension` refer to the X-array with the largest extent?

mantid-roman: * `x` is `const` because `matrixWorkspace` is `const`. Plus method `x(...)` returns a `const`. Do we need another `const` here?
* I am not sure that this will work. My understanding that a MatrixWorkspace can be an MD workspace only when all x vectors are the same.

martyngigg: @mantid-roman The `const` is not strictly necessary here but I think it communicates the intent better.  

I see what you're saying although `getDimension` always worked for `MatrixWorkspace` anyway and has never checked the validity of the common bins. `MWXDimension` has always just arbitrarily referred to `x(0)` but we could just have it refer to he largest X vector and then that will at least ensure the dimension can encompass all of the data.

mantid-roman: @martyngigg There may not be a single x vector that covers the whole range. The axis vector has to be a combination of different vectors. How to bin it? Are you sure this change will not break the code that uses dimensions elsewhere?

martyngigg: @mantid-roman Good point. Let's leave this how it is for the release but I think we should see what effect it has so can we put this change into `master`?

samueljackson92 (approved):

NickDraper:
Build failures are unrelated

PR #20699, opened 2017-09-28
ISIS Reflectometry Re #20675: Fixed Loading and Saving of .tbl files.

### LINKED ISSUE #20675: Regressions in save/load table for DataProcessorWidget
It looks like some bugs have been introduced in the saving/loading of `.tbl` files. I think these are related to the addition of a the new `HiddenOptions` column in the DataProcessorWidget.

### Expected behavior
- You should be able to load the `INTER_NR_test2.tbl` test file into the DataProcessorWidget.
- You should be able to save and re-load the contents of the DataProcessorWidget to a `.tbl` file.

### Actual behavior
- The `INTER_NR_test2.tbl` file cannot be loaded.
- The format of a saved `.tbl` file is incorrect and you cannot re-load it.

*INTER_NR_test2.tbl*
- Open the new reflectometry interface from `Interfaces -> Reflectometry -> ISIS Reflectometry`
- From the `Reflectometry` menu, select `Import .TBL`
- Browse to the `INTER_NR_test2.tbl` file in the [ISIS sample data](http://download.mantidproject.org/), enter a workspace name e.g. `table` and click `Run`
- You get the error `Could not open workspace: Selected table has the incorrect number of columns to be used as a data processor table.`

This file is in an old format which supported the old reflectometry GUI and has more columns than the new GUI. However, this operation works in the new GUI in Mantid 3.10 so we probably had some code for backwards compatibility that has been broken somehow.

*Save/load*
- Open the new reflectometry interface from `Interfaces -> Reflectometry -> ISIS Reflectometry`
- Set the instrument to `POLREF` and the investigation id to `1520249`
- Select several runs e.g. 14885-14890 and click `Transfer`
- From the `Reflectometry` menu click `Save table as` and enter a workspace name e.g. `mytable`.
- From the `Reflectometry` menu click `Export .TBL`, make sure the workspace name you specified is selected and save it to a file.
- From the `Reflectometry` menu click `Import .TBL` and select the file you saved.
- You get an error in the log, `A line must contain 8 cell-delimiting commas. Found 10`

Note that if you look at the contents of the saved `.tbl` file, the `HiddenOptions` header is on the second line rather than the first, which looks like a bug in the export. This is what causes the 10 commas noted in the error message, along with the fact that the new `HiddenOptions` column means there should now be 9 cell-delimiting commas rather than 8, so the check also needs updating.

### Platforms affected
All

edwardb96:
PR #20136 added hidden columns to the `DataProcessorWidget` however the `SaveTBL` and `LoadTBL` algorithms were not updated. This lead to the regression #20675.

This PR updates these algorithms (in the case of `SaveTBL` by using a more robust mechanism to determine the last column) and therefore fixes this regression.

**To test:**
- Follow the instructions in #20675.
- Code review.

Fixes #20675. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


gemmaguest (approved):
- The changes look good and both of the bugs in the linked issue are fixed :+1:
- `LoadTBL` and `SaveTBL` can both handle different numbers of columns and rows.
- They can deal with special characters in the content and the headers, except for some minor existing bugs (see #20708)
- `LoadTBL` reports a sensible error if the rows contain different numbers of columns
- The reflectometry GUI reports a sensible error if you try to import a table with too few or too many columns.

PR #20710, opened 2017-09-28
Fix performance regression in MaxEnt performance test

peterfpeterson:
There was a [regression in performance of `MaxEnt`](http://builds.mantidproject.org/job/master_performancetests2/Master_branch_performance_tests/AlgorithmsTest.MaxEntTestPerformance.testExecReal.htm) that was introduced by the `blocksize()` changes of #20451. This addresses that.

**To test:**

`ctest -R MaxEntTest` a few times before and after the change and marvel at how much faster it is now.

*Does not need to be in the release notes* because it fixes a performance issue that was recently introduced and shouldn't have been seen by users of stable releases.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/src/MaxEnt.cpp
+++ b/Framework/Algorithms/src/MaxEnt.cpp
@@ -224,7 +224,7 @@ void MaxEnt::exec() {
   size_t nspec = inWS->getNumberHistograms();
   // Number of data points - assumed to be constant between spectra or
   // this will throw an exception
-  size_t npoints = inWS->y(0).blocksize() * resolutionFactor;
+  size_t npoints = inWS->blocksize() * resolutionFactor;
   // Number of X bins
   const size_t npointsX = inWS->isHistogramData() ? npoints + 1 : npoints;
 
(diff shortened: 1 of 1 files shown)


Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -227,8 +222,8 @@ void MaxEnt::exec() {
   MatrixWorkspace_const_sptr inWS = getProperty("InputWorkspace");
   // Number of spectra
   size_t nspec = inWS->getNumberHistograms();
-  // Number of data points
-  size_t npoints = inWS->blocksize() * resolutionFactor;
+  // Number of data points - assumed to be constant between spectra
+  size_t npoints = inWS->y(0).size() * resolutionFactor;
```
SimonHeybrock: So what happens now if you call the algorithm with a workspace that does not have constant histogram size?

martyngigg (approved):
With the addition of the `blocksize` check this looks okay to me.

LamarMoore (approved):
Changes look good and performance has significantly improved.

PR #20721, opened 2017-09-29
Fix RayTracerTesterTestPerformance indicated issues 

OwenArnold:
Fixes #20274

**Notes**

Implementation is similar to what `CompAssembly::getBoundingBox` did internally, but pushes the cache onto the client, InstrumentRayTracer in this case, including creation, locks, cache invalidation etc which we do not want to manage centrally.

I do not distinguish between any type of component when adding to the cache. Detectors etc also end up in the cache. Reduces the code, but let me know if you see this becoming a problem for other use-cases.

As further optimisations, that would help us keep away from Component level cache territory, we have discussed the possibility of recognising tubes for special treatment (`ComponentInfo::boundingBox`) based on top and bottom pixel calculations for whole-tube bounding box. We may also be able to do something with "eightpacks" based on the CNCS definition, which gives exactly the same object tree as `RecangularDetectors`, which we already optimise for. However, Mantid lacks a reliable way to identify tubes or eightpacks (type system not rich enough for this in current IComponent hierachy) so we are not implementing this now.

**Tester**
* Review code
* Maybe check `RayTracerTesterTestPeformance` too to ensure it won't time-out

### DIFF
--- a/Framework/Algorithms/src/RayTracerTester.cpp
+++ b/Framework/Algorithms/src/RayTracerTester.cpp
@@ -55,6 +55,7 @@ void RayTracerTester::exec() {
   int NumAzimuth = getProperty("NumAzimuth");
   int NumZenith = getProperty("NumZenith");
   Progress prog(this, 0.3, 1.0, NumAzimuth);
+  InstrumentRayTracer tracker(ws->getInstrument());
   for (int iaz = 0; iaz < NumAzimuth; iaz++) {
     prog.report();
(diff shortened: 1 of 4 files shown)


martyngigg (approved):
Confirmed that the test time is back to an acceptable level. Good to go.

martyngigg:
The test failures are unrelated.

@mantidproject/gatekeepers Can we get this merged ASAP, please? It will fix the performance test builds.

PR #20732, opened 2017-09-29
Added a new except statement to catch runtime error

### LINKED ISSUE #20728: SANS Gui not propagating errors when it can't find the correct file
### Expected behavior
When it can't find the file GUI should exit with an error specifying it can't find the file.
### Actual behavior
The python code exits but then 
Download the ISIS sample data.
OPen sans GUI v2
select MaskFile.txt as the relevant mask file
type 74044 as the sample scatter run
hit process
### Platforms affected
All

Matthew-Andrew:
Description of work.

**To test:**
Download the ISIS sample data.
OPen sans GUI v2
select MaskFile.txt as the relevant mask file
type 74044 as the sample scatter run
hit process

The expected result is that this will raise an exception saying that it cannot find the file. 

Fixes #20728  

**Release Notes** 
Bug fix so no release notes needed.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/scripts/SANS/sans/gui_logic/presenter/run_tab_presenter.py
+++ b/scripts/SANS/sans/gui_logic/presenter/run_tab_presenter.py
@@ -847,6 +847,9 @@ class RunTabPresenter(object):
                     raise RuntimeError("There was a bad entry for row {}. Ensure that the path to your files has "
                                        "been added to the Mantid search directories! See here for more "
                                        "details: {}".format(row, str(e)))
+                except:
+                    self._view.halt_process_flag()
+                    raise
         return states
 
     def _populate_row_in_table(self, row):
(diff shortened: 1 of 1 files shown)


AntonPiccardoSelg (changes requested):
I think this should be in `on_processed_clicked` in the `RunTabPresenter`. We should have one general `try-except` clause there. Any exception in there should be caught and halt the reduction. This also means we would only have one `halt` call. 

`_create_states` is called on paths which don't relate to the `DataProcessorWidget`, hence calling `halt_process_flag` does not make sense there.

What do you think?

Mantid-Matthew:
That does sound cleaner I'll look at changing that now.

AntonPiccardoSelg (approved):
I tried to run a LOQ reduction but with ZOOM enabled, this meant that the data could not be found. An exception text was produced and the reduction did not proceed -- as expected. All looks good. 

Can be merged when all tests have passed

PR #20736, opened 2017-10-02
Phase quad hard coded

### LINKED ISSUE #20714: PhaseQuad is hard coded to read table
The PhaseQuad algorithm is currently hard coded in how it reads the phase table. At present the order has to be 

detector ID, asymmetry, phase

### LINKED ISSUE #20734: PhaseQuad
Some new tests will be added to PhaseQuad to check that the new reading of the columns works.

AnthonyLim23:
PhaseQuad now works out which columns contain the relevant data.

**To test:**

Run the following code: 
```

tab.addColumn('double', 'Asymmetry')
tab.addColumn('double', 'Phase')
tab.addColumn('double', 'ID')

ws=Load("MUSR00062260.nxs")

nspectra = ws[0].getNumberHistograms()

for i in range(0, nspectra):
   tab.addRow([i, nspectra-i,i])
   
out=PhaseQuad(InputWorkspace='ws', PhaseTable='tab')
```

it should work. If you then run:

```

tab.addColumn('double', 'Phase')
tab.addColumn('double', 'Asymmetry')
tab.addColumn('double', 'ID')

ws=Load("MUSR00062260.nxs")

nspectra = ws[0].getNumberHistograms()

for i in range(0, nspectra):
   tab.addRow([nspectra-i,i,i])
   
outSwaped=PhaseQuad(InputWorkspace='ws', PhaseTable='tab')
```
you should find that the ws matches the other one (I just swapped the phase and asymmetry). 

If you put unreasonable (repeated names, missing names) column headers it should throw an error.

Fixes #20714. and fix #20734 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/inc/MantidAlgorithms/PhaseQuadMuon.h
+++ b/Framework/Algorithms/inc/MantidAlgorithms/PhaseQuadMuon.h
@@ -63,6 +63,7 @@ private:
   API::MatrixWorkspace_sptr squash(const API::MatrixWorkspace_sptr &ws,
                                    const API::ITableWorkspace_sptr &phase,
                                    const std::vector<double> &n0);
+ size_t findName(const std::string pattern, const std::vector<std::string> &names);
 };
 
 } // namespace Algorithms
--- a/Framework/Algorithms/inc/MantidAlgorithms/PhaseQuadMuon.h
+++ b/Framework/Algorithms/inc/MantidAlgorithms/PhaseQuadMuon.h
@@ -63,8 +63,6 @@ private:
   API::MatrixWorkspace_sptr squash(const API::MatrixWorkspace_sptr &ws,
                                    const API::ITableWorkspace_sptr &phase,
                                    const std::vector<double> &n0);
-  int findName(const std::vector<std::string> patterns,
-               const std::vector<std::string> &names);
 };
 
 } // namespace Algorithms
--- a/Framework/Algorithms/inc/MantidAlgorithms/PhaseQuadMuon.h
+++ b/Framework/Algorithms/inc/MantidAlgorithms/PhaseQuadMuon.h
@@ -63,7 +63,7 @@ private:
   API::MatrixWorkspace_sptr squash(const API::MatrixWorkspace_sptr &ws,
                                    const API::ITableWorkspace_sptr &phase,
                                    const std::vector<double> &n0);
- size_t findName(const std::string pattern, const std::vector<std::string> &names);
+ int findName(const std::string pattern, const std::vector<std::string> &names);
 };
 
 } // namespace Algorithms
--- a/Framework/Algorithms/inc/MantidAlgorithms/PhaseQuadMuon.h
+++ b/Framework/Algorithms/inc/MantidAlgorithms/PhaseQuadMuon.h
@@ -63,7 +63,8 @@ private:
   API::MatrixWorkspace_sptr squash(const API::MatrixWorkspace_sptr &ws,
                                    const API::ITableWorkspace_sptr &phase,
                                    const std::vector<double> &n0);
- int findName(const std::string pattern, const std::vector<std::string> &names);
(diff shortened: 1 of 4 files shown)


Framework/Algorithms/src/PhaseQuadMuon.cpp
```diff
@@ -92,7 +92,36 @@ std::map<std::string, std::string> PhaseQuadMuon::validateInputs() {
   if (tabWS->columnCount() != 3) {
     result["PhaseTable"] = "PhaseTable must have three columns";
   }
-
+  auto names = tabWS->getColumnNames();
+  for (int j = 0; j < 3; j++) {
+    std::transform(names[j].begin(), names[j].end(), names[j].begin(),
+                   ::tolower);
+  }
+  std::vector<std::string> goodNames = {"phase", "asym", "asymm", "asymmetry"};
```
martyngigg: I think it would be good to define these names in an anonymous namespace using a `std::array` so that they are defined only once. Perhaps separate the phase name from the asymmetry? We could even just use a prefix for the asymmetry and use [`boost::istarts_with`](http://www.boost.org/doc/libs/1_41_0/doc/html/boost/algorithm/istarts_with.html) to compare case insensitively to avoid having to transform?

Framework/Algorithms/test/PhaseQuadMuonTest.h
```diff
@@ -103,6 +106,15 @@ class PhaseQuadMuonTest : public CxxTest::TestSuite {
     TS_ASSERT_DELTA(specImE[20], 0.0031, 0.0001);
     TS_ASSERT_DELTA(specImE[50], 0.0035, 0.0001);
   }
+  // add test for no phase
```
martyngigg: Can we add these tests?

martyngigg (changes requested):

Framework/Algorithms/test/PhaseQuadMuonTest.h
AnthonyLim23: I was concerned about having time to complete them for the release. I should get time later. If not they will be the next thing that I do

AnthonyLim23:
@martyngigg I should have time this afternoon to add in the tests. 

I don't think `boost::isstarts_with` is a good idea. From what I understand it removes the upper case letters, then asks if the string starts with the pattern. So if our pattern was asym then the chemical formula AsYMg would match, which makes me nervous.

martyngigg:
@AnthonyLim23 Fair enough. I think it would be good to define this array of names once at the top of the file though rather than having the strings scattered throughout the file.

Framework/Algorithms/inc/MantidAlgorithms/PhaseQuadMuon.h
```diff
@@ -63,9 +63,11 @@ class DLLExport PhaseQuadMuon : public API::Algorithm {
   API::MatrixWorkspace_sptr squash(const API::MatrixWorkspace_sptr &ws,
                                    const API::ITableWorkspace_sptr &phase,
                                    const std::vector<double> &n0);
+  int findName(const std::vector<std::string> patterns,
```
martyngigg: Do we need this to be a member function? It could be in an anonymous namespace? Also the patterns `vector` should be taken by reference.

Framework/Algorithms/src/PhaseQuadMuon.cpp
```diff
@@ -6,6 +6,11 @@
 #include "MantidKernel/PhysicalConstants.h"
 #include "MantidKernel/Unit.h"
 
+namespace {
+std::vector<std::string> phaseNames = {"phase", "phi"};
```
martyngigg: We should prefer `std::array` with static array sizes.

martyngigg (approved):

PR #20745, opened 2017-10-02
Fix slow update of slice viewer on refresh

### LINKED ISSUE #20690: Slice Viewer replotting very slowly with live data updates
Spotted by OFFSPEC users and confirmed in unscripted testing.

Start live data with the following settings:

livedatasettings_offspec

and use this post-processing script:

```python
Rebin(input,"100,50,100000",OutputWorkspace=output)
ConvertUnits(output, Target="Wavelength",AlignBins=True,OutputWorkspace=output)
CropWorkspace(output,StartWorkspaceIndex=300, EndWorkspaceIndex=550, XMin=1.0,XMax=14.0,OutputWorkspace=output)
Integration(output,OutputWorkspace=output+"I")
GroupDetectors(output,SpectraList="395:440",OutputWorkspace=output+"group")
Transpose(output+"I",OutputWorkspace=output+"I")
```

Once started show the slice viewer on one of the workspaces. On subsequent updates the rendering takes a long time.

# Platforms
All

### LINKED ISSUE #20717: Crash due to memory leak with slice viewer and live data
Jos reported that a user had experienced a crash with Mantid 3.10.1 when it was left running live data overnight. I think this is related to the slice viewer as I have been able to reproduce it, but only with the slice viewer open - after about an hour it was using around 3GB and then crashed. With the colour fill plot open instead, it ran for around 4 hours with no problems (using only around 300MB). This was using a debug build on a Windows 10 laptop with 16GB RAM.

### Expected behavior
Mantid should not leak memory or crash. Mantid should probably not be using more than around 300-500MB in this example even in debug mode.

### Actual behavior
Mantid leaks memory and crashes. It is frequently using more than 1GB in debug mode when the slice viewer is open.

- Start live data with the settings and post-processing script shown below.
- When the workspaces appear, show the slice viewer for the LIVEDATA workspace.
- Leave it running for an hour or so and periodically check the memory usage in task manager.
- You should see the memory usage go up over time and eventually Mantid will probably crash.

livedatasettings_offspec

```
Rebin(input,"100,50,100000",OutputWorkspace=output)
ConvertUnits(output, Target="Wavelength",AlignBins=True,OutputWorkspace=output)
CropWorkspace(output,StartWorkspaceIndex=300, EndWorkspaceIndex=550, XMin=1.0,XMax=14.0,OutputWorkspace=output)
Integration(output,OutputWorkspace=output+"I")
GroupDetectors(output,SpectraList="395:440",OutputWorkspace=output+"group")
Transpose(output+"I",OutputWorkspace=output+"I")
```

### Platforms affected
Windows 10

martyngigg:
**Description of work**

Avoids repeatedly calls to redraw the slice viewer widget when a workspace update occurs. 

**To test:**

Use the testing instructions in #20690 to compare the refresh performance before and after the fix.

Fixes #20690 and fixes #20717 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/qt/widgets/sliceviewer/inc/MantidQtWidgets/SliceViewer/SliceViewer.h
+++ b/qt/widgets/sliceviewer/inc/MantidQtWidgets/SliceViewer/SliceViewer.h
@@ -424,6 +424,8 @@ private:
   AspectRatioType m_lastRatioState;
   QwtScaleDrawNonOrthogonal *m_nonOrthAxis0;
   QwtScaleDrawNonOrthogonal *m_nonOrthAxis1;
+
+  bool m_holdDisplayUpdates;
 };
 
 } // namespace SliceViewer
--- a/qt/widgets/sliceviewer/src/SliceViewer.cpp
+++ b/qt/widgets/sliceviewer/src/SliceViewer.cpp
@@ -89,7 +89,8 @@ SliceViewer::SliceViewer(QWidget *parent)
       m_peaksPresenter(boost::make_shared<CompositePeaksPresenter>(this)),
       m_proxyPeaksPresenter(
           boost::make_shared<ProxyCompositePeaksPresenter>(m_peaksPresenter)),
-      m_peaksSliderWidget(NULL), m_lastRatioState(Guess) {
+      m_peaksSliderWidget(NULL), m_lastRatioState(Guess),
+      m_holdDisplayUpdates(false) {
 
   ui.setupUi(this);
   std::string enableNonOrthogonal;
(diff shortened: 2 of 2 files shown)


AntonPiccardoSelg (commented):
I have tested the PR with 

* Non-orthogonal data
* overlayed peaks workspaces
* auto-binning

It seems to work fine. I have not been able to test this on Live data however

VickieLynch (approved):
Works well with TOPAZ data

gemmaguest:
I've linked this to #20717 regarding a memory leak which is also fixed by this PR. I ran the test described in that issue (basically the same as the testing instructions for #20690 but in debug mode on a low spec machine) for around 20 hours with no problems - Mantid remained using around 300MB, whereas before this PR it was getting to 3GB in about an hour and crashing. The GUI also did not completely hang, as it had been doing before this PR, although it was painfully slow - I think this is normal when running slice viewer in debug mode, especially on a low spec machine.

I'm just testing again in release mode to compare the refresh performance with v3.8, which is the last version before these problems were introduced.

gemmaguest (approved):
The code changes look good to me and the refresh performance is greatly improved compared to before the fix :+1:.

There is also no noticeable lag in interacting with the slice viewer or other GUI components while the slice viewer is refreshing. The performance in both respects seems to be back to what it was in v3.8.

martyngigg:
Thanks for the thorough testing everyone.

PR #20797, opened 2017-10-05
Fix for crash when UserSubWindow is closed while Python script is running

AntonPiccardoSelg:
Mantid would crash when a `UserSubWindow` was closed while a Python script was executing.  Here we ignore the `closeEvent` of a `UserSubWindow` when a script is running. In addition we send an abort request to the script runner.

This should be tested on instances of the `UserSubWindow` that can have a long running Python script execution, e.g. SANS (where it was discovered), but maybe also Muon

**To test:**

1. Open the old ISIS SANS GUI and load a user file which takes a long time to load, e.g. something with a calibration file (newer SANS2D reductions should do)
2. While it is loading, close the ISIS SANS GUI. 
  * Confirm that it cannot be closed and you get a pop up which says that the Window cannot be closed

No issue number or release notes

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidPlot/src/ApplicationWindow.cpp
+++ b/MantidPlot/src/ApplicationWindow.cpp
@@ -15104,10 +15104,7 @@ bool ApplicationWindow::runPythonScript(const QString &code, bool async,
 /**
  * Aborts a running Python script.
  */
-void ApplicationWindow::abortRunningPythonScript() {
-  m_iface_script->abort();
-}
-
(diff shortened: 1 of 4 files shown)


martyngigg:
I still observe a crash with the fix here. I'm going to close this as it effects in legacy C++ GUIs that are being rewritten in a scenario that is not too likely that users would hit.

PR #20798, opened 2017-10-05
Add switchable mutli-period columns to ISIS SANS GUI

### LINKED ISSUE #20748: Multiperiod settings in the ISIS SANS data table should be optional
Steve King remarked that the multi-period settings in the ISIS SANS data table should be optional.

AntonPiccardoSelg:
Fixes #20748

This PR adds the option to dynamically add or remove the multi-period columns in the data table in the new ISIS SANS GUI. This has been requested by Steve King. Since the vast majority of reductions don't make use of a multi-period set-up, it makes sense to have these columns off by default.

# To test:

- Open the sans v2 GUI
- Get the SANS data from _\\olympic\Babylon5\Scratch\Matthew Andrew\GUI_TEST_
- Check that the Multi period box below the table is initially unselected.
- Select it and check that the period columns appear.
- With the box unselected load a user file and batch file, click process.
- Try the same thing with the box selected. 
- Code Review

**Release Notes** 

Have added to release notes.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


NickDraper:
@Mantid-Matthew Can you resolve the conflicts please

scripts/Interface/ui/sans_isis/sans_data_processor_gui.py
```diff
@@ -558,6 +575,13 @@ def update_simple_line_edit_field(self, line_edit, value):
             gui_element = getattr(self, line_edit)
             gui_element.setText(str(value))
 
+    def is_multi_period_view(self):
+        return self.multi_period_check_box.isChecked()
+
+    def switch_to_multi_period_view(self):
```
josephframsay: Would it be a bit clearer (and potentially more useful) if this method was implemented as a setter? ie
```
def set_multi_period_view_mode(self, mode):
    self.multi_period_check_box.setChecked(mode)
```

scripts/Interface/ui/sans_isis/sans_data_processor_gui.py
```diff
@@ -259,9 +260,19 @@ def _setup_main_tab(self):
         self.data_processor_table.processingFinished.connect(self._processing_finished)
         self.data_processor_widget_layout.addWidget(self.data_processor_table)
         self.data_processor_table.dataChanged.connect(self._data_changed)
-
         self.data_processor_table.instrumentHasChanged.connect(self._handle_instrument_change)
 
+    def _setup_main_tab(self):
+        # --------------------------------------------------------------------------------------------------------------
+        # Header setup
+        # --------------------------------------------------------------------------------------------------------------
```
josephframsay: Are these big header comments necessary now that `setup_layout` has been refactored a bit? I would probably argue that the method name should be descriptive enough without, especially if they're this short - since `_setup_main_tab` now appears to set up the header, it should probably be renamed as such

scripts/SANS/sans/gui_logic/presenter/run_tab_presenter.py
```diff
@@ -850,21 +853,22 @@ def _get_table_model(self):
...
-            can_scatter = self._view.get_cell(row=row, column=CAN_SCATTER_INDEX, convert_to=str)
-            can_scatter_period = self._view.get_cell(row=row, column=CAN_SCATTER_PERIOD_INDEX, convert_to=str)
-            can_transmission = self._view.get_cell(row=row, column=CAN_TRANSMISSION_INDEX, convert_to=str)
-            can_transmission_period = self._view.get_cell(row=row, column=CAN_TRANSMISSION_PERIOD_INDEX, convert_to=str)
-            can_direct = self._view.get_cell(row=row, column=CAN_DIRECT_INDEX, convert_to=str)
-            can_direct_period = self._view.get_cell(row=row, column=CAN_DIRECT_PERIOD_INDEX, convert_to=str)
-            output_name = self._view.get_cell(row=row, column=OUTPUT_NAME_INDEX, convert_to=str)
-            user_file = self._view.get_cell(row=row, column=USER_FILE_INDEX, convert_to=str)
+            sample_scatter = self._view.get_cell(row=row, column=self.table_index['SAMPLE_SCATTER_INDEX'], convert_to=str)
+            sample_scatter_period = self._view.get_cell(row=row, column=self.table_index['SAMPLE_SCATTER_PERIOD_INDEX'], convert_to=str) if is_multi_period_view else ""  # noqa
```
josephframsay: This is all super dense and pretty difficult to make sense of. I don't know if this suggestion is even worse (and you may well be able to think of something better), but could you do something like the following: 
- Wrap all the calls to `self._view.get_cell(...)` up in a nested helper:
```
def cell_value(column_name):
    return self._view.get_cell(row=row, column=self.table_index[column_name], convert_to=str)
```
- So retrieving values from the view looks much nicer:
```
sample_scatter_period = cell_value("SAMPLE_SCATTER_PERIOD_INDEX") if is_multi_period_view else ""
```
As I say, I'm not sure whether this is good practice or shoddy Pythonry so feel free to dismiss, but it would at least remove the need for those `noqa` flags

scripts/SANS/sans/gui_logic/presenter/run_tab_presenter.py
```diff
@@ -964,22 +968,39 @@ def get_string_period(_tag):
         can_direct_period = get_string_entry(BatchReductionEntry.CanDirectPeriod, row)
         output_name = get_string_entry(BatchReductionEntry.Output, row)
 
+        # If one of the periods is not null, then we should switch the view to multi-period view
+        if sample_scatter_period or sample_transmission_period or sample_direct_period or can_scatter_period or \
```
josephframsay: It might be more readable to use `any` here:
```
if any((sample_scatter_period, sample_transmission_period, sample_direct_period, ...blah)):
```

josephframsay (changes requested):
The columns appear and disappear as expected - is there anything else I should be looking out for? The changes are mostly pretty nice - there's just a couple of places where the code could possibly be made a little bit less dense

scripts/Interface/ui/sans_isis/sans_data_processor_gui.py
Mantid-Matthew: I think you are right about this header comment not really needed anymore in a 4 line function. I actually prefer setup_main_tab as a name though.

scripts/Interface/ui/sans_isis/sans_data_processor_gui.py
Mantid-Matthew: Changed this as you suggested as I think it is better.

scripts/SANS/sans/gui_logic/presenter/run_tab_presenter.py
Mantid-Matthew: Changed this as you suggested as I think it is better.

scripts/SANS/sans/gui_logic/presenter/run_tab_presenter.py
Mantid-Matthew: Changed this as you suggested as I think it is better.

josephframsay (approved):

scripts/SANS/sans/gui_logic/presenter/run_tab_presenter.py
josephframsay: Ooh this is nicer than what I suggested - nice change

PR #20804, opened 2017-10-05
NormaliseToMonitor Detector Scan and Single Count Workspaces

### LINKED ISSUE #20619: NormaliseToMonitor detector scan support
`NormaliseToMonitor` should be able to support detector scans.

The options to use spectrum number and workspace index should throw for detector scan workspaces - this is not a safe way of doing the normalisation. Only the Monitor ID option should be used. No changes to the signature of the algorithm should be required.

ianbush:
These changes have 2 significant parts
 * Supporting workspaces with scanning detectors  - here there are more than one monitor spectrum to normalise to, so the input must accept the MonitorID only
 * Supporting workspaces with point data and single counts

**To test:**

Check various inputs, point data with more than one bin should fail, all else should pass.

Try loading the attached file for D2B, and check normalisation works correctly (monitor values and detector are artificially changed in this file to help test [508093_monitors_changed.nxs.zip](https://github.com/mantidproject/mantid/files/1375143/508093_monitors_changed.nxs.zip). This file is both scanning and single count per spectrum.

Further tests could be done with `CreateSampleWorkspace`, using the option to set the number of scan points and/or making a single bin workspace, then converting it to point data. Tests can also be performed for the previously supported cases, but these should have been picked up by the unit tests.

Fixes #20619.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


ianbush:
@SimonHeybrock A couple of points we discussed:

Framework/Algorithms/src/NormaliseToMonitor.cpp
```diff
@@ -112,7 +115,10 @@ bool MonIDPropChanger::monitorIdReader(
   }
   // are these monitors really there?
   // got the index of correspondent spectra.
-  std::vector<size_t> indexList = inputWS->getIndicesFromDetectorIDs(mon);
+  // Skip this for performance if number of histograms is large.
```
SimonHeybrock: How is it skipped?

Framework/Algorithms/src/NormaliseToMonitor.cpp
```diff
@@ -121,28 +127,26 @@ bool MonIDPropChanger::monitorIdReader(
       return true;
     }
   }
-  // index list can be less or equal to the mon list size (some monitors do not
+  // index list can be less or equal to the mon list size (some monitors do
+  // not
   // have spectra)
```
SimonHeybrock: Bad linebreaks.

Framework/Algorithms/src/NormaliseToMonitor.cpp
```diff
@@ -510,44 +544,111 @@ bool NormaliseToMonitor::setIntegrationProps() {
...
+  size_t monitorWorkspaceIndex = 0;
+
+  const auto &specInfo = inputWorkspace->spectrumInfo();
+  for (const auto workspaceIndex : m_workspaceIndexes) {
+    // Errors propagated according to
+    // http://docs.mantidproject.org/nightly/concepts/ErrorPropagation.html#error-propagation
+    // This is similar to that in MantidAlgorithms::Divide
+
+    size_t timeIndex = 0;
+    if (m_syncScanInput)
```
SimonHeybrock: Can remove `if`?

Framework/Algorithms/src/NormaliseToMonitor.cpp
```diff
@@ -510,44 +544,111 @@ bool NormaliseToMonitor::setIntegrationProps() {
...
+    const auto newYFactor =
+        1.0 / m_monitor->histogram(monitorWorkspaceIndex).y()[0];
+    const auto divisorError =
+        m_monitor->histogram(monitorWorkspaceIndex).e()[0];
+    const double yErrorFactor = pow(divisorError * newYFactor, 2);
+    monitorWorkspaceIndex++;
+
+    for (size_t i = 0; i < outputWorkspace->getNumberHistograms(); ++i) {
+      const auto &specDef = specInfo.spectrumDefinition(i);
+      if (specDef.size() > 0 && specDef[0].second != timeIndex)
```
SimonHeybrock: What happens if there are grouped detectors (with or without same time index)?

Framework/Algorithms/src/NormaliseToMonitor.cpp
```diff
@@ -561,99 +662,123 @@ void NormaliseToMonitor::normaliseBinByBin(
...
-          X.rawData(), Y.mutableRawData(), E.mutableRawData(), false);
-      // Recalculate the overall normalization factor
-      this->normalisationFactor(X, Y, E);
-    }
+  std::shared_ptr<SpectrumInfo> inputSpecInfo;
+  std::shared_ptr<SpectrumInfo> monitorSpecInfo;
+  if (m_syncScanInput) {
+    inputSpecInfo =
+        std::make_shared<SpectrumInfo>(inputWorkspace->spectrumInfo());
+    monitorSpecInfo = std::make_shared<SpectrumInfo>(m_monitor->spectrumInfo());
```
SimonHeybrock: This takes copies of `SpectrumInfo`? Is this needed? Why does this need to be in an `if`? Getting the reference is cheap I think, even if it is not needed later.

Framework/Algorithms/src/NormaliseToMonitor.cpp
```diff
@@ -561,99 +662,123 @@ void NormaliseToMonitor::normaliseBinByBin(
...
-  PARALLEL_CHECK_INTERUPT_REGION
-  if (hasZeroDivision) {
-    g_log.warning() << "Division by zero in some of the bins.\n";
+  for (auto &workspaceIndex : m_workspaceIndexes) {
+    // Get hold of the monitor spectrum
+    const auto &monX = m_monitor->binEdges(workspaceIndex);
+    auto monY = m_monitor->counts(workspaceIndex);
+    auto monE = m_monitor->countStandardDeviations(workspaceIndex);
+    size_t timeIndex = 0;
+    if (m_syncScanInput)
```
SimonHeybrock: Can remove `if`?

Framework/Algorithms/src/NormaliseToMonitor.cpp
```diff
@@ -561,99 +662,123 @@ void NormaliseToMonitor::normaliseBinByBin(
...
+    bool hasZeroDivision = false;
+    Progress prog(this, 0.0, 1.0, numHists);
+    // Loop over spectra
+    PARALLEL_FOR_IF(
+        Kernel::threadSafe(*inputWorkspace, *outputWorkspace, *m_monitor))
+    for (int64_t i = 0; i < int64_t(numHists); ++i) {
+      PARALLEL_START_INTERUPT_REGION
+      prog.report();
+
+      if (m_syncScanInput) {
```
SimonHeybrock: Can remove `if`?

Framework/Algorithms/src/NormaliseToMonitor.cpp
```diff
@@ -561,99 +662,123 @@ void NormaliseToMonitor::normaliseBinByBin(
...
+        // --- EventWorkspace ---
+        EventList &outEL = outputEvent->getSpectrum(i);
+        outEL.divide(X.rawData(), Y.mutableRawData(), E.mutableRawData());
+      } else {
+        // --- Workspace2D ---
+        auto &YOut = outputWorkspace->mutableY(i);
+        auto &EOut = outputWorkspace->mutableE(i);
+        const auto &inY = inputWorkspace->y(i);
+        const auto &inE = inputWorkspace->e(i);
+        outputWorkspace->mutableX(i) = inputWorkspace->x(i);
```
SimonHeybrock: Can set up sharing X instead of making copy?

Framework/Algorithms/test/NormaliseToMonitorTest.h
```diff
@@ -441,6 +445,172 @@ class NormaliseToMonitorTest : public CxxTest::TestSuite {
...
+    NormaliseToMonitor alg;
+    alg.setChild(true);
+    alg.setRethrows(true);
+    TS_ASSERT_THROWS_NOTHING(alg.initialize())
+    TS_ASSERT_THROWS_NOTHING(alg.setProperty("InputWorkspace", merged))
+    TS_ASSERT_THROWS_NOTHING(alg.setProperty("OutputWorkspace", "outputWS"))
+    TS_ASSERT_THROWS_NOTHING(alg.setProperty("MonitorID", "9"))
+    TS_ASSERT_THROWS_EQUALS(alg.execute(), std::runtime_error & e,
+                            std::string(e.what()),
+                            "More then one spectra corresponds to the "
```
SimonHeybrock: spectrum

Framework/Algorithms/test/NormaliseToMonitorTest.h
```diff
@@ -441,6 +445,172 @@ class NormaliseToMonitorTest : public CxxTest::TestSuite {
...
+                            "requested monitor ID, which is unheard of")
```
SimonHeybrock: "unheard of" sounds a bit too colloquial?

Framework/Beamline/src/DetectorInfo.cpp
```diff
@@ -102,6 +102,12 @@ size_t DetectorInfo::scanSize() const {
   return m_positions->size();
 }
 
+/**
+ * Returns true if all of the detectors all have the same scan interval. Will
+ * return false if DetectorInfo is not scanning.
```
SimonHeybrock: I am now thinking that `true` would make more sense since no scanning implies that all intervals are identical, i.e., sync?

docs/source/algorithms/NormaliseToMonitor-v1.rst
```diff
@@ -55,20 +55,26 @@ technically dimensionless.
 Restrictions on the input workspace
 ###################################
 
-The data must be histogram, non-distribution data.
+The data must be histogram, non-distribution data. The exception to this is for workspaces that contain point data with a single
```
SimonHeybrock: `The exception` refers only to the "histogram" bit, not the "not-distribution" bit, right?

SimonHeybrock (changes requested):
Looks generally really good, tests added and documentation updated 👍. `NormaliseToMonitor` is quite messy and some small cleanup was started here, but a full cleanup is beyond the scope of this PR, I think.

Small comments:

Framework/Algorithms/src/NormaliseToMonitor.cpp
ianbush: I've removed this now is is not needed. I was thinking an error was thrown if no instrument is present, but getting the reference is not an issue.

Framework/Algorithms/src/NormaliseToMonitor.cpp
ianbush: This `if` is required to avoid a crash in the unit tests when getting the spectrum definitions for a workspace without and instrument.

Framework/Algorithms/src/NormaliseToMonitor.cpp
ianbush: Removed.

Framework/Algorithms/test/NormaliseToMonitorTest.h
ianbush: This made me laugh the first time I triggered when trying with a scanning workspace. I'll change it to something more in keeping with the normal error messages.

Framework/Beamline/src/DetectorInfo.cpp
ianbush: I have been back and forth with this. I think the name implies the current behaviour, i.e. it is a scan with identical time indexes. We could have `hasIdenticalScanIntervals` instead?

The question would be what the more common use would be? In `NormaliseToMonitor` I would then `isScanning && hasIdenticalScanIntervals` or similar. Perhaps this is clearer?

docs/source/algorithms/NormaliseToMonitor-v1.rst
ianbush: Correct, fixed in the docs.

ianbush (commented):
Pushed some fixes. The main question is whether to change the behaviour of `isSyncScan` to `true` for non-scanning workspaces.

Framework/Algorithms/src/NormaliseToMonitor.cpp
SimonHeybrock: I see, I guess the problem is that no detector is associated to the monitor so the spectrum definition is empty.

Framework/Beamline/src/DetectorInfo.cpp
SimonHeybrock: Maybe what we should do is avoid overlap with the `isScanning()` method and make the naming convention agree?

Maybe `isScanningSync()`? Well actually that is also unclear, depending on whether "sync" stands for "synchronous" or "synchronously" it could be implying `isScanning` or not? So then in long form `isScanningSynchronously()` to make it clear?

Your suggestions also sounds ok (even though "identical" leaves a it a bit unclear what is identical with what, it could also be taken to mean that all intervals have the same length or something). To decide which option you go with maybe ask yourself whether most client code would be interested in identical scan intervals, or in actual scanning.

Framework/Algorithms/src/NormaliseToMonitor.cpp
```diff
@@ -152,6 +148,15 @@ bool MonIDPropChanger::monitorIdReader(
   return values_redefined;
 }
 
+bool spectrumDefinitionsMatchTimeIndex(const SpectrumDefinition &specDef,
+                                       const size_t timeIndex) {
+  if (specDef.size() > 0)
```
SimonHeybrock: Do not need this `if`?

Framework/Beamline/src/DetectorInfo.cpp
ianbush: As per Slack conversation we will keep this as `isSyncScan` with the current behaviour. There is no perfect option, but this does offer all the functionality needed.

SimonHeybrock (approved):

martyngigg:
@ianbush The `cppcheck` build is failing.

ianbush:
Thanks @martyngigg, I had missed that failure. All fixed now.

martyngigg:
@ianbush One of the ISIS inelastic system tests is segfaulting when running `NormaliseToMonitor`

ianbush:
@martyngigg @SimonHeybrock This is passing now, I think the issue was a workspace which did not have an instrument properly associated.

martyngigg:
Thanks, @ianbush.

PR #20813, opened 2017-10-05
Make WorkspaceProperty::isDefault() to work with workspaces not in the ADS

### LINKED ISSUE #20790: WorkspaceProperty::isDefault() doesn't check for workspaces not in the ADS
`WorkspaceProperty::isDefault()` returns `true` even if the property is set to a valid non-default workspace when the workspace is not in the ADS. This behavior can be reproduced with the following mock algorithm:

```python

class A(PythonAlgorithm):
    def PyInit(self):
        self.declareProperty(MatrixWorkspaceProperty('P', defaultValue='', direction=Direction.Input, optional=PropertyMode.Optional))
    def PyExec(self):
        if self.getProperty('P').isDefault:
            raise RuntimeError('P is default.')
    def category(self):
        return 'Examples'

AlgorithmFactory.subscribe(A)
```

and with the following script:

```python
alg = AlgorithmManager.Instance().createUnmanaged('CreateSampleWorkspace')
alg.setChild(True)
alg.setAlwaysStoreInADS(False)
alg.initialize()
alg.setProperty('OutputWorkspace', 'unused')
alg.execute()
ws = alg.getProperty('OutputWorkspace').value
alg = AlgorithmManager.Instance().create('A')
alg.initialize()
alg.setProperty('P', ws)
alg.execute()
```

soininen:
This PR fixes a case where, when one sets a `WorkspaceProperty` to a workspace which is not in the ADS, `WorkspaceProperty::isDefault()` would still return `true`. This enables checking if the user has given a non-ADS workspace for an optional input property by `isDefault`.

**To test:**

Create a mock algorithm using the first script provided in the issue. The second script should raise an error in Mantid logs before these changes are applied while after applying, the mock algorithm should run fine. 

Fixes #20790.

**Release Notes** 

The user facing changes to the Python API are mentioned in the Framework's release notes.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


josephframsay (changes requested):

josephframsay (approved):
Nice work, looks good to me!

Framework/API/inc/MantidAPI/WorkspaceProperty.tcc
```diff
@@ -182,7 +190,7 @@ std::string WorkspaceProperty<TYPE>::setDataItem(
   boost::shared_ptr<TYPE> typed = boost::dynamic_pointer_cast<TYPE>(value);
   if (typed) {
     std::string wsName = typed->getName();
```
martyngigg: I don't think `wsName` is now used anywhere else other than the assign below so we could just do
```
m_workspaceName = typed->getName()
```
to avoid the temporary copy?

martyngigg (changes requested):

martyngigg (approved):

PR #20850, opened 2017-10-06
Remove legacy bounding box code

### LINKED ISSUE #20848: Remove Legacy Bounding Box code
Follow on from #20264 and #20720. Remove  cache in `ParameterMap` used by Component Assemblies for caching bounding boxes.

OwenArnold:
fixes #20848 by deleting unused code from `ParameterMap` and removing all code that used it. The same thing can be done for `m_cacheLocMap` and `m_cacheRotMap` easily now since we use `ComponentInfo` for all of this now. I am going to do that in a separate PR as I want to double check the performance effects of these changes via the master run of the performance tests first.

**Tester**
A code review will be sufficient


SimonHeybrock (approved):
👍

PR #20886, opened 2017-10-10
Remove Position/Rotation Cache

### LINKED ISSUE #20853: Remove legacy rotation and location cache
Follow on from #20848 to remove position and location caches from `ParameterMap`. Not needed since the introduction of `ComponentInfo`

OwenArnold:
fixes #20853

Delete unused code. Will need to be careful around performance possibly.

**WIP Do not merge**.


OwenArnold:
It's possible that putting these changes into effect would be undesirable at this point in time. @SimonHeybrock  and I have discussed this briefly.

OwenArnold:
The physical instrument in Mantid does not go via `ComponentInfo` or `DetectorInfo` for fetching positions or rotations. Therefore the position & rotation cache is still relevant to performance in those cases.

PR #20890, opened 2017-10-10
Frequency Domain Analysis

### LINKED ISSUE #20655: Maintenance on Frequency Domain Analysis (FDA)
Some of the files in FDA need to be renamed to be lower case and the changes need to be followed through into the code. 

The code will also be cleaned up where possible and tests added if they were missed.

AnthonyLim23:
This was a clean up of the FDA GUI. It has included more comments and doc strings. Some refactoring to reduce repetition between files.  

**To test:**
Open Muon Analysis and load some data.
Just have a play around with the GUI

Fixes #20655. 

**Release Notes** 

There is no new functionality so does not need to be in the release notes. 
---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


scripts/Muon/fft_presenter.py
```diff
@@ -100,6 +101,7 @@ def handleButton(self):
         self.thread.start()
         self.view.setPhaseBox()
 
+    # kills the thread at end of execution
     def handleFinished(self):
         self.activate()
         self.thread.deleteLater
```
soininen: Is this really how the thread gets cleaned up? It is not actually a call?

scripts/Muon/fft_view.py
```diff
@@ -127,7 +130,7 @@ def tableClick(self,row,col):
     def buttonClick(self):
         self.buttonSignal.emit()
 
-    #functions
+    # responses to cpmmands
```
soininen: cpmmands -> commands

scripts/Muon/fft_model.py
```diff
@@ -7,10 +7,16 @@
...
+    """
+    A class to wrap the different parts
+    of the FFT and its preprocessing.
+    This keeps the main FFT class simple.
+    """
     def __init__(self,FFT):
         self.name="FFT"
         self.model=FFT
 
+    #store the data in the wrapper for later
```
soininen: This (and similarly below) could be a real docstring, I think.

soininen (changes requested):
Played around with the Frequency domain analysis interface and as far as I can tell, it works fine. Except for the few minor comments the code looks good, too.

scripts/Muon/fft_presenter.py
AnthonyLim23: deleteLater is a slot, see https://stackoverflow.com/questions/28669195/qt-how-to-clean-up-a-qthread-when-you-do-not-subclass-it

soininen: I know it's a slot, but I was merely wondering if you would actually need to call the attribute in Python, i.e. `self.thread.deleteLater()`. This is what I found people doing on the few examples I found on the Internet. Anyway, I'm not familiar with PyQt so I trust you know what you're doing here.

scripts/Muon/fft_model.py
```diff
@@ -16,7 +16,9 @@ def __init__(self,FFT):
         self.name="FFT"
         self.model=FFT
 
-    #store the data in the wrapper for later
+    """
+    store the data in the wrapper for later
+    """
```
soininen: I believe the docstrings should come after the function signature, e.g.
```python
    def loadData(self, inputs):
    """Store the data in the wrapper for later."""
```

soininen (approved):

PR #20895, opened 2017-10-10
Abins: correct naming revisited

dymkowsk:
Change of naming:
DFT -> AbInitio or ab_initio
phonon -> vibrational or phonon
in function documentation, methods names, GUI etc.

Motivation:
 Both CRYSTAL and GAUSSIAN support not only density functional theory (DFT) but also Hartree-Fock (HF) method. The term ab-initio is more general than DFT and takes into account both DFT and HF.
 AbINS can use data from crystal and molecular calculations. Term phonon refers only to crystals and vibrational only to molecules. In general, data which AbINS uses can be phonon or vibrational data.

screenshot from 2017-10-11 11-02-59

**To test:**
Flag excludeInPullRequests enabled. Rerun all system tests manually.

Fixes #xxxx. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/PythonInterface/plugins/algorithms/Abins.py
```diff
@@ -50,17 +50,17 @@ def summary(self):
...
-        self.declareProperty(name="DFTprogram",
+        self.declareProperty(name="AbInitioProgram",
                              direction=Direction.Input,
                              defaultValue="CASTEP",
                              validator=StringListValidator(["CASTEP", "CRYSTAL", "DMOL3", "GAUSSIAN"]),
-                             doc="DFT program which was used for a phonon calculation.")
+                             doc="An ab initio program which was used for vibrational or phonon calculation.")
 
-        self.declareProperty(FileProperty("PhononFile", "",
+        self.declareProperty(FileProperty("VibrationalOrPhononFile", "",
```
martyngigg: Would something like `SimulationFile` be more general for the future?

Framework/PythonInterface/plugins/algorithms/Abins.py
```diff
@@ -50,17 +50,17 @@ def summary(self):
     def PyInit(self):
 
         # Declare all properties
-        self.declareProperty(name="DFTprogram",
+        self.declareProperty(name="AbInitioProgram",
```
martyngigg: Similarly how about `SimulationProgram` or is that too general?

martyngigg: Or even just `Program`?

scripts/test/AbinsCalculateSPowder2DTest.py
```diff
@@ -1,92 +0,0 @@
-from __future__ import (absolute_import, division, print_function)
```
martyngigg: What's the reason for deleting tests?

Framework/PythonInterface/plugins/algorithms/Abins.py
dymkowsk: @martyngigg I asked Sanghamitra about those names (I cc. you).

scripts/test/AbinsCalculateSPowder2DTest.py
dymkowsk: @martyngigg At this version of AbiINS there is no 2D map and this test is obsolete. I planning to add 2dMAp implementation with corresponding system test. So instead of keeping this unit test I plan to replace it with the system test for 2D map in the future.

martyngigg: :+1:

martyngigg (approved):
It has been confirmed via email that the property names should be kept as they are in this changeset.

Good to go.

Anders-Markvardsen:
I agree

PR #20912, opened 2017-10-13
Try  to make MaxEnt faster

mantid-roman:
Description of work.

Made some changes to the MaxEnt code that make sense to me.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/src/MaxEnt.cpp
+++ b/Framework/Algorithms/src/MaxEnt.cpp
@@ -15,9 +15,6 @@
 #include "MantidKernel/VectorHelper.h"
 #include <gsl/gsl_linalg.h>
 #include <numeric>
-#include <iostream>
-
-//#include "D:/Work/mantid_stuff/Testing/class/MyTestDef.h"
 
 namespace Mantid {
 namespace Algorithms {
@@ -286,8 +283,6 @@ void MaxEnt::exec() {
   outEvolTest = WorkspaceFactory::Instance().create(inWS, nspec, niter, niter);
 
   npoints = complexImage ? npoints * 2 : npoints;
-  
-  //_start_filter([](size_t it){return it < 50 && it % 1 == 0;});
 
   outEvolChi->setPoints(0, Points(niter, LinearGenerator(0.0, 1.0)));
   for (size_t s = 0; s < nspec; s++) {
@@ -315,10 +310,6 @@ void MaxEnt::exec() {
     // Run maxent algorithm
     for (size_t it = 0; it < niter; it++) {
 
-      //_item;
-      //_(data);
-      //_(image);
-
       // Iterates one step towards the solution. This means calculating
       // quadratic coefficients, search directions, angle and chi-sq
       maxentCalculator.iterate(data, errors, image, background);
@@ -326,10 +317,10 @@ void MaxEnt::exec() {
       // Calculate delta to construct new image (SB eq. 25)
       double currChisq = maxentCalculator.getChisq();
       auto coeffs = maxentCalculator.getQuadraticCoefficients();
-      auto delta = move(coeffs, chiTarget / currChisq, chiEps, alphaIter, image, background, distEps);
+      auto delta = move(coeffs, chiTarget / currChisq, chiEps, alphaIter);
 
       // Apply distance penalty (SB eq. 33)
-      //delta = applyDistancePenalty(delta, coeffs, image, background, distEps);
+      delta = applyDistancePenalty(delta, coeffs, image, background, distEps);
 
       // Update image
       auto dirs = maxentCalculator.getSearchDirections();
@@ -344,7 +335,7 @@ void MaxEnt::exec() {
       // Stop condition, solution found
       if ((std::abs(currChisq / chiTarget - 1.) < chiEps) &&
           (currAngle < angle)) {
-        std::cerr << "Stopped after " << it << " iterations " << currChisq << ' ' << chiTarget << std::endl;
+        g_log.information() << "Stopped after " << it << " iterations " << currChisq << ' ' << chiTarget << std::endl;
         break;
       }
 
@@ -428,8 +419,7 @@ std::vector<double> MaxEnt::toComplex(API::MatrixWorkspace_const_sptr &inWS,
 */
 std::vector<double> MaxEnt::move(const QuadraticCoefficients &coeffs,
                                  double chiTarget, double chiEps,
-                                 size_t alphaIter, const std::vector<double> &image,
-                                           double background, double distEps) {
+                                 size_t alphaIter) {
 
   double aMin = 0.; // Minimum alpha
   double aMax = 1.; // Maximum alpha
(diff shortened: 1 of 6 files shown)


Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -443,7 +456,7 @@ std::vector<double> MaxEnt::move(const QuadraticCoefficients &coeffs,
 
   std::vector<double> delta(dim, 0); // delta at current alpha
 
-  while ((fabs(eps) > chiEps) && (iter < alphaIter)) {
+  while ((fabs(eps / chiTarget) > chiEps) && (iter < alphaIter)) {
```
AnthonyLim23: is chiEps a relative value or absolute? You seem to have gone for relative and the old code absolute.

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -618,7 +631,7 @@ std::vector<double> MaxEnt::applyDistancePenalty(
   auto newDelta = delta;
   if (dist > distEps * sum / background) {
     for (size_t k = 0; k < delta.size(); k++) {
-      newDelta[k] *= sqrt(sum / dist / background);
+      newDelta[k] *= sqrt(distEps * sum / dist / background);
```
AnthonyLim23: why have you added a distEps?

Framework/Algorithms/src/MaxEnt/MaxentCalculator.cpp
```diff
@@ -245,7 +258,7 @@ void MaxentCalculator::iterate(const std::vector<double> &data,
       m_coeffs.s1[k][0] += m_directionsIm[k][i] * sgrad[i];
       m_coeffs.c1[k][0] += m_directionsIm[k][i] * cgrad[i];
     }
-    m_coeffs.c1[k][0] /= chiSq;
+    m_coeffs.c1[k][0] /= fac * npoints;
```
AnthonyLim23: this will be 4*npoints larger. Why have you changed this?

Framework/Algorithms/src/MaxEnt/MaxentCalculator.cpp
```diff
@@ -271,7 +284,7 @@ void MaxentCalculator::iterate(const std::vector<double> &data,
           m_coeffs.c2[k][l] += directionsDat[k][i] * directionsDat[l][i] /
                                m_errors[i] / m_errors[i];
       }
-      m_coeffs.c2[k][l] *= 2.0 / chiSq;
+      m_coeffs.c2[k][l] *= 2.0 / fac;
```
AnthonyLim23: this would be a factor of 4 larger. What is the logic?

AnthonyLim23:
the tests are failing. The y values are slightly different.

Framework/Algorithms/src/MaxEnt.cpp
mantid-roman: It has to be relative because `chiTarget` in this function is a relative quantity in the range 0- 1

Framework/Algorithms/src/MaxEnt.cpp
mantid-roman: If `dist` is too large you want to make `delta` smaller. The factor is the solution of the equation
```
factor^2 * dist == distEps * sum / background
```
After applying this factor the above `if` statement would fail if tried again.

Framework/Algorithms/src/MaxEnt/MaxentCalculator.cpp
mantid-roman: Those probably come from the Fourier transform, I am not sure but after that change the `c1` and `c2` actually become the first and second derivatives of the chi squared at least in the `m_directionsIm[0]` direction.

I just realised that I didn't check the other direction. I'll do that.

Framework/Algorithms/src/MaxEnt/MaxentCalculator.cpp
mantid-roman: To make the quadratic function be equal to the actual chi squared in the search subspace. Without that they were different.

mantid-roman:
They are expected to be different. I don't want to change them yet because they may change again.

Framework/Algorithms/src/MaxEnt.cpp
AnthonyLim23: looking at the code I would have thought it was only alpha that had been constrained

Framework/Algorithms/src/MaxEnt/MaxentCalculator.cpp
AnthonyLim23: if it was Fourier transform surely we would expect some multiple of pi?

Framework/Algorithms/src/MaxEnt.cpp
mantid-roman: Alpha is the Lagrange multiplier for the constraint on the chi squared.

Framework/Algorithms/src/MaxEnt/MaxentCalculator.cpp
mantid-roman: I don't know I just compared the numbers.

AnthonyLim23:
@mantid-builder it is that the numbers are slightly different in the maxEnt test

AnthonyLim23:
let me know when you would like me to test it

mantid-roman:
You can test it now.

AnthonyLim23:
I have had some problems with the MaxEnt. I have been having difficulty in getting it to work on MUSR 62642 - long (load into muon analysis to get it). 

In most cases I get the error message "no convergence in alpha chop" and the only way I have found to fix this is to set chi to 1. However, in the old version it happily chugs along  and gets what looks to be a reasonable answer

AnthonyLim23:
The changes do give a good result with the new defaults. However, changing the A value back to 0.4 causes the results to be very poor. @Anders-Markvardsen do you think it is sensible to change the A value in this way?

mantid-roman:
The result depends on the values of `A` and `chiTarget`. What is special about value 0.4?

AnthonyLim23:
I have some recollection of reading that 0.4 was a good default value if you do not know much about the system. I thought it was in Anders' thesis but I cannot find it now. I may be mistaken about the whole thing, but Anders will know

AnthonyLim23:
I have had a play around and A=0.4 and chi target = 10000 gives a reasonable answer. Could we make it clearer that the default chi target goes to the number of points (without the user having to read the documentation)?

Anders-Markvardsen:
The value of A can critically affect the reconstructions. However, not in all cases. E.g. if the 'zero' time or zero fourier coef is known the reconstruction is not sensitive to this value. In my thesis through paintaking analysis I provided a recipe for automatically select a 'best' value for this parameter when this matter (for posneg image case) but this has not been implemented (yet) with this version of MaxEnt. 

The value of 0.4 I am guessing Rachel, from tried and error, found worked well for the muon dataset she tested with this algorithm. @mantid-roman the value you have chosen is 0.001 is very different and likely to break any muon scripts using the algorithms.  Hence I will revert this back to 0.4. However, I agree that 0.4 seems odd without an explanation, hence I agree with Anthony that it would be good to make it clearer where this number comes from. I will do this in this pull request. Note, this value should not be changed before checking with the MUON group who are currently the main user of this algorithm.  

About chi-squared then the standard in the literature is to select it to be equal to about the number of data points. Selecting it to a very small value is the equivalent of ignoring the errors in the data. chi-squared = 100 may not always be perfect but it is in general better than 1.0e-12. I will revert it back to 100 and add some text explaning this choice better. The algorithm could be 'extended/modified' so that it default chi-squared = number of data points. Or equivalent chi-squared could be changed chi-squared divided by number of data point and this quantity default to 1.0.

AnthonyLim23:
the chi squared does not use a small value, in the code if it equal or less than the current default value then the chi squared is replaced with the number of data points. My original point was this was unclear

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -119,10 +129,18 @@ void MaxEnt::init() {
...
+                  "A maximum entropy constant. This algorithm was first developed for the "
+                  "ISIS muon group where the default 0.4 was found to give good reconstructions."
+                  "In general the user will need to experiment with this value. Chosen a "
+                  "small value may lead to unphysical spiky reconstructions and a too large "
+                  "value the reconstruction will start to resamble that of a direct fourier "
+                  "transform reconstruction. However, where data contain a strongly determined zero "
+                  "Fourier data point the reconstruction will be insensitive to the choice "
+                  "of this property.");
+
+  declareProperty(make_unique<PropertyWithValue<double>>("ChiTarget", 1.0e-12,
```
Anders-Markvardsen: but @AnthonyLim23 the default has here been changed from 100 to 1.0e-12

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -228,6 +246,13 @@ void MaxEnt::exec() {
   // Number of X bins
   const size_t npointsX = inWS->isHistogramData() ? npoints + 1 : npoints;
 
+  // Default target chi squared = the number of data points:
+  // it is what is expected from a good fit with a smooth function
+  // (given that the error values are correct)
+  if (chiTarget <= 1.0e-12) {
```
AnthonyLim23: @Anders-Markvardsen  here the default value is replaced with N points. I still think this is unclear for a user, but it does happen

Anders-Markvardsen:
just spoken to roman and he will change the the expression of chi-squared to be divided by number of data point, change the default to 1.0 and the doc accordingly. this will make it less confusing we think for users to understand this property for this alg

AnthonyLim23:
that sounds very sensible

AnthonyLim23:
the two examples that I have from James gave the same results in the frequency domain as using PaddingAndApodization and then FFT. So looks good to me.

mantid-roman:
Replaced by https://github.com/mantidproject/mantid/pull/21079

PR #20922, opened 2017-10-16
Indirect - Elwin - Add check in case of invalid value within IPF file

### LINKED ISSUE #20858: Add more robust checks around reading IPF elements in IDA
On Friday we found that an instrument parameter file with a misspelled XML element was causing Mantid to crash unexpectedly. We corrected this in #20856, but we should make the code more robust as well.

### Expected behavior
Mantid should not crash. 

### Actual behavior
Mantid crashes because it expects the element to exist and be correctly spelled. If this is not the case it gets a `nullptr` and falls over at line 291 in Elwin.cpp.

https://github.com/mantidproject/mantid/blob/21bccfe66a959f3a56e6b0f6860fed37eab64dd9/qt/scientific_interfaces/Indirect/Elwin.cpp#L286-L291

See #20856 

### Platforms affected
All.

brandonhewer:
In the Elwin interface (Elwin.cpp), it is possible to crash Mantid, when there is an attempt to load an IPF file which contains an invalid value.

This PR adds a check for when this condition occurs and provides a warning message to the user.

**To test:**
1. Add the \*\_Parameters file to your instrument parameters directory.
2. Navigate to Interfaces > Indirect > Data Analysis Elwin.
3. Supply the \*\_red file as the input.
4. Run the Elwin interface and ensure Mantid does not crash and an error message is displayed.

**Test Files:**
[elwin-ipf-test-files.zip](https://github.com/mantidproject/mantid/files/1422106/elwin-ipf-test-files.zip)

Fixes #20858. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/qt/scientific_interfaces/Indirect/Elwin.cpp
+++ b/qt/scientific_interfaces/Indirect/Elwin.cpp
@@ -293,7 +293,7 @@ void Elwin::setDefaultResolution(Mantid::API::MatrixWorkspace_const_sptr ws,
       auto params = comp->getNumberParameter("resolution", true);
 
       // set the default instrument resolution
-      if (params.size() > 0) {
+      if (!params.empty()) {
         double res = params[0];
         m_dblManager->setValue(m_properties["IntegrationStart"], -res);
         m_dblManager->setValue(m_properties["IntegrationEnd"], res);
--- a/qt/scientific_interfaces/Indirect/Elwin.cpp
+++ b/qt/scientific_interfaces/Indirect/Elwin.cpp
@@ -288,19 +288,24 @@ void Elwin::setDefaultResolution(Mantid::API::MatrixWorkspace_const_sptr ws,
 
   if (analyser.size() > 0) {
     auto comp = inst->getComponentByName(analyser[0]);
-    auto params = comp->getNumberParameter("resolution", true);
 
-    // set the default instrument resolution
-    if (params.size() > 0) {
-      double res = params[0];
-      m_dblManager->setValue(m_properties["IntegrationStart"], -res);
-      m_dblManager->setValue(m_properties["IntegrationEnd"], res);
+    if(comp) {
+      auto params = comp->getNumberParameter("resolution", true);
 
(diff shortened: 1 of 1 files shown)


qt/scientific_interfaces/Indirect/Elwin.cpp
```diff
@@ -288,19 +288,25 @@ void Elwin::setDefaultResolution(Mantid::API::MatrixWorkspace_const_sptr ws,
...
-      double res = params[0];
-      m_dblManager->setValue(m_properties["IntegrationStart"], -res);
-      m_dblManager->setValue(m_properties["IntegrationEnd"], res);
+    if (comp) {
+      auto params = comp->getNumberParameter("resolution", true);
 
-      m_dblManager->setValue(m_properties["BackgroundStart"], -10 * res);
-      m_dblManager->setValue(m_properties["BackgroundEnd"], -9 * res);
+      // set the default instrument resolution
+      if (params.size() > 0) {
```
soininen: Please use `if (!params.empty())` instead. It conveys the intention better.

soininen (changes requested):

soininen:
Could there be a test file attached to this pull request so it could be tested?

soininen:
@brandonhewer I couldn't make vanilla Mantid 'master' to crash with the test files provided. The interface seems to work just fine. What am I supposed to see when these changes are not applied?

NickDraper:
@brandonhewer This appears to be stuck with you

brandonhewer:
This PR adds a simple check, currently I've been unable to reproduce the issue with the file I supplied.
This PR shouldn't break any existing code (please see the diff) and it may be best to check it works with the supplied test.

Anders-Markvardsen (approved):
The code adds an extra null pointer check. Bug can't been reproduced from this branch and previous test suggest can't easily do this with older version, but this extra change provides an extra safety and therefore approve

soininen (approved):

PR #20927, opened 2017-10-16
Re #19126: Fixed LoadDialog crash by disabling ok button while updating dialog.

### LINKED ISSUE #19126: Exception/crash when `Load Dialog` is updating itself
An unexpected exception/crash can occur if the user tries to load a file while the `Load Dialog` is still in the process of automatically updating the file path and/or `OutputWorkspace` name from an input search string such as "OFFSPEC39347".

This was found while working on #18829 

### Expected behavior
Mantid should not crash or throw an unexpected exception

### Actual behavior
Mantid crashes or throws an unexpected exception

- Go to `Load -> File`
- In the `Load File` dialog, type `OFFSPEC39347` in the `File` box.
- Press `Tab` twice followed by `Enter`.

Mantid will usually catch an unexpected exception, or sometimes crash. It depends a bit on the timing. I think you need to have hit Enter (or pressed the `Run` button) before the `OutputWorkspace` box has been updated. The crash seems to happen if you hit Enter more than once and the exception if you hit it only once.

### Platforms affected
Maybe all. 
- Seen on Windows 7 and 10 with mantid v3.9.0, v3.9.1 and v3.9.1-1537-gde3198. 
- Seen on Fedora 26 with branch based off master-69917f1683.

edwardb96:
Disabled signals for the LoadDialog ok button while updating the ui.

Description of work.

**To test:**
1. Follow the instructions from the original issue and ensure that Mantid no longer crashes.
2. Ensure that the Load dialog still works as expected.
3. Code Review.

Fixes #19126 . 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/qt/widgets/plugins/algorithm_dialogs/src/LoadDialog.cpp
+++ b/qt/widgets/plugins/algorithm_dialogs/src/LoadDialog.cpp
@@ -35,6 +35,20 @@ struct HoldFlag {
 };
 }
 
+// Blocks signals sent to the load button for the duration
+// of it's lifetime.
+class PreventLoadRequests {
+public:
+  PreventLoadRequests(LoadDialog &dialog) : m_dialog(dialog) {
+    m_dialog.disableLoadRequests();
+  }
+
+  ~PreventLoadRequests() { m_dialog.enableLoadRequests(); }
+
+private:
+  LoadDialog &m_dialog;
+};
+
 // Declare the dialog. Name must match the class name
 DECLARE_DIALOG(LoadDialog)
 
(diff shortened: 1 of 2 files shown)


edwardb96:
I am not sure how (if) this interacts with #20954

qt/widgets/plugins/algorithm_dialogs/src/LoadDialog.cpp
```diff
@@ -111,29 +111,32 @@ void LoadDialog::enableNameSuggestion(const bool on) {
...
  * Called when the run button is clicked
  */
 void LoadDialog::accept() {
   // The file widget may have been edited but not lost focus so that the search
-  // wasn't
-  // attempted for the new contents. Force one here.
+  // wasn't attempted for the new contents. Force one here.
   // The widget does nothing if the contents have not changed so it will be
   // quick for this case
+  disableLoadRequests();
```
martyngigg: I'm wondering if something that might throw an exception in the middle of this and `enable` could leave the button disabled? For memory allocation we'd use the RAII pattern, how about using the same thing here to ensure that any route out leaves the button enabled again?

martyngigg (approved):
This seems much more reliable now.

Nice :+1:

PR #20939, opened 2017-10-17
Adjust SaveDiffFittingAscii to save TOFTOF workspaces

### LINKED ISSUE #20859: Implement option SaveAscii in TOFTOF GUI
The option SaveAscii in the TOFTOF GUI should save the S(Q,W) in the following format:
```
W  Q1  S(Q1,W) Err1  Q2  S(Q2,W)  Err2 ... Qn  S(Qn,W) Err_n
```
This format is very similar to one produced by `SaveDiffFittingAscii` algorithm. Therefore, to my opinion, the implementation of this option should be done in a few steps:

- [x] Adjust `SaveDiffFittingAscii` to save TOFTOF Ascii. Following needs to be done (PR #20939):
   - a new line is needed between the header and a data rows
   - Properties `RunNumber` and `Bank` should be optional. If they are not specified, they don't present in the output file.
- [ ] Create `ConvertToTableWorkspace(InputWorkspace, OutputWorkspace, Layout)` algorithm which would create a table workspace to save it using the `SaveDiffFittingAscii` algorithm. 
- [ ] Add functionality to save Ascii to the TOFTOF GUI.

I think, this approach is better than a new SaveTOFTOFAscii algorithm, but I am open for discussion.

milestone 3.12
I plan to work on it.

mganeva:
This PR makes properties `RunNumber` and `Bank` optional. Now there are possible options:
1. both `RunNumber` and `Bank` are empty. In this case they are not present in the saved file
2. both `RunNumber` and `Bank` are set to some value. In this case the algorithm should work the same way as before.

**To test:**
I did not find any unit tests for this algorithm. One can test it with a simple script:
```python
# create a table workspace
table = CreateEmptyTableWorkspace()
table.addColumn(type='double', name="DeltaE")
for i in range(10):
    table.addColumn(type='double', name="Q_{}".format(i))
    table.addColumn(type='double', name="S(Q_{},W)".format(i))
    table.addColumn(type='double', name="S(Q_{},W)Error".format(i))
# add some numbers to the table    
for i in range(10):
    row = [i]
    for j in range(10):
        row.append(0.1*j + 0.1)
        row.append(100.0+j)
        row.append(10 + j)
    table.addRow(row)
    
# save to ascii
# save without run number and bank
SaveDiffFittingAscii(table, Filename="test.txt", OutMode='OverwriteFile')
# save with run number and bank
SaveDiffFittingAscii(table, Filename="test1.txt", Bank="0", RunNumber="0", OutMode='OverwriteFile')
# this will produce an error, because settings are non consistent
SaveDiffFittingAscii(table, Filename="test2.txt", RunNumber="0", OutMode='OverwriteFile')
SaveDiffFittingAscii(table, Filename="test3.txt", Bank="0", OutMode='OverwriteFile')

```

Fixes #20859 p.1 . 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/DataHandling/src/SaveDiffFittingAscii.cpp
+++ b/Framework/DataHandling/src/SaveDiffFittingAscii.cpp
@@ -189,8 +189,8 @@ void SaveDiffFittingAscii::writeInfo(const std::string &runNumber,
 void SaveDiffFittingAscii::writeHeader(
     const std::vector<std::string> &columnHeadings, std::ofstream &file) {
   for (const auto &heading : columnHeadings) {
-    // Chi being the last header in the table workspace
-    if (heading == "Chi") {
+    // if last header in the table workspace, put eol
+    if (&heading == &columnHeadings.back()) {
       writeVal(heading, file, true);
     } else {
       writeVal(heading, file, false);
--- a/Framework/DataHandling/src/SaveDiffFittingAscii.cpp
+++ b/Framework/DataHandling/src/SaveDiffFittingAscii.cpp
@@ -153,7 +153,7 @@ void SaveDiffFittingAscii::processAll(
     std::string bank = splitBank[m_counter];
 
     if (!runNum.empty() || !bank.empty())
-        writeInfo(runNum, bank, file);
+      writeInfo(runNum, bank, file);
 
     // write header
     std::vector<std::string> columnHeadings = input_ws[i]->getColumnNames();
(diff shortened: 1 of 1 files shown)


Framework/DataHandling/src/SaveDiffFittingAscii.cpp
```diff
@@ -274,11 +276,11 @@ std::map<std::string, std::string> SaveDiffFittingAscii::validateInputs() {
   std::string bankNumber = getPropertyValue("Bank");
   std::vector<std::string> splitBank = splitList(bankNumber);
 
-  if (runNumber.empty()) {
-    errors["Bank"] = "Please provide a valid bank list";
-  }
-  if (splitBank.empty()) {
-    errors["Bank"] = "Please provide a valid bank list";
+  if (bankNumber.empty()) {
```
peterfpeterson: This cross-check of parameters belongs in a [`validateInputs`](https://github.com/mantidproject/mantid/blob/master/Framework/API/inc/MantidAPI/IAlgorithm.h#L113) method. That way it is run *before* the `exec` starts and can annotate the invalid parameters in the algorithm dialog.

peterfpeterson (changes requested):

peterfpeterson (approved):
I didn't see that the code was already in the right place.

PR #20942, opened 2017-10-17
Integrate parallel event loader

### LINKED ISSUE #20887: Integrate parallel event loader components
Loader and parser components have been provided by previous PRs. Now integrate those with `LoadEventNexus`.

Part of #20344.

SimonHeybrock:
Put together some previously implemented pieces into a parallel event loader used in `LoadEventNexus`.

This adds a new parallel loader for events in `LoadEventNexus`. It is always disabled for non-MPI builds, i.e., does not influence anyone since we are not providing MPI builds yet.

**To test:**

Core review.

Fixes #20887. 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Parallel/test/PulseTimeGeneratorTest.h
```diff
@@ -0,0 +1,118 @@
...
+    PulseTimeGenerator pulseTimes({0, 2, 2, 3}, {4, 8, 12, 16}, 1000);
+    pulseTimes.seek(1);
+    TS_ASSERT_EQUALS(pulseTimes.next().totalNanoseconds(), 1004);
+    TS_ASSERT_EQUALS(pulseTimes.next().totalNanoseconds(), 1012);
+    pulseTimes.seek(1);
+    TS_ASSERT_EQUALS(pulseTimes.next().totalNanoseconds(), 1004);
+    TS_ASSERT_EQUALS(pulseTimes.next().totalNanoseconds(), 1012);
+  }
+};
+
```
OwenArnold: 👍 for thorough testing

Framework/Types/inc/MantidTypes/Core/DateAndTime.h
```diff
@@ -147,10 +147,66 @@ class MANTID_TYPES_DLL DateAndTime {
 private:
   /// A signed 64-bit int of the # of nanoseconds since Jan 1, 1990.
   int64_t _nanoseconds;
+
+  /// Max allowed nanoseconds in the time; 2^62-1
+  static constexpr int64_t MAX_NANOSECONDS = 4611686018427387903LL;
```
OwenArnold: This can't be done with `std::numeric_limits::max` presumably? not sure why this isn't 2^63-1?

OwenArnold (changes requested):
Looking good. A few suggestions and questions.

Framework/Types/inc/MantidTypes/Core/DateAndTime.h
SimonHeybrock: This is unrelated to this PR (just moved code). The reason seems to be to allow addition that does not overflow.

Framework/DataHandling/src/LoadEventNexus.cpp
```diff
@@ -855,13 +857,28 @@ void LoadEventNexus::loadEvents(API::Progress *const prog,
...
+  if (canUseParallelLoader(haveWeights, oldNeXusFileNames, classType)) {
+    auto ws = m_ws->getSingleHeldWorkspace();
+    m_file->close();
+    try {
+      ParallelEventLoader::load(*ws, m_filename, m_top_entry_name, bankNames);
+      loaded = true;
+      shortest_tof = 0.0;
+      longest_tof = 1e10;
+    } catch (const std::runtime_error &) {
+    }
```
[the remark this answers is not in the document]
SimonHeybrock: Yes, it will fall back to the default loader. In fact I will need to make more changes such that the parallel loader is never used by default, unless enabled explicitly.

Framework/DataHandling/src/LoadEventNexus.cpp
```diff
@@ -1646,6 +1663,31 @@ void LoadEventNexus::safeOpenFile(const std::string fname) {
   }
 }
 
+bool LoadEventNexus::canUseParallelLoader(const bool haveWeights,
+                                          const bool oldNeXusFileNames,
+                                          const std::string &classType) const {
+  if (m_ws->nPeriods() != 1)
```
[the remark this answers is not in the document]
SimonHeybrock: Some of them would require changes on how the loader is working internally. The plan is to add this functionality later of required as high level branching in the loader, if required (the default loader has a lot of these checks on a low level, leading to performance deterioration). Currently all special cases are disallowed to make implementation easier. Some will probably never be supported.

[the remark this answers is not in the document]
SimonHeybrock: Regarding multiperiod files: I simply do not know how these are working and if the comply with the Nexus standard. I do not know if we should support/work on anything non-standard at this point.

Framework/DataHandling/src/ParallelEventLoader.cpp
```diff
@@ -0,0 +1,69 @@
...
+  const auto instrument = ws.getInstrument();
+  const auto &compInfo = ws.componentInfo();
+  const auto &detInfo = ws.detectorInfo();
+  const auto &detIDs = detInfo.detectorIDs();
+
+  // Monitors are not loaded by LoadEventNexus, so we have to exclude them when
+  // computing an offset based on detector IDs. Currently this is computed in a
+  // naive way and works only if all monitors have IDs smaller than any
+  // detector.
+  const auto monitors = instrument->getMonitors();
```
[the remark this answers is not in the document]
SimonHeybrock: You can iterate `DetectorInfo`, I'll see if that looks reasonable.

Framework/DataHandling/src/ParallelEventLoader.cpp
```diff
@@ -0,0 +1,69 @@
...
+  int32_t monitorOffset = static_cast<int32_t>(monitors.size());
+  for (size_t i = 0; i < monitors.size(); ++i)
+    if (monitors[i] != detIDs[i])
+      throw std::runtime_error(
+          "Monitors are not corresponding to the first detector IDs in the "
+          "instrument. This is currently not supported by ParallelEventLoader");
+
+  std::vector<int32_t> bankOffsets;
+  for (const auto &bankName : bankNames) {
+    // Removing "_events" from bankName
```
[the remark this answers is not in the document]
SimonHeybrock: Not by the standard as far as I know, but I had trouble thinking of another way to associate an `NXevent_data` group in the Nexus file with a bank in a Mantid instrument. Is there?

Framework/DataHandling/src/LoadEventNexus.cpp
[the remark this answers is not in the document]
OwenArnold: Ok that makes sense

Framework/DataHandling/src/ParallelEventLoader.cpp
```diff
@@ -20,13 +20,20 @@ std::vector<int32_t> bankOffsets(const API::ExperimentInfo &ws,
...
+    if (detInfo.isMonitor(i)) {
+      if (sawDetector)
+        throw std::runtime_error("Monitors are not corresponding to the first "
+                                 "detector IDs in the instrument. This is "
+                                 "currently not supported by "
+                                 "ParallelEventLoader");
+      ++monitorOffset;
+    } else {
+      sawDetector = true;
+    }
```
OwenArnold: Reads better

OwenArnold (commented):
Changes look good.

OwenArnold (approved):

Framework/Parallel/inc/MantidParallel/IO/PulseTimeGenerator.h
```diff
@@ -0,0 +1,99 @@
...
+    while (m_pulse < m_index.size() - 1 && m_event == m_index[m_pulse + 1]) {
+      ++m_pulse;
+      m_pulseTime = getPulseTime(m_timeZeroOffset, m_timeZero[m_pulse]);
+    }
+    ++m_event;
+    return m_pulseTime;
+  }
+
+private:
+  int64_t cast(const uint64_t time) const { return static_cast<int64_t>(time); }
```
LamarMoore: why not 
```cpp
template <typename T1, typename T2> 
T1 cast (const T2 time) const {return static_cast<T1>(time); }
```
also don't understand the cast from double to double in the last case.

SimonHeybrock: The cast of `double` is not necessary of course. Your suggested template would not work because then I would need to explicitly specify `T1` in `getPulseTime`, i.e., I would need to specialize it for integer and floating points types?

Framework/Parallel/inc/MantidParallel/IO/PulseTimeGenerator.h
```diff
@@ -0,0 +1,99 @@
...
+  int64_t cast(const int64_t time) const { return static_cast<int64_t>(time); }
+  int64_t cast(const uint32_t time) const { return static_cast<int64_t>(time); }
+  int64_t cast(const int32_t time) const { return static_cast<int64_t>(time); }
+  double cast(const double time) const { return static_cast<double>(time); }
```
SimonHeybrock: I think this is actually not what we want. If `time` is a floating point type we need to read its unit from the Nexus file (could be seconds or microseconds, etc., I think).

SimonHeybrock: Same goes also for integer types. Will be solved in https://github.com/mantidproject/mantid/issues/21015.

LamarMoore: 👍

PR #20952, opened 2017-10-18
Fix excluded ranges in unweighted fitting

### LINKED ISSUE #20949: Fit with unweighted least squares doesn't respect excluded ranges
`FitMW` sets the weights of excluded data points to zero but these get overwritten by `CostFuncUnweightedLeastSquares` which forces a weight of 1 for all points.

soininen:
This PR fixes an issues where setting the `CostFunction` property to 'Unweighted least squares' would lead discarding the `Exclude` property.

**To test**

The following script should print something close to zero as the only non-zero number is excluded from the unweighted fit:
```python
import numpy

ys = numpy.array([0, 0, 0, 10, 0])
xs = numpy.array([0, 1, 2, 3, 4, 5])
ws = CreateWorkspace(DataX=xs, DataY=ys)
result = Fit('name=FlatBackground, A0=0.1', ws,
             Exclude=[3, 4],
             Minimizer='Levenberg-MarquardtMD',
             CostFunction='Unweighted least squares')
print(result.Function.A0)
```

Fixes #20949.

**Release Notes** 

The bug fix is mentioned in the Framework release notes.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


josephframsay (approved):
Nice changes, and the cost function now behaves as expected. Looks good to me!

Issue #15188, opened 2016-02-05
Pixel segregation issue in ConvertToReflectometryQ


raquelalvarezbanos:
Jos sent the following script:

```
dref=s.OurLoad(r[36688])
dtrans=s.OurLoad(r[36689])

Trans=CreateTransmissionWorkspaceAuto(dtrans,AnalysisMode="MultiDetectorAnalysis",ProcessingInstructions="380-428",WavelengthMin=2.0)

IvsQ2, IvsLam2, thetaOut=ReflectometryReductionOneAuto(dref,AnalysisMode="MultiDetectorAnalysis",ProcessingInstructions="390-450", FirstTransmissionRun=Trans, ThetaIn="0.6",WavelengthMin=2.0,CorrectionAlgorithm='None',StrictSpectrumChecking=False)#,NormalizeByIntegratedMonitors=False
IvsQ, IvsLam, thetaOut=ReflectometryReductionOneAuto(dref,AnalysisMode="MultiDetectorAnalysis",ProcessingInstructions=','.join(map(str, range(250, 557))), FirstTransmissionRun=Trans, ThetaIn="0.6",WavelengthMin=2.0,CorrectionAlgorithm='None',StrictSpectrumChecking=False)#,NormalizeByIntegratedMonitors=False

TvL = ConvertSpectrumAxis(IvsLam,'SignedTheta')

ConvertToReflectometryQ(TvL,IncidentTheta=0.6 ,DumpVertexes=True,OverrideIncidentTheta=True, OutputWorkspace="Qmap",OutputVertexes="Qtable",Extents=[-5e-4,2e-4,0.008,0.1],NumberBinsQz=500,NumberBinsQx=500,Method='NormalisedPolygon',OutputAsMDWorkspace=False)
```

which produces a `QxQz` map with lines (segments) of dark intensity (we should obtain a smooth color map):

![pixelsegregationissue](https://cloud.githubusercontent.com/assets/8956985/12843564/1e777608-cbf2-11e5-83c0-ae3ae8edb3de.png)

raquelalvarezbanos:
I've modified the script above to work with a smaller number of bins:

```
[...]

PR #20953, opened 2017-10-18
Correct fractional binning in 2D workspaces (RebinnedOutput)

### LINKED ISSUE #20599: Procedure for error calculations used by Rebin2D is incorrect.
### Expected behavior
The code:
```Python

Load(Filename=r'MAR22152_Ei15.00meV.nxspe', OutputWorkspace='Yb2_Ei15.00meV')

# # To generate 1D files for Yb2 data after background subtraction
SofQWNormalisedPolygon(InputWorkspace='Yb2_Ei15.00meV', OutputWorkspace='Yb2_Ei15.00meV_subt_SQW_coarse', QAxisBinning='0,0.001,5', EMode='Direct')
#SofQWCentre(InputWorkspace='Yb2_Ei15.00meV_subt', OutputWorkspace='Yb2_Ei15.00meV_subt_SQW_coarse', QAxisBinning='0,0.001,5', EMode='Direct')
SofQWNormalisedPolygon(InputWorkspace='Yb2_Ei15.00meV', OutputWorkspace='Yb2_Ei15.00meV_subt_SQW_fine', QAxisBinning='0,0.01,5', EMode='Direct')
#SofQWCentre(InputWorkspace='Yb2_Ei15.00meV_subt', OutputWorkspace='Yb2_Ei15.00meV_subt_SQW_coarse', QAxisBinning='0,0.001,5', EMode='Direct')

Rebin2D(InputWorkspace='Yb2_Ei15.00meV_subt_SQW_coarse', OutputWorkspace='Yb2_Ei15.00meV_bcksubt_1D_coarseQ', Axis1Binning='-7,0.02,15', Axis2Binning='0,3,3', UseFractionalArea=True)
Rebin2D(InputWorkspace='Yb2_Ei15.00meV_subt_SQW_fine', OutputWorkspace='Yb2_Ei15.00meV_bcksubt_1D_fineQ', Axis1Binning='-7,0.02,15', Axis2Binning='0,3,3', UseFractionalArea=True)

```
calculates signal and error along the trajectory, chosen by user, with different accuracy. 
The data file used by the script can be downloaded from `ftp://ftp.nd.rl.ac.uk/scratch/abuts/Ticket_20599/MAR22152_Ei15.00meV.nxspe`.

One would expect the signal and error calculated by the script to be independent on the resolution

### Actual behaviour

While the signal does not depend on the resolution when fractional area is selected, the error does. In fact, by changing resolution one can get almost any value of the error.  In addition to that, the calculated error is substantially different from the one calculated by reference code (Mslice)
error_bars_problem

See above
### Platforms affected
any.

mducle:
The handling of errors in workspaces with fractional bin weights (class `RebinnedOutput`) was found to be incorrect in #20599. This resolves that issue and also corrects code elsewhere which handled `RebinnedOutput` incorrectly (see below for details).

The major changes is in the code for `RebinnedOutput` and specifically the `rebinToFractionalOutput` method of `FractionalRebinning`. This affects three algorithms: `SofQWNormalisedPolygons` (where the bug was identified), `Rebin2D` and `ConvertToReflectometryQ` (using `ReflectometryTransform`). Users of these algorithms (principally powder inelastic neutron scattering and reflectometry) should check that they are still getting reasonable numbers out. (All the tests pass, but possibly not enough tests have been put in to cover all use cases).

In addition, the code for handling `RebinnedOutput` workspaces in the `Integration` algorithm has been corrected (it cast `RebinnedOutput` to `Workspace2D` and ignored the fractional bin weights), and that in `SumSpectra` has been verified to be correct. 

---

**Theory**

`rebinToFractionalOutput` works by taking an input quadrilateral (`inputQ`) and rebinning it into an output grid by sharing its count and errors amongst the new bins. First consider the case where `inputQ` is bigger than the output grid:

text4172

For each output bin (regular grid squares), `rebinToFractionalOutput` calculates the overlap area `IntersectOverlap` of that bin with the input quadrilateral, and uses the fractional overlap area as a weight for the counts and errors. In the original (incorrect) code, it calculates for each bin:

```
signal_in_bin = signal_in_bin + ( signal_in_inputQ * weight )
variance_in_bin = variance_in_bin + ( error_in_inputQ * weight )**2
```

where `error` is the standard deviation and `variance=error**2`. Whilst the handling of the signal is correct, the handling of the errors is not, because the weights should be not be squared. To see this consider rebinning the data from the new grid back (regular squares) into the old grid. The error of the re-rebinned `inputQ` would then be:

```
new_variance_in_inputQ = sum_over_i ( weight_i**2 ) * error_in_inputQ**2
```

However, the sum over the squares of the *fractional* weights is **smaller** than unity - so successive rebinning from one grid to the other would yield progressively smaller errors. In fact the correct transformations should be:

```
signal_in_bin = signal_in_bin + ( signal_in_inputQ * weight )
variance_in_bin = variance_in_bin + ( error_in_inputQ**2 * weight )
```

e.g. it should be the variances and not the errors which should be weighted by the fractional area. Now the sum is over just the weights which will be unity.

Another problem with the implementation of `rebinToFractionalOutput` is that it does not "carry over the weights" when applied to a workspace which already has fractional bin weights (it treats all input workspaces as having unit weights). This means that if we were to rebin the previous workspace again into another grid with a different bin size which is does not fully cover the area of `inputQ`, e.g.:

path4260

then the `weight` which is used to scale the `signal` and `variance` will be wrong. In the above diagram, in the original code, the `weight` would be the ratio of the area of the whole of the blue square to the area of the green rectangle. The reweighted signal and error would then be an overestimate because the original data only covers the smaller area where the blue square and red quadrilateral intersects. Thus the correct weight should be:

```
new_weight = old_weight * intersectOverlap / inputQ
```

The final point is that in order to make the data appear sensible (accounting for the possibility that bin weights may vary by a lot in nearby bins which would otherwise make the signal/error appear to vary by alot between adjacent bins), the data in the workspace is `finalized` by dividing by the weights at the end of most algorithms. In this case, because the data is just scaled, the code correctly divides the signal and *error* (not variance) by the weights (it's not added in quadrature). However, for subsequent processing (e.g. in `Rebin2D` after `SofQWNormalisedPolygon`), the data should be `unfinalized` or treated in a different way which accounts for the fact that it has been scaled by the weights. This was not done in `Rebin2D` or `Integration`.

---

**To test:**

Use the code in issue #20599 and check that the rebinned outputs for both coarse and fine Q bin steps now agree exactly in signal and errors. 

Run a standard inelastic neutron powder reduction (e.g. from #15953) and check that everything still works as usual and the data appears ok.

Run a standard reflectometry reduction (e.g. from #15188 and its PR) and check that everything still looks ok.

Fixes #20599. 

**Release Notes** 

To follow.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/DataObjects/src/RebinnedOutput.cpp
```diff
@@ -132,6 +137,35 @@ void RebinnedOutput::finalize(bool hasSqrdErrs) {
...
+ * This function "unfinalizes" the workspace by taking the data/error arrays
+ * and multiplying them by the corresponding fractional area array.
+ * @param hasSqrdErrs :: does the workspace have squared errors?
+ */
+void RebinnedOutput::unfinalize(bool hasSqrdErrs) {
+  if (!m_finalized) {
+    return;
+  }
+  std::size_t nHist = this->getNumberHistograms();
+  for (std::size_t i = 0; i < nHist; ++i) {
```
martyngigg: Can we do this in parallel?

Framework/DataObjects/inc/MantidDataObjects/RebinnedOutput.h
```diff
@@ -62,6 +62,11 @@ class DLLExport RebinnedOutput : public Workspace2D {
 
   /// Create final representation
   void finalize(bool hasSqrdErrs = true);
+  void unfinalize(bool hasSqrdErrs = false);
+
+  /// Returns if finalize has been called, or sets this flag
+  bool isFinalized() const { return m_finalized; }
+  void setFinalized(bool isFinal) { m_finalized = isFinal; }
```
martyngigg: When would it be necessary to call this function? It feels like it could be error prone to remember we have to call it.

mducle: Yeah... I don't quite like this very much either. 

The problem is for algorithms which act on a `RebinnedOutput` but which create a blank new output workspace (which would have `m_finalized = false`) and then populate the elements of this output workspace with the transformed value of the input workspace without calling `finalize()` or `unfinalize()`. The ones I've found are `Transpose` and `UnaryOperation`. So, if you call `SofQW3` on something then call `Transpose` on the output and then call `Rebin2D` on the output of that, `Rebin2D` will think the data has not been "finalized" and will then handle the data incorrectly.

I can't really think of some way to get around this though.

Framework/DataObjects/src/RebinnedOutput.cpp
mducle: I don't know - is `std::transform` thread safe. Someone on SO implies that it is not, but I can't find anything more concrete...

martyngigg: As long as it is treating the histograms independently it will be safe.

Framework/DataObjects/inc/MantidDataObjects/RebinnedOutput.h
martyngigg: So the real problem here is setting the flag appropriately on construction from a `RebinnedOutput` workspace?

mducle: Yes - essentially we need an "empty" copy constructor... it's a bit specialized. Actually, for `UnaryOperation` a normal copy constructor should work because it does the transformation in-place.

Framework/DataObjects/src/RebinnedOutput.cpp
```diff
@@ -132,6 +137,35 @@ void RebinnedOutput::finalize(bool hasSqrdErrs) {
               std::ostream_iterator<double>(g_log.debug(), " "));
     g_log.debug() << '\n';
```
mducle: @martyngigg Should I also parallelise this function in `RebinnedOutput`? I think I would need to remove all the logging in order to parallelise it - but it seems like we don't really need the debug information any more anyway.

martyngigg: @mducle I think I would prefer to remove the logging at this level. I think anyone looking to debug it in the future would be perfectly capable of putting the prints back in temporarily. This should improve the speed to as the `std::copy` will happen regardless of the log level - it will just not be displayed!

Framework/DataObjects/inc/MantidDataObjects/RebinnedOutput.h
martyngigg: @mducle We do have a `cloneEmpty()` method on workspaces. Will that not do the job?

mducle: @martyngigg Yes. I'll look into changing the algorithms to use `cloneEmpty()` if they don't already do.

Framework/Algorithms/src/Integration.cpp
```diff
@@ -242,10 +246,20 @@ void Integration::exec() {
...
+    double Fmax = 0.0;
     if (distmax <= distmin) {
       sumY = 0.;
       sumE = 0.;
     } else {
+      if (rebinned_input) {
+        // Workspace has fractional area information, need to take into account
+        const MantidVec &F = rebinned_input->readF(i);
+        sumF = std::accumulate(F.begin() + distmin, F.begin() + distmax, 0.0);
+        Fmin = F[distmin - 1];
```
martyngigg: If no limits are set I think `distmin=0` doesn't it? This line would then be invalid

martyngigg (changes requested):

Framework/Algorithms/src/Integration.cpp
mducle: Yes. I'll add a check for `distmin > 0`

mducle:
@martyngigg 

Whilst looking at the MSlice issue https://github.com/mantidproject/mslice/issues/181 I found that the Rebin2D code was not also not working correctly:

```
Load(Filename='/home/vqq25957/src/mslice/MAR21335_Ei60.00meV.nxspe', OutputWorkspace='MAR21335_Ei60.00meV')
SofQWNormalisedPolygon(InputWorkspace='MAR21335_Ei60.00meV', OutputWorkspace='MAR21335_Ei60.00meV_sqw', QAxisBinning='0.28303,0.01246,11.0258', EMode='Direct')
Rebin2D(InputWorkspace='MAR21335_Ei60.00meV_sqw', OutputWorkspace='MAR21335_Ei60.00meV_line', Axis1Binning='-1.5,3,1.5', Axis2Binning='0,0.0125,10', UseFractionalArea=True, Transpose=True)
```

mslice_nan_bug3

This is because `Rebin2D::createOutputWorkspace` calls the factory `create<RebinnedOutput>` function which in turns calls `RebinnedOutput::cloneEmpty`. This copies the `m_finalized` flag from the parent to the child, meaning that the call to `RebinnedOutput::finalize()` in line 161 has no effect and the output data is not correctly weighted. 

To get around this, I've added a `force` flag to `RebinnedOutput::finalize`, rather than use `WorkspaceFactory::Instance().create` instead of `create<T>` which does not call `cloneEmpty()` since this seems to be quite complicated, and might break other things. Is this ok?

martyngigg:
Duc, I think this is an acceptable compromise. The only other option is a setter method to reset the flag but that feels worse really. At least the force flag can just be used when you know you need it to be because of the way the workspace has been created.

martyngigg:
This looks good to me now.

The error bars on the original issue are now fixed w.r.t to the bin sizes. I've run a few workflows that use the 2D binning and they look as expected to me.

martyngigg (dismissed):

peterfpeterson (changes requested):
You don't have release notes and I think the documentation for `Rebin2D` could benefit from your discussion in the PR, especially the nice images.

mducle:
@peterfpeterson I've added release notes, but instead of changing the `Rebin2D` documentation (`Rebin2D` should be used in the same way as before), I've put the detailed explanations into `SofQWNormalisedPolygon` which is the only algorithm which creates a `RebinnedOutput`, and in the `Workspace2D` concepts page. 

Also, whilst looking through this, I think that the weights are ignored by all the workspace operations (`UnaryOperation` and `BinaryOperation` effectively cast `RebinnedOutput` to `Workspace2D` - but this could be another issue and PR.

mducle:
**Please don't merge this pull request yet.**

After some discussions with TG Perring I think there is a mistake in the error propagation code which somehow gives consistent but possibly consistently wrong results for the errors.

mducle:
Decision made to not go forward with this solution, but to make changes to `SofQW` to allow it to rebin directly from the reduced data to the desired cuts. See #21500 for further information.

mducle:
Following discussions in #21688, it was decided that the handling of uncertainties in this PR is correct after all, so it has been resurrected.

martyngigg:
@mducle There seems to be a system test failure on one of the BASIS tests.

mducle:
@martyngigg Yes. It's code in `UnaryOperation.cpp` which I'm trying to debug now...

martyngigg (approved):

martyngigg:
@peterfpeterson I think your review here is from the original PR. Could you update that now?

martyngigg:
@peterfpeterson Could you say yes/no on this PR so we can get it moving, please?

peterfpeterson (approved):
Thanks for the documenation updates.

NickDraper:
@peterfpeterson Could you take another look at your review on this PR, it would be good to get this one into the release.  It was a big issue at the User meeting and the PMB

PR #20954, opened 2017-10-18
Prevent GUI thread locking using MWRunFiles

### LINKED ISSUE #20855: Bug with MWRunFiles and aysnc operation
The issue becomes apparent when using the `Engineering Diffraction` interface. 

The steps to reproduce this are:
* Open  `Engineering Diffraction` 
* Change the sample run number and close the GUI
* You should see it freeze

The issue is the `wait` call on `m_thread` in `MWRunFiles`. We tried to mitigate it with a Qt eventloop but this can create the scenario where the MWRunFile is being deleted while the `findRuns` method is still in progress.

samueljackson92:
This prevents a very long running file search from blocking the GUI event thread. It also gets what was the `FindFilesThread` under unit test.

Instead of the blocking approach we were using before, MWRunFiles now using a thread pool to create a worker that searches for files. Importantly, if a new search is initiated we will not block and wait for the thread to finish.

**To test:**
 - Make sure that data archive search is turned on.
 - The original bug that this was reported for is described in the issue. Please check that this no longer occurs.
 - Also check that opening and quickly closing the GUI does not cause it to hang or crash.

This widget has wide use in Mantid. The interfaces that currently use this widget are:
 - The Load Dialog
 - The Fit Dialog
 - ISIS Engineering Diffraction Interface
 - ISIS Muon Interface
 - ISIS Indirect Interfaces
 - ISIS SANS Interface
 - Anywhere that uses the `DataProcessor` widget (ISIS Reflectometry)

Fixes #20855 

**Release Notes** 

See [here](https://github.com/mantidproject/mantid/pull/20954/commits/30bc63610b8a50f51bedd6666d349806d6a14845).

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


NickDraper:
Sorry I'm getting a crash on starting the engineering GUI, it looks like a non gui thread it attempting to update a GUI element
image

NickDraper (changes requested):
There's a problem starting the Engineering GUI on windows

NickDraper:
Tests of the Engineering GUI and Indirect reduction (another interface using the widget, picked at random) seems fine.

No delay starting, or quiting the interface, runfiles lookup seems to operate as normal.

NickDraper (dismissed):

samueljackson92:
@NickDraper just realised that this appears to break the Load dialog.

martyngigg:
@samueljackson92 Does this mean further changes are required?

samueljackson92:
@martyngigg Yes I think so.

martyngigg (changes requested):
Marking as changes requested to make sure it doesn't get merged.

samueljackson92:
@martyngigg I think it should be fixed now.

martyngigg:
Sorry for not getting back to this.

The finding seems to work still and the GUI is responsive. I've found one thing that I think is still worth investigating. If you open the `Load` dialog wait for a find to finish, close the dialog and then quit Mantid I get messages like `QThreadStorage: Thread 0x2f80210 exited after QThreadStorage 3 destroyed` in the terminal. From the [Qt source code](https://code.woboq.org/qt5/qtbase/src/corelib/thread/qthreadstorage.cpp.html#197) it looks like it warns about this as it has the potential to crash the application (see the comment a few lines down). I did see a crash when I was opening the Load dialog and then immediately closed mantid - I don't know if this is related or not.

martyngigg:
I have also created #21243 to track that the failure mode takes a long time to fail but that is not related to this.

samueljackson92:
@martyngigg Thanks for taking a look. I shall see if I can find what is causing it.

martyngigg (approved):

PR #20984, opened 2017-10-20
Add support for non-constant bins to ConjoinWorkspaces

peterfpeterson:
To help reproduce a PDFgetN style workflow in mantid, `ConjoinWorkspaces` need to have the requirement of constant number of bins removed.

**To test:**

Since there is a new unit test, a code review should be sufficient.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


SimonHeybrock (changes requested):
Changes look good I think, apart from two questions:
1. I think you need to address how the changes you did on `WorkspaceJoiners` affects the other child class, `AppendSpectra`. Does it now (accidentally?) also support non-constant binning?
2. In `validateInputs` there is a branch `if(checkBinning)`. Would it make sense to remove that since now non-constant binning is support in the output? It seems a bit counter-intuitive to put restrictions on the input that the output violates, i.e., you cannot use `ConjoinWorkspaces` to merge 3 workspaces?

SimonHeybrock (approved):

PR #20559, opened 2017-09-21
LoadEventNexus MPI support

SimonHeybrock:
Adds MPI support for `LoadEventNexus`. The implementation is not particularly efficient and not meant for production. An actual parallel loader will be added later.

Depends on https://github.com/mantidproject/mantid/pull/20511 (merged).

Note that `LoadInstrument` is now working around #9365: `LoadParameterFile` causes segfaults in MPI related unit tests since it is modifying the base instrument and is thus not thread safe. This should not be an issue in an actuall MPI run.

**To test:**

Code review.

Fixes #20555.

No release notes, MPI is not officially supported yet.

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)?
- [ ] Are the unit tests small and test the class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] Do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant (user and developer) documentation been added/updated?

Does everything look good? Mark the review as **Approve**. A member of `@mantidproject/gatekeepers` will take care of it.

PR #20474, opened 2017-09-14
Parallel event loader components

SimonHeybrock:
Add some key components for the parallel event loader. Evidently it is incomplete (pending parser integration), but that will be completed in a follow-up.

**To test:**

Code review. This is not integrated anywhere yet so nothing can break.

Fixes #20438.

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)?
- [ ] Are the unit tests small and test the class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] Do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant (user and developer) documentation been added/updated?

Does everything look good? Mark the review as **Approve**. A member of `@mantidproject/gatekeepers` will take care of it.

PR #20996, opened 2017-10-23
Parallel event loader

SimonHeybrock:
This feature branch includes a series of changes undertaken to add MPI support with a faster loader to `LoadEventNexus`. Changes are as follows, details can be found in the original PRs (original authors are @LamarMoore and me):

*(A) Changes to `LoadEventNexus`*

1. https://github.com/mantidproject/mantid/pull/20425 -- Some cleanup in `LoadEventNexus` as preparation for subsequent changes.
2. https://github.com/mantidproject/mantid/pull/20511 -- Refactor index handling in `LoadEventNexus`. This includes a series of fixes for small index inconsistencies.
3. https://github.com/mantidproject/mantid/pull/20559 -- Basic MPI support in `LoadEventNexus` and related refactoring.

*(B) Additions for the new event loader. This parallel loader is disabled in non-MPI builds, i.e., these changes should have no effect for any of our released or otherwise supported builds of Mantid. Does does not affect the SNS MPI build either.*

1. https://github.com/mantidproject/mantid/pull/20526 (parser)
2. https://github.com/mantidproject/mantid/pull/20474 (other components)
3. https://github.com/mantidproject/mantid/pull/20942 (combine components from previous PRs)
4. https://github.com/mantidproject/mantid/pull/21021 (load units from NeXus)

Preliminary tests show a 2x to 3x speed up when loading large files on a workstation with an SSD. Tests on a cluster with a parallel file system are pending.

**To test:**

The PRs that are part of this PR have been reviewed and gatekeepered by several people. Nevertheless having another look at the code might be a good idea.

Issues fixed are mentioned in the linked PRs above.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Parallel/inc/MantidParallel/IO/NXEventDataLoader.h
```diff
@@ -0,0 +1,237 @@
...
+  attr.read(attr.getDataType(), value);
+  return value;
+}
+
+template <class TimeOffsetType, class IndexType, class TimeZeroType>
+std::unique_ptr<AbstractEventDataPartitioner<TimeOffsetType>>
+makeEventDataPartitioner(const H5::Group &group, const int numWorkers) {
+  const auto timeZero = group.openDataSet("event_time_zero");
+  int64_t time_zero_offset{0};
+  if (timeZero.attrExists("offset")) {
```
ianbush: I get an error here when building on Ubuntu 14.04. I guess this is a HDF5 version issue?

Framework/API/inc/MantidAPI/TriviallyParallelAlgorithm.h
```diff
@@ -0,0 +1,56 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+class MANTID_API_DLL TriviallyParallelAlgorithm : public Algorithm {
```
ianbush: The idea of this class is great, but I am not a big fan of the name. Would Easy or Simple be better than Trivially? I know Trivially has the right connotations when talking about parallelisation, but it is a bit cumbersome.

SimonHeybrock: As you say, `TriviallyParallel` probably has a commonly understood meaning, while I would not know what `EasyParallel` or `SimpleParallel` would imply. How about `EmbarrassinglyParallelAlgorithm`?

SimonHeybrock:
@ianbush Thanks for pointing out the HDF5 API version issue, it should be fixed. Could you try again?

quantumsteve:
On Fedora 26, I had to add `/usr/lib64/python2.7/site-packages/openmpi` to my `PYTHONPATH`.

Framework/DataHandling/src/LoadEventNexus.cpp
```diff
@@ -1511,96 +863,28 @@ void LoadEventNexus::loadEvents(API::Progress *const prog,
...
+    auto ws = m_ws->getSingleHeldWorkspace();
+    m_file->close();
+    try {
+      ParallelEventLoader::load(*ws, m_filename, m_top_entry_name, bankNames);
+      loaded = true;
+      shortest_tof = 0.0;
+      longest_tof = 1e10;
+    } catch (const std::runtime_error &) {
     }
+    safeOpenFile(m_filename);
```
quantumsteve: Is this in the right place? We only need to open the file after the parallel loader?

quantumsteve (changes requested):
I wasn't able to run the tests due to the `PYTHONPATH` issue mentioned earlier. 

I tried running `VisionReduction`, which uses `LoadEventNexus`, and it failed when I added `UseParallelLoader=True`. I'll try to simplify this down to a smaller test.

Framework/DataHandling/src/LoadEventNexus.cpp
SimonHeybrock: `LoadEventNeus` is opening the file via the Nexus API before reaching this piece of code. The parallel loader uses HDF5 directly and I didn't get it to work without first closing the file via the Nexus API, then let the parallel loader open and close it via HDF5, and in the end reopen it here such that `LoadEventNexus` can proceed.

Framework/DataHandling/inc/MantidDataHandling/DefaultEventLoader.h
```diff
@@ -0,0 +1,172 @@
...
+
+  /// chunk number
+  int chunk;
+  /// number of chunks
+  int totalChunks;
+  /// for multiple chunks per bank
+  int firstChunkForBank;
+  /// number of chunks per bank
+  size_t eventsPerChunk;
+
```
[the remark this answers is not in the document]
SimonHeybrock: I am certain that it will not make a difference since it is a high level class, but I changed it anyway :+1:.

Framework/Types/inc/MantidTypes/Core/DateAndTime.h
```diff
@@ -147,10 +147,66 @@ class MANTID_TYPES_DLL DateAndTime {
...
+  constexpr double minimum = static_cast<double>(MIN_NANOSECONDS);
+  constexpr double maximum = static_cast<double>(MAX_NANOSECONDS);
+  // Use these limits to avoid integer overflows
+  if (nano > maximum)
+    return MAX_NANOSECONDS;
+  else if (nano < minimum)
+    return MIN_NANOSECONDS;
+  else
+    return int64_t(nano);
+}
```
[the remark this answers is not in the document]
SimonHeybrock: Cannot use `return static_cast<int64_t>(boost::algorithm::clamp(name, min, max))` since casting to `double` and back to `int64_t` changes the value of min and max.

SimonHeybrock:
@quantumsteve Thanks for the comments. Regarding PYTHONPATH, I suppose this is only necessary if you build with MPI? You should be able to build and run tests also in a non-MPI build. Most Mantid unit tests will not run with MPI.

If you let me know in more detail how I can reproduce the issues you found with the Vision reduction I can also have a look myself. One possible reason I can think of is that (at least some) SNS files have some kind of "end" pointer in the `event_index` in `NXevent_data`, which is not standard and not supported by the parallel loader. I have seen it produce workspaces with event count off by one, compared to the default loader.

Framework/API/inc/MantidAPI/TriviallyParallelAlgorithm.h
SimonHeybrock: For me `PerfectlyParallelAlgorithm` makes a statement about the performance, computational complexity, or scaling behavior of the algorithm. In general algorithms that would use the base class are *not* scaling perfectly, in some cases not at all.

As an example, `LoadInstrument` is loading the *complete* instrument on each MPI rank. There is absolutely no speed up, but the implementation is trivial -- there are no changes or pieces of code that deal with the fact that there may be many other ranks running the same algorithm for the "same" workspace.

This is why I prefer `TriviallyParallelAlgorithm`, or something similar. I do however completely understand your criticism -- I also think it is a bit unwieldy and, while imho better than the other alternatives, also not entirely clear. How about `ImplicitlyParallelAlgorithm`?

@martyngigg Can we have your opinion as an "outsider" on this naming question?

martyngigg: I see the concerns here. I think it's important that the name convey something about the implementation requirements for the developer and I do like that `EmbarrassinglyParallel` is a more commonly understood term with regards to parallelization so that's what would get my vote.

SimonHeybrock: Reading https://en.wikipedia.org/wiki/Embarrassingly_parallel, it also seems to imply that the problem is easily split into sub-tasks. Are people fine with that, even if it is not correct for cases like `LoadInstrument`?

SimonHeybrock: Another option would be to simply go with `ParallelAlgorithm`. That does not imply much in particular and details can be found in the documentation.

Framework/Types/inc/MantidTypes/Core/DateAndTime.h
```diff
@@ -147,10 +147,66 @@ class MANTID_TYPES_DLL DateAndTime {
...
+ */
+inline DateAndTime::DateAndTime(const int64_t total_nanoseconds) {
+  // Make sure that you cannot construct a date that is beyond the limits...
+  if (total_nanoseconds > MAX_NANOSECONDS)
+    _nanoseconds = MAX_NANOSECONDS;
+  else if (total_nanoseconds < MIN_NANOSECONDS)
+    _nanoseconds = MIN_NANOSECONDS;
+  else
+    _nanoseconds = total_nanoseconds;
+}
```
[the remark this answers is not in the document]
SimonHeybrock: Reverted my change since it caused linker errors on OSX/`clang`, which I did not want to track down given that this was simply code moved from source to header.

Framework/API/inc/MantidAPI/TriviallyParallelAlgorithm.h
martyngigg: This would be fine with me.

quantumsteve:
Building on :redhat: 7 with `MPI_EXPERIMENTAL=TRUE` fails because the macro `BOOST_MPI_HAS_NOARG_INITIALIZATION` is not defined for MPI_VERSION 3 and later. 

This was fixed in Boost 1.56. Using software collections (which contains boost 1.57) is one possible alternative. 
https://github.com/boostorg/mpi/commit/ff6f4c64f7cdf926d1031a05ed0ef127c7aeb516

I'll create a self-contained repro and submit it to the redhat buzilla.

Framework/API/inc/MantidAPI/TriviallyParallelAlgorithm.h
quantumsteve: :+1:

quantumsteve (approved):
Between the unit tests, system tests, and the manual testing I did with VisionReduction, I'm fairly confident about functionality. We have time before the release to address any minor errors or omissions.

The main issues I'm seeing in code review are linting and modernization that may shouldn't change end-user functionality. They can and should be done in later pull request. 

Similarly issues encountered by the unsupported `MPI_EXPERIMENTAL` build can be addressed in later PRs.

OwenArnold:
Thorough review by @quantumsteve with previous PRs and reviews into the long-running feature branch. There are a lot of changes here, but much of this is new functionality and should not affect existing users. @LamarMoore can you please monitor performance tests later today to try to identify any (hopefully unlikely) side-effects.

PR #17476, opened 2016-09-13
Updated instrument definitions for POLREF

PranavBahuguna:
This pull request updates the instrument definitions for POLREF that are required to be shipped for the next Mantid install. As stated in the issue it fixes, this fixes a few issues and maintains some semblance of backwards data reduction compatibility.

Release notes have been updated.

Fixes #17475.

**To test:**

Simply ensure that the instrument definition files have been updated correctly.

---
#### Reviewer

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):
##### Code Review
- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)? Is it well structured with small focussed classes/methods/functions?
- [ ] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [x] If there are changes in the release notes then do they describe the changes appropriately?
##### Functional Tests
- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant documentation been added/updated?
- [ ] Is user-facing documentation written in a user-friendly manner?
- [ ] Has developer documentation been updated if required?
- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

Issue #16218, opened 2016-05-16
OFFSPEC IDF does not have slits defined


raquelalvarezbanos:
Discuss this with scientists. They are currently trying to harmonize their IDFs across instruments, as part of this harmonization, slits should be defined for OFFSPEC. This is needed for instance to run `CalculateResolution` successfully on OFFSPEC data.

Also while investigating #16518 it was found the OFFSPEC parameter file did not contain any values for `Correction`. These should also be added if scientists need them.

gemmaguest:
The new (experimental) IDF supplied by Jos should fix this

PR #20997, opened 2017-10-23
Re #20991: Updated Reflectometry IDFs.

### LINKED ISSUE #20991: INTER IDF should position detectors at twoTheta
### Expected behavior
The INTER IDF should reposition detectors to twoTheta on Load.

### Actual behavior
If using a linear detector, the detector is positioned at theta instead of twoTheta.

Run the following script. Both the point and linear detectors should be re-positioned on load to twoTheta=2.6, but currently the linear detector is being placed at theta=1.3 instead. (Note that the output of this script is not exact because the pixel centre is not positioned precisely at twoTheta. This is expected.)
```
import math
 
def getTwoThetaForPixel(wkspace, detId):
    inst = wkspace.getInstrument()
    det = inst.getDetector(detId)
    return wkspace.detectorTwoTheta(det) * 180 / math.pi
 
print "Expected twoTheta=2.6 in both cases"
 
print "INTER43572 point-detector puts detectors at twoTheta (nearest pixel is 4):"
pixel=4
ws=Load(Filename='INTER43572')
detTwoTheta = getTwoThetaForPixel(ws,pixel)
print '  Detector ID',pixel,'twoTheta =', detTwoTheta
 
print "INTER43583 linear-detector puts detectors at theta instead of twoTheta (nearest pixel is 2047):"
pixel=2047
ws=Load(Filename='INTER43583')
detTwoTheta = getTwoThetaForPixel(ws,pixel)
print '  Detector ID',pixel,'twoTheta =', detTwoTheta
```

### Platforms affected
All

edwardb96:
Updates the IDFs for INTER and SURF.
Removes `ReflectometryReductionOneAuto` v1
This also fixes #20991.

**To test:**
- In the ISIS Reflectometry interface, try autoreducing some `INTER` investigations dated both before and after the date the IDF changed (14/2/2017 for INTER), e.g. before: `1620396`, after: `1620384`. In v3.11 the latter gave an error about slit gaps not being found. You don't need to wait for all of the runs to reduce, as long as the first few groups process ok then the correct IDF is being picked up.
- Repeat the test for the following `SURF` investigations: before: `1610531`, after: `1700019`
- Run this script. The output for the nearest pixels for INTER should be around twoTheta=2.6 rather than theta=1.3. (Note that the output is not exact because the pixels are not centred exactly at twoTheta.) For SURF the output should be around twoTheta=3.04.
```python
import math
def getTwoThetaForPixel(wkspace, detId):
    inst = wkspace.getInstrument()
    det = inst.getDetector(detId)
    return wkspace.detectorTwoTheta(det) * 180 / math.pi
def doCheck(run, pixel):
    print run, "(nearest pixel is ", pixel, "):"
    ws=Load(Filename=run)
    detTwoTheta = getTwoThetaForPixel(ws,pixel)
    print '  Detector ID',pixel,'twoTheta =', detTwoTheta

print "INTER: Expect twoTheta to be around 2.6 in both cases"
doCheck("INTER43572", 4)
doCheck("INTER43583", 2047)
print
print "SURF: Expect twoTheta to be around 3.04"
doCheck("SURF121400", 30001)
```

Expected output:
```
INTER: Expect twoTheta to be around 2.6 in both cases
INTER43572 (nearest pixel is  4 ):
  Detector ID 4 twoTheta = 2.63604652684
INTER43583 (nearest pixel is  2047 ):
  Detector ID 2047 twoTheta = 2.59952111409

SURF: Expect twoTheta to be around 3.04
SURF121400 (nearest pixel is  30001 ):
  Detector ID 30001 twoTheta = 3.03571136117
```

Fixes #20991. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


gemmaguest:
Since the INTER IDF contains a breaking change (a change in the log name used for slit gaps) I think we need to version this file in a similar manner to PR #17476, i.e. add it as a new file with a different name and an appropirate `valid-from` date, and revert the original `INTER_Definition.xml` but change its `valid-to` date to the appropriate date. I'm not sure whether we can find out the date the log name changed or whether Max could tell us.

gemmaguest:
I've added the Science Validation label because we should check with Max that your fixes are ok with him, i.e. the fixes for `ystart` and `TransRunEndOverlap`, and the version dates for the INTER IDF (was there anything else?).

gemmaguest:
When we test this, we should test reducing INTER runs both before and after the version date, e.g. on the ISIS Reflectometry interface, enter investigation 1620384 (16 Feb 2017) and click Autoreduce - this currently fails due to the changed log name, whereas investigation 1620396 (Dec-2016) works. Both should work after this PR.

NickDraper:
Note: When this is merged these will be downloaded to users immediately, they will not wait for the next release.

gemmaguest:
Max said he is fine for these to go live whenever we are ready.

gemmaguest:
This PR also adds slits for offspec which is part of issue #16218

Scientists are happy with the INTER IDF now but not with the bench rotation for OFFSPEC, so we should hold off merging this for now.

instrument/INTER_Definition_2017.xml
```diff
@@ -0,0 +1,246 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- For help on the notation used to specify an Instrument Definition File
+     see http://www.mantidproject.org/IDF -->
+<instrument xmlns="http://www.mantidproject.org/IDF/1.0"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xsi:schemaLocation="http://www.mantidproject.org/IDF/1.0 http://schema.mantidproject.org/IDF/1.0/IDFSchema.xsd"
+            name="INTER"
+            valid-from="2017-02-14 00:00:00"
+            valid-to=  "2100-01-31 23:59:59"
+			last-modified="2010-11-04 00:00:00">
```
gemmaguest: We should put a more sensible last-modified date in here. I think this is just for reference so doesn't need to be accurate. Maybe 26 October as that was the last commit date?

gemmaguest (changes requested):
Just one minor comment

gemmaguest:
I think I made a mistake when I said the old version of the INTER IDF should not be changed. We're making two sets of changes here - the positions, and the log names. The log names need to remain the same to match the older runs, but the other changes should be included in the old IDF, otherwise we'll get different behaviour when using old vs new runs. I therefore think we should change the old IDF to be the same as the new one for everything apart from the slit log names (S1_VG etc), and of course the valid dates. I'll email Max and Becky to check this is what they expect.

gemmaguest:
@edwardb96 We had some more minor changes to the INTER and SURF IDFs requested today, so I've pushed those changes. Also the doctest was failing again so I'm not sure the last fix was correct. I've reverted it for now and will check it on Monday.

gemmaguest (approved):
The changes look good and behave as expected :+1:

There are a lot of changes in tests etc. and removal of old versions of our algorithms which no longer work due to the significant changes in the IDFs. Scientists have confirmed that the old algorithms are not in use and have verified the new IDFs.

It might be worth looking at creating replacement system tests for the OFFSPEC tests that have been removed. I think this should be done as a separate PR though because we don't want to hold up these IDF changes any more.

PR #21021, opened 2017-10-25
Correctly handle units in parallel event loader

### LINKED ISSUE #21015: Read unit of event data in parallel event loader
Must read the unit attribute for `event_time_zero` and `event_time_offset`.

SimonHeybrock:
Previously this made assumptions about default units, now they are read from the `units` attribute defined by the Nexus standard.

**To test:**

Code review.

Fixes #21015.

No release notes since this is not used in anything public.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Parallel/inc/MantidParallel/IO/EventParser.h
+++ b/Framework/Parallel/inc/MantidParallel/IO/EventParser.h
@@ -54,11 +54,6 @@ namespace detail {
 void MANTID_PARALLEL_DLL eventIdToGlobalSpectrumIndex(int32_t *event_id_start,
                                                       size_t count,
                                                       const int32_t bankOffset);
-
-template <class TimeOffsetType>
-void redistributeDataMPI(
-    Communicator &comm, std::vector<Event<TimeOffsetType>> &result,
-    const std::vector<std::vector<Event<TimeOffsetType>>> &data);
 }
 
 template <class TimeOffsetType> class EventParser {
@@ -84,6 +79,7 @@ private:
                  const TimeOffsetType *event_time_offset_start,
                  const Chunker::LoadRange &range);
 
+  void redistributeDataMPI();
   void populateEventLists();
 
   // Default to 0 such that failure to set unit is easily detected.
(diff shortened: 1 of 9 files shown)


Framework/Parallel/inc/MantidParallel/IO/EventParser.h
```diff
@@ -104,6 +78,12 @@ template <class TimeOffsetType> class EventParser {
   void doParsing(int32_t *event_id_start,
                  const TimeOffsetType *event_time_offset_start,
                  const Chunker::LoadRange &range);
+
+  void redistributeDataMPI();
```
LamarMoore: Were these moved out of testing scope because we only need to test against the resulting `EventList`?

LamarMoore (commented):
Looks good 👍

Framework/Parallel/inc/MantidParallel/IO/EventParser.h
SimonHeybrock: There was no test for the methods so making them private was ok. If we would later write tests for them, they can be made free functions again.

LamarMoore (approved):

Framework/Parallel/inc/MantidParallel/IO/EventParser.h
```diff
@@ -146,80 +126,94 @@ void EventParser<TimeOffsetType>::setEventDataPartitioner(
...
-    Communicator &comm, std::vector<Event<TimeOffsetType>> &result,
-    const std::vector<std::vector<Event<TimeOffsetType>>> &data) {
-  using Event = Event<TimeOffsetType>;
-  if (comm.size() == 1) {
-    result = data.front();
+void EventParser<TimeOffsetType>::setEventTimeOffsetUnit(
+    const std::string &unit) {
+  constexpr char second[] = "second";
+  constexpr char microsecond[] = "microsecond";
+  constexpr char nanosecond[] = "nanosecond";
```
OwenArnold: I suppose it's reasonable to assume that case will not be an issue we would have to consider?

SimonHeybrock: What case?

OwenArnold: A few lines below, where the comparison is made (sorry should have put my comment there) on the `==` comparisons. i.e. we are being strict about file authors not having "Microseconds" instead of "microseconds"

SimonHeybrock: I am not sure how precise the standard is formulated in that respect. Changing this in case it causes issues should be trivial in the future, I think.

PR #21023, opened 2017-10-25
Load isis live timeout for histogram data

### LINKED ISSUE #20746: LoadLiveData does not time out
After testing the fix for #20703 it was noticed that `LoadLiveData` never times out if it can't connect to the server. Mantid is left in a state that requires a user to kill it rather than being able to quit cleanly.

The bug can be reproduced with the same testing instructions that were used for producing #20703.

NickDraper:
This has added a timeout to the ISIS Histogram DAE listener.  Unfortunately the timeout cannot seperate network failure from long running successful operations, so the value of the timeout has been set to 120 seconds by default.

You can configure it in the user.properties file with the key:
```
ISISDAE.Timeout = 20 #seconds
```

**To test:**

**NOTE: This has some windows/linux specific code, it has been developed on windows so I'm pretty confident about it working their, but it needs testing on Linux/Mac.**

1. Run Mantid, Load->Live Data
1. Pick an instrument ("Offspec", "Engin-X" etc avoid the really big ones like MAPS, MERLIN)
1. Use Histogram mode
1. Rest of the load live settings do not matter much.
1. Run the live load.
1. After the dataset has first appeared pull out the network cable from your PC.
1. Mantid should remain responsive, the monitorLiveData algorithm should remain until the timeout, when you will also get an error message (not a great one, but ok) to the log, then it should disappear.
1. Mantid should close cleanly (after the timeout) without hanging.

1. You should add this line to your user.properties file and repeat the test above, this time the timeout should be only 20 seconds. You will need to restart mantid for the change to be picked up.
```
ISISDAE.Timeout = 20 #seconds
```

Fixes #20746
**Release Notes** 
docs/source/release/.../framework.rst

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


NickDraper:
The least build did complete successfully, but Jenkins did not register it

josephframsay (commented):
The test with timeout at 120 seconds works in my Ubuntu virtual machine, but changing the value in either `Mantid.user.properties` or `MantidTest.user.properties` doesn't seem to have any effect on the value in the config service - should I be setting this somewhere else?

Matthew-Andrew (approved):
This works for me on Ubuntu 16.04.

PR #21033, opened 2017-10-26
Re #21011 Fix missaligned ticks on 2D plots

### LINKED ISSUE #21011: Ticks missaligned on 2D plot.
### Expected behavior
The ticks line up on the left to right and bottom to top regardless of the zoom level.

### Actual behavior
When zooming in the plot ticks become misaligned

Open a 2D plot and experiment with different zoom levels.

### Platforms affected
Probably all, verified on Windows 10 and Fedora 26

edwardb96:
Fixes #21011 by updating the secondary axes and re-plotting after a zoom.

Additionally this PR reduces tidies up some of the code in `Plot.cpp` - originally thought to be the cause of the bug.

**To test:**
1. Open a 2D plot
2. Zoom into an area of the plot.
3. Ensure that the markers on the top axes line up with the markers on the bottom and the same for the left and right axes.

- I am not sure if the changes to `Graph.cpp` have other undesirable effects on other plots.
- I am not sure if the contents of `PlotTick.cpp/.h` should be split into more, separate files. 

*Please sanity check this PR*

Fixes #21011 . 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidPlot/src/ApplicationWindow.cpp
+++ b/MantidPlot/src/ApplicationWindow.cpp
@@ -7557,7 +7557,7 @@ void ApplicationWindow::zoomIn() {
   QList<Graph *> layers = plot->layersList();
   foreach (Graph *g, layers) {
     if (!g->isPiePlot())
-      g->zoom(true);
+      g->zoomMode(true);
   }
 }
 
--- a/MantidPlot/src/Graph.cpp
+++ b/MantidPlot/src/Graph.cpp
@@ -62,8 +62,8 @@
 #include "Mantid/ErrorBarSettings.h"
 #include "Mantid/MantidMDCurve.h"
 #include "Mantid/MantidMatrixCurve.h"
-#include "MantidKernel/Strings.h"
 #include "MantidAPI/AnalysisDataService.h"
+#include "MantidKernel/Strings.h"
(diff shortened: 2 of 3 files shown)


edwardb96:
@martyngigg I've tried running the test RHEL7 keeps failing on but I can't seem to re-create the segmentation fault locally. Any thoughts? Is it worth trying a local build with gcc rather than clang?

martyngigg:
@edwardb96 That is odd. I wonder if it is to do with the version of the Qwt library on RHEL 7. It is most likely an older version there. A VM with CentOS might be the only way to reproduce this.

martyngigg:
One additional comment/concern. I agree that it's good to generally refactor and improve code as we go but in this case we know `MantidPlot` will be deleted so I would caution against putting too much effort in to refactoring in these cases.

In this case for example it's impossible to see where the "real" fix for the issue was added. Is it worth reverting the code, fixing it as it was?

martyngigg:
Thanks, @edwardb96. That's much clearer now.

martyngigg:
Would it be possible to clean up the history by using `rebase -i` to remove the commits that we don't want?

martyngigg (approved):
Thanks, @edwardb96. This looks good now.

PR #21066, opened 2017-10-27
ISIS SANS gui does not use q ranges when fitting merged data if they are specified 

### LINKED ISSUE #21041: Sans GUI does not use Q ranges when fitting merged data if they are specified.
### Expected behavior
When a range is entered which is within bounds this is used instead of the default range calculated by the algorithm.

Matthew-Andrew:
There are currently some options in both sans GUI which are never used in the code. These relate to specifying a custom range to use to find the fit parameters for stitching together the HAB and LAb DATA

**To test:**
- Code Review
- Confirm new unit tests work

Fixes #21041 . 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


NickDraper (changes requested):
Unit test seems reasonable and passes.

However I disagree that this does not need to be in the release notes.  It should be added to the SANS page.

NickDraper (approved):
All good now with the release notes updated

PR #21079, opened 2017-10-30
Try to make MaxEnt faster

### LINKED ISSUE #20983: Try to make MaxEnt faster
From my point of view MaxEnt does too many iterations. Also there are inefficiencies in the code.

mantid-roman:
Changed calculations of the quadratic sub-space.

**To test:**

Compare the numbers of iterations it takes the algorithm to converge before and after the change.

Fixes #20983.

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -117,12 +120,24 @@ void MaxEnt::init() {
...
+      "reconstructions."
+      "In general the user will need to experiment with this value. Chosen a "
+      "small value may lead to unphysical spiky reconstructions and a too "
+      "large "
+      "value the reconstruction will start to resamble that of a direct "
+      "fourier "
+      "transform reconstruction. However, where data contain a strongly "
+      "determined zero "
+      "Fourier data point the reconstruction will be insensitive to the choice "
+      "of this property.");
```
AnthonyLim23: just a bit of cleaning up on the wording:

`In general the user will need to experiment with this value. Choosing a small value may lead to unphysical spiky reconstructions. Whereas, a large value will generate a reconstruction that resembles that of a direct Fourier transform reconstruction. However, when the data contains a strongly determined zero Fourier data point the reconstruction will be insensitive to the choice of this property.`

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -117,12 +120,24 @@ void MaxEnt::init() {
...
 
   declareProperty(make_unique<PropertyWithValue<double>>(
-                      "ChiTarget", 100.0, mustBeNonNegative, Direction::Input),
+                      "ChiTarget", 1.0, mustBeNonNegative, Direction::Input),
```
AnthonyLim23: can we change the name to something like `ChiTargetOverN`?

Framework/Algorithms/test/MaxEntTest.h
```diff
@@ -313,7 +308,6 @@ class MaxEntTest : public CxxTest::TestSuite {
     alg->setChild(true);
     alg->setProperty("InputWorkspace", ws);
     alg->setProperty("A", 0.01);
-    alg->setProperty("ChiTarget", 50.);
```
AnthonyLim23: should we not be setting a smaller `ChiTarget` value e.g. 0.5?

AnthonyLim23 (dismissed):
I think at least one of the examples should have a ChiTarget that is not the defaults (so users can see how to use it). I also think we should have a test that alters the ChiTarget value.

AnthonyLim23:
it appears to run ok

Framework/Algorithms/test/MaxEntTest.h
mantid-roman: What information would it give us? We are testing the resolution factor here.

AnthonyLim23: based on the new definition of chiTarget setting it to 50 means that we are using 50*N (N = number of data points). This just seemed a bit excessive to me

mantid-roman: This line is deleted. No ChiTarget is set.

AnthonyLim23:
I have just noticed that the progress bar is sometimes wrong. I ran the below script and the progress bar would say it had over a min but then reset (I assume it completes) after a second or two


```
import mantid.simpleapi as mantid
import numpy as np

Load(Filename=r'\\isis\inst$\ndxemu\instrument\data\cycle_10_1\EMU00020711.nxs', OutputWorkspace='EMU00020711')
CropWorkspace(InputWorkspace='EMU00020711', OutputWorkspace='phaseWS', XMin=0, XMax=16)

Load(Filename=r'\\isis\inst$\ndxemu\instrument\data\cycle_10_1\EMU00020730.nxs', OutputWorkspace='input')
CropWorkspace(InputWorkspace='input', OutputWorkspace='croped', XMin=0, XMax=16)
#dirty remove the exp
EstimateMuonAsymmetryFromCounts(InputWorkspace='croped', OutputWorkspace='dh')

# do maxEnt
MaxEnt(InputWorkspace='dh', ComplexImage=True, AutoShift=True, MaxIterations='1000', A=1.e-3,ResolutionFactor=5,AlphaChopIterations='25', EvolChi='chi', EvolAngle='angle', ReconstructedImage='image', ReconstructedData='data')
fft= mantid.AnalysisDataService.retrieve("image")
# get phases
CalMuonDetectorPhases(InputWorkspace='phaseWS', FirstGoodData=0.5, LastGoodData=10, Frequency=1.35, DetectorTable='phaseTable', DataFitted='jkb5')
phaseTable = mantid.AnalysisDataService.retrieve("phaseTable")
phases= phaseTable.column(2)
#useful
j= complex(0,1)
# create empty complex array
yData=[]
for k in range(len(fft.readY(0))):
    yData.append(complex(0.0,0.0))
# sum freqs
N=64 #input.getNumberHistograms()
for k in range(N):
    phi =phases[k] 
    # 64 is the number of histograms -> need to make general
    yData+=(fft.readY(k)+j*fft.readY(k+N))*np.exp(-j*phi)

outRe=CreateWorkspace(fft.readX(0),yData.real)
outIm=CreateWorkspace(fft.readX(0),yData.imag)
```

mantid-roman:
@AnthonyLim23 I believe it's the same as with Fit: the number of actual iterations is unknown and the progress bar at 100% after max number of iterations.

AnthonyLim23:
that makes sense

AnthonyLim23:
@Anders-Markvardsen  could you double check that this is doing the same thing as before?

Anders-Markvardsen:
I have tested this PR this morning. Will need to do some more. Hence, yes, don't merge this PR into master yet

Anders-Markvardsen:
Using the usage example (here in format of this PR):

Load(Filename=r'EMU00020884.nxs', OutputWorkspace='ws')
CropWorkspace(InputWorkspace='ws', OutputWorkspace='ws', XMin=0.17, XMax=4.5, EndWorkspaceIndex=0)
ws = RemoveExpDecay(InputWorkspace='ws')
ws = Rebin(InputWorkspace='ws', Params='0.016')
evolChi1, evolAngle1, image1, data1 = MaxEnt(InputWorkspace='ws', A=0.0001, MaxIterations=2500, ResolutionFactor=1)
evolChi2, evolAngle2, image2, data2 = MaxEnt(InputWorkspace='ws', A=0.0001, MaxIterations=5000, ResolutionFactor=2)

then plotting image1 (both spectra) I see:

capture

Anders-Markvardsen:
Then plotting the image1 with the old code I get:

capture

This answer is more believable because it does what you expect ironing out uncessary features. If you do the same for image2 you get a worse agreement. @mantid-roman perhaps the maxent is no longer converting?

Framework/Algorithms/src/MaxEnt/MaxentCalculator.cpp
```diff
@@ -292,17 +306,27 @@ void MaxentCalculator::calculateChisq() {
...
+MaxentCalculator::calculateChiSquared(const std::vector<double> &data) const {
   size_t npoints = m_data.size();
+  auto dpoints = static_cast<double>(npoints);
 
   // Calculate
   // ChiSq = sum_i [ data_i - dataCalc_i ]^2 / [ error_i ]^2
-  m_chisq = 0;
+  double chisq = 0;
   for (size_t i = 0; i < npoints; i++) {
     if (m_errors[i] != 0.0) {
```
Anders-Markvardsen: You did not add "if (m_errors[i] != 0.0)" but probably it was added for a reason (unfortunately no comment). However with that if statement you may be dividing by the wrong number! Could you please look into understanding what this is doing and add a comment to explain why it is there (and if not remove). If the if statement is needed then the number of data points equals the number of times the content of this if statement is called.

Also please change the method description to include the 1/N

Anders-Markvardsen (changes requested):
As per comments I just made

mantid-roman:
@Anders-Markvardsen I don't see the image that you show with mantid 3.11
What I also want to note is that the data is a sine function making exactly 6 periods. It's integral must be 0. But if you look at the central point on the black line you'll see that the old algorithm gives a spike there (the central point is the zero-frequency amplitude == integral over the data). With the new algorithm it's practically 0 as expected
image1
.

Framework/Algorithms/src/MaxEnt/MaxentCalculator.cpp
mantid-roman: I just noticed that dividing every term by `dpoints` is very inefficient.

I can see the reason for the `if` in protecting against accidental (practically impossible) zero errors. 
I am afraid that if I change `dpoints` to the number of non-zero errors it might affect the code that calculates `c1, c2, s1, s2` values. I don't see benefits in trying to do it because:
1. probability of having significant number of zeros is very low
2. inaccuracy in the `chi2` can be offset at run-time by adjusting the target chi property.

AnthonyLim23:
I find that both 3.10 and 3.11 gives the same as Roman.

Anders-Markvardsen:
For information, the older version I was using to plot image1 above was 3.8.20170117.1529 !

AnthonyLim23:
Just tried 3.9 and I got the below image. It looks like the bug was introduced between 3.8 (Oct 2016) and 3.9 (Feb 2017)

image

AnthonyLim23:
it looks like MaxEnt.cpp was not edited between 3.8 and 3.9....

Anders-Markvardsen:
I agree.... none of the maxent files have been touched in this period - but perhaps the FFT have been touched

AnthonyLim23:
my 3.8 had the bug in it. I will try some earlier versions

image

AnthonyLim23:
The noise was due to the choice of chiTarget.

Anders-Markvardsen:
A not essential (I think) comment to one of your previous comments @mantid-roman note that the integral over the rebinned EMU00020884.nxs data above times bin-width is 0.097. Hence an FFT on this data would always produce this exact value. On the basis of the data along it is not obvious to me that the MaxEnt should favour zero value for this point

Anders-Markvardsen:
On the old code running using ChiTarget=270 instead of ChiTarget=300 (there are 270 data points) I get

Load(Filename=r'EMU00020884.nxs', OutputWorkspace='ws')
CropWorkspace(InputWorkspace='ws', OutputWorkspace='ws', XMin=0.17, XMax=4.5, EndWorkspaceIndex=0)
ws = RemoveExpDecay(InputWorkspace='ws')
ws = Rebin(InputWorkspace='ws', Params='0.016')
evolChi1, evolAngle1, image1, data1 = MaxEnt(InputWorkspace='ws', A=0.0001, ChiTarget=270, MaxIterations=2500, ResolutionFactor=1)

capture

which as a stronger central peak as you may expect from the above comment, but also some small-ish evidence of what looks like noise

Anders-Markvardsen:
Running this PR code with a smaller than the default MaxAngle value:

Load(Filename=r'EMU00020884.nxs', OutputWorkspace='ws')
CropWorkspace(InputWorkspace='ws', OutputWorkspace='ws', XMin=0.17, XMax=4.5, EndWorkspaceIndex=0)
ws = RemoveExpDecay(InputWorkspace='ws')
ws = Rebin(InputWorkspace='ws', Params='0.016')
evolChi1, evolAngle1, image1, data1 = MaxEnt(InputWorkspace='ws', A=0.0001, MaxIterations=2500, ResolutionFactor=1, MaxAngle=0.001)

capture

which looks more similar to the result with the old code, in fact I would say promising similar.

AnthonyLim23:
there still seems to be some noise, but it does look good

Anders-Markvardsen:
If instead keeping MaxAngle at default and instead reducing ChiEps to 0.000001 makes no improvement:

capture

Anders-Markvardsen:
In the old code running

Load(Filename=r'EMU00020884.nxs', OutputWorkspace='ws')
CropWorkspace(InputWorkspace='ws', OutputWorkspace='ws', XMin=0.17, XMax=4.5, EndWorkspaceIndex=0)
ws = RemoveExpDecay(InputWorkspace='ws')
ws = Rebin(InputWorkspace='ws', Params='0.016')
evolChi2, evolAngle2, image2, data2 = MaxEnt(InputWorkspace='ws', A=0.0001, ChiTarget=270, MaxIterations=5000, ResolutionFactor=2)

gives

capture

Anders-Markvardsen:
Running the above with default MaxAngle gives:

capture

Anders-Markvardsen:
running with MaxAngle = 0.01

capture

AnthonyLim23:
@Anders-Markvardsen  in Romans' version do  you have muon analysis available to select? It seems to have vanished from my interfaces list. I will try a clean build.

Anders-Markvardsen:
running with MaxAngle = 0.001

capture

Anders-Markvardsen:
@AnthonyLim23  can you try with the modified MaxAngle value to double check that speed gain still there?

AnthonyLim23:
I will do that now

AnthonyLim23:
I have ran the following script and altered the value of maxAngle.
```
from datetime import datetime 

Load(Filename=r'EMU00020884.nxs', OutputWorkspace='ws')
CropWorkspace(InputWorkspace='ws', OutputWorkspace='ws', XMin=0.17, XMax=15)
ws = RemoveExpDecay(InputWorkspace='ws')
ws = Rebin(InputWorkspace='ws', Params='0.016')
startTime= datetime.now() 
evolChi1, evolAngle1, image1, data1 = MaxEnt(InputWorkspace='ws', A=0.0001, MaxIterations=2500, ResolutionFactor=1, MaxAngle=0.05)
timeElapsed=datetime.now()-startTime 

print('Time elpased (hh:mm:ss.ms) {}'.format(timeElapsed))
```

A summary of the results (min:sec) and the PR is a release build:

maxAngle       time (PR)          time (3.11)
0.05                14:33                  
0.001              13:59
0.0001            13:54

AnthonyLim23:
repeating yesterdays test but with a different run EMU00020711.

maxAngle _____ time (PR) _______ time (3.11)
0.05 ____________00:06  ___________ 
0.001 ___________00:18  ___________
0.0001 __________ 00:18 ___________

Here are the plots for the evolution of the angle (roman)

image

V3.11

AnthonyLim23:
image
image

mantid-roman:
@AnthonyLim23 Are you sure that your analysis of the `EMU00020884.nxs` dataset was accurate?
On my machine version v.3.11 took about 13 mins while the new version ~ 11 sec. 

v.3.11
max_ent_v 3 11

v.new
max_ent_v new

AnthonyLim23:
@mantid-roman , I just had another and you were right, I must have had the cmake wrong (I re did it between the two sets of tests)

Anders-Markvardsen (approved):

Anders-Markvardsen:
Changes provides speedup while maintaining previous outputs

AnthonyLim23 (approved):
looks good and works as described

PR #21080, opened 2017-10-30
Fix workspace operators when workspaces are not in the ADS

### LINKED ISSUE #21064: AddSampleLog doesn't work with non-ADS workspace
[`AddSampleLog`](http://docs.mantidproject.org/nightly/algorithms/AddSampleLog-v1.html) fails when its *Workspace* property is set to a non-ADS workspace. The following script
```python
ws = CreateSampleWorkspace(StoreInADS=False)
AddSampleLog(ws, 'log_entry', 'log_value', 'String')
```
gives the following error:
```
RuntimeError: Add Data Object with empty name
```

### LINKED ISSUE #21065: Binary and unary operators do not work with non-ADS workspaces in Pyhton
Binary and unary operators (`+`, `-`, `~`, etc.) don't work when the workspaces are not stored in the ADS.

The following script
```python
ws = CreateSampleWorkspace(StoreInADS=False)
ws *= 0.0
```
give the error
```
ValueError: Invalid value for property LHSWorkspace (MatrixWorkspace) "": Enter a name for the Input/InOut workspace
```

soininen:
The binary operators (`+`, `-`, `*=`, etc.) did not work if one or both of the operands was not stored in the Analysis data service. This PR fixes the issue.

Additionally, this PR changes the copy constructor of `Workspace` to not to copy the workspace name field. The copy constructor is used in cloning. The clones are no longer in the ADS and therefore should have empty names. The clone's name was causing WorkspaceOpOverloads to confuse the original workspace in the ADS with the non-ADS clone.

**To test:**

Play with workspaces both in and out of the ADS and binary operators. The following Python script tests some cases and the single unary operator available in Mantid:
```python
ws = CreateSampleWorkspace(StoreInADS=False)
initialY = ws.readY(0)[0]
ws_ads = CreateSampleWorkspace(StoreInADS=True)
result1 = ws + ws
result2 = ws + ws_ads
result3 = ws_ads + ws
result4 = ws_ads + ws_ads
result5 = ws + 1
result6 = 1 + ws
result7 = ws_ads + 1
result8 = 1 + ws_ads
ws += 1
ws_ads += 1
if ws.readY(0)[0] != initialY + 1 or ws_ads.readY(0)[0] != initialY + 1:
    raise RuntimeError('Operator += did not work.')
mdws = CreateMDHistoWorkspace(SignalInput=[0], ErrorInput=[0], Dimensionality=1, Extents=[0, 1], NumberOfBins=1, Names=['a'], Units=['TOF'], StoreInADS=False)
mdws_ads = CreateMDHistoWorkspace(SignalInput=[0], ErrorInput=[0], Dimensionality=1, Extents=[0, 1], NumberOfBins=1, Names=['a'], Units=['TOF'], StoreInADS=True)
result9 = ~mdws
result10 = ~mdws_ads
```

Fixes #21065, #21129.

**Release Notes** 

The Framework's release notes were updated accordingly.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/PythonInterface/test/python/mantid/api/WorkspaceBinaryOpsTest.py
+++ b/Framework/PythonInterface/test/python/mantid/api/WorkspaceBinaryOpsTest.py
@@ -1,0 +1,38 @@
+from __future__ import (absolute_import, division, print_function)
+
+from mantid.api import mtd
+from mantid.simpleapi import CreateSampleWorkspace
+import unittest
+
+
+class WorkspaceBinaryOpsTest(unittest.TestCase):
+    def tearDown(self):
+        mtd.clear()
+
+    def test_binary_ops_with_workspaces_not_in_ADS(self):
+        ws = CreateSampleWorkspace(StoreInADS=False)
+        initialY = ws.readY(0)[0]
+        ws_ads = CreateSampleWorkspace(StoreInADS=True)
+        result1 = ws + ws
+        self.assertTrue(mtd.doesExist('result1'))
+        result2 = ws + ws_ads
+        self.assertTrue(mtd.doesExist('result2'))
+        result3 = ws_ads + ws
+        self.assertTrue(mtd.doesExist('result3'))
+        result4 = ws_ads + ws_ads
+        self.assertTrue(mtd.doesExist('result4'))
+        result5 = ws + 1
+        self.assertTrue(mtd.doesExist('result5'))
+        result6 = 1 + ws
+        self.assertTrue(mtd.doesExist('result6'))
+        result7 = ws_ads + 1
+        self.assertTrue(mtd.doesExist('result7'))
+        result8 = 1 + ws_ads
+        self.assertTrue(mtd.doesExist('result8'))
+        ws += 1
+        self.assertFalse(mtd.doesExist('ws'))
+        ws_ads += 1
+        self.assertTrue(mtd.doesExist('ws_ads'))
+
+if __name__ == '__main__':
+    unittest.main()
--- a/Framework/TestHelpers/inc/MantidTestHelpers/FakeObjects.h
+++ b/Framework/TestHelpers/inc/MantidTestHelpers/FakeObjects.h
@@ -191,16 +191,6 @@ public:
 
   const std::string id() const override { return "WorkspaceTester"; }
 
-  /// Returns a clone of the workspace
-  std::unique_ptr<WorkspaceTester> clone() const {
-    return std::unique_ptr<WorkspaceTester>(doClone());
-  }
-
-  /// Returns a default-initialized clone of the workspace
-  std::unique_ptr<WorkspaceTester> cloneEmpty() const {
-    return std::unique_ptr<WorkspaceTester>(doCloneEmpty());
-  }
-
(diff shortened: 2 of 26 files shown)


SimonHeybrock (commented):
Question: Instead of clearing the name in the `clone()` method, would it make (more?) sense to clear it in the copy constructor?

Framework/TestHelpers/inc/MantidTestHelpers/FakeObjects.h
```diff
@@ -191,16 +191,6 @@ class WorkspaceTester : public AxeslessWorkspaceTester {
 
   const std::string id() const override { return "WorkspaceTester"; }
 
-  /// Returns a clone of the workspace
-  std::unique_ptr<WorkspaceTester> clone() const {
```
SimonHeybrock: Why are these gone?

Framework/PythonInterface/test/python/mantid/api/WorkspaceBinaryOpsTest.py
```diff
@@ -0,0 +1,38 @@
...
+        result6 = 1 + ws
+        self.assertTrue(mtd.doesExist('result6'))
+        result7 = ws_ads + 1
+        self.assertTrue(mtd.doesExist('result7'))
+        result8 = 1 + ws_ads
+        self.assertTrue(mtd.doesExist('result8'))
+        ws += 1
+        self.assertFalse(mtd.doesExist('ws'))
+        ws_ads += 1
+        self.assertTrue(mtd.doesExist('ws_ads'))
```
SimonHeybrock: Good testing :+1:

SimonHeybrock (changes requested):
Are there missing changes to a series `clone()` methods? For example, `EventWorkspace::clone()` does not seem to be refactored for the new behavior? I suspect running the test with, e.g.,` and `EventWorkspace` would fail?

Given that there are many workspace types with `clone()` that would need to be refactored, maybe clearing the name in the `Workspace` copy constructor is the easiest way to go (unless this is breaking something else?)?

Edit: Given that the copy constructor is not public and is used (I think) only for cloning, I think such a change would not break anything.

soininen:
@SimonHeybrock The changes to workspace cloning here are only half done, indeed. In addition to changing the cloning behavior of all workspaces, the Python documentation of `clone()` should be updated as well. As the changes to cloning are largish but independent of what this PR is (or should be) about, I think I'll open another issue and PR to handle `clone()`.

SimonHeybrock:
@soininen My point was rather that I think you can avoid all those changes to `clone()`.

soininen:
@SimonHeybrock Yes, I'll have a look if I can get away with just changing the copy constructors, as you suggested.

SimonHeybrock:
@soininen I think you only need to change it for `Workspace`, all derived classes should be unchanged, I think.

SimonHeybrock:
Changes look good now. Also verified that `CloneWorkspace` still puts a workspace with the requested name into the ADS (it is using `clone()` internally).

SimonHeybrock (approved):

PR #21084, opened 2017-10-31
SumOverlappingTubes Algorithm

### LINKED ISSUE #20285: Combine detector scans into a single workspace for D2B
As part of the D2B workflow a number of detector scans need to be merged into a single workspace. The new algoirthm should:

1. Accept a number of input workspaces with scan information (and possibly just start/end times)
1. Allow automatically or manually choosing a 2&theta; range
1. Choosing a step size between bins
1. Output a workspace of detector tube height/pixel number against 2&theta;
1. Any detector tubes within a certain tolerance should be put into the correct bin
1. Any detector tubes that are outside that tolerance should be split between bins

For now the algorithm should not need to do any normalisation.

ianbush:
This is a new algorithm to combine detectors scans for the D2B powder diffractometer. Please read the algorithm documentation for a full description of behaviour.

**To test:**

1. Read docs, do they explain what the algorithm does?
1. Test with attached files. Try different options, and different cases such as choosing two files or a group workspace. Use a colour fill plot to see the output in the 2D cases.
1. Code review.

Fixes #20285.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


mantid-roman:
There are conflicts here

Framework/Algorithms/src/RunCombinationHelpers/RunCombinationHelper.cpp
```diff
@@ -85,5 +86,72 @@ RunCombinationHelper::checkCompatibility(MatrixWorkspace_sptr ws,
...
+      ws = AnalysisDataService::Instance().retrieveWS<MatrixWorkspace>(
+          inputWorkspaces[i]);
+      if (!ws) {
+        g_log.error() << "Input workspace " << inputWorkspaces[i]
+                      << " not found.\n";
+        throw Kernel::Exception::NotFoundError("Data Object",
+                                               inputWorkspaces[i]);
+      }
+      inWS.push_back(ws);
+    } catch (Exception::NotFoundError &) {
```
mantid-roman: I understand that it's a cut and paste edit but while we are on it can we improve things a little?

At the moment if I pass a name of an existing table workspace to `SumOverlappingTubes` I get this list of errors:

```
Input workspace Table1 not found.
Input workspace Table1 not found.
Error in execution of algorithm SumOverlappingTubes:
Data Object search object Table1
```
The first message is from the `if (!ws) {...` block, the second one is from the `catch` clause and the third is logged by the Algorithm base class when the exception is re-thrown. I think logging in this method is unnecessary and the message is wrong: the workspace exists but its type is wrong.

docs/source/algorithms/SumOverlappingTubes-v1.rst
```diff
@@ -0,0 +1,86 @@
...
+-----------
+
+This algorithm takes an input workspace, a group workspace or a selection of workspaces, with a PSD tube instrument, and converts the workspace to a one or two-dimensional workspace of counts as a function of height and scattering angle. There are options to either see the data as it is, or apply a correction to straighten the Debye-Scherrer cones.
+
+This algorithm is written for the geometry of D2B, where the detector tubes have the same radial distance from the sample, and rotate around the sample in the horizontal plane during a detector scan. This algorithm could be generalised to support other instrument geometries in the future.
+
+General Behavior
+#################
+
+For the output workspace a grid is made, the x-axis is the scattering angle and the y-axis is the tube height. Counts from the input workspaces are then put onto this grid. Where one of the entries in the scattering angle is between two angles the count are split between the two, unless the ScatteringAngleTolerance option is set.
```
mantid-roman: The values on the y-axis can be negative. Is `tube height` the right name for them?

Framework/Algorithms/src/SumOverlappingTubes.cpp
```diff
@@ -0,0 +1,366 @@
...
+  return result;
+}
+
+void SumOverlappingTubes::exec() {
+  getInputParameters();
+
+  HistogramData::Points x(m_numPoints, LinearGenerator(m_startScatteringAngle,
+                                                       m_stepScatteringAngle));
+
+  MatrixWorkspace_sptr outputWS = create<Workspace2D>(m_numHistograms, x);
```
mantid-roman: The output workspace doesn't seem to have an instrument attached. Is it intentional?

mantid-roman (changes requested):

Framework/Algorithms/src/RunCombinationHelpers/RunCombinationHelper.cpp
ianbush: A good point. I will remove the try/catch as I don't think it is required, it should throw from `retrieveWS`, but in most cases it will be caught on input validation.

Framework/Algorithms/src/SumOverlappingTubes.cpp
ianbush: Yes, that is intentional. Because the counts are put onto a regular 2D grid they are sometimes split, so keeping an instrument associated would not make sense anymore.

docs/source/algorithms/SumOverlappingTubes-v1.rst
ianbush: Hmm, a good question. I think using height is consistent with the terminology used at the ILL, so I would prefer to leave it as that for now. I will clarifiy what this means here a bit though.

ianbush:
@mantid-roman I am just going to make a couple of minor changes to this so it can be generalised for use on D20, I will mark it as `InProgress` for the moment.

ianbush:
@mantid-roman This is ready to test again, sorry about making changes part way through.

The changes made were to support flipping the scattering angles in the output, i.e. multiplying by `-1`, so that D2B can have the expected output. This is done via setting `mirror_detector_angles` in the instrument parameters to true.

There is now an option to crop the negative scattering angles too, before in the `1DStraight` and `2DStraight` options the negative parts were just added to the positive scattering angle entries.

mantid-roman (approved):

PR #21089, opened 2017-10-31
Use IArchiveSearch more selectively

### LINKED ISSUE #20352: Check for data archive connection once
Looking for files in system tests when the data archive is not reachable (e.g. using ICAT from offsite) waits for a timeout from the service. Rather than waiting for the timeout with each file, add a static variable and associated instance method to `IArchiveSearch` objects which check the connection once (per startup of mantid) and allow for skipping the `IArchiveSearch` being used in `FileFinder`.

The actual work to be done:
1. Add a static variable `IArchiveSearch::m_canReachArchive` to the various `IArchiveSearch` classes.
2. Initialize the variable in the constructor of each constructors using [`std::call_once`](http://en.cppreference.com/w/cpp/thread/call_once) so the connection is only tested if the archive search is going to be used.
3. Add a instance method `bool IArchiveSearch::canReachArchive()` which returns the result.
4. Modify [`FileFinder`](https://github.com/mantidproject/mantid/blob/master/Framework/API/src/FileFinder.cpp#L482) to utilize this new method.

peterfpeterson:
Copied and rephrased from the issue:

The value of `datasearch.searcharchive` (case insensitive) will be interpreted differently in `FileFinder`:

* empty string or `off` means don't do archive search
* `on` means search your default facility only
* `all` means search in all facilities
* list of names means search facilities in the list (example is `hfir,sns` will search those and skip `isis`)

**To test:**

Turn logging up to `debug`, set your default facility to `isis` or `sns`, set your `datasearch.searcharchive` to various values, then search for `OSI12345` and `NOM12345`. I don't know if both exist, but you'll see from the logging whether or not it used archive search for the files.

Fixes #20352.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


rosswhitfield (changes requested):
All this appears to be working correctly!

But can you add the new options to the [ManageUserDirectories](https://www.mantidproject.org/ManageUserDirectories) UI?

At the moment it [will appear to be off while it may still be searching](https://github.com/mantidproject/mantid/blob/master/qt/widgets/common/src/ManageUserDirectories.cpp#L73-L76) and [then be overridden when you close the ManageUserDirectories](https://github.com/mantidproject/mantid/blob/master/qt/widgets/common/src/ManageUserDirectories.cpp#L90-L93)

peterfpeterson:
I'll change the setting to a drop-down list of "default facility", "all", "off", or the list of facilities (if the user has set it). I'll also fix that populating that bit of the dialog is case sensitive.

qt/widgets/common/src/ManageUserDirectories.cpp
```diff
@@ -82,15 +100,21 @@ void ManageUserDirectories::loadProperties() {
...
-  QString newSaveDir;
-
-  if (m_uiForm.ckSearchArchive->isChecked())
-    newSearchArchive = "On";
-  else
-    newSearchArchive = "Off";
+  QString newSearchArchive = m_uiForm.cbSearchArchive->currentText().toLower();
+  if (newSearchArchive == "all") {
+    // do nothing
+  } else if (newSearchArchive == "none") {
```
rosswhitfield: Wouldn't it be easier if the options matched between the ManageUserDirectories UI and the Mantid properties file?

qt/widgets/common/src/ManageUserDirectories.cpp
```diff
@@ -82,15 +100,21 @@ void ManageUserDirectories::loadProperties() {
...
+    newSearchArchive = "off";
+  } else if (newSearchArchive.startsWith("default facility only")) {
+    newSearchArchive = "on";
+  } else {
+    // the only way "custom" gets set is by using the value in ConfigService
+    // already, so just copy it
+    QString::fromStdString(Mantid::Kernel::ConfigService::Instance().getString(
+                               "datasearch.searcharchive"))
```
rosswhitfield: You are prepending `"custom - "` to the custom option above which is getting saved, after opening ManageUserDirectories a few time my properties file looks like
```
datasearch.searcharchive=custom - custom - custom - custom - custom - sns
```
😀

rosswhitfield (changes requested):

rosswhitfield (approved):

PR #21114, opened 2017-11-03
Load new D2B NeXus files

### LINKED ISSUE #20881: Allow LoadILLDiffraction to read calibrated data
NeXus files with calibrated data already applied will be produced for D2B (and eventually D20). This issue is to add the option to load the calibrated data.

ianbush:
The NeXus files for D2B have been updated* to provide calibrate data. This is stored in `data_scan/detector_data/data`, with the presence of the entry `data_scan/detector_data/raw_data` telling you the other data block is calibrated data.

Some scanned variables were also added with the same name, so the name when loaded into Mantid is now `name.property`.

*The NeXus files are not being used yet so they are still changed sometimes.

**To test:**

Check code and unit test logic. Try out the attached D2B and D20 files, check both load with no `DataType` given, the D2B loads for `Calibrated` or `Raw` and the D20 file loads for `Calibrated` and throws for `Raw`.

[example_data.zip](https://github.com/mantidproject/mantid/files/1440578/example_data.zip)

Fixes #20881.

**Release Notes** 

Updated.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/DataHandling/src/LoadILLDiffraction.cpp
```diff
@@ -701,5 +724,20 @@ LoadILLDiffraction::getInstrumentFilePath(const std::string &instName) const {
...
+/**
+ * Returns true if the file contains calibrated data
+ *
+ * @param filename The filename to check
+ * @return True if the file contains calibrated data, false otherwise
+ */
+bool LoadILLDiffraction::containsCalibratedData(
+    const std::string &filename) const {
+  NexusDescriptor descriptor(filename);
+  // This is unintuitive, but if the file has calibrated data there are entries
```
gvardany: :1st_place_medal:

gvardany (commented):
Looks good, just a comment about the default.
Maybe we should also add a sample log about what it loaded, but I'd rather do this in my reduction branch, where I added many other logs.

Framework/DataHandling/src/LoadILLDiffraction.cpp
```diff
@@ -93,6 +94,22 @@ void LoadILLDiffraction::init() {
   declareProperty(make_unique<WorkspaceProperty<MatrixWorkspace>>(
                       "OutputWorkspace", "", Direction::Output),
                   "The output workspace.");
+  std::vector<std::string> calibrationOptions{"", "Raw", "Calibrated"};
```
gvardany: So the default is "Raw"? Maybe we should have "Auto" as default, which will do what the empty string does now?

ianbush: The default is the empty string. I agree calling this `Auto` would be a bit nicer - makes it clear you should check what is going to happen!

ianbush:
Having a way to confirm whether the data is raw/calibrated is a good idea. The D2B files have a new log, `Detetor.calibration_file`. It is currently loaded in any case, but how about we set that to 'none' for loading raw data?

Framework/DataHandling/src/LoadILLDiffraction.cpp
```diff
@@ -94,8 +94,8 @@ void LoadILLDiffraction::init() {
   declareProperty(make_unique<WorkspaceProperty<MatrixWorkspace>>(
                       "OutputWorkspace", "", Direction::Output),
                   "The output workspace.");
-  std::vector<std::string> calibrationOptions{"", "Raw", "Calibrated"};
-  declareProperty("DataType", "Raw",
+  std::vector<std::string> calibrationOptions{"Auto", "Raw", "Calibrated"};
+  declareProperty("DataType", "Auto",
                   boost::make_shared<StringListValidator>(calibrationOptions),
                   "Select the type of data, with or without calibration "
                   "already applied. If blank then the calibrated data is "
```
gvardany: It is not blank anymore, maybe better to say `by default the calibrated data....`

Framework/DataHandling/src/LoadILLDiffraction.cpp
```diff
@@ -226,6 +227,14 @@ void LoadILLDiffraction::loadMetaData() {
     m_loadHelper.addNexusFieldsToWsRun(nxHandle, m_outWorkspace->mutableRun());
     nxStat = NXclose(&nxHandle);
   }
+
+  if (mutableRun.hasProperty("Detector.calibration_file")) {
+    if (getPropertyValue("DataType") == "Raw") {
+      mutableRun.getProperty("Detector.calibration_file")->setValue("none");
+      g_log.error() << "Set to none!" << std::endl;
```
gvardany: I hope this is a leftover?

Framework/DataHandling/src/LoadILLDiffraction.cpp
ianbush: This is now fixed, blank -> Auto.

Framework/DataHandling/src/LoadILLDiffraction.cpp
ianbush: Whoops, indeed it is!

gvardany (approved):
Should be good now.

PR #21118, opened 2017-11-03
Remove crash in debug build in LoadISISNexus

### LINKED ISSUE #21116: LoadISISNexus2 trips assert statement in std::sort comparator on OSX
This seems to be a weird case of implementation defined behaviour cause this crash to be seen only on some systems. The `assert` is triggered because the implementation of `std::sort` in `libc++` seems to compare the same item against itself.

### Expected behavior
`Load` should load PEARL files correctly in debug mode.

### Actual behavior
`Load` crashes in debug mode on the assert statement at line 700 in LoadISISNexus2

 - Compile in debug mode using `libc++`
 - Run `Load` on PEARL 97532
 - It should trip the assert statement in this file:

https://github.com/mantidproject/mantid/blob/528410251c830ce5cdc7b8124b3f5b8b682fd7a9/Framework/DataHandling/src/LoadISISNexus2.cpp#L700-L710
 
### Platforms affected
Seems to trip on OSX with the `libc++` implementation

samueljackson92:
This removes an assert statement that could be tripped depending on the implementation details of `std::sort`. 

**To test:**

See instructions in issue for how to reproduce. Data file is attached.
[PEARL00097532.nxs.zip](https://github.com/mantidproject/mantid/files/1441492/PEARL00097532.nxs.zip)

Fixes #21116 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/DataHandling/inc/MantidDataHandling/LoadISISNexus2.h
+++ b/Framework/DataHandling/inc/MantidDataHandling/LoadISISNexus2.h
@@ -147,6 +147,9 @@ private:
   void buildSpectraInd2SpectraNumMap(bool range_supplied, bool hasSpectraList,
                                      DataBlockComposite &dataBlockComposite);
 
+  /// Check if any of the spectra block ranges overlap
+  void checkOverlappingSpectraRange();
+
   /// The name and path of the input file
   std::string m_filename;
   /// The instrument name from Nexus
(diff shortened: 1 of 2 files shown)


soininen (approved):
The crash when loading the test file is gone after applying the changes in this PR. The code looks fine to me, as well.

PR #21131, opened 2017-11-06
Add support for iterating over a Histogram

### LINKED ISSUE #21130: Add an iterator interface to HistogramData
After discussions with simon, this would be a nice feature and it could potentially provide a performance increase where you only need to provide conversions on a small subset of points/bins from a large Histogram (e.g. all the non-zero values).

samueljackson92:
Add support for iterating over a `Histogram`. 

Note that there will be a increase in the performance test for HistogramTestPerformance as a new test case has been added.

**To test:**
 - Unit tests pass
 - Code review

Fixes #21130.

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/HistogramData/inc/MantidHistogramData/HistogramItem.h
```diff
@@ -0,0 +1,81 @@
...
+
+private:
+  const Histogram &histogramRef() const;
+  // Deleted copy & assignment operators as a HistogramItem is not copyable
+  HistogramItem(const HistogramItem &) = delete;
+  HistogramItem operator=(const HistogramItem &) = delete;
+
+  const Histogram &m_histogram;
+  const size_t m_index;
+};
```
SimonHeybrock: Without having tried, I would say that `HistogramItem` would probably make sense to have all its methods `inline`, given that they are all very small and called for every bin in every histogram of a workspace, i.e., billions of times.

Framework/HistogramData/inc/MantidHistogramData/HistogramIterator.h
```diff
@@ -0,0 +1,72 @@
...
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+class MANTID_HISTOGRAMDATA_DLL HistogramIterator
+    : public boost::iterator_facade<HistogramIterator, HistogramItem,
+                                    boost::bidirectional_traversal_tag> {
+
+public:
+  HistogramIterator(const Histogram &histogram, const size_t index = 0)
+      : m_histogram(histogram), m_index(index),
+        m_currentItem(Kernel::make_unique<HistogramItem>(histogram, index)){};
```
SimonHeybrock: Why the `unique_ptr`? Can we avoid the memory allocation?

Framework/HistogramData/inc/MantidHistogramData/HistogramIterator.h
```diff
@@ -0,0 +1,72 @@
...
+
+  void increment();
+  bool equal(const HistogramIterator &other) const;
+  HistogramItem &dereference() const;
+  void decrement();
+  void advance(int64_t delta);
+  uint64_t distance_to(const HistogramIterator &other) const;
+
+  const Histogram &m_histogram;
+  size_t m_index;
```
SimonHeybrock: The histogram reference and the index are also stored in the `HistogramItem`, is there a need for the duplication or could it be avoided?

Framework/HistogramData/inc/MantidHistogramData/HistogramIterator.h
```diff
@@ -0,0 +1,72 @@
...
+  std::unique_ptr<HistogramItem> m_currentItem;
+};
```
SimonHeybrock: Just like `HistogramItem`, I would think having everything `inline` would make sense.

Framework/HistogramData/src/HistogramItem.cpp
```diff
@@ -0,0 +1,106 @@
...
+using Mantid::HistogramData::BinEdges;
+using Mantid::HistogramData::Points;
+
+HistogramItem::HistogramItem(const Histogram &histogram, const size_t index)
+    : m_histogram(histogram), m_index(index) {}
+
+double HistogramItem::counts() const {
+  const auto &yMode = histogramRef().yMode();
+  if (yMode == Histogram::YMode::Counts) {
+    return histogramRef().counts()[m_index];
```
SimonHeybrock: I think all these methods should avoid calling the higher-level methods like `counts()` and `binEdges()`, since they return by value, which, while not a copy of the data, takes a copy of the `cow_ptr` (internally a `shared_ptr`). Doing so once per histogram is usually not too expensive, but here it is done once for every bin so I think we must avoid it to see good performance.

That is, work with `x`, y`, etc. from `Histogram` directly.

Framework/HistogramData/src/HistogramItem.cpp
```diff
@@ -0,0 +1,106 @@
...
+  } else {
+    const Frequencies frequency{histogramRef().frequencies()[m_index]};
+    return Counts(frequency, binEdges())[0];
+  }
+}
+
+double HistogramItem::countVariance() const {
+  return histogramRef().countVariances()[m_index];
```
SimonHeybrock: This can be terribly expensive, computing all the variances. Same goes for several of the other methods below.

Framework/HistogramData/src/HistogramItem.cpp
```diff
@@ -0,0 +1,106 @@
...
+  return histogramRef().countStandardDeviations()[m_index];
+}
+
+double HistogramItem::frequency() const {
+  const auto &yMode = histogramRef().yMode();
+  if (yMode == Histogram::YMode::Frequencies) {
+    return histogramRef().frequencies()[m_index];
+  } else {
+    const Counts counts{histogramRef().counts()[m_index]};
+    return Frequencies(counts, binEdges())[0];
```
SimonHeybrock: Similar to what I said above: For performance reasons we cannot rely on all the math done by classes like `Frequencies`, you need to implement this yourself, based on the underlying data.

Framework/HistogramData/src/HistogramItem.cpp
```diff
@@ -0,0 +1,106 @@
...
+double HistogramItem::center() const { return point()[0]; }
+
+double HistogramItem::width() const {
+  const auto &edges = binEdges();
+  const auto &lower = edges[0];
+  const auto &upper = edges[1];
+  return upper - lower;
+}
+
+const BinEdges HistogramItem::binEdges() const {
```
SimonHeybrock: I hesitated for a second here, the return type kind of works, but in the end I think it is probably too confusing. Additionally, creating `BinEdges` involves two memory allocation (`shared_ptr` and `vector`) so it should be too expensive for this purpose.

Same goes for using `Points` below.

Framework/HistogramData/src/HistogramItem.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+  if (xMode == Histogram::XMode::BinEdges) {
+    const auto &edges = histogramRef().binEdges();
+    const auto &lower = edges[m_index];
+    const auto &upper = edges[m_index + 1];
+    return BinEdges{lower, upper};
+  } else {
+    const auto &points = histogramRef().points();
+    // Guess the bin edges from the points
+    if (points.size() == 0) {
```
SimonHeybrock: In that case `histogram.begin() == histogram.end()`, i.e., this should never happen.

Framework/HistogramData/src/HistogramItem.cpp
```diff
@@ -0,0 +1,106 @@
...
+      // We have no points, just return empty bin edges
+      return BinEdges{};
+    } else if (points.size() == 1) {
+      // We have a single point, let BinEdges guess the edges
+      return BinEdges(points);
+    } else {
+      // We have 2 or more points. Attempt to use up to 4 points
+      // to interpolate the correct bin edges.
```
SimonHeybrock: Nice idea, however this is not what what the old conversion from points to bin edges did. `HistogramData` is implemented to follow the old conversion for keeping things consistent. Thus, we have to do the same here.

If you want to do such a change it would need to be done consistently for all cases (legacy implementation, current histogram, and new iterator), but it would definitely need to go by the TSC, and I predict mixed opinions.

In any case, such a change should not be part of this PR.

Framework/HistogramData/src/HistogramIterator.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+using Mantid::HistogramData::Histogram;
+using Mantid::HistogramData::HistogramItem;
+using Mantid::HistogramData::HistogramIterator;
+using Mantid::Kernel::make_unique;
+
+void HistogramIterator::increment() {
+  if (m_index < m_histogram.size()) {
+    ++m_index;
+    m_currentItem = make_unique<HistogramItem>(m_histogram, m_index);
```
SimonHeybrock: `m_histogram` should not change, just the index so why create a new `HistogramItem` (also see note above on duplicate data in iterator)?

Framework/HistogramData/src/HistogramIterator.cpp
```diff
@@ -0,0 +1,41 @@
...
+  }
+}
+
+bool HistogramIterator::equal(const HistogramIterator &other) const {
+  return (&m_histogram == &other.m_histogram) && (m_index == other.m_index);
+}
+
+HistogramItem &HistogramIterator::dereference() const { return *m_currentItem; }
```
SimonHeybrock: `const HistogramItem`?

Framework/HistogramData/test/HistogramItemTest.h
```diff
@@ -0,0 +1,160 @@
...
+using Mantid::HistogramData::BinEdges;
+
+class HistogramItemTest : public CxxTest::TestSuite {
+public:
+  // This pair of boilerplate methods prevent the suite being created statically
+  // This means the constructor isn't called when running other tests
+  static HistogramItemTest *createSuite() { return new HistogramItemTest(); }
+  static void destroySuite(HistogramItemTest *suite) { delete suite; }
+
+  static constexpr double tolerance = 1e-6;
```
SimonHeybrock: Why so large? Doesn't `1e-12` or `1e-14` work?

Framework/HistogramData/test/HistogramItemTest.h
```diff
@@ -0,0 +1,160 @@
...
+  void test_get_countVariance_from_histogram_with_counts() {
+    Histogram hist(Points{0.1, 0.2, 0.4}, Counts{1, 2, 4});
+    HistogramItem item(hist, 1);
+    TS_ASSERT_DELTA(item.countVariance(), 2.0, tolerance)
+  }
+
+  void test_get_countVariance_from_histogram_with_frequencies() {
+    Histogram hist(Points{0.1, 0.2, 0.4}, Frequencies{1, 2, 4});
+    HistogramItem item(hist, 1);
+    TS_ASSERT_DELTA(item.countVariance(), 0.045, 1e-12)
```
SimonHeybrock: I think I would write most (or all?) of these test simply comparing to elements of the corresponding method on `Histogram`, for example:
```cpp
TS_ASSERT_DELTA(item.countVariance(), hist.countVariance()[index]);
```

Edit:
Maybe doing all the testing on the iterator is more useful/simple, testing the whole range? Consider changing `HistogramItem` such that it can be constructed only by the iterator (no one else needs to?).

Framework/HistogramData/test/HistogramTest.h
```diff
@@ -1108,6 +1108,15 @@ class HistogramTest : public CxxTest::TestSuite {
     TS_ASSERT(!h.sharedE());
     TS_ASSERT(!h.sharedDx());
   }
+
+  void test_that_can_iterate_hsitogram() {
```
SimonHeybrock: typo

Framework/HistogramData/test/HistogramTest.h
```diff
@@ -1108,6 +1108,15 @@ class HistogramTest : public CxxTest::TestSuite {
...
+    Histogram hist(Points{0.1, 0.2, 0.4}, Counts{1, 2, 4});
+    double total = 0;
+    for (const auto &item : hist) {
+      total += item.counts();
+    }
+    TS_ASSERT_EQUALS(total, 7)
+  }
```
SimonHeybrock: Test also:
- `begin()` and `end()` correct for both cases, point-histograms and normal histograms.
- Iterator comparison.
- Adding/subtracting from iterators.
- More?

Framework/HistogramData/test/HistogramTest.h
```diff
@@ -1119,8 +1128,9 @@ class HistogramTestPerformance : public CxxTest::TestSuite {
 
   HistogramTestPerformance() : xData(histSize, LinearGenerator(0, 2)) {
     BinEdges edges(histSize, LinearGenerator(0, 2));
+    Counts counts(histSize - 1, LinearGenerator(0, 2));
     for (size_t i = 0; i < nHists; i++)
-      hists.push_back(Histogram(edges));
+      hists.push_back(Histogram(edges, counts));
```
SimonHeybrock: Why do you need the data in a performance test? `Histogram` initializes it to zero and that should be fine, I think.

Framework/HistogramData/test/HistogramTest.h
```diff
@@ -1140,6 +1150,15 @@ class HistogramTestPerformance : public CxxTest::TestSuite {
       i.setSharedX(x);
   }
 
+  void test_iterate() {
+    double total = 0;
+    for (size_t i = 0; i < nHists / 10; ++i) {
```
SimonHeybrock: I suppose this is quite slow right now?

SimonHeybrock (changes requested):
Hi Sam! Good start, but several comments:

Framework/HistogramData/inc/MantidHistogramData/HistogramIterator.h
samueljackson92: I want to avoid this but I'm not sure how. I can't simply create and return a reference to `HistogramItem` as that would mean returning reference to a stack variable. `HistogramItem` is not copyable so I can't return a copy. I also can't have an instance variable `HistogramItem` because the `HistogramIterator` is copied when returned from the `begin()` call.

Framework/HistogramData/src/HistogramItem.cpp
samueljackson92: Ah ok. I didn't want to create any extra work for myself, but if there's a likely performance gain then that's fine.

Framework/HistogramData/src/HistogramItem.cpp
samueljackson92: Again, I thought it might be better to let the Histogram handle the computations itself, but if there's a likely to be a large performance hit then I'll do it manually.

Framework/HistogramData/src/HistogramItem.cpp
samueljackson92: I see that this is different and agree it should not be changed here. I'll make sure it is consistent.

Framework/HistogramData/inc/MantidHistogramData/HistogramIterator.h
SimonHeybrock: But the pointer is not solving that, is it? You will have an invalided reference to `HistogramItem` once you, e.g., increment the iterator.

Framework/HistogramData/test/HistogramTest.h
samueljackson92: Not particularly fast.

Framework/HistogramData/inc/MantidHistogramData/HistogramIterator.h
SimonHeybrock: Why not just store `HistogramItem` in `HistogramIterator` (instead of `m_histogram` and `m_index`) and return a reference to that?

samueljackson92: Ok, but what about the last point? `begin` constructs and returns a `HistogramIterator`. If it has an instance of `HistogramItem` then `HistogramIterator` has an implicitly deleted copy constructor?

SimonHeybrock: Right, I guess then `HistogramItem` should be copyable after all?

samueljackson92: I guess. 😄

Framework/HistogramData/src/HistogramItem.cpp
samueljackson92: Writing a test for the case that `histogram.begin() == histogram.end()` it seems that `histogram.size()` does not correctly report `0`. For example, the following fails:
```c++
    Histogram hist(Histogram::XMode::BinEdges, Histogram::YMode::Counts);
    assert(hist.size() == 0);
```

SimonHeybrock: Actually I fixed that yesterday in https://github.com/mantidproject/mantid/tree/21128_ExtractSpectra_refactor since it also causes trouble there. In general this cannot (could not) happen in Mantid because `MatrixWorkspace` does not allow creation with X or Y length 0, so I had not paid much attention to that case. However, with support for non-constant histogram size added recently this becomes more relevant.

I will extract the relevant changes from that branch and make it a separate PR (since the bigger changes in the aforementioned branch will take slightly longer to get into `master`).

SimonHeybrock: @samueljackson92 I have assigned https://github.com/mantidproject/mantid/pull/21153 to you.

samueljackson92: Awesome. Thanks.

Framework/HistogramData/inc/MantidHistogramData/HistogramItem.h
```diff
@@ -0,0 +1,99 @@
...
+  double frequency() const;
+  double frequencyVariance() const;
+  double frequencyStandardDeviation() const;
+  double width() const;
+  double center() const;
+
+  void incrementIndex();
+  void advance(int64_t delta);
+
+  inline void decrementIndex() {
```
SimonHeybrock: Defining a method inside the class declaration implies `inline`, i.e., you do not need the keyword.

Framework/HistogramData/inc/MantidHistogramData/HistogramItem.h
```diff
@@ -0,0 +1,99 @@
...
+
+  inline size_t getIndex() const { return m_index; }
+
+  inline void setIndex(const size_t index) { m_index = index; }
+
+private:
+  friend class HistogramIterator;
+  /// Private constructor, can only be created by HistogramIterator
+  HistogramItem(const Histogram &histogram, const size_t index);
+  /// Get a refernce to the histogram
```
SimonHeybrock: Typo.

Framework/HistogramData/inc/MantidHistogramData/HistogramIterator.h
```diff
@@ -0,0 +1,80 @@
...
+                                    boost::bidirectional_traversal_tag> {
+
+public:
+  HistogramIterator(const Histogram &histogram, const size_t index = 0)
+      : m_item(histogram, index){};
+
+private:
+  friend class boost::iterator_core_access;
+
+  inline void increment() { m_item.incrementIndex(); }
```
SimonHeybrock: As above, can drop all the `inline` keywords.

Framework/HistogramData/src/HistogramItem.cpp
```diff
@@ -0,0 +1,127 @@
...
+  const auto &e = histogramRef().e();
+  if (!yModeIsCounts()) {
+    return e[m_index];
+  } else {
+    const auto binWidth = width();
+    return e[m_index] / binWidth;
+  }
+}
+
+const Histogram &HistogramItem::histogramRef() const { return m_histogram; }
```
SimonHeybrock: Why is this wrapped in a function call?

Framework/HistogramData/src/HistogramItem.cpp
```diff
@@ -0,0 +1,127 @@
...
+double HistogramItem::center() const {
+  const auto &x = histogramRef().x();
+  if (xModeIsPoints()) {
+    return x[m_index];
+  } else {
+    return (x[m_index + 1] + x[m_index]) / 2.0;
+  }
+}
+
+double HistogramItem::width() const {
```
SimonHeybrock: `width()` or `binWidth()`?

Framework/HistogramData/src/HistogramItem.cpp
```diff
@@ -0,0 +1,127 @@
...
+      upper = 0.5 * (x[m_index + 1] + x[m_index]);
+      lower = x[0] - (upper - x[0]);
+    } else if (m_index == numPoints - 1) {
+      // last point
+      lower = 0.5 * (x[m_index] + x[m_index - 1]);
+      upper = x[numPoints - 1] + (x[numPoints - 1] - lower);
+    } else {
+      // everything inbetween
+      lower = 0.5 * (x[m_index] + x[m_index - 1]);
+      upper = 0.5 * (x[m_index + 1] + x[m_index]);
```
SimonHeybrock: `return 0.5*(x[m_index + 1] - x[m_index - 1])`

Framework/HistogramData/src/HistogramItem.cpp
```diff
@@ -0,0 +1,127 @@
...
+double HistogramItem::width() const {
+  const auto &x = histogramRef().x();
+  if (xModeIsPoints()) {
+    auto numPoints = histogramRef().size();
+    double lower = 0;
+    double upper = 0;
+    if (m_index == 0) {
+      // first point
+      upper = 0.5 * (x[m_index + 1] + x[m_index]);
+      lower = x[0] - (upper - x[0]);
```
SimonHeybrock: `return x[1] - x[0]`

Framework/HistogramData/src/HistogramItem.cpp
```diff
@@ -0,0 +1,127 @@
...
+    } else if (m_index == numPoints - 1) {
+      // last point
+      lower = 0.5 * (x[m_index] + x[m_index - 1]);
+      upper = x[numPoints - 1] + (x[numPoints - 1] - lower);
```
SimonHeybrock: `return x[m_index] - x[m_index - 1]`

Framework/HistogramData/test/HistogramIteratorTest.h
```diff
@@ -0,0 +1,412 @@
...
+using Mantid::HistogramData::HistogramIterator;
+using Mantid::HistogramData::Points;
+using Mantid::HistogramData::Counts;
+using Mantid::HistogramData::Frequencies;
+using Mantid::HistogramData::BinEdges;
+using Mantid::HistogramData::CountVariances;
+using Mantid::HistogramData::CountStandardDeviations;
+using Mantid::HistogramData::FrequencyVariances;
+using Mantid::HistogramData::FrequencyStandardDeviations;
+using Mantid::HistogramData::LinearGenerator;
```
SimonHeybrock: Maybe `using namespace Mantid::HistogramData` is in order here?

Framework/HistogramData/test/HistogramIteratorTest.h
```diff
@@ -0,0 +1,412 @@
...
+    Histogram hist(Points{1, 2, 3}, expectedFrequencies);
+
+    // Act
+    auto result =
+        std::equal(hist.begin(), hist.end(), expectedFrequencies.begin(),
+                   [](const HistogramItem &item, const double &frequency) {
+                     return item.frequency() == frequency;
+                   });
+
+    // Assert
```
SimonHeybrock: I don't think comments like this add a lot? They just make the code longer.

Framework/HistogramData/test/HistogramIteratorTest.h
```diff
@@ -0,0 +1,412 @@
...
+    TSM_ASSERT("Frequencies did not match", result);
```
SimonHeybrock: Also just a minor (matter of taste?) suggestion: You have a test that tests only a single thing, so the test name alone gives a good indication as to what may have gone wrong. Personally I would simply write

```TS_ASSERT(std::equal(...));```

For me that makes clear what is going on, and if you have a failure you get the name of the failed test from ctest, telling you what went wrong. No need to write fancy assert messages, I think.

File: Framework/HistogramData/test/HistogramIteratorTest.h [hunk 30]
```diff
@@ -0,0 +1,412 @@
...
+                   });
+
+    // Assert
+    TSM_ASSERT("Bin widths did not match", result);
+  }
+
+  void test_iterate_over_histogram_count_variances_when_histogram_has_counts() {
+    // Arrange
+    Histogram hist(BinEdges{1, 2, 3, 5}, Counts{2, 3, 4},
+                   CountVariances{3, 2, 1});
```
SimonHeybrock: Note that `Histogram` computes the uncertainties automatically if you do not specify them.

Framework/HistogramData/test/HistogramIteratorTest.h
```diff
@@ -0,0 +1,412 @@
...
+    Counts counts(histSize - 1, LinearGenerator(0, 2));
+    for (size_t i = 0; i < nHists; i++)
+      hists.push_back(Histogram(edges, counts));
+  }
+
+  void test_iterate_and_access_each_item() {
+    double total = 0;
+    for (auto &hist : hists)
+      for (auto &item : hist)
+        total += item.frequency();
```
SimonHeybrock: Would be interesting to write exactly the same test without iterators so we can compare the performance.

Framework/HistogramData/test/HistogramIteratorTest.h
```diff
@@ -0,0 +1,412 @@
...
+  HistogramIteratorTestPerformance() {
+    BinEdges edges(histSize, LinearGenerator(0, 2));
+    Counts counts(histSize - 1, LinearGenerator(0, 2));
+    for (size_t i = 0; i < nHists; i++)
+      hists.push_back(Histogram(edges, counts));
+  }
+
+  void test_iterate_and_access_each_item() {
+    double total = 0;
+    for (auto &hist : hists)
```
SimonHeybrock: I think I would reuse the same histogram. Using different histograms makes mainly sense if you are interested in the memory bandwidth of your computer, I think.

SimonHeybrock (changes requested):
Good update! Some more detailed comments:

Framework/HistogramData/src/HistogramItem.cpp
samueljackson92: I think I prefer `binWidth`. More specific.

Framework/HistogramData/inc/MantidHistogramData/HistogramItem.h
```diff
@@ -0,0 +1,97 @@
...
+
+public:
+  double counts() const;
+  double countVariance() const;
+  double countStandardDeviation() const;
+  double frequency() const;
+  double frequencyVariance() const;
+  double frequencyStandardDeviation() const;
+  double binWidth() const;
+  double center() const;
```
SimonHeybrock: I would recommend inlining all methods of `HistogramItem`. It probably makes a performance difference, given that the methods are called for every single bin. If you are concerned about having too much code in the header that is included by basically all algorithms via `MatrixWorkspace.h`, do not include `HistogramIterator.h` in `Histogram.h` (implying that `Histogram::begin()` and `Histogram::end()` cannot be inline, which should be less of an issue). Anyone using iteration will then need to include that header directly.

Framework/HistogramData/inc/MantidHistogramData/HistogramIterator.h
```diff
@@ -0,0 +1,80 @@
...
+
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+class MANTID_HISTOGRAMDATA_DLL HistogramIterator
+    : public boost::iterator_facade<HistogramIterator, const HistogramItem &,
+                                    boost::bidirectional_traversal_tag> {
+
+public:
+  HistogramIterator(const Histogram &histogram, const size_t index = 0)
```
SimonHeybrock: Is the default argument needed?

Framework/HistogramData/src/HistogramItem.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+HistogramItem::HistogramItem(const Histogram &histogram, const size_t index)
+    : m_histogram(histogram), m_index(index) {}
+
+double HistogramItem::center() const {
+  const auto &x = m_histogram.x();
+  if (xModeIsPoints()) {
+    return x[m_index];
+  } else {
+    return (x[m_index + 1] + x[m_index]) / 2.0;
```
SimonHeybrock: Division is slow. Use `0.5 * xxx` instead of `xxx / 2.0`.

Framework/HistogramData/src/HistogramItem.cpp
```diff
@@ -0,0 +1,122 @@
...
+  }
+}
+
+double HistogramItem::binWidth() const {
+  const auto &x = m_histogram.x();
+  if (xModeIsPoints()) {
+    auto numPoints = m_histogram.size();
+    double lower = 0;
+    double upper = 0;
```
SimonHeybrock: Unused.

Framework/HistogramData/src/HistogramItem.cpp
```diff
@@ -0,0 +1,122 @@
...
+      // first point
+      return x[1] - x[0];
+    } else if (m_index == numPoints - 1) {
+      // last point
+      return x[m_index] - x[m_index - 1];
+    } else {
+      // everything inbetween
+      return 0.5 * (x[m_index + 1] - x[m_index - 1]);
+    }
+    return upper - lower;
```
SimonHeybrock: Unreachable.

Framework/HistogramData/test/HistogramIteratorTest.h
```diff
@@ -0,0 +1,329 @@
...
+      }
+    }
+  }
+
+  void test_convert_counts_to_frequency_once_per_histogram_sparse() {
+    double total = 0;
+    for (size_t i = 0; i < nHists; i++) {
+      auto counts = m_hist.frequencies();
+      auto frequencies = m_hist.frequencies();
+      for (size_t j = 0; j < nHists; ++j)
```
SimonHeybrock: `histSize`?

Also you are getting `frequencies()` twice for some reason? I suppose you wanted `counts()`?

SimonHeybrock (changes requested):
This is almost there, mainly consider/try inlining for hopefully better performance:

SimonHeybrock (approved):

SimonHeybrock:
@samueljackson92 Nice work. How are the performance tests performaning now?

PR #21144, opened 2017-11-08
Enabled config file in ISIS Powder scripts HRPD

### LINKED ISSUE #21142: ISIS Powder HRPD scripts do not support a config file
The ISIS_Powder documentation says that all supported instruments can be used with a config file instead of having to pass in parameters which never change every time the scripts are used. This feature is not implemented on HRPD, but is nice to have as it saves instrument scientists time

josephframsay:
The `config_file` parameter of an instrument in `isis_powder` was exposed to HRPD, meaning the instrument object can be initialised with a yaml file instead of passing in all the required config as parameters.

**To test:**
Create a file called `config.yaml` somewhere with the following contents:
```
user_name: "Polly"
```
Running the following script should raise `AttributeError: The parameter with name: 'calibration_directory' is required but was not set or passed.`. This means that the config file was found by HRPD but is missing some parameters (which is fine, because it is).

```
from isis_powder import HRPD
hrpd = HRPD(config_file="path/to/config.yaml")
```

Fixes #21142 
Addresses part of #21148 

**Release Notes** 
Does not need to be in the release notes, as should have worked all along

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/scripts/Diffraction/isis_powder/hrpd_routines/hrpd_param_mapping.py
+++ b/scripts/Diffraction/isis_powder/hrpd_routines/hrpd_param_mapping.py
@@ -7,6 +7,7 @@ attr_mapping = \
     [
         ParamMapEntry(ext_name="calibration_directory",     int_name="calibration_dir"),
         ParamMapEntry(ext_name="calibration_mapping_file",  int_name="cal_mapping_path"),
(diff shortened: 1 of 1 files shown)


ElliotAOram (approved):
Performed the test as described and output was as expected. Simple change, all good!

PR #21146, opened 2017-11-08
Scanning ComponentInfo

### LINKED ISSUE #20986: Scanning ComponentInfo
Time dependent positions/rotations for ALL components.

OwenArnold:
This PR provides Sync scan functionality at the ComponentInfo level. Async scans are not supported (originally added, then removed). Things to note:

* Only components holding detectors are currently allowed to scan.
* `ComponentInfo` forwards to `DetectorInfo` detector scanning, the two implementations are separate.
* merging two `ComponentInfos` will result in a nested merge of associated `DetectorInfos`
* Copy behaviour for `ComponentInfo` has been deleted. Since `ComponentInfo` has non-owning pointer to `DetectorInfo` consistent copies via copy construction/assignment are not possible

Fixes #20986

**Tester**
Code review should be sufficient. Should be tested by @ianbush


Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -266,6 +362,20 @@ void ComponentInfo::failIfScanning() const {
...
+ComponentInfo::linearIndex(const std::pair<size_t, size_t> &index) const {
+  // The most common case are beamlines with static components. In that case the
+  // time index is always 0 and we avoid expensive map lookups. Linear indices
+  // are ordered such that the first block contains everything for time index 0
+  // so even in the time dependent case no translation is necessary.
+  if (index.second == 0)
+    return index.first;
+  // Calculate the linear index without a lookup
+  if (m_isSyncScan)
+    return index.first + size() * index.second;
```
SimonHeybrock: Is `size()` correct here, since it includes detectors, I think?

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -359,5 +471,112 @@ bool ComponentInfo::isStructuredBank(const size_t componentIndex) const {
...
+  const auto detectorScanSize = m_detectorInfo ? m_detectorInfo->scanSize() : 0;
+  if (!m_positions)
+    return 0 + detectorScanSize;
+  return m_positions->size() + detectorScanSize;
+}
+
+bool ComponentInfo::isScanning() const {
+  if (m_detectorInfo && m_detectorInfo->isScanning()) {
+    return true;
+  } else if (!m_positions)
```
SimonHeybrock: Weird bracing.

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -213,6 +243,35 @@ void ComponentInfo::setPosition(const size_t componentIndex,
...
+  // Optimization: Split loop over detectors and other components.
+  const auto range = detectorRangeInSubtree(componentIndex);
+  for (const auto &subIndex : range) {
+    m_detectorInfo->setPosition(
+        {subIndex, timeIndex},
+        m_detectorInfo->position({subIndex, timeIndex}) + offset);
+  }
+
+  for (const auto &index : componentRangeInSubtree(componentIndex)) {
+    size_t offsetIndex = compOffsetIndex(index);
```
SimonHeybrock: This method does not exist (yet?)?

SimonHeybrock (commented):
I did not look at all details at this point, but this is looking very reasonable. For now just small comments:

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -113,21 +114,46 @@ ComponentInfo::componentsInSubtree(const size_t componentIndex) const {
 size_t ComponentInfo::size() const { return m_size; }
 
 Eigen::Vector3d ComponentInfo::position(const size_t componentIndex) const {
+  checkNoTimeDependence();
```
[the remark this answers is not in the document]
OwenArnold: Agreed. This would only be relevant if there was scanning of a non-detector component. Based on our previous discussion about scanning samples/sources (which we agreed to forbid) I don't see that happening. However I'd like to keep this check in place until I am confident that we are reasonably able to verify that we cannot have a scanning ComponentInfo without a scanning DetectorInfo.

Framework/Beamline/src/ComponentInfo.cpp
OwenArnold: Good point.

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -113,21 +136,46 @@ ComponentInfo::componentsInSubtree(const size_t componentIndex) const {
...
 
+Eigen::Vector3d
+ComponentInfo::position(const std::pair<size_t, size_t> &index) const {
+
+  const auto componentIndex = index.first;
+  if (isDetector(componentIndex)) {
+    // Time index info must be same between detector and component infos!
+    return m_detectorInfo->position(index);
+  }
+  const auto rangesIndex = compOffsetIndex(componentIndex);
```
SimonHeybrock: Would it make sense to merge `compOffsetIndex()` and `linearIndex()`? Do you ever need them separately?

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -359,5 +511,235 @@ bool ComponentInfo::isStructuredBank(const size_t componentIndex) const {
...
+}
+
+bool ComponentInfo::isScanning() const {
+  if (m_detectorInfo && m_detectorInfo->isScanning())
+    return true;
+  else if (!m_positions || !m_componentRanges)
+    return false;
+  else
+    // TODO what if you have a scan that causes no position change, such as a
+    // rotation. Not scanning???
```
SimonHeybrock: Any scan will (should?) duplicate all parameters, even if they are kept fixed. At least that is what `DetectorInfo` is doing, I think.

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -359,5 +511,235 @@ bool ComponentInfo::isStructuredBank(const size_t componentIndex) const {
...
+  if (!m_scanCounts)
+    this->initScanCounts();
+  const auto toMerge = buildMergeIndices(other);
+  if (m_isSyncScan)
+    throw std::runtime_error(
+        "Sync scans not currently supported for ComponentInfo");
+  initIndices();
+  auto scanCounts(m_scanCounts);
+  for (size_t linearIndex = 0; linearIndex < other.m_positions->size();
+       ++linearIndex) {
```
SimonHeybrock: We had discussed forbidding scanning anything that is not something like a detector bank (in particular source and sample). How is that handled here?

SimonHeybrock (commented):
A couple of questions, but things seem to be moving in the right direction. Do you have plans already regarding how to handle/prevent scanning of specific components?

OwenArnold:
@SimonHeybrock. Yes I will prevent scanning of source and sample as we discussed already. That should be easy to do as I already know what the component index is for those components and can prevent at the point that the `ComponentInfo::setScanInterval` is called.

Framework/Beamline/src/ComponentInfo.cpp
OwenArnold: If that's happening, that is OK. I think you're right, this should happen in the `merge`

SimonHeybrock:
@OwenArnold Maybe you could also simply check the size of the detectors-in-subtree range? That would be more generic and prevent choppers-scans or similar?

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -642,27 +656,48 @@ this has no time dependence prior to this operation.
...
+      size_t indexEnd = indexStart + nonDetectorSize();
+      positions.insert(positions.end(), other.m_positions->begin() + indexStart,
+                       other.m_positions->begin() + indexEnd);
+      rotations.insert(rotations.end(), other.m_rotations->begin() + indexStart,
+                       other.m_rotations->begin() + indexEnd);
+    }
+
+  } else {
+    const auto toMerge = buildMergeIndices(other);
+    if (m_isSyncScan)
```
SimonHeybrock: Something is in the wrong order here. This in within the `else` block of `if (m_isSyncScan)`.

But I suppose it will need to change anyway, based on our discussion on Slack yesterday.

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -642,27 +656,48 @@ this has no time dependence prior to this operation.
...
+      auto &positions = m_positions.access();
+      auto &rotations = m_rotations.access();
+      m_scanCounts.access()[0]++;
+      scanIntervals.push_back((*other.m_scanIntervals)[timeIndex]);
+      const size_t indexStart = other.linearIndex({0, timeIndex});
+      size_t indexEnd = indexStart + nonDetectorSize();
+      positions.insert(positions.end(), other.m_positions->begin() + indexStart,
+                       other.m_positions->begin() + indexEnd);
+      rotations.insert(rotations.end(), other.m_rotations->begin() + indexStart,
+                       other.m_rotations->begin() + indexEnd);
```
SimonHeybrock: I am wondering if as a first stage it would simplify things if we supported *only* sync scans in `ComponentInfo`? Async scans could still be done directly via `DetectorInfo` of course.

Framework/Beamline/test/ComponentInfoTest.h
```diff
@@ -946,12 +1005,12 @@ class ComponentInfoTest : public CxxTest::TestSuite {
     b.setPosition(b.root(), pos2);
     std::pair<int64_t, int64_t> interval1(0, 1);
     std::pair<int64_t, int64_t> interval2(1, 2);
-    a.setScanInterval(a.root(), interval1);
-    b.setScanInterval(b.root(), interval2);
+    functor(a, a.root(), interval1);
+    functor(b, b.root(), interval2);
```
SimonHeybrock: `functor`? It is completely unclear what this does.

Framework/Beamline/src/ComponentInfo.cpp
OwenArnold: If we are happy not to support async scans I'll remove the code. I think that now we have decided to prevent scanning for any components not holding detectors, then the async scan is probably irrelevant. I'll delete this code and associated functionality.

Framework/Beamline/test/ComponentInfoTest.h
OwenArnold: Just a function that can be injected to apply different `setScanInterval` member function calls. Between many of the tests (sync/async) this is the only difference. However, in light of the async removal (comment above), I'll get rid of the template parameter.

Framework/Beamline/src/ComponentInfo.cpp
SimonHeybrock: 👍 We may want to support it later, but for the time being and given the complexity maybe introducing this one step at a time is better, until we know that it works and is useful in practice?

OwenArnold: I'll remove it as as single commit if I can, that way it might help to reintroduce it if we find that it is needed.

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -359,5 +507,180 @@ bool ComponentInfo::isStructuredBank(const size_t componentIndex) const {
...
+    return m_scanCounts;
+  }
+}
+
+size_t ComponentInfo::scanSize() const {
+  const auto detectorScanSize = m_detectorInfo ? m_detectorInfo->scanSize() : 0;
+  if (!m_positions)
+    return 0 + detectorScanSize;
+  return m_positions->size() + detectorScanSize;
+}
```
ianbush: I wonder if we also need an `isSyncScan()` method to check whether a component has all sync detector scans. I can't see the `DetectorInfo::isSyncScan()` methods on this branch, perhaps it would be good to merge master?

ianbush (commented):
This looks great, the interface and code closely follows that in `ComponentInfo` as I would expect.

I have made some minor comments. A bigger question for me is if we need to prevent scanning of the sample a this point - as we do have a use case for this at the ILL. Would it be a big technical change to allow that in this PR?

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -359,5 +507,180 @@ bool ComponentInfo::isStructuredBank(const size_t componentIndex) const {
...
+
+/**
+ * As part of the public API, the ComponentInfo treats any component to be the
+ * same whether detector or otherwise. However, internally we often need to know
+ * what the non-detector size is as composed non-detector members are managed as
+ * part of this type.
+ * @return size of the component info in terms of non-detectors.
+ */
+size_t ComponentInfo::nonDetectorSize() const {
```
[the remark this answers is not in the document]
OwenArnold: I thought the comments here would have clarified things, but since we only store data in the ComponentInfo, for non detector components, we often need to translate between a component index (includes detectors), and an index that we need to use to fetch something at an address.

For example, imagine you had a beamline with 5 detectors, 1 source, 1 sample, and 1 root node. That's 8 components in total. Calling `ComponentInfo::position(0)` get the first detector, `ComponentInfo::position(7)` gets the root node. However, we only internally store a vector sized 3 for the non-detector components. 

`nonDetectorSize()` is a private function to allow us to do translations for some of the problems above. 

I'm happy to change the name, but I don't think your suggestion makes it any clearer.

OwenArnold:
@ianbush I don't think it does make sense to merge `linearIndex` and `compOffsetIndex` together. Willing to discuss. If you had something specific in mind, let me know.

ianbush (approved):
Change are good to go. Some minor clarification of code may be done in later PRs as discussed with @OwenArnold .

PR #21161, opened 2017-11-09
Change from setting PYTHONPATH to using .pth files

rosswhitfield:
This will allow both python 2 and 3 packages to be installed and imported into native python correctly

At the moment if you try to import python 2 mantid into python 3 you get:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/Mantid/bin/mantid/__init__.py", line 25, in <module>
    from . import pyversion
  File "/opt/Mantid/bin/mantid/pyversion.py", line 18, in <module>
    raise ImportError(message % (TARGET_VERSION, _running_vers))
ImportError: Python version mismatch, cannot continue.
Mantid was built against version '2.7' but you are running version '3.6'. These versions
must match.
```
So we should not be using `PYTHONPATH`

**To test:**
Build release packages for python 2 and 3, install and test importing into python 2 and 3

**Does not need to be in the release notes.**

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/buildconfig/CMake/LinuxPackageScripts.cmake
+++ b/buildconfig/CMake/LinuxPackageScripts.cmake
@@ -39,9 +39,8 @@ file ( WRITE ${CMAKE_CURRENT_BINARY_DIR}/mantid.sh
   "MANTIDPATH=${CMAKE_INSTALL_PREFIX}/${BIN_DIR}\n"
   "PV_PLUGIN_PATH=${CMAKE_INSTALL_PREFIX}/${PVPLUGINS_DIR}/${PVPLUGINS_DIR}\n"
   "PATH=$PATH:$MANTIDPATH\n"
-  "PYTHONPATH=$MANTIDPATH:$PYTHONPATH\n"
 
-  "export MANTIDPATH PV_PLUGIN_PATH PATH PYTHONPATH\n"
+  "export MANTIDPATH PV_PLUGIN_PATH PATH\n"
 )
 
 # c-shell
@@ -50,17 +49,26 @@ file ( WRITE ${CMAKE_CURRENT_BINARY_DIR}/mantid.csh
   "setenv MANTIDPATH \"${CMAKE_INSTALL_PREFIX}/${BIN_DIR}\"\n"
   "setenv PV_PLUGIN_PATH \"${CMAKE_INSTALL_PREFIX}/${PVPLUGINS_DIR}/${PVPLUGINS_DIR}\"\n"
   "setenv PATH \"\${PATH}:\${MANTIDPATH}\"\n"
-
-  "if ($?PYTHONPATH) then\n"
-  "  setenv PYTHONPATH \"\${MANTIDPATH}:\${PYTHONPATH}\"\n"
-  "else\n"
-  "  setenv PYTHONPATH \"\${MANTIDPATH}\"\n"
-  "endif\n"
 )
 
 install ( PROGRAMS ${CMAKE_CURRENT_BINARY_DIR}/mantid.sh
   ${CMAKE_CURRENT_BINARY_DIR}/mantid.csh
+  ${CMAKE_CURRENT_BINARY_DIR}/mantid.pth
   DESTINATION ${ETC_DIR}
+)
+
+###########################################################################
+# Find python site-packages dir and create mantid.pth
+###########################################################################
+execute_process(
+  COMMAND "${PYTHON_EXECUTABLE}" -c "if True:
+    from distutils import sysconfig as sc
+    print(sc.get_python_lib(plat_specific=True))"
+  OUTPUT_VARIABLE PYTHON_SITE
+  OUTPUT_STRIP_TRAILING_WHITESPACE)
+
+file ( WRITE ${CMAKE_CURRENT_BINARY_DIR}/mantid.pth
+  "${CMAKE_INSTALL_PREFIX}/${BIN_DIR}\n"
 )
 
 ############################################################################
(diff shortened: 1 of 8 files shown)


buildconfig/CMake/LinuxPackageScripts.cmake
```diff
@@ -46,19 +45,28 @@ file ( WRITE ${CMAKE_CURRENT_BINARY_DIR}/mantid.csh
...
-  "endif\n"
 )
 
 install ( PROGRAMS ${CMAKE_CURRENT_BINARY_DIR}/mantid.sh
   ${CMAKE_CURRENT_BINARY_DIR}/mantid.csh
+  ${CMAKE_CURRENT_BINARY_DIR}/mantid.pth
   DESTINATION ${ETC_DIR}
 )
 
+###########################################################################
```
peterfpeterson: I'm surprised that the `install` command can be before the file is created.

peterfpeterson:
The ubuntu, ubuntu-py3, and rhel7 builds should all have packages.

peterfpeterson:
It would be good to add a templated file (`dont_have_a_good_name.py.in`) for people wanting to add their development version into python (e.g. `virtualenv` and `conda`):
```python
import os
from distutils import sysconfig as sc

# get location of mantid this was run with
mantidpath = "${CMAKE_CURRENT_BINARY_DIR}"
if not mantidpath.endswith(os.sep):
    mantidpath += os.sep

# where path file should go
pathfile = os.path.join(sc.get_python_lib(plat_specific=True), 'mantid.pth')

print('writing', mantidpath, 'to', pathfile)
with open(pathfile, 'w') as handle:
    handle.write(mantidpath)
    handle.write("\n")
```

peterfpeterson (changes requested):
Other than the script for developers...the installers have everything "right."

buildconfig/CMake/LinuxPackageScripts.cmake
```diff
@@ -46,19 +45,28 @@ file ( WRITE ${CMAKE_CURRENT_BINARY_DIR}/mantid.csh
...
+# Find python site-packages dir and create mantid.pth
+###########################################################################
+execute_process(
+  COMMAND "${PYTHON_EXECUTABLE}" -c "if True:
```
martyngigg: Just out of curiosity, why is the `if True:` needed?

martyngigg (dismissed):
I checked this on Ubuntu. I built a package and was then able to `import mantid` successfully.

I get a warning about the Vates algorithms: `Could not open library /opt/mantidunstable/plugins//libMantidVatesAlgorithms.so: libvtkIOLegacy-pv5.4.so.1: cannot open shared object file: No such file or directory` but that is not related to this.

martyngigg (changes requested):
After removing the package I'm still left with a file in `/usr/lib/python2.7/dist-packages/mantid.pth`. Trying to install a new package on top then results in an error.

martyngigg:
@rosswhitfield If we slightly update the `CMAKE_INSTALL_RPATH` like

```
diff --git a/buildconfig/CMake/LinuxPackageScripts.cmake b/buildconfig/CMake/LinuxPackageScripts.cmake
index c35609b..15502e6 100644
--- a/buildconfig/CMake/LinuxPackageScripts.cmake
+++ b/buildconfig/CMake/LinuxPackageScripts.cmake
@@ -23,8 +23,10 @@ if ( CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT )
   set ( CMAKE_INSTALL_PREFIX /opt/mantid${CPACK_PACKAGE_SUFFIX} CACHE PATH "Install path" FORCE )
 endif()
 
-# We are only generating Qt4 packages at the moment
-set ( CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_PREFIX}/${LIB_DIR};${CMAKE_INSTALL_PREFIX}/${PLUGINS_DIR}/qt4 )
+set ( CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_PREFIX}/${LIB_DIR};${CMAKE_INSTALL_PREFIX}/${PLUGINS_DIR};${CMAKE_INSTALL_PREFIX}/${PLUGINS_DIR}/qt4; )
+if ( MAKE_VATES )
+  list ( APPEND CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_PREFIX}/${LIB_DIR}/paraview-${ParaView_VERSION_MAJOR}.${ParaView_VERSION_MINOR} )
+endif ()
 
 # Tell rpm that this package does not own /opt /usr/share/{applications,pixmaps}
 # Required for Fedora >= 18 and RHEL >= 7
```
then we can just do `python -c "import mantid"` and everything imports cleanly without using `mantidpython`. Shall we apply that here or do it separately?

peterfpeterson:
I vote for putting it on this PR.
________________________________

buildconfig/CMake/LinuxPackageScripts.cmake
rosswhitfield: It's not. I wrote this 6 months ago and I can't remember why I had that, probably from where I copied it. 😆

martyngigg: :)

peterfpeterson:
[mantidunstable-python3.deb](http://builds.mantidproject.org/job/pull_requests-ubuntu-python3/5498/artifact/build/mantidunstable-python3_3.11.20171110.1349-0ubuntu1~xenial1_amd64.deb)

martyngigg (approved):
This looks good now. `mantid` imports cleanly into a vanilla Python instance and the `.pth` files are cleaned up on package removal.

peterfpeterson (approved):

PR #21166, opened 2017-11-10
Cleanup Run sharing in ExperimentInfo

SimonHeybrock:
`WorkspaceFactory` shares the `Run` in a workspace, whereas the copy constructor of `ExperimentInfo` makes a clone (the copy constructor would be called, e.g., when cloning a workspace, without use of the factory).

Changes:
- `ExperimentInfo` now shares the run when copies are made.
- Use `cow_ptr` instead of hand-written copy-on-write based on `shared_ptr`.
- Add `Experiment::setSharedRun` to avoid code duplication, which apparently was added because SNS had trouble with large runs being copied. @wdzhou  you should have a look if this still works for you as expected.
- Workspace factory now also copies `ExperimentInfo::m_moderatorModel` and `ExperimentInfo::m_choppers` from the parent. I believe these have been previously overlooked in the implementation?

**To test:**

Code review.
No issue, no release notes (potentially there might be performance improvements for certain cases, but I cannot name any).

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/inc/MantidAPI/ExperimentInfo.h
+++ b/Framework/API/inc/MantidAPI/ExperimentInfo.h
@@ -104,6 +104,8 @@ public:
   const Run &run() const;
   /// Writable version of the run object
   Run &mutableRun();
+  void setSharedRun(Kernel::cow_ptr<Run> run);
+
   /// Access a log for this experiment.
   Kernel::Property *getLog(const std::string &log) const;
   /// Access a single value from a log for this experiment.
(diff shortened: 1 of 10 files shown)


SimonHeybrock:
Need to solve one more problem, `PeaksWorkspace` is doing horribly things:
```cpp
API::LogManager_sptr PeaksWorkspace::logs() {
  if (m_logCash)
    return m_logCash;

  m_logCash = API::LogManager_sptr(&(this->mutableRun()), NullDeleter());
  return m_logCash;
}
```

martyngigg (approved):
The changes look sensible to me and should be functionally equivalent.

I have no idea why the COW mechanism was hand crafted for the `Run` & `Sample` before.

SimonHeybrock:
Note that the OSX build seems to have passed, but is not picket up by the PR for some reason: http://builds.mantidproject.org/job/pull_requests-osx/17453/consoleFull.

PR #21170, opened 2017-11-13
Refactor CropToComponent to use the new instrument API

SimonHeybrock:
Refactor `CropToComponent` to use the new instrument API. This also serves as a preparation for adding MPI support in the next step.

Note that the code for doing the mapping from detectors to spectra is likely to be absorbed into some lower-level abstraction like `IndexInfo` in the future.

I observed a speedup from 0.7 seconds to 0.4 seconds when extracting a single bank from a 4-bank instrument with 160k pixels per bank.

**To test:**

Code review. Existing unit tests look quite reasonable and should hopefully catch any issues.
No related issue.

Release notes updated with performance improvement.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/src/CropToComponent.cpp
+++ b/Framework/Algorithms/src/CropToComponent.cpp
@@ -1,12 +1,12 @@
 #include "MantidAPI/AlgorithmManager.h"
 #include "MantidAPI/MatrixWorkspace.h"
-#include "MantidAPI/SpectrumInfo.h"
 #include "MantidAlgorithms/CropToComponent.h"
 #include "MantidGeometry/Instrument.h"
 #include "MantidGeometry/Instrument/ComponentInfo.h"
-#include "MantidGeometry/Instrument/DetectorInfo.h"
 #include "MantidKernel/ArrayProperty.h"
-#include "MantidTypes/SpectrumDefinition.h"
+#include "MantidIndexing/Conversion.h"
+#include "MantidIndexing/GlobalSpectrumIndex.h"
+#include "MantidIndexing/IndexInfo.h"
 
 namespace {
 std::vector<size_t>
@@ -22,31 +22,6 @@ getDetectorIndices(const Mantid::API::MatrixWorkspace &workspace,
     detIndices.insert(detIndices.end(), indices.begin(), indices.end());
   }
   return detIndices;
-}
-
-std::vector<size_t>
-getWorkspaceIndices(const Mantid::API::MatrixWorkspace &workspace,
-                    const std::vector<size_t> &detectorIndices) {
-  std::vector<bool> detectorMap(workspace.detectorInfo().size(), false);
-  for (const auto &index : detectorIndices)
-    detectorMap[index] = true;
-
-  const auto &spectrumInfo = workspace.spectrumInfo();
-  std::vector<size_t> spectrumIndices;
-  for (size_t i = 0; i < spectrumInfo.size(); ++i) {
-    const auto &spectrumDefinition = spectrumInfo.spectrumDefinition(i);
-    if (spectrumDefinition.size() == 1 &&
-        detectorMap[spectrumDefinition[0].first])
-      spectrumIndices.push_back(i);
-    if (spectrumDefinition.size() > 1)
-      throw std::runtime_error("Cannot map from detector ID to workspace index "
-                               "-- spectrum in MatrixWorkspace maps to more "
-                               "than 1 detector.");
-  }
-  if (detectorIndices.size() != spectrumIndices.size())
-    throw std::runtime_error(
-        "Some of the requested detectors do not have a corresponding spectrum");
-  return spectrumIndices;
 }
 }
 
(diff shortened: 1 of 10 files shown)


Framework/Indexing/inc/MantidIndexing/Conversion.h
```diff
@@ -0,0 +1,79 @@
...
+#include "MantidIndexing/SpectrumNumber.h"
+
+namespace Mantid {
+namespace Indexing {
+
+/** Conversion helpers from and to (vectors of) integer types such as
+  SpectrumNumber and GlobalSpectrumIndex.
+
+  @author Simon Heybrock
+  @date 2016
```
AntonPiccardoSelg: 2017

Framework/Algorithms/src/CropToComponent.cpp
```diff
@@ -99,12 +81,13 @@ void CropToComponent::exec() {
       getProperty("InputWorkspace");
 
   // Get all detectors
-  std::vector<Mantid::Geometry::IDetector_const_sptr> detectors;
-  getDetectors(inputWorkspace, componentNames, detectors);
+  const auto &detectorIndices =
```
AntonPiccardoSelg: Just out of interest. The exposure of detector indices is what you eventually want to remove, right?

AntonPiccardoSelg (changes requested):
This looks good. The speed improvement is great.

Framework/Algorithms/src/CropToComponent.cpp
SimonHeybrock: Not really. Instrument-2.0 works preferably with detector indices, and can thus avoid map lookups. In general I think detector IDs should only be used for UI or in files.

Framework/Indexing/src/IndexInfo.cpp
```diff
@@ -193,6 +193,55 @@ SpectrumIndexSet IndexInfo::makeIndexSet(
...
+                             "globalSpectrumIndicesFromDetectorIndices -- no "
+                             "spectrum definitions available");
+  std::vector<char> detectorMap;
+  for (const auto &index : detectorIndices) {
+    // IndexInfo has no knowledge of the maximum detector index so we workaround
+    // this knowledge gap by assuming below that any index beyond the end of the
+    // map is 0.
+    if (index >= detectorMap.size())
+      detectorMap.resize(index + 1, 0);
+    detectorMap[index] = 1;
```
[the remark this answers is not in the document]
SimonHeybrock: I will be sparse, but has a known and limited size (number of pixels in the instrument). I believe a `map` or `unordered_map` would be (much) slower here, both in terms of building it as well as during lookup. I deliberately chose a vector for this reason.

Framework/Indexing/src/IndexInfo.cpp
```diff
@@ -193,6 +193,55 @@ SpectrumIndexSet IndexInfo::makeIndexSet(
...
+  }
+
+  std::vector<GlobalSpectrumIndex> spectrumIndices;
+  for (size_t i = 0; i < size(); ++i) {
+    const auto &spectrumDefinition = m_spectrumDefinitions->operator[](i);
+    if (spectrumDefinition.size() == 1) {
+      const auto detectorIndex = spectrumDefinition[0].first;
+      if (detectorMap.size() > detectorIndex &&
+          detectorMap[detectorIndex] != 0) {
```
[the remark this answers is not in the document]
SimonHeybrock: I check if the `detectorMap` is larger than `detectorIndex`, i.e., `detectorIndex` is smaller than the size, so I think this is ok, just written in different order?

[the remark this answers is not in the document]
AntonPiccardoSelg: It is a facepalm moment, I read it as the other way around. :-(

Framework/Indexing/src/IndexInfo.cpp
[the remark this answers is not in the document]
AntonPiccardoSelg: Ok, that is fine.

Framework/Algorithms/src/CropToComponent.cpp
AntonPiccardoSelg: I thought that detector indices were an implementation detail with no real meaning, except for internal book-keeping. It seems odd that this needs to be handled on an algorithm level.

AntonPiccardoSelg (approved):

Framework/Algorithms/src/CropToComponent.cpp
SimonHeybrock: Detector indices are to detector IDs a bit like workspace indices are to spectrum numbers. To be more precise, detector index is even more well defined than workspace index: A workspace index for a spectrum can change for many reasons, such as loading only some spectra, extracting spectra, or reordering spectrum. A detector index is always the same -- it is simply the index into a sorted vector of detector IDs. It only changes if the instrument changes by adding detectors.

You are right though that we do want to keep them "internal" and not expose them to users via algorithms since adding a fourth index would be too confusing. Any access to the Instrument-2.0 API via `DetectorInfo` is using detector indices though.

AntonPiccardoSelg: It is a good point that the workspace index itself is only an implementation detail

Issue #11321, opened 2014-11-03
Multi-dataset fitting interface: add a wiki page


mantid-roman:
This issue was originally [TRAC 10479](http://trac.mantidproject.org/mantid/ticket/10479)

This ticket is blocked by : 
- [X] [TRAC8911](http://trac.mantidproject.org/mantid/ticket/8911)

This ticket blocks :  #11927.

Add a description wiki page with examples.

PR #21174, opened 2017-11-13
Output multidataset fit parameters in columns

### LINKED ISSUE #21173: Multi-dataset fitting interface: output parameters in columns
After a simultaneous fit create a table workspace with parameter values organised similar to the output from `PlotPeakByLogValue` algorithm. The parameters are put into columns: two columns per parameter, one column for the value and the other for the error. Each row corresponds to a dataset. The table workspace must be ready for plotting: the user selects a column and can plot it.

mantid-roman:
After a simultaneous fit the MultiDatasetFit interface creates a table workspace with parameter values put into columns. It's easier to plot.

**To test:**

Do a simultaneous fit using the MultiDatasetFit interface. It must create a TableWorkspace with a name ending on `_vs_dataset`. Open the workspace and plot some columns.

Fixes #21173 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/qt/scientific_interfaces/MultiDatasetFit/MultiDatasetFit.cpp
+++ b/qt/scientific_interfaces/MultiDatasetFit/MultiDatasetFit.cpp
@@ -8,7 +8,10 @@
 #include "MantidAPI/FunctionFactory.h"
 #include "MantidAPI/ITableWorkspace.h"
 #include "MantidAPI/MatrixWorkspace.h"
+#include "MantidAPI/MultiDomainFunction.h"
 #include "MantidAPI/Run.h"
+#include "MantidAPI/TableRow.h"
+#include "MantidAPI/WorkspaceFactory.h"
 #include "MantidAPI/WorkspaceGroup.h"
 #include "MantidKernel/TimeSeriesProperty.h"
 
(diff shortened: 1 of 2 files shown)


qt/scientific_interfaces/MultiDatasetFit/MultiDatasetFit.cpp
```diff
@@ -25,6 +28,49 @@ namespace {
...
+  // Add columns for parameters and their errors
+  const auto &fun = *mdFunction.getFunction(0);
+  for (size_t iPar = 0; iPar < fun.nParams(); ++iPar) {
+    table->addColumn("double", fun.parameterName(iPar));
+    table->addColumn("double", fun.parameterName(iPar) + "_Err");
+  }
+
+  // Fill in the columns
+  for (size_t iData = 0; iData < nDomains; ++iData) {
+    Mantid::API::TableRow row = table->appendRow();
```
soininen: Since the number of rows is known beforehand, why not create enough rows before this loop instead of appending at each iteration? `table->setRowCount(nDomains)` should do the job. This might be beneficial performance-wise.

qt/scientific_interfaces/MultiDatasetFit/MultiDatasetFit.cpp
```diff
@@ -25,6 +28,49 @@ namespace {
...
+    // A warning will be shown elsewhere.
+    return;
+  }
+
+  auto table =
+      Mantid::API::WorkspaceFactory::Instance().createTable("TableWorkspace");
+  auto col = table->addColumn("double", "Dataset");
+  col->setPlotType(1); // X-values inplots
+
+  auto nDomains = mdFunction.getNumberDomains();
```
soininen: Could `nDomains` be declared `const`?

qt/scientific_interfaces/MultiDatasetFit/MultiDatasetFit.cpp
```diff
@@ -25,6 +28,49 @@ namespace {
 const int zoomToolPage = 0;
 const int rangeToolPage = 1;
 Mantid::Kernel::Logger g_log("MultiDatasetFit");
+
+/// Copy parameter values into a table workspace that is ready for
+/// plotting them against a dataset index.
+void formatParametersForPlotting(const Mantid::API::IFunction &function,
+                                 const std::string &parametersPropertyName) {
+  const auto &mdFunction =
+      dynamic_cast<const Mantid::API::MultiDomainFunction &>(function);
```
soininen: Would it be worth checking if the cast is successful?

qt/scientific_interfaces/MultiDatasetFit/MultiDatasetFit.cpp
```diff
@@ -25,6 +28,49 @@ namespace {
...
+  col->setPlotType(1); // X-values inplots
+
+  auto nDomains = mdFunction.getNumberDomains();
+  assert(nDomains == mdFunction.nFunctions());
+
+  // Add columns for parameters and their errors
+  const auto &fun = *mdFunction.getFunction(0);
+  for (size_t iPar = 0; iPar < fun.nParams(); ++iPar) {
+    table->addColumn("double", fun.parameterName(iPar));
+    table->addColumn("double", fun.parameterName(iPar) + "_Err");
```
soininen: [`FindEPP`](http://docs.mantidproject.org/nightly/algorithms/FindEPP-v2.html) produces very similar output table, and there the error column names are appended by `Error` instead of `_Err`. Would it make sense to be consistent with `FindEPP` here?

soininen (changes requested):
I played around with the interface and the new parameters table looks very handy. The code looks also good despite the few minor comments. Is there any user-facing documentation for this interface available?

qt/scientific_interfaces/MultiDatasetFit/MultiDatasetFit.cpp
mantid-roman: Yes. I need to do it. If you run a fit with a single data set it gives you a crash :(

qt/scientific_interfaces/MultiDatasetFit/MultiDatasetFit.cpp
mantid-roman: Yes

qt/scientific_interfaces/MultiDatasetFit/MultiDatasetFit.cpp
mantid-roman: `_Err` is consistent with the output from `PlotPeakByLogValue` algorithm which is used by the same interface to run sequential fits.

qt/scientific_interfaces/MultiDatasetFit/MultiDatasetFit.cpp
mantid-roman: I don't think it's a big issue but I can do it.

mantid-roman:
The issue to write documentation is #11321

soininen (approved):

PR #21207, opened 2017-11-17
Cancel button frequency domain analysis

### LINKED ISSUE #20879: Add a cancel button to FDA
The Frequency domain analysis GUI should allow the user to cancel the calculation. When the window closes the calculations  should end graceful. This amounts to terminating the threads if they are active.

AnthonyLim23:
This adds a cancel button to the MaxEnt widget of the Frequency domain analysis GUI (in muon section). 

**To test:**
1. Open muon analysis and load some data (the data does not matter). Close muon analysis
2. Open Frequency Domain Analysis and select the MaxEnt option from the combo box at the top of the window
3. Run the algorithm, wait a little while then press cancel.
4. The algorithm should stop (the progress bar in the algorithm dialog will disappear) and the calculate button will be enabled
5. Run maxEnt again, this time instead of pressing cancel close the frequency domain analysis window and see if it stops the algorithm (it should)

Fixes #20879. 

**Release Notes** 

Added to release notes 
---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


scripts/test/Muon/MaxEntPresenter_test.py
```diff
@@ -66,11 +67,11 @@ def test_buttonWithoutRaw(self):
```
Mantid-Matthew: With the tests involving handleMaxEntButton do you not want to test that the threads are called correctly as well? I think you could replace handleMaxEntButton with getMaxEntInput in the current tests and still pass? I agree you don't need a test for the cancel button though as whilst it has an if statement it is a trivial one.

AnthonyLim23: no, it will not call thread.start if we just call the get input. The thread.start wraps the maxEnt algorithm so start calls it

Mantid-Matthew: Yep I was being blind and had missed the start thread check.

Mantid-Matthew:
Currently works for me. When I click cancel however or close the window a stack trace is produced in the logger window. This doesn't stop it working but it would be nice if the trace was not produced when the cancel button is working as expected.

_MaxEnt: Execution terminated by user.
Exception RuntimeError: 'wrapped C/C++ object of type ThreadModel has been deleted' in <bound method ThreadModel.__del__ of <Muon.thread_model.ThreadModel object at 0x000002AA5DA105E8>> ignored
Traceback (most recent call last):
  File "C:/Users/tgz88916/Documents/GitHub/mantid/scripts\Muon\thread_model.py", line 20, in run
    self.model.execute()
  File "C:/Users/tgz88916/Documents/GitHub/mantid/scripts\Muon\maxent_model.py", line 21, in execute
    self.alg.execute()
KeyboardInterrupt_

AnthonyLim23:
I will look into how to suppress the stack trace. May be a little while.

AnthonyLim23:
I have had a look at removing the traceback. Removing it without changing the sys settings seems to be non-trivial.

scripts/Muon/thread_model.py
```diff
@@ -17,8 +20,13 @@ def __del__(self):
         self.wait()
 
     def run(self):
-        self.model.execute()
-        self.model.output()
+        try:
+            self.model.execute()
+            self.model.output()
+        except:
```
Mantid-Matthew: I think it is considered good practice to specify the type of exception you want to catch here. In this case it appears to be KeyboardInterrupt that you want to catch.

Matthew-Andrew (approved):
Once the tests pass this is good to go.

Issue #21171, opened 2017-11-13
Engineering Diffraction GUI - Move ADS and Algorithm code to Model


josephframsay:
Following on from #21082 (where the fitting tab was made to load and plot multiple runs), the fitting tab of the ED GUI should be able to apply its fitting routines to these runs.

This ticket should finish off moving fitting/Algorithm code from `EnggDiffFittingPresenter` to `EnggDiffFittingModel` in order to make the system more decoupled.

NOTE: This issue originally covered applying fitting to multiple runs, but the refactoring job ended up being large enough that it was split out into its own PR.

josephframsay:
Completed by #21226

Issue #21082, opened 2017-10-30
Engineering Diffraction GUI - Fitting tab should support loading multiple runs


josephframsay:
The fitting tab in the ED GUI currently only allows the user to load one run at a time. The user should be able to load a range of runs and select which to fit from the list box at the side of the plot

PR #21226, opened 2017-11-21
Engineering Diffraction GUI move Algorithm and ADS code to model

josephframsay:
As part of #21171, all code related to Mantid algorithms and the ADS has been moved from `EnggDiffFittingPresenter` to the new `EnggDiffFittingModel`, giving a more decoupled system.

This PR was originally also meant to take care of enabling fitting of multiple runs at the same time, but the refactoring job was sufficiently large that I thought it best to make an intermediate PR here (apologies for the huge diff).

Please note: the unit test at `EnggDiffFittingPresenterTest.h:428` is disabled, as the functionality is not yet properly implemented. This will be re-enabled and fixed when **Fit all** is finally complete.

**To test:**

Some sample data is [here](https://github.com/mantidproject/mantid/files/1460806/ENGINX_focused_runs.zip).

Load a run from the above data by browsing with the first **Browse** button, then clicking **Load**.
Select the loaded run from the list widget by clicking its label - it should be plotted in the plot window.
Click **Fit** to do a fit - some workspaces with fit parameters should appear in the Workspaces area, and the fitted peaks should be plotted over the original plot.

Try changing the peaks with the Peaks line-edit, fitting while multiple runs are loaded, and anything else you think might break things.

Check that fitting information was saved (as `.csv`). On Windows this will be in `C:\EnginX_Mantid\User`, on Linux `~/EnginX_Mantid/User`.

Partially addresses #21171 #21082 

**Release Notes** 
The change is not complete - I will add it to the release notes once multi-run fitting has been implemented.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


gvardany:
I get this error when clicking the `Load` button. Am I doing something wrong here?
screen shot 2017-11-23 at 13 34 32

josephframsay:
Ah, nope this is a bug - it looks like the GUI can only work out the bank number if it's loading multiple files

gvardany:
The loading seems fine now. Then when trying to fit with the default parameters, I get error (see the log in red). Is this expected? If not, what can I supply in `Peaks` field to get fit run?

screen shot 2017-11-24 at 08 58 47

gvardany:
Also I could not find the .csv file (I'm on OS X). I'm a little uncomfortable with a hardcoded path for the output file, shouldn't this be save in Mantid save directory set by the user?

josephframsay:
@gvardany I agree about the paths, but this might be out of the scope of this PR - the output files should be in the location given by `Poco::Path().home()`. I'm not quite sure where 'File not found' is coming from - would it be possible to get a screenshot of the log on a slightly lower setting such as Information?

josephframsay:
(For reference in future, since it's hard to find, the output directory is set in `EnggDiffractionPresenter::outFilesRootDir`)

gvardany:
Here is the tip of the information log. It does produce the fit table, although it claims the fit has failed. Regarding the file path, what is the `Output Folder` under the `Settings` tab? Is this related to the fit output or not? 
screen shot 2017-11-28 at 09 51 20

josephframsay:
The Output folder only appears to be used for output from the Focus tab, not Fitting.
Regarding the fit errors, the default values for peak centres mean the algorithm is trying to fit peaks where there are none, so I suppose it makes sense that there would be errors here. Putting `0.95, 1.23, 1.25, 1.48, 2.25, 2.56` in the Peaks text box gives you some nice peaks with no errors for `277208_1`.

gvardany:
I think that indeed the output file path should be the same for Focus and Fitting, and be the one set up in Settings. This would make the most sense to me. It can be done in a separate ticket though.

Regarding the fitting, the list of numbers you gave does run a proper fit, but when leaving the `Peaks` to default (empty), it fails. My only concern is whether the default values are not so smart, or the test data we run over are somehow uncommon? Where do the default parameters come from?

josephframsay:
The default values appear to come from the defaults of the `EnggFitPeaks` algorithm. However I'm not so sure how valuable these defaults are in practice anyway, as presumably the peaks will be in a different place on every experiment, right?

A bigger priority for me would be to get the peak picker working properly so that the user can select the peaks they want to fit easily, as I would guess that they would want to do this more often than use the defaults anyway

gvardany (approved):
Ok, the explanation sounds good to me. I did not realise that the peak picker is not yet operational, once in place I would also assume no one would rely on the defaults.

ianbush:
@josephframsay I assume you do not want tickets #21171 #21082 closed, but if either of them are completed by this please close them manually.

PR #21232, opened 2017-11-21
Suppress clang-tidy for generated code

peterfpeterson:
It turns out that if `CMAKE_CXX_CLANG_TIDY` is not set when the target is created, the tool won't be run on the compiled files.

**To test:**

A code review should be sufficient. However, setting up a build to use `clang-tidy` via `-DENABLE_CLANG_TIDY=on` would be better.

*Does not need to be in the release notes* because it is only for users.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


buildconfig/CMake/QtTargetFunctions.cmake
```diff
@@ -75,6 +75,8 @@ endfunction()
 # keyword: INSTALL_DIR_BASE Base directory the build output. The final product goes into a subdirectory based on the Qt version.
 # keyword: OSX_INSTALL_RPATH Install path for osx version > 10.8
 function (mtd_add_qt_target)
+  set(CMAKE_CXX_CLANG_TIDY "") # disable clang-tidy for sip generated code
```
martyngigg: This comment is not quite accurate. This is nothing to to with sip generation. It will in factor disable clang-tidy for all of the Qt code. Is that what we want?

peterfpeterson:
It turns out that `CMAKE_CXX_CLANG_TIDY` is the default value for all targets. You can override (or turn off) the value by setting the `CXX_CLANG_TIDY` property on an individual target.

martyngigg (approved):
This works for me and clang-tidy no longer segfaults.

At some point we may want to look into enabling the Qt library targets.

PR #21233, opened 2017-11-21
Update the ISIS_Powder system test reference files (Not including Pearl)

### LINKED ISSUE #21229: ISIS_Powder system tests not comparing data correctly
### Expected behavior
ISIS_Powder system test should actually compare data from output to the reference files
### Actual behavior
The comparison statement: `if not (mantid.CompareWorkspaces(Workspace1=ws, Workspace2=ref)):` in the system test comparison makes the assumption that `CompareWorkspaces` returns ONLY a boolean. This is not the case and it actually returns the result boolean and the `compare_msg` table workspace as a tuple which when printed looks something like:
```
CompareWorkspaces_returns(Result=True, Messages=TableWorkspace
Columns: 3
Rows: 0
0 kB)
``` 

Initial work in changing this using something like `if not (mantid.CompareWorkspaces(Workspace1=ws, Workspace2=ref)[0]):` to check the results item in the tuple shows that the POLARIS data does not match the reference. A data comparison has not been done for the other ISIS_Powder instruments, the the comparison must be changed for those. 

### Platforms affected
* Observed on Windows 10

ElliotAOram:
Ensured that the system tests for ISIS_Powder actually test for comparison.
Additionally I have remove the `_compare_ws` function from each of the system tests in favour of using the standard Mantid validate (which calls `CompareWorkspace` anyway)

**To test:**

* Ensure the system tests for ISIS_Powder run on the build server
* Ensure that the reference files have been updated
* Ensure that the ISIS_Powder system tests run locally

Fixes #21229

**Release Notes** 
*Does not need to be in the release notes - internal change.*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


ElliotAOram:
Once the problem has been confirmed on the build server the reference files can be updated and the more standard system test comparison method can be implemented.

josephframsay:
Unfortunately the PEARL create vanadium tests all fail on my machine. Am I missing something to do with updating the reference files?

ElliotAOram:
@josephframsay Did you re-run CMake before performing the system tests? I assume so actually - given that the other pass. I'll take a look at it with you later today

ElliotAOram:
Looking at this shows differences of the order `1e-5`. As currently none of the system tests are working for all ISIS_Powder instruments. I think it might be best to push the Pearl tolerance up and then solve this in a follow up issue.

martyngigg (approved):
These tests all now pass on Ubuntu.

Merging now as they are broken on `master`.

PR #21234, opened 2017-11-22
Use fixed Windows SDK Version on builders

martyngigg:
Description of work.

Ensures that all Windows builders target the same version of the platform SDK. This will allow us to build on Windows 10 but still guarantee and app that runs on Windows 7.

**To test:**

* Code review as the changes are to the build server script alone. The default value cannot be
specified with CMake itself.
* You can simulate this locally by running (from the command line as it is not available in the GUI)

`cmake -DCMAKE_SYSTEM_VERSION=8.1`

from the build directory. When the generation has complete:

* open VS
* select a target, say Kernel
* right-click on it and select properties
* check that the "Target Platform Version" matches 8.1

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


ElliotAOram (approved):
Checked various targets after generation in VS all are 8.1.
Code changes to build script all look good. Built fine too.

PR #21267, opened 2017-11-27
ComponentInfo needs Component Names

### LINKED ISSUE #21256: ComponentInfo needs component names
Need an equivalent of `ICompoentAssembly::getComponentByName` or order to move away from old `Instrument` code. Also will be needed by @LamarMoore for the `InstrumentView` refactor or order to display component names in that GUI.

OwenArnold:
ComponentInfo needs ability to store and search via component names. fixes #21256 

**Tester**

Code review should be sufficient. No instances of `IComponent::getComponentByName` have been replaced here, so should not affect existing behaviour.

### DIFF
--- a/Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
+++ b/Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
@@ -9,6 +9,7 @@
 #include <utility>
 #include <cstddef>
 #include <Eigen/Geometry>
+#include <string>
 
 namespace Mantid {
 namespace Beamline {
@@ -54,6 +55,7 @@ private:
   Mantid::Kernel::cow_ptr<std::vector<Eigen::Quaterniond>> m_rotations;
   Mantid::Kernel::cow_ptr<std::vector<Eigen::Vector3d>> m_scaleFactors;
   Mantid::Kernel::cow_ptr<std::vector<bool>> m_isStructuredBank;
+  boost::shared_ptr<const std::vector<std::string>> m_names;
 
   const size_t m_size = 0;
   const int64_t m_sourceIndex = -1;
@@ -96,6 +98,7 @@ public:
                 boost::shared_ptr<std::vector<Eigen::Quaterniond>> rotations,
                 boost::shared_ptr<std::vector<Eigen::Vector3d>> scaleFactors,
                 boost::shared_ptr<std::vector<bool>> isStructuredBank,
+                boost::shared_ptr<const std::vector<std::string>> names,
                 int64_t sourceIndex, int64_t sampleIndex);
   /// Copy assignment not permitted because of the way DetectorInfo stored
   ComponentInfo &operator=(const ComponentInfo &other) = delete;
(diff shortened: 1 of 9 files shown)


KarlPalmen (approved):
Looks good

LamarMoore (commented):
Looks good. I will merge this.

Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
```diff
@@ -138,6 +141,8 @@ class MANTID_BEAMLINE_DLL ComponentInfo {
   size_t sample() const;
   size_t root() const;
   double l1() const;
+  std::string name(const size_t componentIndex) const;
```
SimonHeybrock: Why return by value?

Framework/Geometry/inc/MantidGeometry/Instrument/ComponentInfo.h
```diff
@@ -111,6 +112,7 @@ class MANTID_GEOMETRY_DLL ComponentInfo {
   size_t sample() const;
   double l1() const;
   Kernel::V3D scaleFactor(const size_t componentIndex) const;
+  std::string name(const size_t componentIndex) const;
```
SimonHeybrock: Return by reference instead?

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -494,6 +503,21 @@ Eigen::Vector3d ComponentInfo::scaleFactor(const size_t componentIndex) const {
   return (*m_scaleFactors)[componentIndex];
 }
 
+std::string ComponentInfo::name(const size_t componentIndex) const {
+  return (*m_names)[componentIndex];
+}
+
+size_t ComponentInfo::indexOf(const std::string &name) const {
```
SimonHeybrock: I feel that this method is not providing a good interface, since it is misleading. It implies a unique mapping from `name` to index exists, but it does not.

A method with this signature should maybe throw if there are multiple components with the same name. Maybe we need a method returning all indices for a given name?

Furthermore, the reverse search makes the behavior different from the old `ComAssembly::getComponentByName`, which return the *first* match, instead of the last.

SimonHeybrock (commented):
@OwenArnold Some comments on this:

OwenArnold:
@SimonHeybrock I agree on these points. I'll make a further PR to update the API.

Framework/Beamline/src/ComponentInfo.cpp
OwenArnold: @SimonHeybrock I believe that naming was supposed to be unique but never enforced as such. As the reverse search will be generally faster, I would prefer to keep it and make fixes as and where problems are found. Loosing the ability to terminate the search (after a match has been found) could affect performance quite badly.

How about we instead create a map with component names as part of the `ComponentInfo` creation and throw if any duplicates exist. I did this to find and eliminate all Detector ID duplicates. The alternative would be to store a multimap, but I'm less keen on that.

SimonHeybrock: I would avoid introducing complications like maps if possible. For starters, renaming the method to something like `indexOfAny` might be enough?

OwenArnold: @SimonHeybrock I think that would probably be enough. If we did find that the wrong (first, last whatever) index was being returned, I think my first question would be why are names being duplicated in this particular IDF.

Issue #21238, opened 2017-11-22
Engineering Diffraction GUI bank number combo box is now obsolete


josephframsay:
Following on from #21082 (where the fitting tab was made to load and plot multiple runs) and #21171 (a major refactoring job on the presenter and model), the combo box for selecting bank is now obsolete. This widget, and all associated code, needs to be removed from the GUI.

josephframsay:
Closed by #21273

PR #21164, opened 2017-11-10
Engineering Diffraction GUI can load and plot multiple runs

josephframsay:
Loading and plotting multiple focused runs in the fitting tab of the ED GUI has been implemented. In order to achieve this, work was started on implementing an explicit model (as opposed to essentially using the ADS) for the fitting tab.

Note: the fitting tab is a work in progress. Attempting to do a fit will probably crash Mantid, and some of the code is a little rough around the edges as I haven't yet moved all the model-ly stuff into the model

**To test:**
[Re-enable the GUI](https://github.com/mantidproject/mantid/commit/8011de68c96312cf2ccf746ac34c8d433a07534e) and build.

Some sample data is [here](https://github.com/mantidproject/mantid/files/1460806/ENGINX_focused_runs.zip).

Open the ED GUI (Interfaces > Diffraction > Engineering Diffraction) and go to the fitting tab. Click the first Browse button and select all (or some combination of) the attached files, then hit Load. The loaded workspaces should appear in the list next to the plot, labelled with run number and bank number. Clicking on these should plot the corresponding workspace. Try and break things, and also try renaming the files for run 277208 and seeing if the load still works.

Partially fixes #21082 

**Release Notes** 

The change isn't complete, so shouldn't go in the release notes. I'll add it once fitting is wired up

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)?
- [ ] Are the unit tests small and test the class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] Do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant (user and developer) documentation been added/updated?

Does everything look good? Mark the review as **Approve**. A member of `@mantidproject/gatekeepers` will take care of it.

PR #21273, opened 2017-11-27
Engineering Diffraction GUI bank combo box removed and assorted refactoring

josephframsay:
After the work in #21164, the combo box for bank number in the ED GUI was made obsolete, as the list widget takes care of both run and bank numbers. This PR took care of removing the combo box and all related code.

A large amount of refactoring was also done, which involved moving some helper functions to anonymous namespaces (instead of having them as methods) and renaming some methods and attributes whose purpose did not match their original name, 

Addresses #21238 

**To test:**

Essentially, make sure everything still works as it did after #21226.

Some sample data is [here](https://github.com/mantidproject/mantid/files/1460806/ENGINX_focused_runs.zip).

Load a run from the above data by browsing with the first **Browse** button, then clicking **Load**.
Select the loaded run from the list widget by clicking its label - it should be plotted in the plot window.
Click **Fit** to do a fit - some workspaces with fit parameters should appear in the Workspaces area, and the fitted peaks should be plotted over the original plot. **Update:** The x axis label should be updated when workspaces are plotted so that the workspace units and plot label actually match.

Try changing the peaks with the Peaks line-edit, fitting while multiple runs are loaded, and anything else you think might break things.

Check that fitting information was saved (as `.csv`). On Windows this will be in `C:\EnginX_Mantid\User`, on Linux `~/EnginX_Mantid/User`.

**Release Notes**
The changes to the fitting tab are not complete - this change is a small part of the effort to enable multi-run fitting, which will be included in the release notes once it's been put to bed.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


ElliotAOram:
Might not be 100% related to this PR, but when you load in data initially, the x axis on the plot window is wrong. I assume this should be TOF not d-spacing?

image

ElliotAOram:
Is it worth doing a sanity check on the data range? e.g. if the peak list contains a value outside the X range of what it is trying to fit (e.i. data ranges from 10,000->60,000 in TOF, peak list should not contain anything outside of this range)

ElliotAOram:
Finally when you come to do peak selection, I might be worth making the distinction between the units for the peak location (e.g. store them all in d-spacing or TOF - not a mix)

ElliotAOram (approved):
Refactoring looks good and changes in PR function as described. The changes I have mentioned are beyond the scope of this PR, so I'm happy to approve this.

martyngigg (changes requested):
Can we fix the incorrect units on the X axis easily? Units are very important and it doesn't look great if some this simple is not correct.

martyngigg (approved):

PR #21258, opened 2017-11-23
Polaris set sample material

ElliotAOram:
**This should be merged after #21233 - This PR includes the changes from that PR as they were required to avoid later merge conflicts.**

This PR ensures that the Sample Material (chemical formula) is ALWAYS present in the output from the Polaris scripts. Currently, regardless of if you set the chemical formula in the script or not, the chemical formula is only added to the output workspace if you use absorption corrections.

**To test:**

Point Mantid data search directory to `\\olympic\babylon5\Public\ElliotOram\polaris` and run the following script:
```
import os
from isis_powder import Polaris, SampleDetails

#====Setup
DIRS = config['datasearch.directories'].split(';')
working_folder_name = "polaris"

# Relative to input folder
calibration_folder_name = "calibration"
calibration_map_rel_path = "polaris_mapping.yaml"
config_rel_path = "Master_copy_of_grouping_file_with_essential_masks.cal"

# Generate paths
input_dir = os.path.join(DIRS[0], 'input')
output_dir = os.path.join(DIRS[0], 'output')

calibration_map_path = os.path.join(input_dir, calibration_map_rel_path)
calibration_dir = os.path.join(input_dir, calibration_folder_name)
config_path =os.path.join(calibration_dir, config_rel_path)

#set up polaris instrument
polaris = Polaris(user_name="Mantid-Test",
                      calibration_mapping_file=calibration_map_path,
                      calibration_directory=calibration_dir,
                      output_directory=output_dir,
                      mode="PDF")


#====Load Vanadium

vanadium_run = '97942'  # Vanadium
polaris.create_vanadium(first_cycle_run_no=vanadium_run,
                                 do_absorb_corrections=True,
                                 multiple_scattering=False)


#====Load sample and focus

run_number = '97947' #Sample
sample_empty = '97948' # Container

sample_material = SampleDetails(height=4.0, radius=0.2985, center=[0.,0.,0.], shape='cylinder')
sample_material.set_material(chemical_formula='Si', temperature=10, pressure=10)

polaris.set_sample_details(sample=sample_material)

polaris.focus(run_number=run_number,
                  input_mode="Individual",
                  do_van_normalisation=True,
                  do_absorb_corrections=True,
                  multiple_scattering=False,
                  sample_empty=sample_empty,
                  sample_empty_scale=1.0)
```
* Running this script should populate the sample material of the output workspaces
[...]

PR #21276, opened 2017-11-27
Revert "Polaris set sample material"

josephframsay:
Reverts mantidproject/mantid#21258

This PR was prematurely merged - it cannot be merged before #21233. This was probably my bad for approving it before it was ready


josephframsay:
The parent branch was merged cleanly, so this revert is no longer necessary

Issue #20086, opened 2017-07-25
Multiple plot normalisation issue


gvardany:
When 2 spectra are plotted in separate windows, MantidPlot confuses the toggling of their normalisations (bin width/none).

### Expected behaviour
Normalise each spectrum individually in it's window.

### Actual behavior
They are mixed

### Steps to reproduce the behavior
```
ws1 = CreateSampleWorkspace()
ws2 = CloneWorkspace(ws1)
```
Plot the spectrum 1 of each of the 2 workspaces in separate windows.
On one of the windows toggle the normalisation option (to none).
Then click on the 2nd window, right click and you will see that the normalisation is shown as none, but in reality it is still bin-width normalised.
If you now toggle the normalisation on window 2, it will act on the spectrum in window 1.

### Platforms affected
Found on OS X

mantid-roman:
I don't see this on Windows. Could it be mac only issue?

gvardany:
Yes I just realised that. Seems it's mac os that confuses the focus.

PR #21277, opened 2017-11-27
Fix wrong graph axis being scaled on MacOS.

quantumsteve:
Description of work.

On MacOS [right-clicks don't change the focus of the window](https://superuser.com/questions/292895/mac-os-focus-on-right-click), letting one right-click on one plot and modify the axes of another. This PR follows suggestions from @martyngigg and @mantid-roman to override this behavior.

**To test:**

This is best tested on MacOS.
On Master:
1. Open or create a workspace where one can plot more than one spectra 
     i. I used `ExternalData/Testing/Data/UnitTest/IRS26173.RAW`
2. Right-click the workspace and use the "Plot Spectrum..." menu option to plot two spectra in separate windows.
3. Right click on the y-axis and click the option to scale axis.
4. Repeat step 3 with the other window and repeat until the axes change in the other (unclicked) window.

On this branch:
1. Repeat steps 1-3 above.
2. Repeat step 4 above, verifying that the modified axis matches the one clicked on.

This issue was reported by @mganeva. There is no GitHub issue associated with the pull request.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidPlot/src/ApplicationWindow.h
+++ b/MantidPlot/src/ApplicationWindow.h
@@ -1611,6 +1611,7 @@ private:
 
   friend class MantidUI;
   friend class ConfigDialog;
+  friend class Graph;
   QString m_nexusInputWSName;
 
   // Store initialized script environments
--- a/MantidPlot/src/Graph.cpp
+++ b/MantidPlot/src/Graph.cpp
@@ -3963,6 +3963,7 @@ void Graph::showAxisTitleMenu() {
 }
 
 void Graph::showAxisContextMenu(int axis) {
+  this->multiLayer()->applicationWindow()->setActiveWindow(this->multiLayer());
   QMenu menu(this);
 
   menu.addAction(getQPixmap("unzoom_xpm"), tr("&Rescale to show all"), this,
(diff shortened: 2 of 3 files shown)


gvardany:
I was wondering if https://github.com/mantidproject/mantid/issues/20086 is related or can be easily fixed here?

gvardany:
I wonder if this can be fixed in a more general way by setting the focus on right click anywhere on the window (not just the axis), it seems all the context menus are getting wrong.

gvardany:
Can one test this branch  `20086_focus_right_click` ?

quantumsteve:
@gvardany I agree that this is a more general issue, though I'm increasingly less comfortable with this workaround. Would @martyngigg's second suggestion of emitting a pointer to the layer better match the expectations of a MacOS user?

quantumsteve:
I tried opening two finder windows and right-clicks works as expected, but the focus doesn't change.

gvardany:
True, that the focus does not change, but the right click is attributed with the right finder window. This is similar to the multiple plot windows in the branch I've posted. I agree though that this is a workaround, if you have a better solution please go ahead.

quantumsteve:
Closing this since we haven't settled on a solution. Ideally we'd avoid such platform-specific interaction patterns in the new workbench.

PR #21288, opened 2017-11-29
New algorithm LoadAndMerge

### LINKED ISSUE #21156: New python algorithm LoadMergeRuns
A lightweight python wrapper algorithm is needed to take MultipleFileProperty as input, loop over the files and load them. This is similar to the generic Load, except this will call MergeRuns instead of Plus, when summing is requested.
Due to this metadata merging will be handled properly.
This will be needed basically for all the ILL workflow algorithms, which currently implement the same logic within them in a custom way.

gvardany:
This adds a new algorithm to load and merge multiple runs.
Practically it does the same as `Load`, except uses `MergeRuns` for summing.
This allows to handle the merging of the metadata.
This is particularly needed for all the data reduction workflows at the ILL.
The ad-hoc implementation of this in direct and indirect ILL reductions are refactored to use the new algorithm. Future ILL reductions (Powder Diffraction, Reflectometry, SANS) will also rely on this.

**To test:**
Use the new algorithm to load some list of nexus files from your favourite instrument using the [multiple file loading scheme ](http://docs.mantidproject.org/nightly/api/python/mantid/api/MultipleFileProperty.html).

Fixes #21156 . 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


docs/source/algorithms/LoadMergeRuns-v1.rst
```diff
@@ -0,0 +1,62 @@
...
+
+This algorithm loads multiple runs conforming the rules in :py:obj:`MultipleFileProperty <mantid.api.MultipleFileProperty>`.
+When listing is requested, the output is a :ref:`WorkspaceGroup <WorkspaceGroup>`, containing items for each run.
+When summing is requested, it is performed pair-wise *in-situ* with :ref:`MergeRuns <algm-MergeRuns>`.
+By doing this, the metadata (SampleLogs) are also merged governed by the rules specified in the Instrument Parameter File (IPF)
+as explained in :ref:`MergeRuns <algm-MergeRuns>`.
+This rules can be overridden by the **MergeRunsOptions** input.
+Specific loader can be given by the **LoaderName** and **LoaderVersion**.
+If left as defaults, the algorithm will automatically find out the corresponding specific loader, deducting from the first run in the list.
+It will set the name and the version of the suitable loader found as output properties in **LoaderName** and **LoaderVersion**.
```
KarlPalmen: is "deducting" the correct verb?

KarlPalmen: Oops I intended to comment the previous line.

gvardany: Nope, I meant deduce

KarlPalmen (changes requested):
Works well. Just one concern. When I did sum of two MUSR files each with two workspaces. I found the workspaces of the first file were shown along with the sum (of all 4 workspaces), but the two workspaces of the second file were not shown. Is this intended behaviour?

KarlPalmen:
I tried it on the three ARGUS files available in the unit test data. They load OK separately, but not when added.

gvardany:
Regarding ARGUS, I can see that run 26577 loads to a group workspace containing 3 workspaces (is this what is referred to as multi period?)
Run 31800 produces a group of 2, whereas run 26287 just a single workspace.
Therefore, MergeRuns fails when trying to merge these together. 
If you try to use Load over these files summed, it will also fail for the same reason.
So I think this is the correct behaviour.

Regarding, MUSR, can you send some files to try it out? I can see only one file in the unit tests.
But if each single file produces a group itself, something might be going wrong. 
Although I am not sure what to expect from MergeRuns in this case.

KarlPalmen:
When running load on the ARGUS workspaces they fail with error messages that explain why. I would like the same error messages to occur with LoadAndMerge.

KarlPalmen:
Dear Gagik

I attach two MUSR files. They both have a group of 2.

Karl

gvardany:
I agree that the error message produced by `Load` is more explicit. However, I think it is fine to receive different errors for `Load` and `LoadAndMerge`, since in once case `Plus` is called, in another `MergeRuns`.  The latter is a special algorithm inheriting from `MultiPeriodGroupAlgorithm` with special treatment for multi-period data. The error we see arises [here ](https://github.com/mantidproject/mantid/blob/master/Framework/API/src/MultiPeriodGroupWorker.cpp#L290) .  And I'm not sure it is useful to put the `MergeRuns` call in try-catch. 

For MUSR, I could not see the attachment in the email, can you post here perhaps?

KarlPalmen:
I still think the original Load error messages tell the user about the situation better. If this would take a long time to implement, I'd accept it as a separate issue.

KarlPalmen:
Also I don't know how to send files (such as the MUSR files) through GIT.

gvardany:
Yes I think this is more of an issue of MergeRuns itself with multi period data, which seems beyond this PR.
Can you send the files on slack? (normally you can also upload here after appending .zip to bypass the extension check)

KarlPalmen:
Dear Gagik

I try with zip file.

Karl

KarlPalmen:
Do ILL use data files with a group of workspaces? 

If no, I can approve this for files without groups and raise a separate issue to make it work with groups.

gvardany:
There is no loader at ILL that produces a group workspace. However the issue you note about MUSR sounds like a bug, so I would prefer to have it fixed here in the first place.

gvardany:
I have tried with `MUSR00022725.nxs` (1 period) and `MUSR00015192.nxs` (2 periods) in system test data. When I use `LoadAndMerge` with +, I get a single output workspace. But when I `Load` them individually and call `MergeRuns` I get identical output.

gvardany:
Could you try again in a clean session? I suspect you run into confusing behaviour, since while loading it names the runs after the run number, so if you have the same runs already loaded into your ADS, it will overwrite and shuffle the existing groups.

KarlPalmen:
I set the instrument to MUSR and ran LoadAndMerge with `15196+15197` in input and `sum` as output and got 3 workspaces shown, which were the two workspaces of 15196 and `sum`. 

Afterwards I ran it again for `22725+15192` and `sum2` and only `sum2` was shown.

If you want the two MUSR files I tried. I suggest you E-mail me at the address I'll give you via slack. Then I can E-mail them directly to you.

gvardany:
I think I have fixed the problem, it was renaming the group workspace inappropriately, hence leaving it's items orphaned. Please test again.
Also, could you please test the hidden output option? That is if your output name starts with double underscore, you should not see anything, unless you select to show hidden workspaces in mantid properties.

KarlPalmen (approved):
This has worked out well. Also I appreciate the extra unit tests. 
The issue of error messages when group workspaces do not match has been put into #21349 .

PR #21297, opened 2017-11-29
Fix common widgets tests when Qt5 libraries are present 

martyngigg:
**Description of work**

Th plugin library loader has been extended to allow the excluded set of libraries to be defined externally. This allows the `InterfaceManager` at the Qt layer to exclude plugins that are not built for the version of Qt it is compiled against. The properties file has also been extended to allow exclusions of libraries for just the framework so prevent any Qt libraries being loaded when the framework is used alone.

**To test:**

* build the project with `-DENABLE_WORKBENCH=ON`.
* run the common widgets tests: `ctest -R MantidQtWidgetsCommonTest`
* there should be no segfaults

*No associated issue*

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


rosswhitfield (dismissed):
LGTM All the `MantidQtWidgetsCommonTest` unit tests (including `Qt5`) now pass with Arch and RHEL7.

peterfpeterson (commented):
The problem still persists on f26 with python3. It works fine with f26 on python2. Don't merge this quite yet.

rosswhitfield:
It works on Arch with python 3. What are you seeing?

peterfpeterson:
It is the exact issue that this supposedly fixes. Since it is python3 specific, I'm curious if it is a missing python3 library.

peterfpeterson:
There is a positive correlation with `-DMAKE_VATES` and this not working.

rosswhitfield:
I see the same thing, building with paraview causes the failures again

peterfpeterson:
The segfault happens while attempting to load `libMantidVatesAlgorithms.so`.

quantumsteve:
@peterfpeterson This [patch file](https://github.com/mantidproject/mantid/files/1519784/qt5.txt) appears to remove the Qt dependence from `libMantidVatesAlgorithms.so`.

martyngigg:
Thanks for spotting this. Strangely enough on my machine the Qt dependence doesn't appear in the library - I'm not sure why.

@quantumsteve I couldn't quite get it to build with that patch as I had to add VatesAPI back to the link list. It turns out though that `VatesAPI` was including the `qwt_double_interval` header  and also linking to Qwt5 but not actually using any functionality. I've just removed this dependency so hopefully this I will improve things. Let me know.

peterfpeterson:
The same tests are still broken. 
```
$ ldd bin/libMantidVatesAlgorithms.so  | grep -i qt
	libvtkGUISupportQt-pv5.4.so.1 => .../paraview/build/lib/libvtkGUISupportQt-pv5.4.so.1 (0x00007f0621dc7000)
	libQtGui.so.4 => /lib64/libQtGui.so.4 (0x00007f06210a3000)
	libQtCore.so.4 => /lib64/libQtCore.so.4 (0x00007f0620ba0000)
```

martyngigg:
I don't understand why this doesn't show up in my build. Is this Python 3 only?

rosswhitfield:
No I see it in python 2 and 3.

martyngigg:
Could I have something different about my ParaView build? I don't see the Qt dependencies in my tree at all, even before my changes.

martyngigg:
@quantumsteve Your patch does actually work for me now. I've pushed that and let's see what happens.

rosswhitfield:
Maybe this is a good time to set `-DENABLE_WORKBENCH=ON` on PR builds, then let the build server decide? We just need to make sure we have all the required dependencies installed.

martyngigg:
@rosswhitfield I like that idea

peterfpeterson:
Looking at the dependencies from the tip of this branch
```
$ ldd bin/libMantidVatesAlgorithms.so  | grep -i qt
	libvtkGUISupportQt-pv5.4.so.1 => .../paraview/build/lib/libvtkGUISupportQt-pv5.4.so.1 (0x00007ff1e9cbe000)
	libQtGui.so.4 => /lib64/libQtGui.so.4 (0x00007ff1e8f9a000)
	libQtCore.so.4 => /lib64/libQtCore.so.4 (0x00007ff1e8a97000)
```
It looks like our problems come from linking in `vtkGUISupportQt` which is connected to Qt4. I'm still confused as to why it doesn't fail all the Qt5 tests.

martyngigg:
@peterfpeterson Thanks. The annoying thing is I can't find what brings that dependency in and it's not listed in my `ldd` output for some reason. I think it is only failing some Qt5 tests because it's only the ones that load the plugins that have a problem.

martyngigg:
So I've tracked down where the Qt dependence actually comes from on RHEL7. `VatesAlgorithms` needs `VatesAPI` and `VatesAPI` needs `vtkPVVTKExtensionsDefault` for the `vtkPVGlyphFilter`.  `vtkPVVTKExtensionsDefault` is linked to `vtkPVVTKExtensionsRendering` which in turn links to `vtkGUISupportQt`.

Annoyingly `ldd --unused libvtkPVVTKExtensionsDefault.so  | grep -i qt` shows that `vtkGUISupportQt` is unused but this doesn't seem to make any difference.

I'm not sure what to do here.

martyngigg:
This patch:
```
diff --git a/ParaViewCore/VTKExtensions/Rendering/module.cmake b/ParaViewCore/VTKExtensions/Rendering/module.cmake
index f92b157..942c6ea 100644
--- a/ParaViewCore/VTKExtensions/Rendering/module.cmake
+++ b/ParaViewCore/VTKExtensions/Rendering/module.cmake
@@ -13,10 +13,6 @@ if(PARAVIEW_ENABLE_MATPLOTLIB)
   list(APPEND __dependencies vtkRenderingMatplotlib)
 endif()
 
-if (PARAVIEW_ENABLE_QT_SUPPORT)
-  list(APPEND __dependencies vtkGUISupportQt)
-endif()
-
 if("${VTK_RENDERING_BACKEND}" STREQUAL "OpenGL")
   list(APPEND __dependencies vtkRenderingLIC)
   if (PARAVIEW_USE_MPI)
```

to ParaView seems to do the trick. @quantumsteve Does this seem resonable? We should probably also submit it upstream.

peterfpeterson:
The tip of this branch works on f26 with paraview 5.4.1

peterfpeterson (approved):
I think this is ready once the builds pass

martyngigg:
The builds have finally passed :tada:

Framework/API/src/FrameworkManager.cpp
```diff
@@ -461,18 +391,110 @@ bool FrameworkManagerImpl::deleteWorkspace(const std::string &wsName) {
...
+ * @param excludeKey A string
+ */
+void FrameworkManagerImpl::loadPluginsUsingKey(const std::string &locationKey,
+                                               const std::string &excludeKey) {
+  const auto &cfgSvc = Kernel::ConfigService::Instance();
+  const auto pluginDir = cfgSvc.getString(locationKey);
+  if (pluginDir.length() > 0) {
+    std::vector<std::string> excludes;
+    const auto excludeStr = cfgSvc.getString(excludeKey);
+    boost::split(excludes, excludeStr, boost::is_any_of(";"));
```
quantumsteve: This works fine, but I'm trying to standardize on `Kernel::StringTokenizer`.

quantumsteve:
At some point we should add the ParaView 5.4.1 update to the release notes.

PR #21299, opened 2017-11-30
Notify GDPP on Settings Change

### LINKED ISSUE #21353: Expose the Q-summation functionality on the Settings tab
As described in #21167, this issue is to Expose the Q-summation functionality on the Settings tab via via SummationType and ReductionType drop downs as on the RROA dialog.

edwardb96:
This PR removes the need for the m_settingsChanged flag in the GenericDataProcessorPresenter by making the Reflectometry GUI actively notify it when the settings are changed, this eliminates the need for the `areSettingsChanged` routine and should make it easier to switch the parallelisation model in the `GenericDataProcessorPresenter`.

Additionally it includes the changes from #21364 with the test + compilation failures fixed.
> Exposes the `ReductionType` and `SummationType` options of `ReflectometryReductionOneAuto` on the > Reflectometry GUI settings tab via two comboboxes.

**To test:**
1. Ensure that the Reflectometry Settings tab still correctly passes the settings to the reduction algorithms.
2. Ensure that the SANS v2 GUI continued to function as expected.

Fixes #21353. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


qt/widgets/common/src/DataProcessorUI/TwoLevelTreeManager.cpp
```diff
@@ -111,6 +111,12 @@ std::vector<Command_uptr> TwoLevelTreeManager::publishCommands() {
   return commands;
 }
 
+void TwoLevelTreeManager::invalidateAllProcessed() {
+  for (auto i = 0; i < m_model->rowCount(); i++) {
+    setProcessed(false, i);
+  }
```
gemmaguest: I think this needs to loop through the group/row hierarchy in a similar way to [here](https://github.com/mantidproject/mantid/blob/master/qt/widgets/common/src/DataProcessorUI/QTwoLevelTreeModel.cpp#L221)

gemmaguest (changes requested):
The changes look good, but I think there is an error in setting the processed state of rows in the two level tree - see inline comments

gemmaguest:
The changes look good now and the reflectometry GUI works as expected. Rows are reprocessed when settings/data change but not otherwise.

There seems to be a problem on the SANS GUI - if you load the maskfile and batch file from the sample data and tick `PlotResults`, then the workspaces are created ok but the plot window does not show any graphs.

gemmaguest:
The SANS GUI seems to be working fine now. Changes to the settings on the runs tab or the settings tab seem to be picked up ok. Therefore all looks good to me :+1:

gemmaguest (approved):

martyngigg:
@edwardb96 It looks like there are some conflicts after #21343 was merged.

Issue #21230, opened 2017-11-21
Instrument View in-memory Instrument Refactor


LamarMoore:
This issue involves re-factoring the instrument view to make use of `ComponentInfo` and `DetectorInfo` which should result in speed-ups by getting around the recursion used by `ComponentAssemblyActor`

- [x] Renderer/Handler Cleanup #21248
- [x] Replace Actors with ComponentInfo #21339
- [x] Performance optimizations #21582

PR #21310, opened 2017-12-01
21248 renderer cleanup

### LINKED ISSUE #21248: Cleanup around Geometry Renderers and Handlers
Code cleanup to remove code duplication and consolidate Rendering features.

LamarMoore:
**NB. This will be merged into a long running feature branch #21230**
This work involves a code cleanup around Geometry Renderers/Handlers which will allow for extending capabilities to perform volume rendering (required for voxel detectors in ESS instruments). This also helps to clarify the interplay between objects (and components) and the associated handler/renderer. Although many instances of unnecessary virtualisation were removed, I do not expect major speed improvements from this PR.

Points to note:
- The `Renderer` class centralizes rendering of all geometry types (replaces `OCGeometryRenderer` etc)
- The `GeometryTriangulator` class handles triangulation of non-standard types (replaces `OCGeometryGenerator` and `CacheGeometryGenerator`)
- The `ShapeInfo` class stores object shape information (replaces `GluGeometryHandler` specifics)
- The `GeometryHandler` class has been modified, is no longer a base class, and handles all geometry types, hence the removal of virtualization (consolidates `CacheGeometryHandler`, `BitmapGeometryHandler`, `StructuredGeometryHandler`, `GluGeometryHandler`, `OCGeometryHandler`)

**To test:**

-Thorough code review.
- Ensure all unit tests related to Object pass
- Ensure instruments are still displayed correctly in the instrument view.

Fixes #21248 . 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


mantid-roman (changes requested):
If I delete a vtp file for an instrument from the geometryCache folder I cannot load that instrument: the algorithm hangs.

Framework/Geometry/inc/MantidGeometry/Rendering/ShapeInfo.h
```diff
@@ -0,0 +1,96 @@
...
+class MANTID_GEOMETRY_DLL ShapeInfo {
+public:
+  enum class GeometryShape {
+    NOSHAPE = 0,
+    CUBOID,            ///< CUBOID
+    HEXAHEDRON,        ///< HEXAHEDRON
+    SPHERE,            ///< SPHERE
+    CYLINDER,          ///< CYLINDER
+    CONE,              ///< CONE
+    SEGMENTED_CYLINDER ///< Cylinder with 1 or more segments (along the axis).
```
mantid-roman: I believe `SEGMENTED_CYLINDER` is never used. Could you remove it please?

mantid-roman (changes requested):

Framework/Geometry/inc/MantidGeometry/Rendering/ShapeInfo.h
LamarMoore: It's used here https://github.com/mantidproject/mantid/blob/143ecb071b1db696d625c50b141e6a3e5e1bbbcf/Framework/Geometry/src/Rendering/Renderer.cpp#L252

mantid-roman: I don't think any instrument creates a `SEGMENTED_CYLINDER`

LamarMoore: But should we introduce an instrument which requires this we will have to replace it.

mantid-roman (approved):

martyngigg:
I've always wondered whether it is necessary that we have the rendering code within the framework - I think this is the only reason we need the OpenGL dependencies linked to Geometry? It would be nice to be able to build the framework without these, for headless machines for example. Any thoughts?

LamarMoore:
@martyngigg I guess now the renderers have been consolidated, they could probably be moved out but where? Qt4 seems like the wrong place for this stuff.

mantid-roman:
@martyngigg That would be great but I also wanted to get rid of the actors.

martyngigg:
@LamarMoore We don't really have an intermediate layer so my first though was actually in the InstrumentView library as that's the only place that currently uses it but that then means nothing else can. Perhaps we should just create a separate library?

@mantid-roman That would be good.

I'm going to merge this as this is unrelated to these changes.

Framework/Geometry/src/Objects/Object.cpp
```diff
@@ -1120,10 +1113,10 @@ double Object::triangleSolidAngle(const V3D &observer,
     //
     return rayTraceSolidAngle(observer); // so is this
   }
-  double *vertices = this->getTriangleVertices();
-  int *faces = this->getTriangleFaces();
+  const auto &vertices = this->getTriangleVertices().get();
```
OwenArnold: You have gone through the trouble of making this `boost::optional` but you don't need to check that it is valid before extracting?

Framework/Geometry/test/ShapeInfoTest.h
```diff
@@ -0,0 +1,138 @@
...
+  void testEquality() {
+    V3D center(0, 2, 1);
+    double radius = 10;
+    m_shapeInfo.setSphere(center, radius);
+    ShapeInfo shapeInfo2, shapeInfo3;
+    shapeInfo2.setSphere(center, radius);
+    shapeInfo3.setCuboid(V3D(), V3D(), V3D(), V3D());
+    TS_ASSERT_EQUALS(shapeInfo2, m_shapeInfo);
+    TS_ASSERT_DIFFERS(shapeInfo3, m_shapeInfo);
+  }
```
OwenArnold: Good

Framework/Geometry/inc/MantidGeometry/Rendering/ShapeInfo.h
```diff
@@ -0,0 +1,91 @@
...
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+File change history is stored at: <https://github.com/mantidproject/mantid>
+Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+namespace detail {
+class MANTID_GEOMETRY_DLL ShapeInfo {
```
OwenArnold: I'm not sure about this type name. This looks like a creation object based on it's signature. We have tried to use our modern `...Info` classes (`SpectrumInfo`, `DetectorInfo`, `ComponentInfo`) as flat objects that can be indexed into, i.e. used as "Structure Of Arrays", maybe we can discuss.

Framework/Geometry/inc/MantidGeometry/Rendering/Renderer.h
```diff
@@ -0,0 +1,120 @@
...
+namespace Geometry {
+class RectangularDetector;
+class StructuredDetector;
+class IObjComponent;
+
+namespace detail {
+class GeometryTriangulator;
+}
+
+/** Renderer : Handles rendering details of geometry within mantid.
```
OwenArnold: Could probably say a bit more about this. Does not handle rendering of geometry concerned to md workspaces for example,  this is concerned with objects of geometry.

Framework/Geometry/inc/MantidGeometry/Rendering/Renderer.h
```diff
@@ -0,0 +1,120 @@
...
+  void renderBitmap(const RectangularDetector &rectDet,
+                    RenderMode mode = RenderMode::Basic) const;
+  /// Renders structured geometry (used for rendering StructuredDetector)
+  void renderStructured(const StructuredDetector &structDet,
+                        RenderMode mode = RenderMode::Basic) const;
+
+private:
+  mutable RenderMode m_renderMode;
+  /// Renders a sphere
+  void doRenderSphere(const ShapeInfo &shapeInfo) const;
```
OwenArnold: Minor or minor points, but you could drop the "do" from these function names. There's no namespace clash with these.

Framework/Geometry/src/Instrument/RectangularDetector.cpp
```diff
@@ -45,7 +46,7 @@ RectangularDetector::RectangularDetector()
       m_minDetId(0), m_maxDetId(0) {
 
   init();
-  setGeometryHandler(new BitmapGeometryHandler(this));
+  setGeometryHandler(new GeometryHandler(this));
```
OwenArnold: I can see that this was done originally, but would be better to move away from explicit new and pointers for this API. Since `ObjComponent` should die off, maybe this wouldn't be worth fixing here?

OwenArnold (commented):
Pretty good. Mostly questions to resolve before merging.

Framework/Geometry/src/Instrument/RectangularDetector.cpp
LamarMoore: `RectangularDetector` and other things like this should die off so it can probably be fixed later on in the release branch.

PR #21326, opened 2017-12-04
NormaliseToMonitor detector scan improvements

### LINKED ISSUE #21025: NormaliseToMonitor Improvements
When testing for D20 at the ILL a number of issues were noticed when testing `NormaliseToMonitor` on a large workspace.

1. The algorithm is slow for detector scan workspaces - ~~it should be possible to take advantage of the guaranteed indexing in the sync scan case - see `DetectorInfo::linearIndex()`~~ there is a nested loop which could be inverted or avoided
1. The algorithm dialogue is slow to open - the check for a monitor ID being present in a workspace should be moved to the execution
1. There is a redundant warning for single count workspaces about integration ranges
1. Having a separate monitor workspace is not supported properly for detector scan workspaces, either it should support these fully or throw and error
1. Progress tracking should be added (if the algorithm still takes some time after the above changes)

ianbush:
Some minor tidy-ups to the algorithm.

1. The divide step for detector scan workspaces is now done in an OpenMP loop
1. The check for a monitor ID being in a workspace is only done if there are less than 10,000 histograms present
1. There is a redundant warning for single count workspaces about integration ranges that has been removed
1. Having a separate monitor workspace is not supported properly for detector scan workspaces - now a proper error is thrown
1. Progress tracking has been added

**To test:**

Use the attached file for D2B - [d20_det_scan.zip].(https://github.com/mantidproject/mantid/files/1526808/d20_det_scan.zip)

1. Check running `NormaliseToMonitor` is quicker on this branch than master
1. Check the algorithm dialogue is quicker to open for this workspace than on master
1. Check no "Integration range" warning is shown
1. Use `ExtractMonitors` to create a separate detector and monitor workspace. Try running `NormaliseToMonitor` giving the separate monitor workspace - an error should be thrown.
1. Check progress tracking is sensible for running on the D20 workspace

Fixes #21025.

**Release Notes** 
*Does not need to be in the release notes.* This is a tidy-up for a previous ticket to support scanning workspaces in `NormaliseToMonitor`.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Algorithms/src/NormaliseToMonitor.cpp
```diff
@@ -115,7 +115,14 @@ bool MonIDPropChanger::monitorIdReader(
   }
   // are these monitors really there?
   // got the index of correspondent spectra.
-  std::vector<size_t> indexList = inputWS->getIndicesFromDetectorIDs(mon);
+  // Only check if number of histograms is small, else this takes too long!
+  std::vector<size_t> indexList;
+  if (inputWS->getNumberHistograms() < 100000) {
+    indexList = inputWS->getIndicesFromDetectorIDs(mon);
+  } else {
+    indexList = std::vector<size_t>(mon.begin(), mon.end());
```
SimonHeybrock: Not sure I understand this. This seems to be setting `workspace index = detector ID`?

SimonHeybrock (changes requested):
Question below regarding the monitor indices.

Probably the mapping from monitor detector ID to workspace index could also be done faster, via `SpectrumInfo` and `DetectorInfo` (albeit with a little more code).

Framework/Algorithms/src/NormaliseToMonitor.cpp
ianbush: Hmm, looking into this more closely it gets confusing. This is meant to be a list of monitor IDs, so I think the else part is fine. In a previous update I made some changes to this, but it looks like if monitors are missing from the workspace then the vector is reduced in size, but does not correspond to what is actually present?

https://github.com/mantidproject/mantid/pull/20804/files#diff-f54358cb406d12adce22fb09ef086959L124

ianbush: I will see if I can rework this to something sensible. I am also tempted to remove the check as it covers a fairly uncommon case.

SimonHeybrock: I am not sure why you think the `else` part is fine. `getIndicesFromDetectorIDs` translates from detectors IDs to workspace indices. "monitor IDs" is not a defined concept, so I am not sure what you are referring to?

ianbush: To be clear when I mention monitor IDs I mean detector IDs (for detectors that are monitors). This how it has been defined in the `NormaliseToMonitor` docs - see [`MonitorID`](http://docs.mantidproject.org/nightly/algorithms/NormaliseToMonitor-v1.html). I can make the documentation a little clearer, but I am afraid we are stuck with the name until a v2 of the algorithm comes along.

The idea in this bit of code is to get a list of the monitor (detector) IDs in the workspace. First the monitor IDs are obtained from the instrument, then a check is made to see if corresponding spectra exist in the workspace. I skip this second check as it is slow on large workspaces.

SimonHeybrock: What I am saying is that this is not just doing a check, but also a translation to workspace index. You seem to be skipping that translation when skipping the check.

ianbush: I'll also rename this from `indexList`, the name is just wrong.

ianbush: Yes, but the only the size of the translated vector is used, not the contents.

SimonHeybrock: Hmm, indeed, how weird. Where does it find the corresponding spectrum then?

ianbush: I don't see how this would ever have worked correctly before. I have pushed some updates that should improve things, and I think the performance is good enough to remove the check on the number of spectra.

ianbush: It can be tested with:
```python
CreateSampleWorkspace(OutputWorkspace='big1', NumBanks=100, NumMonitors=10, BankPixelWidth=100, NumScanPoints=10)
ExtractSpectra(InputWorkspace='big1', OutputWorkspace='extracted_mons', StartWorkspaceIndex=10000090, EndWorkspaceIndex=10000099)
```

ianbush: I will mark this as in progress and do some tidy up tomorrow.

ianbush: @SimonHeybrock I have pushed the changes, this should work as expected now. You can try the script, then open the dialogue from MantidPlot. It is still a little slow on this workspace, but I think the performance is sufficient to remove the checks on histogram numbers. In most real cases it will now be quick to open now, as only the first spectra are checked if all the monitors are found.

SimonHeybrock (approved):

PR #21328, opened 2017-12-04
ILL Powder Diffraction Detector Scan Reduction

ianbush:
Add a workflow algorithm for powder diffraction reduction at the ILL with detector scans. Should support D2B and D20, although we are lacking data to test D20 with at the moment.

**To test:**

Download the attached test files and add the data directory to Mantid. Be sure to set the facility to ILL. Try running the workflow algorithm with the various options. Check documentation.

[D2B_Example.zip](https://github.com/mantidproject/mantid/files/1625765/D2B_Example.zip)

Fixes #21456.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/PowderDiffILLDetScanReduction.py
```diff
@@ -0,0 +1,156 @@
...
+                             doc='A pair of values, comma separated, to give the minimum and maximum height range (in m). If not specified '
+                                 'the full height range is used.')
+
+        self.declareProperty(WorkspaceGroupProperty('OutputWorkspace', '',
+                                                    direction=Direction.Output),
+                             doc='Output workspace containing the reduced data.')
+
+    def PyExec(self):
+        input_workspaces = self._load()
+        instrument_name = input_workspaces[0].getInstrument().getName()
```
gvardany: I would also validate if the loaded workspaces are time-indexed. This [trick](https://github.com/mantidproject/mantid/blob/master/Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/PowderDiffILLDetEffCorr.py#L164) will do, but I wonder if we should export at least `isScanned` method to python?

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/PowderDiffILLDetScanReduction.py
```diff
@@ -0,0 +1,156 @@
...
+                             + str(supported_instruments))
+
+        self._progress.report('Normalising to monitor')
+        if self.getPropertyValue('NormaliseTo') == 'Monitor':
+            input_workspaces = NormaliseToMonitor(InputWorkspace=input_workspaces, MonitorID=0)
+
+        height_range = ''
+        height_range_prop = self.getProperty('HeightRange').value
+        if (len(height_range_prop) == 2):
+            height_range = str(height_range_prop[0]) + ', ' + str(height_range_prop[1])
```
gvardany: `','.join(height_range_prop)`?

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/PowderDiffILLDetScanReduction.py
```diff
@@ -0,0 +1,156 @@
...
+                                           HeightAxis=height_range,
+                                           OutputWorkspace=output_workspace_name + '_1D')
+            output_workspaces.append(output1D)
+        DeleteWorkspace('input_workspaces')
+
+        self._progress.report('Finishing up...')
+
+        grouped_output_workspace = GroupWorkspaces(output_workspaces)
+        RenameWorkspace(InputWorkspace=grouped_output_workspace, OutputWorkspace=output_workspace_name)
+        self.setProperty('OutputWorkspace', output_workspace_name)
```
gvardany: `
GroupWorkspaces(InputWorkspaces=output_workspaces, OutputWorkspace=output_workspace_name)
self.setProperty('OutputWorkspace', output_workspace_name)
`

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/PowderDiffILLDetScanReduction.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def _load(self):
```
gvardany: You should be able to use LoadAndMerge now.

gvardany (changes requested):
Some comments on the code for the moment.
Just to confirm, was this requested to have all different outputs (1D, 2D, ...) together? I thought only one would be asked at a time. What about ROI and time normalisations?

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/PowderDiffILLDetScanReduction.py
ianbush: Hmm, a good question, this will still work for non-time indexed data so I am tempted to skip this check. I don't know of any, but there could conceivably be a case for D2B where you want to run this where you do not have any time indexing, but still want to do some of the tube magic.

ianbush:
@gvardany Changes should all be pushed. I chose to give the option of multiple outputs as it saves some time, although this is something to check with the scientists to see if is useful (before the release). Normalisation to time and ROI can come in subsequent PRs if required.

gvardany (commented):
Thanks for the update, it's much better now.
I think for D20 we should forbid the 2D options, since they are particularly meaningless and confusing. (I have tried and they have 64 spectra, no idea where this number comes from). Maybe this validation should also be done `SumOverlappingTubes` ?

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/PowderDiffILLDetScanReduction.py
```diff
@@ -0,0 +1,134 @@
...
+            self.log.warning('Running for unsupported instrument, use with caution. Supported instruments are: '
+                             + str(supported_instruments))
+
+        self._progress.report('Normalising to monitor')
+        if self.getPropertyValue('NormaliseTo') == 'Monitor':
+            input_group = NormaliseToMonitor(InputWorkspace=input_group, MonitorID=0)
+
+        height_range = ''
+        height_range_prop = self.getProperty('HeightRange').value
+        if (len(height_range_prop) == 2):
```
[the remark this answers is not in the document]
ianbush: No, it should not be now.

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/PowderDiffILLDetScanReduction.py
```diff
@@ -0,0 +1,134 @@
...
+            height_range = ','.join(height_range_prop)
```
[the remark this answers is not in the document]
ianbush: Indeed, I will revert this to the old way. As it is a float array property it is not a string, as `join` seems to expect.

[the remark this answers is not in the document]
ianbush: I thought I had a test for the height range, but it is missing, so I'll add it to catch problems like this.

docs/source/algorithms/SumOverlappingTubes-v1.rst
```diff
@@ -24,18 +24,18 @@ If the Normalise option is set then the counts for each point in the OutputWorks
...
++++++++++
 
 In this case the x-axis is the true scattering angle of a pixel, so the Debye-Scherrer cones are effectively straightened. The x-axis is strictly positive in this case, the negative angles from the 2D case are put in the same bins as the equivalent positive angles. The height is the pixel height in the tube as for the 2D Option.
 
-1DStraight Option
-+++++++++++++++++
+1D Option
++++++++++
 
 This is effectively the 2DStraight option, with a single bin in the y-axis. For this case only the HeightAxis option can be given as a minimum and maximum.
```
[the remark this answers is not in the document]
ianbush: Ah, this is leftover, I intended to remove all the 'straight' references. It is indeed Lamp terminology, I will fix this.

ianbush:
I need to fix the D20 issue in this PR - https://github.com/mantidproject/mantid/pull/21595. The problem is I gave `panel_1` as the component to get the height axis from insteado of `cell_1`. In any case it makes little sense as you say, but I think the best is for `SumOverlappingTubes` to just generate a warning where the detector is actually 1D.

ianbush:
@gvardany Fixes for the height range pushed. I will try and fix the D20 2D options as soon as this is merged.

gvardany (approved):
Ok I think this one is good to go now. Still for 1D detector, I think it should be a stopping error, when a 2D option is requested.

martyngigg:
The Windows failure is unrelated.

PR #21330, opened 2017-12-04
c and c++ compiler version may be different and hence their allowed compiler options

### LINKED ISSUE #21329: Build failure with gcc , when version of gcc and g++ differ.
With gcc-6 and g++-7, building maintid fails in the make step (that is after cmake configure), because gcc-6 does not recognize the warning option "-Wimplicit-fallthrough". 

This is because the variable GCC_C_COMPILER_VERSION , set in GNUSetup.cmake only 
depends on CMAKE_CXX_COMPILER_VERSION, and does not check CMAKE_C_COMPILER_VERSION.

### Expected behavior
expected make to finish the build

### Actual behavior
build compilation step fails to build minigzip.c

cmake; make

### Platforms affected
All with gcc-6 and g++-7 as default compilers

ebknudsen:
Introduces a test in cmake's code-configuring step that takes the most conservative choice for compiler options, as the c version and c++ compiler options may be different.

In the particular case gcc was 6.4 and g++ 7.2 which leads to compilation failure since the "-Wimplicit-fallthrough" option is unknown to gcc-6.

**To test:**

Fixes #21329. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/buildconfig/CMake/GNUSetup.cmake
+++ b/buildconfig/CMake/GNUSetup.cmake
@@ -17,8 +17,15 @@ endif()
 
 # Set our own compiler version flag from the cmake one and export it globally
 if ( CMAKE_COMPILER_IS_GNUCXX )
-  set( GCC_COMPILER_VERSION ${CMAKE_CXX_COMPILER_VERSION} CACHE INTERNAL "")
-  message( STATUS "gcc version: ${GCC_COMPILER_VERSION}" )
+  set( GCC_C_COMPILER_VERSION ${CMAKE_C_COMPILER_VERSION} CACHE INTERNAL "")
+  message( STATUS "gcc version: ${GCC_C_COMPILER_VERSION}" )
+  set( GCC_CXX_COMPILER_VERSION ${CMAKE_CXX_COMPILER_VERSION} CACHE INTERNAL "")
+  message( STATUS "g++ version: ${GCC_CXX_COMPILER_VERSION}" )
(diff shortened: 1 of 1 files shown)


quantumsteve:
@ebknudsen When would this situation normally come up? Is it something we want to actively support?

ebknudsen:
I honestly don't know when this normally comes up, nor if it commonly happens.
I happens to be the case on my system (debian testing/buster) and I could make a fix for it without too much effort :-), which I thought I'd share. 
IMHO the present situation where all the configure checks pass but the compile fails due to version mismatch, is not desirable. It is better for configure to fail with a message on version mismatch - is that preferrable? 
I suppose another option to avoid the compiler flag entirely - IMHO it issues a warning about a standard language construct which is more often used by choice than by accident.

quantumsteve:
It looks like [gcc-7](https://packages.debian.org/buster/gcc-7) is available for debian testing/buster and g++-7 requires it. Hopefully there's a better workaround than manually specifying the C and C++ compilers.

My preference would be for configure to fail with a message about the version mismatch.

ebknudsen:
Cool. I have commited an edit to this effect. CMake will now fail to configure if gcc and g++ version don't match.

quantumsteve (approved):
Adds a useful error message during configuration.

quantumsteve:
This only only potentially causes an error in the configure stage. The build server failures are unrelated.

PR #21332, opened 2017-12-04
PropertyManager::asJson should not serialize properties that are not enabled

### LINKED ISSUE #21321: PropertyManager::asJson should not serialize properties that are not enabled
`PropertyManager::asJson` will serialize a property `A` depending on the truthiness of other property `B` independent of the value of `B`. This leads to confusion when `B` is `false`. 
In the picture, property `NormRunNumbers` has a value that will not be used. However, the string representation of the algorithm becomes:

```
{"name":"BASISReduction","properties":{"DoIndividual":"1","EnergyBins":"-120,0.4,120","MaskFile":"/SNS/BSS/shared/autoreduce/new_masks_08_12_2015/BASIS_Mask_default_111.xml",
"NoMonitorNorm":"1","NormRunNumbers":"76185-76201","RunNumbers":"75778-75803"},"version":1}
```
Anyone reading this string may think that `NormRunNumbers` was in fact used

screenshot

jmborr:
- Included a condition to check if property is enabled
- unit test

**To test:**
Run the following algorithm declaration in MantidPlot's python window:
```
from __future__ import print_function
class Hello(PythonAlgorithm):
    def __init__(self):
        PythonAlgorithm.__init__(self)
    def name(self):
        return 'Hello'
    def category(self):
        return "General"
    def PyInit(self):
        self.declareProperty('Greet', True, direction=Direction.Input,
                             doc='Do I personally greet you?')
        ifGreet = EnabledWhenProperty('Greet', PropertyCriterion.IsDefault)
        self.declareProperty('Name', '', 'Enter a name')
        self.setPropertySettings('Name', ifGreet)
    def PyExec(self):
        print(self)
AlgorithmFactory.subscribe(Hello)
```
Now execute `Hello` from MantidPlot with some value for `Name`, first with `Greet` option enabled  and then with `Greet` disabled. The  output shown in the log widget for the first case should be:

```
{"name":"Hello","properties":{"Name":"Horace"},"version":1}
```

(I chose "Horace" for `Name`), and for the second case:

```
{"name":"Hello","properties":{"Greet":"0"},"version":1
```

"Horace" is not printed because of disabled `Greet`.

Deserialization:

```
import json
for json_str in ('{"name":"Hello","properties":{"Name": "Horace"},"version":1}',
                 '{"name":"Hello","properties":{"Greet":"0"},"version":1}'):
    x = json.loads(json_str)
    alg = AlgorithmManager.create('Hello')
    alg.initialize()
    for name,value in x['properties'].items():
        alg.setProperty(str(name), value)
    alg.execute()
```

Should result in the following output:

```
{"name":"Hello","properties":{"Name":"Horace"},"version":1}

{"name":"Hello","properties":{"Greet":"0"},"version":1}
```

Fixes #21321

[Release Notes ](https://github.com/mantidproject/mantid/blob/a9c794d35de2a5966c8b37876ea1216d653f83a1/docs/source/release/v3.12.0/framework.rst#core-functionality)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Kernel/src/PropertyManager.cpp
+++ b/Framework/Kernel/src/PropertyManager.cpp
@@ -502,7 +502,7 @@ std::string PropertyManager::asString(bool withDefaultValues) const {
   for (size_t i = 0; i < count; ++i) {
     Property *p = getPointerToPropertyOrdinal(static_cast<int>(i));
     if (p->isValueSerializable() && (withDefaultValues || !p->isDefault()) &&
-            p->getSettings()->isEnabled(this) ) {
+        p->getSettings()->isEnabled(this)) {
       jsonMap[p->name()] = p->value();
     }
   }
--- a/Framework/Kernel/src/PropertyManager.cpp
+++ b/Framework/Kernel/src/PropertyManager.cpp
@@ -501,8 +501,11 @@ std::string PropertyManager::asString(bool withDefaultValues) const {
   const size_t count = propertyCount();
   for (size_t i = 0; i < count; ++i) {
     Property *p = getPointerToPropertyOrdinal(static_cast<int>(i));
-    if (p->isValueSerializable() && (withDefaultValues || !p->isDefault()) &&
-        p->getSettings()->isEnabled(this)) {
+    bool is_enabled = true;
+    if (p->getSettings()) {
+      is_enabled = p->getSettings()->isEnabled(this);
+    }
+    if (p->isValueSerializable() && (withDefaultValues || !p->isDefault()) && is_enabled) {
       jsonMap[p->name()] = p->value();
     }
   }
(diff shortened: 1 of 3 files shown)


peterfpeterson (approved):
It works as advertised.

Framework/Kernel/src/PropertyManager.cpp
```diff
@@ -500,7 +501,12 @@ ::Json::Value PropertyManager::asJson(bool withDefaultValues) const {
   const size_t count = propertyCount();
   for (size_t i = 0; i < count; ++i) {
     Property *p = getPointerToPropertyOrdinal(static_cast<int>(i));
```
quantumsteve: (linting issue) Given that `getPointerToPropertyOrdinal` accepts an `int`, shouldn't the loop variable type be `int` and `count` be cast outside of the loop?

quantumsteve (changes requested):
Needs a little cleanup before being merged.

quantumsteve (approved):

jmborr:
@peterfpeterson @quantumsteve :zap: thanks! :zap:

PR #21363, opened 2017-12-06
Diffraction Reduction - OSIRIS - Remove validation involving comparison of the number of runs

### LINKED ISSUE #21357: OSIRISDiffractionReduction - Providing differing number of Sample and Container Runs fails validation check
Currently, in the OSIRISDiffractionReduction algorithm (and consequently the Indirect > Diffraction Reduction interface), providing a differing number of sample and container runs fails the validation check.

This check should be removed, as files are now associated by D-Range and the smallest subset of runs will be selected, it is therefore possible to have a different number of container and sample runs.

brandonhewer:
Currently, in the OSIRISDiffractionReduction algorithm, the number of input sample, container and vanadium runs are compared during the validation step (must be more sample than vanadium runs, number of container runs and sample runs must be equal if there are one or more container runs). 

Since runs are no longer combined by the order they are supplied and instead by their respective D-Ranges, thus the smallest intersection of the three sets of runs is used. This validation should be removed, as it is both unnecessary and is producing an error when user's wish to change the number of sample/container/vanadium files and not change the other files (e.g. entering D1-5 for Vanadium and Container and varying the sample runs used without being required to change the other two inputs).

**To test:**
1. Navigate to Interfaces > Indirect > Diffraction.
2. Select 'OSIRIS' in the Instrument drop-down menu.
3. Enter `119976, 119977, 119978, 119981` into the 'Sample' field.
4. Enter `71988, 71991, 71990` into the 'Container' field.
5. Enter `119965, 119967` into the 'Vanadium' field.
**Note the runs are provided in varying orders - this is to test that the order is not considered, runs should be organised by d-range**
6. Provide the supplied calibration file into the 'Calibration' field.
7. Click 'Run'.
8. Open the sample logs of the '\*\_dRange' workspace.
9. Ensure the 'D-Ranges' log value is `3.1 - 4.3 Angstrom, 5.2 - 6.2 Angstrom`

**Test files:**
[osiris_041_RES10.zip](https://github.com/mantidproject/mantid/files/1568749/osiris_041_RES10.zip)

Fixes #21357 . 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/OSIRISDiffractionReduction.py
+++ b/Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/OSIRISDiffractionReduction.py
@@ -105,7 +105,7 @@ class DRangeToWorkspaceMap(object):
 
         :param morphism:    The morphism to apply.
         """
-        self._map = {d_range : morphism(d_range, workspace)
+        self._map = {d_range: morphism(d_range, workspace)
                      for d_range, workspace in self._map.items()}
 
     def combine(self, d_range_map, combinator):
@@ -391,7 +391,9 @@ def diffraction_calibrator(calibration_file):
                              XMin=d_range[0], XMax=d_range[1],
                              OutputWorkspace="calibrated_sample",
                              StoreInADS=False, EnableLogging=False)
+
     return calibrator
+
 
 def create_loader(ipf_filename, minimum_spectrum, maximum_spectrum, load_logs, load_opts):
     """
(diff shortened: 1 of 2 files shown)


MatthewBowles (changes requested):
An extra `divided_ws` workspace appears in the ADS (which is unnecessary?)
Also, the D-ranges in the sample logs only includes the 3.1 - 4.3 range, not the 6.2 - 7.3 as stated.

MatthewBowles (changes requested):
The D-range values of the output are correct, but transposed (it says 5.2 - 6.2 first). Also I don't think the workspaces used in the algorithm are deleted properly - there were several left over in the ADS.

MatthewBowles (approved):

PR #21369, opened 2017-12-06
Basic Nexus Geometry Loader

OwenArnold:
This feature gives `LoadEmptyInstrument` the capability to load new-style Nexus Geometry files, which are a complete alternative to IDF based instrument creation. This feature is meant as a demonstrator, but the `NexusGeometry` library that comes with this should easily support future adoption of this format.

Despite the number of changes, given that that all existing behaviour of `LoadEmptyInstrument` is still supported, this feature should be low risk.

**Tester**
* Code Review (I appreciate that is going to be tricky with this volume of changes)
* Ensure that the demonstrator nexus files can be loaded via `LoadEmptyInstrument` 
* Highlight any regressions in existing unit or system tests (there should be absolutely none)


KarlPalmen:
I see there is code to read the Nexus OFF format. How would this code be used in an algorithm to read this format as described in issue #21553 ?

KarlPalmen:
Thank you. I can now use the nexus shape factory instead of the parser.

OwenArnold:
```
Loading Times (s) LoadEmptyInstrument vs LoadEmptyNexusGeometry
loki: old 0.00869798660278, new 0.114725828171
sans2d: old 0.132910013199, new 0.0698039531708
wish: old 1.05436086655, new 0.044634103775
```
Loki (new) uses 2D mesh(s), whereas Loki IDF uses StructuredDetector, this is probably the source of the longer load times.

docs/source/algorithms/LoadEmptyNexusGeometry-v1.rst
```diff
@@ -0,0 +1,38 @@
...
+:ref:`algm-LoadEmptyInstrument`
+
+.. properties::
+
+Description
+-----------
+This algorithm performs a similar function to :ref:`algm-LoadEmptyInstrument`, but builds the instrument geometry from a `NeXus format <https://www.nexusformat.org/>`_ (hdf5) file instead of an Instrument Definition File. 
+
+A fake workspace is created, 1 spectra for every detector in the instrument file. The workspace is comprised of histograms with two bin edges 0 and 1. The data is meanlingless. The workspace merely provides a host for attaching the loaded instrument to. 
+
```
KarlPalmen: 1 spectrum?

OwenArnold: Good spot, thanks Karl

Framework/NexusGeometry/src/NexusGeometryParser.cpp
```diff
@@ -595,6 +602,109 @@ void parseMonitors(const H5::Group &root, InstrumentBuilder &builder) {
   }
 }
 
+class Tube {
+private:
+  Eigen::Vector3d axis;
```
OwenArnold: `m_` prefix consistency

Framework/NexusGeometry/src/NexusGeometryParser.cpp
```diff
@@ -595,6 +602,109 @@ void parseMonitors(const H5::Group &root, InstrumentBuilder &builder) {
   }
 }
 
+class Tube {
```
OwenArnold: `NexusGeometryParser` is big enough, so move to another file.

Framework/NexusGeometry/src/NexusGeometryParser.cpp
```diff
@@ -595,6 +602,109 @@ void parseMonitors(const H5::Group &root, InstrumentBuilder &builder) {
...
+private:
+  Eigen::Vector3d axis;
+  double height = 0;
```
OwenArnold: Should be `const`

Framework/NexusGeometry/src/NexusGeometryParser.cpp
```diff
@@ -595,6 +602,109 @@ void parseMonitors(const H5::Group &root, InstrumentBuilder &builder) {
...
+    // Find point on outer circle
+    auto norm = axis.norm();
+    auto factor = radius / norm;
+    auto pointOnCircle = (axis * factor) + (p1 - centre);
+    points.col(1) = pointOnCircle;
+
+    return NexusShapeFactory::createCylinder(points);
+  }
+
+  Tube(Eigen::Vector3d axis, double radius, Eigen::Vector3d pos, int detID) {
```
OwenArnold: Inputs here need explaining. For example, the magnitude of the axis is important in the calculations (gives pixel height), so must be un-normalised. `pos` needs description.

Framework/NexusGeometry/src/NexusGeometryParser.cpp
```diff
@@ -595,6 +602,109 @@ void parseMonitors(const H5::Group &root, InstrumentBuilder &builder) {
...
+      return true;
+
+    auto denom = denomVec.norm();
+    auto d = num / denom;
+
+    return d == 0.0;
+  }
+};
+
+std::vector<Tube> findTubes(const Mantid::Geometry::IObject &shape,
```
OwenArnold: This important auxiliary function should also have it's own class?/file. Looks like the kind of thing we can and should unit test too.

Framework/NexusGeometry/src/NexusGeometryParser.cpp
OwenArnold: Description of tube required. 

i.e. Supports intermediate calculation. Mutable structure built to house detectors that form part of the tube.

Framework/NexusGeometry/src/NexusGeometryParser.cpp
```diff
@@ -595,6 +602,109 @@ void parseMonitors(const H5::Group &root, InstrumentBuilder &builder) {
...
+  Eigen::Vector3d axis;
+  double height = 0;
+  double radius;
+  std::vector<Eigen::Vector3d> m_positions;
+  std::vector<int> m_detIDs;
+  Eigen::Vector3d p1;
+  Eigen::Vector3d p2;
+
+public:
+  const Eigen::Vector3d &location() const { return m_positions[0]; }
```
OwenArnold: Not clear what location refers to? Is the the start or centre of the tube.

Framework/NexusGeometry/src/NexusGeometryParser.cpp
```diff
@@ -595,6 +602,109 @@ void parseMonitors(const H5::Group &root, InstrumentBuilder &builder) {
...
+    this->axis = axis;
+    this->radius = radius;
+    m_positions.push_back(pos);
+    m_detIDs.push_back(detID);
+
+    p1 = axis + pos;
+    p2 = pos;
+  }
+
+  bool addIfCoLinear(const Eigen::Vector3d &pos, int detID) {
```
OwenArnold: addDetector. 👍 for returning status!

Framework/NexusGeometry/src/NexusGeometryParser.cpp
```diff
@@ -595,6 +602,109 @@ void parseMonitors(const H5::Group &root, InstrumentBuilder &builder) {
...
+    if (checkCoLinear(pos)) {
+      m_positions.push_back(pos);
+      m_detIDs.push_back(detID);
```
OwenArnold: I really think we should be working with detector indexes not IDs if at all possible.

Framework/NexusGeometry/src/NexusGeometryParser.cpp
OwenArnold: Wold be nicer/faster if we knew the approximate size of the tube up front. Is that possible?

Framework/NexusGeometry/src/NexusGeometryParser.cpp
```diff
@@ -595,6 +602,109 @@ void parseMonitors(const H5::Group &root, InstrumentBuilder &builder) {
...
+    auto temp = pos + axis; // unecessary
+    auto numVec = ((p2 - p1).cross(p1 - temp));
+    auto denomVec = (p2 - p1);
+    auto num = numVec.norm();
+
+    if (num == 0)
+      return true;
+
+    auto denom = denomVec.norm();
+    auto d = num / denom;
```
OwenArnold: You might not need to calculate `denom` at all given that you only want to know if it is 0 or not.

Framework/NexusGeometry/src/NexusGeometryParser.cpp
OwenArnold: I would say that the following should be put under test and then refactored. Can probably reduce this to much fewer lines.

Framework/NexusGeometry/src/NexusGeometryParser.cpp
```diff
@@ -595,6 +602,109 @@ void parseMonitors(const H5::Group &root, InstrumentBuilder &builder) {
...
+
+    return d == 0.0;
+  }
+};
+
+std::vector<Tube> findTubes(const Mantid::Geometry::IObject &shape,
+                            const Pixels &positions,
+                            const std::vector<int> detIDs) {
```
OwenArnold: Do we really need detIDs? do we really want to copy by value?

Framework/NexusGeometry/src/NexusGeometryParser.cpp
```diff
@@ -595,6 +602,109 @@ void parseMonitors(const H5::Group &root, InstrumentBuilder &builder) {
...
+  auto axis =
+      Kernel::toVector3d(shape.getGeometryHandler()->shapeInfo().points()[1]);
+  auto radius = shape.getGeometryHandler()->shapeInfo().radius();
+  std::vector<Tube> tubes;
+
+  tubes.emplace_back(Tube(axis, radius, positions.col(0), detIDs[0]));
```
OwenArnold: axis and radius relate to individual pixels of potential tube.

Framework/NexusGeometry/src/NexusGeometryParser.cpp
```diff
@@ -595,6 +602,109 @@ void parseMonitors(const H5::Group &root, InstrumentBuilder &builder) {
...
+        newEntry = false;
+        break;
+      }
+    }
+
+    if (newEntry)
+      tubes.push_back(Tube(axis, radius, positions.col(i), detIDs[i]));
+  }
+
+  return tubes;
```
OwenArnold: I think there is an assumption here that any bank shape that involves cylinders automatically implies tubes. I would say that if a tube only ends up containing a single detector then it is not a tube and should not be treated as such? If you agree maybe `Tube` needs an `isValidTube` method? I guess it partly depends on the overhead of having the wrapping Assembly?

Framework/NexusGeometry/src/NexusGeometryParser.cpp
```diff
@@ -625,16 +735,29 @@ extractInstrument(const H5File &file, const Group &root) {
...
-      Eigen::Vector3d relativePos = detectorPixels.col(index);
-      builder.addDetectorToLastBank(name, detectorIds[index], relativePos,
-                                    shape);
+    auto shape = parseNexusShape(detectorGroup, searchTubes);
+
+    std::vector<Tube> tubes;
+    if (searchTubes) {
+      tubes = findTubes(*shape, pixelOffsets, detectorIds);
+      for (size_t i = 0; i < tubes.size(); ++i) {
+        auto name = bankName + "_tube_" + std::to_string(i);
```
OwenArnold: Minor point as before with constructing strings for these names.

Framework/NexusGeometry/src/NexusGeometryParser.cpp
```diff
@@ -543,6 +548,7 @@ parseNexusShape(const Group &detectorGroup) {
   }
   // Give shape group to correct shape parser
   if (shapeType == NX_CYLINDER) {
+    searchTubes = isGroup;
```
OwenArnold: yes could be tubes. not definite though.

Framework/NexusGeometry/src/NexusGeometryParser.cpp
LamarMoore: I think I should probably pass the shape instead of the axis and radius separately.

Framework/NexusGeometry/src/NexusGeometryParser.cpp
LamarMoore: Searching for tubes only occurs where there are a group of cylinders as determined by the contents of the Nexus files. If the Nexus file has a group of detectors where there is only one detector, isn't that a problem with the Nexus file?

Framework/NexusGeometry/src/NexusGeometryParser.cpp
LamarMoore: This is however the only scenario where one could have tubes in a Nexus file. The search which follows would then be definitive.

LamarMoore:
The algorithm `LoadEmptyNexusGeometry` is missing. I think some merge conflicts still exist.

OwenArnold:
@LamarMoore thanks. Could you elaborate on any merge conflict issue you have found that do not result in compiler errors/warnings.

I remove `LoadEmpyNexusGeometry` and rolled everything into `LoadEmpyInstrument`

SimonHeybrock (changes requested):
Looks nice overall with many added tests and good code structure :+1:.  I have only (many) small comments and questions below.

I do not understand the changes to `PanelsSurface`. Is that related to fixing issues with what was previously implemented?

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -57,8 +57,8 @@ const Kernel::Material MeshObject::material() const { return m_material; }
  */
 bool MeshObject::hasValidShape() const {
   // May enclose volume if there are at
-  // at least 4 triangles and 4 vertices
-  return (numberOfTriangles() >= 4 && numberOfVertices() >= 4);
+  // at least 2 triangles and 4 vertices
+  return (numberOfTriangles() >= 2 && numberOfVertices() >= 4);
```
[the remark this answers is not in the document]
OwenArnold: Thanks! good catch, that was a temporary fix that Lamar put in place. This needs to be reverted. This is the exact reason we have `MeshObject2D`

Testing/Data/UnitTest/LOKI_Definition.hdf5.md5
```diff
@@ -0,0 +1 @@
+dfd4bddc159307410168cbd93a6577fd
```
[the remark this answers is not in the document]
OwenArnold: Not easily, but they were generated the other way round using a tool that Matt and I worked on. So going IDF->HDF5 can be done in an automated way. https://github.com/ess-dmsc/python-nexus-utilities.git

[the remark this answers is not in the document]
SimonHeybrock: Would adding such a test be worthwhile?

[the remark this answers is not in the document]
OwenArnold: @SimonHeybrock Yes possibly. I guess we could add such a test to `LoadEmptyInstrument` itself. Unfortunately we would have to use `CompareWorkspaces` algorithm or equivalent to check that they are actually the same.

Framework/DataHandling/src/LoadEmptyInstrument.cpp
```diff
@@ -23,6 +26,25 @@ using namespace Geometry;
...
+    return ((descriptor.isAscii() && descriptor.extension() == ".xml"));
+  }
+  return !instrumentname.empty();
+}
+
+bool isNexus(const std::string &filename) {
+  if (!filename.empty() && !FileDescriptor(filename).isAscii(filename)) {
+    NexusDescriptor descriptor(filename);
+    return descriptor.isHDF(filename) &&
+           descriptor.classTypeExists("NXinstrument");
```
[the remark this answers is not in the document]
OwenArnold: I'm going to use `NXcylindrical_geometry` || `NXoff_geometry` as the test.

Framework/DataHandling/src/LoadEmptyInstrument.cpp
```diff
@@ -53,16 +75,18 @@ int LoadEmptyInstrument::confidence(Kernel::FileDescriptor &descriptor) const {
...
-      "specifying an instrument definition file. Note Filename or "
+      "specifying an instrument definition file. Files can also be .hdf5 or "
+      ".nxs for usage with NeXus Geometry files. Note Filename or "
       "InstrumentName must be specified but not both.");
-  declareProperty(
-      "InstrumentName", "",
-      "Name of instrument. Can be used instead of Filename to specify an IDF");
+  declareProperty("InstrumentName", "",
+                  "Name of instrument. Can be used instead of Filename to "
+                  "specify an IDF");
```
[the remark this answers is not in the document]
OwenArnold: No have not thought that far ahead. For the time being, specifying an `InstrumentName` is tied to loading via the IDF route. In principle, this could be made to work in an identical way to the current route, but with a NeXus file with only the geometry aspects supplying the geometry. However, I believe for the ESS at least, there is a strong desire to always provide the consistent geometry alongside the data itself. i.e. move away from separate or inconsistent (parameter maps make IDF strings inconsistent) definitions.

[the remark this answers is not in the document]
SimonHeybrock: Neverhteless we need to support viewing an instrument without having data?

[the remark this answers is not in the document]
OwenArnold: Yes, we do. I would prefer that this was done via the aforementioned converter, with users still writing IDFs as the entry point for the time being. I don't like the `InstrumentName` parameter as which instrument definition is really being loaded is opaque and this has caused real problems in the past for users so would rather not apply that to the NeXus files if possible.

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -446,6 +446,9 @@ const BoundingBox &MeshObject::getBoundingBox() const {
         minZ = std::min(minZ, vz);
         maxZ = std::max(maxZ, vz);
       }
+      if (minZ == maxZ)
+        maxZ += 0.001;
```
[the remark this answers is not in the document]
OwenArnold: Correct. I'm not 100% clear on why Lamar added this, I assume this was before we could do 2D properly. With the possibility of having a proper 2D representation now, I'm going to remove this.

qt/widgets/instrumentview/src/PanelsSurface.cpp
```diff
@@ -256,15 +255,10 @@ void PanelsSurface::addFlatBankOfDetectors(
...
-
   for (auto detector : detectors) {
     addDetector(detector, pos0, index, info->rotation);
-    // update the outline polygon
     UnwrappedDetector &udet = m_unwrappedDetectors[detector];
-    auto p2 = QPointF(udet.u, udet.v);
-    vert.clear();
-    vert << p0 << p1 << p2;
-    info->polygon = info->polygon.united(QPolygonF(vert));
+    info->polygon << QPointF(udet.u, udet.v);
```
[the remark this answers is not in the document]
OwenArnold: This is Lamar's change, but I think this must be on the basis of 2D detector banks not having overlapping detectors.

Framework/Geometry/src/Objects/MeshObject2D.cpp
```diff
@@ -0,0 +1,472 @@
...
+                                const Kernel::V3D &b, const Kernel::V3D &c) {
+
+  // in change of basis, barycentric coordinates p = A + u*v0 + v*v1. v0 and
+  // v1 are basis vectors
+  // rewrite as v0 = u*v0 + v*v1
+  // i) v0.v0 = u*v0.v0 + v*v1.v0
+  // ii) v0.v1 = u*v0.v1 + v*v1.v1
+  // solve for u, v and check u and v >= 0 and u+v <=1
+
+  // TODO see MeshObject::rayIntersectsTriangle and compare!
```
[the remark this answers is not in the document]
OwenArnold: Yes, I think this need to be left. I'm dealing with `MeshObject` & `MeshObject2D` cross-over in another PR.

Framework/Geometry/src/Objects/MeshObject2D.cpp
```diff
@@ -0,0 +1,472 @@
...
+}
+
+bool MeshObject2D::pointsCoplanar(
+    const std::vector<Mantid::Kernel::V3D> &vertices) {
+  if (!CoplanarChecks::sufficientPoints(vertices))
+    return false;
+
+  Mantid::Kernel::V3D normal = CoplanarChecks::surfaceNormal(vertices);
+  // Check that a valid normal was found amongst collection of vertices
+  if (normal.norm2() == 0) {
```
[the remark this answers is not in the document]
OwenArnold: This is similar to `MeshObject`. I'll address this in the follow up PR.

Framework/Geometry/src/Objects/MeshObject2D.cpp
```diff
@@ -0,0 +1,472 @@
...
+}
+
+double MeshObject2D::volume() const {
+  return 0; // Volume is always 0 for a plane
+}
+
+/**
+ * Returns an axis-aligned bounding box that will fit the shape
+ *
+ * This is not threadsafe
```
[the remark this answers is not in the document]
OwenArnold: This affects `MeshObject` too, I'll tackle both together as part of the cleanup.

Framework/NexusGeometry/src/InstrumentBuilder.cpp
```diff
@@ -0,0 +1,202 @@
...
+  m_lastBank = assembly;
+  m_instrument->add(assembly);
+}
+
+std::unique_ptr<const Geometry::Instrument>
+InstrumentBuilder::createInstrument() const {
+  verifyMutable();
+  // Lock this from further modification. Temporary releases on destruction.
+  Transaction transaction(&m_finalized);
+  (void)transaction;
```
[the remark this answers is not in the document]
OwenArnold: This is not about thread safety, more about reuse of the builder after the move of instrument. Copying an instrument on return is very expensive (my original behaviour). I used this as a mechanism of allowing an incremental build-up of the instrument, but to lock-down further modification after the instrument had been moved out. I'll review this as part of my further cleanup pr.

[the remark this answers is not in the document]
OwenArnold: Transaction as a stack allocated object, is always cleaned up post return. This allows me to switch behaviour as part of the successful return.

Framework/NexusGeometry/src/NexusShapeFactory.cpp
```diff
@@ -0,0 +1,209 @@
...
+}
+
+void createTrianglesFromPolygon(const std::vector<uint16_t> &windingOrder,
+                                std::vector<uint16_t> &triangularFaces,
+                                int &startOfFace, int &endOfFace) {
+  int polygonOrder = endOfFace - startOfFace;
+  auto first = windingOrder.begin() + startOfFace;
+
+  triangularFaces.reserve(triangularFaces.size() + 3 * polygonOrder);
+  for (int polygonVertex = 1; polygonVertex < polygonOrder - 1;
```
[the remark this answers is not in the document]
OwenArnold: Unfortunately nexus is not very exact about the width of the types you use. http://download.nexusformat.org/sphinx/datarules.html?highlight=types#nexus-data-types

Framework/NexusGeometry/src/NexusShapeFactory.cpp
```diff
@@ -0,0 +1,209 @@
...
+                          nexusVertices[vertexNumber + 1],
+                          nexusVertices[vertexNumber + 2]);
+  }
+
+  return NexusShapeFactory::createMesh(std::move(triangularFaces),
+                                       std::move(vertices));
+}
+
+std::vector<Mantid::Kernel::V3D>
+toVectorV3D(const std::vector<Eigen::Vector3d> &nexusVertices) {
```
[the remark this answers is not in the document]
OwenArnold: I'm not going to do this as part of this though.

Framework/NexusGeometry/src/TubeHelpers.cpp
```diff
@@ -0,0 +1,42 @@
...
+        break;
+      }
+    }
+
+    // Create a new tube if detector does not belong to any tubes
+    if (newEntry)
+      tubes.emplace_back(shape, positions.col(i), detIDs[i]);
+  }
+
+  // Remove "tubes" with only 1 element
```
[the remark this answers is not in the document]
OwenArnold: I think what Lamar has done here is to strip out the assemblies that would be created if there is only a single detector as part of the "tube". We could probably just live without this.

Framework/NexusGeometry/src/InstrumentBuilder.cpp
```diff
@@ -0,0 +1,185 @@
...
+  assembly->setRot(Kernel::toQuat(rotation));
+  m_lastBank = assembly;
+  m_instrument->add(assembly);
+}
+
+std::unique_ptr<const Geometry::Instrument>
+InstrumentBuilder::createInstrument() const {
+  verifyMutable();
+  sortDetectors();
+  m_finalized = true;
```
SimonHeybrock: Wouldn't it be simpler to remove the whole `verifyMutable` mechanism? Simply allocate a new empty instrument tree when `createInstrument()` is called, so the builder can be used again if required.

Framework/DataHandling/test/LoadEmptyInstrumentTest.h
```diff
@@ -1043,6 +1043,33 @@ class LoadEmptyInstrumentTest : public CxxTest::TestSuite {
...
+    alg.initialize();
+    alg.setPropertyValue("Filename", "WISH_Definition_10Panels.hdf5");
+    alg.setPropertyValue("OutputWorkspace", "dummy");
+    alg.execute();
+    Mantid::API::MatrixWorkspace_sptr wish_nexus =
+        alg.getProperty("OutputWorkspace");
+
+    // Now re-run
+    alg.setPropertyValue("Filename", "WISH_Definition_10Panels.xml");
+    alg.execute();
```
SimonHeybrock: Are you sure that the second load will not simply return the instrument from the `InstrumentDataService`, based on a matching name?

Framework/DataHandling/test/LoadEmptyInstrumentTest.h
```diff
@@ -1043,6 +1043,33 @@ class LoadEmptyInstrumentTest : public CxxTest::TestSuite {
...
+    Mantid::API::MatrixWorkspace_sptr wish_xml =
+        alg.getProperty("OutputWorkspace");
+
+    const auto &wish_nexus_detinfo = wish_nexus->detectorInfo();
+    const auto &wish_xml_detinfo = wish_xml->detectorInfo();
+    TS_ASSERT_EQUALS(wish_nexus_detinfo.size(), wish_xml_detinfo.size());
+    for (size_t i = 0; i < wish_nexus_detinfo.size(); ++i) {
```
SimonHeybrock: Why not use `DetectorInfo::isEquivalent()`?

Framework/NexusGeometry/src/NexusGeometryParser.cpp
```diff
@@ -0,0 +1,670 @@
...
+}
+
+// Parse sample and add to instrument
+void parseAndAddSample(const H5File &file, const Group &root,
+                       InstrumentBuilder &builder) {
+  std::string sampleName = "sample";
+  H5std_string samplePath = "raw_data_1/sample";
+  Group sampleGroup = root.openGroup(samplePath);
+  auto sampleTransforms = getTransformations(file, sampleGroup);
+  auto samplePos = sampleTransforms * Eigen::Vector3d(0.0, 0.0, 0.0);
```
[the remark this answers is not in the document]
OwenArnold: Given that the Eigen guidance on this is be explicit, I'll change this to be the correct Matrix type rather than an expression.

SimonHeybrock (approved):

PR #21387, opened 2017-12-08
Stop ws locks causing context menus to wait

### LINKED ISSUE #21263: Locks causing menus to not display
Open Mantid and load #MUSR00062260

Then run MaxEnt with the default values (it should be quick, you may need to merge in #20983)

The open the MaxEnt algorithm box again and increase the iterations to 1000. Then run maxEnt

The code should be slower. Now try to right click on one the output workspaces from MaxEnt. A new process will start and will not complete until the lock on the workspace has been removed. Once the lock is removed the right click menu will appear

NickDraper:
The workspace context menu was getting stuck on using an algorithm HasUB to determine weether to show a "Clear UB" option.

These changes add a direct HasOrientedLattice to all IMDWorkspaces and anything that inherits from it, so you don't need to use the algorithm and get stopped by the ws locking.

**To test:**
To test this you need to run something slow like:
``` python
CreateSampleWorkspace(OutputWorkspace='base', Function='Powder Diffraction', NumBanks=50, BankPixelWidth=100)
```
1. Run the alg to create a workspace, right click on the workspace and see the context menu
2. Rerun the same algorithm with the same output ws (this will cause the ws write lock)
3. While the algorithm is running right click on the workspace and get the context menu up (it might be a bit sluggish, but should appear.

Fixes #21263 

**Release Notes** 

docs/source/release/... /framework.rst

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/API/src/IMDEventWorkspace.cpp
```diff
@@ -41,8 +43,18 @@ bool IMDEventWorkspace::threadSafe() const { return !this->isFileBacked(); }
 
 //-----------------------------------------------------------------------------------------------
 
-/**
+/** Returns true is any experiment info sample has an oriented lattice
```
AnthonyLim23: I don't understand this sentence.

docs/source/release/v3.12.0/framework.rst
```diff
@@ -39,6 +39,7 @@ Core Functionality
 - Fixed an issue where certain isotopes could not be accessed using the `Atom` classes, e.g Si28.
 - Added new functionality to ``datasearch.searcharchive`` :ref:`property <Properties File>` to only search the default facility
 - Condition to check if a property is enabled when serializing.
+- Workspace locking no longer prevents simple the read operations required to display the workspace conext menu in Mantidplot.
```
AnthonyLim23: I dont think the 'the' is needed between simple and read

AnthonyLim23 (changes requested):
The IMDEventWorkspace and IMDHistoWorkspace both have a function called hasOrientedLattice. As far as I can tell, the code is duplicated. I think it would be better to put it into the base class or to have a helper class. Then we could only use one test

AnthonyLim23:
it is slow, but seems to work

AnthonyLim23:
there seems to be a conflict. I will check if it works after the recent changes.

AnthonyLim23:
just had a go and it hangs for a while on 100% complete. I also noticed if you run the suggested script twice. Then while it is running the second time if you click show data it hangs for a while and the context menu no longer appears when the workspace is clicked (until the hang is resolved).

NickDraper:
Just merged in master to resolve the conflict, 

Any interaction will be slow as while the context menu operation is not blocked, the CPU will be busy, quite possibly 100%.

Also this change only stops the lock that was preventing the right click context menu appearing.  Other operations that this enables might still be blocked on read locks (although these might be for very good reasons).  Whereas the context menu blocking was an artifact of how it has been implemented previously.

AnthonyLim23 (approved):

PR #21390, opened 2017-12-08
Update KafkaEventListener for new message schema

### LINKED ISSUE #19358: Update KafkaEventListener to use new cross-institution message schema
Use new event schema from here: https://github.com/ess-dmsc/streaming-data-types
Seeking to start of run should be done by timestamp in run messages rather than offset.

These changes will permit multipartition streaming for event data and make the listener suitable for use by ESS and SINQ as well as ISIS.

### LINKED ISSUE #19487: Update KafkaEventListener to use new sample env log schema
Use sample environment log schema from here: https://github.com/ess-dmsc/streaming-data-types

These changes will permit using the EPICS to Kafka forwarder, which was developed at the PSI for the ESS.

### LINKED ISSUE #21389: Stop at end of run option is not robust in Kafka event listener
### Expected behavior
When the "stop" option is chosen for run transition behaviour of the KafkaEventListener, the algorithm should terminate when the end of a run is reached.

### Actual behavior
Occasionally the algorithm does not terminate at the end of run. The probability of this occurring is _much_ higher when running the Kafka broker on the same machine as Mantid (for example using [this docker-compose](https://gist.github.com/matthew-d-jones/151602c00fc88b6e56c914797b33fcd6) script to launch a broker)

matthew-d-jones:
ISIS, PSI and ESS have agreed on message schema which should be suitable for use at the three institutions. These changes update KafkaEventListener for compatibility with these schema.

**To test:**
- Use [this docker-compose script](https://gist.github.com/matthew-d-jones/151602c00fc88b6e56c914797b33fcd6) to launch a Kafka broker and publish example data to it. You'll need docker and docker-compose installed (on ubuntu `apt install docker docker-compose`), then put the script somewhere and run `docker-compose up`
- Launch MantidPlot and set your default facility to TEST_LIVE
- Enter these options in the Live Data interface:
livedata_options
- Hit run, you should end up with a workspace like this (compare the number of events):
result
- You may want to clean up the docker containers
```
docker-compose rm -v
```
and docker images, use `docker images` to list IDs of images, then
```
docker rm <IMAGE_ID>
```

Fixes #19358
Fixes #19487
Fixes #21389

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


martyngigg:
A question regarding the run-transition behaviour: what is the expected behaviour when `restart` is selected? I was expecting it to discard the existing data but it does nothing. Is there only 1 runs-worth of data in the test broker?

The `stop` behaviour works as expected though.

matthew-d-jones:
Yes, the producer container only publishes a single run by default.
You can append the following to the producer service in the docker-compose script:
```yaml
environment:
      SINGLE_RUN: "false"
```
I've just updated the image to expose that, so you'll need to `docker-compose pull` before you `docker-compose up` for it to work.

martyngigg (commented):
Should we add the minimum version required to  `find_package ( LibRDKafka )` ?

martyngigg:
Thanks for the docker script updates. I now see that the workspace is cleared at the run transition and I checked that the `run_number` sample log has been incremented. However, I don't see any additional events coming through. Am I doing something wrong?

matthew-d-jones:
Thanks for catching these things.
- A minimum version for librdkafka is now specified and I've updated the Find Module to figure out what version it has found. For example, if you change the required version to 0.12 you should see a CMake error.
- The disabled tests have been removed. I think you added them when you first created the listener, they required the broker to be accessible to work.
- I've updated `KafkaEventStreamDecoderTest::test_Multiple_Period_Event_Stream` to actually test what it claims to test.
- The "rename" run transition behaviour is fixed... I don't think it is straightforward to cover with tests, but I have an idea I'll try on Monday before asking you to review again.

matthew-d-jones:
@martyngigg this is ready for review again.
The behaviour at the end of the run now has test coverage.
There remains a possibility to lose messages at run transitions, this is not a regression and I've created an issue here #21481

martyngigg:
@matthew-d-jones Thanks. I'll take a look.

martyngigg (approved):
I've tested the new changes and I'm happy with the run transition behaviour now as it stands pending the future fixes in #21481

PR #21396, opened 2017-12-08
add navbar to the DGSPlannerGUI

dpaj:
Andre Podlesnyak had complained about lack of zoom on the DGSPlanner and the inability to see the position of the particular peak in a given trajectory.  Easiest fix was to just add in the standard nagivation toolbar.

### DIFF
--- a/scripts/DGSPlanner/DGSPlannerGUI.py
+++ b/scripts/DGSPlanner/DGSPlannerGUI.py
@@ -18,8 +18,7 @@ except ImportError:
     except ImportError:
         sys.exit("Cannot import navigation toolbar")
 from matplotlib.figure import Figure
-from  mpl_toolkits.axisartist.grid_helper_curvelinear import GridHelperCurveLinear
-import mpl_toolkits.axisartist.floating_axes as floating_axes
+from mpl_toolkits.axisartist.grid_helper_curvelinear import GridHelperCurveLinear
 from mpl_toolkits.axisartist import Subplot
 import numpy
 import copy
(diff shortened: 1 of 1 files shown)


scripts/DGSPlanner/DGSPlannerGUI.py
```diff
@@ -10,6 +10,7 @@
 import mantidqtpython as mqt
 from .ValidateOL import ValidateOL
 from matplotlib.backends.backend_qt4agg import FigureCanvasQTAgg as FigureCanvas
+from matplotlib.backends.backend_qt4agg import NavigationToolbar2QTAgg as NavigationToolbar
```
AndreiSavici: This fails to run on analysis or on my computer It should be `NavigationToolbar2QT` not `NavigationToolbar2QTAgg`

scripts/DGSPlanner/DGSPlannerGUI.py
```diff
@@ -82,6 +83,8 @@ def __init__(self, ol=None, parent=None):
         self.trajfig = Subplot(self.figure, 1, 1, 1, grid_helper=self.grid_helper)
         self.trajfig.hold(True)
         self.figure.add_subplot(self.trajfig)
+        self.toolbar = NavigationToolbar(self.canvas, self)
```
AndreiSavici: I would create a vertical layout, add the canvas widget to it, then the navigation toolbar, and add this layout to `self.layout()`.
In addition, you should remove the save button, or override the `save_figure` function of the navigation toolbar to run the save function of DGSPlannerGUI widget. This will add information about instrument/goniometer/UB, integration ranges

AndreiSavici (changes requested):

scripts/DGSPlanner/DGSPlannerGUI.py
dpaj: OK, this actually depends on the version.  I used on my home machine.  Thank you for changing.

AndreiSavici:
Please fix the Flake8 warnings. After that, it looks fine

AndreiSavici (approved):

Issue #21167, opened 2017-11-10
Reflectometry GUI fixes


gemmaguest:
This is an umbrella issue to keep track of bug fixes and minor features required for the new reflectometry GUI

### Transmission runs
We have had continual problems with providing transmission runs via the Settings tab. The bugs need to be fixed as a priority and we should add more unit tests to make sure we don't break this again in the future. It is also a priority to have some way of specifying transmission runs on a per-angle basis.

We may want to look at this in conjunction with #21218, which does some refactoring to make working with this code much easier.

- [x] (fixed by #21417) Transmission runs entered on the Settings tab are not found #21076
- [x] (fixed by #21417) You should be able to specify the transmission run on the settings tab or runs tab as a run number or a workspace name in the ADS - currently this works on the Runs tab but gives an unhandled exception on the settings tab.
- [x] (fixed by #21495) No parameters are getting passed through to CreateTransmissionWorkspaceAuto #21457
- [x] (fixed by #21762) @gemmaguest Allow transmission runs to be specified on a per-angle basis #20088 

### Other
- [x] (fixed by #21808) Improve time sliced workspace names #21004
- [x] Output workspace names should have suffix instead of prefix #21004
- [x] ~~(#21317 helps with this)~~ When adding a new group to the table, the new group has a single child row, which is where users usually want to start entering run numbers. However, the new row is hard to see and the group is collapsed by default, which users find confusing. We should do something to help prompt users where to enter text, e.g. expand the group, select the child row and focus the first cell ready for editing.
- [x] (fixed by #21417) Properties on the Runs tab should take precedence over properties on the Settings tab

## Priority - Quick wins
These changes are considered important and should be relatively easy:

### General
- [x] (fixed by #21680) Editing settings on the Events and Settings tabs should be disabled during processing to improve robustness and avoid confusion.

### Parameters on the GUI
- [x] (fixed by #21364) Expose the Q-summation functionality on the Settings tab (via `SummationType` and `ReductionType` drop downs as on the RROA dialog). #21353 
- [x] (fixed by #21343) Expose the `CorrectDetectors` tick box on the Settings tab similarly to the RROA dialog. #21337
- [x] (fixed by #21314) Remove the DirectBeam box from the GUI and the underlying `RegionOfDirectBeam` property from `ReflectometryReductionOneAuto`. Scientists at ISIS do not use this and don't think it would work well anyway. #21313
[...]

PR #21417, opened 2017-12-11
Refactoring and bug fixes for DPW and Refl GUI

### LINKED ISSUE #21076: Transmission run entered on Settings tab not found
### Expected behavior
The ISIS Reflectometry GUI should be able to find transmission runs specified on the Settings tab.

### Actual behavior
Mantid reports an error about the workspace not being found in the ADS.

- Open the ISIS Reflectometry GUI
- Set instrument to OFFSPEC
- On the first child row in the tree, enter run number 42746, angle 0.3 and dQ/Q=0.02
- In the Options column enter `ProcessingInstructions=390-415`
- In the Settings tab, enter Transmission Run(s) as 42744
- Back on the Runs tab, process the group. You get an error that Workspace 42744 could not be found in the ADS. (It works ok if you enter this transmission run on the Runs tab rather than in the Settings).
- 42744 has actually been loaded successfully into the ADS, but with the name `TRANS_42744`. I tried entering this in the Settings tab and reprocessing, but then Mantid caught an unexpected exception with the message `Failed to find an instrument with this name in any facility: 'TRANS_' - search object TRANS_`. (Again, this worked ok doing the same thing in the Runs tab.)

### Platforms affected
Ubuntu 16.04, probably all

### LINKED ISSUE #21218: Refactor GDPP::reduceRow
[GenericDataProcessorPresenter::reduceRow](https://github.com/mantidproject/mantid/blob/master/qt/widgets/common/src/DataProcessorUI/GenericDataProcessorPresenter.cpp#L878) is difficult to understand and I'm not sure it's correct. We should attempt to split this into smaller functions and improve comments/naming/testing where appropriate.

In particular, I'm not sure why [this function](https://github.com/mantidproject/mantid/blob/master/qt/scientific_interfaces/ISISReflectometry/ReflRunsTabPresenter.cpp#L406) needs to hard-code the preprocessing properties when they are part of the [PreprocessingAlgorithm](https://github.com/mantidproject/mantid/blob/master/qt/scientific_interfaces/ISISReflectometry/ReflGenericDataProcessorPresenterFactory.cpp#L96). I suspect this may be related to a regression where properties are not being passed through to the `CreateTransmissionWorkspaceAuto` algorithm (mentioned on #21167).

gemmaguest:
This PR adds the following fixes and under-the-hood changes:
- Refactor GenericDataProcessorPresenter::reduceRow into a much shorter, more readable function
- Refactor ReflSettingsPresenter functions to return options as a map rather than constructing a csv string from them which then has to be re-parsed
- Consolidate csv parsing functions so that all components use the functions in ParseKeyValueString.h 
- Fix some issues from #21167:
  - Transmission runs entered on the Settings tab are not found #21076
  - You should be able to specify the transmission run on the settings tab or runs tab as a run number or a workspace name in the ADS - check that this still works.
  - Properties on the Runs tab should take precedence over properties on the Settings tab
- Improve naming of preprocessed output workspaces to preserve "+" in rather than replacing it with "_"

**To test:**

- Open `ISIS Reflectometry`
- Under `Run(s)`, enter `13460` on one row and add another row with `13461+13462`
- On the Settings tab enter the transmission runs = `13463,13464`
- Back on the Runs tab, select the group and hit Process
- Check that output workspaces include `IvsQ_13460_13461+13462`, `TRANS_13463+13464`, `TOF_13460` and `TOF_13461+13462`
- Delete all workspaces except these. Rename the `IvsQ_` workspace to e.g. `original` so that it doesn't get overwritten.
- Repeat with transmission runs = `"TRANS_13436+13464"` (including the quotes) and replacing the run numbers with `TOF_13460` (no quotes) and `"TOF_13461+13462"` (including quotes).
- You should get the same results as before except that `IvsQ_` is now named `IvsQ_TOF_` etc. Run `CompareWorkspaces` against `original` to check.
- There should be no workspaces named `TOF_` or `TRANS_` other than the ones used (i.e. no more workspaces were loaded).

Also check that the bug described in #21076 is fixed

Also open `ISIS SANS v2 experimental` and check that it still works by loading the mask and batch files from the sample data and processing them

Fixes #21218 and fixes #21076 

**Release Notes** 
*Added to the release notes*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Mantid-Matthew:
This all seems fine to me with one exception. 

In the sans GUI the processing options no longer seem to be being passed into the data processor algorithm correctly. 

For example if you open the sans Gui load some data click the plot results checkbox on the bottom right and then click process the data is not plotted. This is most likely to do with the changes made to getProcessingOptionsAsString as this is the function used to pass these variables in.

Matthew-Andrew (approved):
The options are now passed correctly and the original tests of this PR still pass.

PR #21420, opened 2017-12-12
ISIS Powder POLARIS validate chopper mode and make its value persistent

### LINKED ISSUE #21350: ISIS Powder Polaris PDF mode
Having `mode="PDF"` when setting up the Polaris instrument in isis_powder should propagate the default values (such as `pdf_focused_cropping_values`) to the create_vanadium and focus methods. Additionally, it would be nice be able to enter `"pdf"` or `"PDF"` and for a warning to be displayed in the results log if the mode entered is not recognised.
we should check it is still possible to override these in create_vanadium/focus e.g.:
* set mode of inst to pdf
* run vanadium with in rietveld mode
* run focus in pdf using default instrument values from set up

josephframsay:
Previously, the `mode` parameter of `isis_powder` (which can have 2 values, `PDF` and `Rietveld`) did not behave as described in the [documentation](http://docs.mantidproject.org/nightly/techniques/ISISPowder-Polaris-v1.html#mode). It was not case-insensitive, and its value was not held in state in the instrument settings.

`mode` was made case-insensitive, and its value was made to persist after it had been set in one method call according to the scheme set out [here](http://docs.mantidproject.org/nightly/techniques/ISISPowder-Tutorials.html#how-objects-hold-state-in-isis-powder). In addition, the nasty behaviour that `Rietveld` was used as the default (so `mode="pdf"` meant that `Rietveld` values were used) was replaced with proper validation.

In addition the system test reference file, which was using the wrong cropping values, was updated.

**To test:**

Point your data directories to `\\olympic\Babylon5\Public\JoeRamsay\POLARIS` and make sure this is at the top of the list. Setup your instrument with the script below.

```
import os
from isis_powder import Polaris

data_dirs = config["datasearch.directories"].split(";")
config_file = os.path.join(data_dirs[0], "polaris_shared_config.yaml")

pol = Polaris(config_file=config_file)
```
First,
```
pol.focus(mode="blah")
```
should fail with the error
```
ValueError: Invalid chopper mode "blah"
```
Now run
```
pol.focus(run_number=100017, mode="PDF")
pol.focus(run_number=100018)
pol.focus(run_number=100019, mode="rietveld")
```
This may take a few minutes.

The workspace `100017-ResultTOF-1` (a member of the group `100017-Results-TOF-Grp`) should be cropped between 1500 and 19900 (ie the PDF cropping values). `100018-ResultTOF-1` should be cropped in the same window.

`100019-ResultTOF-1` should be cropped between 700 and 30000 - the Rietveld cropping values.

Fixes #21350 

**Release Notes** 
[Here](https://github.com/mantidproject/mantid/commit/f52dc3ace5809ede4429fb6a2fbd9ec5599632b5)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


josephframsay:
*Note*: after a short conversation with Ron about `mode`, the future of this PR is a little unclear. Proper validation and case-insensitivity are good, but POLARIS seem to like having to specify `mode` on every focus, so this behaviour possibly shouldn't be changed. I'll put of making any further decisions until I've discussed with Elliot how `mode` is (or will be) used in `create_total_scattering_pdf`

josephframsay:
*Update*: I thought it was probably best to go ahead with the change as intended - it won't stop POLARIS specifying `mode` on every run, and it makes all the parameters consistent with each other **and** the documentation

ElliotAOram:
Thanks for the update, I'll give it a test once it's through build servers

ElliotAOram (commented):
Code changes look good. Workspace X values don't actually start/end were stated in the PR (they are marginally different). @josephframsay is this because of logarithmic binning?

ElliotAOram:
see difference in binning below:
image

josephframsay:
@ElliotAOram I think this is happening because the bin boundaries don't quite line up with the cropping window. Is this avoidable (and would it actually be worth avoiding)? It seems to me that another rebin would slow the whole thing down quite a lot, and powder people have never complained about this behaviour

ElliotAOram (approved):
@josephframsay I think it's within tolerance and normally data at the min/max is not really used so this should be fine

PR #21423, opened 2017-12-12
LoadMcStas single, multiple and sum of scattering output

### LINKED ISSUE #20369: Allow multiple scattering data to be distinguished with LoadMcStas

ElliotAOram:
Previously the `LoadMcStas` algorithm only returned the sum of single and multiple scattering eventdata. 
This change means the algorithm now outputs both the single and multiple scattering data as well as the sum of the two. 
 
**To test:**

Run the LoadMcStas algorithm
*Note: The required file for the below is in the DocTest data directory*
```
test = LoadMcStas('mcstas_event_hist.h5')
```
* Observe that there are now 3 `EventData` workspaces: `EventData_test, EventData_test_1, EventData_test_2`
* `EventData_test_1` and `_2` should sum to `EventData_test`.
* When showing the instrument on these workspaces you should see that `EventData_test` has two banks, in `EventData_test_1`, `EventData_test_2` contain one of these banks each.

Fixes #20369

**Release Notes** 
Update to framework page in release notes [here](https://github.com/mantidproject/mantid/blob/ec789a914d7075d8b3c679a5a771c10abaafd377/docs/source/release/v3.12.0/framework.rst)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


ElliotAOram:
Doc tests need updating

Anders-Markvardsen:
Good first start.

What I now see loading mcstas_event_hist.h5 is 

capture

Could you change this so that the additional workspaces are listed at the end and with the same name convention as for the histogram workspaces, i.e. for this example EventData_ws_1 -> k01_events_dat_list_p_x_y_n_it_t and Eventdata_ws -> k02_events_dat_list_p_x_y_n_it_t. The reason for the latter is that it allows users of mcstas to easily know what these workspaces are in Mantid. The reason for the former is the benefit that it will not break backward compatability of scripts which (accidentally use index rather the name to refer to a given workspace), and putting these additional event workspaces at the end is fine. 

could you also add a usage example that demonstrate how EventData_ws is the sum of the two additional workspaces for this example please

Anders-Markvardsen:
Looks good when loading into my windows laptop now. However, as you probably already noticed usage example now failing on Ubuntu. I would hope not, but I am wondering if this is a precision issue on Ubuntu, it worrying seem to have put no data in k02_events_dat_list_p_x_y_n_it_t! To reduce future need to change the usage example could you see if you could get the workspaces by name instead of by index

ElliotAOram:
I'm building this now on my ubuntu VM so I can take a better look. Getting workspace by name seems like a good change regardless so I will implement this.

ElliotAOram:
@Anders-Markvardsen this is ready to test again now

Anders-Markvardsen:
Please note when I run an older version of Mantid (in this case 3.8 something) the file in \\olympic\Babylon5\Public\Anders loads fine (although note outputting the event workspace separtely). However, running it with the code from the PR I get: 

capture

Anders-Markvardsen:
This is what I get when I use the Load button on the zip (using mantid 3.8 something):

capture

which is attached with the issue also. I will download a 3.11 version also and try in on a more recent version also

Anders-Markvardsen:
Tested and loading mccode.h5 on 3.11 on my windows machine works file also in that it does crash mantid. Could you look get this PR to work with mccode.h5 also

ElliotAOram:
The reason the mccode file can not be loaded is:
The instrument xml has detectors with the same id. By looking at the xml, this can be seen in the lines:
```
<component type="MonNDtype-0" name="nD_Mantid_0" idstart="0" idfillbyfirst="x" idstepbyrow=" 50">
...
<component type="MonNDtype-1" name="nD_Mantid_1" idstart="0" idfillbyfirst="x" idstepbyrow=" 50">
...
<component type="MonNDtype-2" name="nD_Mantid_2" idstart="0" idfillbyfirst="x" idstepbyrow=" 50">
```
note above the `idstart=0` in all three cases.

By contrast, in the `mcstat_event_hist` file (see doctests), the XML shows that the detectors start at different IDs:
```
<component type="MonNDtype-0" name="nD_Mantid_1" idstart="0" idfillbyfirst="x" idstepbyrow=" 64">
...
<component type="MonNDtype-1" name="nD_Mantid_2" idstart="1000000" idfillbyfirst="x" idstepbyrow=" 64">
```
If you supply mantid with 2 identical detector IDs for one instrument it throws at [exception](https://github.com/mantidproject/mantid/blob/master/Framework/Geometry/src/Instrument.cpp#L756). My guess (without investigating this in too much depth) is this was added with the updates in instrument geometry. I don't know too much about how the mcstas file is created, but if possible, I suggest we should attempt to update the instrument XML to have unique detector Ids.

Anders-Markvardsen:
Well spotted with the detector IDs. However, since the instrument Geometry change Mantid should not simply die and print out a non-sensible message if an IDF accidentally contains same detector IDs. Hence, could you feed this back to the core sub-team please?

ElliotAOram:
@Anders-Markvardsen I'll let them know - as well as looking into why the tests are failing.

Anders-Markvardsen:
When running the attached mcstas nexus file (with unique IDs this time!) 

[mccode_with_unique_IDs.zip](https://github.com/mantidproject/mantid/files/1606269/mccode_with_unique_IDs.zip)

I see:

capture

Hence almost there, but could try to understand why 'single_list_...' is not within the group 'mccode_with_unique_IDs'

ElliotAOram:
@Anders-Markvardsen I can take a look into why that file isn't grouped today (or first thing Monday). Thanks for the improvement to the error handling!

ElliotAOram:
This should now add all partial workspaces to the output group

Anders-Markvardsen (approved):
Looks good and also tested with additional mcstas output

martyngigg:
Failing tests are unrelated.

PR #21333, opened 2017-12-04
Indirect - Modify ApplyPaalmanPings to work with corrections from CalculateMonteCarloAbsorption

brandonhewer:
Currently, the ApplyPaalmanPings interface (and by extension the Apply Paalman Pings interface), will not work when supplied the output from the CalculateMonteCarloAbsorption algorithm (alternatively the Calculate Monte Carlo Absorption interface). This is as a result of four of the Paalman Pings factors being required in the ApplyPaalmanPings algorithm.

This PR adds a two-factor approximation to the ApplyPaalmanPings algorithm (used when only two correction factors are provided) and changes the name of the Apply Paalman Pings interface to Apply Absorption Corrections.

Unit Tests and documentation changes (including workflow diagram) have been included.

**To test:**
1. Navigate to Interfaces > Indirect > Corrections > Calculate Monte Carlo Absorption.
2. Provide the supplied sample and container files.
3. Set 'Sample Width' to `0.3`.
4. Set 'Sample Thickness' to `0.1`.
5. Set 'Sample Height' to `0.5`.
6. Set 'Container Front Thickness' to `0.1`.
7. Set 'Container Back Thickness' to `0.1`.
8. Enter `H2-0` as the chemical formula under 'Sample Details'
9. Enter `Al` as the chemical formula under 'Container Details'
10. Click 'Run'.
11. Navigate to the Apply Absorption Corrections tab.
12. Provide the sample file into the Input field.
13. Check the 'Use Container' check-box and provide the container file.
14. Check the 'Use Corrections' check-box, change the drop-down menu to 'Workspace' and select the created `*_Corrections` workspace group.
15. Click 'Run'.
16. Ensure the algorithm is run successfully, a corrected workspace is added to the ADS and a plot of the sample, container and corrected data are shown in the interface.

**Test Files**
[test-files.zip](https://github.com/mantidproject/mantid/files/1528245/test-files.zip)

Fixes #21316. 

**Release Notes** 

#### Changes have been added to the release notes.
---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [x] Is the code of an acceptable quality?
- [x] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)?
- [x] Are the unit tests small and test the class in isolation?
- [x] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

[...]

PR #21424, opened 2017-12-12
Indirect - CorrectionsTabs - Error handling in the case of a WorkspaceGroup being provided

### LINKED ISSUE #21157: CalculatePaalmanPings GUI crash
The GUI tab crashes if a file containing a workspace group is supplied.

Steps to reproduce:
1. Navigate to Interfaces->Indirect->Corrections->CalculatePaalmanPings Tab.
2. Choose the attached nexus file in the first input file field. (first rename it to get rid of '.zip')
3. Count to 5 -> 💣 

The problem is that it tries to get the instrument from the group workspace directly.

[16658_166445_red.nxs.zip](https://github.com/mantidproject/mantid/files/1457331/16658_166445_red.nxs.zip)

brandonhewer:
~~**Must be merged after #21333**~~

Currently, in the Interfaces > Indirect > Corrections tabs, attempting to supply a WorkspaceGroup to the sample/container input fields, provides no informative error message. Upon clicking the 'Run' button in these interfaces, prior to supplying a WorkspaceGroup, Mantid will currently crash.

This PR adds correct error handling, in the circumstance in which a WorkspaceGroup, or any invalid workspace type is provided to the input fields of the Indirect CorrectionsTabs.

A message box will be displayed, when the invalid workspace type is first selected:
corr_err_message

Along with a message output in the log:
corr_err_log

This PR also improves the error message when the 'Run' button is clicked and a workspace of an invalid type, occupies the sample/container input field:
corr_err_message_run

**To test:**
1. Navigate to Interfaces > Indirect > Corrections.
2. For each tab in the Indirect Corrections interface:
    1. Supply the provided `sample_group_red.nxs` file into the sample field.
    2. Ensure an error message, similar to the one shown above, is displayed and a warning is output to the log.
    3. Click the 'Run' button.
    4. Ensure an error message, similar to the one shown above, is displayed, which includes a line about the sample input not being of the correct type.
    5. If there is a container field, supply the provided `container_group_red.nxs` file into the container field.
    6. In the 'ContainerSubtraction' interface, the same error message received when providing the sample group, should be produced. Otherwise, no error should be produced.
    7. Click the 'Run' button.
    8. Ensure an error message, similar to the one shown above, is displayed, which includes both a line about the sample input not being of the correct type and the same for the container input.

**Test Files:**
[group_test_files.zip](https://github.com/mantidproject/mantid/files/1552406/group_test_files.zip)

Fixes #21157. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


gvardany:
@brandonhewer This has developed a conflict.

gvardany (changes requested):
The error handling seems fine, except I wonder if we should put also the reason of the error in the pop-up message instead of the log behind? This would make it more user friendly. If not at least the log level should be set to error, since it is a blocking issue. 
My only concern is that validation seems to take too long, and in fact, when you browse the file, and immediately click Run, it still crashes. Could you investigate disabling the Run button, while the validation is in progress?

brandonhewer:
@gvardany I agree that showing a message box would be more appropriate here and the extra message you've pointed out, would be very useful here.

On the topic of the issue with the Run button - I've come across this problem before and included it on the MSG board backlog. This issue appears to be universal across Indirect tabs, so I'll create a new issue for this.

gvardany:
I agree the run button issue can be fixed in a separate PR.

gvardany (approved):

PR #21425, opened 2017-12-12
Add negation operators to V2D and V3D

### LINKED ISSUE #19958: V2D and V3D are missing negate operators
The negation `V?D operator-() const;` methods and Python exports could be added to `V2D` and `V3D` classes.

soininen:
This PR adds the unary `operator-()` functions to `V2D` and `V3D`. The negation operator of `V3D` is also exposed to Python (`V2D` itself in not available in Python).

**To test:**

Check that negation of `V3D` objects produce vectors of the same length but in opposite direction. Testing `V2D` might be a bit tricky as it is not available in Python, but since it's not used too much, maybe it's enough to do a throughout code review.

Fixes #19958.

**Release Notes** 

Framework release notes were updated accordingly.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Kernel/inc/MantidKernel/V2D.h
+++ b/Framework/Kernel/inc/MantidKernel/V2D.h
@@ -76,6 +76,8 @@ public:
   V2D operator*(const double factor) const;
   /// Scale this
   V2D &operator*=(const double factor);
+  /// Negate
+  V2D operator-() const noexcept;
 
   ///@}
 
--- a/Framework/Kernel/test/V2DTest.h
+++ b/Framework/Kernel/test/V2DTest.h
@@ -114,7 +114,7 @@ public:
   }
 
   void test_Negate_Works_With_Special_Values() {
-    const V2D p1(1. / 0., 0. / 0.);
+    const V2D p1(INFINITY, std::nan(""));
     const V2D p2 = -p1;
     TS_ASSERT_EQUALS(p2.X(), -INFINITY);
     TS_ASSERT(std::isnan(p2.Y()));
--- a/Framework/Kernel/test/V2DTest.h
+++ b/Framework/Kernel/test/V2DTest.h
@@ -107,6 +107,12 @@ public:
     TS_ASSERT_EQUALS(p1, V2D(9.0, 27.0));
   }
 
+  void test_Negate_Gives_Same_Length_But_Opposite_Direction() {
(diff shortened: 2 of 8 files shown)


Framework/Kernel/inc/MantidKernel/V2D.h
```diff
@@ -76,6 +76,8 @@ class MANTID_KERNEL_DLL V2D {
   V2D operator*(const double factor) const;
   /// Scale this
   V2D &operator*=(const double factor);
+  /// Negate
+  V2D operator-() const noexcept;
```
josephframsay: Can this safely be `noexcept`? For instance, is it OK to assume that `-NaN` or `-inf` won't throw anything?

josephframsay (commented):
These changes are nice, and the new operator behaves as expected on my machine. Just one small query about `noexcept`, but apart from that everything looks good!

Framework/Kernel/inc/MantidKernel/V2D.h
soininen: Yes. I've added unit tests to make sure the behavior won't change unnoticed.

josephframsay: Cool - once the build servers pass this looks good to go from my point of view then

josephframsay (approved):

Framework/Kernel/test/V2DTest.h
```diff
@@ -107,6 +107,19 @@ class V2DTest : public CxxTest::TestSuite {
...
   }
 
+  void test_Negate_Gives_Same_Length_But_Opposite_Direction() {
+    const V2D p1(-3, 9);
+    const V2D p2 = -p1;
+    TS_ASSERT_EQUALS(p2, V2D(3, -9))
+  }
+
+  void test_Negate_Works_With_Special_Values() {
+    const V2D p1(INFINITY, std::nan(""));
```
quantumsteve: `std::nan("")` is correct (no need to change it), but I find [`std::numeric_limits<double>::quiet_NaN()`](http://en.cppreference.com/w/cpp/types/numeric_limits/quiet_NaN) is a bit clearer.

soininen: I didn't know this exists in `std::numeric_limits`. Thanks for the hint!

PR #21435, opened 2017-12-14
MaxEnt not outputing evolution of chi and MaxAngle once it has converged

### LINKED ISSUE #21392: Cut diagnosis MaxEnt output workspaces to shown only up to converged
Change Maxent to output EvolChi and EvolAngle to

* Only to up convergence or if fails until max number of iterations
* With x-axis label 'number of iterations' instead of other label such as 'Time'
* With y-axis likewise showing what is plotted instead of e.g. 'asymmetry'
* For all usage examples add check on number of iterations
* Add unit test where reach max number of iterations, (warning to user that did not converge)

Noticed while testing #21079

AnthonyLim23:
This is to prevent MaxEnt from outputting workspaces containing mostly zeros. Instead we do not return the values after the algorithm has converged. 

**To test:**
run the following script and look at the data for evolChi1 and evolAngle1. Do this with both this version and master (at the same time). The last entry into the workspaces for this PR should be followed by zeros in master. 

```
from datetime import datetime 

Load(Filename=r'EMU00020884.nxs', OutputWorkspace='ws')
CropWorkspace(InputWorkspace='ws', OutputWorkspace='ws', XMin=0.17, XMax=15, EndWorkspaceIndex=10)
ws = RemoveExpDecay(InputWorkspace='ws')
ws = Rebin(InputWorkspace='ws', Params='0.016')
startTime= datetime.now() 
evolChi1, evolAngle1, image1, data1 = MaxEnt(InputWorkspace='ws', A=0.0001, MaxIterations=2500, ResolutionFactor=1, MaxAngle=0.05)
timeElapsed=datetime.now()-startTime 

print('Time elpased (hh:mm:ss.ms) {}'.format(timeElapsed))
```

Fixes #21392. 

**Release Notes** 

In release notes, also added to the release notes for #20983 
---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Anders-Markvardsen (approved):
Looks good and results from it is interesting. I have created one separate issue to understand why simply model takes longer to converge compared to real data, where luckily this is the most important use. An an issue to feed info back on converges per spectrum

Anders-Markvardsen:
I think the build server issues are noise, I am re-running some build to be surer, but gatekeeper could you just double check - in particular the Doxygen failure

martyngigg:
Unrelated test failures.

PR #21443, opened 2017-12-15
Fix leak of H5 attribute in parallel event loader

SimonHeybrock:
The leak caused crashes when using the NeXus API after the even loading via HDF5 in the parallel event loader was complete. The parallel loader is only enabled in MPI builds so this does not affect any normal Mantid build.

Also cleans `LoadEventNexus`, to remove raw owning pointer.

**To test:**

Code review.
No isse and no release notes. The crash was only happing in MPI builds, which are not public.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


AndreiSavici (approved):
Code looks good. Since the builds pass, I would say it's good to go.

PR #21449, opened 2017-12-15
Fix RHEL 7 - GCC7 build

DavidFair:
Description of work.
As part of some personal projects with compilers I've fixed some GCC 7 compilation errors on RHEL7. I thought I'd contribute this back to the project for others to benefit too.

---

Currently this uses the software collections within RHEL to switch between compilers. 
The other nice feature RHEL has is any changes between `libstdc++` 4.8.x (default compiler) and `libstdc++` 5.x onwards is statically compiled into the final binaries. This means it is backwards compatible without requiring end users to install the devtoolset package.

---
There is no reason this shouldn't work on Ubuntu/Debian based systems. However it will fail to link as the default ABI changed between GCC5 and 6. This means Ubuntu 16.04, which uses the C++03 ABI cannot easily use GCC6 onwards without rebuilding boost. (see [here](https://bugs.launchpad.net/ubuntu/+source/llvm-toolchain-3.8/+bug/1664321)). 
Ubuntu 16.10 onwards will work correctly as the default compiler was changed to using the C++14 ABI.

**To test:**
Setup RHEL7 ([see](https://www.mantidproject.org/Building_with_CMake#Red_Hat_7.2FCentOS_7))
Then:
```
sudo subscription-manager repos --enable rhel-server-rhscl-7-rpms
sudo yum install devtoolset-7
```
Switch to bash instance which uses GCC 7 as default compiler
```
scl enable devtoolset-7 bash
mkdir build_gcc7
cd build_gcc7
cmake <path_to_mantid_src>
```

Edit: An upstream bug [here](https://bugzilla.redhat.com/show_bug.cgi?id=1519073) has been introduced into the standard library causing it to fail to link. 
Instead run make with `-i` to ignore link failure and check all files compile correctly, or wait for the patch release on the devtoolset and it should link properly this time.

Run all tests to verify it compiled correctly

Fixes N/A

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/test/PDFFourierTransformTest.h
+++ b/Framework/Algorithms/test/PDFFourierTransformTest.h
@@ -1,21 +1,22 @@
 #ifndef MANTID_ALGORITHMS_PDFFOURIERTRANSFORMTEST_H_
 #define MANTID_ALGORITHMS_PDFFOURIERTRANSFORMTEST_H_
 
-#include <cxxtest/TestSuite.h>
-#include <numeric>
-#include <cmath>
-#include "MantidKernel/Timer.h"
-#include "MantidKernel/System.h"
-#include "MantidKernel/UnitFactory.h"
-#include "MantidDataObjects/Workspace2D.h"
 #include "MantidAPI/AnalysisDataService.h"
 #include "MantidAPI/Axis.h"
 #include "MantidAPI/FrameworkManager.h"
 #include "MantidAPI/IAlgorithm.h"
 #include "MantidAPI/MatrixWorkspace.h"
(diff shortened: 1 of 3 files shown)


quantumsteve:
@DavidFair Do the linking errors exist on earlier versions of the devtoolset?

quantumsteve (approved):
LGTM. I don't see any controversial changes and the builds still pass.

DavidFair:
@quantumsteve - No there wasn't linking errors previously, I compiled and ran all system tests in November with the above changes.

martyngigg:
Thanks, @DavidFair!

PR #21451, opened 2017-12-18
Fix issues in LoadMask

### LINKED ISSUE #21450: Reusing LoadMask instances masks detectors from previous call
`LoadMask` is loading detectors to mask into a vector that is a field of the class and is not cleared before or after use. Using the same `LoadMask` instance to load several mask files will result in the second and subsequent calls including all previously masked detectors.

SimonHeybrock:
Fix two issues of `LoadMask`:

- Detector ID buffers need to be cleared, otherwise reuse will see previously loaded masked detectors.
- Memory leak of Poco::XML::Document.

See also issue description.

**To test:**

Code review.

Fixes #21450.

Not reported by a user, so no release notes, in particular since this is probably not observable with the GUI or the *functional* Python algorithm interface (need to create algorithm instance yourself to see the issue).

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


AntonPiccardoSelg (commented):
Everything looks good to me. The failing test is not related to this PR.

I am wondering how many user scripts out there rely on `LoadMask` to store the state between executions. It should be none, but there is a chance that some do. Would it be worth to make users aware that this behaviour has changed with a remark in the release notes?

SimonHeybrock:
@AntonPiccardoSelg I think it is actually unlikely that users will experience this. If the loaded masks are all applied to the *same* workspace no difference will be visible. I will add a note to the release notes nevertheless.

AntonPiccardoSelg (approved):

PR #21453, opened 2017-12-18
Fix mantidqt.widgets.algorithmselector test under Python 3

martyngigg:
**Description of work**

Fixes the test used to inspect if an attribute is an method of a class for compatibility with both Python 2 and 3. Python 3 no longer has the concept of unbound methods and returns a type `FunctionType` and not `MethodType`.

Note that I have not added the check to the general `py3compat` functions as `isfunction` would only be correct to check if you already no the attributes "belongs" to a class. 

**To test:**

Code review and builds passing should be sufficient.

*No issue*

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


mantid-roman (approved):
Looks good to me.

PR #21455, opened 2017-12-18
Export algorithm dialogs to python

### LINKED ISSUE #21281: Export base algorithm dialog  & custom dialogs to mantidqt
Following on from #21180. After selecting the algorithm the user will run want to run it. This will require access to the algorithm dialogs, which in turn will require defining the code for executing an algorithm in the background.

Tasks:
* Export `AlgorithmDialog`, `GenericDialog`, `InterfaceManager` (plus related classes) to Python inside `mantidqt`. See #21251 
* Define functionality to run the algorithms asynchronously in the background of the workbench. My suggestion would be that we create a module `workbench.external.mantid` and add this functionality there.

mantid-roman:
Used sip to export algorithm GenericDialog and InterfaceManager to python.

**To test:**

Code review. Open dialogs via exported InterfaceManager. Start AlgorithmSelector with `run_test_app.py` and execute algorithms. Check logs to see that it works.

Fixes #21281

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Kernel/src/DllOpen.cpp
```diff
@@ -10,6 +10,8 @@
 
 #include <boost/algorithm/string/predicate.hpp>
 
+#include <boost/algorithm/string/predicate.hpp>
```
AndreiSavici: Why include twice?

qt/python/mantidqt/widgets/src/_widgetscore.sip
```diff
@@ -0,0 +1,98 @@
+%ModuleCode
+#include "MantidQtWidgets/Common/Message.h"
```
martyngigg: I don't think we need the following 2 includes at this level - just `#include "MantidQtWidgets/Common/Message.h"`. The main reason is to allow the `using namespace MantidQt::MantidWidgets`. Can we move the classes in `MantidQt::API` to `MantidQt::MantidWidgets`? I'd likely to slowly kill of the `MantidQt::API` namespace as we don't have the library anymore.

qt/python/mantidqt/test/test_algorithm_dialog.py
```diff
@@ -0,0 +1,59 @@
+import unittest
```
martyngigg: Missing GPL header.

qt/python/mantidqt/test/test_algorithm_dialog.py
```diff
@@ -0,0 +1,59 @@
...
+
+from qtpy.QtWidgets import QWidget, QLineEdit
+
+from mantid.api import PythonAlgorithm
+from mantid.kernel import Direction, FloatArrayProperty
+from mantid.simpleapi import AlgorithmManager, AlgorithmFactory
```
martyngigg: These classes can also be imported from `mantid.api`. I'd like to keep the`simpleapi` as a high-level module for users.

qt/python/mantidqt/test/test_algorithm_dialog.py
martyngigg: Also can we move the test to `mantidqt.widgets.test` with the others? I think this is also missing from the `python/CMakeLists.txt` file so that it is not being picked up by `ctest`.

qt/python/mantidqt/utils/test/test_modal_tester.py
```diff
@@ -0,0 +1,101 @@
...
+from mock import patch
+import unittest
+
+from qtpy.QtWidgets import QInputDialog
+
+from mantidqt.utils.qt.plugins import setup_library_paths
+from mantidqt.utils.qt.testing import ModalTester
+
+
+setup_library_paths()
```
martyngigg: The `requires_qapp` now sets this up so it shouldn't be needed here.

qt/python/mantidqt/test/test_algorithm_dialog.py
```diff
@@ -0,0 +1,59 @@
...
+from mantidqt.utils.qt.plugins import setup_library_paths
+from mantidqt.utils.qt.testing import gui_test_case
+
+from _widgetscoreqt5 import AlgorithmDialog, GenericDialog, InterfaceManager
```
martyngigg: We shouldn't import from `_widgetscoreqt5` directly as this package will need to support Qt4 & Qt5 later. Please add a thin wrapper module as is done in `messagedisplay.py`

qt/python/mantidqt/test/test_algorithm_dialog.py
```diff
@@ -0,0 +1,59 @@
+import unittest
+
+from qtpy.QtWidgets import QWidget, QLineEdit
+
+from mantid.api import PythonAlgorithm
+from mantid.kernel import Direction, FloatArrayProperty
+from mantid.simpleapi import AlgorithmManager, AlgorithmFactory
+from mantidqt.utils.qt.plugins import setup_library_paths
+from mantidqt.utils.qt.testing import gui_test_case
```
martyngigg: I had to rename `gui_test_case` to remove the `_test_` to be compatible be python's `nose` testing framework in my branch, which I think you have merged in. I've renamed it to `requires_qapp`.

martyngigg (changes requested):

Framework/Kernel/src/DllOpen.cpp
martyngigg: This looks like it could be a merge error?

qt/python/mantidqt/widgets/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+
+set ( COMMON_INC_DIR ../../../widgets/common/inc )
+set ( HEADER_DEPENDS
+  ${COMMON_INC_DIR}/MantidQtWidgets/Common/AlgorithmDialog.h
+  ${COMMON_INC_DIR}/MantidQtWidgets/Common/AlgorithmDialog.h
+  ${COMMON_INC_DIR}/MantidQtWidgets/Common/Message.h
+  ${COMMON_INC_DIR}/MantidQtWidgets/Common/MessageDisplay.h
+)
+
+find_package ( Boost COMPONENTS python REQUIRED )
```
martyngigg: I think the failure of the Python 3 build is due to this - it's find boost linked to the wrong version of Python. See [PythonInterface/CMakeLists.txt(mantidproject/mantid/blob/master/Framework/PythonInterface/CMakeLists.txt#L19) here for what happens in the framework. Perhaps we should add a `FindBoostPython.cmake` file to `buildconfig/CMake` to put this in a common place? It can then be called as `find_package ( BoostPython REQUIRED )`?

qt/python/mantidqt/widgets/src/_widgetscore.sip
```diff
@@ -0,0 +1,98 @@
...
+
+
+class AlgorithmDialog: QDialog {
+%TypeHeaderCode
+#include "MantidQtWidgets/Common/AlgorithmDialog.h"
+#include "MantidAPI/AlgorithmObserver.h"
+#include <boost/python/extract.hpp>
+%End
+public:
+  void setShowKeepOpen(const bool showOption);
```
martyngigg: Do we need to expose all of these methods?

qt/python/mantidqt/test/test_algorithm_dialog.py
mantid-roman: I thought the new workbench was Qt5 only

martyngigg: It will be but the `mantidqt` package will need to be used from other code that still uses Qt4 so I'd like to keep this one Qt4 compatible. I think it will also prevent headaches in the future when moving forward with Qt versions.

qt/python/mantidqt/widgets/src/_widgetscore.sip
mantid-roman: I don't know.

qt/python/mantidqt/utils/test/test_modal_tester.py
martyngigg: Do we need this line anymore?

qt/python/mantidqt/utils/test/test_modal_tester.py
```diff
@@ -0,0 +1,101 @@
...
+            raise RuntimeError('')
+
+        def testing_function(widget):
+            pass
+        with patch('traceback.print_exc') as print_exc:
+            tester = ModalTester(create, testing_function)
+            tester.start()
+            print_exc.assert_called_once_with()
+        self.assertTrue(tester.widget is None)
+        self.assertFalse(tester.passed)
```
martyngigg: I think this file is missing the usual
```
if __name__ == "__main__":
    unittest.main()
```
and has not been added in the CMakeLists.txt?

martyngigg:
@mantid-roman Could you merge in `master`. The PR is showing many more changes than are strictly there and maybe merging in master will help.

martyngigg:
This is looking good. I was able to create a dialog from the selector widget. One thing I noticed is that we are missing the icons from the dialogs so I think we will need a `resources.qrc` for the `mantidqt` library too, much like the [workbench](/mantidproject/mantid/blob/master/qt/applications/workbench/CMakeLists.txt).

mantid-roman:
@martyngigg The linuxes give me this error when `resources.py` is imported:
```
RuntimeError: the PyQt5.QtCore and PyQt4.QtCore modules both wrap the QObject class
```
Do you know how to fix it?

qt/python/resources.qrc
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE RCC><RCC version="1.0">
+<qresource>
+    <file alias="MantidPlot_Icon_32offset.png">../../images/MantidPlot_Icon_32offset.png</file>
+    <file alias="data_replace.png">../../images/data_replace.png</file>
+    <file alias="fileclose.png">../../images/fileclose.png</file>
```
martyngigg: I'm seeing an error saying `Cannot find file: ../../images/fileclose.png`. It looks like this file is in `MantidPlot/icons` and not `images`

qt/python/resources.qrc
```diff
@@ -0,0 +1,10 @@
...
+    <file alias="history.png">../../images/history.png</file>
```
martyngigg: I'm still getting a `QPixmap::scaleHeight: Pixmap is a null pixmap` warning. I think we are missing the `history_off.png` image that is used by https://github.com/mantidproject/mantid/blob/master/qt/widgets/common/src/PropertyWidget.cpp#L372.

martyngigg (changes requested):
Can we add the generated `resources.py` file to `.flake8` excludes too?

This and the missing `history_off.png` are the last things to do I think.

martyngigg (approved):
This looks good now. The list of algorithms is there and you see the expected dialog when requesting one.

PR #21471, opened 2017-12-20
FindPeaksMD performance issue for tube based instruments

### LINKED ISSUE #21334: Peak::findDetector now ~1000X slower
Tested on RHEL7

Using CORELLI
```
cor = LoadEmptyInstrument(InstrumentName='CORELLI')
ws = CreatePeaksWorkspace(InstrumentWorkspace='cor')
p = ws.createPeak([1,1,1])
%timeit p.findDetector()
```
On `3.11.0` this is quick
```
%timeit p.findDetector()
1000 loops, best of 3: 793 µs per loop
```
But on `master`, `3.11.20171201.1727` as tested, this is very slow! About 1000X!
```
%timeit p.findDetector()
1 loops, best of 3: 551 ms per loop
```

Strangely for TOPAZ it's now faster
```
topaz = LoadEmptyInstrument(InstrumentName='TOPAZ')
ws = CreatePeaksWorkspace(InstrumentWorkspace='topaz')
p = ws.createPeak([1,1,1])
%timeit p.findDetector()
```
On `3.11.0`:
```
%timeit p.findDetector()
1000 loops, best of 3: 407 µs per loop
```
On `master`:
```
%timeit p.findDetector()
1000 loops, best of 3: 215 µs per loop
```
I suspect this has to do CORELLI using tubes instead of, as for TOPAZ, rectangular detectors.

This problem has carry on effects, I first observed this using `FindPeaksMD` with CORELLI which was running in about ~25 seconds with `3.11.0` but taking over 25 minutes with nightly.

Is this because of Instrument 2.0 changes as nothing has changed in `Peak`?

@OwenArnold @SimonHeybrock ?

OwenArnold:
Fixes #21334. Gets the performance comparable to what it was previously.

Special handling for tubes and banks of tubes.

**Tester**
- Code Review
- Performance Testing using issue #21334

### DIFF
--- a/Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
+++ b/Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
@@ -55,7 +55,7 @@ private:
   Mantid::Kernel::cow_ptr<std::vector<Eigen::Vector3d>> m_positions;
   Mantid::Kernel::cow_ptr<std::vector<Eigen::Quaterniond>> m_rotations;
   Mantid::Kernel::cow_ptr<std::vector<Eigen::Vector3d>> m_scaleFactors;
-  Mantid::Kernel::cow_ptr<std::vector<ComponentType>> m_componentTypeFlag;
+  Mantid::Kernel::cow_ptr<std::vector<ComponentType>> m_componentType;
   boost::shared_ptr<const std::vector<std::string>> m_names;
 
   const size_t m_size = 0;
@@ -98,7 +98,7 @@ public:
                 boost::shared_ptr<std::vector<Eigen::Vector3d>> positions,
                 boost::shared_ptr<std::vector<Eigen::Quaterniond>> rotations,
                 boost::shared_ptr<std::vector<Eigen::Vector3d>> scaleFactors,
-                boost::shared_ptr<std::vector<ComponentType>> isStructuredBank,
+                boost::shared_ptr<std::vector<ComponentType>> componentType,
                 boost::shared_ptr<const std::vector<std::string>> names,
                 int64_t sourceIndex, int64_t sampleIndex);
   /// Copy assignment not permitted because of the way DetectorInfo stored
@@ -147,7 +147,7 @@ public:
   Eigen::Vector3d scaleFactor(const size_t componentIndex) const;
   void setScaleFactor(const size_t componentIndex,
                       const Eigen::Vector3d &scaleFactor);
-  ComponentType componentFlag(const size_t componentIndex) const;
+  ComponentType componentType(const size_t componentIndex) const;
 
   size_t scanCount(const size_t index) const;
   size_t scanSize() const;
--- a/Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
+++ b/Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
@@ -1,6 +1,7 @@
 #ifndef MANTID_BEAMLINE_COMPONENTINFO_H_
 #define MANTID_BEAMLINE_COMPONENTINFO_H_
 
+#include "MantidBeamline/ComponentType.h"
 #include "MantidBeamline/DllConfig.h"
 #include "MantidKernel/cow_ptr.h"
 #include <Eigen/Geometry>
@@ -54,7 +55,7 @@ private:
   Mantid::Kernel::cow_ptr<std::vector<Eigen::Vector3d>> m_positions;
   Mantid::Kernel::cow_ptr<std::vector<Eigen::Quaterniond>> m_rotations;
   Mantid::Kernel::cow_ptr<std::vector<Eigen::Vector3d>> m_scaleFactors;
-  Mantid::Kernel::cow_ptr<std::vector<bool>> m_isStructuredBank;
+  Mantid::Kernel::cow_ptr<std::vector<ComponentType>> m_componentTypeFlag;
   boost::shared_ptr<const std::vector<std::string>> m_names;
 
   const size_t m_size = 0;
@@ -97,7 +98,7 @@ public:
                 boost::shared_ptr<std::vector<Eigen::Vector3d>> positions,
                 boost::shared_ptr<std::vector<Eigen::Quaterniond>> rotations,
                 boost::shared_ptr<std::vector<Eigen::Vector3d>> scaleFactors,
-                boost::shared_ptr<std::vector<bool>> isStructuredBank,
+                boost::shared_ptr<std::vector<ComponentType>> isStructuredBank,
                 boost::shared_ptr<const std::vector<std::string>> names,
                 int64_t sourceIndex, int64_t sampleIndex);
   /// Copy assignment not permitted because of the way DetectorInfo stored
@@ -146,7 +147,7 @@ public:
   Eigen::Vector3d scaleFactor(const size_t componentIndex) const;
   void setScaleFactor(const size_t componentIndex,
                       const Eigen::Vector3d &scaleFactor);
-  bool isStructuredBank(const size_t componentIndex) const;
+  ComponentType componentFlag(const size_t componentIndex) const;
 
   size_t scanCount(const size_t index) const;
   size_t scanSize() const;
--- a/Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
+++ b/Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
@@ -3,13 +3,13 @@
 
 #include "MantidBeamline/DllConfig.h"
 #include "MantidKernel/cow_ptr.h"
-#include <boost/shared_ptr.hpp>
-#include <boost/iterator/reverse_iterator.hpp>
-#include <vector>
-#include <utility>
-#include <cstddef>
 #include <Eigen/Geometry>
+#include <boost/iterator/reverse_iterator.hpp>
+#include <boost/shared_ptr.hpp>
+#include <cstddef>
 #include <string>
+#include <utility>
+#include <vector>
 
 namespace Mantid {
 namespace Beamline {
@@ -141,7 +141,7 @@ public:
   size_t sample() const;
   size_t root() const;
   double l1() const;
-  std::string name(const size_t componentIndex) const;
+  const std::string &name(const size_t componentIndex) const;
   size_t indexOf(const std::string &name) const;
   Eigen::Vector3d scaleFactor(const size_t componentIndex) const;
   void setScaleFactor(const size_t componentIndex,
--- a/Framework/Beamline/src/ComponentInfo.cpp
+++ b/Framework/Beamline/src/ComponentInfo.cpp
@@ -1,13 +1,13 @@
 #include "MantidBeamline/ComponentInfo.h"
 #include "MantidBeamline/DetectorInfo.h"
 #include "MantidKernel/make_cow.h"
-#include <boost/make_shared.hpp>
-#include <numeric>
 #include <algorithm>
+#include <boost/make_shared.hpp>
+#include <iterator>
+#include <numeric>
+#include <sstream>
 #include <utility>
 #include <vector>
-#include <sstream>
-#include <iterator>
 
 namespace Mantid {
 namespace Beamline {
@@ -23,8 +23,8 @@ void checkScanInterval(const std::pair<int64_t, int64_t> &interval) {
     throw std::runtime_error(
         "ComponentInfo: cannot set scan interval with start >= end");
 }
-}
-}
+} // namespace
+} // namespace
 
 ComponentInfo::ComponentInfo()
     : m_assemblySortedDetectorIndices(
@@ -503,7 +503,7 @@ Eigen::Vector3d ComponentInfo::scaleFactor(const size_t componentIndex) const {
   return (*m_scaleFactors)[componentIndex];
 }
 
-std::string ComponentInfo::name(const size_t componentIndex) const {
+const std::string &ComponentInfo::name(const size_t componentIndex) const {
   return (*m_names)[componentIndex];
 }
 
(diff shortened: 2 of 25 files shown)


Framework/Geometry/src/Instrument/ComponentInfo.cpp
```diff
@@ -239,6 +241,167 @@ double ComponentInfo::solidAngle(const size_t componentIndex,
...
+    std::stack<std::pair<size_t, size_t>> &mutableDetExclusions,
+    IteratorT &mutableIterator) const {
+
+  auto rangeDet = m_componentInfo->detectorRangeInSubtree(index);
+  if (!rangeDet.empty()) {
+    auto startIndex = *rangeDet.begin();
+    auto endIndex = *(rangeDet.end() - 1);
+    mutableBB.grow(componentBoundingBox(startIndex, reference));
+    mutableBB.grow(componentBoundingBox(endIndex, reference));
+    mutableDetExclusions.emplace(std::make_pair(startIndex, endIndex));
```
SimonHeybrock: I am a bit confused (also in the sibling methods, apart from the `byDetectors` variant): We are creating exclusions, but not handling any. Where is the handling happening?

Framework/Geometry/src/Instrument/ComponentInfo.cpp
```diff
@@ -239,6 +241,167 @@ double ComponentInfo::solidAngle(const size_t componentIndex,
...
+ * @param index : Index of the component to get the bounding box for
+ * @param mutableBB : Output bounding box. This will be grown.
+ * @param mutableDetExclusions : Output detector exclusions to append to. These
+ * are ranges of detector indices that we do NOT need to consider for future
+ * bounding box calculations for detectors.
+ * @param detectorExclusions : ranges of detector indices NOT to consider.
+ */
+void ComponentInfo::growBoundingBoxByDetectors(
+    size_t index, const BoundingBox *reference, BoundingBox &mutableBB,
+    std::stack<std::pair<size_t, size_t>> detectorExclusions) const {
```
SimonHeybrock: Have you tried using a `std::vector` instead of `std::stack`? Might be faster. Are there any features relying heavily in `stack`?

Framework/Geometry/src/Instrument/ComponentInfo.cpp
```diff
@@ -239,6 +241,167 @@ double ComponentInfo::solidAngle(const size_t componentIndex,
...
+  auto rangeDet = m_componentInfo->detectorRangeInSubtree(index);
+  auto detIterator = rangeDet.begin();
+  auto *exclusion =
+      detectorExclusions.empty() ? nullptr : &detectorExclusions.top();
+  while (detIterator != rangeDet.end()) {
+
+    // Handle detectors in exclusion ranges
+    if (exclusion && (*detIterator) >= exclusion->first &&
```
SimonHeybrock: Does this mechanism rely on some ordering in the stack? It is not so obvious what is happening here.

Framework/Geometry/src/Instrument/ComponentVisitorHelper.cpp
```diff
@@ -0,0 +1,66 @@
...
+ *behaviour.
+ *
+ * @param visitor : The mutable vistor on which the appropriate register method
+ *will be called
+ * @param visitee : The visitee that will be interrogated
+ * @param nameHint : Naming hint that can be used to determine what type of
+ *Component we have
+ */
+size_t visitAssembly(ComponentVisitor &visitor, const ICompAssembly &visitee,
+                     const std::string &nameHint) {
```
SimonHeybrock: Is the regex matching in this function causing noticeable performance changes anywhere? As far as I understand this now runs for all instruments?

Framework/Geometry/test/ComponentInfoTest.h
```diff
@@ -409,7 +416,7 @@ class ComponentInfoTest : public CxxTest::TestSuite {
     instrument.markAsSource(source);
 
     // A sample
-    ObjComponent *sample = new ObjComponent("some-surface-holder");
+    ObjComponent *sample = new ObjComponent("some-surfbankNameace-holder");
```
SimonHeybrock: clicko

Framework/MDAlgorithms/src/FindPeaksMD.cpp
```diff
@@ -322,6 +325,8 @@ void FindPeaksMD::findPeaks(typename MDEventWorkspace<MDE, nd>::sptr ws) {
   for (uint16_t iexp = 0; iexp < ws->getNumExperimentInfo(); iexp++) {
     ExperimentInfo_sptr ei = ws->getExperimentInfo(iexp);
     this->readExperimentInfo(ei, boost::dynamic_pointer_cast<IMDWorkspace>(ws));
+
+    Geometry::InstrumentRayTracer tracer(inst);
```
SimonHeybrock: This is (one of?) the changes for better performance, caching the ray tracer?

`inst` does not look specific to each loop iteration, why not move outside? Is it being updated by `readExperimentInfo`?

Framework/Geometry/inc/MantidGeometry/Instrument/ComponentInfo.h
```diff
@@ -69,6 +71,27 @@ class MANTID_GEOMETRY_DLL ComponentInfo {
 
   /// Private copy constructor. Do not make public.
   ComponentInfo(const ComponentInfo &other);
+  template <typename IteratorT>
```
SimonHeybrock: I couldn't tell why these are templated, I saw only one use below?

SimonHeybrock (changes requested):
This looks really good overall, the mechanism looks like a good strategy.

Some comments, but mainly questions:

Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
```diff
@@ -141,12 +142,12 @@ class MANTID_BEAMLINE_DLL ComponentInfo {
...
   size_t root() const;
   double l1() const;
-  std::string name(const size_t componentIndex) const;
+  const std::string &name(const size_t componentIndex) const;
   size_t indexOf(const std::string &name) const;
   Eigen::Vector3d scaleFactor(const size_t componentIndex) const;
   void setScaleFactor(const size_t componentIndex,
                       const Eigen::Vector3d &scaleFactor);
-  bool isStructuredBank(const size_t componentIndex) const;
+  ComponentType componentFlag(const size_t componentIndex) const;
```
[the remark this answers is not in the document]
OwenArnold: No good reason. I'll move to `ComponentType` and `componentType`.

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -42,7 +42,7 @@ ComponentInfo::ComponentInfo(
     boost::shared_ptr<std::vector<Eigen::Vector3d>> positions,
     boost::shared_ptr<std::vector<Eigen::Quaterniond>> rotations,
     boost::shared_ptr<std::vector<Eigen::Vector3d>> scaleFactors,
-    boost::shared_ptr<std::vector<bool>> isStructuredBank,
+    boost::shared_ptr<std::vector<ComponentType>> componentTypeFlag,
```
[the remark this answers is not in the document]
OwenArnold: Agree

Framework/Geometry/inc/MantidGeometry/Instrument/ComponentInfo.h
OwenArnold: It's a boost reverse iterator being used. But it's simply being incremented, wanted to hide full type details from future readers.

Framework/Geometry/src/Instrument/ComponentInfo.cpp
```diff
@@ -239,6 +241,167 @@ double ComponentInfo::solidAngle(const size_t componentIndex,
...
+void ComponentInfo::growBoundingBoxAsRectuangularBank(
+    size_t index, const Geometry::BoundingBox *reference,
+    Geometry::BoundingBox &mutableBB,
+    std::stack<std::pair<size_t, size_t>> &mutableDetExclusions,
+    IteratorT &mutableIterator) const {
+  const auto innerRangeComp = m_componentInfo->componentRangeInSubtree(index);
+  const auto innerRangeDet = m_componentInfo->detectorRangeInSubtree(index);
+  auto nSubComponents = innerRangeComp.end() - innerRangeComp.begin() - 1;
+  auto nSubDetectors =
+      std::distance(innerRangeDet.begin(), innerRangeDet.end());
```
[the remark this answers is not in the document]
OwenArnold: Yes exactly. I'll add a comment about that.

Framework/Geometry/src/Instrument/ComponentInfo.cpp
```diff
@@ -239,6 +241,167 @@ double ComponentInfo::solidAngle(const size_t componentIndex,
...
+void ComponentInfo::growBoundingBoxAsBankOfTubes(
+    size_t index, const BoundingBox *reference, BoundingBox &mutableBB,
+    std::stack<std::pair<size_t, size_t>> &mutableDetExclusions,
+    IteratorT &mutableIterator) const {
+  auto innerRangeComp = m_componentInfo->componentRangeInSubtree(index);
+  auto nSubComponents = innerRangeComp.end() - innerRangeComp.begin() - 1;
+  auto innerRangeDet = m_componentInfo->detectorRangeInSubtree(index);
+  auto nSubDetectors =
+      std::distance(innerRangeDet.begin(), innerRangeDet.end());
+  auto nY = nSubDetectors / nSubComponents;
```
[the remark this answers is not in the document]
OwenArnold: Good question. I've not ever seen anything like this, to my knowledge such banks do not exist. I will add a comment about that though.

Framework/Geometry/src/Instrument/ComponentInfo.cpp
```diff
@@ -239,6 +241,167 @@ double ComponentInfo::solidAngle(const size_t componentIndex,
...
+  size_t bottomIndex = *innerRangeDet.begin();
+  size_t lastIndex = *(innerRangeDet.end() - 1);
+  while (bottomIndex < lastIndex) {
+    auto topIndex = bottomIndex + (nY - 1);
+    mutableBB.grow(componentBoundingBox(bottomIndex, reference));
+    mutableBB.grow(componentBoundingBox(topIndex, reference));
```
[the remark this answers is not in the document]
OwenArnold: I'll try to make that refactor.

Framework/Geometry/src/Instrument/ComponentInfo.cpp
OwenArnold: They are passed into `growBoundingBoxByDetectors`. I've put that into it's own method, but the workings are essentially unchanged from before.

Framework/Geometry/src/Instrument/ComponentInfo.cpp
OwenArnold: Yes, but I think I'll try to factor those out, if not in this PR, than in another. The reason is that the iteration over Components (non-detectors) is reverse, as that allows us to do the skipping over sub-components. iteration over Detectors is done forwards, so the stack gives us last exclusion ranges first. As I say  though, this probably needs to be looked at. I'm confident that I have sufficient testing in place to modify this safely now.

Framework/Geometry/src/Instrument/ComponentVisitorHelper.cpp
OwenArnold: This is true, it would run for all instruments, but the regexes only get evaluated for components of type `IComponentAssembly`, of which there should be relatively few in an instrument. In any case, given that the regex is essentially static, I may move to pre-compiling the regex.

Framework/MDAlgorithms/src/FindPeaksMD.cpp
OwenArnold: function `readExperimentInfo` Line 327 actually resets the `inst`. Potentially every experiment info in the multiple experiment info has a different instrument associated.

Framework/Geometry/src/Instrument/ComponentInfo.cpp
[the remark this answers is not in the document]
OwenArnold: Looking at the implementations I think it's probably best left as it is. Unless it is determined that there really are situations where we have banks of tubes where the number of pixels in the tubes is not  the same across the bank. 

The current mechanism should be slightly faster, with only one call to `ComponentInfo::detectorRangeInSubtree` rather than one call per tube.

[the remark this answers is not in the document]
SimonHeybrock: Isn't `detecRangeInSubtree` basically zero cost?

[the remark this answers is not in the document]
OwenArnold: Yes. I'll see if making the change has any effect on overall performance.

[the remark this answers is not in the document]
OwenArnold: @SimonHeybrock unfortunately I can't do this without degrading performance to some extent. The reason is that the current implementation allows us to ignore all detectors that form part of that bank in one go. Trying the alternative method requires us to store and evaluate many more individual ranges of detectors to exclude from detector bounding box calculations. I did however just slightly rewrite the `getBoundingBoxByDetectors` method to avoid deallocations of the map.

SimonHeybrock:
As discussed, a refactor with cleanup will be done as part of a follow-up PR. If this works for @rosswhitfield I have no reason to hold this up now.

SimonHeybrock (approved):

AndreiSavici:
It is still 10 times slower than `3.11`. I get 6.92 ms with this branch and 668 microseconds with `3.11`.

OwenArnold:
@AndreiSavici using Ross's script using `test.md`, `FindPeaksMD` takes ~9 seconds (3.11) and ~17 seconds with changes on this branch, so 2x slower, but not a factor of 10. I'm using Centos7. Nevertheless I'll see if I can get this back to as good or better than before.

rosswhitfield (approved):
👍 Thanks! I'm happy with the performance of FindPeaksMD again.

PR #21472, opened 2017-12-20
Simpler MasterOnly MPI mechanism

SimonHeybrock:
Replaces dummy workspaces by NULL/None workspaces. This seems to make handling `MasterOnly` execution and implementation of such algorithms easier.

There is a minor drawback to this approach: Previously Python scripts could rely on obtained a default-constructed (but uninitialized) workspace on non-master ranks. This allow for, e.g.,

```python
ws = MyAlg()
for i in range ws.getNumberHistograms():
  pass
```
`getNumberHistograms()` returns 0 if the worksapce is not initialized so this piece of code worked. Now `ws` is `None`, so we have to write:

```python
ws = MyAlg()
if ws is not None:
  for i in range ws.getNumberHistograms():
    pass
```
This is slightly more verbose and might necessitate slightly more changes in some reduction scripts. On the other hand it may also catch some actual issues -- code that is actually broken for a parallel workflow -- so this change could maybe also be seen as having a positive effect.

**To test:**

Code review.
No issue, no release notes, since MPI is still internal.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/test/AlgorithmMPITest.h
+++ b/Framework/API/test/AlgorithmMPITest.h
@@ -390,8 +390,6 @@ void run1To1StorageModeTransition(const Parallel::Communicator &comm) {
   auto alg = create<FakeAlg1To1StorageModeTransition<modeOut>>(comm);
   if (modeIn != Parallel::StorageMode::MasterOnly || comm.rank() == 0) {
     alg->setProperty("InputWorkspace", in);
-  } else {
-    alg->setProperty("InputWorkspace", in->cloneEmpty());
   }
   TS_ASSERT_THROWS_NOTHING(alg->execute());
   TS_ASSERT(alg->isExecuted());
--- a/Framework/API/test/AlgorithmMPITest.h
+++ b/Framework/API/test/AlgorithmMPITest.h
@@ -172,11 +172,9 @@ protected:
   }
 };
 
-class FakeAlgNTo1StorageModeFailure : public Algorithm {
+class FakeAlgNTo1 : public Algorithm {
 public:
-  const std::string name() const override {
-    return "FakeAlgNTo1StorageModeFailure";
-  }
+  const std::string name() const override { return "FakeAlgNTo1"; }
   int version() const override { return 1; }
   const std::string category() const override { return ""; }
   const std::string summary() const override { return ""; }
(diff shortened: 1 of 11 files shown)


Framework/API/test/AlgorithmMPITest.h
```diff
@@ -362,79 +337,50 @@ void runNTo0(const Parallel::Communicator &comm) {
...
   TS_ASSERT_THROWS_NOTHING(alg->execute());
   TS_ASSERT(alg->isExecuted());
   Workspace_const_sptr out = alg->getProperty("OutputWorkspace");
-  TS_ASSERT_EQUALS(out->storageMode(), storageMode);
-  TS_ASSERT_EQUALS(out->id(), "FakeWorkspaceA");
+  if (storageMode != Parallel::StorageMode::MasterOnly || comm.rank() == 0) {
+    TS_ASSERT_EQUALS(out->storageMode(), storageMode);
+    TS_ASSERT_EQUALS(out->id(), "FakeWorkspaceA");
+  } else {
+    TS_ASSERT_EQUALS(out, nullptr);
```
OwenArnold: Looks reasonable to me.

OwenArnold (approved):
Changes look reasonable to me. I would suggest a second review from another TSC member before merging.

SimonHeybrock:
I have encountered another potential drawback of this approach: Some code may base decisions on the existance of a workspaces (example: SANS2D reused reduced can run, result is only on master rank, non-master ranks thus re-reduce). Any more complex check on a workspace would fail, so unless the check is *purely* based on the workspace name this drawback does not apply.

I will give this some more thought based on those thoughts.

SimonHeybrock:
Regarding my previous comment: I believe testing for existing workspaces in the ADS is a corner case (which should probably be discouraged anyway) and we should not let this influence the overall design, i.e., this should go ahead as is. Removing the in-progress flag.

martyngigg (approved):
I think this looks sensible. I like that we're able to remove the boilerplate `execNonMaster` implementations - that feels cleaner.

I agree that a non-master rank receiving "null" workspace is a better indication that the previous uninitialized value.

PR #21473, opened 2017-12-20
New loader for LAMP exports

### LINKED ISSUE #21454: A new loader for .hdf files processed by LAMP at ILL.
A new mantid loader is needed to load the processed .hdf files produced by the old data reduction package LAMP at the ILL.
This will be useful in the transition period, and for the validation of mantid routines against the results produced by the old software.

gvardany:
This PR adds a skinny convenience algorithm for loading HDF5 files produced by LAMP (ILL).
It will facilitate the validation of new mantid algorithms against the results produced by the older software LAMP. It will enable hosting the Mantid/LAMP comparisons in Mantid.

**To test:**
Use the algorithm to load the attached file (chop off the .zip extension).
Open it also with e.g. HDFView and make sure the DATA, errors, X and Y axes match with the loaded mantid workspace.
[967076_LAMP.hdf.zip](https://github.com/mantidproject/mantid/files/1575746/967076_LAMP.hdf.zip)

Fixes #21454 . 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/PythonInterface/plugins/algorithms/LoadLamp.py
```diff
@@ -0,0 +1,67 @@
...
+        self.declareProperty(FileProperty(name="InputFile", defaultValue="", action=FileAction.Load, extensions=["hdf"]))
+        self.declareProperty(WorkspaceProperty(name="OutputWorkspace", defaultValue="", direction=Direction.Output))
+
+    def PyExec(self):
+        input_file = self.getProperty("InputFile").value
+        output_ws = self.getPropertyValue("OutputWorkspace")
+
+        with h5py.File(input_file, 'r') as hf:
+            DATA = numpy.array(hf.get('entry1/data1/DATA'), dtype='float')
+            if len(DATA.shape) > 2:
```
josephframsay: It's not immediately obvious what `len(foo.shape)` does - could this be put in a helper such as `def num_dimensions(array): return len(array.shape)` to make this slightly clearer?

Framework/PythonInterface/plugins/algorithms/LoadLamp.py
```diff
@@ -0,0 +1,67 @@
...
+    def PyInit(self):
+        self.declareProperty(FileProperty(name="InputFile", defaultValue="", action=FileAction.Load, extensions=["hdf"]))
+        self.declareProperty(WorkspaceProperty(name="OutputWorkspace", defaultValue="", direction=Direction.Output))
+
+    def PyExec(self):
+        input_file = self.getProperty("InputFile").value
+        output_ws = self.getPropertyValue("OutputWorkspace")
+
+        with h5py.File(input_file, 'r') as hf:
+            DATA = numpy.array(hf.get('entry1/data1/DATA'), dtype='float')
```
josephframsay: Is there any reason for the variable names being in upper case? As far as I can tell, they're not globals? `DATA` is also quite a confusing name - what's the relationship between `DATA`, `Y` and `y_axis`, and could this be made any clearer?

Framework/PythonInterface/plugins/algorithms/LoadLamp.py
```diff
@@ -0,0 +1,67 @@
+from __future__ import (absolute_import, division, print_function)
+from mantid.api import FileProperty, WorkspaceProperty, PythonAlgorithm, NumericAxis, AlgorithmFactory, FileAction
+from mantid.kernel import Direction
+from mantid.simpleapi import CreateWorkspace, AddSampleLogMultiple, mtd
+
+import numpy
+import h5py
+
+
+class LoadLamp(PythonAlgorithm):
```
josephframsay: Could Lamp ever use another data format? Possibly `LoadLampHDF` might be more descriptive?

Framework/PythonInterface/plugins/algorithms/LoadLamp.py
```diff
@@ -0,0 +1,67 @@
...
+        if len(DATA.shape) == 2:
+            nspec = DATA.shape[0]
+            if len(X.shape) == 1:
+                X = numpy.tile(X, nspec)
+            y_axis = NumericAxis.create(nspec)
+            for i in range(nspec):
+                y_axis.setValue(i, Y[i])
+
+        CreateWorkspace(DataX=X, DataY=DATA, DataE=E, NSpec=nspec, OutputWorkspace=output_ws)
+        if len(DATA.shape) == 2:
```
josephframsay: This check has already taken place on line 35: instead of writing in incorrect data and then replacing it, would it not make more sense to assign the right thing to `y_axis` the first time?

Framework/PythonInterface/plugins/algorithms/LoadLamp.py
```diff
@@ -0,0 +1,67 @@
...
+        for log in LOGS.split('\n'):
+            split = log.strip().split('=')
+            if len(split) == 2:
+                name = split[0]
+                value = split[1]
+                if name and value:
+                    log_names.append(name)
+                    log_values.append(value)
+        if log_names:
+            AddSampleLogMultiple(Workspace=output_ws, LogNames=log_names, LogValues=log_values, ParseType=False)
```
josephframsay: Why is `ParseType=False`? Would it not be more compatible with the rest of Mantid to have the values parsed nicely?

Framework/PythonInterface/test/python/plugins/algorithms/LoadLampTest.py
```diff
@@ -0,0 +1,36 @@
...
+
+class LoadLampTest(unittest.TestCase):
+
+    def setUp(self):
+        config.appendDataSearchSubDir('ILL/LAMP/')
+
+    def tearDown(self):
+        mtd.clear()
+
+    def test_1D(self):
```
josephframsay: Could we test a few of the values we read from the file as well, just to be sure that they are being read in correctly? This could also apply to `test_2D_ragged` and `test_2D_tile`

docs/source/algorithms/LoadLamp-v1.rst
```diff
@@ -0,0 +1,32 @@
...
+.. summary::
+
+.. alias::
+
+.. properties::
+
+Description
+-----------
+
+This algorithm loads processed HDF5 files produced by LAMP software at the ILL. Only 1D and 2D data are supported. The output is a point data workspace.
```
josephframsay: Perhaps a small explanation of the structure of a LAMP HDF5 file might be useful thing to have noted down here?

josephframsay (changes requested):
The algorithm works as expected on my machine and the quality of changes is mostly good - I just have a couple of queries about how Y data is handled in `PyExec` (and a few nitpicky comments as well)

Framework/PythonInterface/plugins/algorithms/LoadLamp.py
gvardany: The reason I had in mind, is to make one to one correspondence with the entry name in the HDF file. But it is true that they are not globals, so maybe it is more correct to rename them.  `y_axis` is just a mantid object.

docs/source/algorithms/LoadLamp-v1.rst
gvardany: Yes you are right.

Framework/PythonInterface/plugins/algorithms/LoadLamp.py
gvardany: Lamp can export other formats, but they are ASCII, which we publicly do not support in Mantid. So I think adding HDF5 to the algorithm name is a bit of an overkill. The documentation and summary should make it clear enough.

Framework/PythonInterface/plugins/algorithms/LoadLamp.py
gvardany: 👍

Framework/PythonInterface/test/python/plugins/algorithms/LoadLampTest.py
gvardany: 👍

Framework/PythonInterface/plugins/algorithms/LoadLamp.py
gvardany: This is a good point, for some reason the type parsing goes wrong in python3, I will investigate this further.

Framework/PythonInterface/plugins/algorithms/LoadLamp.py
gvardany: I should have used `ndim` directly.

gvardany:
I hope that's all.

josephframsay:
Yep, everything looks good now!

josephframsay (approved):

PR #21489, opened 2017-12-28
Add buildscript to create UsageData.zip

peterfpeterson:
It has been noticed that the usage data is not updated on sourceforge often enough. This buildscript will be used to fix that issue.

**To test:**

Go into the root of your build tree and run this twice and look at the contents of the archive are reasonable (only works on linux). This is to insure that running it multiple times on a build server works as expected.

*Does not need to be in the release notes* because it is a buildscript.

**Update the [rhel7 clean build](http://builds.mantidproject.org/view/Master%20Pipeline/job/master_clean-rhel7/) once this is merged.**

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/buildconfig/Jenkins/usagedata.sh
+++ b/buildconfig/Jenkins/usagedata.sh
@@ -28,6 +28,6 @@ if [ -f UsageData.zip ]; then
 fi
 
 # create the zip archive
-zip -r UsageData UsageData -x \*hash-stamp -x README\*
+zip -r UsageData UsageData -x \*hash-stamp README\*
 cd -
 mv ExternalData/Testing/Data/UsageData.zip .
--- a/buildconfig/Jenkins/usagedata.sh
+++ b/buildconfig/Jenkins/usagedata.sh
@@ -28,6 +28,6 @@ if [ -f UsageData.zip ]; then
 fi
 
 # create the zip archive
-zip -r UsageData UsageData -x \*hash-stamp README\*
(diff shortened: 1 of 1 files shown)


peterfpeterson:
The RHEL7 build got lost. Since this adds a script that isn't called by anything (yet), I wouldn't expect it to fail anyhow.

buildconfig/Jenkins/usagedata.sh
```diff
@@ -0,0 +1,33 @@
...
+fi
+
+# remove the file if it exists - which shouldn't ever happen
+if [ -f UsageData.zip ]; then
+    echo "removing old version of UsageData.zip"
+    rm UsageData.zip
+fi
+
+# create the zip archive
+zip -r UsageData UsageData -x \*hash-stamp README\*
```
martyngigg: I'm getting the stamp files added to the archive. Should the exclude argument be `-x \*md5-stamp README\*` ?

martyngigg (changes requested):

buildconfig/Jenkins/usagedata.sh
peterfpeterson: Mine end in `\*hash-stamp` so I'll add `\*md5-stamp` rather than replace. I'm guessing it is a cmake version thing.

martyngigg (approved):
This looks good to go to me.

PR #21495, opened 2018-01-03
Fix CTWA and generate workspace

### LINKED ISSUE #21457: Parameters not passed to CTWA
On the `ISIS Reflectometry` interface, there is a bug where no parameters are being passed through from the GUI widgets to the `CreateTransmissionWorkspaceAuto` algorithm, other than the transmission run numbers themselves.

Parameters for this algorithm can be specified via the same input widgets as parameters for the main processing algorithm on the Settings tab. They can also be overridden by entering values on the Runs tab.

### Expected behavior
Parameters specified on the Settings or Runs tab that are applicable to `CreateTransmissionWorkspaceAuto` should get passed through to that function, with the Runs tab taking precedence.

### Actual behavior
No parameters get passed through other than the run numbers

Note that this depends on #21417 to avoid another bug where transmission runs can't be loaded.

- Open `ISIS Reflectometry`
- Enter `13460` under `Run(s)`
- Enter `13463,13464` under `Transmission Run(s)`
- Hit `Process` and plot the spectrum for the output workspace `TRANS_13463+13464`. Note that the max wavelength is 17.
- On the `Settings` tab, enter `14` in `LambdaMax` and re-process. Note that the max wavelength should be 14 but it is unchanged.
- On the `Runs` tab, enter `WavelengthMax=12` in the `Options` column and reprocess. Note that the max wavelength should be 12 but it is unchanged.

### Platforms affected
All

gemmaguest:
***Note: PR #21417 should be merged before this PR***. Until that is merged it might be easier to look at the [comparison with that branch](https://github.com/mantidproject/mantid/compare/21218_refactor_GDPP_reduceRow...21457_fix_CTWA_params).

This PR fixes some bugs on the `ISIS Reflectometry` GUI to do with creating the transmission workspace and generating a python notebook:

- Fix a bug where parameters are not getting passed through to `CreateTransmissionWorkspaceAuto` from the ISIS Reflectometry GUI.
- Fix `GenerateNotebook` so that special characters are allowed in the workspace names. This is required to preserve characters such as `+` which are used in names for workspaces that have been summed.
- Fix `GenerateNotebook` so that all algorithm property values are enclosed in quotes. Previously, the notebook would not run due to some unquoted properties in `ReflectometryReductionOneAuto`.
- Make post-processing optional in `GenerateNotebook` and remove an assertion that was happening when saving a notebook from the `ISIS SANS v2 experimental` GUI.

**To test:**

Check that the bug described in #21457 is fixed.

* Open the `ISIS Reflectometry` GUI and select the `INTER` instrument.
* Expand the group and enter `13460` in the first row. Add another row with `13461+13462`.
* Tick `Output Notebook`
* On the Settings tab, enter:
  * Transmission run(s): `13463,13464`
  * dQ/Q: `0.02`
  * Stitch1DMany: `Params=-0.02`
  * IntegratedMonitors: tick yes
* Back on the Runs tab, select the group and click `Process` and select a location to save the notebook.
* [Launch](https://www.mantidproject.org/Using_IPython_Notebook) IPython Notebook and open the saved notebook. Check that it all runs ok.
* Check each of the plots in the notebook against a plot of the same workspace(s) that were generated in MantidPlot (note they are log-log plots with errors).

Fixes #21457 

**Release Notes** 
*Added to the release notes*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Mantid-Matthew:
This all seems to work the only minor inconsistency I found was that the last graph in the python notebook looked slightly different. In mantid (the bottom image below) the left hand side has a plateau missing from the notebook graph uncertain whether this is a problem or not though.

image

image

gemmaguest:
The difference in the final plot is due to issue #21501 and is not related to this PR. It's because the `dQ/Q` value should override the `Params` but this doesn't work at the moment. I've updated the test instructions to make  `Params=-0.02`, then we get the same results (note that `Params` should be negative for logarithmic binning but it was `+0.01` in my original instructions, which is why the plateau was missing).

Matthew-Andrew (approved):
With the change to params this now works as expected.

PR #21499, opened 2018-01-04
resurrect the old isis refl gui

### LINKED ISSUE #21496: Old ISIS Reflectometry Reduction GUI Won't Start
Trying to start the old ISIS reflectometry GUI produces an error:

```
ImportError: No module named reflectometer
  at line 6 in '<Interface>'
```

It looks like some modules that this interface uses were accidentally deleted in #21145. We need to figure out what should be kept.

NickDraper:
The code for the GUI was incorrectly removed when the SNS refl gui was removed.  This Pr brings the files back.

**To test:**
1. Check that the REFL, REFM and REF SF interfaces are still gone from the interfaces->reflectometry menu, and the underlying code has gone
1. Start and test the ISIS Reflectometry (old) interface.  Before this it would not start.

Fixes #21496

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


scripts/Interface/ui/reflectometer/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+
+# List of UIs to Auto convert
+set( UI_FILES
+refl_columns.ui   
+data_refl_simple.ui
```
gemmaguest: I think data_refl_simple.ui can be removed - it looks like this was part of the ORNL stuff

scripts/Interface/ui/reflectometer/data_refl_simple.ui
```diff
@@ -0,0 +1,4648 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
gemmaguest: As above, this file can be removed. There's also a file `buildconfig/ui_data_refl_simple.ui` which looks like a stray file that should be removed.

gemmaguest (changes requested):
The old interface now starts up ok and both dialogs from the Options menu work, as well as the slit calculator from the Help menu. Import, save, search, transfer, process, stitch and plot all work :+1:

There are just a couple of files left in that should be removed.

gemmaguest (approved):

PR #21509, opened 2018-01-05
Fix regression in GSASIIRefineFitPeaks and enable tests

### LINKED ISSUE #21482: GSASIIRefineFitPeaks crash with zero-size array error
The algorithm GSASIIRefineFitPeaks has been broken by some recent change which went undetected. The script from #20966 no longer produces output, instead it crashes with 
```
RuntimeError: zero-size array to reduction operation minimum which has no identity
  at line 105 in '/home/joe/mantid/Framework/PythonInterface/plugins/algorithms/GSASIIRefineFitPeaks.py'
  caused by line 160 in '/home/joe/mantid/Framework/PythonInterface/plugins/algorithms/GSASIIRefineFitPeaks.py'
  caused by line 945 in '/home/joe/g2conda/GSASII/GSASIIscriptable.py'
  caused by line 709 in '/home/joe/g2conda/GSASII/GSASIIscriptable.py'
```
Somehow the way `GSASIIscriptable` loads in powder histograms has changed. This isn't a change in GSAS-II, as the crash still happens on the revision used to test the original PR.

This ticket should fix this bug, and also come up with some way of making the unit test for GSASIIRefineFitPeaks actually run on the build servers so that this kind of regression can be caught in future.

josephframsay:
Previously, the unit tests for GSASIIRefineFitPeaks did not run on the build servers, as they did not have GSAS-II installed. This PR fixes a bug which, for this reason, went unnoticed.

The unit test was converted to a system test to be run on the nightly server. Scripts for installing GSAS-II automatically were implemented to make this as painless as possible.

**To test:**

Install GSAS-II using one of the new scripts, then make sure the system test is not skipped and passes. Also code review.

Fixes #21482

**Release Notes** 

Doesn't need to be in the release notes, as it's an internal change, and no scientists were aware that the algorithm was unstable in the first place

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Testing/SystemTests/tests/analysis/GSASIIRefineFitPeaksTest.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+class GSASIIRefineFitPeaksRietveldTest(stresstesting.MantidStressTest, _GSASIIRefineFitPeaksTestHelper):
+
+    rwp = None
+    gof = None
+    gsas_proj_path = None
+    _REFERENCE_FILE_NAME = "GSASIIRefineFitPeaksRietveldFitParams.nxs"
+    _GSAS_PROJ_FILE_NAME = "GSASIIRefineFitPeaksRietveldTest.gpx"
+
```
martyngigg: I think we also need to add a `skipTests` method, see [here](https://github.com/mantidproject/mantid/blob/0a677423d0f19c332481ba069fa60b19a7ca0dd4/Testing/SystemTests/tests/analysis/ISIS_PowderPearlTest.py#L53) for an example, that skips the tests if GSAS is not available. Most developers won't have this installed and all of the standard nightly jobs won't have it. Essentially we can probably just pull up the `path_to_gsas` bit below into `skipTests`.

Testing/SystemTests/tests/analysis/GSASIIRefineFitPeaksTest.py
```diff
@@ -0,0 +1,168 @@
...
+
+class GSASIIRefineFitPeaksPawleyTest(stresstesting.MantidStressTest, _GSASIIRefineFitPeaksTestHelper):
+
+    rwp = None
+    gof = None
+    gsas_proj_path = None
+    _REFERENCE_FILE_NAME = "GSASIIRefineFitPeaksPawleyFitParams.nxs"
+    _GSAS_PROJ_FILE_NAME = "GSASIIRefineFitPeaksPawleyTest.gpx"
+
```
martyngigg: Same again with the `skipTests` stuff.

scripts/GSAS-II/install_gsas_vetted.bat
```diff
@@ -0,0 +1,17 @@
+@echo off
+@echo Running install_gsas_proxy.bat
+
+@set DEV_PYTHON_EXE=%~dp0..\..\external\src\ThirdParty\lib\python2.7\python.exe
+@set RELEASE_PYTHON_EXE=%~dp0..\..\bin\python.exe
+
+if exist %DEV_PYTHON_EXE% (
+   @echo Using %DEV_PYTHON_EXE%
+   %DEV_PYTHON_EXE% install_gsas_proxy.py -v 3216
```
martyngigg: Can we define the vetted revision once as a variable at the top of the script?

scripts/GSAS-II/install_gsas_latest.bat
```diff
@@ -0,0 +1,17 @@
...
+   @echo Using %DEV_PYTHON_EXE%
+   %DEV_PYTHON_EXE% install_gsas_proxy.py
+) else (if exist %RELEASE_PYTHON_EXE% (
+  @echo Using %RELEASE_PYTHON_EXE%
+   %RELEASE_PYTHON_EXE% install_gsas_proxy.py
+) else (
+   echo Could not find Mantid Python executable
+)
+)
+pause
```
martyngigg: The script looks good. I can see that the `pause` is necessary so that users can see the script has finished but for automation we'll need to be able to run this in a non-interactive manner. How about adding a flag for the script, something like `--non-interactive` that when passed to the script it will skip the `pause` part?

martyngigg (changes requested):
This is nearly there. Just a few comments inline.

scripts/GSAS-II/install_gsas_proxy.py
```diff
@@ -0,0 +1,101 @@
...
+    response = urllib2.urlopen(url)
+    bootstrap_file = response.read()
+    response.close()
+
+    with open(target_location, "w") as out_file:
+        out_file.write(bootstrap_file)
+
+
+def gsas_install_directory():
+    root_dir = os.path.abspath(os.sep)
```
martyngigg: I just had a thought while looking over this again. Python has the concept of a user site directory where users can install packages and Python automatically knows how to pick them up without any additional configuration of `sys.path`. The module `site` defines the [`USER_SITE`](https://docs.python.org/2/library/site.html#site.USER_SITE)) variable to point at this location.

Perhaps that is a better default for the installation location? The system test would then not need to do any directory searching and `def skipTests` can just check if `import GSASIIscriptable` work without raising an `ImportError`?

josephframsay: I prefer this as a default for the installation location, as this way if a user installs GSAS-II via the script then it's in a convenient place for them to use outside of the algorithm, instead of being squirelled away somewhere weird

martyngigg: I see your point - that's fine. Could we swap it around though so that the python script default location is `site.USER_SITE` and then the install bat script puts it into the location that is better for the users? That should have the same effect?

I'd like the system tests to not have to search for the folder and just test for `import GSASIIscriptable` as it would be better if the system test didn't have hardcoded values like "look 3 levels way etc". It makes the tests harder to run on systems that are not setup like that.

josephframsay: Yup, that sounds like a more sensible way of doing it!

scripts/GSAS-II/install_gsas_vetted.bat
```diff
@@ -0,0 +1,3 @@
+@echo off
+@set VETTED_REVISION_NUMBER=3216
```
martyngigg: Quick question. Do the scripts handle the upgrade when we change this number and do they effectively just update the install or does it have to be wiped out?

josephframsay: It doesn't do any updating - if this number changes we want to just delete the installation and start again

josephframsay:
`install_gsas_proxy.py` now uses `bootstrap.py` to update GSAS-II if it's already installed, instead of needing to wipe out the installation and start again

martyngigg (approved):
Looks good now. Thanks.

PR #21522, opened 2018-01-09
Update WorkspaceTreeWidget to not require shared ptr

### LINKED ISSUE #21442: Add MantidTreeWidget for workspaces to workbench
The mantid workbench requires a workspace widget just as mantid plot has. This is currently the [MantidTreeWidget](https://github.com/mantidproject/mantid/blob/master/qt/widgets/common/inc/MantidQtWidgets/Common/MantidTreeWidget.h).

This will be wrapped using sip to allow python access. However, in order to use it, several steps are required:
* [x] Remove dependency on MantidUI - Create a new MantidDisplayBase (`MantidTreeModel`) which only includes the functionality we require (current implementation is MantidPlot dependant)
* [x] Rewrite `QWorkspaceDockView`  to not be a `DockWidget` but rather a `QWidget`. Let parent class make decisions on docking. #21461 
* [x] `sip` wrapper
* [x] Minor test to ensure construction on workbench side - assume that tests for functionality are on the c++ side

ElliotAOram:
The `WorkspaceTreeWidget` (previously `QWorkspaceDockWidget`) uses `shared_from_this()` when creating it's presenter. The problem with this is this relies upon there being a shared_ptr already created for this object which causes a problem with the workbench (because it's python - no shared_ptrs). 

This covers removing the shared pointer and refactoring code to use a pointer reference instead. As its a reference, not null checks can be removed. Additionally this PR covers removing the `init()` function from `WorkspaceTreeWidget` as this is no longer required as `shared_from_this` is not used. Finally this PR makes the changes in MantidUI to accommodate the new way of creating a `WorkspaceTreeWidget`.

Quick turn around on this issue would be appreciated as it blocks #21442 - hence high priority assigned

**To test:**

* Code review - specifically ensuring pointer safety / memory management is still okay!

* Run MantidPlot and try to break the Workspace Widget

Fixes - no issue number

**Release Notes** 
*Internal change not required in release notes*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidPlot/src/Mantid/MantidUI.cpp
+++ b/MantidPlot/src/Mantid/MantidUI.cpp
@@ -98,8 +98,6 @@
 #include "MantidQtWidgets/SpectrumViewer/SpectrumView.h"
 #include <typeinfo>
 
-#include "MantidQtWidgets/Common/MantidTreeModel.h"
-
 using namespace std;
 
 using namespace Mantid::API;
--- a/MantidPlot/src/Mantid/MantidUI.cpp
+++ b/MantidPlot/src/Mantid/MantidUI.cpp
@@ -98,6 +98,8 @@
 #include "MantidQtWidgets/SpectrumViewer/SpectrumView.h"
 #include <typeinfo>
 
+#include "MantidQtWidgets/Common/MantidTreeModel.h"
+
 using namespace std;
 
 using namespace Mantid::API;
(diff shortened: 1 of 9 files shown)


LamarMoore (approved):
Everything seems to work fine. Could do with a second pair of eyes on the code. Would be nice if we didn't have to use raw pointers somehow. However this may be unavoidable given the intended use case.

martyngigg (approved):

PR #21525, opened 2018-01-10
Replace Instrument 1.0 access with ComponentInfo/DetectorInfo

### LINKED ISSUE #21339: Replace Actors with ComponentInfo
Replace instrument/component actors with the `ComponentInfo` object developed as part of instrument 2.0

LamarMoore:
This PR involves the following work items **NB: A lot of the changes are files which have been removed.**:
* remove gui actors and replace with `ComponentInfo`
* remove references to `Instrument`
* remove references to `IObjComponent` `IComponentAssembly` `ObjComponentAssembly` `ComponentAssembly`

as far as possible use `ComponentInfo` and/or `DetectorInfo` for instrument access.

**To test:**

* Thorough code review
* Load **Instrument View** and ensure a wide selection of instruments are displayed correctly:
  * ensure all detectors are displayed correctly in each view (unwrapped views and FULL3D)
  * ensure picking, masking and selection all work correctly
  * ensure the instrument tree displays the correct detectors (can be verified using pick tab).

Fixes #21339. 

**Release Notes** 
*Does not need to be in the release notes. The performance improvements will be included in the release notes as part of the main feature branch.*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Geometry/src/Instrument/ComponentInfo.cpp
```diff
@@ -107,8 +107,33 @@ ComponentInfo::componentsInSubtree(size_t componentIndex) const {
...
+const std::vector<size_t> &
+ComponentInfo::children(size_t componentIndex) const {
+  return m_componentInfo->children(componentIndex);
+}
+
 size_t ComponentInfo::size() const { return m_componentInfo->size(); }
 
+ComponentInfo::StructuredPanel
+ComponentInfo::structuredPanel(const size_t componentIndex) const {
+  StructuredPanel corners;
```
OwenArnold: This is the same calculation performed for Bounding Boxes for rectangular banks within this class. Can this be re-factored to make the implementation common somehow?

Framework/Geometry/test/InstrumentVisitorTest.h
```diff
@@ -438,6 +437,57 @@ class InstrumentVisitorTest : public CxxTest::TestSuite {
     TS_ASSERT_EQUALS(detScaling, compInfo->scaleFactor(0));
     TS_ASSERT_EQUALS(instrScaling, compInfo->scaleFactor(compInfo->root()));
   }
+
+  void test_instrumentTreeWithMinimalInstrument() {
```
OwenArnold: Good testing

qt/widgets/instrumentview/src/InstrumentTreeModel.cpp
```diff
@@ -184,37 +175,27 @@ QModelIndex InstrumentTreeModel::parent(const QModelIndex &index) const {
...
+      const auto &componentInfo =
+          m_instrWidget->getInstrumentActor().getComponentInfo();
+      auto index = extractIndex(parent);
+      const auto &children = componentInfo.children(index);
+      if (children.size() > 1)
+        return static_cast<int>(children.size());
     }
   } catch (...) {
-    // std::cout<<"Exception: in rowCount"<<'\n';
+    // do nothing
```
OwenArnold: How is this expected to happen?

OwenArnold (changes requested):
A few comments to start with. Id like to re review once closer to completion.

Framework/Geometry/inc/MantidGeometry/Instrument/DetectorInfo.h
```diff
@@ -90,6 +90,8 @@ class MANTID_GEOMETRY_DLL DetectorInfo {
   double twoTheta(const std::pair<size_t, size_t> &index) const;
   double signedTwoTheta(const size_t index) const;
   double signedTwoTheta(const std::pair<size_t, size_t> &index) const;
+  double getPhi(const size_t index) const;
```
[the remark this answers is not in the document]
LamarMoore: Ah yes. Oversight on my part.

Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
```diff
@@ -86,27 +87,30 @@ class MANTID_BEAMLINE_DLL ComponentInfo {
...
+      boost::shared_ptr<std::vector<ComponentType>> componentType,
+      boost::shared_ptr<const std::vector<std::string>> names,
+      int64_t sourceIndex, int64_t sampleIndex);
   /// Copy assignment not permitted because of the way DetectorInfo stored
   ComponentInfo &operator=(const ComponentInfo &other) = delete;
   /// Clone method
   std::unique_ptr<ComponentInfo> cloneWithoutDetectorInfo() const;
   std::vector<size_t> detectorsInSubtree(const size_t componentIndex) const;
   std::vector<size_t> componentsInSubtree(const size_t componentIndex) const;
+  const std::vector<size_t> &children(const size_t componentIndex) const;
```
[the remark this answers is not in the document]
LamarMoore: That's a good idea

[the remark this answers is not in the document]
OwenArnold: Please do same with `componentsInSubtree` + doxygen

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -154,12 +155,13 @@ ComponentInfo::componentsInSubtree(const size_t componentIndex) const {
 const std::vector<size_t> &
 ComponentInfo::children(const size_t componentIndex) const {
   static std::vector<size_t> emptyVec;
-  auto dets = detectorRangeInSubtree(root());
-  auto numDets = static_cast<size_t>(std::distance(dets.begin(), dets.end()));
-  auto index = componentIndex - numDets;
 
-  if (index < m_instrumentTree->size())
-    return (*m_instrumentTree)[index];
+  if (!isDetector(componentIndex)) {
```
OwenArnold: minor point, but might be slightly more readable as 

```cpp
if(isDetector(componentIndex))
  return emptyVec;
...
```

Framework/Beamline/test/ComponentInfoTest.h
```diff
@@ -66,7 +66,7 @@ makeFlatTree(PosVec detPositions, RotVec detRotations) {
 
   std::vector<size_t> children(detPositions.size());
   std::iota(children.begin(), children.end(), 0);
-  auto instrumentTree =
+  auto assemblyImmediateChildren =
```
OwenArnold: I like the `ImmedateChildren` part, but as this is also filled in for non-assemblies (i.e. regular components), maybe we should change the `assembly` part of the name to `component`?

Framework/Geometry/test/ComponentInfoTest.h
```diff
@@ -648,6 +650,33 @@ class ComponentInfoTest : public CxxTest::TestSuite {
...
   }
+
+  void test_StructuredPanel_for_single_rectangular_bank() {
+    auto instrument =
+        ComponentCreationHelper::createTestInstrumentRectangular2(1, 4);
+    auto wrappers = InstrumentVisitor::makeWrappers(*instrument);
+    const auto &componentInfo = std::get<0>(wrappers);
+
+    // find structured panel
+    size_t structuredIndex = 0;
```
OwenArnold: Minor point. Would it be so bad to hard-code the index as  `root - N`? I think that what you have is more flexible, but as this is test code, if the `cretateTestInstrumentRetangular` is changed, these tests would break and that would get fixed. Might save a few lines of code here as part of your search.

mantid-roman (changes requested):
There is a problem with the tube-based instruments: individual detectors are drawn instead of the "outline" or maybe the outline tube is also drawn on top of the detectors. This leads to slow rendering and some colour artefacts. Example of instrument: LARMOR.

Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
```diff
@@ -105,8 +109,9 @@ class MANTID_BEAMLINE_DLL ComponentInfo {
   ComponentInfo &operator=(const ComponentInfo &other) = delete;
   /// Clone method
   std::unique_ptr<ComponentInfo> cloneWithoutDetectorInfo() const;
-  std::vector<size_t> detectorsInSubtree(const size_t componentIndex) const;
+  std::vector<size_t> detectorsInFullSubtree(const size_t componentIndex) const;
```
SimonHeybrock: Why is this renamed? What is a full subtree in comparison to a subtree?

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -51,6 +53,7 @@ ComponentInfo::ComponentInfo(
       m_detectorRanges(std::move(detectorRanges)),
       m_componentRanges(std::move(componentRanges)),
       m_parentIndices(std::move(parentIndices)),
+      m_assemblyImmediateChildren(std::move(assemblyImmediateChildren)),
```
SimonHeybrock: How about `m_children`?

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -139,6 +152,20 @@ ComponentInfo::componentsInSubtree(const size_t componentIndex) const {
   return indices;
 }
 
+const std::vector<size_t> &
+ComponentInfo::children(const size_t componentIndex) const {
+  static std::vector<size_t> emptyVec;
```
SimonHeybrock: `const`?

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -139,6 +152,20 @@ ComponentInfo::componentsInSubtree(const size_t componentIndex) const {
...
+
+  if (!isDetector(componentIndex)) {
+    auto numDets = m_assemblySortedDetectorIndices->size();
+    auto index = componentIndex - numDets;
```
SimonHeybrock: I thought there is a helper method for getting the index? Can't remember the name.

Framework/Geometry/inc/MantidGeometry/Instrument/ComponentInfo.h
```diff
@@ -94,6 +94,15 @@ class MANTID_GEOMETRY_DLL ComponentInfo {
       const std::map<size_t, size_t> &detectorExclusions) const;
 
 public:
+  typedef struct {
```
SimonHeybrock: Why the C syntax and not just `struct StructuredPanel`?

Framework/Geometry/src/Objects/CSGObject.cpp
```diff
@@ -2145,27 +2145,23 @@ size_t CSGObject::numberOfVertices() const {
 /**
 * get vertices
 */
-boost::optional<const std::vector<double> &> CSGObject::getTriangleVertices() const {
-  if (m_handler == nullptr)
```
SimonHeybrock: Is `m_handler` never NULL?

Framework/Geometry/src/Rendering/GeometryHandler.cpp
```diff
@@ -79,17 +79,18 @@ size_t GeometryHandler::numberOfPoints() {
   return 0;
 }
 
-boost::optional<const std::vector<double> &>
-GeometryHandler::getTriangleVertices() {
+const std::vector<double> &GeometryHandler::getTriangleVertices() {
+  static std::vector<double> empty;
```
SimonHeybrock: `const`?

Nice use of local `static` variables for thread-safe defaults here and above btw. 👍

MantidPlot/src/Mantid/MantidUI.cpp
```diff
@@ -2172,9 +2173,14 @@ void MantidUI::showMantidInstrument() {
 }
 
 void MantidUI::showMantidInstrumentSelected() {
+  auto start = std::chrono::system_clock::now();
```
SimonHeybrock: Is this temporary for testing?

SimonHeybrock (changes requested):
A partial review, covering only changes not part of the actual instrument-view code:

Framework/Beamline/src/ComponentInfo.cpp
LamarMoore: I chose this name because I wanted to make the distinction between what is returned by `ComponentInfo::detectorsInSubtree` now `ComponentInfo::detectorsInFullSubtree` and the fact that this only returns children in the next (immediate) tree level. Which could be "Component Assemblies" or leaf nodes.

Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
LamarMoore: This is to reflect the fact that all detectors below this are returned.  For example, doing this on the root will return all detectors in the instrument.

SimonHeybrock: That is implied by the word `tree`, isn't it?

Framework/Beamline/src/ComponentInfo.cpp
SimonHeybrock: Well, anything that is not in the next tree level are grandchildren, or grand-grand children. I believe "tree" and "children" have a commonly understoof and unique meaning?

Framework/Beamline/src/ComponentInfo.cpp
LamarMoore: No this `index` is specific to accessing `m_children`

qt/widgets/instrumentview/inc/MantidQtWidgets/InstrumentView/InstrumentActor.h
```diff
@@ -271,86 +268,30 @@ class EXPORT_OPT_MANTIDQT_INSTRUMENTVIEW InstrumentActor : public GLActor {
...
-  /// Colors in order of workspace indexes
+  /// Colors in order of component info
   mutable std::vector<GLColor> m_colors;
   /// Colour of a masked detector
   GLColor m_maskedColor;
   /// Colour of a "failed" detector
   GLColor m_failedColor;
   /// The collection of actors for the instrument components
-  GLActorCollection m_scene;
+  // GLActorCollection m_scene;
```
mantid-roman: Delete commented out  code?

mantid-roman:
* For non-detector components (choppers) a special colour should be used (`defaultDetectorColor`). See MARI as an example.
* Also a special colour should be used for detectors for which there are no data.

Framework/Geometry/src/Instrument/InstrumentVisitor.cpp
```diff
@@ -203,6 +205,7 @@ InstrumentVisitor::registerGenericComponent(const IComponent &component) {
   // Unless this is the root component this parent is not correct and will be
   // updated later in the register call of the parent.
   m_parentComponentIndices->push_back(componentIndex);
+  m_children->emplace_back(std::vector<size_t>());
```
OwenArnold: Code comment above to the effect that generic components are not assemblies and cannot have children.

Framework/Geometry/src/Objects/CSGObject.cpp
```diff
@@ -2145,27 +2145,29 @@ size_t CSGObject::numberOfVertices() const {
 /**
 * get vertices
 */
-boost::optional<const std::vector<double> &> CSGObject::getTriangleVertices() const {
+const std::vector<double> &CSGObject::getTriangleVertices() const {
+  static std::vector<double> empty;
```
OwenArnold: consider making const too. See other `static` vectors in this file too.

Framework/Geometry/test/ComponentInfoTest.h
```diff
@@ -108,10 +108,11 @@ std::unique_ptr<Beamline::ComponentInfo> makeSingleBeamlineComponentInfo(
   using Mantid::Beamline::ComponentType;
   auto isStructuredBank =
```
OwenArnold: Not your changes but, `isStructured` bank is a bad name for this variable now. `componentType` would be better. Change if you have time.

qt/widgets/instrumentview/inc/MantidQtWidgets/InstrumentView/UnwrappedSurface.h
```diff
@@ -92,9 +59,9 @@ class UnwrappedSurface : public ProjectionSurface {
 
   /** @name Implemented public virtual methods */
   //@{
-  void componentSelected(Mantid::Geometry::ComponentID = nullptr) override;
-  void getSelectedDetectors(QList<int> &dets) override;
-  void getMaskedDetectors(QList<int> &dets) const override;
+  void componentSelected(size_t componentIndex) override;
+  void getSelectedDetectors(std::vector<size_t> &dets) override;
```
OwenArnold: Would be useful to have a comment describing whether the container held ids or indexes. I expect it to be the latter. Consider renaming argument.

qt/widgets/instrumentview/src/InstrumentActor.cpp
```diff
@@ -86,45 +103,53 @@ InstrumentActor::InstrumentActor(const QString &wsName, bool autoscaling,
...
   // If the instrument is empty, maybe only having the sample and source
-  const int nelements = instrument->nelements();
-  if ((nelements == 0) || (nelements == 1 && (instrument->getSource() ||
-                                              instrument->getSample())) ||
-      (nelements == 2 && instrument->getSource() && instrument->getSample())) {
+  auto nelements = componentInfo.size();
+  if ((nelements == 0) || (nelements == 1 && (componentInfo.hasSource() ||
+                                              componentInfo.hasSample())) ||
+      (nelements == 2 && componentInfo.hasSource() &&
+       componentInfo.hasSample())) {
```
OwenArnold: Couldn't it also have no detectors, even if there were a large number of non-detectors? The one thing not checked here is the actual number of detectors.

qt/widgets/instrumentview/src/InstrumentActor.cpp
```diff
@@ -354,101 +385,90 @@ const MantidColorMap &InstrumentActor::getColorMap() const {
...
+  const auto &detInfo = getDetectorInfo();
+  if (pickID < detInfo.size()) {
+    return detInfo.detectorIDs()[pickID];
   }
   return -1;
 }
 
+QList<Mantid::detid_t>
+InstrumentActor::getDetIDs(const std::vector<size_t> &dets) const {
+  QList<Mantid::detid_t> detIDs;
```
OwenArnold: consider preallocate the `QList` to the known size? has a reserve method.

qt/widgets/instrumentview/src/InstrumentActor.cpp
```diff
@@ -615,50 +627,52 @@ void InstrumentActor::sumDetectorsRagged(QList<int> &dets,
...
+    for (const auto &det : specDef) {
+      auto detIndex = det.first;
 
-      if (mask) {
-        masked = mask->isMasked(dets);
-      } else {
-        masked = spectrumInfo.hasDetectors(wi) && spectrumInfo.isMasked(wi);
-      }
+      if (mask)
+        masked = mask->isMasked(detectorIDs[detIndex]);
```
OwenArnold: Idea being that if any detectors in the spectrum are masked, then for colouring they should all be treated as masked?

qt/widgets/instrumentview/src/InstrumentActor.cpp
```diff
@@ -1304,5 +1336,39 @@ void InstrumentActor::loadFromProject(const std::string &lines) {
   }
 }
 
-} // MantidWidgets
-} // MantidQt
+const Mantid::Geometry::ComponentInfo &
+InstrumentActor::getComponentInfo() const {
```
OwenArnold: These are the neutronic versions. We need to add in the physical versions.

qt/widgets/instrumentview/src/InstrumentWidgetPickTab.cpp
```diff
@@ -51,6 +53,17 @@ namespace MantidWidgets {
 
 using namespace boost::math;
 
+namespace {
+double getPhi(const Mantid::Kernel::V3D &pos) {
```
OwenArnold: Comments required. Implies beam along x, y vertical? pos is offset from sample, or just works because sample is at 0,0,0?

OwenArnold: Same bellow with offset.

qt/widgets/instrumentview/src/InstrumentWidgetPickTab.cpp
```diff
@@ -51,6 +53,17 @@ namespace MantidWidgets {
...
+}
+
+double getPhiOffset(const Mantid::Kernel::V3D &pos, const double offset) {
+  double avgPos = getPhi(pos);
+  return avgPos < 0 ? -(offset + avgPos) : offset - avgPos;
+}
+} // namespace
+
 /// to be used in std::transform
 struct Sqrt {
```
OwenArnold: I know you didn't add this, but surely this is solving some pre-lambda problem.

qt/widgets/instrumentview/src/InstrumentWidgetPickTab.cpp
```diff
@@ -1237,27 +1190,25 @@ void DetectorPlotController::plotSingle(int detid) {
...
-  boost::shared_ptr<const Mantid::Geometry::IComponent> parent =
-      det.getParent();
-  Mantid::Geometry::ICompAssembly_const_sptr ass =
-      boost::dynamic_pointer_cast<const Mantid::Geometry::ICompAssembly>(
-          parent);
-  if (parent && ass) {
+  const auto &componentInfo = actor.getComponentInfo();
+
+  if (componentInfo.hasParent(detindex) &&
+      componentInfo.detectorsInSubtree(detindex).size() > 0) {
```
OwenArnold: This can be expensive `detectorsInSubtree` as it will involve a large allocation. And you only want to know the size. Maybe consider adding a method to `ComponentInfo` to get you the number of detectors in the subtree. Can be calculated from the ranges.

qt/widgets/instrumentview/src/InstrumentWidgetRenderTab.cpp
```diff
@@ -141,6 +141,11 @@ InstrumentWidgetRenderTab::InstrumentWidgetRenderTab(
   m_displayAxes->setCheckable(true);
   m_displayAxes->setChecked(true);
   connect(m_displayAxes, SIGNAL(toggled(bool)), this, SLOT(showAxes(bool)));
+  m_volumeRender = new QAction("VolumeRender", this);
```
OwenArnold: Has this accidentally been included. I think this is another issue being dealt with here?

OwenArnold (changes requested):
Overall, very good. Some improvement suggestions here.

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -153,20 +156,22 @@ ComponentInfo::componentsInSubtree(const size_t componentIndex) const {
...
 
   return emptyVec;
 }
 
 size_t ComponentInfo::size() const { return m_size; }
 
+size_t
+ComponentInfo::numberOfDetectorsInSubtree(const size_t componentIndex) const {
+  auto range = detectorRangeInSubtree(componentIndex);
+  return std::distance(range.begin(), range.end());
```
OwenArnold: Looks good.

Framework/Geometry/src/Instrument/ComponentInfo.cpp
```diff
@@ -114,9 +114,19 @@ ComponentInfo::children(size_t componentIndex) const {
...
-ComponentInfo::StructuredPanel
-ComponentInfo::structuredPanel(const size_t componentIndex) const {
-  StructuredPanel corners;
+size_t ComponentInfo::numberOfDetectorsInSubtree(size_t componentIndex) const {
+  return m_componentInfo->numberOfDetectorsInSubtree(componentIndex);
+}
+
+ComponentInfo::QuadrilateralComponent
+ComponentInfo::quadrilateralComponent(const size_t componentIndex) const {
+  auto type = componentType(componentIndex);
```
OwenArnold: Tests added for this?

Framework/Geometry/test/ComponentInfoTest.h
```diff
@@ -649,24 +649,38 @@ class ComponentInfoTest : public CxxTest::TestSuite {
...
-        break;
-      }
-    }
+    // find quadrilateral component
+    size_t structuredIndex = componentInfo->root() - 3;
+    // Does not throw for valid index
+    TS_ASSERT_THROWS_NOTHING(
+        componentInfo->quadrilateralComponent(structuredIndex));
+    // Throws for other non quadrilateral component
+    TS_ASSERT_THROWS(
```
OwenArnold: very minor point, catch as `std::runtime_error&` not `std::runtime_error` some compilers complain about this.

qt/widgets/instrumentview/src/InstrumentActor.cpp
```diff
@@ -165,45 +205,47 @@ void InstrumentActor::setUpWorkspace(
...
-*
-* @param visitor :: Visitor to be accepted bu this actor.
-* @param rule :: A rule defining visitor acceptance by assembly actors.
-*/
-bool InstrumentActor::accept(GLActorVisitor &visitor, VisitorAcceptRule rule) {
-  bool ok = m_scene.accept(visitor, rule);
-  visitor.visit(this);
-  invalidateDisplayLists();
-  return ok;
+void InstrumentActor::setupPhysicalInstrument() {
```
OwenArnold: I would break this up somehow or rename. The usage implies that you are setting up a physical instrument, but actually you do nothing if `isPhysicalView()` does not return true.

qt/widgets/instrumentview/src/InstrumentActor.cpp
```diff
@@ -165,45 +205,47 @@ void InstrumentActor::setUpWorkspace(
...
+  auto sharedWorkspace = getWorkspace();
```
OwenArnold: Shouldn't this shared_ptr and the lock only be set after we know that this is a physical instrument, i.e. move into conditional block?

qt/widgets/instrumentview/src/InstrumentActor.cpp
```diff
@@ -1304,5 +1329,53 @@ void InstrumentActor::loadFromProject(const std::string &lines) {
   }
 }
 
-} // MantidWidgets
-} // MantidQt
+const Mantid::Geometry::ComponentInfo &InstrumentActor::componentInfo() const {
+  return getComponentInfo();
```
OwenArnold: Why are there `getComponentInfo` and `componentInfo` methods?. Surely only `componentInfo` is needed? Also, very important to comment that this will return EITHER the neutronic or physical version. Other similiar APIs (on `MatrixWorkspace` for example) only return the neutronic version.

qt/widgets/instrumentview/src/InstrumentActor.cpp
```diff
@@ -1304,5 +1329,53 @@ void InstrumentActor::loadFromProject(const std::string &lines) {
...
+}
+
+const Mantid::Geometry::DetectorInfo &InstrumentActor::detectorInfo() const {
```
OwenArnold: Same question as above

qt/widgets/instrumentview/src/InstrumentActor.cpp
```diff
@@ -165,45 +205,47 @@ void InstrumentActor::setUpWorkspace(
...
-  invalidateDisplayLists();
-  return ok;
+void InstrumentActor::setupPhysicalInstrument() {
+  auto sharedWorkspace = getWorkspace();
+  Mantid::Kernel::ReadLock _lock(*sharedWorkspace);
+
+  if (isPhysicalView()) {
+    auto instr = sharedWorkspace->getInstrument()->getPhysicalInstrument();
+    if (instr) {
+      m_isPhysicalInstrument = true;
```
OwenArnold: Very minor point, but I would move the setting of this flag to the last statement in the block to give stronger exception guarantees.

qt/widgets/instrumentview/src/InstrumentActor.cpp
```diff
@@ -323,132 +373,112 @@ void InstrumentActor::clearMasks() {
...
-Instrument_const_sptr InstrumentActor::getInstrument() const {
-  Instrument_const_sptr retval;
-
-  // Look to see if we have set the property in the properties file
-  // to define our 'default' view
-  std::string view = Mantid::Kernel::ConfigService::Instance().getString(
-      "instrument.view.geometry");
+std::vector<size_t> InstrumentActor::getMonitors() const { return m_monitors; }
 
+Instrument_const_sptr InstrumentActor::getInstrument() const {
```
OwenArnold: What is this still required for? Presumably there is good reason to have it.

qt/widgets/instrumentview/src/InstrumentActor.cpp
```diff
@@ -1304,5 +1329,53 @@ void InstrumentActor::loadFromProject(const std::string &lines) {
...
+                            static_cast<unsigned char>(qGreen(c)),
+                            static_cast<unsigned char>(qBlue(c)));
+}
+
+const Mantid::Geometry::ComponentInfo &
+InstrumentActor::getComponentInfo() const {
+  if (m_isPhysicalInstrument)
+    return *m_physicalComponentInfo;
+  else
+    return getWorkspace()->componentInfo();
```
OwenArnold: Logic here and below looks reasonable to me.

OwenArnold (changes requested):
A few points to clear up or clarify.

mantid-roman:
Tube picking isn't working. If you switch to "Select whole tube" tool the plot should show the integrated signal of the parent component.

Framework/Beamline/src/ComponentInfo.cpp
```diff
@@ -139,8 +154,26 @@ ComponentInfo::componentsInSubtree(const size_t componentIndex) const {
...
+
+  return emptyVec;
+}
+
 size_t ComponentInfo::size() const { return m_size; }
 
+size_t
+ComponentInfo::numberOfDetectorsInSubtree(const size_t componentIndex) const {
+  if (isDetector(componentIndex))
+    return 0;
```
SimonHeybrock: This is inconsistent, the return of this method should be them same as `detectorsInSubtree().size()`. That being said, isn't it sufficient to use that to avoid bloating the interface?

SimonHeybrock (changes requested):
Small comment on `ComponentInfo` interface:

Framework/Beamline/src/ComponentInfo.cpp
LamarMoore: `detectorsInSubtree()` results in an allocation. I do not want the indices. Just the size. I will look into it.

qt/widgets/instrumentview/src/InstrumentActor.cpp
LamarMoore: see `InstrumentActor::getParameterInfo`

qt/widgets/instrumentview/src/InstrumentWidgetPickTab.cpp
```diff
@@ -1237,50 +1189,53 @@ void DetectorPlotController::plotSingle(int detid) {
...
-  if (parent && ass) {
+  const auto &componentInfo = actor.componentInfo();
+
+  if (!componentInfo.hasParent(detindex)) {
+    m_plot->clearCurve();
+    return;
+  }
+
+  auto parent = componentInfo.parent(detindex);
+  if (componentInfo.numberOfDetectorsInSubtree(parent) > 1) {
```
SimonHeybrock: So this plots only if there are two detectors or more?

Also, is it really relevant for performance to not use `detectorsInSubtree` here? Having a extra method just for this single case would not be justified otherwise.

mantid-roman (approved):

qt/widgets/instrumentview/src/InstrumentActor.cpp
OwenArnold: Ok. Maybe try and strip off the `shared_ptr` bit if you can, otherwise leave it as it is.

SimonHeybrock (approved):

OwenArnold (approved):

OwenArnold:
Merging to long-running feature branch

PR #21526, opened 2018-01-10
No Regex for BoundingBox strategy in ComponentInfo

### LINKED ISSUE #21506: Cleanup and Further Generalisation for BoundingBox in Instrument 2.0
1. See @SimonHeybrock comments from #21334 (now merged), where we should try to remove the `detectorExclusions`. This would give the code a little more clarity if we can possibly achieve it.
2. Registering any `ObjCompAssembly` (has shape), should follow the mechanism introduced for tubes. Currently the regex would also need too match before this strategy was considered. 

I think there are probably two PRs here.

OwenArnold:
Do not use regex matches for banks of tubes or banks. This reduces the amount of code required quite considerably, and only leads to a 5% drop in performance, which should be acceptable.

Fixes #21506. Removing detectorExclusion ranges will be done as a further step.

Changes are internal, no need to update release notes.

**Tester**
Code review, check performance


LamarMoore (approved):
Much more compact and definitely easier to follow.

PR #21536, opened 2018-01-11
FlatPlatePaalmanPings input fixes

### LINKED ISSUE #21520: FlatPlatePaalmanPingsCorrection input validation
There are some unexpected results in the algorithm `FlatPlatePaalmanPingsCorrection` correction after changes made in issue #19959. For example setting the `CanBackThickness` to `0` results in NaNs in the output workspace, although this is a slightly unrealistic use case.

The issue was noticed as this is not set properly in the `Indirect Corrections -> Calculate Paalman Pings -> Container Back Thickness` setting - this should be fixed too.

ianbush:
Changes:
1. Interface `Indirect Corrections -> Calculate Paalman Pings -> Container Back Thickness` has been fixed to set the container back thickness correctly.
1. The algorithm should now deal with cases of no sample, no can back thickness and no can front thickness (and any combinations of these) without generaing `NaNs`. A warning is logged in these cases.

**To test:**

Look at updated system test. Try running these manually to verify that the output workspaces are free of `NaNs`. In the case of no sample and no can the workspace should be all `1`s.

Fixes #21520 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/FlatPlatePaalmanPingsCorrection.py
```diff
@@ -504,83 +552,142 @@ def _flat_abs(self, angle):
...
+                acsc = acc1
+            elif self._has_can_back_in:
+                acsc = acc2 * np.exp(-ki_s - kf_s)
+            else:
+                acsc = 0.5 + 0.5 * np.exp(-ki_s - kf_s)
+        else:
+            acsc = acc
+
+        if self._has_can_front_in and self._has_can_back_in:
+            assc = ass * np.exp(-ki_c1 - kf_c1)
```
brandonhewer: This is the second time np.exp(-ki_c1 - kf_c1) is calculated in this function ('_container_reflection_calc').
Is it unlikely that the factors ki_c1 and kf_c1 contain a large amount of data?

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/FlatPlatePaalmanPingsCorrection.py
```diff
@@ -504,83 +552,142 @@ def _flat_abs(self, angle):
...
+                acsc = 0.5 * np.exp(-kf_s) + 0.5 * np.exp(-ki_s)
+        else:
+            acsc = acc
+
+        if self._has_can_front_in and self._has_can_back_in:
+            assc = ass * np.exp(-ki_c1 - kf_c2)
+        elif self._has_can_front_in:
+            assc = ass * np.exp(-ki_c1)
+        elif self._has_can_back_in:
+            assc = ass * np.exp(-kf_c2)
```
brandonhewer: np.exp(-ki_c1) and np.exp(-kf_c2) are being calculated for the second time in this function, are these factors likely to be small in size?

brandonhewer (commented):
The updated unit tests, when run manually, produce the expected output.

I've included a few comments on repeated calculations, although the calculations are perhaps small, in which case readability is perhaps more important.

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/FlatPlatePaalmanPingsCorrection.py
ianbush: This should not actually get called twice, if `self._has_can_front_in` and  `self._has_can_back_in` then `np.exp(-kf_c2)` is called on line 622, if only `self._has_can_back_in` then it is only called on line 649. I agree though it is a bit hard to follow this, so I re-arranged these methods. I hope they read a bit more clearly now.

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/FlatPlatePaalmanPingsCorrection.py
ianbush: As comment above.

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/FlatPlatePaalmanPingsCorrection.py
```diff
@@ -619,35 +619,34 @@ def _can_thickness_calc(self, can_x_section, can_x_section_efixed, can_thickness
...
         if self._has_can_front_in and self._has_can_back_in:
-            acc = self._can_front_thickness * acc1 * np.exp(-kf_c2)
-            acc += self._can_back_thickness * acc2 * np.exp(-ki_c1)
-            acc /= (self._can_front_thickness + self._can_back_thickness)
+            acc = (self._can_front_thickness * acc1 * np.exp(-kf_c2) + self._can_back_thickness * acc2 * np.exp(-ki_c1)) \
+                  / (self._can_front_thickness + self._can_back_thickness)
+            if self._has_sample_in:
+                acsc = (self._can_front_thickness * acc1 * np.exp(-kf_s - kf_c2) +
+                        self._can_back_thickness * acc2 * np.exp(-ki_c1 - ki_s)) / \
+                       (self._can_front_thickness + self._can_back_thickness)
```
brandonhewer: This is a nice re-arrangement :smile:, it is clear how each term is used in calculating each factor.
There is a further way to prevent recalculating the same values on multiple occasions:

The acc factor appears to be the sum of two product terms. These same product terms are recalculated for the acsc.
i.e.
let `x = self._can_front_thickness * acc1 * np.exp(-kf_c2)` (the first term used to calculate the acc correction factor)
then for the calculation of the ascs factor: 
`self._can_front_thickness * acc1 * np.exp(-kf_s - kf_c2)` can be rewritten as `x * np.exp(-kf_s)`

Similarly for the second term,
let `y = self._can_back_thickness * acc2 * np.exp(-ki_c1)` (the second term used to calculate the acc correction factor)
then in calculation of the ascs factor: 
`self._can_back_thickness * acc2 * np.exp(-ki_s - ki_c1)` can be rewritten as `y * np.exp(-ki_s)`

This could also allow for using pre-calculated `np.exp(-ki_s)` and `np.exp(-kf_s)` terms, which are used later in this function.

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/FlatPlatePaalmanPingsCorrection.py
```diff
@@ -656,35 +655,33 @@ def _container_transmission_calc(self, acc, acc1, acc2, ki_s, kf_s, ki_c1, kf_c2
...
     def _container_reflection_calc(self, acc, acc1, acc2, ki_s, kf_s, ki_c1, kf_c1, ass):
         if self._has_can_front_in and self._has_can_back_in:
-            acc = self._can_front_thickness * acc1
-            acc += self._can_back_thickness * acc2 * np.exp(-ki_c1 - kf_c1)
-            acc /= (self._can_front_thickness + self._can_back_thickness)
+            acc = (self._can_front_thickness * acc1 + self._can_back_thickness * acc2 * np.exp(-ki_c1 - kf_c1)) \
+                  / (self._can_front_thickness + self._can_back_thickness)
+            if self._has_sample_in:
+                acsc = (self._can_front_thickness * acc1 + self._can_back_thickness * acc2 * np.exp(-ki_c1 - ki_s - kf_s - kf_c1)) \
+                       / (self._can_front_thickness + self._can_back_thickness)
```
brandonhewer: Similar to the previous comment, a larger number of operations than necessary would be performed here. The (possible) solution could be the same as the previous one.

brandonhewer (commented):
I've left a few additional comments to outline where there could potentially be a reduction in the number of operations performed on the function inputs.

This may be unnecessary if the numpy arrays are likely to be small, or if the changes could effect readability.

Framework/PythonInterface/plugins/algorithms/WorkflowAlgorithms/FlatPlatePaalmanPingsCorrection.py
ianbush: You are right in saying this could be simplified this way. However, I am not sure the gains are worth while. In either case you have to calculate the exponential of a numpy array, which will be the slow part. Doing the array subtraction such as `-kf_s - kf_c2` should be quick in comparison.

I think I would rather leave it as it is, so that it is more readable, as the code is already a bit hard to follow. As far as I understand the arrays would normally be small, thought perhaps there are exceptions to this.

brandonhewer (approved):
Everything functions as expected and the code is well-structured.

Issue #20441, opened 2017-09-12
Cannot sort peaks workspace by sigma/I column in slice viewer


samueljackson92:
As the name suggests, it does not seem possible to sort by this column in the slice viewer. For other columns it appears to be possible. This column is actually quite a useful one to sort by.

It is probably just a case of enabling it somewhere.

ElliotAOram:
Investigation into this shows that the `Int/SigInt` value in the SliceViewer is created on the fly when the peaks workspace is added (using the overlay peaks functionality).

The sorting for the data in the SliceViewer uses the `SortPeaksWorkspace` algorithm and changes the underlying data in the PeaksWorkspace which is then used to update the table in the sliceViewer. As this is this case, it is not currently possible to sort the data.

A possible solution for this would be to add the additional data (`Int/SigInt`) to the PeaksWorkspace either during integration or when it is added to the SliceViewer. A change such as this should be discussed with other members of the team as (although ideally it shouldn't) changing the number of columns in a PeaksWorkspace could have repercussions else where (e.g. saving/loading etc.).

PR #21556, opened 2018-01-15
Add python extensions for new IPeak methods

samueljackson92:
#20441 added two new methods to `IPeak` but did not export them to python.

**To test:**
 - Load/Create a PeaksWorkspace and then access the peak via the python interface
 - The new methods to be checked are:
   - `getEnergy()`
   - `getIntensityOverSigma()` 

*There's no issue associated with this PR.*

**Release Notes** 
[See here](https://github.com/mantidproject/mantid/pull/21556/commits/aa1fa6c3f92cbcb00906d74cc26686489119ab20)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/DataObjects/src/Peak.cpp
+++ b/Framework/DataObjects/src/Peak.cpp
@@ -714,7 +714,9 @@ double Peak::getFinalEnergy() const { return m_finalEnergy; }
 double Peak::getInitialEnergy() const { return m_initialEnergy; }
 
 /** Get the difference between the initial and final neutron energy */
-double Peak::getEnergyTransfer() const { return getInitialEnergy() - getFinalEnergy(); }
+double Peak::getEnergyTransfer() const {
+  return getInitialEnergy() - getFinalEnergy();
+}
 
 //----------------------------------------------------------------------------------------------
 /** Get the H index of the peak */
--- a/Framework/DataObjects/src/Peak.cpp
+++ b/Framework/DataObjects/src/Peak.cpp
@@ -714,7 +714,7 @@ double Peak::getFinalEnergy() const { return m_finalEnergy; }
 double Peak::getInitialEnergy() const { return m_initialEnergy; }
 
 /** Get the difference between the initial and final neutron energy */
-double Peak::getEnergy() const { return getInitialEnergy() - getFinalEnergy(); }
+double Peak::getEnergyTransfer() const { return getInitialEnergy() - getFinalEnergy(); }
 
 //----------------------------------------------------------------------------------------------
 /** Get the H index of the peak */
--- a/Framework/DataObjects/src/Peak.cpp
+++ b/Framework/DataObjects/src/Peak.cpp
@@ -707,13 +707,13 @@ void Peak::setMonitorCount(double m_monitorCount) {
 }
 
 //----------------------------------------------------------------------------------------------
-/** Get the final neutron energy */
+/** Get the final neutron energy in meV */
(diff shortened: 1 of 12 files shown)


Framework/PythonInterface/mantid/api/src/Exports/IPeak.cpp
```diff
@@ -128,6 +128,8 @@ void export_IPeak() {
            "Get the initial (incident) neutron energy")
       .def("getFinalEnergy", &IPeak::getFinalEnergy, arg("self"),
            "Get the final neutron energy")
+      .def("getEnergy", &IPeak::getEnergy, arg("self"),
```
OwenArnold: Did you consider referring to this as a delta in energy, or is there a good reason why this is just referred to as the energy?

OwenArnold (commented):
Did you consider adding tests to cover the python exports?

Framework/PythonInterface/mantid/api/src/Exports/IPeak.cpp
samueljackson92: I did consider that, but this name was consistent with what the column in the slice viewer was called where the value was originally being calculated. I'm not sure which is better. `Energy` is a bit vague.

samueljackson92:
> Did you consider adding tests to cover the python exports?

It appears that there are no tests at all for `IPeak` python bindings.

Framework/PythonInterface/mantid/api/src/Exports/IPeak.cpp
OwenArnold: I think that @AndreiSavici should provide some advice here. Deprecating parts of the python api is a difficult process, so better to get it right first time if we can.

AndreiSavici: Not sure anyone is using that. But yes, it should be `DeltaE` or `EnergyTransfer`. You might want to change the C++ code as well

samueljackson92: The name in the SliceViewer should also be changed then.

Framework/DataObjects/src/Peak.cpp
```diff
@@ -714,7 +714,9 @@ double Peak::getFinalEnergy() const { return m_finalEnergy; }
 double Peak::getInitialEnergy() const { return m_initialEnergy; }
 
 /** Get the difference between the initial and final neutron energy */
```
OwenArnold: In what units? I think there should be a comment about this. Can do as a follow up change.

Framework/PythonInterface/test/python/mantid/geometry/IPeakTest.py
```diff
@@ -0,0 +1,185 @@
...
+        self.assertEqual(self._peak.getL1(), expected_l1)
+
+    def test_get_l2(self):
+        det_id = 101
+        expected_l2 = 0.2444125610784556
+        self._peak.setDetectorID(det_id)
+        self.assertEqual(self._peak.getL2(), expected_l2)
+
+
+if __name__ == '__main__':
```
OwenArnold: 👍

OwenArnold (approved):

Framework/DataObjects/src/Peak.cpp
samueljackson92: Nah might as well do it here. Initial and final are in meV so should also be in meV.

PR #21571, opened 2018-01-17
Engineering Diffraction GUI - make RunMap its own class

### LINKED ISSUE #21565: Engineering Diffraction GUI RunMap needs to be its own class
`EnggDiffFittingModel` makes extensive use of a typedef called `RunMap`, which allows workspaces and other data associated with fitting to be indexed by run number and bank ID. The same kind of behaviour is needed for the model for the upcoming GSAS tab, so in order to not have to repeat the definition, the typedef and associated helper functions should be removed from `EnggDiffFittingModel` and put into a new class of their own.

Work on this issue could also take care of removing the 'data clump' of run number and bank id.

josephframsay:
The typedef `RunMap`, basically just a container used to index run data by run number and bank ID, was previously only used in `EnggDiffFittingModel`. However it will also be useful for the GSAS fitting tab as well, so it was extracted out into its own class.

**To test:**

Code review, make sure unit tests for `RunMap` and `EnggDiffFittingModel` still pass locally and remotely.

Fixes #21565 

**Release Notes** 
The change is internal, so doesn't need to be in the release notes

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/qt/scientific_interfaces/EnggDiffraction/RunMap.h
+++ b/qt/scientific_interfaces/EnggDiffraction/RunMap.h
@@ -10,6 +10,14 @@ namespace CustomInterfaces {
 template <size_t NumBanks, typename T> class RunMap {
 
 public:
+  void add(const int runNumber, const size_t bank, const T itemToAdd) {
+    if (bank < 1 || bank > NumBanks) {
+      throw std::invalid_argument("Tried to access invalid bank: " +
+                                  std::to_string(bank));
+    }
+    m_map[bank - 1][runNumber] = itemToAdd;
+  }
+
   bool contains(const int runNumber, const size_t bank) const {
     return bank > 0 && bank <= NumBanks &&
            m_map[bank - 1].find(runNumber) != m_map[bank - 1].end();
@@ -26,14 +34,6 @@ public:
                                   std::to_string(bank));
     }
     return m_map[bank - 1].at(runNumber);
-  }
-
-  void add(const int runNumber, const size_t bank, const T itemToAdd) {
-    if (bank < 1 || bank > NumBanks) {
-      throw std::invalid_argument("Tried to access invalid bank: " +
-                                  std::to_string(bank));
-    }
-    m_map[bank - 1][runNumber] = itemToAdd;
   }
 
 private:
--- a/qt/scientific_interfaces/EnggDiffraction/RunMap.h
+++ b/qt/scientific_interfaces/EnggDiffraction/RunMap.h
@@ -21,6 +21,8 @@ public:
 
   std::vector<std::pair<int, size_t>> getRunNumbersAndBankIDs() const;
 
+  size_t size() const;
+
 private:
   std::vector<int> getAllRunNumbers() const;
 
--- a/qt/scientific_interfaces/EnggDiffraction/RunMap.h
+++ b/qt/scientific_interfaces/EnggDiffraction/RunMap.h
@@ -8,19 +8,34 @@
 namespace MantidQt {
 namespace CustomInterfaces {
 
+/**
+Templated container class holding information relating to runs, indexed by
+run number and bank ID.
+*/
 template <size_t NumBanks, typename T> class RunMap {
 
 public:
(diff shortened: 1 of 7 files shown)


qt/scientific_interfaces/EnggDiffraction/RunMap.h
```diff
@@ -0,0 +1,52 @@
...
+template <size_t NumBanks, typename T> class RunMap {
+
+public:
+  /**
+   Add an item to the map
+   @param runNumber The run number associated with the item
+   @param bank The bank ID associated with the item
+   @param itemToAdd The item to add
+   */
+  void add(const int runNumber, const size_t bank, const T itemToAdd);
```
soininen: How large items are going to be added? Would `const T &itemToAdd` make any sense?

qt/scientific_interfaces/EnggDiffraction/RunMap.tpp
```diff
@@ -0,0 +1,94 @@
...
+         m_map[bank - 1].find(runNumber) != m_map[bank - 1].end();
+}
+
+template <size_t NumBanks, typename T>
+const T &RunMap<NumBanks, T>::get(const int runNumber,
+                                  const size_t bank) const {
+  if (bank < 1 || bank > NumBanks) {
+    throw std::invalid_argument("Tried to access invalid bank: " +
+                                std::to_string(bank));
+  }
```
soininen: Would it make sense to refactor this test to a separate function? It is needed here, in `RunMap::add()` and (possibly) `RunMap::remove()`.

qt/scientific_interfaces/EnggDiffraction/RunMap.tpp
```diff
@@ -0,0 +1,94 @@
...
+    throw std::invalid_argument("Tried to access invalid run number " +
+                                std::to_string(runNumber) + " for bank " +
+                                std::to_string(bank));
+  }
+  return m_map[bank - 1].at(runNumber);
+}
+
+template <size_t NumBanks, typename T>
+void RunMap<NumBanks, T>::remove(const int runNumber, const size_t bank) {
+  m_map[bank - 1].erase(runNumber);
```
soininen: Should `bank` validity be tested?

qt/scientific_interfaces/EnggDiffraction/RunMap.tpp
```diff
@@ -0,0 +1,94 @@
...
+template <size_t NumBanks, typename T>
+std::vector<int> RunMap<NumBanks, T>::getAllRunNumbers() const {
+  std::vector<int> runNumbers;
+
+  for (const auto &bank : m_map) {
+    for (const auto &runBankPair : bank) {
+      const auto runNumber = runBankPair.first;
+      if (std::find(runNumbers.begin(), runNumbers.end(), runNumber) ==
+          runNumbers.end()) {
+        insertInOrder(runNumber, runNumbers);
```
soininen: Would using `std::set` make more sense here?

soininen (changes requested):
Generally the code looks really nice and clean to me. I just spotted a few possible minor issues.

qt/scientific_interfaces/EnggDiffraction/RunMap.h
josephframsay: They are almost always shared pointers to workspaces (such as [here](https://github.com/mantidproject/mantid/blob/master/qt/scientific_interfaces/EnggDiffraction/EnggDiffFittingModel.h#L83)). Is it considered better to pass these by value or reference?

soininen: If passed by reference, one has to make sure the reference doesn't outlive the actual pointer. I think a reference would be safe here unless you are calling `add()` in some obscure asynchronous way. The possible performance impact for copying shared pointers is not a big deal, though, so if you feel like staying on the safe side, you can leave the method as-is.

soininen (approved):

PR #21245, opened 2017-11-23
Make space for MeshObject

KarlPalmen:
The geometry objects are rearranged to enable a MeshObject to be defined.

IObject class is defined. Object is renamed CGSObject and made to inherit from IObject.
ObjComponent is modified to use IObject instead of CSGObject.

**To test:**

Code Review and you may check that object can be rendered in instrument view or elsewhere.

Fixes #12827, except that MeshObject still needs to be added. The merger of this change beforehand will help with this.

CSGObject has a performance test. Run it (as the Object performance test) without this fix and with this fix to see whether there is a significant loss in performance.

Are you satisfied with the set of methods of IObject. Could any be removed?

**Release Notes** 
Will be added after MeshObject has been added.

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)?
- [ ] Are the unit tests small and test the class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] Do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant (user and developer) documentation been added/updated?

Does everything look good? Mark the review as **Approve**. A member of `@mantidproject/gatekeepers` will take care of it.

PR #21581, opened 2018-01-18
MeshObject added

### LINKED ISSUE #12827: Extend Shape classes to support triangle based shapes
This is to support the ability to load in scanned images of sample geometries.

This will require inheriting from object or surface class and providing a class that can support all the same methods but using a triangular based mesh as the source data.

KarlPalmen:
This work is to add a triangular mesh object that implements IObject.h

Code review. Check the tests are adequate and pass. 

Fixes #12827. 

**Release Notes** 
When there is a way the user can make use of this mesh object, there can be a release note for it.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,749 @@
...
+* @param observer :: point to measure solid angle from
+* @param scaleFactor :: V3D giving scaling of the object
+* @return :: estimate of solid angle of object.
+*/
+double MeshObject::solidAngle(const Kernel::V3D &observer,
+                              const Kernel::V3D &scaleFactor) const
+
+{
+  MeshObject scaledObject;
+  std::vector<V3D> scaledVertices;
```
brandonhewer: Given the number of elements to be held in the scaledVertices vector, would it be appropriate to reserve that space before pushing to the vector in an iteration, i.e.:
scaledVertices.reserve(m_vertices.size());

brandonhewer (commented):
All tests pass as expected, the new class 'MeshObject' and all new code appears well written. There was only a single case in which it may be appropriate to reserve memory in a vector.

KarlPalmen:
Most recent change has passed all checks.

brandonhewer (approved):
All tests pass and the code appears well-written.

Framework/Geometry/src/Objects/BoundingBox.cpp
```diff
@@ -65,60 +65,60 @@ bool BoundingBox::doesLineIntersect(const V3D &startPoint,
   if (startPoint.X() > xMax()) {
     if (lineDir.X() < -tol) {
       lambda = (xMax() - startPoint.X()) / lineDir.X();
-      if (yMin() < startPoint.Y() + lambda * lineDir.Y() &&
-          yMax() > startPoint.Y() + lambda * lineDir.Y())
-        if (zMin() < startPoint.Z() + lambda * lineDir.Z() &&
-            zMax() > startPoint.Z() + lambda * lineDir.Z())
+      if (yMin() <= startPoint.Y() + lambda * lineDir.Y() &&
```
martyngigg: Can I ask why it was necessary to change this calculation?

Framework/Geometry/inc/MantidGeometry/Rendering/CacheGeometryGenerator.h
```diff
@@ -40,15 +40,19 @@ namespace Geometry {
...
+class MeshObject;
+
 class MANTID_GEOMETRY_DLL CacheGeometryGenerator {
 private:
-  CSGObject *Obj;     ///< Input Object
-  int mNoOfVertices;  ///< number of vertices
-  int mNoOfTriangles; ///< number of triangles
-  double *mPoints;    ///<double array or points
-  int *mFaces;        ///< Integer array of faces
+  CSGObject *csgObj;   ///< Input CSGObject
```
martyngigg: What is the reason that we need to keep 2 pointers here rather than a single `IObject` pointer?

Framework/Geometry/inc/MantidGeometry/Rendering/GeometryHandler.h
```diff
@@ -49,14 +50,17 @@ class MANTID_GEOMETRY_DLL GeometryHandler {
 
 protected:
   IObjComponent *ObjComp; ///< ObjComponent that uses this geometry handler
-  CSGObject *Obj;         ///< Object that uses this geometry handler
+  CSGObject *csgObj;      ///< CSG Object that uses this geometry handler
```
martyngigg: Same question as above.

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,749 @@
+#include "MantidGeometry/Objects/MeshObject.h"
+
+#include "MantidGeometry/Objects/Rules.h"
```
martyngigg: This is a very long list of headers. Do we need them all? For example we include `Kernel/MersenneTwister.h` & `<boost/random/mersenne_twister.hpp>` and I can't see where these are used.

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,749 @@
...
+#include <boost/accumulators/statistics/error_of_mean.hpp>
+#include <boost/accumulators/statistics/stats.hpp>
+#include <boost/make_shared.hpp>
+#include <boost/random/mersenne_twister.hpp>
+#include <boost/random/ranlux.hpp>
+#include <boost/random/uniform_01.hpp>
+
+#include <array>
+#include <deque>
+#include <iostream>
```
martyngigg: Do we need this header? It should generally not be required unless printing to screen and it is quite a heavy header to compile.

martyngigg (changes requested):

Framework/Geometry/src/Objects/BoundingBox.cpp
KarlPalmen: I found it failed when the ray enters the bounding box at one edge and left the bounding box at the opposite edge. Equality is necessary to catch this edge case.

KarlPalmen: I've also added a unit test for this case.

Framework/Geometry/inc/MantidGeometry/Rendering/CacheGeometryGenerator.h
KarlPalmen: We need to keep 2 pointers, because behaviour is different for MeshObject than CSGObject, see the code for `CacheGeometry::Generate() `  (line 44 of cpp file) for example.

Framework/Geometry/inc/MantidGeometry/Rendering/GeometryHandler.h
KarlPalmen: I don't remember why in this case, but in the previous pull request #21245 , I was not able to replace usage CSGObject in the geometry handlers by IObject, either because a concrete instance was required or a CSG specific method was called in one of the geometry handlers.

Framework/Geometry/src/Objects/MeshObject.cpp
KarlPalmen: I've shortened this list of headers considerably and will push this change.

Framework/Geometry/src/Objects/MeshObject.cpp
KarlPalmen: Removed in above-mentioned change.

Framework/Geometry/src/Objects/BoundingBox.cpp
martyngigg: I see. Thanks for the explanation.

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,725 @@
...
+
+MeshObject::~MeshObject() = default;
+
+/*
+ * Initialise with vectors.
+ * @param faces :: Triangles specified by vertex index in anticlockwise order
+ * as seen from outside.
+ * @param vertices :: Vertex positions
+ */
+void MeshObject::initialize(const std::vector<int> &faces,
```
OwenArnold: Do we really need this public method? Why are these arguments not provided to the constructor instead? 

I notice that's there is quite a lot of code protecting for corrupted/incomplete state.

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,725 @@
...
+  // at least 4 triangles and 4 vertices
+  return (numberOfTriangles() >= 4 && numberOfVertices() >= 4);
+}
+
+/**
+* Determines whether Pt is within the object or on the surface
+* @param Pt :: Point to be tested
+* @returns 1 if true and 0 if false
+*/
+bool MeshObject::isValid(const Kernel::V3D &Pt) const {
```
OwenArnold: uppercase `Pt` shouldn't this just be `point`

Framework/Geometry/src/Objects/MeshObject.cpp
OwenArnold: Is there a good reason to use a signed type as the type argument to std::vector<T>? I think this should be unsigned since you are referring to point indices.

Framework/Geometry/inc/MantidGeometry/Objects/MeshObject.h
```diff
@@ -0,0 +1,212 @@
...
+class Track;
+class vtkGeometryCacheReader;
+class vtkGeometryCacheWriter;
+
+/**
+\class MeshObject
+\brief Triangular Mesh Object
+\version 1.0
+\date Dec 2017
+\author S. Ansell
```
OwenArnold: Wrong author

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,725 @@
...
+  type = 0;
+  if (m_handler == nullptr)
+    return;
+  m_handler->GetObjectGeom(type, vectors, myradius, myheight);
+}
+
+/** Getter for the shape xml
+@return the shape xml.
+*/
+std::string MeshObject::getShapeXML() const { return ""; }
```
OwenArnold: If this method can't be erased from `IObject`, the correct behaviour might be for this to throw. I think it's worth reviewing how client code is currently using this. This method does not make sense for this type.

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,725 @@
...
+namespace Geometry {
+
+using Kernel::Material;
+using Kernel::V3D;
+using Kernel::Quat;
+
+/**
+*  Default constuctor
+*/
+MeshObject::MeshObject() : MeshObject("") {}
```
OwenArnold: Unless it is really necessary to have one, I would avoid defining a default constructor for this type.

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,725 @@
...
+
+MeshObject::MeshObject(const std::string &string)
```
OwenArnold: I can't see why we would want this constructor at all, since it isn't actually possible for anything to be "constructed" with the string xml.

Framework/Geometry/test/MeshObjectTest.h
```diff
@@ -0,0 +1,1089 @@
...
+    for (size_t i = 0; i < npoints; ++i) {
+      dummy = lShape->generatePointInObject(rng, maxAttempts);
+    }
+  }
+
+  boost::shared_ptr<MeshObject> createOctahedron() {
+    /**
+    * Create octahedron with vertices on the axes at -1 & +1.
+    */
+    boost::shared_ptr<MeshObject> retVal =
```
OwenArnold: 1. `boost::make_shared`,  not `new MeshObject`
1. but should be a `unique_ptr` anyway as no sharing needed.
1. Description does not match comments for this. Does this really return an Octohedron?
1. Either make this a `const` method or define it as a non-member function

Framework/Geometry/test/MeshObjectTest.h
```diff
@@ -0,0 +1,1089 @@
...
+        boost::shared_ptr<MeshObject>(new MeshObject);
+    retVal->initialize(triangles, vertices);
+    return retVal;
+  }
+};
+
+// -----------------------------------------------------------------------------
+// Performance tests
+// -----------------------------------------------------------------------------
+class MeshObjectTestPerformance : public CxxTest::TestSuite {
```
OwenArnold: 👍 for thinking to include performance tests. Could you please.

1. Ensure that all the performance critical behaviour is covered. such as `isValid`, `isOnSide`, see public methods on `IObject` a guide.
1. Ensure that the test run durations are reasonable in length (~1/10th to 1 second) per test method so that we have a good ability to track performance changes.

Framework/Geometry/test/MeshObjectTest.h
```diff
@@ -0,0 +1,1089 @@
...
+    triangles.push_back(1);
+    triangles.push_back(5);
+
+    boost::shared_ptr<MeshObject> retVal =
+        boost::shared_ptr<MeshObject>(new MeshObject);
+    retVal->initialize(triangles, vertices);
+    return retVal;
+  }
+
+  boost::shared_ptr<MeshObject> createLShape() {
```
OwenArnold: Similar issue as mentioned below. No need to use `shared_ptr` to wrap these objects.

Framework/Geometry/inc/MantidGeometry/Objects/MeshObject.h
```diff
@@ -0,0 +1,212 @@
...
+  /// Assignment operator
+  MeshObject &operator=(const MeshObject &);
+  /// Destructor
+  virtual ~MeshObject();
+  /// Clone
+  IObject *clone() const override {
+    if (m_vertices.size() > 0)
+      return new MeshObject(*this);
+    else {
+      throw("Unintialised MeshObject cannot be copied");
```
OwenArnold: As i mention in the implementation (cpp file) Would be much better if an uninitialized implementation was not possible.

OwenArnold (changes requested):
A great start Karl. I'll re-review after changes made.

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,725 @@
...
+* @returns 1 if true and 0 if false
+*/
+bool MeshObject::isValid(const Kernel::V3D &Pt) const {
+
+  BoundingBox bb = getBoundingBox();
+  if (!bb.isPointInside(Pt)) {
+    return false;
+  }
+
+  V3D direction(0.0, 0.0, 1.0); // direction to look for intersections
```
OwenArnold: I'm confused by this, method. We only check along the z-axis for intersections with the triangles of the mesh?

KarlPalmen: The purpose of this code is to check whether a point is inside. We need only check along one direction for this. I chose upwards for this direction.

OwenArnold: Ok. so the assumption is that the mesh always forms a closed surface.

KarlPalmen: Yes

OwenArnold: Thanks for clarifying @KarlPalmen. I think that this assumption needs to be clearly stated, perhaps in the documentation on the header file.

Framework/Geometry/test/MeshObjectTest.h
KarlPalmen: CSGObjectTest does the same thing with `createCappedCylinder()` and others.

Framework/Geometry/test/MeshObjectTest.h
KarlPalmen: If you think this does not return an octahedron, then exactly what do you think it returns. Note that opposite vertices have indices differing by 3. I'll mention that in the comments.

OwenArnold: This is not creating an Octahedron. This is creating a NULL/Invalid/empty shape of some sorts. This is the implementation of `createOctahdron` in your `MeshObjectPerformanceTest`. You also have one in your `MeshObjectTest` line 885, which is doing the right thing. Maybe it would be best to define a free function for `createOctohedron` that would be visible to tests in both test classes.

OwenArnold (changes requested):

Framework/Geometry/src/Objects/MeshObject.cpp
KarlPalmen: No code calls this.

Framework/Geometry/test/MeshObjectTest.h
KarlPalmen: The current performance test is similar to the test used by CSGObject (was object).

KarlPalmen: I regard the extension of the performance tests beyond that of CSGObject beyond the scope of this issue. The current test covers isValid and intersection.

Framework/Geometry/inc/MantidGeometry/Objects/MeshObject.h
```diff
@@ -34,9 +34,10 @@ class vtkGeometryCacheWriter;
 \brief Triangular Mesh Object
 \version 1.0
 \date Dec 2017
-\author S. Ansell
+\author Karl Palmen
 
-Mesh Object of Traingles
+Mesh Object of Triangles assumed to form one or more
```
OwenArnold: 👍

Framework/Geometry/inc/MantidGeometry/Objects/MeshObject.h
```diff
@@ -160,6 +158,8 @@ class MANTID_GEOMETRY_DLL MeshObject : public IObject {
   void updateGeometryHandler();
 
 private:
+  /// Default constructor - should never be called
+  MeshObject();
```
OwenArnold: Delete it since it should never be called.

Framework/Geometry/inc/MantidGeometry/Objects/MeshObject.h
```diff
@@ -61,10 +62,10 @@ Code Documentation is available at: <http://doxygen.mantidproject.org>
...
-  /// Default constructor
-  MeshObject();
-  /// Constructor from string.
-  MeshObject(const std::string &shapeXML);
+  /// Main constructor
+  MeshObject(const std::vector<size_t> &faces,
+             const std::vector<Mantid::Kernel::V3D> &vertices);
+
   /// Copy constructor
   MeshObject(const MeshObject &);
```
OwenArnold: Where is copy and copy-assignment actually required? I believe that in most cases, `IObject` will be wrapped in a `shared_pointer` (in the production code), and those objects would be immutable. So i'm not sure why we would need the ability to copy. Maybe you could investigate why and comment back. I guess you could delete the method and see what the compiler says when you try to compile mantid.

Framework/Geometry/inc/MantidGeometry/Objects/MeshObject.h
```diff
@@ -80,9 +81,6 @@ class MANTID_GEOMETRY_DLL MeshObject : public IObject {
     }
   }
 
-  void initialize(const std::vector<int> &faces,
```
OwenArnold: 👍

Framework/Geometry/inc/MantidGeometry/Objects/MeshObject.h
```diff
@@ -202,7 +202,7 @@ class MANTID_GEOMETRY_DLL MeshObject : public IObject {
   boost::shared_ptr<vtkGeometryCacheWriter> m_vtk_cache_writer;
 
   /// Contents
-  std::vector<int> m_triangles;
+  std::vector<size_t> m_triangles;
```
OwenArnold: Since you are providing these as constructor arguments. You can make these const now, particularly if copy-assignment is disabled too. i.e. `const std::vector<size_t> m_triangles;

OwenArnold: Could you also add some better comments around this `m_triangles` and `m_vertices` are linked. `m_triangles` indexes into `m_vertices`.

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -22,11 +22,14 @@ using Kernel::Quat;
...
-MeshObject::MeshObject() : MeshObject("") {}
+MeshObject::MeshObject() {} // Should never be called
 
-MeshObject::MeshObject(const std::string &string)
-    : m_boundingBox(), m_object_number(0), m_handler(), m_string(string),
-      m_id(), m_material(), m_triangles(), m_vertices() {}
+MeshObject::MeshObject(const std::vector<size_t> &faces,
+                       const std::vector<V3D> &vertices)
+    : m_boundingBox(), m_object_number(0), m_id(), m_material(),
+      m_triangles(faces), m_vertices(vertices) {
```
OwenArnold: 👍 for your changes....

This is a bit advanced, but as it stands, a `MeshObject` contains a copy of the vertices and faces provided via the constructor arguments. That could be quite expensive. One solution could be to declare both constructor arguments as rvalue references, which allows you to safely`move` (steal) them into the held member vectors. To do this you would have to decalre the Constructor as `MeshObject(const std::vector<size_t> &&faces, std::vector<V3D> &&vertices)`

Come and talk to me if this doesn't make sense to you.

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -101,55 +87,55 @@ bool MeshObject::hasValidShape() const {
 }
 
 /**
-* Determines whether Pt is within the object or on the surface
-* @param Pt :: Point to be tested
+* Determines whether point is within the object or on the surface
+* @param point :: Point to be tested
 * @returns 1 if true and 0 if false
```
OwenArnold: Comment is wrong. Returns True if within surface, does not return an int, so 0, 1 do not make sense in comment.

Framework/Geometry/test/MeshObjectTest.h
```diff
@@ -50,21 +50,238 @@ class MockRNG final : public Mantid::Kernel::PseudoRandomNumberGenerator {
   MOCK_CONST_METHOD0(max, double());
   GCC_DIAG_ON_SUGGEST_OVERRIDE
 };
+
+boost::shared_ptr<MeshObject> createCube(const double size, const V3D &centre) {
```
OwenArnold: 👍 for making these free functions.

The return values should not be wrapped as `boost::shared_ptr` unless there really is sharing happening, and as far as I can tell, there isn't. `std::unique_ptr` would be a much better choice for return value for these methods.

OwenArnold (changes requested):
Thanks for making last round of changes.

Since the solution that you have chosen involves writing the mesh triangulation, and implementing methods by hand, excellent functional and performance testing really is required. Please make sure you consider lots of edge cases in the functional tests, and that the performance tests cover the really important aspects of mesh geometry that need to work quickly.

Framework/Geometry/inc/MantidGeometry/Objects/MeshObject.h
```diff
@@ -0,0 +1,212 @@
...
+  /// material composition
+  std::unique_ptr<Kernel::Material> m_material;
+
+  /// a pointer to a class for reading from the geometry cache
+  boost::shared_ptr<vtkGeometryCacheReader> m_vtk_cache_reader;
+  /// a pointer to a class for writing to the geometry cache
+  boost::shared_ptr<vtkGeometryCacheWriter> m_vtk_cache_writer;
+
+  /// Contents
+  std::vector<size_t> m_triangles;
```
OwenArnold: I wonder if defining this as `std::vector<uint16_t>` would be better. Since I cannot see for a MeshObject (defines a single shape), that 2^64 vertices would ever be used.  @martyngigg thoughts?

martyngigg: I agree that we should use a fixed-size type. Is 65535 enough for say a laser scan of some complex sample? Would `uint32_t` be safer?

OwenArnold: Agreed. `uint32_t` would be safer. I don't see that we would use more than 65K points, but 4G points for a single shape should be easily enough to describe a single object. 

@KarlPalmen could you add a check in the constructor of `MeshObject` as part of these changes along the lines of:

```cpp
if (m_vertices.size() > std::numeric_limits<m_triangles::value_type>::max()) {
  throw std::invalid_argument("More vertices given than possible to triangulate"); // or some better message.
}
```
And add a test for this to `MeshObjectTest`.

Framework/Geometry/inc/MantidGeometry/Objects/MeshObject.h
KarlPalmen: MeshObject is not quite immutable. It material may be changed. Also it is good practice to have a copy constructor so we can control copying rather than leave it to a compiler generated function, likewise with default constructor.

Framework/Geometry/src/Objects/MeshObject.cpp
KarlPalmen: I've come across `&&name` in a course I recently did.

Framework/Geometry/inc/MantidGeometry/Objects/MeshObject.h
KarlPalmen: I made it private to ensure this, else the compiler will create a default constructor that can be called.

Framework/Geometry/inc/MantidGeometry/Objects/MeshObject.h
OwenArnold: You are right, there are complex rules governing when a copy or copy-assignment operator would be provisioned automatically. However, it would be best to disable the copy and copy-assignment operators, not just fail to provide an implementation. i.e.

```cpp
MeshObject(const MeshObject& other) = delete;
MeshObject& operator=(const MeshObject& other) = delete;
``` 

Since you have already provided a non-default constructor the compiler will NOT generate a default constructor for you can safely delete the declaration of the default constructor. I see no reason to allow anyone to create a default-constructed `MeshObject` that has no vertex entries.

Does the material really need to be changed after the lifetime of the `MeshObject` has started? i.e. is a setter for the Material type really necessary? Can this be done via a constructor argument instead?

Framework/Geometry/inc/MantidGeometry/Objects/MeshObject.h
OwenArnold: This is better, but as I just commented elsewhere, you can safely delete it as the compiler is already forbidden from creating one for you.

KarlPalmen: Thanks for letting me know.

Framework/Geometry/inc/MantidGeometry/Objects/MeshObject.h
KarlPalmen: The material does need to be changed after construction, because it is a requirement of `IObject`.

OwenArnold: From a quick search of the code base a pattern I'm seeing a lot is of the form:

```cpp
    auto shape = boost::shared_ptr<IObject>(in_WS->sample().getShape().clone());
    shape->setMaterial(Material("SetInMultipleScattering", neutron, coeff2));
```

i.e. clients of this part of the code are able to provide the material at the same time that the lifetime of the `IObject` derivative starts. In other words, the `IObject` interface may be wrong and you should not be afraid to change it. For example, the above may be better as:

```cpp
    const auto shape = boost::shared_ptr<IObject>(in_WS->sample().getShape().cloneWithMaterial(Material("SetInMultipleScattering", neutron, coeff2)));
```

Notice how the new shape is now also immutable

KarlPalmen: When I investigated this I found cases where an `IObject_sptr` returned by `ComponentCreationHelper` has its material set. Perhaps I could change this `IObject_sptr` back to a shared pointer to `CGSObject` having got rid of most of them.

OwenArnold: Can you point me to where in `ComponentCreationHelper`?

KarlPalmen: Also I've found this done to examples returned by `shapeFactory::createShape`.

KarlPalmen: I found the following functions that return an `IObject_sptr` that subsequently gets its material set.
```
ComponentCreationHelper::createSphere
ComponentCreationHelper::createCappedCylinder
ShapeFactory().createShape with at least 4 argument lists
Container::createSampleShape

```

Framework/Geometry/src/Objects/MeshObject.cpp
KarlPalmen: Would this place any restrictions on the caller of the constructor?

Framework/Geometry/inc/MantidGeometry/Objects/MeshObject.h
OwenArnold: Yes there are quite a few. The point is that you don't have to have a setter in order to achieve this. As I said above. Adding a method to `IObject` such as:

```cpp
std::unique_ptr<IObject> cloneWithMaterial(const Material& material) const;
```

Would allow you to remove the `setMaterial` method.

Framework/Geometry/src/Objects/MeshObject.cpp
OwenArnold: Yes, they would have to pass a temporary or it won't compile. In most cases I think they would want to do this.

Framework/Geometry/inc/MantidGeometry/Objects/MeshObject.h
KarlPalmen: How would this work for the functions I listed, when there is no cloning.

OwenArnold: A simplier approach for many of the cases you listed would simply be to pass the the `Material` as an additional argument to `createSphere`, `createCappedCylinder`, that way you can provide it as part of the constructor to shape, or via the public `setMaterial` method on the derived object at the point that you make it (if you have to).

That means that you are not calling the setter on the `IObject` interface. If this is still causing confusion I can pass by the office tomorrow and talk you through it.

Framework/Geometry/src/Objects/MeshObject.cpp
KarlPalmen: By temporary do you mean something created on the stack (without a new() )?

Framework/Geometry/inc/MantidGeometry/Objects/MeshObject.h
KarlPalmen: I thought it might be something like this.

Framework/Geometry/src/Objects/MeshObject.cpp
KarlPalmen: I've tried it and it was OK in the unit test after I used `std::move` on the constructor arguments, but in `MeshObject` I create a scaled version of the `this` to calculate the solid angle with scale factor and the fails for the triangles and I can't fix it with `std::move` or a temporary copy of the triangles.

OwenArnold: @KarlPalmen I'm a bit tied up this morning. Could you please ask one of the experienced members of the team in the office such as @martyngigg to have a look at this with you?

KarlPalmen: After consulting Martyn, I made two constructors; A flexible constructor that works the old way and an efficient constructor that works the way you suggested.

Framework/Geometry/inc/MantidGeometry/Objects/MeshObject.h
```diff
@@ -202,7 +203,8 @@ class MANTID_GEOMETRY_DLL MeshObject : public IObject {
   boost::shared_ptr<vtkGeometryCacheWriter> m_vtk_cache_writer;
 
   /// Contents
-  std::vector<size_t> m_triangles;
+  /// Triangles are specified by indices into a list of vertices.
+  std::vector<uint16_t> m_triangles;
```
OwenArnold: We discussed on the PR with @martyngigg that this should be `uint32_t`. `unit16_t` is probably sufficient but we agreed on `uint32_t`

As a side point, as that would give >65K addresses for vertexes. Assuming a `V3D` is is 3x8bytes, in the worst case this would give a `MeshObject` the potential size of 1.5GB.  Using `uint32_t`, we get a max size of 96GB in the worst case, per `MeshObject`

martyngigg: @OwenArnold Now you phrase it that way maybe `uint16_t` is sufficient. We probably don't want a 96Gb mesh! At least more will fit in the cache if they are smaller.

Framework/Geometry/inc/MantidGeometry/Objects/MeshObject.h
KarlPalmen: We also need to get rid of `setName` from `IObject.h` .

KarlPalmen: and also `setID`, `setVtkGeometryCacheReader` and `setVtkGeometryCacheWriter`. Every remaining function in `IObject.h`  ends with `const =0`.

KarlPalmen: Another thing to investigate is why 
```
virtual boost::shared_ptr<GeometryHandler> getGeometryHandler() = 0;
```
has no `const` before the ` = 0;`.

Framework/Geometry/test/MeshObjectTest.h
OwenArnold: We need a baseline to start tracking performance from and this means putting the tests in place so that we can begin tracking. Lamar Moore tracks the performance across the framework weekly looking for regressions.  As the implementation of math for `MeshObject` has been provided completely bespoke, I consider the performance tests to be very important.  I don't consider `CSGObjectTestPerformance` as a good example to follow when it comes to complete performance tests.

KarlPalmen: I think a MeshObject with a large number vertices and triangles would be needed for this and am think of an algorithm to generate one.

OwenArnold: A better approach would be to call the method lots of times to build up enough statistics.

KarlPalmen: I was thinking that there might be performance issues that only arise on a larger mesh.

OwenArnold: I don't think that's such a concern. The way your are storing the information looks quite efficient to me. However you have had to implement all aspects of the mesh related math yourself, and that's worth covering with performance tests.

Framework/Geometry/test/MeshObjectTest.h
```diff
@@ -1105,7 +1105,48 @@ class MeshObjectTestPerformance : public CxxTest::TestSuite {
     }
   }
 
+  void test_solid_angle() {
+    const size_t dim(20);
+    V3D test_pt;
+    double dummy;
+    for (size_t i = 0; i < dim*dim*dim; ++i) {
```
OwenArnold: 👍 For adding important performance tests to this performance test class. However, the `create_test_point` calls are going to pollute the performance test timings. We are only interested in the `MeshObject` methods. Can you move the creation of the test points to outside the loops in these test methods so that they have minimal impact, or ideally create the test point in the `MeshObjectTestPerformance` constructor?

KarlPalmen: OK I will deal with that. 

I'll be away from tonight till Tuesday. Also after attempting to rebase and fix a merge conflict my branch has become unpushable and I will be dealing with that when I come back. Then I'll respond to this and any other reviews you make while I am away.

Framework/API/src/Sample.cpp
```diff
@@ -363,7 +363,10 @@ int Sample::loadNexus(::NeXus::File *file, const std::string &group) {
     }
     Kernel::Material material;
     material.loadNexus(file, "material");
-    m_shape->setMaterial(material);
+    // CSGObject expected, if so, set its material
+    if (auto csgObj = boost::dynamic_pointer_cast<Geometry::CSGObject>(m_shape)) {
```
OwenArnold: I was speaking to @martyngigg about this last week, and I think that the main issue is that `ShapeFactory`, should not have been modified to return a `IObject`. `ShapeFactory` should create `CSGObject` only. If you make that change then we can remove these `dynamic_pointer_casts`. That should be very easy to do, and make the code cleaner.

KarlPalmen: This may take a long time, because of the many users of `shapeFactory'.

martyngigg: Do we need to update every usage?

martyngigg: Doing
```
IObject *shape = shape_factory.create(...)
```
will still work as `CSGObject*` can be converted automatically to an `IObject*`. I think it's just the places where we can avoid a dynamic cast?

KarlPalmen: Thanks for the information. That explains why nearly everything builds OK. Only `DetectorEfficiencyCorTest.h` fails.

KarlPalmen: I also find that similar changes are needed in `ComponentCreationHelper`. I'll make those where necessary to remove the dynamic casts and possibly throughout.

KarlPalmen: I find I've already made these changes (only 2 functions needed modifying).

OwenArnold: Once you've made the minor change to the `ShapeFactory` `create` to return the `CSGObject`, you should only need to remove the `dynamic_ptr_casts` you added earlier.

KarlPalmen:
I've done all the removals of the dynamic cast that can be simply removed. More complicated cases such as those in `instrumentDefinitionParser.cpp` remain.

KarlPalmen:
I looked at `InstrumentDefinitionParser` and found this would require changing the `Iobject` in
```
  /// map which maps the type name to a shared pointer to a geometric shape
  std::map<std::string, boost::shared_ptr<Geometry::IObject>>
      mapTypeNameToShape;
```
which I think is a bad idea because it would rule out non-CSG objects from IDFs in future.

Framework/API/src/Sample.cpp
OwenArnold: This dynamic_cast can be removed, you don't need it.

Framework/API/test/SampleTest.h
```diff
@@ -261,7 +261,10 @@ class SampleTest : public CxxTest::TestSuite {
                       Mantid::PhysicalConstants::getNeutronAtom(23, 0), 0.072);
     Sample sample;
     IObject_sptr shape = Mantid::Geometry::ShapeFactory().createShape("");
```
OwenArnold: @martyngigg said to change to `ShapeFactory::createShape` to return CSGObject not IObject types, can you do this so that you can remove the dynamic_cast below.

KarlPalmen:
Another example is `SetSample.cpp`, which would require `workspace->sample()` to return a `CSGObject`.

KarlPalmen:
@OwenArnold  you have replied to out of date code. However, the issue concerning 'Sample.cpp` is still outstanding. It would require the private member variable `m_shape` to be changed to `CSGObject`. I'm not sure we want this with a sample, because it is for samples we want a MeshObject. One could have duplicate pointers, one for `IObject` and one for `CSGObject` both pointing the same way.

OwenArnold:
@KarlPalmen I think you are right about `Sample`. I think leave that as it is.

OwenArnold (approved):
In its current state. I would approve this being merged in to a long-running feature branch. We should avoid merging this into Master this close to the release.

martyngigg:
@OwenArnold I agree.

martyngigg (approved):

martyngigg:
@KarlPalmen There are a few more warnings on Ubuntu to deal with. Once those are taken care of I'll merge this.

PR #21589, opened 2018-01-18
Beamline for Physical Instrument

### LINKED ISSUE #21587: Beamline for Physical Instrument
We need (instrument 2.0/beamline) for the physical instrument as well as the neutronic one.

OwenArnold:
Fixes #21587 by adding const ref `DetectorInfo` and `ComponentInfo` access to new objects created around a physical instrument if present.

**WIP - Do not merge** I'm just making this available for early review.

### DIFF
--- a/Framework/API/inc/MantidAPI/ExperimentInfo.h
+++ b/Framework/API/inc/MantidAPI/ExperimentInfo.h
@@ -163,12 +163,14 @@ public:
                                            const std::string &date = "");
 
   const Geometry::DetectorInfo &detectorInfo() const;
+  const Geometry::DetectorInfo &detectorInfoPhysical() const;
   Geometry::DetectorInfo &mutableDetectorInfo();
 
   const SpectrumInfo &spectrumInfo() const;
   SpectrumInfo &mutableSpectrumInfo();
 
   const Geometry::ComponentInfo &componentInfo() const;
+  const Geometry::ComponentInfo &componentInfoPhysical() const;
   Geometry::ComponentInfo &mutableComponentInfo();
 
   void invalidateSpectrumDefinition(const size_t index);
--- a/Framework/Geometry/inc/MantidGeometry/Instrument.h
+++ b/Framework/Geometry/inc/MantidGeometry/Instrument.h
@@ -212,6 +212,7 @@ public:
   // Methods for use with indirect geometry instruments,
   // where the physical instrument differs from the 'neutronic' one
   boost::shared_ptr<const Instrument> getPhysicalInstrument() const;
+  bool hasPhysicalInstrument() const;
   void setPhysicalInstrument(std::unique_ptr<Instrument>);
 
   void getInstrumentParameters(double &l1, Kernel::V3D &beamline,
--- a/Framework/Geometry/inc/MantidGeometry/Instrument/ParameterMap.h
+++ b/Framework/Geometry/inc/MantidGeometry/Instrument/ParameterMap.h
@@ -339,9 +339,12 @@ public:
 
   bool hasDetectorInfo(const Instrument *instrument) const;
   bool hasComponentInfo(const Instrument *instrument) const;
+  bool hasPhysicalBeamline() const;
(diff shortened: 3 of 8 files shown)


Framework/API/inc/MantidAPI/ExperimentInfo.h
```diff
@@ -163,12 +163,14 @@ class MANTID_API_DLL ExperimentInfo {
                                            const std::string &date = "");
 
   const Geometry::DetectorInfo &detectorInfo() const;
+  const Geometry::DetectorInfo &detectorInfoPhysical() const;
```
SimonHeybrock: Adding two more methods to the already huge `MatrixWorkspace`, is it necessary, given that this is used basically nowhere apart form once in `InstrumentView`?

Framework/Geometry/inc/MantidGeometry/Instrument.h
```diff
@@ -212,6 +212,7 @@ class MANTID_GEOMETRY_DLL Instrument : public CompAssembly {
   // Methods for use with indirect geometry instruments,
   // where the physical instrument differs from the 'neutronic' one
   boost::shared_ptr<const Instrument> getPhysicalInstrument() const;
+  bool hasPhysicalInstrument() const;
```
SimonHeybrock: Why extend the interface of legacy code? Isn't `if(getPhysicalInstrument()` just as good?

Framework/Geometry/inc/MantidGeometry/Instrument/ParameterMap.h
```diff
@@ -380,6 +383,14 @@ class MANTID_GEOMETRY_DLL ParameterMap {
   /// associated with an ExperimentInfo object.
   std::unique_ptr<Geometry::ComponentInfo> m_componentInfo;
 
+  /// Pointer to the DetectorInfo wrapper. NULL unless the instrument is
+  /// associated with the instrument and that has a physical instrument.
+  std::unique_ptr<Geometry::DetectorInfo> m_detectorInfoPhysical;
```
SimonHeybrock: Ok, this brings us to the discussion we had yesterday: now that we have this, how do you make sure `Component::getPos` works correctly on the physical instrument? https://github.com/mantidproject/mantid/blob/master/Framework/Geometry/src/Instrument/Component.cpp#L343.

SimonHeybrock (changes requested):
My main comment/question is regarding the discussion we had yesterday, see below.

I wonder if a simpler option would be to parse the physical instrument on-the-fly when opening the `InstrumentView`?

Framework/Geometry/inc/MantidGeometry/Instrument/ParameterMap.h
OwenArnold: Yes. I see your point. As it stands, you would have to know that the component in question belonged to either the physical or neutronic instrument. The two options I see are.

1. We try to lookup the component (via ID) in the `ComponentInfo` and if it's not there for the neutronic one, we look for it in the physical one.
1. We set a flag on the Instrument to indicate that it is either a physical or neutronic one. We know this at the time that they are created. However, this would looking up the root (instrument object) every time you performed an operation.

Framework/Geometry/inc/MantidGeometry/Instrument.h
OwenArnold: Yes, it's just as good. It is simply a readability aid.

Framework/API/inc/MantidAPI/ExperimentInfo.h
OwenArnold: Good point.

OwenArnold: Though providing this addition does get us closer to the possibility that we don't need to store the Physical Instrument at all. Creating the `Instrument 2` bits in a lazy way at the point of use, means keeping the Physical Instrument.

OwenArnold:
This will be done directly in the InstrumentView, and I will help @LamarMoore with that.

PR #21595, opened 2018-01-19
Speed up SumOverlappingTubes

### LINKED ISSUE #21512: Speed up SumOverlappingTubes
Look at improving performance of `SumOverlappingTubes`. A performance test should be added too.

ianbush:
This speeds up the algorithm `SumOverlappingTubes` by removing a debugging statement and paralleling a couple of loops.

**To test:**

1. Compare the performance test speed for 6a5455a and d778d7c and check there is *big* a speed up.
1. Compare the performance test speed for 2a4eb13 and 4694f45 and check there is a *small* speed up.

Fixes #21512. 

**Release Notes** 
*Does not need to be in the release notes - changes to a recently introduced algorithm.*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


ianbush:
@OwenArnold Changes to use `ComponentInfo` in `SumOverlappingTubes` made here.

Framework/Algorithms/src/SumOverlappingTubes.cpp
```diff
@@ -212,21 +214,17 @@ void SumOverlappingTubes::getHeightAxis(const std::string &componentName) {
...
-    compAss.getChildren(children, false);
-    for (const auto &child : children) {
-      const auto posY = child->getPos().Y();
+    const auto &componentInfo = m_workspaceList.front()->componentInfo();
+    const auto &detectorInfo = m_workspaceList.front()->detectorInfo();
+    const auto componentIndex = componentInfo.indexOf(componentName);
+    const auto &detsInSubtree =
+        componentInfo.detectorsInSubtree(componentIndex);
+    for (const auto detIndex : detsInSubtree) {
+      const auto posY = detectorInfo.position({detIndex, 0}).Y();
```
OwenArnold: You can make this call directly through `componentInfo`
```cpp
const auto posY = componentInfo.position({detIndex, 0}).Y()
```

ianbush: Is this possible because `detectorIndex == componentIndex`?

OwenArnold: Yes. A detector index is guaranteed to be the same a component index http://docs.mantidproject.org/nightly/concepts/InstrumentAccessLayers.html#access-to-components-and-working-with-detectors

ianbush:
@OwenArnold This should be ready for review now, the `ComponentInfo` changes are included here.

Framework/Algorithms/src/SumOverlappingTubes.cpp
```diff
@@ -271,11 +267,17 @@ SumOverlappingTubes::performBinning(MatrixWorkspace_sptr &outputWS) {
   std::vector<std::vector<double>> normalisation(
       m_numHistograms, std::vector<double>(m_numPoints, 0.0));
 
+  std::vector<MatrixWorkspace_sptr> workspaceVector{std::begin(m_workspaceList),
```
OwenArnold: Should the `vector` and the type argument for the vector both be `const`?

Framework/Algorithms/test/SumOverlappingTubesTest.h
```diff
@@ -25,6 +26,41 @@ using namespace Mantid::Indexing;
...
+
+  Points x(nBins, LinearGenerator(0.0, 1.0));
+  Counts y(std::vector<double>(nBins, 2.0));
+  builder.setHistogram(Histogram(x, y));
+
+  auto testWS = builder.buildWorkspace();
+
+  // This has to be added to the ADS so that it can be used with the string
+  // validator used in the algorithm.
+  AnalysisDataService::Instance().add(name, testWS);
```
OwenArnold: It's a real shame you have to do this, and I can see the `ADSValidator` on the input workspaces. @LamarMoore was working on a `WorkspaceListProperty` to try to rectify this exact situation. I don't recall what the status of that is now.

OwenArnold (changes requested):
Minor change request and a question. Also, if this algorithms was part of any previous release, I think that you should add a note about performance to the release notes.

Seems you have this about 10x faster than before from my performance test runs.

Framework/Algorithms/test/SumOverlappingTubesTest.h
ianbush: I just checked with him, unfortunately it is not ready yet, but something to bear in mind for future tidy-up.

OwenArnold: Of course, this would have to be done carefully. Once it makes a release, we have to assume that users have scripts that use the current property mechanism.

ianbush: I was expecting a `WorkspaceListProperty` would look the same to the user as the current mechanism in Python, perhaps this would not be the case though?

ianbush:
I have removed the parallel loops completely, with the way the errors are calculated the loops cause a race condition. I might revisit this to see if it can be worked around at some point, but in any case the biggest speedup is from removing the debugging.

OwenArnold (approved):

PR #21601, opened 2018-01-19
Remove unnecessary copy in SaveEventNexus::appendEventListData.

quantumsteve:
Description of work.

The vector of events is being passed by value when it could be passed by a const reference. Also several checks can be removed from the hot loop.

**To test:**

1. Code review
2. Do we have a test or user script to test for performance?

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/DataHandling/src/SaveNexusProcessed.cpp
+++ b/Framework/DataHandling/src/SaveNexusProcessed.cpp
@@ -331,8 +331,8 @@ void SaveNexusProcessed::appendEventListData(const std::vector<T> &events,
   if (events.empty())
     return;
 
-  auto it = events.cbegin();
-  auto it_end = events.cend();
+  const auto it = events.cbegin();
+  const auto it_end = events.cend();
 
   // Fill the C-arrays with the fields from all the events, as requested.
   if (tofs) {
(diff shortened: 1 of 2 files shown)


Framework/DataHandling/src/SaveNexusProcessed.cpp
```diff
@@ -331,21 +331,29 @@ void SaveNexusProcessed::appendEventListData(std::vector<T> events,
   if (events.empty())
     return;
 
-  typename std::vector<T>::const_iterator it;
-  typename std::vector<T>::const_iterator it_end = events.end();
-  size_t i = offset;
+  auto it = events.cbegin();
```
peterfpeterson: Should this be `const auto`? I'm suspicious that the compiler will figure it out from the `cbegin()`.

quantumsteve: This shouldn't matter (`events` is const and `.cbegin()` and `.cend()` returns a const_interator), but never hurts to be explicit.

peterfpeterson (approved):

peterfpeterson (approved):

peterfpeterson (approved):

PR #21604, opened 2018-01-21
Crystalfield bugfixes

### LINKED ISSUE #20990: CrystalField and CrystalFieldMulti bugs
This is a general issue for bugs related to the new implementation of `CrystalField` and `CrystalFieldMulti` to handle multiple rare-earth sites in a single compound. The workflow example scripts (`yb2ti2o7.py`, `ndos2al10.py` and `prrh3.py`) in the attached zip should work without errors and provide a good fit to the data. At the moment, they fail with various errors; most are related to the Python interface to the new C++ code.

[crystalfield_tests.zip](https://github.com/mantidproject/mantid/files/1404347/crystalfield_tests.zip)

mducle:
This is a collection of bugfixes for the CrystalField calculations routines since the back-end was rewritten last autumn. Due to the changes in the back-end some workflows (which don't have system tests) became broken, and this was not picked up in the unit tests.

The major changes are in the handling of peak widths (setting a resolution model disables peak with fitting and invalidates the `.FWHM` property; setting `.FWHM` invalidates a resolution model; previously the behaviour when both are set was ambiguous and in some cases wrong (e.g. fixed FWHM at starting value given in `.FWHM`)), and in the multi-site routines (particularly handling of background).

In addition, due to requests from users, two additional features were added:

1. The ability to calculate the dipole transition matrix
2. The ability to add a residual (background) susceptibility (`chi0`) to the susceptibility calculations.

The first feature is purely Python, the second involved some C++ changes.

Most of the bugs were in the Python interface (e.g. the arbitrary `J` calculations was mistakenly removed in the Python interface but the C++ remained), but there were some C++ changes (particularly handling the peak widths) needed.

**To test:**

Code review and check automated tests run ok.

Test scripts and related datafiles are in the following zip:
[cf_test_scripts.zip](https://github.com/mantidproject/mantid/files/1744795/cf_test_scripts.zip)
Unzip and add the `data` folder to the Mantid data path and `reduction_scripts` to the Mantid scripts path and run. 

A new system test based on the `cf_doc.py` file has been added to check that the python syntax works (it runs some fits but the results are not checked - the unit tests are used for checked the numerical validity of the computations).

Fixes #20990

**Release Notes** 

Release notes included.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/CurveFitting/src/Functions/CrystalFieldControl.cpp
```diff
@@ -55,10 +55,31 @@ void CrystalFieldControl::setAttribute(const std::string &name,
...
       buildControls();
     } else if (name == "FWHMs") {
+      const size_t nSpec = m_temperatures.size();
       m_FWHMs = attr.asVector();
-      if (m_FWHMs.size() == 1 && m_FWHMs.size() != m_temperatures.size()) {
-        m_FWHMs.assign(m_temperatures.size(), m_FWHMs.front());
+      if (m_FWHMs.size() == 1 && m_FWHMs.size() != nSpec) {
+        m_FWHMs.assign(nSpec, m_FWHMs.front());
+      }
+      if (attr.asVector().size() > 0) {
```
mantid-roman: Could you use `empty()`? `attr.asVector()` creates a new copy of the vector. Can you use `m_FWHMs` instead?

Framework/CurveFitting/src/Functions/CrystalFieldControl.cpp
```diff
@@ -55,10 +55,31 @@ void CrystalFieldControl::setAttribute(const std::string &name,
...
+            control.setAttribute("FWHMY", Attribute(std::vector<double>()));
+          } else {
+            setAttribute("FWHMX", Attribute(std::vector<double>()));
+            setAttribute("FWHMY", Attribute(std::vector<double>()));
+          }
+        }
       }
+    } else if ((name.compare(0, 5, "FWHMX") == 0 ||
+                name.compare(0, 5, "FWHMY") == 0) &&
+               attr.asVector().size() > 0) {
```
mantid-roman: The same as above

mantid-roman (changes requested):
`CrystalFieldControl` has field `std::vector<int> m_physprops` which I think is never used. Can it be deleted?

Framework/CurveFitting/src/Functions/CrystalFieldControl.cpp
mducle: @mantid-roman I can't find a way to avoid using `asVector` in this case. `attr.isEmpty()` implicitly assumes the attribute is a string, and so raises an exception in this case (where the attribute is a vector), and there's no other public method to access the contents of the attribute except to using the `as*` methods...

mantid-roman: Could you then use `attr.asVector().empty()` ?

mducle: Yes. I'll do that.

scripts/Inelastic/CrystalField/fitting.py
```diff
@@ -261,6 +314,10 @@ def makePhysicalPropertiesFunction(self, i=0):
         if len(fieldParams) > 0:
             out += ',%s' % ','.join(['%s=%s' % item for item in fieldParams.items()])
         ties = self._getFieldTies()
+        if ppobj.TypeID == 2:
```
mantid-roman: Using numbers as type ids is no ideal. Can they be replaced by named constants?

mantid-roman (changes requested):
Just a couple of minor issues.

mantid-roman (approved):

martyngigg:
@rosswhitfield Sure. I'm in the deb packaging zone with bibtex anyway!

martyngigg:
@rosswhitfield The updated version is in the PPA and has the Python 3 packages too.

PR #21627, opened 2018-01-23
Include time spent in Algorithm::execute in reported time.

### LINKED ISSUE #20445: Algorithm timer shows incorrect runtime
When running an algorithm, the log outputs notices that look like this:
```
AlgorithmName started
AlgorithmName successful, Duration 0.01 seconds
```
The duration here can be extremely misleading.  For example:
```python
import time
start = time.time()
ws = CreateWorkspace(DataX=range(1000000), DataY=range(1000000), NSpec=1000)
print "%.2f sec" % (time.time() - start)
```
Output:
```
6.63 sec
```
Log:
```
CreateWorkspace started
CreateWorkspace successful, Duration 0.07 seconds
```
This happens because the timer is only measuring the `exec` of the algorithm, but there is considerable overhead (separate issue, see #20443) that occurs prior to that. Additionally, the "Algorithm started" message appears long before `exec` is called, so you will actually see the "started" notice, a long delay, followed by a claim of a short duration.

Since the overhead of calling an Algorithm can be so much heavier than the execution of the Algorithm itself, the duration of `exec` alone probably isn't very useful information and we should begin measuring the runtime much sooner.

In any case, a user would expect the timer to start when the "Algorithm started" notice is issued at the latest.

### LINKED ISSUE #21626: Incomplete Algorithm time reporting.
`Algorithm` does not include property validation and other setup code in the reported time. This can lead to large confusing discrepancies between observed and logged time.

SimonHeybrock:
See issue description. Also add debug output regarding time within execute.

**To test:**

Code review.

Fixes #21626.
Fixes #20445.

**Release Notes** 

Updated.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/API/src/Algorithm.cpp
```diff
@@ -460,6 +461,7 @@ bool Algorithm::execute() {
     logAlgorithmInfo();
 
   // Check all properties for validity
+  float timingInit = timer.elapsed();
```
OwenArnold: I find the API for `Timer::elapsed` to be counter intuitive. Reading the code I initially thought that each time point you assign was a total offset since `time` is constructed, however the reset argument defaults to true. Thus your code gives the correct behaviour. I think an improvement to readability for future maintainers would be to explicitly pass the flag indicating that the timer is reset as part of the `elapsed` call.

Framework/API/src/Algorithm.cpp
```diff
@@ -584,6 +592,13 @@ bool Algorithm::execute() {
       setExecuted(true);
 
       // Log that execution has completed.
+      getLogger().debug() << "Time to validate properties: "
```
martyngigg: We've observed in the past that there is quite a difference in performance with the loggers if you build the string up externally, either with a `stringstream` or just a `string` and `to_string` and then pass it to getLogger().debug(message_str) in one shot. Is that worth doing here?

OwenArnold (changes requested):
Works. One minor code change requested.

Framework/API/src/Algorithm.cpp
martyngigg: :+1:

Framework/API/src/Algorithm.cpp
```diff
@@ -584,6 +591,16 @@ bool Algorithm::execute() {
       setExecuted(true);
 
       // Log that execution has completed.
+      getLogger().debug() << "Time to validate properties: " +
```
martyngigg: I'd meant something more like:
```
std::string dbgMsg = "Time to validate properties: ";
dbgMsg.append(std::to_string(timingPropertyValidation)).append(" seconds\n");

+ similar append lines for other info...

# and finally send it to the debug log stream in one shot. The stream api flushes each time it encouters a "\n" so
# this will just send the whole lot out once.
getLogger().debug(dbgMsg); 
```

OwenArnold (approved):

OwenArnold:
Do not merge until Martn's review is complete and tests all pass.

Framework/API/src/Algorithm.cpp
```diff
@@ -584,6 +591,14 @@ bool Algorithm::execute() {
       setExecuted(true);
 
       // Log that execution has completed.
+      getLogger().debug("Time to validate properties: " +
```
martyngigg: :+1:

martyngigg (approved):
Looks good to me too.

PR #21639, opened 2018-01-23
Unicode workspace name

peterfpeterson:
Allow for people to use unicode when requesting workspaces in scripts with python2. 

**To test:**

There are two scripts to try out (python2 will show an error, python3 auto-converts already):
```python
from __future__ import (absolute_import, division, print_function, unicode_literals)

CreateSampleWorkspace(OutputWorkspace='w')
w = mtd['w']
print(w)
```
and
```python

CreateSampleWorkspace(OutputWorkspace='w')
w = mtd[unicode('w')]
```
Both of which should work.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/PythonInterface/inc/MantidPythonInterface/kernel/DataServiceExporter.h
```diff
@@ -27,11 +27,38 @@
...
+#include <boost/python/list.hpp>
+#include <boost/python/str.hpp>
 
 #include <set>
 
+/**
+ * Convert a python object to a string or throw an exception. This will convert
+ * unicode strings in python2 via utf8.
+ *
+ * THIS IS DUPLICATED FROM
```
martyngigg: Why does this need to be duplicated?

peterfpeterson: It was having some issue with the function not being available from `_api.so` even though other things in `Converters` were.

martyngigg (approved):
This looks good and works as needed.

PR #21647, opened 2018-01-24
Fix possible crash when plotting from an asynchronous script in workbench

martyngigg:
Description of work.

Wraps key calls in the `matplotlib` backend so that they occur on the main GUI thread.

**To test:**

The following script either crashed the workbench immediately or on exit before these changes. 

```python
from __future__ import (absolute_import, print_function)

import matplotlib.pyplot as plt
import numpy as np

x = np.linspace(0, 10, 1000)
y = np.sin(x)

# plot
fig, ax = plt.subplots()
ax.plot(x, y)
fig.show()
```

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidPlot/pymantidplot/mpl/backend_mtdqt4agg.py
+++ b/MantidPlot/pymantidplot/mpl/backend_mtdqt4agg.py
@@ -96,12 +96,15 @@ class ThreadAwareFigureManagerQT(FigureManagerQT):
         FigureManagerQT.__init__(self, canvas, num)
         self._destroy_orig = self.destroy
         self.destroy = QAppThreadCall(self._destroy_orig)
+        self._show_orig = self.show
+        self.show = QAppThreadCall(self._show_orig)
 
 
 # ----------------------------------------------------------------------------------------------------------------------
-# Wrap the required functions
+# Patch known functions
+# ----------------------------------------------------------------------------------------------------------------------
 show = QAppThreadCall(show)
-# Use our figure manager
+
 FigureManager = ThreadAwareFigureManagerQT
 
 if MPL_HAVE_GIVEN_FIG_METHOD:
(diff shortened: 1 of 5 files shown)


AndreiSavici:
It still crashes for me:
```
QPixmap: It is not safe to use pixmaps outside the GUI thread
QPixmap: It is not safe to use pixmaps outside the GUI thread
QPixmap: It is not safe to use pixmaps outside the GUI thread
QPixmap: It is not safe to use pixmaps outside the GUI thread
QPixmap: It is not safe to use pixmaps outside the GUI thread
QPixmap: It is not safe to use pixmaps outside the GUI thread
QPixmap: It is not safe to use pixmaps outside the GUI thread
QPixmap: It is not safe to use pixmaps outside the GUI thread
QPixmap: It is not safe to use pixmaps outside the GUI thread
QPixmap: It is not safe to use pixmaps outside the GUI thread
QCoreApplication::sendPostedEvents: Cannot send posted events for objects in another thread
[xcb] Unknown request in queue while dequeuing
[xcb] Most likely this is a multi-threaded client and XInitThreads has not been called
[xcb] Aborting, sorry about that.
MantidPlot: ../../src/xcb_io.c:165: dequeue_pending_request: Assertion `!xcb_xlib_unknown_req_in_deq' failed.
MantidPlot: Fatal IO error 11 (Resource temporarily unavailable) on X server :0.
Aborted (core dumped)
```
I run the script from the script window, but it's the same in script interpreter. The only way to make it work it's with Execute->Mode->Serialized.

I am running Ubuntu 17.04

martyngigg:
@AndreiSavici This is for the workbench not MantidPlot but you make a good point that I hadn't considered we had the same problem there too. I'll see if I can apply the same fixes to our current MantidPlot backend.

mantid-roman (approved):
No crash. The active plot updates if the script window doesn't freeze

AndreiSavici:
Works :grinning:

PR #21669, opened 2018-01-31
Bugs fixes for Eigen 3.3, use Eigen CMake target

quantumsteve:
Description of work.

This attempts to use Eigen3's CMake target instead of simply specifying the `EIGEN3_INCLUDE_DIR`. 

Previous test failures appear to be related to the lifetime of temporary `Eigen::Vector3d` objects.  

`Eigen::Quaterniond` is a [Fixed-size vectorizable Eigen object](https://eigen.tuxfamily.org/dox/group__TopicFixedSizeVectorizable.html) and so I believe some precautions need to be taken when using them inside [std::vector](https://eigen.tuxfamily.org/dox/group__TopicStlContainers.html#title2)

**To test:**

* Code Review
* Check that the builds still pass

**Release Notes** 

Developers should be notified about support for Eigen v3.3.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Beamline/inc/MantidBeamline/ComponentInfo.h
```diff
@@ -53,7 +53,9 @@ class MANTID_BEAMLINE_DLL ComponentInfo {
       m_componentRanges;
   boost::shared_ptr<const std::vector<size_t>> m_parentIndices;
   Mantid::Kernel::cow_ptr<std::vector<Eigen::Vector3d>> m_positions;
-  Mantid::Kernel::cow_ptr<std::vector<Eigen::Quaterniond>> m_rotations;
+  Mantid::Kernel::cow_ptr<std::vector<
+      Eigen::Quaterniond, Eigen::aligned_allocator<Eigen::Quaterniond>>>
```
OwenArnold: After including StdVector -> includes src/StlSupport/StdVector specialisation, I think you will automatically get the right allocator. i.e. no need to express here.

quantumsteve: The Eigen documentation says that one needs to **both** the `StdVector` header an the aligned allocator. 

https://eigen.tuxfamily.org/dox/group__TopicStlContainers.html#title2

I tried using the `EIGEN_DEFINE_STL_VECTOR_SPECIALIZATION` that tries to hide the second requirement. Ensuring they are both defined everywhere it is necessary and not defined twice was more trouble than it is worth.

Framework/Beamline/CMakeLists.txt
```diff
@@ -34,12 +34,13 @@ add_library ( Beamline ${SRC_FILES} ${INC_FILES} )
...
                                  INSTALL_RPATH "@loader_path/../MacOS" )
-target_include_directories( Beamline SYSTEM PUBLIC ${EIGEN3_INCLUDE_DIR} )
+
+target_include_directories( Beamline SYSTEM PUBLIC ${Boost_INCLUDE_DIRS})
 
 # Add to the 'Framework' group in VS
 set_property ( TARGET Beamline PROPERTY FOLDER "MantidFramework" )
 
-target_link_libraries ( Beamline LINK_PRIVATE ${TCMALLOC_LIBRARIES_LINKTIME} 
+target_link_libraries ( Beamline LINK_PUBLIC Eigen3::Eigen LINK_PRIVATE ${TCMALLOC_LIBRARIES_LINKTIME}
```
SimonHeybrock: 👍 for more modern use of cmake.

SimonHeybrock (approved):
Looks good. Alignment can be very important for performance and should maybe be used more within Mantid. 👍

PR #21671, opened 2018-02-01
Test of google-test integration.

SimonHeybrock:
**Do not merge, this is only a build test**

Notes:
- When using `GTEST_ADD_TESTS`, only RHEL7 seems to produce output that Jenkins picks up. Probably using `add_test` is the better choice, so this is a non issue.
- `add_test` does not need to be used with filters as I did to set up two different `ctest` tests. If we do not mind having all tests of a module in the same `ctest` test all that needs to be done is setting the output xml name such that Jenkins finds it.


NickDraper:
Closing this testing PR

Issue #21640, opened 2018-01-24
Engineering Diffraction GUI shared code between fitting tabs needs refactoring


josephframsay:
The Fitting tab and the GSAS tab share a lot of code relating to the way they handle multiple workspaces being loaded. The plot area with a list of loaded runs could do with being refactored into its own widget.

For 3.13, aim to integrate `MultiRunFittingWidget` into the Fitting tab - this isn't feasible for 3.12

PR #21695, opened 2018-02-06
Engineering Diffraction GUI integrate multi-run widget into GSAS tab

josephframsay:
The new multi-run widget has been integrated with the GSAS tab of the ED GUI. This widget contains focused runs and their fitted peaks workspaces, indexed by run number and bank ID. Workspaces can be plotted, with or without their fitted peaks if available, removed from the widget, and plotted to a separate window. This replicates the behaviour of the fitting tab - the aim is to replace the relevant section of the fitting tab with this widget as well.

**To test:**

Sample data files [here](https://github.com/mantidproject/mantid/files/1652777/engggui_pr_data.zip). 

Load up the ED GUI (`Interfaces -> Diffraction -> Engineering Diffraction`) then click `GSAS Fitting` tab.

Click the first **Browse** button to browse for focused run data (in the zip file). Load all 3 runs, and make sure `277208_1`, `277208_2` and `277209_2` are displayed in the list widget next to the plot. Clicking each of them in turn should plot the run. Try only loading 2 runs, and make sure the third is added correctly once loaded on its own (you'll need to exit the GUI to get rid of the loaded runs).

Make sure the zoom tool works nicely as well - zoom by clicking and dragging, and reset by right-clicking

Try removing runs and plotting to a separate window. Note that, since GSASIIRefineFitPeaks isn't yet hooked up to the GUI, plotting fitted peaks is not yet supported.

Partially addresses #21640 

**Release Notes** 
Internal change, doesn't need to be in the release notes

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


ElliotAOram (commented):
Updating the data (e.g. running crop on the workspace in the ADS) does not update the mini plot (This is just observed behaviour - not sure if we want it to or not anyway?)

ElliotAOram (changes requested):
* Occasionally the error messages are delayed e.g. if you select a run and check the `plot Fitted Peaks` option then click `Plot to separate window`. It still makes the plot, but no error message for the fact there are no fitted peaks is displayed. Instead, when you do another operation (e.g. try to load data) you get the error message pop up
image
* Similarly when `Plot fitted peaks` is true, if you switch between run numbers you get the error message shown above for every time you switch
* tldr; ensure that the error message pops up when the invalid action takes place

ElliotAOram (changes requested):
If you load a run in the GUI, delete it in the Workspace Widget (ADS), then try to plot it in the interface, Mantid Crashes:
image

josephframsay:
@ElliotAOram the second problem is something I anticipated happening in a different PR - the error message isn't delayed at all, when we do an action the `selectrun` event fires. The handler for this in `EnggDiffGSSAFittingPresenter` doesn't check whether there are fit results, only that the user has opted to plot them. This is dealt with in [this commit](https://github.com/mantidproject/mantid/pull/21804/commits/17f86989a18f1d2b1e59227c50b2542e55ca0083) but this is dependent on quite a bit else from that PR - is it OK if we be a bit naughty and just let this slide?

josephframsay:
@ElliotAOram This is ready for review again. Re the problem with updating the data in the ADS not updating the mini plot, I discovered that updating the data in the ADS doesn't update the data in the model at all. I don't really mind this behaviour - it seems logical to me that once a workspace is in the GUI that's it. I changed the way plotting to separate window works, so now it always plots the workspace from the model. This is at least consistent, if not quite the Mantid-y way of doing things. If scientists complain I'd be happy to change it, but I can't see getting it working properly happening before the release

ElliotAOram:
@josephframsay 
* For the first point, that's fine: we just need to make sure it's working correctly in the other PR (which I will look at after this one) and that it gets merged for the Release.
* Re. second point: I agree that it is not entirely mantid-like but given the circumstance I think the approach you've taken is more logical 👍 As you say, if the users do prefer the other way, we can address this but it's not so major that it should be causing anyone pain for the next release cycle.

I'll test this again now

ElliotAOram:
Couple of Nits:
* Preview plots remain after you delete the data
  * Load data into the GUI
  * Select the data in the `Run number` - plot comes up in the preview as expected
  * Remove the run  using `Remove Run` button
  * Data removed but plot still shows. I think it would make most sense to at this point clear the model of the data? 
* Super Nit: Disable Load button on empty
  * It might be nice to disable the load button while the `Focused Run File` text field is empty and add a tooltip to the button while disable that says something like "Please specify a file to load via the browse menu or by typing the full path to the file in the text field"
  * Main reasoning behind this is (atleast to me) when I click load I expect to get the file directory browser like you would if you hit load on the Workspace Widget.

ElliotAOram (approved):
This look good now. Thanks for making those changes! One thing to think about for the future (Definitely not this release and will depend on what users want) is the possibility of some keyboard control for `Run Number`
For me at least I'd like to be able to:
* Highlight a run and press the delete key to remove it
* Click on a run at the top of the box then shift click a run at the bottom of the box to highlight it and all those in between.

But this does rely on users being interested really! 
Anyway... Once tests pass this is good to go. 👍

PR #21702, opened 2018-02-06
Fix relying on undocumented bug in older jinja2.

AndreiSavici:
On new linux operating systems building documentation yields ugly pages 
bad
This is what I expect
good

The reason is that the template file used relies on an undocumented bug of jinja2. In the documentation, setting a variable inside a scope has no effect on that variable outside of it. In some older versions this was not enforced. In the new versions, setting a variable inside the `for` loop does not affect the value in the next iteration.
I've used this [solution](https://stackoverflow.com/a/42888347/6259479)

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


rosswhitfield (approved):
LGTM on my computer. I also checked that it didn't break RHEL7

PR #21715, opened 2018-02-07
ISIS Powder PEARL add option to disable empty instrument subtraction

### LINKED ISSUE #21712: ISIS Powder issues when focusing empty instrument runs
`isis_powder` focusing scripts run into problems when focusing empty instrument runs. This is because we end up with one of two situations:
- We're subtracting last cycle's empty from this cycle's, which gives us noisy data with low intensity, which is useless
- We load this cycle's empty, subtract it from this cycle's empty, then delete it from the ADS. Trying to use the run then causes a crash, as the workspace has been deleted

Focusing empties should be made possible by either detecting whether this run is an empty or adding a flag to disable subtracting. Or both

josephframsay:
The option to **not** subtract the empty instrument runs from a run during focusing was added in `isis_powder` for PEARL. Subtraction is now disabled under one of the following two conditions:
- The user has passed `subtract_empty_instrument=False` to `pearl.focus`. This is useful for focusing empties right at the start of a cycle, when we're still using cycle mapping parameters from the last cycle, in order to avoid subtracting last cycle's empty from this cycle's, which gives us bogus data
- The run numbers being focused overlap with the empty run numbers. This prevents a crash which occurs when using this cycle's mapping parameters, as the empty is subtracted from itself and then deleted, so trying to access it for the next step of focusing is not valid

**To test:**

If not reviewing from ISIS, you'll need to change the second line and do some setup locally - ping me message for an explanation of how to set up ISIS_Powder directories, as it can be fiddly.

```
from isis_powder import Pearl

pearl = Pearl(config_file=r"\\olympic\Babylon5\Public\JoeRamsay\PEARL\pearl_shared_config.yaml")

subtracted = pearl.focus(run_number="97574")
RenameWorkspaces(InputWorkspaces=[subtracted.getName()] + list(subtracted.getNames()), 
                 Suffix="-empties")

not_subtracted = pearl.focus(run_number="97574", subtract_empty_instrument=False)

res, messages = CompareWorkspaces(Workspace1=subtracted, Workspace2=not_subtracted)
print(res) # Should print False

focused_empty = pearl.focus(run_number="97545-97552")
```

Make sure `subtracted` and `not_subtracted` don't match. They'll be very similar, but `subtracted` should have slightly lower intensity.

The final line should run with no crash, and give you a nice curve something like:
capture

Fixes #21712

**Release Notes** 
[Here](https://github.com/mantidproject/mantid/compare/21712_isis_powder_empty_runs?expand=1#diff-6d316275f01d64b8113c082a773943ae)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


ElliotAOram:
@josephframsay can you rebase this to master when you get a chance (it has the google test failures in it)

ElliotAOram:
Change works as expected. Would it be possible to put a unit test case in for this?

ElliotAOram (approved):

ElliotAOram:
Merge conflicts just in release notes so I've fixed it myself (see above commit)

PR #21717, opened 2018-02-08
Adding Muon Maxent

### LINKED ISSUE #21630: Adding MuonMaxEnt
One of the scientists has provided a python algorithm, which they are happy with, that needs to be added into the Mantid code base. The scientists' code will be cleaned up and added into the code base.

AnthonyLim23:
This work was to add James' Maxent code to the Mantid code base. This has included: 

- creating tests (unit and system)
- fixing warnings/errors (including reducing the complexity of the code)
- adding documentation

This will be removed once MaxEnt can do the equivalent calculations. 

**To test:**
I have not changed how the code works and James was happy with how it worked. 

Fixes #21630. 

**Release Notes** 

In release notes
---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Anders-Markvardsen:
@AnthonyLim23  do you want to fix the flake8 errors before I start testing this PR?

AnthonyLim23:
Just working through them now. Should be done in about an hour

AnthonyLim23:
@Anders-Markvardsen looks like the flake 8 warnings are now fixed

martyngigg (commented):
Can we move the non-algorithm files out of algorithms?

When we start the framework it has to open every Python file to check if there is an algorithm in there. This can make it slow when the number of files increases. Can we have a muon package in scripts?

AnthonyLim23:
I could move them to the scripts, but will take some time as the paths will need changing. @martyngigg is there an easy way to get the algorithm to import from the script repository?

Framework/PythonInterface/plugins/algorithms/Muon/MuonMaxent.py
```diff
@@ -0,0 +1,441 @@
+import numpy as np
+import math
+from MuonMaxentTools.multimaxalpha import MULTIMAX
+from mantid.api import *
+     # PythonAlgorithm, registerAlgorithm, WorkspaceProperty
+from mantid.kernel import *
+from mantid.simpleapi import *
```
martyngigg: It would be better to import `import mantid.simpleapi as api` and then qualify each algorithm usage.

martyngigg:
@AnthonyLim23 I don't quite understand what you mean?

Framework/PythonInterface/plugins/algorithms/Muon/MuonMaxent.py
AnthonyLim23: I agree, this is suppose to be a quick port of a scientist's code as a temporary solution for MaxEnt. I did not want to make the change as it could be time consuming to work out

AnthonyLim23:
the maxEnt alg needs to import the stuff thats now in the script repository. Do I need to worry about paths?

martyngigg:
Files in the codebase can't have any dependencies on the script repository. It is purely for external/optional code

docs/source/algorithms/MuonMaxent-v1.rst
```diff
@@ -0,0 +1,64 @@
...
+where :math:`\Re(z)` is the real part of :math:`z`, :math:`\mathrm{IFFT}` is the inverse fast Fourier transform, :math:`\phi_k` is the phase and :math:`A_k` is the asymmetry of the of the  :math:`k^\mathrm{th}` spectrum. 
+The asymmetry is normalised such that :math:`\sum_k A_k = 1`. 
+The instrumental frequency response function, :math:`R(\omega)`, is  is in general complex (due to a 
+non-symmetric pulse shape) and is the same for all spectra. The values of the phases and asymmetries are fitted in the outer loop of the algorithm. 
+
+The :math:`\chi^2` value is calculated via the equation
+
+.. math:: \chi^2 = F\frac{\sum_{k,t} (D_k(t)-g_k(t))^2 }{E_k(t)^2},
+
+where :math:`F` is the :code:`Factor` and is of order 1.0 (but can be adjusted within the algorithm for a better fit).
```
Anders-Markvardsen: By "(but can be adjusted within the algorithm for a better fit)" do you mean that the algorithm adjust F during run time or that the user can select what this value should be?

AnthonyLim23: ah, sorry. The F value can be changed by the user (run a new algorithm)

docs/source/algorithms/MuonMaxent-v1.rst
```diff
@@ -0,0 +1,64 @@
...
+added with infinite errors. The time domain data prior to :code:`FirstGoodTime` also have their errors set to infinity. The algorithm will produce the frequency spectra :math:`f(\omega)` and this is assumed to be real and positive. 
+The upper limit of the frequency spectra is determined by :code:`MaxField`. The maximum frequency, :math:`\omega_\mathrm{max}` can be less than the Nyquist limit :math:`\frac{\pi}{\delta T}` if the instrumental frequency response function for 
+:math:`\omega>\omega_\mathrm{max}` is approximatley zero. The inital estimate of the frequency spectrum is flat. 
+
+The algorithm calculates an estimate of each time domain spectra, :math:`g_k(t)` by the equation
+
+.. math::  g_k(t)=(1+A_k \Re(\mathrm{IFFT}(f(\omega) R(\omega))\exp(-j\phi_k) ) ),
+
+where :math:`\Re(z)` is the real part of :math:`z`, :math:`\mathrm{IFFT}` is the inverse fast Fourier transform, :math:`\phi_k` is the phase and :math:`A_k` is the asymmetry of the of the  :math:`k^\mathrm{th}` spectrum.
```
Anders-Markvardsen: There exist two ways of defining IFFT, e.g. from https://www.gnu.org/software/gsl/doc/html/fft.html there is the paragraph: "In general there are two possible choices for the sign of the exponential in the transform/ inverse-transform pair. GSL follows the same convention as FFTPACK, using a negative exponential for the forward transform. The advantage of this convention is that the inverse transform recreates the original function with simple Fourier synthesis. Numerical Recipes uses the opposite convention, a positive exponential in the forward transform.". Which one is used here?

AnthonyLim23: I will double check in the morning

Anders-Markvardsen (changes requested):
Tested with usage example and where I played around with a couple of the parameters and executed for all these cases. 

Can you answer the two questions I had otherwise looks good and nice .rst doc page

docs/source/algorithms/MuonMaxent-v1.rst
AnthonyLim23: So looking at the code we dont seem to use the FFT module (I should try to delete it) and instead numpy is used. According to the docs https://docs.scipy.org/doc/numpy-1.12.0/reference/routines.fft.html#module-numpy.fft the forward transform has a minus sign

Anders-Markvardsen: OK. Note I am away Thursday-Friday this week. Simply e.g. adding the above reference to number to numpy's definition I would say is fine

AnthonyLim23:
that should be everything

Anders-Markvardsen (approved):

AnthonyLim23:
@mantidproject/gatekeepers could we please get this merged?

PR #21728, opened 2018-02-09
Add V20 IDFs

OwenArnold:
Fixes #21728 

I'm proposing that these are added as two separate instruments
**V20_DIFFRACTION** and 
**V20_IMAGING**

This would allow us to make a small modification [here](https://git.esss.dk/wedel/wfm_stitching/blob/master/wfm_stitching/wfm_processing.py#L79)

which would allow users to simplify something existing like [this](https://git.esss.dk/wedel/wfm_stitching/blob/master/diffraction_example.py#L18)

```python
instrument_filename = 'sample-data/HZB_V20_Definition_4-tubes-90-degree.xml'
...
sample_processed = p.process(
    sample_workspace, instrument_filename, '0,64,70000', scale=1e3,
    delete_temporary_workspaces=True)

```

to this
```
sample_processed = p.process(
    sample_workspace, 'V20_DIFFRACTION', '0,64,70000', scale=1e3,
    delete_temporary_workspaces=True)
```
**Discussion**
There are pros and cons to the multi-name approach to the same instrument. We should consider these before merging.
**Notes**
According to [this](http://docs.mantidproject.org/nightly/algorithms/DownloadInstrument-v1.html) once changes are pushed to master these instruments should be available immediately.
**Testing**
* Run `LoadEmptyInstrument` setting the instrument name to be either of the above.
* Check the facilities xml file. HZB should now have three instruments.

### DIFF
--- a/instrument/Facilities.xml
+++ b/instrument/Facilities.xml
@@ -797,6 +797,12 @@
     <technique>TOF Direct Geometry Spectroscopy</technique>
     <technique>Single Crystal Diffraction</technique>
   </instrument>
+  <instrument name="V20_DIFFRACTION">
+    <technique>Neutron Diffraction</technique>
+  </instrument>
+  <instrument name="V20_IMAGING">
+    <technique>Neutron Imaging</technique>
+  </instrument>
 </facility>
 
 <!--  Test Facility to allow example usage of Live listeners against "Fake" instrument sources -->
--- a/instrument/V20_DIFFRACTION_Definition.xml
+++ b/instrument/V20_DIFFRACTION_Definition.xml
@@ -1,0 +1,62 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- For help on the notation used to specify an Instrument Definition File
+see http://www.mantidproject.org/IDF -->
+<instrument xmlns="http://www.mantidproject.org/IDF/1.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.mantidproject.org/IDF/1.0 Schema/IDFSchema.xsd"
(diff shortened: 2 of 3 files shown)


SimonHeybrock:
@OwenArnold If we would not go with the multi-name approach, what would the alternative be?

Related problem: V20 plans to add another different detector bank for the diffraction runs. This might mean that we need multiple configurations for the same (diffraction) instrument. How can we handle that?

OwenArnold:
@SimonHeybrock, yes this is the weakness. The IDFs do have a concept of a single valid "window of time" for each IDF, i.e. you can specify dates within which an IDF would be valid and used with the collected data. But the core assumption with the current mechanism is that the IDF gets superseded, and the old IDF is only needed for historical runs. I think this is not the case with V20 where the bench-style setup of the beamline is going to demand a lot of flexibility.

My fear is that we will not be able to address all the setups we require by name alone. One option might be to keep the current distribution mechanism (IDFs in mantid repo) but to use something more similar to this in the scripts, i.e. provide full file names. 

```python
from os import path
from mantid.kernel import ConfigService as config
to_load = path.join(config['instrumentDefinition.directory'], 'V20_IMAGING_Definition_with_256_detector.xml')
LoadInstrument(Filename=to_load, ...)
```

SimonHeybrock:
As long as they have not much more than 5 different configurations I would not worry. We can just go ahead and add them as separate instruments. If the number grows much larger we can come up with another solution.

Any reason for marking this as "in progress"? We can just go forward with this and see if it works for them?

SimonHeybrock (approved):

OwenArnold:
@SimonHeybrock. "In Progress" was just to ensure we had the above discussion before going ahead :)

PR #21758, opened 2018-02-16
Fix complex run summation in MultiFileNameParser

### LINKED ISSUE #21757: MultipleFileProperty does not work with `1-2+3-4` like string.
`MutlipleFileProperty` does not support complex summation formulas such as `1-2+4` or `1+3-6`.

soininen:
This PR adds the ability to specify complex sum ranges, like `1-3+5` and `2+5-23` to `MultiFileNameParser`.

**To test:**

Try, for example, the `Load` or `LoadAndMerge` algorithms with complex summation both from MantidPlot (or Workbench, if you dare) and from Python.

Fixes #21757.

**Release Notes** 

The fix is mentioned in the Framework release notes.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


docs/source/release/v3.12.0/framework.rst
```diff
@@ -65,6 +65,7 @@ Core Functionality
 - Added new classes ``ConfigObserver`` for listening for changes to any configuration property and ``ConfigPropertyObserver`` for listening to changes to an individual config property of interest.
 - Fixed the calculation of scattering length and scattering length squared for :py:obj:`Material <mantid.kernel.Material>`
 - Fixed the behaviour of ``UpdateInstrumentDefinitions.OnStartup`` in the :ref:`properties file <Properties File>`. It was not being used correctly for using the updated ``Facilities.xml`` file.
+- ``MultiFileProperty`` now accepts complex summation ranges for run numbers, such as ``111-113+115`` and ``111-115+123-132``.
```
SimonHeybrock: Question:
Is it potentially confusing to have `-` for defining a range but `+` combining ranges? Is this convention following any existing behavior?

SimonHeybrock (changes requested):
Looks good overall, sensible tests were added.

See one question below.

Furthermore: Should this new option also be documented somewhere (tooltips, for example?).

docs/source/release/v3.12.0/framework.rst
soininen: `+` and `-` deal with adding (merging) of data. The documentation of `MultipleFileProperty` states that 'Added Ranges and Stepped Added Ranges may be plussed however.' At least to me the behavior of adding 'added ranges' with `+` feels logical, and the use pattern will be common at here at ILL where users frequently merge several files into one workspace.

I'll see if I can improve the documentation. At least `MultipleFileProperty` could include examples of this kind, maybe `Load`, `LoadAndMerge` and some tooltip could be improved as well

SimonHeybrock: My point is: Is there a risk that having `-` for the range is interpreting as *subtracting* that run from the others? If so, would `:` work better?

soininen: There is indeed that risk, but `-` is and has always been an 'Added Range', also according to `MultipleFileProperty` documentation. `:` is a range of individual files so it cannot be used for summing. I don't think `n:m+k:l` makes sense, and is explicitly prohibited in `MultipleFileProperty` documentation.

SimonHeybrock: Good point, I did not know that. I guess then we do not have a choice and just keep it as it is. :)

SimonHeybrock (approved):

martyngigg:
Test failure is unrelated.

Issue #18990, opened 2017-02-24
Mantid crashes when checking the sample log of some workspaces


AntonPiccardoSelg:
Querying the sample log of the system test data `SANS2D00005512.nxs` creates an unhandled exception. Strangely this is only true for the child workspaces 1 to 11. Querying the sample log from child workspaces 12 and 13 works as expeceted.

### Expected behavior

The sample logs can be queried on all child workspaces of a group workspace;

### Actual behavior

Mantid crashes when querying the sample logs on some child workspaces.

### Steps to reproduce the behavior

See above

### Platforms affected

Seen on Ubuntu 16.04

AntonPiccardoSelg:
Initial scoping of the issue. The issue occurs for the `selog>Tank_Pressure` entry. The point fo failure is
[here](https://github.com/mantidproject/mantid/find/master#L2023:L2024).

martyngigg:
This is most likely a filtering issue.

When you say querying, can I just extract the log in python?

AntonPiccardoSelg:
I stumbled across it via the GUI. Right-click on the workspace, then select `Sample Logs...`. This will cause the crash.

martyngigg:
It looks like the problem here is that log has 1 value in it and then filtering has removed that value and we are left with an empty log. I will have to investigate further as to whether the filtering has been applied appropriately.

martyngigg:
@wdzhou Would you have time to look into this issue? You understand the log filtering better than I do.

martyngigg:
![](https://github.trello.services/images/mini-trello-icon.png) [Filtering Crash](https://trello.com/c/xdbSs4wJ/312-filtering-crash)

Pasarus:
I've found a similar bug which throws an unexpected exception, this happens when you attempt to show sample logs for a group workspace. This can be done by highlighting the group workspace and then using the option workspace->show logs in MantidPlot as it isn't disabled here like in the right click menu.

martyngigg:
This appears to have been fixed in the current version.

PR #21778, opened 2018-02-19
Crash reporting service

### LINKED ISSUE #21709: Impliment Crash reporting service

Matthew-Andrew:
This adds crash reporting to the last chance error handler.
**To test:**

- To test you first need to provoke an uncaught exception. One method is here https://github.com/mantidproject/mantid/issues/18990

- When you open mantid turn usage reporting on in the splashscreen. This is no longer necessary and error reporting should work if usage reporting is turned off. **However where the usage reporting gets its url from has been changed so this should be checked to still send usage reports to the right places.**

- Check that the usage reporting startup message is sent correctly it should go to errorreports.mantidproject.org/api/error

- Confirm that when you cause an uncaught exception a dialog window appears asking you whether you wish to continue using mantid and what information you want to share.

- Check that Mantid continues or terminates correctly.

- Check that the correct information (including importantly nothing if you decide to share nothing) is sent in a post request. 

- Also requires a code review.

ToDo
Fixes #21709 . 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


scripts/ErrorReporter/errorreport.ui
```diff
@@ -0,0 +1,199 @@
...
+      </item>
+     </layout>
+    </widget>
+   </item>
+   <item>
+    <layout class="QHBoxLayout" name="horizontalLayout_2">
+     <item>
+      <layout class="QVBoxLayout" name="verticalLayout">
+       <item>
+        <widget class="QTextBrowser" name="requestTextBrowser">
```
gvardany: Just noticed: the text area does not seem to be editable now, is this readonly?

scripts/ErrorReporter/errorreport.ui
```diff
@@ -0,0 +1,199 @@
...
+      <widget class="QPushButton" name="fullShareButton">
+       <property name="text">
+        <string>Yes, share information</string>
+       </property>
+      </widget>
+     </item>
+     <item>
+      <widget class="QPushButton" name="nonIDShareButton">
+       <property name="text">
+        <string>Share non-identifiable information</string>
```
gvardany: I think tooltips would be useful for the buttons, especially to explain what is non-identifiable.

gvardany:
Am I right to assume that if running from `build/bin`, it won't report anything even if `Report usage data` is checked?

Mantid-Matthew:
I have been running it from build/bin using the launcher script and it has been reporting things when Report Usage data is checked. It is currently configured to send things to a test server rather than the usage reporting system though.

scripts/ErrorReporter/errorreport.ui
Mantid-Matthew: The text area is read only

gvardany:
GUI is very nice and shows up in the provided crash scenario as expected. Continue/terminate works. Still my reports do not appear in the test receiver stage while running from built mantid with crash reporting on. Am I doing something wrong here? 
The code is fine, except some const consistency and missing doxygen in C++ side. \
Also I was wondering if the message box should be made editable with the free typed text reported as well? I think macOS system crash offers this. Not sure though how good is Qt in validating against javascript injections.

gvardany:
It seems I have some general connectivity issue on :OS X: , so I cannot test the main functionality, I let someone else to test.

NickDraper (changes requested):
I discussed some GUI changes to the form.

1. reducing the white empty box after the question about more details
1. Change the question to ask for contact details to allow follow up.

NickDraper (approved):

NickDraper:
Looking good now, Just add the database access details to the wiki protected information page

https://www.mantidproject.org/Private:Protected_Information

NickDraper (changes requested):
We do need a release notes entry, ideally with a screenshot and the different options explained in the release notes.

Framework/Properties/Mantid.properties.template
```diff
@@ -54,6 +54,8 @@ CheckMantidVersion.DownloadURL = http://download.mantidproject.org
 
 # Whether to report usage statistics back to central server
 usagereports.enabled = @ENABLE_USAGE_REPORTS@
+error_root_url = http://errorreports.mantidproject.org/api/error
+usage_root_url = http://reports.mantidproject.org
```
martyngigg: Can we follow the standard property config style for names, please?

I think this should be:

```
usagereports.rooturl =
```

and

```
errorreports.rooturl = 
```

Also I think the root errors should be "http://errorreports.mantidproject.org" and then the `api/error` appended internally so that both are consistent. If I'm testing then I just want to be able to change that to `localhost` and have it work.

martyngigg (changes requested):

Framework/Kernel/inc/MantidKernel/UsageService.h
```diff
@@ -1,6 +1,7 @@
 #ifndef MANTID_KERNEL_USAGESERVICE_H_
 #define MANTID_KERNEL_USAGESERVICE_H_
 
+#include "MantidKernel/DateAndTime.h"
```
martyngigg: This should only need `#include <MantidTypes/Core/DateAndTime.h>`

Framework/Kernel/src/ErrorReporter.cpp
```diff
@@ -0,0 +1,141 @@
...
+namespace {
+/// static logger
+Logger g_log("ErrorReporter");
+}
+
+//----------------------------------------------------------------------------------------------
+// Constructor for ErrorReporter
+/** Constructor
+*/
+ErrorReporter::ErrorReporter(std::string application,
```
martyngigg: We can use [delegating constructors](http://en.cppreference.com/w/cpp/language/initializer_list) (search for delegating constructors on the page) here to reduce the copy-and-pasted code:

```
ErrorReporter::ErrorReporter(std::string application,
                             Types::Core::time_duration upTime,
                             std::string exitCode, bool share) : ErrorReporter(application, upTime, exitCode, share, "", "")
```

This way we only define the constructor logic once along with the general case of all parameters and then simply use it elsewhere. This was only possible in >=C++11.

Framework/Kernel/src/ErrorReporter.cpp
```diff
@@ -0,0 +1,141 @@
...
+*/
+int ErrorReporter::sendReport(const std::string &message,
+                              const std::string &url) {
+  int status = -1;
+  try {
+    Kernel::InternetHelper helper;
+    std::stringstream responseStream;
+    helper.setTimeout(20);
+    helper.setBody(message);
+    status = helper.sendRequest(url, responseStream);
```
martyngigg: Maybe a `.debug` log message to indicate the status and response for future debugging?

Framework/Kernel/src/UsageService.cpp
```diff
@@ -73,6 +65,13 @@ UsageServiceImpl::UsageServiceImpl()
       m_startupActiveMethod(this, &UsageServiceImpl::sendStartupAsyncImpl),
       m_featureActiveMethod(this, &UsageServiceImpl::sendFeatureAsyncImpl) {
   setInterval(60);
+  int retval = Mantid::Kernel::ConfigService::Instance().getValue(
+      "usagereports.rooturl", m_url);
+  if (retval == 0) {
+    g_log.error() << "Failed to load usage report url\n";
+  } else {
+    g_log.information() << "Root usage reporting url is " << m_url << "\n";
```
martyngigg: I think this is debug level information.

martyngigg (changes requested):
This looks really nice. Just a few minor coding issues.

Framework/Kernel/src/ErrorReporter.cpp
Mantid-Matthew: There is already some debug level information about the post request within Internet helper with responds with for example,

Sending http POST request to: http://errorreports.mantidproject.org/api/error
Answer from web: 201 Created
Continue working.

Do we need more in addition to this ?

NickDraper (approved):

martyngigg (approved):
Thanks, @Mantid-Matthew. Good to go from my perspective.

Framework/Kernel/src/ErrorReporter.cpp
martyngigg: No that's okay. I'd forgotten about the other level of logging.

PR #21815, opened 2018-02-22
ISIS Powder disable normalisation and empty corrections when current is 0

### LINKED ISSUE #21805: ISIS Powder NormaliseByCurrent crash when current is 0
`NormaliseByCurrent` raises a divide by zero error (`The proton charge found in the input workspace run information is zero`) for POLARIS when:
- input mode is `individual` and one of the runs has `tot_prtn_chrg = 0` in the sample logs
- input mode is `summed` and all of the runs have `tot_prtn_chrg = 0` in the sample logs

I presume that this also affects the other `isis_powder` instruments but none have come across it yet (or they may just have not reported it, as it's obvious why it happens).

POLARIS never care about runs with zero current (I would also presume that other instruments don't either), so we can throw these away as we see fit. The cleanest, from my point of view, would be to just skip `NormaliseByCurrent` altogether when `tot_prtn_chrg = 0`. This should be safe, as if there's no current then there should also be no data, so when we sum up these runs they'll have no effect on the end product

`103106` has valid current, `103107` and `103108` do not.
```
pol.focus(run_number="103106-103107", input_mode="individual")
pol.focus(run_number="103107-102108")
```
Both lines cause the crash

josephframsay:
Previously, when processing runs with total current of 0, `isis_powder` scripts would crash with a divide by 0 error from `NormaliseByCurrent`. This is annoying if scientists are focusing and summing up a long list of runs, some of which may have no current - we just want these runs to contribute nothing to the final result.

When current is 0 (`tot_prtn_chrg` in sample logs), we don't bother running `NormaliseByCurrent`, and we also don't subtract the empty, as this gives us negative counts.

The end result should be:
- For `input_mode="summed"`, if the runs all have 0 current, the end result should be a workspace with 0 counts. If some of the runs have counts, the output workspace should be the same as if we hadn't included the empty runs
- For `input_mode="individual"`, we should be able to focus runs with 0 current on their own, and just get a workspace with 0 counts

**To test:**

If you're not at ISIS, you'll need to change the second line - ping me a message about setting up ISIS Powder directories, as it can be fiddly. Bear in mind `103106 has current, `103107` does not.
```
from isis_powder import Polaris
pol = Polaris(config_file="D:\ISIS_Powder\POLARIS\polaris_config.yaml",
              do_van_normalisation=False)
```
Focus a single workspace with 0 current, this not crash, and the output workspaces should be flat lines at 0
```
pol.focus(run_number="103107")
```
Now sum up `103107` and `1031081` - again, we should get workspaces with no counts
```
pol.focus(run_number="103107-103108")
```
Now focus `103106` and compare it to the result of summing `103106` with `103107` - they should be the same
```
individual = pol.focus(run_number="103106")
summed = pol.focus(run_number="103106-103107", input_mode="summed")
CompareWorkspaces(individual, summed)
```

Fixes #21805 

**Release Notes** 
[Here](https://github.com/mantidproject/mantid/compare/21805_isispowder_normalisebycurrent?expand=1#diff-6d316275f01d64b8113c082a773943ae)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


ElliotAOram:
@josephframsay Would it be worth adding a warning message to the result log when it's a 0 current workspace? Or are we expecting that scientists should already be aware of this?

ElliotAOram (changes requested):
When summing the workspaces with:
`pol.focus(run_number="103107-103108")`
I get a flat workspace:
image
I have amended the script slightly from the above:
```
from isis_powder import Polaris
pol = Polaris(config_file="\\olympic\babylon5\Public\JoeRamsay\POLARIS\polaris_shared_config.yaml",
              do_van_normalisation=False,
              mode="Rieveld")
```

josephframsay:
@ElliotAOram `103107-103108` should be flat, as both workspaces have 0 current, so this is fine!

ElliotAOram:
Ah yes, it says it right there in the PR test description.... My bad!

ElliotAOram (approved):
I've successfully managed to read the rest of the test instructions properly... and this all works as expected. This is another change that can not be  unit tested due to the structure of ISIS_Powder (myself and Joe will be discussing how to deal with this going forward in the upcoming maintenance period)

This is good to merge once it passes the build servers

ElliotAOram (changes requested):
@josephframsay looks like system tests have caught something

ElliotAOram:
By chance @josephframsay I have found something that we should possible watch out for as it seem like it will cause us pain in the future (unless this is expected behaviour):

* If you do not define a chopper mode in the instrument and then focus you get a sensible error about the chopper mode not being set - Good!
* If you then define the mode in the focus e.g. ` pol.focus(run_number="103107", mode="Rietveld")` it works and runs correctly - Good!
* All subsequent runs then default to that chopper mode without specifying the chopper mode e.g.:
```
from isis_powder import Polaris
pol = Polaris(config_file="...", do_van_normalisation=False)

# None of the below run as they complain about no chopper mode - as expected
pol.focus(run_number="103107")
pol.focus(run_number="103107-103108")
pol.focus(run_number="103108")

# Just specifying the chopper mode for the first focus, means all the subsequent focuses use the same chopper mode. e.i. the chopper mode persists through - is this what we want?
pol.focus(run_number="103107", mode="Rietveld")
pol.focus(run_number="103107-103108")
pol.focus(run_number="103108")
```

josephframsay:
@ElliotAOram I also thought this was a bit nasty, but apparently this is what POLARIS want - see #21420 and docs [here](http://docs.mantidproject.org/nightly/techniques/ISISPowder-Tutorials.html#how-objects-hold-state-in-isis-powder)

ElliotAOram:
Ah yeah sorry I remember about this now. Thanks

ElliotAOram (approved):
Changes look good and the tests that were failing now pass locally. I've hit rebuild on the failing tests as they look unrelated but it's still probably best to make sure they pass before merging. Functionally it's ready to merge. Thanks for the changes!

PR #21819, opened 2018-02-22
Fix LoadEventNexus and LoadNexusMonitors for OFFSPEC 44754+

### LINKED ISSUE #21818: Fix LoadEventNexus/LoadNexusMonitors bug for OFFSPEC 44754+
### Expected behavior
Running
```python
LoadEventNexus("OFFSPEC0045523", OutputWorkspace='test', LoadMonitors=True)
```
should produce two workspaces `test` and `test_monitors` where test contains the event data and `test_monitors` contains the monitors histogram data.

### Actual behavior
The following error message is produced:

```
Supplied null pointer to getData
Error while loading the monitors from the file. File may contain no monitors.
```

This is due to a fault in the instrument software which caused files to be generated with monitors 
containing histogram data and the usual event-like entries but an empty `event_id` array. This fault is being fixed but the runs still need to be loaded.

### Platforms affected
All

edwardb96:
Loading of runs for OFFSPEC 44754+ fails due to an empty `event_id` array being generated in the nexus file alongside the histogram data for the monitor. This PR patches this in order to allow these runs to be properly loaded.

**To test:**
Attempt to load a combination of event and histogram files using `LoadEventNexus` and `LoadNexusMonitors`.

Fixes #21818 . 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/DataHandling/inc/MantidDataHandling/LoadNexusMonitors2.h
+++ b/Framework/DataHandling/inc/MantidDataHandling/LoadNexusMonitors2.h
@@ -102,6 +102,14 @@ private:
   bool isEventMonitor(::NeXus::File &monitorFileHandle) const;
   std::size_t sizeOfUnopenedEntry(::NeXus::File &file,
                                   const std::string &entryName) const;
+  bool eventIdNotEmptyIfExists(
+      ::NeXus::File &monitorFileHandle,
+      std::map<std::string, std::string> const &entries) const;
+  bool hasAllEventLikeAttributes(
+      std::map<std::string, std::string> const &entries) const;
+  bool keyExists(std::string const &key,
+                 std::map<std::string, std::string> const &entries) const;
+
   bool
   createOutputWorkspace(size_t numHistMon, size_t numEventMon,
                         bool monitorsAsEvents,
--- a/Framework/DataHandling/inc/MantidDataHandling/LoadNexusMonitors2.h
+++ b/Framework/DataHandling/inc/MantidDataHandling/LoadNexusMonitors2.h
@@ -99,6 +99,9 @@ private:
                         std::map<int, std::string> &monitorNumber2Name,
                         std::vector<bool> &isEventMonitors);
 
+  bool isEventMonitor(::NeXus::File &monitorFileHandle) const;
+  std::size_t sizeOfUnopenedEntry(::NeXus::File &file,
+                                  const std::string &entryName) const;
   bool
   createOutputWorkspace(size_t numHistMon, size_t numEventMon,
                         bool monitorsAsEvents,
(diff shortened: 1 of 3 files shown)


Framework/DataHandling/src/LoadEventNexus.cpp
```diff
@@ -1153,9 +1153,10 @@ bool LoadEventNexus::hasEventMonitors() {
         break;
       }
     }
-    m_file->openData("event_id");
```
SimonHeybrock: Why is this deleted? Why are we still calling `closeGroup` a few lines further down, given that we do not open anymore?

SimonHeybrock (dismissed):
See one comment below.

I do not know enough about monitors in Nexus files, so I recommend having someone else review this as well.

Framework/DataHandling/src/LoadEventNexus.cpp
edwardb96: This is deleted because we added the call to `numEvents` which does the same thing in this case.
The call to `closeGroup` corresponds to `openGroup` on line 1152 not the call to `openData` in `numEvents`.

edwardb96:
@wdzhou Can you take a look at this and make sure it doesn't break anything for you?
@martyngigg Who else did you say should check this?

martyngigg:
@peterfpeterson Could you also run through some SNS data to check this is still okay your end?

Framework/DataHandling/inc/MantidDataHandling/LoadNexusMonitors2.h
```diff
@@ -99,6 +99,17 @@ class DLLExport LoadNexusMonitors2 : public API::ParallelAlgorithm {
                         std::map<int, std::string> &monitorNumber2Name,
                         std::vector<bool> &isEventMonitors);
 
+  bool isEventMonitor(::NeXus::File &monitorFileHandle) const;
+  std::size_t sizeOfUnopenedEntry(::NeXus::File &file,
+                                  const std::string &entryName) const;
+  bool eventIdNotEmptyIfExists(
+      ::NeXus::File &monitorFileHandle,
+      std::map<std::string, std::string> const &entries) const;
```
martyngigg: If we're going with "east const" then I think we should convert the whole file. Consistency at the file-level is still worth having I think...

martyngigg:
This fixes the issue for the bad ISIS files. @peterfpeterson @wdzhou Are you able to try out a few SNS files to confirm that is okay there?

peterfpeterson:
This is fine with SNS.

martyngigg (approved):
Discussion with others around `const` location lead to an attempt to use clang to work the whole framework.

This is high priority so let's get it in.

PR #21832, opened 2018-02-22
Fix bad background colour in Matrix View.

### LINKED ISSUE #21821: Fix Monitor Background Colour in the MantidPlot Matrix View
PR  #21668 causes the MatrixView to shade the cells containing monitor data with a black colour, the same colour as the text within them. The data table view should be modified to shade the cells with a lighter colour whilst keeping the tooltips visible.

## Platforms
Linux only.

edwardb96:
Fixes #21821 
Uses a hard-coded colour value instead of the tooltip colour for the background of the monitor rows in the matrix data view.

**To test:**
Ensure that the monitor cells in the Matrix Data View are readable on both Linux and Windows.

Fixes #21821 . 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidPlot/src/Mantid/MantidMatrixModel.cpp
+++ b/MantidPlot/src/Mantid/MantidMatrixModel.cpp
@@ -44,8 +44,7 @@ void MantidMatrixModel::setup(const Mantid::API::MatrixWorkspace *ws, int rows,
   m_colNumCorr = 1;
   m_endRow = m_rows - 1;
   m_startRow = start >= 0 ? start : 0;
-  m_mon_color =
-      QApplication::palette().color(QPalette::Active, QPalette::ToolTipBase);
+  m_mon_color = QColor(255, 253, 209);
   m_mask_color =
       QApplication::palette().color(QPalette::Disabled, QPalette::Background);
 


Mantid-Matthew:
This works fine on linux just building on windows

josephframsay (approved):
Monitors look good on Windows

MantidPlot/src/Mantid/MantidMatrixModel.cpp
```diff
@@ -44,8 +44,7 @@ void MantidMatrixModel::setup(const Mantid::API::MatrixWorkspace *ws, int rows,
   m_colNumCorr = 1;
   m_endRow = m_rows - 1;
   m_startRow = start >= 0 ? start : 0;
-  m_mon_color =
-      QApplication::palette().color(QPalette::Active, QPalette::ToolTipBase);
+  m_mon_color = QColor(255, 253, 209);
```
AndreiSavici: Can't we do better than hardcoded numbers?

edwardb96: I don't think there's a simple way to pass the original colour from [here](https://github.com/mantidproject/mantid/blob/master/MantidPlot/src/ApplicationWindow.cpp#L11561-L11585) to where we are now.

AndreiSavici: I would get the color of whatever is the background and change the color of the text in the cell to have ```
text_hue=(bg_hue+180)%360;
text_saturation=(bg_saturation+128)%256;
text_value=(bg_value+128)%256;
```
This idea is from https://doc.qt.io/archives/qq/qq26-adaptivecoloring.html

PR #21842, opened 2018-02-23
Live Spectrum Viewer Issue

### LINKED ISSUE #21574: Correct Tabbing / Comparison Behaviour in Spectrum Viewer
Follow-up to #21405.

As part of making the Spectrum Viewer available in Python so that developers at MLZ could integrate it with the NICOS control software, we found that the Spectrum Viewer had some inconsistent behaviour when handling multiple tabs and workspace updates.

Overview of current behaviour:

- Every time `renderWorkspace` is called, a new tab is created
- Even if a tab for a workspace already exists, calling `renderWorkspace` creates another tab
- Every time a workspace that has a tab is updated, another tab for the same workspace is created automatically
- The only way to create new tabs for a *different* workspace, that I could find, is to drag and drop it
- The horizontal and vertical graphs appear to try to render comparisons between the tabs, but the values are off
- For example, two tabs with the same data produce lines with the same shape, but off by thousands of counts
- Switching between tabs produces very inconsistent behaviour instead of showing the workspace data associated with that tab

Some thoughts and observations:

- The tabbing behaviour was added by PR #13670 in response to issue #10149
- The intent is to be able to compare *different* workspaces in different tabs
- This is a desirable feature, but currently broken and should be fixed
- Opening multiple tabs for the same workspace is not desirable and should be removed
- This includes manually calling `renderWorkspace` with the same workspace and automatic update on workspace being changed
- Only dragging and dropping should produce a new tab, and only if that workspace doesn't have one already

This issue is to correct all these problems and make Spectrum Viewer behave more consistently.

**Reproduce**
Add the path to Mantid to the PYTHONPATH. Then run something like the following:
```python
from threading import Thread
import time
import sys
import PyQt4
import mantid.simpleapi as simpleapi
import mantidqtpython as mpy
import os

SpectrumView = mpy.MantidQt.SpectrumView.SpectrumView
app = PyQt4.QtGui.QApplication(sys.argv)
wsOut = simpleapi.Load("INTER00013460.nxs")

sv = SpectrumView()
sv.renderWorkspace("wsOut")
# Second workspace should be rendered the same but is not
sv.renderWorkspace("wsOut")
sv.show()
app.exec_()
```

OwenArnold:
Fixes #21574 

Fixes ECP related bug described in 21574. 

After replace hook was used to call `renderWorkspace`, which is built for display of new workspaces. Dealing with the updates to the numerous objects involved in the SpectrumViewer was tricky, but this should be fully working.

This PR does not address the existing scaling issues described in #21574. However it would now allow end users to correctly visualise single updating MatrixWorkspace. I suggest that the scaling issue is investigated as part of another PR.

### DIFF
--- a/qt/widgets/spectrumviewer/inc/MantidQtWidgets/SpectrumViewer/SpectrumView.h
+++ b/qt/widgets/spectrumviewer/inc/MantidQtWidgets/SpectrumViewer/SpectrumView.h
@@ -4,6 +4,7 @@
 #include <QMainWindow>
 #include <QMdiSubWindow>
 #include <QList>
+#include <vector>
 
 #include "MantidAPI/MatrixWorkspace_fwd.h"
 #include "MantidQtWidgets/Common/WorkspaceObserver.h"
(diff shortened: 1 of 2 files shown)


MikeHart85 (approved):
Everything looks good to me, works as described.

Since this (intentionally) doesn't address all the problems in #21574, we should probably keep the issue open or create a new one for what is still left over.

PR #21844, opened 2018-02-23
SaveGSS shouldn't segfault with unwrittable path

peterfpeterson:
Change so SaveGSS errors out instead. 

Description of work.

**To test:**

There is an additional test and a fix so a code review should be sufficient.

*Does not need to be in the release notes* because it is an edge case that was discovered by the dev team.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/DataHandling/test/SaveGSSTest.h
+++ b/Framework/DataHandling/test/SaveGSSTest.h
@@ -115,6 +115,28 @@ public:
     AnalysisDataService::Instance().remove(wsName);
   }
 
+  void test_unwritablePath() {
+    if (Poco::File("/usr").exists()) {
+      // Save a 2 banks diffraction data with instrument using SLOG format
+      const std::string wsName = "unwritablePath";
+      auto dataws = generateTestMatrixWorkspace(wsName, m_defaultNumHistograms,
+                                                m_defaultNumBins);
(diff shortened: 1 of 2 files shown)


Framework/DataHandling/test/SaveGSSTest.h
```diff
@@ -115,6 +115,28 @@ class SaveGSSTest : public CxxTest::TestSuite {
     AnalysisDataService::Instance().remove(wsName);
   }
 
+  void test_unwritablePath() {
+    if (Poco::File("/usr").exists()) {
```
ElliotAOram: Unless I'm mistaken, this means this test will always be skipped on windows. Is this behaviour deliberate?

peterfpeterson: You are correct. I wasn't sure of a location on windows that both exists and isn't user writable. If you have a location, I can modify the test. I don't know if it is a big deal since it is run everywhere else.

ElliotAOram: Good point I can't say I know of an equivalent on windows myself either. I'm happy to approve this now I understand the reasoning.

ElliotAOram (approved):

PR #21845, opened 2018-02-23
MR reduction: allow passing a direct beam workspace

mdoucet:
The magnetism reflectometer reduction currently takes in a run number for the direct beam run.
This is inefficient for several reasons:

 - In an application like a UI where the reflectivity might be recalculated often, we don't want to reload the direct beam every time.

 - More importantly, if we have more than a single cross-section to calculate (which is 99% of the time for MR), you don't want to reload the direct beam for each cross-section.

We'll load the data once and pass the workspace as a parameter to the main workflow algorithm.

**To test:**
- Inspect the code.
- Verify that the system tests pass.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


ricleal (commented):
Tests passed. Code is readable. Looks good to me.

ricleal (approved):

PR #21871, opened 2018-02-26
Make sphinx optional again

peterfpeterson:
The developer site accidentally added a hard requirement of sphinx which broke the conda builds. This makes it optional again.

**To test:**

See if this branch can be build with conda. Alternatively, change the `buildconfig/CMake/FindSphinx.py` script to try to `import sphinxx` and see that cmake now completes and doesn't give the `docs-html` or `dev-docs-html`.

*Does not need to be in the release notes* because it is only for developers and should only be exercised with the conda builds.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


yxqd (approved):
Ran conda build (python 2) and it fixed the build problem.

PR #21874, opened 2018-02-26
Convert from byte array in py3

peterfpeterson:
#21847 generates the wrong text in python3. The result of the system call to git in python3 is a byte array that needs to be converted to a string. This does that in a way that should work in python2 and python3.

**To test:**

Build the docs and look at the "last modified" text on any algorithm. It should be sensible independent of what version of python you're using.

*Does not need to be in the release notes* because it is fixing an error in functionality introduced during this release cycle.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


ElliotAOram (approved):
Tested on Py27 and Py3 and it works as expected 👍

PR #21880, opened 2018-02-27
Performance optimisations for detector Banks

LamarMoore:
This PR is related to the overarching feature branch #21230 for refactoring the instrument view to use `ComponentInfo/DetectorInfo`.  These changes add optimisations for Banks of tubes, Rectangular banks and structured banks.

**To test:**
* Thorough Code Review
* Launch Instrument View and ensure that Rectangular banks (RectangularDetector e.g SANS2D, LOQ), Banks of Tubes (e.g WISH, LARMOR) and Structured banks (e.g LOKI) are loaded and displayed correctly. 

Fixes #21582 

**Release Notes** 
Release notes updated for 3.13

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)?
- [ ] Are the unit tests small and test the class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] Do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant (user and developer) documentation been added/updated?

Does everything look good? Mark the review as **Approve**. A member of `@mantidproject/gatekeepers` will take care of it.

PR #21881, opened 2018-02-27
Instrument View in-memory instrument refactor

### LINKED ISSUE #21230: Instrument View in-memory Instrument Refactor

LamarMoore:
Long-running feature branch which includes instrument view changes for refactoring the in-memory instrument. The following PRs have been tested and merged into this branch:

#21310 - renderer cleanup
#21525 - replace actors with componentInfo/DetectorInfo

NB there is another branch [here](https://github.com/mantidproject/mantid/pull/21880) with some further optimizations which will be merged after this branch.

**To test:**
- Ensure Mantid builds cleanly and test that a wide range of instruments are loaded correctly.

Fixes #21230. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Geometry/src/Instrument/ComponentInfo.cpp
```diff
@@ -107,8 +107,39 @@ ComponentInfo::componentsInSubtree(size_t componentIndex) const {
...
+  auto innerRangeDet = m_componentInfo->detectorRangeInSubtree(componentIndex);
+  auto nSubDetectors =
+      std::distance(innerRangeDet.begin(), innerRangeDet.end());
+  corners.nY = nSubDetectors / corners.nX;
+
+  corners.bottomLeft = *innerRangeDet.begin();
+  corners.topRight = corners.bottomLeft + nSubDetectors - 1;
+  corners.topLeft = corners.bottomLeft + (corners.nY - 1);
+  corners.bottomRight = corners.topRight - (corners.nY - 1);
+  return corners;
```
OwenArnold: Didn't we decide that this can give the wrong results for Rectangular banks where the id ordering is by row? I think it would be advisable to bring those changes across. Potentially, even to push a branch to master as part of another PR, as it's technically a bug affecting the upcoming release.

Framework/Geometry/inc/MantidGeometry/Rendering/ShapeInfo.h
```diff
@@ -0,0 +1,91 @@
...
+namespace Mantid {
+namespace Kernel {
+class V3D;
+}
+namespace Geometry {
+class RectangularDetector;
+class StructuredDetector;
+class IObjComponent;
+
+/** ShapeInfo : Stored Object related shape information for rendering.
```
OwenArnold: * Minor point. Description could be made clearer.
* This class looks like it has a creational/building job for other objects an attributes rather than `info` status? 
* Has it been checked that it is not possible to get into a bad state with this type? What happens if you call different setters on the same `ShapeInfo` object?

OwenArnold (changes requested):
Not a full review. Will do another pass soon.

Framework/Geometry/src/Instrument/ComponentInfo.cpp
LamarMoore: This is changed in the next PR with the optimisations [here](https://github.com/mantidproject/mantid/pull/21880)

Framework/Geometry/inc/MantidGeometry/Rendering/ShapeInfo.h
LamarMoore: * All members are overwritten with every setter so calling the setters multiple times should not result in invalid configurations, in the absence of concrete types for Cones, Spheres etc, there is no point forcing this in the constructors.
* I did think of `ShapeCreator` but although this object seems to create shapes, it actually stores information about the shape of an object which can be retrieved at the point of rendering. I cannot think of a better name.
*

Framework/Geometry/inc/MantidGeometry/Rendering/GeometryHandler.h
```diff
@@ -13,103 +16,88 @@ namespace Geometry {
...
 
 protected:
-  IObjComponent *ObjComp; ///< ObjComponent that uses this geometry handler
-  CSGObject *Obj;         ///< Object that uses this geometry handler
-  bool boolTriangulated;  ///< state of the geometry triangulation
-  bool
-      boolIsInitialized; ///< state of the geometry initialization for rendering
+  std::unique_ptr<detail::Renderer> m_renderer;
+  std::shared_ptr<detail::ShapeInfo> m_shapeInfo;
+  std::unique_ptr<detail::GeometryTriangulator> m_triangulator;
```
OwenArnold: Review use of smart pointers. If possible move to stack-allocated objects. You mentioned that these could be null, in which case leave as they are.

Framework/Geometry/inc/MantidGeometry/Rendering/GeometryTriangulator.h
```diff
@@ -0,0 +1,86 @@
...
+  File change history is stored at: <https://github.com/mantidproject/mantid>
+  Code Documentation is available at: <http://doxygen.mantidproject.org>
+*/
+class MANTID_GEOMETRY_DLL GeometryTriangulator {
+private:
+  bool m_isTriangulated;
+  size_t m_nFaces;
+  size_t m_nPoints;
+  std::vector<double> m_points; ///< double array or points
+  std::vector<int> m_faces;     ///< Integer array of faces
```
OwenArnold: I don't think you need any signed values. Would `uint16_t` be a better choice?

Framework/Geometry/inc/MantidGeometry/Rendering/GeometryTriangulator.h
```diff
@@ -0,0 +1,86 @@
...
+                    /// OpenCascade analysis of object surface
+  void OCAnalyzeObject();
+  size_t numPoints() const;
+  size_t numFaces() const;
+  void setupPoints();
+  void setupFaces();
+
+public:
+  /// Return OpenCascade surface.
+  boost::shared_ptr<TopoDS_Shape> getOCSurface();
```
OwenArnold: Is shared access really needed?

Framework/Geometry/inc/MantidGeometry/Rendering/GeometryTriangulator.h
```diff
@@ -0,0 +1,86 @@
...
+  size_t numTriangleVertices();
+  /// get a pointer to the 3x(NumberOfPoints) coordinates (x1,y1,z1,x2..) of
+  /// mesh
+  const std::vector<double> &getTriangleVertices();
+  /// get a pointer to the 3x(NumberOFaces) integers describing points forming
+  /// faces (p1,p2,p3)(p4,p5,p6).
+  const std::vector<int> &getTriangleFaces();
+#ifdef ENABLE_OPENCASCADE
+private:
+  boost::shared_ptr<TopoDS_Shape>
```
OwenArnold: See question below about shared access.

OwenArnold (changes requested):

Framework/Geometry/inc/MantidGeometry/Rendering/GeometryTriangulator.h
```diff
@@ -61,10 +62,10 @@ class MANTID_GEOMETRY_DLL GeometryTriangulator {
   const std::vector<double> &getTriangleVertices();
   /// get a pointer to the 3x(NumberOFaces) integers describing points forming
   /// faces (p1,p2,p3)(p4,p5,p6).
-  const std::vector<int> &getTriangleFaces();
+  const std::vector<uint32_t> &getTriangleFaces();
 #ifdef ENABLE_OPENCASCADE
 private:
-  boost::shared_ptr<TopoDS_Shape>
+  std::unique_ptr<TopoDS_Shape>
```
OwenArnold: One effect of this change is that the default behaviour of the `GeometryTriangulator` class would have now changed. `GeometryTriangulator` should not longer be copyable via an ordinary copy-constructor. As this has not caused a compiler error, I presume that this class in not meant to be copied, although it would have been possible previously.

This is very minor, but would it be worth stating that to some extent in the header, or marking the copy and copy-assignment operators as deleted? This could even be done in a follow on PR if there are other areas of the code that would benefit from this.

OwenArnold (approved):
Minor changes, would be happy if fixed in another PR if at all.

Framework/Geometry/inc/MantidGeometry/Rendering/GeometryTriangulator.h
LamarMoore: @OwenArnold  I deleted the copy constructor as part of [this](https://github.com/mantidproject/mantid/pull/21880) PR

martyngigg:
@LamarMoore There are some merge conflicts here now.

LamarMoore:
@martyngigg I expected this. I will deal with these now.

Framework/Geometry/src/Rendering/GeometryTriangulator.cpp
```diff
@@ -58,23 +59,37 @@ Kernel::Logger g_log("GeometryTriangulator");
...
+    generateMesh();
+    m_nPoints = m_meshObj->numberOfVertices();
+    m_nFaces = m_meshObj->numberOfTriangles();
+    m_points = m_meshObj->getVertices();
+    m_faces = m_meshObj->getTriangles();
+  }
   m_isTriangulated = true;
 }
 
+void GeometryTriangulator::generateMesh() {
```
LamarMoore: @KarlPalmen The changes in this PR consolidate the GeometryHandlers/Generators. The `GeometryTriangulator` is now responsible for generating the faces and vertices for triangulated surfaces. I have added this method to replace `MeshGeometryGenerator::Generate` which was previously unimplemented. We can discuss a strategy for added the mesh generation in if you want but I think it is straightforward and easy to see now that the code for all of this has been significantly compressed.

CMakeLists.txt
```diff
@@ -153,11 +153,12 @@ include_directories ( Framework/Kernel/inc )
 include_directories ( Framework/HistogramData/inc )
 include_directories ( Framework/Indexing/inc )
 include_directories ( Framework/Parallel/inc )
+include_directories ( Framework/Beamline/inc )
 include_directories ( Framework/Geometry/inc )
 include_directories ( Framework/API/inc )
 include_directories ( Framework/Types/inc )
 
-set ( CORE_MANTIDLIBS Kernel HistogramData Indexing Geometry API Types )
+set ( CORE_MANTIDLIBS Kernel HistogramData Indexing Beamline Geometry API Types )
```
martyngigg: Should `Beamline` be here? `CORE_MANTIDLIBS` gets linked with everything passed this point. Can we just link it in the projects that need it? I'd like to get rid of this variable entirely at some point.

LamarMoore: @martyngigg  `Beamline` is fundamental and is used by `Geometry` therefore it needs to be wherever `Geometry` is used.

LamarMoore: @martyngigg `Beamline` is used by `Geometry` so technically speaking it is core. It needs to be wherever `Geometry` is used so I could take it out but I think it makes as much sense and removing `Geometry` from `CORE_MANTIDLIBS` at this stage.

martyngigg: @LamarMoore Ok. Thanks for the explanation.

mantid-roman (changes requested):
Found a couple of issues:
1. 2D views do not fill the whole window probably because monitors are included. Example: MARI in Cylindrical X or Spherical X.
2. Adding a crystal peak gives an error. In a 2D view go to the Pick tab and select the `Add single crystal peak` tool. Click on a detector then click inside the mini-plot. Error dialog with the message: `invalid vector<T> subscript`.
3. When using the free drawing tool holding the right mouse button switches to the eraser. When you release the button a popup menu shows. It shouldn't.

LamarMoore:
@mantid-roman I was able to reproduce the 1 but not 2 and 3. What platform/instruments were you using?

mantid-roman:
Windows. MARI, WISH

LamarMoore:
@mantid-roman  I tried an another version of mantid and noticed that the issue in 3 is actually still there. If there are peaks selected while erasing something which is being drawn, a context menu appears which provides the option to clear peaks. Is this what you are seeing?

mantid-roman:
@LamarMoore Yes

LamarMoore:
@mantid-roman I think this is a feature as opposed to a bug. I can create a separate PR to deal with this separately if you believe this is incorrect functionality.

mantid-roman:
Maybe it is a feature. It was there before your changes.

mantid-roman:
@LamarMoore Those issues are gone. Do you know why WISH shows a grey infinite cylinder in 3D?

mantid-roman:
No infinite geometry any more. But I have found the instrument for which the side-by-side projection was developed - VISION. And unfortunately it crashes mantid.

mantid-roman (approved):

Framework/Geometry/src/Rendering/RenderingHelpers.cpp
```diff
@@ -0,0 +1,438 @@
...
+
+  glTexCoord2f(0.0, static_cast<GLfloat>(tex_frac_y));
+  pos = rectDet.getRelativePosAtXY(0, rectDet.ypixels() - 1);
+  pos += V3D(rectDet.xstep() * (-0.5), rectDet.ystep() * (+0.5),
+             0.0); // Adjust to account for the size of a pixel
+  glVertex3f(static_cast<GLfloat>(pos.X()), static_cast<GLfloat>(pos.Y()),
+             static_cast<GLfloat>(pos.Z()));
+
+  glEnd();
+  if (glGetError() > 0)
```
martyngigg: `cout` will be buried on Windows. Can we get more information in this case and send it to a logger?

LamarMoore: @martyngigg in the follow up [PR](https://github.com/mantidproject/mantid/pull/21880) these are completely removed. There is no sense fixing it for this PR.

martyngigg: Sounds good.

PR #21923, opened 2018-03-02
Stop the resetting of group/pair selection in muon analysis

### LINKED ISSUE #21812: Resetting of groups/pair selection in Multi Muon Fitting
In muon analysis if multiple fitting mode is selected the selection for the group/pair selection resets if the user leaves the data analysis tab. 

This is to change the code to keep the (valid subset) of the same selection, unless none of them are available in which case it will reset.

AnthonyLim23:
Fixes a bug which prevented the group/pair selection in muon analysis being kept when changing tabs. 

**To test:**
Load HIFI 108379 into muon analysis
go to the settings tab and make sure multiple fitting is ticked
go to the data analysis tab, the group/pair selection should be "all pairs"
Change the selection to all groups
go to the grouping tab and add a new group (just put "1" for the detectors)
go back to the grouping tab and it should now have the new group in the list of groups

change the groups selection to custom and only have long and the new group ticked
go to the home tab and press the back arrow (load previous run)
go to the data analysis tab and the group selection should be the same (long + new group)
change the group selection to all pairs

go to the home tab and load MUSR 62260 (it does not have any pairs)
go to the data analysis tab and all groups should be selected

Fixes #21812. 

**Release Notes** 

in release notes
---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/qt/scientific_interfaces/Muon/MuonAnalysis.cpp
+++ b/qt/scientific_interfaces/Muon/MuonAnalysis.cpp
@@ -367,14 +367,7 @@ void MuonAnalysis::setChosenGroupAndPeriods(const QString &wsName) {
   if (!periodToSet.isEmpty() && !periods.contains(periodToSet)) {
     m_uiForm.fitBrowser->setChosenPeriods(periodToSet);
   }
-  return; /*
-  const QString &groupToSet = QString::fromStdString(wsParams.itemName);
-
-const auto &groups = m_dataSelector->getChosenGroups();
-     if (!groups.contains(groupToSet)) {
-     m_uiForm.fitBrowser->setChosenGroup(groupToSet);
-     }
-*/
+  return; 
 }
 
 /**
--- a/qt/scientific_interfaces/Muon/MuonAnalysis.cpp
+++ b/qt/scientific_interfaces/Muon/MuonAnalysis.cpp
@@ -358,25 +358,23 @@ void MuonAnalysis::initLayout() {
 }
 
 void MuonAnalysis::setChosenGroupAndPeriods(const QString &wsName) {
-	const auto wsParams =
-		MuonAnalysisHelper::parseWorkspaceName(wsName.toStdString());
+  const auto wsParams =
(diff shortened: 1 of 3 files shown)


qt/scientific_interfaces/Muon/MuonAnalysis.cpp
```diff
@@ -361,16 +361,20 @@ void MuonAnalysis::setChosenGroupAndPeriods(const QString &wsName) {
...
   const auto &periods = m_dataSelector->getPeriodSelections();
-  if (!groups.contains(groupToSet)) {
-    m_uiForm.fitBrowser->setChosenGroup(groupToSet);
-  }
+
   if (!periodToSet.isEmpty() && !periods.contains(periodToSet)) {
     m_uiForm.fitBrowser->setChosenPeriods(periodToSet);
   }
+  return; /*
+  const QString &groupToSet = QString::fromStdString(wsParams.itemName);
```
NickDraper: Is this code commented out for a reason?

There is no note to explain this, in which case it should either be uncommented or deleted.

It will only cause confusion in the codebase laster.

NickDraper (changes requested):

NickDraper:
Another usability hitch that I noticed (may be another worth splitting into another issue, or it may be simple to solve)

In the custom groups selection dialog, it takes two mouse clicks to change any of the checkboxes, this is irritating and not what anyone wold expect.

qt/scientific_interfaces/Muon/MuonAnalysis.cpp
AnthonyLim23: I will delete this. I thought it was for periods and might be needed later.

AnthonyLim23:
just had a quick look at the clicking bug, it is present for all of the tickboxes the first time you click within a table. I think the first click is to select the cell and not the tickbox. I could change the code so that a cell click will connect to the tickbox

AnthonyLim23:
see #21941 for clicking checkboxes

NickDraper (approved):

PR #21930, opened 2018-03-02
Fix tooltip colours on RHEL7

### LINKED ISSUE #21921: Tooltip background colour is too dark on RHEL7
This was introduced by #21668 and #21832

Release is fine
tooltip_3 11

Nightly is unreadable
tooltip_nightly

edwardb96:
Fixes the colour pallete so that the hints are readable on RHEL7.

> On Linux (tested on Fedora 26-7 and Ubuntu 14.4) the palette colour for
> ToolTipBase has no effect on the colour of tooltips, but does change
> the colour of 'What's This' boxes and and Line Edit hints. The palette
> colour for ToolTipText on the other hand affects all three of these.
>
> The default pallete shows light text on a pale background which, although not
> affecting tooltips, makes LineEdit hints and 'What's This' boxes difficuilt
> if not impossible to read.
> 
> Changing the tooltip text to a darker colour fixes the problem for 'LineEdit'
> hints and 'What's This' boxes but creates one for ordinary tooltips.
>
> Setting the background colour to a darker colour works fine on
> Fedora 26-7+ and Ubuntu 14.04 but not for RHEL7 where the tooltip text
> colour is also dark.
>
>  At the moment the fix is to simply hardcode the values to a dark on
>  light colour scheme which works consistently on Fedora 26-7 and Ubuntu
>  14.04 where ToolTipBase has no effect

**To test:**
Check that the tool-tips, line edit hints and whats this help have appropriate colours.

Fixes #21921 . 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidPlot/src/ApplicationWindow.cpp
+++ b/MantidPlot/src/ApplicationWindow.cpp
@@ -2603,8 +2603,7 @@ void ApplicationWindow::exportMatrix() {
     QMessageBox::critical(
         this, tr("MantidPlot - Export error"), // Mantid
         tr("Could not write to file: <br><h4> %1 </h4><p>Please verify that "
-           "you have the right to write to this location!")
-            .arg(file_name));
+           "you have the right to write to this location!").arg(file_name));
     return;
   }
 
@@ -2971,8 +2970,8 @@ void ApplicationWindow::setAutoUpdateTableValues(bool on) {
     QList<MdiSubWindow *> folderWindows = f->windowsList();
     foreach (MdiSubWindow *w, folderWindows) {
       if (w->inherits("Table"))
-        dynamic_cast<Table *>(w)->setAutoUpdateValues(
-            d_auto_update_table_values);
+        dynamic_cast<Table *>(w)
+            ->setAutoUpdateValues(d_auto_update_table_values);
     }
     f = f->folderBelow();
   }
@@ -3575,12 +3574,11 @@ MantidTable *ApplicationWindow::convertTableToTableWorkspace(Table *t) {
   }
   std::string wsName = t->objectName().toStdString();
   if (Mantid::API::AnalysisDataService::Instance().doesExist(wsName)) {
-    if (QMessageBox::question(this, "MantidPlot",
-                              "Workspace with name " + t->objectName() +
-                                  " already exists\n"
-                                  "Do you want to overwrite it?",
-                              QMessageBox::Yes | QMessageBox::No) ==
-        QMessageBox::Yes) {
+    if (QMessageBox::question(
+            this, "MantidPlot", "Workspace with name " + t->objectName() +
+                                    " already exists\n"
+                                    "Do you want to overwrite it?",
+            QMessageBox::Yes | QMessageBox::No) == QMessageBox::Yes) {
       Mantid::API::AnalysisDataService::Instance().addOrReplace(wsName, tws);
     } else {
       return nullptr;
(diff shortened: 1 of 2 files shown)


edwardb96:
The diff is larger than it needs to be. I'm working from home today and I seem to have applied a newer version of clang-format. The only substantiative changes are at 11575-11595.

rosswhitfield (changes requested):
Yes it readable now but ugly
new

Can you invert the colours, does this still work for ubuntu/etc?

```diff
diff --git a/MantidPlot/src/ApplicationWindow.cpp b/MantidPlot/src/ApplicationWindow.cpp
index 5739e38..f693e60 100644
--- a/MantidPlot/src/ApplicationWindow.cpp
+++ b/MantidPlot/src/ApplicationWindow.cpp
@@ -11582,8 +11582,8 @@ void ApplicationWindow::setUpdateCurvesFromTable(Table *table, bool on) {
   14.04 where ToolTipBase has no effect.
 */
 void ApplicationWindow::patchPaletteForLinux(QPalette &palette) const {
-  auto tooltipBaseColor = QColor("black");
-  auto tooltipTextColor = QColor("white");
+  auto tooltipBaseColor = QColor("white");
+  auto tooltipTextColor = QColor("black");
 
   palette.setColor(QPalette::ColorGroup::Inactive, QPalette::ToolTipBase,
                    tooltipBaseColor);
```

swap

rosswhitfield:
Re-reading above I see that it won't work on Fedora 26-7 and Ubuntu 14.04, but we also don't care about those after this release.

But did we have this problem before? Does reverting #21668 and #21832 fix this? There must be a better solution. I can't approve white on black text.

edwardb96:
The problem existed before for Fedora and Ubuntu, #21668 fixed it but broke RHEL. Looking at it again #21832 is largely irrelevant it just fixed one of the problems created as a side effect off #21668. So reverting those PRs will fix for RHEL but break Fedora and Ubuntu again. What do you mean about not caring about Fedora and Ubuntu after this release, just those versions or are we dropping Ubuntu support?

rosswhitfield:
This is the last release that will support Ubuntu 14.04, 16.04 will still be supported, Fedora isn't officially supported at all.

edwardb96:
If I remember correctly the problem #21668 solved also existed in the more recent versions of Ubuntu. @gemmaguest Do you know the most recent version you saw this on?

gemmaguest:
I'm using Ubuntu 16.04 and have the problem of white text on yellow background in v3.11. If changing the background colour is problematic could we change the text colour on Ubuntu instead?

edwardb96:
We could change the text colour but then normal tooltips would show up as black text on a black background. 

- On Ubuntu (and Fedora) the background colour for 'tooltip like things' only applies to What's Thises and Line Edit Hints, not normal tooltips, normal tooltips are stuck as black colour, I believe this is technically a bug in the implementation of Qt rather than the way we have it configured.
- On Ubuntu (and Fedora) the text colour for 'tooltip like things' applies to What's Thises, Line Edit Hints **and** normal tooltips.

Bearing this in mind we have a few options.
1. Write some code to determine which linux distro the user has and set the colours based on what the system theme is likely to be.
2. Set the background of LineEdit Hints and What Thises to be dark to match the unchangeable dark background of normal tooltips, this allows us to set the tooltip text colour to be white.
The tooltips are then readable on all distros, if a little ugly on RHEL.
3. Let the text be unreadable one distro or the other.
4. Add a configuration option somewhere.

At the moment I have implemented option 2 because we're close to a release deadline, otherwise I'd consider option 1 or 4 but I'm not sure if it's worth it given MantidPlot is being replaced.

edwardb96:
Further investigation has revealed a workaround for Fedora is to tell mantid to use the `Adwaita` theme (instead of GTK+) from the adwaita-qt plugin. It also revealed that certain versions of gnome when running on ubuntu are not affected by the original issue. For now I propose to limit the scope of the original fix to Unity users who do not have the option to switch to the `Adwaita` theme without installing extra packages.

edwardb96:
@rosswhitfield and @gemmaguest please can you check that this now works for RHEL and Ubuntu respectively?

gemmaguest:
Looks good on Ubuntu 16.04 :+1:

rosswhitfield (approved):
LGTM

PR #21938, opened 2018-03-05
Fix a bug where threads don't exit

gemmaguest:
The reflectometry GUI creates many threads and was not calling exit() on them, so you could end up with several hundred threads hanging around.

**To test:**

- Run MantidPlot in a debugger
- Open the `ISIS Reflectometry` GUI
- Run a reduction with many rows, e.g. select the instrument `POLREF`, search for investigation `1520249`, select all the rows apart from the first 3, click `Transfer` and then click `Process`.
- After the reduction is finished, check how many threads there are still running. Before this fix it was 700+. After the fix it should be more like 10 or 15.

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

)


martyngigg (approved):
Looks good.

Confirmed that before the fix there were 768 threads alive after the reduction. After the fix I can see only 14 from various Qt-related functions.

PR #21946, opened 2018-03-05
Fix tiled window's drag and drop functionality

### LINKED ISSUE #21929: Drag and drop into tiled windows not working
Following these instructions tiled windows are not accepting drag and drop of other graphs.
http://www.mantidproject.org/MBC_Displaying_data_Complex 

I've found this in windows 8, Gemma found something similar during the NTC

### Expected behavior
A red box around the area for the graph to insert and then insertion of the graph into the tiled window

### Actual behavior
No interaction with drag and drop of graphs

We should fix this if it is easy to do so, otherwise we could consider disabling this form of Tiled window and removing it from the mantid basics course.

mantid-roman:
`QWidget::visibleRegion()` doesn't work the way this functionality (drag and drop) expects of it. In this PR the visible region is replaced by its bounding rect in the test for a point being inside the tiled window. This is not ideal as a tiled window hidden behind other windows can still accept drops. I don't see a way to fix this properly.

**To test:**

* Open a new tiled window (File->New->Tiled Window).
* Open some plots. The plots must be floating windows.
* Drag the plots onto the tiled window. They must be inserted into one of the empty cells.

Fixes #21929.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidPlot/src/ApplicationWindow.cpp
+++ b/MantidPlot/src/ApplicationWindow.cpp
@@ -16663,8 +16663,8 @@ TiledWindow *ApplicationWindow::getTiledWindowAtPos(QPoint pos) {
   foreach (QMdiSubWindow *w, wl) {
     TiledWindow *tw = dynamic_cast<TiledWindow *>(w->widget());
     if (tw) {
-      QPoint mdiOrigin = mapFromGlobal(pos);
-      auto r = w->visibleRegion();
+      QPoint mdiOrigin = mapToGlobal(w->pos());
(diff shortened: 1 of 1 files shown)


NickDraper:
The operation of the tiled windows remains a bit clumsy, but that is much better than not working at all.

We have removed the section of the Mantid basics course that refers people to these types of windows, and they will be implemented differently in the new workbench.

As such, there is little point investing more effort into this rarely used feature.

NickDraper (approved):
However, saying that the review of this change is approved.

PR #21974, opened 2018-03-07
Tie/fix/constrain in multi fitting

### LINKED ISSUE #21829: Tie/fix/constrain in multi fitting
In the multi fitting function browser (muon and general) adding a tie/constraint/fix will be added to all of the workspaces (even if it is not global)

mantid-roman:
Corrected update of the gui for local parameters. Added constraints to the local parameters.

**To test:**

* Start the multi fit interface. Add some data, function.
* Set ties, fixes, constraints to parameters of functions of different data sets.
* Switch between data sets. Check that the gui updates correctly.
* Run a few fits selecting different current data set each time. Results must be the same regardless of the selected data set.

Fixes #21829

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/qt/widgets/common/src/FunctionBrowser.cpp
+++ b/qt/widgets/common/src/FunctionBrowser.cpp
@@ -2240,8 +2240,8 @@ Mantid::API::IFunction_sptr FunctionBrowser::getGlobalFunction() {
       if (fun1->getConstraint(j)) {
         fun1->removeConstraint(parameterName);
       }
-      const auto &localParam = m_localParameterValues[parName][i];
       checkLocalParameter(parName);
+      const auto &localParam = m_localParameterValues[parName][i];
       if (localParam.fixed) {
         // Fix this particular local parameter
         fun1->setParameter(j, localParam.value);
--- a/qt/widgets/common/src/FunctionBrowser.cpp
+++ b/qt/widgets/common/src/FunctionBrowser.cpp
@@ -175,6 +175,8 @@ void FunctionBrowser::createBrowser() {
           SLOT(attributeVectorSizeChanged(QtProperty *)));
   connect(m_tieManager, SIGNAL(propertyChanged(QtProperty *)), this,
           SLOT(tieChanged(QtProperty *)));
+  connect(m_constraintManager, SIGNAL(propertyChanged(QtProperty *)), this,
+          SLOT(constraintChanged(QtProperty *)));
   connect(m_parameterManager, SIGNAL(valueChanged(QtProperty *, double)),
           SLOT(parameterChanged(QtProperty *)));
 
@@ -1079,6 +1081,18 @@ FunctionBrowser::addConstraintProperties(QtProperty *prop, QString constraint) {
     }
   }
 
+  
+  // In case of multi-dataset fitting store the tie for a local parameter
+  if (prop->hasOption(globalOptionName) &&
+      !prop->checkOption(globalOptionName)) {
+    auto parName = getParameterName(prop);
+    auto &localValues = m_localParameterValues[parName];
+    if (m_currentDataset >= localValues.size()) {
+      initLocalParameter(parName);
+    }
+    localValues[m_currentDataset].constraint = constraint;
+  }
+
   // add properties
   QList<FunctionBrowser::AProperty> plist;
   AConstraint ac;
@@ -1158,6 +1172,21 @@ bool FunctionBrowser::hasUpperBound(QtProperty *prop) const {
       return true;
   }
   return false;
+}
+
+/// Get a constraint string
+QString FunctionBrowser::getConstraint(const AConstraint& ac) const {
+  QString constraint;
+  if (ac.lower) {
+    constraint += m_constraintManager->value(ac.lower) + "<" +
+                  ac.paramProp->propertyName();
+  } else {
+    constraint += ac.paramProp->propertyName();
+  }
+  if (ac.upper) {
+    constraint += "<" + m_constraintManager->value(ac.upper);
+  }
+  return constraint;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


keeeto:
I'm happy to have a go at testing this too. Can anyone direct me to the multi fitting documentation?

mantid-roman:
@keeeto I am not sure if we have it

keeeto:
Ah, okay. No problem; I guess this is a job I cannot do then :). Maybe after the release is over I will have a go at learning the interface and writing some documentation. @mantid-roman would I be best to talk to you about this?

mantid-roman:
@keeeto Yes

soininen:
I found a problem when using CompositeFunction.
1. Get some data. I used `084446.nxs` from `ExternalData/Testing/Data/UnitTest/ILL/IN4`.
2. Open the Multi dataset fitting interface.
3. Add data. I used ws indices 10-14 from the IN4 workspace.
4. Add composite function.
5. Add FlatBackground to the composite function.
6. Constrain the A0 parameter to 10%.
7. Click Fit. Mantid crashes.

There is even a crash if I just try to change the value of A0...

soininen:
The problem with CompositeFunction seems to be gone now. However, there are some problems with ties: when I write some garbage, like `0.8*` in the tie input field, Mantid crashes.

Also, the tie field sometimes becomes disabled (grayed out) and I cannot modify the value anymore. This seems to happen when I try to write `0.` in the field.

soininen:
I should have been more precise: there is a crash when I right click a parameter, select `Add tie`, and write some garbage like `**` in the `Add a tie` dialog.

soininen:
The problems with the ties are now gone.

However, if I
1. Use the same dataset as above
2. Create a CompositeFunction
3. Add a FlatBackground
4. Add a Gaussian
5. Set the tie of the Sigma parameter to `f1.PeakCentre/25`
6. Click 'Fit'

```
Fit failed:

Unknown function parameter Height

This may happen if there is a CompositeFunction containing only one function
```

mantid-roman:
I am afraid this is unfixable, in this release at least. The message tells the user that using a composite function with one member is an error and they should either add more functions of do without the CompositeFunction.

soininen:
@mantid-roman There was two functions there in this case, FlatBackground and Gaussian, so I was a bit surprised with the error message.

mantid-roman:
That's not good :(

qt/widgets/common/src/FunctionBrowser.cpp
```diff
@@ -1912,47 +1953,72 @@ void FunctionBrowser::attributeVectorDoubleChanged(QtProperty *prop) {
...
-      return;
-    auto tie = QString::fromStdString(getTie(parProp));
-    auto tieExpr = tie.split('=');
-    if (tieExpr.size() == 2) {
-      tie = tieExpr[1];
+/// Called when a constraint property changes
+void FunctionBrowser::constraintChanged(QtProperty *prop) {
+  for (auto it = m_constraints.begin(); it != m_constraints.end(); ++it) {
+    bool isLower = it.value().lower == prop;
+    bool isUpper = it.value().upper == prop;
```
soininen: `const bool`?

qt/widgets/common/src/FunctionBrowser.cpp
```diff
@@ -1912,47 +1953,72 @@ void FunctionBrowser::attributeVectorDoubleChanged(QtProperty *prop) {
...
+    if (isLower || isUpper) {
+      auto paramProp = it.value().paramProp;
+      if (isLocalParameterProperty(paramProp)) {
+        auto parName = getParameterName(paramProp);
```
soininen: `const auto`?

qt/widgets/common/src/FunctionBrowser.cpp
```diff
@@ -2171,7 +2250,8 @@ Mantid::API::IFunction_sptr FunctionBrowser::getGlobalFunction() {
     multiFun->setDomainIndex(i, i);
     auto fun1 = multiFun->getFunction(i);
     for (size_t j = 0; j < fun1->nParams(); ++j) {
-      QString parName = QString::fromStdString(fun1->parameterName(j));
+      auto parameterName = fun1->parameterName(j);
+      QString parName = QString::fromStdString(parameterName);
```
soininen: Could both strings be `const`? It looks a bit strange that the same string is duplicated as C-string and `QString` here, but I guess it cannot be helped.

qt/widgets/common/src/FunctionBrowser.cpp
```diff
@@ -2181,19 +2261,32 @@ Mantid::API::IFunction_sptr FunctionBrowser::getGlobalFunction() {
...
+      if (localParam.fixed) {
         // Fix this particular local parameter
-        fun1->tie(parName.toStdString(),
-                  boost::lexical_cast<std::string>(
-                      getLocalParameterValue(parName, i)));
+        fun1->setParameter(j, localParam.value);
+        fun1->fix(j);
       } else {
-        auto tie = getLocalParameterTie(parName, i);
+        auto tie = localParam.tie;
```
soininen: `const auto &`?

qt/widgets/common/src/FunctionBrowser.cpp
```diff
@@ -1912,47 +1953,72 @@ void FunctionBrowser::attributeVectorDoubleChanged(QtProperty *prop) {
...
-    auto &paramValue = m_localParameterValues[parName][m_currentDataset];
-    if (paramValue.fixed)
-      return;
-    auto tie = QString::fromStdString(getTie(parProp));
-    auto tieExpr = tie.split('=');
-    if (tieExpr.size() == 2) {
-      tie = tieExpr[1];
+/// Called when a constraint property changes
+void FunctionBrowser::constraintChanged(QtProperty *prop) {
+  for (auto it = m_constraints.begin(); it != m_constraints.end(); ++it) {
```
soininen: Perhaps C++11 style `for` loop could be considered here:
```
for (const auto &constraint : m_constraints) {
```

soininen (changes requested):
Tested the changes with some data, and for simple functions, ties, fixes and constraints work fine. I get some error message with ties when CompositeFunction is involved but this is a known problem to my understanding.

The code changes look good, as well. I spotted just a few minor issues.

soininen:
The ties seem to work now with multiple functions in CompositeFunction. Good work!

soininen (approved):

PR #22007, opened 2018-03-08
Muon analysis handles 0 good frames gracefully

### LINKED ISSUE #22000: Multiperiod data not loading correctly into muon analysis
If the data has periods and is the current run then it cannot be loaded into muon analysis.

This is because the number of good frames is set to 0 until the run ends (for this case). As a result two problems occur:
1. If dead time corrections are used then the data will not load at all
2. If a group is being looked at then the asymmetry cannot be calculated (uses dead time)

The fixes are to
1. If number of good frames is 0 then to change the dead time correction drop down to none and disable the from data file option (until some other data is loaded)
2. Throw a warning and use 1 instead of 0

A scientist has provided a temporary data file to use as a 'current run'

AnthonyLim23:
Muon analysis would fail to load the current run if it had multiple periods. The changes will allow muon analysis to handle this gracefully. 

**To test:**
This needs someone at ISIS. The data is located

\\Olympic\Babylon5\Public\Steve Cottrell 

and is called EMUauto_E.tmp

Open muon analysis and make sure that the dead time correction is set to from file
Load  EMUauto_E.tmp (this is our fake current run with multiple periods). You will need to show all files to find it in the browser
It should give a warning but load the data
The dead time correction box is now none

Change from long to a group fwd
It should produce a warning in the mantid dialog box and produce some data

Go to the settings tab and change the finish value to 1 (makes the next bit easier)
Go back to the home page
Change the run to HIFI00108379
A warning should appear about restoring the dead time correction
Some data should appear and the dead time selection will be from file
Check that the data changes if you change the dead time selection to None

Fixes #22000. 

**Release Notes** 

In the release notes
---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/src/EstimateMuonAsymmetryFromCounts.cpp
+++ b/Framework/Algorithms/src/EstimateMuonAsymmetryFromCounts.cpp
@@ -97,8 +97,11 @@ void EstimateMuonAsymmetryFromCounts::exec() {
   double startX = getProperty("StartX");
   double endX = getProperty("EndX");
   const Mantid::API::Run &run = inputWS->run();
-  const double numGoodFrames = std::stod(run.getProperty("goodfrm")->value());
-
+  double numGoodFrames = std::stod(run.getProperty("goodfrm")->value());
+  if (numGoodFrames == 0) {
+	  g_log.warning("The data has no good frames, assuming a value of 1");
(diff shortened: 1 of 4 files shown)


NickDraper (approved):
All seems to work fine for me

PR #22008, opened 2018-03-08
Muon Sample log copying

### LINKED ISSUE #21885: sample logs and algorithms
The phaseQuad algorithm does not save the sample log to the output. This should be changed. 

MuonMaxent should also save the sample log to the output(s). See the version from the script repository for recent  changes.

ElliotAOram:
Ensure that the sample logs are copied from the MuonPhaseQuad input to the output.

**To test:**

* Load some data into muon analysis (MUSR 68389)
* Open the frequency domain anlysis GUI
* Change the workspace to phaseQuad
* Press the calculate button
* In the grouped workspace a PhaseQuad;FTT output has been added.
* Ensure this workspace has the same sample logs as the `MuonAnalysis` workspace.

Fixes #21885 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


ElliotAOram:
@AnthonyLim23 @NickDraper 
After investigating this further, this is caused by the reference file not having the instrument data - which it ideally should. Using a parent workspace on creation copies the instrument as well as the logs - hence the mismatch occuring.
The system test output also references a data-mismatch. This is misleading and is caused by this check in the system test:
```
# FFT of offset should have different phase
FFT(offset, Real=0, Imaginary=1, AcceptXRoundingErrors=True, AutoShift=True, 
    OutputWorkspace='OffsetFFTResults')
(result, _messages) = CompareWorkspaces('MuonFFTResults', 'OffsetFFTResults')
self.assertEqual(result, False)
```
I recommend we update the system test to increase the tolerance to actually ensure we catch errors! In addition, the data matches for all points (within a 1e-8, with the exception of the new 0 frequency reduced peak - a change that was added that slipped through the system tests a while back due to tolerance) tolerance so I suggest we update the reference file to include the instrument data and the new value at 0 Frequency.

NickDraper:
This works, although I ran into fit errors with the suggested data set

NickDraper (approved):

NickDraper:
Good to merge if the builds pass

martyngigg:
I realised after merging that this was marked for `3.12` but the target was set to `master`. I've cherry picked the commits on to the release branch.

ElliotAOram:
Thanks @martyngigg , that was an oversight on my part!

PR #22011, opened 2018-03-08
Prevent script window crash

### LINKED ISSUE #22009: Python script window execution causes MantidPlot to crash
The following Python scripts crash MantidPlot when executed via the script window:
This is a cleaner re-write of issue https://github.com/mantidproject/mantid/issues/21955.

A clearly not valid Python code that should give an error is (actually always crashes MantidPlot):
```Python
A = CreateSampleWorkspace()
B = A.getRun().getProperty('run_title')
A = CreateSampleWorkspace()
A.setTitle(B)
```

A possibly valid Python code is (actually crashes MantidPlot as well - maybe not directly, but consecutive or sequential further executions):
```Python
A = CreateSampleWorkspace()
B = A.getRun().getProperty('run_title')
A = CreateSampleWorkspace()
A.setTitle(B.value)
```

Experienced on Ubuntu 16.04, master

vreimund:
Fixes a bug. See issue(s) for more information.

**To test:**
Make sure this does not crash MantidPlot.

```
A = CreateSampleWorkspace()
B = A.getRun().getProperty('run_title')
A = CreateSampleWorkspace()
A.setTitle(B)
```

**Try to provoke a crash using the Python script window.**

Fixes #22009 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidPlot/src/PythonScript.cpp
+++ b/MantidPlot/src/PythonScript.cpp
@@ -647,7 +647,6 @@ bool PythonScript::executeString() {
   }
   // If an error has occurred we need to construct the error message
   // before any other python code is run
-  QString msg;
   if (!result) {
     emit_error();
     // If a script was aborted we both raise a KeyboardInterrupt and
@@ -658,9 +657,9 @@ bool PythonScript::executeString() {
   } else {
     emit finished(MSG_FINISHED);
     success = true;
-  }
-  if (isInteractive()) {
-    generateAutoCompleteList();
+    if (isInteractive() && !isExecuting()) {
+      generateAutoCompleteList();
+    }
   }
 
   Py_XDECREF(compiledCode);
--- a/MantidPlot/src/PythonScript.cpp
+++ b/MantidPlot/src/PythonScript.cpp
@@ -657,7 +657,7 @@ bool PythonScript::executeString() {
   } else {
     emit finished(MSG_FINISHED);
     success = true;
-    if (isInteractive() && !isExecuting()) {
+    if (isInteractive()) {
       generateAutoCompleteList();
     }
   }


MantidPlot/src/PythonScript.cpp
```diff
@@ -658,9 +657,9 @@ bool PythonScript::executeString() {
   } else {
     emit finished(MSG_FINISHED);
     success = true;
-  }
-  if (isInteractive()) {
-    generateAutoCompleteList();
+    if (isInteractive() && !isExecuting()) {
```
martyngigg: I think the `isExecuting` check here means that the code inside the `if` here will never run. I think `isExecuting` will only return false after `executeString()` has completed.

To demonstrate this first type `ws = CreateSampleWorkspace()` in to the script window and execute the code. Now on a new line start typing `ws.` and nothing happens. In the current version of Mantid the selection box of methods pops up, which is created by a call to `generateAutoCompleteList()`

martyngigg (changes requested):
I'm happy that we only update the auto-complete information on successful execution of a script.

martyngigg:
I think our main problem here is that we are hitting a real problem under the hood with how `getRun().getProperty` works. `getProperty` returns a bare pointer to a log value and when `A` is overwritten the pointer stored in `B` is no longer valid so any further accesses of this pointer will likely cause a segmentation fault.

I think your fix, without the `isExecuting` check, is still valid as it at least allows an error message to be populated.

The issue with `getProperty` is rather harder to solve and the workaround at the moment would be:
```
A = CreateSampleWorkspace()
a_title = A.getRun().getProperty('run_title').value
A = CreateSampleWorkspace()
A.setTitle(a_title)
```
so that `a_title` becomes a copy of the value in the property and having a dangling pointer in the script at all.

I suggest we try and tackle the *real* problem with `getPointer` in the next release as it is to big to put in now.

reimundILL:
Thank you for your careful considerations, I did not check that auto-completion still works throughout.

I will follow your advice to allow auto-completion and open an issue for the remaining bug.  I hope you follow this further since I will likely not have time to do some further investigations.

Some points that stay important for me are:
- this solution does (I was really happy) not produce any crashes in no multiple ways for me.

- the script
```Python
A = CreateSampleWorkspace()
B = A.getRun().getProperty('run_title')
A = CreateSampleWorkspace()
A.setTitle(B.value)
```
in the script interpreter does not produce crashes at all (on master)
- but this script will still crash Mantid in the future when run in the script window.

The script interpreter and the script window should behave the same way, the script window should be fixed first from my point of view.

martyngigg:
@reimundILL Thanks for re-enabling the auto completion. I agree that it's important we fix the underlying issue so that the program does not crash in any instance when accessing the log data in that way.

martyngigg (approved):

PR #21959, opened 2018-03-06
Mesh geometry support

OwenArnold:
Merge long-running feature branch for mesh support into mantid master now that the release 3.12 branch has been made

This PR should not be reviewed by any of the developers for the PRs that make up this changeset.

**To test:**

Code Review
Check builds pass

Issue #21848, opened 2018-02-26
Clean-up around MeshObject


OwenArnold:
Follow on from #12827.

Mainly removing unused code.

PR #22016, opened 2018-03-08
LoadSampleShape algorithm for STL files

### LINKED ISSUE #12828: LoadShape algorithm
Create an algorithm that will load an STL sample geometry file and assign that geometry to the sample, or any named instrument component.

Example algorithm Dialog:

new mockup 1

sample files can be found here
http://files.mantidproject.org/videos/StlFiles/

needs
- [x] #12827

KarlPalmen:
Description of work.

Creation of new algorithm to load and STL file into a MeshObject and place the resulting shape into the sample of the output workspace.

**To test:**

* Careful code review (issue #12828 only).
* Check that unit tests work and are adequate.
* Run the algorithm in Mantid to see that the dialog box is OK and it doesn't crash.
* Build and check the documentation.

Fixes #12828. 

**Release Notes** 
In New Algorithms of Framework in the release notes of 3.13.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


KarlPalmen:
This pull request has been put to master rather than the base branch of #12828's branch as I expected. This issue needs to be addressed before merging.

KarlPalmen:
The base branch has been merged into master. #21959

OwenArnold:
@KarlPalmen there a lot of commits listed in this PR. Have you merged origin/master into your branch recently?

KarlPalmen:
No I have not merged master into this branch. Many of the commits were inherited from the mesh support branch I based this branch upon and so are not part of the issue.

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,690 @@
...
+namespace Geometry {
+
+using Kernel::Material;
+using Kernel::V3D;
+using Kernel::Quat;
+
+/**
+*  Default constuctor
+*/
+MeshObject::MeshObject() {} // Should never be called
```
matthew-d-jones: Use `= default`

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,690 @@
...
+*/
+bool MeshObject::isValid(const Kernel::V3D &point) const {
+
+  BoundingBox bb = getBoundingBox();
+  if (!bb.isPointInside(point)) {
+    return false;
+  }
+
+  V3D direction(0.0, 0.0, 1.0); // direction to look for intersections
+  std::vector<V3D> intesectionPoints;
```
matthew-d-jones: typo in variable name, should be `intersectionPoints`

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,690 @@
...
+  std::vector<int> entryExitFlags;
+
+  getIntersections(point, direction, intesectionPoints, entryExitFlags);
+
+  if (intesectionPoints.size() == 0) {
```
matthew-d-jones: Use `intersectionPoints.empty()` to test for emptiness.

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,690 @@
...
+    return false;
+  }
+
+  // True if point is on surface
+  for (size_t i = 0; i < intesectionPoints.size(); ++i) {
```
matthew-d-jones: Use range-based for loop

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,690 @@
...
+*/
+bool MeshObject::isOnSide(const Kernel::V3D &point) const {
+
+  BoundingBox bb = getBoundingBox();
+  if (!bb.isPointInside(point)) {
+    return false;
+  }
+
+  std::vector<V3D> directions; // directions to look for intersections
+  directions.push_back(V3D(0, 0, 1));
```
matthew-d-jones: Use `emplace_back`

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,690 @@
...
+  directions.push_back(V3D(1, 0, 0));
+  // We have to look in several directions in case a point is on a face
+  // or edge parallel to the first direction or also the second direction.
+  for (size_t i = 0; i < directions.size(); ++i) {
+    std::vector<V3D> intesectionPoints;
+    std::vector<int> entryExitFlags;
+
+    getIntersections(point, directions[i], intesectionPoints, entryExitFlags);
+
+    if (intesectionPoints.size() == 0) {
```
matthew-d-jones: `.empty()`

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,690 @@
...
+    return false;
+  }
+
+  std::vector<V3D> directions; // directions to look for intersections
+  directions.push_back(V3D(0, 0, 1));
+  directions.push_back(V3D(0, 1, 0));
+  directions.push_back(V3D(1, 0, 0));
+  // We have to look in several directions in case a point is on a face
+  // or edge parallel to the first direction or also the second direction.
+  for (size_t i = 0; i < directions.size(); ++i) {
```
matthew-d-jones: Use range-based for loop

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,690 @@
...
+    std::vector<int> entryExitFlags;
+
+    getIntersections(point, directions[i], intesectionPoints, entryExitFlags);
+
+    if (intesectionPoints.size() == 0) {
+      return false;
+    }
+
+    size_t k = intesectionPoints.size();
+    for (size_t i = 0; i < k; ++i) {
```
matthew-d-jones: You are already using `i` in the outer loop, but this would be solved by making the outer loop range based.

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,690 @@
...
+  if (!bb.doesLineIntersect(UT)) {
+    return 0;
+  }
+
+  std::vector<V3D> intesectionPoints;
+  std::vector<int> entryExitFlags;
+
+  getIntersections(UT.startPoint(), UT.direction(), intesectionPoints,
+                   entryExitFlags);
+  if (intesectionPoints.size() == 0)
```
matthew-d-jones: `.empty()`

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,690 @@
...
+  a = edge1.scalar_prod(h);
+
+  const double EPSILON = 0.0000001 * edge1.norm();
+  if (a > -EPSILON && a < EPSILON)
+    return false; // Ray in or parallel to plane of triangle
+  f = 1 / a;
+  s = start - v1;
+  u = f * (s.scalar_prod(h));
+  if (u < 0.0 || u > 1.0)
+    return false; // Intesection with plane outside triangle
```
matthew-d-jones: typo in "Intersection"

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,690 @@
...
+                                double &zmin) const {
+  BoundingBox bb = getBoundingBox();
+  xmin = bb.xMin();
+  xmax = bb.xMax();
+  ymin = bb.yMin();
+  ymax = bb.yMax();
+  zmin = bb.zMin();
+  zmax = bb.zMax();
+
+  return;
```
matthew-d-jones: Remove redundant `return`

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,690 @@
...
+* @param scaleFactor :: V3D giving scaling of the object
+* @return :: estimate of solid angle of object.
+*/
+double MeshObject::solidAngle(const Kernel::V3D &observer,
+                              const Kernel::V3D &scaleFactor) const
+
+{
+  std::vector<V3D> scaledVertices;
+  scaledVertices.reserve(m_vertices.size());
+  for (size_t i = 0; i < m_vertices.size(); ++i) {
```
matthew-d-jones: Use range-based for loop

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,690 @@
...
+    scaledVertices.push_back(V3D(scaleFactor.X() * m_vertices[i].X(),
```
matthew-d-jones: Use `emplace_back`

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,690 @@
...
+    // bounding box, if the cached bounding box is null.
+    if (numberOfVertices() > 0) {
+      // Initial extents to be overwritten by loop
+      constexpr double huge = 1e10;
+      double minX, maxX, minY, maxY, minZ, maxZ;
+      minX = minY = minZ = huge;
+      maxX = maxY = maxZ = -huge;
+
+      // Loop over all vertices and determine minima and maxima on each axis
+      for (size_t i = 0; i < m_vertices.size(); ++i) {
```
matthew-d-jones: Use range-based for loop

Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -0,0 +1,690 @@
...
+* Try to find a point that lies within (or on) the object, given a seed point
+* @param point :: on entry the seed point, on exit point in object, if found
+* @return true if point found
+*/
+bool MeshObject::searchForObject(Kernel::V3D &point) const {
+  //
+  // Method - check if point in object, if not search directions along
+  // principle axes using interceptSurface
+  //
+  Kernel::V3D testPt;
```
matthew-d-jones: `testPt` isn't used.

Framework/Geometry/src/Instrument/SampleEnvironmentSpecParser.cpp
```diff
@@ -222,16 +223,18 @@ Mantid::Geometry::SampleEnvironmentSpecParser::parseComponent(
...
   if (iter != m_materials.end()) {
-    comp->setMaterial(iter->second);
+    mat = iter->second;
   } else {
     throw std::runtime_error("SampleEnvironmentSpecParser::parseComponent() - "
                              "Unable to find material with id=" +
                              materialID);
   }
+  comp->setID(element->getAttribute("id"));
+  comp->setMaterial(mat);
```
matthew-d-jones: Why the change from `comp->setMaterial(iter->second);`? It doesn't look like there is a reason for `mat` to be in function scope.

Framework/DataHandling/src/LoadShape.cpp
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  std::string filename = getProperty("Filename");
+  std::ifstream file(filename.c_str());
+  if (!file) {
+    g_log.error("Unable to open file: " + filename);
+    throw Exception::FileError("Unable to open file: ", filename);
+  }
+
+  std::string solidName = "";
```
matthew-d-jones: Initialisation is redundant, just use `std::string solidName;`

Framework/DataHandling/src/LoadShape.cpp
```diff
@@ -0,0 +1,214 @@
...
+    // Add triangle if all 3 vertices are distinct
+    if (!areEqualVertices(t1, t2) && !areEqualVertices(t1, t3) &&
+        !areEqualVertices(t2, t3)) {
+      triangleIndices.push_back(addSTLVertex(t1, vertices));
+      triangleIndices.push_back(addSTLVertex(t2, vertices));
+      triangleIndices.push_back(addSTLVertex(t3, vertices));
+    }
+  }
+  // Use efficient constructor of MeshObject
+  std::unique_ptr<MeshObject> retVal = std::unique_ptr<MeshObject>(
```
matthew-d-jones: Use `Mantid::Kernel::make_unique`?

Framework/DataHandling/src/LoadShape.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+// Adds vertex to list if distinct and returns index to vertex added or equal
+uint16_t LoadShape::addSTLVertex(V3D &vertex, std::vector<V3D> &vertices) {
+  for (uint16_t i = 0; i < vertices.size(); ++i) {
+    if (areEqualVertices(vertex, vertices[i])) {
+      return i;
+    }
+  }
+  vertices.push_back(vertex);
+  uint16_t index = static_cast<uint16_t>(vertices.size() - 1);
```
matthew-d-jones: Use `auto`, the type is already stated in the cast.

matthew-d-jones (commented):
Here's a quick review of the code, I think it would be good to give Owen opportunity to check it over from the design perspective before this is merged.

Framework/Geometry/test/MeshObjectTest.h
```diff
@@ -0,0 +1,1224 @@
...
+  vertices.push_back(centre + V3D(max, min, max));
+  vertices.push_back(centre + V3D(min, min, max));
+  vertices.push_back(centre + V3D(max, max, min));
+  vertices.push_back(centre + V3D(min, max, min));
+  vertices.push_back(centre + V3D(max, min, min));
+  vertices.push_back(centre + V3D(min, min, min));
+
+  std::vector<uint16_t> triangles;
+  // top face of cube - z max
+  triangles.push_back(0);
```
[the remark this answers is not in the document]
KarlPalmen: I never knew one could insert several elements at a time (in an array).

KarlPalmen:
@OwenArnold has made changes to MeshObject, so I will deal with that after his changes have been merged in. These changes include at least one of @matthew-d-jones requests. I'll deal with other files tomorrow.

KarlPalmen:
The issue in which @OwenArnold cleaned up MeshObject appears to be #21848 and @matthew-d-jones did much of the work there.

KarlPalmen:
I have addressed @matthew-d-jones review.

Framework/API/src/Sample.cpp
```diff
@@ -363,7 +363,11 @@ int Sample::loadNexus(::NeXus::File *file, const std::string &group) {
     }
     Kernel::Material material;
     material.loadNexus(file, "material");
-    m_shape->setMaterial(material);
+    // CSGObject expected, if so, set its material
+    if (auto csgObj =
```
OwenArnold: `ShapeFactory` should already be creating a `CSGObject` for you. All you have to do is assign that to a temporary, set the material (no dynamic cast required), and then set `m_shape` from the temporary.

Framework/Algorithms/test/MonteCarloAbsorptionTest.h
```diff
@@ -52,11 +52,13 @@ void addSample(Mantid::API::MatrixWorkspace_sptr ws,
...
 
     ShapeFactory shapeMaker;
-    auto can = boost::make_shared<Container>(
+    auto canShape =
         shapeMaker.createShape(ComponentCreationHelper::cappedCylinderXML(
-            radius, height, baseCentre, axis, id)));
-    can->setMaterial(Material("CanMaterial",
-                              PhysicalConstants::getNeutronAtom(26, 0), 0.01));
+            radius, height, baseCentre, axis, id));
+    // Set material assuming it's a CSG Object
```
OwenArnold: I don't think you need this comment. There is no assumption. `createShape` will only return a CSGObject

Framework/Algorithms/test/MonteCarloTesting.h
```diff
@@ -90,15 +90,24 @@ inline Mantid::API::Sample createSamplePlusContainer() {
   const double height(0.05), innerRadius(0.0046), outerRadius(0.005);
   const V3D centre(0, 0, -0.5 * height), upAxis(0, 0, 1);
   // Container
-  auto can = boost::make_shared<Container>(ShapeFactory().createShape(
-      annulusXML(innerRadius, outerRadius, height, upAxis)));
-  can->setMaterial(Material("Vanadium", getNeutronAtom(23), 0.02));
+  auto canShape = ShapeFactory().createShape(
+      annulusXML(innerRadius, outerRadius, height, upAxis));
+  // CSG Object assumed
+  if (auto csgObj =
```
OwenArnold: No dynamic_cast needed.

Framework/Algorithms/test/MonteCarloTesting.h
```diff
@@ -90,15 +90,24 @@ inline Mantid::API::Sample createSamplePlusContainer() {
...
+  }
+  auto can = boost::make_shared<Container>(canShape);
   auto environment =
       Mantid::Kernel::make_unique<SampleEnvironment>("Annulus Container", can);
   // Sample volume
   auto sampleCell = ComponentCreationHelper::createCappedCylinder(
       innerRadius, height, centre, upAxis, "sample");
-  sampleCell->setMaterial(Material("Si", getNeutronAtom(14), 0.15));
+  // CSG Object assumed
+  if (auto csgObj = boost::dynamic_pointer_cast<Mantid::Geometry::CSGObject>(
```
OwenArnold: No dynamic_cast needed. No downcasting todo.

Framework/Algorithms/test/MonteCarloTesting.h
```diff
@@ -130,7 +139,11 @@ inline Mantid::API::Sample createTestSample(TestSampleType sampleType) {
     } else {
       throw std::invalid_argument("Unknown testing shape type requested");
     }
-    shape->setMaterial(Material("Vanadium", getNeutronAtom(23), 0.02));
+    // CSG Object assumed
+    if (auto csgObj =
```
OwenArnold: same. No dynamic cast.

Framework/DataHandling/inc/MantidDataHandling/LoadShape.h
```diff
@@ -0,0 +1,83 @@
+#ifndef DATAHANDLING_LOAD_SHAPE_H_
+#define DATAHANDLING_LOAD_SHAPE_H_
+
+#include "MantidAPI/IFileLoader.h"
+#include "MantidGeometry/Objects/MeshObject.h"
+#include "MantidKernel/V3D.h"
+
+namespace Mantid {
+namespace DataHandling {
+/**  Load Shape into an instrument of a workspace
```
OwenArnold: Load a shape for the sample on the an experiment info of the workspace. More detail that that should be provided.

Framework/DataHandling/inc/MantidDataHandling/LoadShape.h
```diff
@@ -0,0 +1,83 @@
...
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+File change history is stored at: <https://github.com/mantidproject/mantid>.
+Code Documentation is available at: <http://doxygen.mantidproject.org>
+ */
+
+class DLLExport LoadShape : public API::IFileLoader<Kernel::FileDescriptor> {
```
OwenArnold: This algorithm should be called `LoadSampleShape`. That is more accurate. Or even `LoadSTLSampleShape`. I think this algorithm has adopted a more generic algorithm name that should be reserved for future use.

Framework/DataHandling/inc/MantidDataHandling/LoadShape.h
```diff
@@ -0,0 +1,83 @@
...
+  std::unique_ptr<Geometry::MeshObject> readSTLSolid(std::ifstream &file,
+                                                     std::string &name);
+  std::unique_ptr<Geometry::MeshObject> readSTLMeshObject(std::ifstream &file);
+  bool readSTLTriangle(std::ifstream &file, Kernel::V3D &v1, Kernel::V3D &v2,
+                       Kernel::V3D &v3);
+  bool readSTLVertex(std::ifstream &file, Kernel::V3D &v);
+  bool readSTLLine(std::ifstream &file, std::string const &linetype);
+  uint16_t addSTLVertex(Kernel::V3D &vertex,
+                        std::vector<Kernel::V3D> &vertices);
+  bool areEqualVertices(Kernel::V3D &v1, Kernel::V3D &v2) {
```
OwenArnold: The math looks fine, but...
1. Does not need to be a member function. Make it a non-member.
2. Does not need to be declared the header at all, move to an anonymous namespace in the implementation file.
3. Tolerance needs some explaining. Why the absolute value 1e-6?

Framework/DataHandling/src/LoadShape.cpp
```diff
@@ -0,0 +1,214 @@
...
+// Register the algorithm into the algorithm factory
+DECLARE_ALGORITHM(LoadShape)
+
+using namespace Kernel;
+using namespace API;
+using namespace Geometry;
+
+void LoadShape::init() {
+  auto wsValidator = boost::make_shared<CompositeValidator>();
+  wsValidator->add<API::InstrumentValidator>();
```
OwenArnold: Why have a composite validator for one rule?

Framework/DataHandling/src/LoadShape.cpp
```diff
@@ -0,0 +1,214 @@
...
+  declareProperty(
+      make_unique<WorkspaceProperty<>>("InputWorkspace", "", Direction::Input,
+                                       wsValidator),
+      "The name of the workspace containing the instrument to add the shape");
+
+  // shape file
+  const std::vector<std::string> extensions{".stl"};
+  declareProperty(
+      make_unique<FileProperty>("Filename", "", FileProperty::Load, extensions),
+      "The name of the file containing the shape. ");
```
OwenArnold: It's not just the name of the file. It's also the path to the file. And that file is an stl file.

Framework/DataHandling/src/LoadShape.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+  // Output workspace
+  declareProperty(make_unique<WorkspaceProperty<>>("OutputWorkspace", "",
+                                                   Direction::Output),
+                  "The name of the workspace that will be same as "
```
OwenArnold: I don't think you are enforcing that, and I don't see that you should or can. Not a good idea to manipulate the ADS within an algorithm execution method.

Also, since this algorithm is NOT modifying the data, either it should clone the instrument from the first workspace (to leave it intact), or the workspace should be specified as `InOut` to indicate that the input geometry is going to be modified.

Framework/Geometry/inc/MantidGeometry/Objects/CSGObject.h
```diff
@@ -77,20 +77,26 @@ class MANTID_GEOMETRY_DLL CSGObject : public IObject {
   /// Clone
   IObject *clone() const override { return new CSGObject(*this); }
 
+  IObject *cloneWithMaterial(const Kernel::Material &material) const override {
```
OwenArnold: The return should be **Covarient** otherwise you need dynamic_casts everywhere and the code becomes a total mess.

Framework/Geometry/inc/MantidGeometry/Objects/MeshObject.h
```diff
@@ -0,0 +1,209 @@
...
+  MeshObject(const MeshObject &) = delete;
+  /// Assignment operator
+  MeshObject &operator=(const MeshObject &) = delete;
+  /// Destructor
+  virtual ~MeshObject();
+  /// Clone
+  IObject *clone() const override {
+    return new MeshObject(m_triangles, m_vertices, *m_material);
+  }
+  IObject *cloneWithMaterial(const Kernel::Material &material) const override {
```
OwenArnold: Same as above, should return `MeshObject*`

Framework/Geometry/inc/MantidGeometry/Objects/MeshObject.h
```diff
@@ -0,0 +1,209 @@
...
+             const Kernel::Material &material);
+
+  /// Copy constructor
+  MeshObject(const MeshObject &) = delete;
+  /// Assignment operator
+  MeshObject &operator=(const MeshObject &) = delete;
+  /// Destructor
+  virtual ~MeshObject();
+  /// Clone
+  IObject *clone() const override {
```
OwenArnold: Should return `MeshObject*`

Framework/Geometry/inc/MantidGeometry/Objects/MeshObject.h
```diff
@@ -0,0 +1,209 @@
...
+  int numberOfVertices() const;
+  double *getVertices() const;
+  int numberOfTriangles() const;
+  int *getTriangles() const;
+
+  void updateGeometryHandler();
+
+private:
+  /// Default constructor - should never be called
+  MeshObject();
```
OwenArnold: Delete this constructor. You do not need it.

OwenArnold (changes requested):
First round of review changes

KarlPalmen:
I see most of @OwenArnold's review refers to out of date code. I'm merged in code from master to address such issues. I'll work on the parts of the review that refer to up-to-date code, when I am ready.

Framework/DataHandling/inc/MantidDataHandling/LoadShape.h
KarlPalmen: What details do you want adding? Which questions would you want these details to answer?

OwenArnold:
@KarlPalmen I started the review and got interrupted by some meetings. Please look at the comments.

Framework/DataHandling/inc/MantidDataHandling/LoadShape.h
KarlPalmen: I take it you are referring `areEqualVertices` only.
I decided to keep it simple and inline. Making it a non-member would require it being moved to the `.cpp` file.  The value of the tolerance can be easily changed if a more appropriate value is found.  If this needs to be calculated from the shape, then it could be moved to the `.cpp` file.

Framework/DataHandling/src/LoadShape.cpp
KarlPalmen: I'll mention it's a path name to the file. I won't mention it's an STL file here, because this is mentioned elsewhere in the documentation and is subject to change (e.g. OFF file will be added).

Framework/DataHandling/inc/MantidDataHandling/LoadShape.h
OwenArnold: - What makes 1e-6 a good value for this? Why is the precision not greater, say 1e-9?
- Regarding keeping it inline, are you seeing measurable performance issues? what justifies having this method inline?
Also, looking at this signature again, the arguments should be passed as const-ref not ref, and the method should be declared const in it's current state.

Framework/DataHandling/src/LoadShape.cpp
OwenArnold: Ok. if OFF is going to be added to the same algorithm (not another), then this makes sense.

Framework/DataHandling/inc/MantidDataHandling/LoadShape.h
KarlPalmen: I'll try 1e-9.

Framework/DataHandling/src/LoadShape.cpp
KarlPalmen: If the algorithm were for STL files only, it would have `STL` in its name.

Framework/DataHandling/inc/MantidDataHandling/LoadSampleShape.h
```diff
@@ -0,0 +1,84 @@
...
+    return "DataHandling\\Instrument";
+  }
+  /// Returns a confidence value that this algorithm can load a file
+  int confidence(Kernel::FileDescriptor &descriptor) const override;
+
+private:
+  // Implement abstract Algorithm methods
+  void init() override;
+  void exec() override;
+  std::unique_ptr<Geometry::MeshObject> readSTLSolid(std::ifstream &file,
```
OwenArnold: All of the methods except for `init` and `exec` in this private section do not access or use any member variable from the class `LoadSampleShape`, or any of its public base classes. It therfore does not make sense to have them as part of the class. Please refactor them as such to be non-members so that the class API is cleaner and accurate.

KarlPalmen:
I've addressed @OwenArnold 's review.

Framework/DataHandling/test/LoadSampleShapeTest.h
```diff
@@ -0,0 +1,211 @@
...
+    clearWorkspaces("InputWS", "InputWS");
+  }
+
+private:
+  // Create workspaces and add them to algorithm properties
+  void prepareWorkspaces(LoadSampleShape &alg, const std::string &inputWS,
+                         const std::string &outputWS) {
+    auto inWs =
+        WorkspaceCreationHelper::create2DWorkspaceWithFullInstrument(10, 4);
+    Mantid::API::AnalysisDataService::Instance().add(inputWS, inWs);
```
OwenArnold: The use of singletons (here the `ADS`), can and should be avoided here. You do not need to put the workspace in the ADS and do not need to clear it out if you set the algoritm...

```cpp
alg.setChild(true);
```
Then you can simply set and fetch the Workspace directly, i.e.
```cpp
alg.setProperty("InputWorkspace", inputWS);
...
MatrixWorkspace_sptr outWS = alg.getProperty("OutputWorkspace");
```
Please refactor the use of the ADS out of this test suite.

Framework/DataHandling/test/LoadSampleShapeTest.h
```diff
@@ -0,0 +1,211 @@
...
+    return getMeshObject(outputWS);
+  }
+
+  const MeshObject *getMeshObject(const std::string &outputWS) {
+    MatrixWorkspace_sptr ws = getWorkspace(outputWS);
+    Sample s;
+    TS_ASSERT_THROWS_NOTHING(s = ws->sample());
+    auto &obj = s.getShape();
+    auto mObj = dynamic_cast<const MeshObject *>(&obj);
+    TSM_ASSERT_DIFFERS("Shape is not a mesh object", mObj, nullptr);
```
OwenArnold: Good test. I think `TSM_ASSERT("Shape is not a mesh object", mObj != nullptr);` as a minor refactor would be better. We only want to know if it is null or not.

Framework/DataHandling/test/LoadSampleShapeTest.h
```diff
@@ -0,0 +1,211 @@
...
+    return mObj;
```
OwenArnold: Be careful! think about what you have done here. You have taken a reference and have tried to extended it's lifetime beyond the scope of the function in which it was obtained by taking it's address as a pointer and returning it. You only get away with this because the current internal implementation of `Sample` stores shape as a field and it is therefore tied to the lifetime of `Sample` [see](https://github.com/mantidproject/mantid/blob/master/Framework/API/src/Sample.cpp#L106) and only because the same thing happens in `ExperimentInfo` [see](https://github.com/mantidproject/mantid/blob/master/Framework/API/src/ExperimentInfo.cpp#L604).

This is not good. You need to do whatever you need to with the `MeshObject` within the same scope in which you obtain it.

Talk to me if this is not clear.

Framework/DataHandling/test/LoadSampleShapeTest.h
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  MatrixWorkspace_sptr getWorkspace(const std::string &outputWS) {
```
OwenArnold: This whole function will go away when you move to having the Algorithm store the workspace, for one thing the Algorithm knows that the type of the property is a `MatrixWorkspace` so no special casting required.

Framework/DataHandling/test/LoadSampleShapeTest.h
```diff
@@ -0,0 +1,211 @@
...
+      TS_ASSERT_THROWS_NOTHING(alg->execute());
+    }
+  }
+
+  void tearDown() override {
+    for (int i = 0; i < numberOfIterations; i++) {
+      delete loadAlgPtrs[i];
+      loadAlgPtrs[i] = nullptr;
+    }
+    API::AnalysisDataService::Instance().remove(outWsName);
```
OwenArnold: No ADS required. See above.

Framework/DataHandling/test/LoadSampleShapeTest.h
```diff
@@ -0,0 +1,211 @@
...
+        WorkspaceCreationHelper::create2DWorkspaceWithFullInstrument(10, 4);
+    Mantid::API::AnalysisDataService::Instance().add(inWsName, inWs);
+    for (int i = 0; i < numberOfIterations; ++i) {
+      loadAlgPtrs.emplace_back(setupAlg());
+    }
+  }
+
+  void testLoadSampleShapePerformance() {
+    for (auto alg : loadAlgPtrs) {
+      TS_ASSERT_THROWS_NOTHING(alg->execute());
```
OwenArnold: Why create separate Algorithm instances? Why not have a single Algorithm instance and call `execute` on it many times? This would be a lot simpler.

Framework/DataHandling/test/LoadSampleShapeTest.h
```diff
@@ -0,0 +1,211 @@
...
+    for (int i = 0; i < numberOfIterations; i++) {
+      delete loadAlgPtrs[i];
+      loadAlgPtrs[i] = nullptr;
+    }
+    API::AnalysisDataService::Instance().remove(outWsName);
+    API::AnalysisDataService::Instance().remove(inWsName);
+  }
+
+private:
+  std::vector<LoadSampleShape *> loadAlgPtrs;
```
OwenArnold: See comment above about just calling execute multiple times. You don't need this.

Framework/DataHandling/test/LoadSampleShapeTest.h
```diff
@@ -0,0 +1,211 @@
...
+    clearWorkspaces("InputWS", "OutputWS");
+  }
+
+  void test_output_workspace_has_MeshObject_1WS() {
+    loadMeshObject("InputWS", "InputWS", "cube.stl");
+    clearWorkspaces("InputWS", "InputWS");
+  }
+
+  void test_cube() {
+    auto cube = loadMeshObject("InputWS", "InputWS", "cube.stl");
```
OwenArnold: Rather than test the `Algorithm` as one big block, did you consider breaking it up into testable functions/classes etc. The reason I ask is that you are only interested in the `Sample` but for your testing (since you do that via the `Algorithm`), you have to mess around with getting things in and out of `Workspaces`. This is just a consideration you should think about.

OwenArnold (dismissed):
Mainly issues around the tests

Framework/DataHandling/test/LoadSampleShapeTest.h
KarlPalmen: What so you mean by the minor refactor?

OwenArnold: It's not worth changing this. Ignore my comment.

Framework/DataHandling/test/LoadSampleShapeTest.h
KarlPalmen: I'm great difficulty in getting this to work without ADS, which does the useful task of setting up an input workspace. If I try this using a workspace pointer of a workspace created by a helper function, it crashes when I attempt to `setProperty` of the algorithm to it. Do you know of any similar algorithm test that avoids the ADS as you suggest? I'll spend some time looking for one myself.

Framework/DataHandling/test/LoadSampleShapeTest.h
KarlPalmen: Would the use of shared pointer to MeshObject fix this?

OwenArnold: Yes, but I don't think you can get one since I think `getShape` returns a `ref` to the `Object`, and I don't think you should be modifying the API of `Sample` to get around this.

Framework/DataHandling/test/LoadSampleShapeTest.h
OwenArnold: Karl, I'm away next week, perhaps @martyngigg or another more experienced developer in the office could help you with this.

Framework/DataHandling/test/LoadSampleShapeTest.h
KarlPalmen: It seems to be causing a problem, when I'm no longer using the ADS.

Framework/DataHandling/test/LoadSampleShapeTest.h
KarlPalmen: Yes it became a one-liner when I removed the ADS.

Framework/DataHandling/test/LoadSampleShapeTest.h
KarlPalmen: Hopefully fixed in the course of removing the ADS.

NickDraper (changes requested):
The documentation in the RST page for the algorithm is inadequate, as was pointed out before.  It needs to be extended to describe the format it is loading.  It should also stste that it loads ASCII STL, not binary STL files.

The PR header states that the release notes have not been done yet.  This means the PR is not ready for testing!

I'll take a look at the changes now, but it is now very hard as master has been merged in.

Framework/DataHandling/inc/MantidDataHandling/LoadSampleShape.h
```diff
@@ -0,0 +1,71 @@
...
+
+  /// Algorithm's version for identification overriding a virtual method
+  int version() const override { return 1; };
+  /// Algorithm's category for identification overriding a virtual method
+  const std::string category() const override {
+    return "DataHandling\\Instrument";
+  }
+  /// Returns a confidence value that this algorithm can load a file
+  int confidence(Kernel::FileDescriptor &descriptor) const override;
+
```
NickDraper: Please add a seeAlso list with CreateSampleShape, and CopySample at least

Framework/DataHandling/inc/MantidDataHandling/LoadSampleShape.h
```diff
@@ -0,0 +1,71 @@
...
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+File change history is stored at: <https://github.com/mantidproject/mantid>.
+Code Documentation is available at: <http://doxygen.mantidproject.org>
+ */
+
+class DLLExport LoadSampleShape
+    : public API::IFileLoader<Kernel::FileDescriptor> {
```
NickDraper: Why does this inherit from IFileLoader?  why have you done that?

Framework/DataHandling/src/LoadSampleShape.cpp
```diff
@@ -0,0 +1,222 @@
...
+                  "shape of the sample");
+}
+
+/**
+* Return the confidence with with this algorithm can load the file
+* @param descriptor A descriptor for the file
+* @returns An integer specifying the confidence level. 0 indicates it will not
+* be used
+*/
+int LoadSampleShape::confidence(Kernel::FileDescriptor &descriptor) const {
```
NickDraper: There is a question if you need this at all, but if you do this is a simplistic and only partial implementation.  In the exec, you throw an error if the first part of the first line is not "SOLID", this should be checked here.

Framework/DataHandling/src/LoadSampleShape.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+using namespace Kernel;
+using namespace API;
+using namespace Geometry;
+
+namespace {
+
+bool areEqualVertices(Kernel::V3D const &v1, Kernel::V3D const &v2) {
+  Kernel::V3D diff = v1 - v2;
+  return diff.norm() < 1e-9;
```
NickDraper: You should include a comment here explaining what 2e-9 means in real world measurements, given that the base unit of distance in Mantid is m.

Framework/DataHandling/src/LoadSampleShape.cpp
```diff
@@ -0,0 +1,222 @@
...
+#include "MantidKernel/EnabledWhenProperty.h"
+#include "MantidKernel/Exception.h"
+
+#include <Poco/File.h>
+
+namespace Mantid {
+namespace DataHandling {
+
+// Register the algorithm into the algorithm factory
+DECLARE_ALGORITHM(LoadSampleShape)
```
NickDraper: You inherit from IFileLoader, but you only Declare_Algorithm, this does not make sense.  Do you understand how the IFileLoader classes work?

Framework/DataHandling/src/LoadSampleShape.cpp
```diff
@@ -0,0 +1,222 @@
...
+}
+
+std::unique_ptr<Geometry::MeshObject> readSTLSolid(std::ifstream &file,
+                                                   std::string &name) {
+  // Read Solid name
+  // We expect line after trimming to be "solid "+name.
+  std::string line;
+  if (getline(file, line)) {
+    boost::trim(line);
+    if (line.size() < 5 || line.substr(0, 5) != "solid") {
```
NickDraper: should this be a case insensitive comparison?

Framework/DataHandling/src/LoadSampleShape.cpp
```diff
@@ -0,0 +1,222 @@
...
+  declareProperty(
+      make_unique<WorkspaceProperty<>>("InputWorkspace", "", Direction::Input,
+                                       wsValidator),
+      "The name of the workspace containing the instrument to add the shape");
+
+  // shape file
+  const std::vector<std::string> extensions{".stl"};
+  declareProperty(
+      make_unique<FileProperty>("Filename", "", FileProperty::Load, extensions),
+      "The path name of the file containing the shape.");
```
NickDraper: Full stop at the end of this property description, but not the other two?  Consistency

Framework/DataHandling/test/LoadSampleShapeTest.h
```diff
@@ -0,0 +1,174 @@
...
+    LoadSampleShape testLoadSampleShape;
+    testLoadSampleShape.initialize();
+    testLoadSampleShape.setPropertyValue("Filename", "cube.stl");
+    std::string path = testLoadSampleShape.getPropertyValue("Filename");
+    auto *descriptor = new Kernel::FileDescriptor(path);
+    TS_ASSERT_EQUALS(90, testLoadSampleShape.confidence(*descriptor));
+    delete descriptor;
+  }
+
+  void testExec_2WS() {
```
NickDraper: Poorly named test, you can't tell what is is testing from the name alone

Framework/DataHandling/test/LoadSampleShapeTest.h
```diff
@@ -0,0 +1,174 @@
...
+    LoadSampleShape testLoadSampleShape;
+    testLoadSampleShape.initialize();
+    testLoadSampleShape.setPropertyValue("Filename", "cube.stl");
+    prepareWorkspaces(testLoadSampleShape, false);
+    TS_ASSERT_THROWS_NOTHING(testLoadSampleShape.execute());
+    TS_ASSERT(testLoadSampleShape.isExecuted());
```
NickDraper: This test only checks that the algorithm did not throw, this is almost pointless.

Framework/DataHandling/test/LoadSampleShapeTest.h
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  void testExec_1WS() {
```
NickDraper: Same two comments as for the test above

Framework/DataHandling/test/LoadSampleShapeTest.h
```diff
@@ -0,0 +1,174 @@
...
+    TS_ASSERT_THROWS_NOTHING(testLoadSampleShape.execute());
+    TS_ASSERT(testLoadSampleShape.isExecuted());
+  }
+
+  void test_output_workspace_has_MeshObject_2WS() {
+    LoadSampleShape alg;
+    loadMeshObject(alg, false, "cube.stl");
+  }
+
+  void test_output_workspace_has_MeshObject_1WS() {
```
NickDraper: Surely this is just a badly named  subset of test_cube

Framework/DataHandling/test/LoadSampleShapeTest.h
NickDraper: also is there a reason these two tests are not using loadmeshobject?

Framework/DataHandling/test/LoadSampleShapeTest.h
```diff
@@ -0,0 +1,174 @@
...
+  static LoadSampleShapeTest *createSuite() {
+    return new LoadSampleShapeTest();
+  }
+  static void destroySuite(LoadSampleShapeTest *suite) { delete suite; }
+
+  void testName() { TS_ASSERT_EQUALS(loadShape.name(), "LoadSampleShape"); }
+
+  void testVersion() { TS_ASSERT_EQUALS(loadShape.version(), 1); }
+
+  void testInit() {
```
NickDraper: Rather pointless test, also this is the only test that uses the class variableloadShpae, all others define there own, why? consistency.  This looks just copied and pasted without understanding.

Framework/DataHandling/test/LoadSampleShapeTest.h
```diff
@@ -0,0 +1,174 @@
...
+#include "MantidTestHelpers/WorkspaceCreationHelper.h"
+#include <cxxtest/TestSuite.h>
+
+using namespace Mantid;
+using namespace Mantid::API;
+using namespace Mantid::DataHandling;
+using namespace Mantid::DataObjects;
+using namespace Mantid::Geometry;
+
+class LoadSampleShapeTest : public CxxTest::TestSuite {
```
NickDraper: Where are the negative tests?
Missing file?
Incorrect file format, you have at least two throw lines inside the algorithm, these should be tested.

NickDraper (changes requested):

Framework/DataHandling/inc/MantidDataHandling/LoadSampleShape.h
KarlPalmen: I had originally hoped one could load an STL file using the load button in MantidPlot, but I no longer think it would work in practice so will remove this feature.

NickDraper: This would not make sense as LoadSampleShape modifies an existing workspace, rather than creating a new one from a file.

Framework/DataHandling/test/LoadSampleShapeTest.h
KarlPalmen: This test only tests whether the algorithm has executed.

NickDraper: Yes, but what is the point of that?

Surely any test has to check that the code actually did something rather than just did not complain that it failed?

An algorithm with an empty exec() method would pass your test here.

Framework/DataHandling/src/LoadSampleShape.cpp
KarlPalmen: I checked the standard at http://www.fabbers.com/tech/STL_Format and found out that keywords do need to be in lower case.

Framework/DataHandling/test/LoadSampleShapeTest.h
KarlPalmen: Missing file is dealt by the file property manager when one sets the filename property, so is not necessary, but I will add tests for malformed files.

Framework/DataHandling/inc/MantidDataHandling/LoadSampleShape.h
```diff
@@ -0,0 +1,72 @@
+#ifndef DATAHANDLING_LOAD_SHAPE_H_
+#define DATAHANDLING_LOAD_SHAPE_H_
+
+#include "MantidAPI/IFileLoader.h"
```
NickDraper: Why are we still including IFileLoader?

Framework/DataHandling/src/LoadSampleShape.cpp
```diff
@@ -0,0 +1,203 @@
...
+    g_log.error("Unable to open file: " + filename);
+    throw Exception::FileError("Unable to open file: ", filename);
+  }
+
+  std::string solidName = "";
+  boost::shared_ptr<MeshObject> shape = nullptr;
+  try {
+    shape = readSTLSolid(file, solidName);
+  } catch (std::exception &e) {
+    throw Exception::FileError(e.what(), filename);
```
NickDraper: What is this try catch intending to do?

This code is at best misleading and potentially dangerous.
How is this catch handling the error?

NickDraper (changes requested):

Framework/DataHandling/src/LoadSampleShape.cpp
KarlPalmen: I now realise it is just passing on an exception thrown lower down, so is not needed.

Framework/DataHandling/inc/MantidDataHandling/LoadSampleShape.h
KarlPalmen: I find I need it for `getline` to make sense. There may be a better way of doing this.

Framework/DataHandling/src/LoadSampleShape.cpp
```diff
@@ -0,0 +1,200 @@
+#include "MantidDataHandling/LoadSampleShape.h"
+
+#include "MantidAPI/IFileLoader.h"
```
NickDraper: Just moving the include to the cpp file is not an answer, you should not include an interface that you are not using.

NickDraper (changes requested):

Framework/DataHandling/src/LoadSampleShape.cpp
```diff
@@ -0,0 +1,201 @@
+#include "MantidDataHandling/LoadSampleShape.h"
+
+#include "MantidAPI/AnalysisDataService.h"
+#include "MantidAPI/FileProperty.h"
+#include "MantidAPI/InstrumentValidator.h"
+#include "MantidAPI/WorkspaceFactory.h"
+#include "MantidAPI/MatrixWorkspace.h"
+#include "MantidAPI/MatrixWorkspace_fwd.h"
```
NickDraper: what is this doing in a cpp file?

Framework/DataHandling/src/LoadSampleShape.cpp
```diff
@@ -0,0 +1,201 @@
...
+#include "MantidAPI/Sample.h"
+
+#include "MantidGeometry/Instrument.h"
+#include "MantidKernel/Material.h"
```
NickDraper: I don't see why you need this either

Framework/DataHandling/src/LoadSampleShape.cpp
```diff
@@ -0,0 +1,201 @@
...
+#include "MantidKernel/CompositeValidator.h"
```
NickDraper: You don't seem to use one of these

Framework/DataHandling/src/LoadSampleShape.cpp
```diff
@@ -0,0 +1,201 @@
...
+#include "MantidKernel/EnabledWhenProperty.h"
```
NickDraper: or one of these

Framework/DataHandling/src/LoadSampleShape.cpp
```diff
@@ -0,0 +1,201 @@
+#include "MantidDataHandling/LoadSampleShape.h"
+
+#include "MantidAPI/AnalysisDataService.h"
+#include "MantidAPI/FileProperty.h"
+#include "MantidAPI/InstrumentValidator.h"
+#include "MantidAPI/WorkspaceFactory.h"
```
NickDraper: Do you need this?

Framework/DataHandling/src/LoadSampleShape.cpp
```diff
@@ -0,0 +1,201 @@
+#include "MantidDataHandling/LoadSampleShape.h"
+
+#include "MantidAPI/AnalysisDataService.h"
```
NickDraper: You aren't using this either, just some other include that are getting included by accident

Framework/DataHandling/inc/MantidDataHandling/LoadSampleShape.h
```diff
@@ -0,0 +1,72 @@
+#ifndef DATAHANDLING_LOAD_SHAPE_H_
+#define DATAHANDLING_LOAD_SHAPE_H_
+
+#include "MantidGeometry/Objects/MeshObject.h"
+#include "MantidAPI/Algorithm.h"
+#include "MantidKernel/V3D.h"
```
NickDraper: Doesn't look like you use this in this file

Framework/DataHandling/inc/MantidDataHandling/LoadSampleShape.h
```diff
@@ -0,0 +1,72 @@
+#ifndef DATAHANDLING_LOAD_SHAPE_H_
+#define DATAHANDLING_LOAD_SHAPE_H_
+
+#include "MantidGeometry/Objects/MeshObject.h"
```
NickDraper: This can be moved to the cpp file

NickDraper (changes requested):

NickDraper (approved):

docs/source/release/v3.13.0/reflectometry.rst
```diff
@@ -24,8 +24,6 @@ Bug fixes
 Features Removed
 ################
 
-* Added deprecation notice to ISIS Reflectometry (Old) due to be removed in March 2019.
```
martyngigg: Was this release note meant to be removed?

martyngigg: It looks important.

docs/source/release/v3.13.0/muon.rst
```diff
@@ -8,18 +8,12 @@ MuSR Changes
...
----------
-
-
 Improvements
 ############
 
 Bug fixes
 #########
 
-- Results table can now detect sequential fits.
```
martyngigg: Was this release note meant to be removed?

docs/source/release/v3.13.0/reflectometry.rst
KarlPalmen: It was not meant to be removed.

docs/source/release/v3.13.0/muon.rst
KarlPalmen: It was not meant to be removed.

martyngigg:
RHEL 7 failure is unrelated.

PR #22035, opened 2018-03-09
Numerical derivatives must apply ties after parameter shift.

### LINKED ISSUE #22004: Results table in muon analysis
If you do a multiple fit in muon analysis with some global values and some fixed values, then the results table will have global values printed for every workspace (the values will also be slightly different).

mantid-roman:
Call `applyTies()` after a shift is added to a parameter when calculating the numerical derivatives.

**To test:**

Instructions are in the issue description.

Fixes #22004

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


AnthonyLim23 (changes requested):
We should add something to the release notes. I have also just noticed that the errors are not copied over for the ties (is this easy to fix)? 

The code change looks good and it works

mantid-roman:
@AnthonyLim23 Can you suggest what to say in the release notes? We cannot say that all issues with ties have been fixed.

AnthonyLim23:
We could just say that tied parameters are updated correctly in multiple fits (or do you think this is too vague)

mantid-roman:
You can still set ties that won't be updated correctly

AnthonyLim23:
ah. Ok we will leave it then. Will the errors being copied be much work?

mantid-roman:
Sorry, I didn't understand the question

AnthonyLim23:
If you look at the screenshots from the issue you will see that the errors are not copied to the other workspaces. How much work would that be to fix?

mantid-roman:
Do they need to be copied?

AnthonyLim23:
just thought it could be confusing

mantid-roman:
I guess it depends on the point of view. But if it's what the scientists want it shouldn't be too hard to add.

Testing/SystemTests/tests/analysis/reference/II.IRISFuryFitMulti.nxs.md5
```diff
@@ -1 +1 @@
-ce0d536337942b5662b8ecd107a5c71f
+d3c125646057bc0d1291c3286ddfab82
```
AnthonyLim23: what is the difference between these two files? Is it just the parameters being copied correctly?

mantid-roman: There was one number different

AnthonyLim23: was it by much?

AnthonyLim23:
for the release notes could we say:

Fix for a bug in calculating numerical derivatives by applying ties correctly. 

?

Testing/SystemTests/tests/analysis/reference/II.IRISFuryFitMulti.nxs.md5
mantid-roman: Less than 0.1%

mantid-roman:
This is more accurate

AnthonyLim23 (approved):

NickDraper:
builds cancelled to free up Q, final change was release notes only

AnthonyLim23:
but the last few tests never finished on the previous build

PR #22052, opened 2018-03-12
Complain if underlying data changes in the phase quad gui

### LINKED ISSUE #22031: PhaseQuad can add results to the wrong group workspace
### Expected behavior
The results should go in the right group workspace

### Actual behavior
If you run another file through Muon Analysis after opening the Frequency analysis gui then the PhaseQuad will use the new workspace, but store the results into the group workspace of the old one.

1. Start the Muon Analysis interface
1. Load MUSR00068389
1. Start the Frequency domain GUI
1. Set the workspace to be PhaseQuad
1. Go back to the Muon Analysis interface
1. Load MUSR00068675
1. Go back to the Frequency domain GUI
1. Click calculate FFT
1. The PhaseQuad;FFT workspace is added to the MUSR00068389 group workspace, but it is based from

image

AnthonyLim23:
The frequency domain analysis would use incorrect workspaces in its calculations if the data in muon analysis changed

**To test:**
Open muon analysis
Load some data
Open frequency domain analysis
Change workspace to PhaseQuad and run it (you may have to change the endX)
See that the name now includes the run name (instrument + run number)
Change the method to maxent
Tick all of the output options
Run maxent
See that the output names now include which muon analysis they worked on (useful for multi period data such as HIFI 84447)

Change the data in Muon analysis (without closing the frequency GUI)
Go back to the frequency GUI
Run maxEnt - it should give an error then update the GUI
Try to run it again and it should add the output to the new group ws with correct names
Go back to FFT and try another phaseQuad - it should give an error and update
Try to run it again and it should work (add the output to the new ws and with correct names)

Fixes #22031. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


NickDraper (approved):
Overall I would have preferred a more visible error message than just a log entry, And a refresh option would have been better than having to reopen the interface.

However, given where we are in the release this is acceptable for now.

AnthonyLim23:
I'm thinking that in future instead of just throwing a log warning we then reload the data automatically. And then resubmit the calculation (but to do this well will take some thought)

NickDraper:
@AnthonyLim23 You have a few flake8 warnings to sort out

AnthonyLim23:
I have thought of a few improvements. Most have now been implemented and I am just waiting to check that it passes the tests. 

I have moved the checking to the thread model and have a method that can be called by the presenters at a later data for reloading the data (#22059). I have also improved the naming convention for the maxent

AnthonyLim23:
I have also improved the naming of the PhaseQuad outputs and the checks. The previous version only compared the run numbers (so if two runs had the same number but different instruments it would have not failed) and now we also look at the instrument.

scripts/Muon/thread_model.py
```diff
@@ -9,14 +10,31 @@ class ThreadModel(QThread):
...
         self.model=model
+        self.run = run
 
     def __del__(self):
         self.wait()
 
+    def digit(self,x):
+        return int(filter(str.isdigit,x) or 0)
+
+    def hasDataChanged(self):
```
martyngigg: I'm not sure the `TheadModel` is the correct place for this check. `ThreadModel` has the responsibility  for starting threads and it shouldn't have other responsibilities that are not related to this.

As discussed offline the `load_utils.py` might make more sense for this as that already knows about the workspace names etc.

martyngigg (changes requested):

AnthonyLim23:
moved the error check and it now reloads the data if it has changed

martyngigg:
I seem to get the name+instrument combination duplicated in the names (see below). Is that intended?

screenshot from 2018-03-13 12-19-04

AnthonyLim23:
yes that is intended. As it is useful if you use one of the grouped workspaces as you then get

run;grouped ws...

martyngigg:
@AnthonyLim23 Thanks for the explanation - I just wanted to check.

martyngigg (approved):

AnthonyLim23:
no worries, it does look strange

AnthonyLim23:
I have fixed the merge conflict and the previous build passed so we can cancel the builds

PR #22077, opened 2018-03-14
SANS V2 Simplified PrintParser

### LINKED ISSUE #22076: SANS v2 user file parser failing for certain files
The user file parser used by the new sans isis backend fails if the user file contains regexp special characters in the print statement.

Matthew-Andrew:
This fixes an issue where the user file import was failing if any print statements contained regexp special characters. 

**To test:**
Check that you can successfully load the following LOQ user file in the new sans GUI. 
[USER_LOQ_174J_M3_Lu_12mm_Changer_MAIN.TXT](https://github.com/mantidproject/mantid/files/1811048/USER_LOQ_174J_M3_Lu_12mm_Changer_MAIN.TXT)

Note that the new GUI currently does not do anything with print statements put inside the user file. This may or may not be by intention I need to check with the instrument scientists. 

Fixes #22076 . 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/scripts/SANS/sans/user_file/user_file_parser.py
+++ b/scripts/SANS/sans/user_file/user_file_parser.py
@@ -2084,11 +2084,11 @@ class PrintParser(UserFileComponentParser):
 
         setting = line.strip()
 
-        if setting.upper().startswith('PRINT'):
-            setting = setting[len('PRINT'):]
+        if setting.upper().startswith(PrintParser.Type):
+            setting = setting[len(PrintParser.Type):]
             setting = setting.strip()
         else:
-            raise RuntimeError("PrintParser: Failed to extract line {} it does not start with PRINT".format(line))
+            raise RuntimeError("PrintParser: Failed to extract line {} it does not start with {}".format(line, PrintParser.Type))
 
         return {PrintId.print_line: setting}
 
--- a/scripts/SANS/sans/user_file/user_file_parser.py
+++ b/scripts/SANS/sans/user_file/user_file_parser.py
@@ -2079,10 +2079,6 @@ class PrintParser(UserFileComponentParser):
     def __init__(self):
         super(PrintParser, self).__init__()
 
-        # Print
-        self._print = "\\s*PRINT\\s+"
-        self._print_pattern = re.compile(start_string + self._print + "\\s*.*\\s*" + end_string)
-
     def parse_line(self, line):
         # Get the settings, ie remove command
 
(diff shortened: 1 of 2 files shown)


scripts/SANS/sans/user_file/user_file_parser.py
```diff
@@ -2085,11 +2085,16 @@ def __init__(self):
```
martyngigg: Should we remove the 
```
self._print = "\\s*PRINT\\s+"
self._print_pattern = re.compile(start_string + self._print + "\\s*.*\\s*" + end_string)
```
lines above?

scripts/SANS/sans/user_file/user_file_parser.py
```diff
@@ -2085,11 +2085,16 @@ def __init__(self):
 
     def parse_line(self, line):
         # Get the settings, ie remove command
-        setting = UserFileComponentParser.get_settings(line, PrintParser.get_type_pattern())
 
-        # Determine the qualifier and extract the user setting
-        original_setting = re.search(setting.strip(), line, re.IGNORECASE).group(0)
-        return {PrintId.print_line: original_setting}
+        setting = line.strip()
```
martyngigg: I think this is much cleaner and easier to follow.

scripts/SANS/sans/user_file/user_file_parser.py
Mantid-Matthew: Yes I think so have now done. I have left in the static member functions below because I think they are used elsewhere when deciding which parser to use and removing them currently breaks the unit tests

martyngigg: That sounds fine. It's good that there are tests that cover this and thanks for adding the new ones.

scripts/SANS/sans/user_file/user_file_parser.py
```diff
@@ -2079,17 +2079,18 @@ class PrintParser(UserFileComponentParser):
...
+
+        if setting.upper().startswith('PRINT'):
```
NickDraper: This and the next line should use the Type class variable rather than hard-coded "PRINT"

NickDraper (changes requested):

scripts/SANS/sans/user_file/user_file_parser.py
```diff
@@ -2079,17 +2079,18 @@ class PrintParser(UserFileComponentParser):
...
+            setting = setting[len('PRINT'):]
+            setting = setting.strip()
+        else:
+            raise RuntimeError("PrintParser: Failed to extract line {} it does not start with PRINT".format(line))
```
NickDraper: And this line should use TYPE too

NickDraper (changes requested):

NickDraper (approved):

martyngigg (approved):

PR #21965, opened 2018-03-06
Indirect  - Files update when instrument is selected

MatthewBowles:
MWRunFiles did not re-find files after the instrument selection was changed. This caused a bug in the diffraction reduction interface.

**To test:**

- Navigate to Interfaces -> Indirect -> Diffraction
- Ensure Instrument Selection is **not** OSIRIS (If it is then change it)
- Enter 119976, 119977, 119981 as the Run numbers.
- Enter 71988, 71991, 71990 into the 'Container' field.
- Change the instrument selection to OSIRIS (the run button at the bottom should change to 'Finding files' briefly).
- Enter 119965, 119967 into the 'Vanadium' field.
- Enter calibration file [osiris_041_RES10.zip](https://github.com/mantidproject/mantid/files/1784938/osiris_041_RES10.zip)
- Click 'Run'
- Should run without errors and give *_dRange and *_tof workspaces in the ADS

Fixes #21634

**Release Notes** 
*Does not need to be in the release notes.*

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)?
- [ ] Are the unit tests small and test the class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] Do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant (user and developer) documentation been added/updated?

Does everything look good? Mark the review as **Approve**. A member of `@mantidproject/gatekeepers` will take care of it.

PR #22084, opened 2018-03-14
MWRunFiles - Emit findingFiles() when searchText is not empty

### LINKED ISSUE #22074: Indirect - Data Reduction - MWRunFiles widget is disabled
In the ```Interfaces > Indirect > Data Reduction``` interfaces, the MWRunFiles widget (in all places it occurs) is disabled. Changing the instrument in this interface does not enable the widget. This currently renders many of the Indirect Data Reduction interfaces unusable, including ISISEnergyTransfer.

brandonhewer:
In the Interfaces > Indirect > Data Reduction interfaces, the MWRunFiles widget (in all places it occurs) is disabled. Changing the instrument in this interface does not enable the widget. This currently renders many of the Indirect Data Reduction interfaces unusable, including ISISEnergyTransfer.

This PR ensures the emit is moved to the correct place.

**To test:**
1. Test widget in Indirect Data Reduction, Indirect Diffraction, Muon Analysis and all other appropriate places.

Fixes #22074. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/qt/widgets/common/src/MWRunFiles.cpp
+++ b/qt/widgets/common/src/MWRunFiles.cpp
@@ -550,9 +550,6 @@ void MWRunFiles::findFiles(bool isModified) {
 * @return search text to create search params with
 */
 const QString MWRunFiles::findFilesGetSearchText(QString &searchText) {
-
-  emit findingFiles();
-
   // If we have an override instrument then add it in appropriate places to
   // the search text
   if (!m_defaultInstrumentName.isEmpty()) {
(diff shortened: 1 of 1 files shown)


martyngigg (approved):
@brandonhewer has explained to me that adding the `findFiles(True)` in #21965 as part of the `setInstrumentOverride` caused this because in all other circumstances `findFiles` ends up being called twice first with `modified=True` and then with `modified=False` and this the final `fileInspectionFinished()` event is called the second time around.

What is not clear is why the widget requires `findFiles` to be called twice to function as one would expect. This needs exploration.

The fix here is sensible and he interfaces are working again. It should probably stay in the long run too as it only really makes sense to emit the `findingFiles` signal when the expensive operations are about to start.

martyngigg:
This needs to be in nightly as man interfaces are broken.

PR #22088, opened 2018-03-14
A few changes to support msvc 2017

LamarMoore:
Fixes warnings coming from the MSVC 2017 build.

**To test:**

Build locally using msvc 2017 to ensure everything builds cleanly and they are no build or warnings. Also ensure the 2015 build still works,

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/MDAlgorithms/src/CalculateCoverageDGS.cpp
```diff
@@ -425,7 +425,9 @@ void CalculateCoverageDGS::exec() {
       std::vector<coord_t> pos(4);
       std::transform(curIntSec.getBareArray(), curIntSec.getBareArray() + 4,
                      prevIntSec.getBareArray(), pos.begin(),
-                     VectorHelper::SimpleAverage<double>());
```
martyngigg: Does `SimpleAverage` produce warnings now?

LamarMoore: Yes because the `double` return value is being converted to a `float` (`coord_t`) in `pos`.

martyngigg:
This will now be done in the maintenance period after 3.13.

Framework/MDAlgorithms/src/CalculateCoverageDGS.cpp
martyngigg: Should we remove the `SimpleAverage` functor now?

buildconfig/CMake/FindBoost.cmake
```diff
@@ -1,3 +1,6 @@
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
```
martyngigg: If we upgrade cmake as discussed can we remove this file?

buildconfig/CMake/FindBoostPython.cmake
```diff
@@ -5,7 +5,11 @@
 #  - windows: boost_python (python2), ????? (python3)
 #  - others?
 if ( MSVC )
-  find_package ( Boost COMPONENTS python REQUIRED )
+  if ( ${Boost_VERSION} LESS 106700)
```
martyngigg: Maybe we don't need the `if` check? We could assume that on windows we always know what version of boost we're running.

LamarMoore: Good point

buildconfig/CMake/FindBoost.cmake
LamarMoore: Won't it still be needed for other platforms? Or is this strictly used for windows?

martyngigg: Good question. I misunderstood and thought we added it here. I'm not sure why we had it in the first place could you ask on Slack if anyone knows?

martyngigg (approved):

PR #22145, opened 2018-03-19
Fix binning params in the SofQW* algorithms

### LINKED ISSUE #22140: SofQWNormalisedPolygon fails if a single value is given as QBinningParams
If *QBinningParams* is supplied with only the bin width, `SofQWNormalisedPolygon` throws the following error:

```
RuntimeError: All arguments to init must be positive and non-zero
```

This seems to be due to `Mantid::Kernel::VectorHelper::createAxisFromRebinParams` doesn't handle the bin width only case properly.

soininen:
The `SofQW*` algorithms failed when they were supplied with 'bin width only' binning parameters. This PR fixes the issue by adding E and Q binning range estimation and by fixing the 'bin width only' case in `VectorHelper::createAxisFromRebinParams`.

**To test:**

The following script should work after these changes are applied. The data can be found in Mantid's unit test data.
```python
DirectILLCollectData(
    Run='ILL/IN6/164192.nxs',
    OutputWorkspace='ws',
)
ws = mtd['ws']
e = ws.run().getProperty('Ei').value
print(e)
ws = ConvertUnits(ws, 'DeltaE', EMode='Direct', EFixed=e)
ws = Rebin(ws, '-10000, 5, 5')

ws = SofQWNormalisedPolygon('ws', QAxisBinning='0.1', EAxisBinning='-10000, 5, 5', EMode='Direct', EFixed=e)
```

Fixes #22140.

**Release Notes** 

The Direct Inelastic release notes were updated.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Kernel/src/VectorHelper.cpp
```diff
@@ -27,6 +27,10 @@ int DLLExport createAxisFromRebinParams(const std::vector<double> &params,
                                         std::vector<double> &xnew,
                                         const bool resize_xnew,
                                         const bool full_bins_only) {
+  if (params.size() == 1) {
```
SimonHeybrock: I am not sure I understand what you did here. Don't other algorithms like `Rebin` support such parameters?

SimonHeybrock (changes requested):

Framework/Kernel/src/VectorHelper.cpp
soininen: They do but in case of `Rebin`, the min and max X are known. The `SofQW*` algorithms convert to momentum transfer and it's a bit tedious to guess suitable Qmin and Qmax beforehand.

Framework/Algorithms/src/SofQCommon.cpp
```diff
@@ -82,5 +85,142 @@ double SofQCommon::getEFixed(const Geometry::IDetector &det) const {
...
   return efixed;
 }
+
+/**
+ * Return a pair of (minimum DeltaE, maximum DeltaE)
+ * @param ws a workspace
+ * @return a pair containing the global minimum and maximum X
+ */
+std::pair<double, double>
+SofQCommon::eBinHints(const API::MatrixWorkspace &ws) const {
```
SimonHeybrock: Isn't this the same as `MatrixWorkspace::getXMinMax`?

Framework/Algorithms/src/SofQCommon.cpp
```diff
@@ -82,5 +85,142 @@ double SofQCommon::getEFixed(const Geometry::IDetector &det) const {
...
+  const auto maxK = std::sqrt(maxKSq);
+  std::array<double, 4> q;
+  q[0] = std::sqrt(incidentKSq + minKSq -
+                   2. * incidentK * minK * std::cos(minTheta));
+  q[1] = std::sqrt(incidentKSq + minKSq -
+                   2. * incidentK * minK * std::cos(maxTheta));
+  q[2] = std::sqrt(incidentKSq + maxKSq -
+                   2. * incidentK * maxK * std::cos(minTheta));
+  q[3] = std::sqrt(incidentKSq + maxKSq -
+                   2. * incidentK * maxK * std::cos(maxTheta));
```
SimonHeybrock: It feels like this conversion must be something that has been implemented before somewhere else?

Framework/Algorithms/src/SofQW.cpp
```diff
@@ -151,21 +142,37 @@ void SofQW::exec() {
...
     const std::vector<double> &qbinParams, std::vector<double> &qAxis,
-    const std::vector<double> &ebinParams) {
+    const std::vector<double> &ebinParams, const SofQCommon &emodeProperties) {
+  using Kernel::VectorHelper::createAxisFromRebinParams;
   // Create vector to hold the new X axis values
   HistogramData::BinEdges xAxis(0);
+  auto eHints = std::make_pair<double, double>(std::nan(""), std::nan(""));
   int xLength;
   if (ebinParams.empty()) {
     xAxis = inputWorkspace->refX(0);
```
SimonHeybrock: `refX` is deprecated.

Framework/Algorithms/src/SofQWNormalisedPolygon.cpp
```diff
@@ -411,18 +413,32 @@ RebinnedOutput_sptr SofQWNormalisedPolygon::setUpOutputWorkspace(
...
+    if (std::isnan(eHints.first)) {
+      eHints = m_EmodeProperties.eBinHints(inputWorkspace);
+    }
+    const auto qHints = m_EmodeProperties.qBinHints(
+        inputWorkspace, eHints.first, eHints.second);
+    yLength = createAxisFromRebinParams(qbinParams, qAxis, true, true,
+                                        qHints.first, qHints.second);
+  } else {
+    yLength = createAxisFromRebinParams(qbinParams, qAxis);
+  }
```
SimonHeybrock: This code looks identical to the one above in `SofQW`, can it be extracted into a shared function/method without too much hassle?

Framework/Algorithms/test/SofQCommonTest.h
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  void test_getEFixedFromDetectorsIndirectGeometry() {
+    using namespace Mantid;
+    using namespace WorkspaceCreationHelper;
+    Algorithms::SofQW alg;
+    alg.initialize();
+    alg.setProperty("EMode", "Indirect");
+    // const double Ei{2.3};
+    // alg.setProperty("EFixed", Ei);
```
SimonHeybrock: Leftover comments.

SimonHeybrock (changes requested):
Looks nice, with a good amount of cleanup and many tests 👍.

Only a couple of minor comments:

SimonHeybrock (approved):
Nice, even more cleanup 👍

Framework/Algorithms/inc/MantidAlgorithms/SofQCommon.h
```diff
@@ -27,12 +27,20 @@ struct DLLExport SofQCommon {
   /// Get the efixed value for the given detector
   double getEFixed(const Geometry::IDetector &det) const;
 
+
+  /// Calculate the Q value
+  double q(const double deltaE, const double twoTheta, const Geometry::IDetector &det) const;
```
SimonHeybrock: Getting the detector from the instrument is a really expensive operation (one of the few that still needs to use the old instrument code), can't it be avoided in direct mode?

Framework/Algorithms/src/SofQCommon.cpp
```diff
@@ -167,20 +205,15 @@ SofQCommon::qBinHintsIndirect(const API::MatrixWorkspace &ws, const double minE,
...
-    const auto theta = detectorInfo.twoTheta(i) / 2.;
-    const auto eFixed = getEFixed(detectorInfo.detector(i));
-    const Kernel::DeltaEMode::Type indirectEMode{Kernel::DeltaEMode::Indirect};
-    Kernel::Units::DeltaE DeltaEUnit;
-    Kernel::Units::MomentumTransfer QUnit;
-    const auto Q1 = Kernel::UnitConversion::run(DeltaEUnit, QUnit, minE, 0., 0.,
-                                                theta, indirectEMode, eFixed);
-    const auto Q2 = Kernel::UnitConversion::run(DeltaEUnit, QUnit, maxE, 0., 0.,
-                                                theta, indirectEMode, eFixed);
+    const auto theta = spectrumInfo.twoTheta(i);
```
SimonHeybrock: Dangerous naming? `twoTheta` assigned to `theta`?

Framework/Algorithms/src/SofQCommon.cpp
```diff
@@ -167,20 +205,15 @@ SofQCommon::qBinHintsIndirect(const API::MatrixWorkspace &ws, const double minE,
   using namespace Mantid::PhysicalConstants;
   auto minQ = std::numeric_limits<double>::max();
   auto maxQ = std::numeric_limits<double>::lowest();
-  const auto &detectorInfo = ws.detectorInfo();
-  for (size_t i = 0; i < detectorInfo.size(); ++i) {
-    if (detectorInfo.isMasked(i) || detectorInfo.isMonitor(i)) {
+  const auto &spectrumInfo = ws.spectrumInfo();
+  for (size_t i = 0; i < spectrumInfo.size(); ++i) {
```
SimonHeybrock: Are you sure about this change? I see you adjusted the doxygen above. One thing to double check: What does `getEFixed` to for a `DetectorGroup`? Unless it returns an average (of the E_f looked up in the parameter map) for all contained detectors I think you cannot change this here.

Framework/Algorithms/src/SofQWNormalisedPolygon.cpp
```diff
@@ -113,23 +112,19 @@ void SofQWNormalisedPolygon::exec() {
...
-    double phiHalfWidth = 0.5 * phiWidth;
+    const double thetaHalfWidth = 0.5 * thetaWidth;
 
     const double thetaLower = theta - thetaHalfWidth;
     const double thetaUpper = theta + thetaHalfWidth;
 
-    const double phiLower = phi - phiHalfWidth;
-    const double phiUpper = phi + phiHalfWidth;
-
-    const double efixed = m_EmodeProperties.getEFixed(spectrumInfo.detector(i));
```
SimonHeybrock: Hmm, this is also using `getEFixed` for a (potential) `DetectorGroup`, does this imply that it does the right thing...?

Edit: tried to track it down: I believe `getEFixed` simply throw in that case?

SimonHeybrock (changes requested):
Some comments:

Framework/Algorithms/src/SofQWNormalisedPolygon.cpp
soininen: `DetectorGroup::getNumberParameter()` always returns an empty `vector` so `SofQCommon::getEFixed()` would indeed throw for `DetectorGroup`.

SimonHeybrock (approved):

PR #22161, opened 2018-03-21
Fix clang-tidy misc-unconventional-assign-operator warnings.

### LINKED ISSUE #22144: Fix misc-unconventional-assign-operator errors.
There are 4 clang-tidy errors caused by the return types of various property assignment operators.
This issue is to fix that.

edwardb96:
Clang tidy issues warnings for unconventional assignment operators which don't return `*this`. This PR fixes a couple of instances where this is the case within Mantid.

**To test:**
1. Ensure builds pass.
2. Code Inspection

Fixes #22144. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


martyngigg (commented):
@LamarMoore @SimonHeybrock Just want to check you're okay with the changes here r.e. `IndexProperty` and `IndexInfo`?

Framework/Indexing/src/IndexInfo.cpp
```diff
@@ -90,7 +90,8 @@ IndexInfo::~IndexInfo() = default;
 
 IndexInfo &IndexInfo::operator=(const IndexInfo &other) {
   auto copy(other);
-  return *this = std::move(copy);
+  *this = std::move(copy);
+  return *this;
```
SimonHeybrock: Isn't the old version doing the same thing?

SimonHeybrock (commented):
@martyngigg Regarding the properties, what `IndexProperty` does seems like it was simply copied from the behavior of `PropertyWithValue`, i.e., the question we should ask is why `PropertyWithValue::operator=` was written in this way?

Framework/Indexing/src/IndexInfo.cpp
edwardb96: Functionally they are the same yes. Clang Tidy doesn't see that though. This is arguably more readable at a quick glance.

martyngigg:
@SimonHeybrock Good point and not one I can answer as no one is left from when that code was originated. If I'm not mistaken the only situations that should be affected are chained assignments, right? I don't think we do that anywhere and I certainly can't see why we'd actually want the current non-standard behaviour so I'm happy for this to be changed - would you agree?

SimonHeybrock:
I agree. In the case of chained assignment I assume we would either get a compiler failure, or there is an implicit conversion from `PropertyWithValue` back to string that covers such cases? In any case I do not see a reason to keep this, given that it compiles and tests pass.

martyngigg (approved):

PR #22163, opened 2018-03-21
Maxent algorithm output of iteration records and use of complex data

### LINKED ISSUE #21444: MaxEnt: return the number of MaxEnt iterations for each dataset
Make the alg return the correct number of maxent iterations when running the maxent alg with multiple input datasets.

When running the script:

[maxentTest.zip](https://github.com/mantidproject/mantid/files/1642342/maxentTest.zip)

it returns:

`Number of iterations dataset1: 11`
`Number of iterations dataset2: 7`
`Number of iterations dataset1: 9`
`Number of iterations dataset2: 9`

whereas it should return:

`Number of iterations dataset1: 11`
`Number of iterations dataset2: 7`
`Number of iterations dataset1: 11`
`Number of iterations dataset2: 7`

* fix this. This means 1) allow individual number of iterations to be extracted from the maxent output 2) ensure the right number of iterations are returned
* add the attached python code as a new Usage example
* ensure the documentation for how to get number of iterations is up to date

KarlPalmen:
Ensure that Maxent algorithm returns for each spectrum, in the two workspaces that record the iterations exactly one element per iteration with no padding with zeros. Update the documentation and provide a usage example covering this and also how to combine two workspaces with complex data.

**To test:**

Build the documentation check that it is OK and up to date.

Run the usage example added to the documentation. 
Then select workspaces `EvolAngleP` & `EvolAngleP2` and plot them, then plot `EvolAngleC` both plots should be identical, except for legend and the second spectrum's plot should not go all the way to the right.

You may then try out other things with Maxent.

Fixes #21444. 

**Release Notes** 
In ...\release\v3.13.0\framework.html of documentation built by you.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


docs/source/algorithms/MaxEnt-v1.rst
```diff
@@ -100,33 +100,33 @@ the output workspaces whether the algorithm was evolving towards the correct sol
...
     +===================+==============================+================+====================================================================================================================================================================================================================================================================================================================+
-    | EvolChi           | M                            | MaxIterations  | Evolution of :math:`\chi^2` until the solution is found. Then all values are set to zero.                                                                                                                                                                                                                          |
+    | EvolChi           | M                            | J              | For spectrum :math:`s` in the input workspace, evolution of :math:`\chi^2` until the solution is found                                                                                                                                                                                                             |
```
AnthonyLim23: can we use a different letter for the spectrum? S is the entropy and I think it might be confusing

KarlPalmen: Perhaps, that is why 's' is used for spectrum rather than 'S'.

AnthonyLim23: to avoid any confusion you could say the jth spectrum

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -397,15 +409,17 @@ void MaxEnt::exec() {
...
+      if (it > maxItCount) {
+        maxItCount = it;
       }
+
       // Stop condition, solution found
       if ((std::abs(currChisq / ChiTargetOverN - 1.) < chiEps) &&
           (currAngle < angle)) {
 
-        g_log.information() << "Stopped after " << it << " iterations"
+        g_log.information() << "Stopped after " << it + 1 << " iterations"
```
AnthonyLim23: why has this got a plus one now?

KarlPalmen: This is because `it` is counted from zero, one needs to add 1 to get the correct number of iterations.

AnthonyLim23:
@Anders-Markvardsen I have had a look at the code and it seems ok, just got a question about it. I have not checked the docs. 

The script works, but I did notice that the table output does have zeros in.

After running the script and looking at the table for chi^2. If you then try to run maxEnt an uncaught exception is thrown 

```
Unexpected exception: blocksize undefined because size of histograms is not equal
Continue working.

```

Framework/Algorithms/src/MaxEnt.cpp
AnthonyLim23: ok

Anders-Markvardsen:
@AnthonyLim23 could you provide a step by step instructions of how to reproduce the crash you report above please

AnthonyLim23:
Run the script (copied below for simplicity)
Load some data (I used MUSR 62260)
Try to run ManEnt from the algorithm algorithm dialog
Mantid throws an uncaught exception


```
 from math import pi, sin, cos
from random import random, seed
seed(0)
# Create complex data for a workspace
X = []
YRe = []
YIm = []
E = []
N = 200
w = 3
for i in range(0,N):
       x = 2*pi*i/N
       X.append(x)
       YRe.append(cos(w*x)+(random()-0.5)*0.3)
       YIm.append(sin(w*x)+(random()-0.5)*0.3)
       E.append(0.1)

seed(0)
# Create complex data for a second workspace
X2 = []
YRe2 = []
YIm2 = []
E2 = []
N2 = 200
w2 = 4
for i in range(0,N2):
       x = 2*pi*i/N2
       X2.append(x)
       YRe2.append(cos(w2*x)+(random()-0.5)*0.3)
       YIm2.append(sin(w2*x)+(random()-0.5)*0.3)
       E2.append(0.5)

# Create two workspaces of one spectrum each    
CreateWorkspace(OutputWorkspace='ws1',DataX=X+X,DataY=YRe+YIm,DataE=E+E,NSpec=2)
evolChiP, evolAngleP, imageP, dataP = MaxEnt(InputWorkspace='ws1', ComplexData=True, A=0.001, PositiveImage=True)
print ("Number of iterations dataset1 separate: "+str( len(evolAngleP.readX(0))))
CreateWorkspace(OutputWorkspace='ws2',DataX=X2+X2,DataY=YRe2+YIm2,DataE=E2+E2,NSpec=2)
evolChiP2, evolAngleP2, imageP2, dataP2 = MaxEnt(InputWorkspace='ws2', ComplexData=True, A=0.001, PositiveImage=True)

print ("Number of iterations dataset2 separate: "+str( len(evolAngleP2.readX(0))))
# Combine the two workspaces
CreateWorkspace(OutputWorkspace='wsCombined',DataX=X+X2+X+X2,DataY=YRe+YRe2+YIm+YIm2,DataE=E+E2+E+E2,NSpec=4)
evolChiC, evolAngleC, imageC, dataC = MaxEnt(InputWorkspace='wsCombined', ComplexData=True, A=0.001, PositiveImage=True)

print ("Number of iterations dataset1 combined: "+str( len(evolAngleC.readX(0))))
print ("Number of iterations dataset2 combined: "+str( len(evolAngleC.readX(1))))
```

Framework/Algorithms/src/MaxEnt.cpp
Anders-Markvardsen: @KarlPalmen could you add a comment above this line with your explanation above

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -53,12 +53,15 @@ const double THRESHOLD = 1E-6;
...
-* @param maxIt :: [input] The max number of iteration this alg used
+* @param itCount [input] The number of iterations this alg used for each
+* spectrum
+* @param maxItCount :: [input] The max number of iteration this alg used
 * @param yLabel :: [input] y-label to use for returned ws
 * @return : ws cut down in lenght to maxIt
 */
-MatrixWorkspace_sptr removeZeros(const MatrixWorkspace_sptr &ws,
-                                 const size_t maxIt,
+MatrixWorkspace_sptr removeZeros(MatrixWorkspace_sptr &ws,
```
Anders-Markvardsen: @KarlPalmen thanks for this implementation and also improvements to same of the variable names used.

Looking at this method and where you have introduced resize() to adjust spectra length, I am wondering if there is a need for this method creating a new workspace? Instead could this method just resize the spectra of the existing input workspace? This should mean 1) avoid creation of an addition workspace 2) potentially less code

Anders-Markvardsen:
Following these steps

* start MantidPlot
* Load the Usage example data set MUSR00015189
* Rebin MUSR00015189_1 with params: 1, 0.1, 5
* Run MaxEnt on rebinned data

This works fine both on a Windows nightly build and on from this branch running on Ubuntu.

Next I tried:

* restart MantidPlot
* Load the Usage example data set MUSR00015189
* Run MaxEnt on MUSR00015189_1: this will fail with "Input data must have non-zero errors" but that is OK
* Rebin MUSR00015189_1 with params: 1, 0.1, 5
* Run MaxEnt on rebinned data

This also works fine on nightly and from this branch.

@AnthonyLim23 can you be more specific on how you get it to fail?

AnthonyLim23:
you have to run the script from my previous post first and then try maxent on some other data. Its fine if you just run it straight onto muon data

Anders-Markvardsen:
@AnthonyLim23 OK, now I can reproduce it. Specific I do

* start MantidPlot
* Run script in Anthony's comment above
* Load the Usage example data set MUSR00015189
* Rebin MUSR00015189_1 with params: 1, 0.1, 5
* Run MaxEnt on rebinned data

and I get " Input WorkspaceGroup found 
Unexpected exception: blocksize undefined because size of histograms is not equal", Anthony which you report above.

For double sanity just tested and the above steps works on a nightly build from today.

@KarlPalmen note ragged workspaces with different length spectra is a relative new feature in Mantid. Please debug

KarlPalmen:
I too get such an exception, but the new user friendly exception GUI has prevented me from reaching the exception with the debugger as I've been able to do so before.

KarlPalmen:
Luckily I found it quite quickly this morning in `Workspace2D.cpp`; first file I searched.

KarlPalmen:
After running it on a different build of Mantid, I found that all the workspaces created were added to a menu of possible input workspaces. I believe the exception arose when it attempted to add a ragged workspace to that list. I now think the code that compiles that list should `try` to add each workspace and if it `catch`-es an exception, it does not add that workspace to the list.

Framework/Algorithms/src/MaxEnt.cpp
KarlPalmen: The presence of separate input and output workspaces did confuse me and caused delay in making this change.

Anders-Markvardsen (approved):
In addtion to fixing this issue, thanks also for fixing bug in EqualBinSizesValidator.cpp, remove the need to create a new workspace in removezeros(), some code clean-ups and adding a new usage example for complex data

Anders-Markvardsen:
@KarlPalmen For the bug you fixed for EqualBinSizesValidator could you separate create an issue to add a unit to EqualBinSizesValidator that test this is now working, and initially assign it to yourself

SimonHeybrock:
There is a merge conflict.

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -73,19 +74,19 @@ MatrixWorkspace_sptr removeZeros(const MatrixWorkspace_sptr &ws,
     label->setLabel("Number of Iterations", "");
   }
 
-  if (maxIt == ws->readY(0).size()) {
-    return ws;
-  }
   const size_t nspec = ws->getNumberHistograms();
-  MatrixWorkspace_sptr outWS =
-      WorkspaceFactory::Instance().create(ws, nspec, maxIt, maxIt);
+  if (itCount.size() == 0) {
```
martyngigg: It is generally better practice to use `itCount.empty()` instead of checking for a `size==0`.

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -73,19 +74,19 @@ MatrixWorkspace_sptr removeZeros(const MatrixWorkspace_sptr &ws,
...
+    return ws; // In case, we don't have any spectra
+  }
   for (size_t spec = 0; spec < nspec; spec++) {
-    outWS->setPoints(spec, Points(maxIt, LinearGenerator(0.0, 1.0)));
-    auto Data = ws->readY(spec);
-    outWS->setCounts(spec,
-                     std::vector<double>(Data.begin(), Data.begin() + maxIt));
+    auto &DataX = ws->dataX(spec);
```
martyngigg: Variable names here should start with a lower case letter -> `dataX`

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -53,12 +53,13 @@ const double THRESHOLD = 1E-6;
...
 * @param yLabel :: [input] y-label to use for returned ws
 * @return : ws cut down in lenght to maxIt
 */
-MatrixWorkspace_sptr removeZeros(const MatrixWorkspace_sptr &ws,
-                                 const size_t maxIt,
+MatrixWorkspace_sptr removeZeros(MatrixWorkspace_sptr &ws,
+                                 const std::vector<size_t> &itCount,
                                  const ::std::string yLabel) {
 
   ws->setYUnitLabel(yLabel);
```
martyngigg: I know the code below regarding the units was not changed here but it looks odd. "Number of Iterations" is not a valid unit identifier so this will always go through the exception. I would suggest removing the `try` and just create the "Label" unit as in the catch case.

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -53,12 +53,13 @@ const double THRESHOLD = 1E-6;
...
+* @param itCount [input] The number of iterations this alg used for each
+* spectrum
 * @param yLabel :: [input] y-label to use for returned ws
 * @return : ws cut down in lenght to maxIt
 */
-MatrixWorkspace_sptr removeZeros(const MatrixWorkspace_sptr &ws,
-                                 const size_t maxIt,
+MatrixWorkspace_sptr removeZeros(MatrixWorkspace_sptr &ws,
+                                 const std::vector<size_t> &itCount,
                                  const ::std::string yLabel) {
```
martyngigg: Can we remove the `::` prefix from `std` and use a `const std::string &` here?

martyngigg (changes requested):
I've got just a couple of minor tidy up comments to add.

Framework/Algorithms/src/MaxEnt.cpp
KarlPalmen: Please comment on code above. That is what GitHub displays in the pull request.

martyngigg: @KarlPalmen Sorry for not being clear - the way the code is displayed makes it hard to comment on lines that have not changed. I was referring to [lines 66-75](https://github.com/mantidproject/mantid/blob/ef9307e26a46c626414ff0b43100d47e808a9452/Framework/Algorithms/src/MaxEnt.cpp#L66) with the `try...catch` around setting the units.

martyngigg (approved):

PR #22168, opened 2018-03-21
Reduce potential triggers for building code on build servers (Linux/macOS)

martyngigg:
Description of work.

As we move the developer documentation into the repository it is becoming more likely that pull requests will be submitted with purely documentation changes . This PR adds further checks so the code + unit tests is only built when necessary.

**To test:**

* The builds should pass
* Review the new logic thoroughly (twice to be sure!)

*No issue*

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/buildconfig/Jenkins/check_for_changes
+++ b/buildconfig/Jenkins/check_for_changes
@@ -18,22 +18,18 @@ case "$TYPE" in
         exit $(git diff --quiet ${BRANCH_TIP} ${BRANCH_BASE} -- \*\.py)
     ;;
     cpp)
-        if $(! git diff --quiet ${BRANCH_TIP} ${BRANCH_BASE} -- \*\.h)
-        then
-            exit $FOUND
-        fi
-        if $(! git diff --quiet ${BRANCH_TIP} ${BRANCH_BASE} -- \*\.cpp)
-        then
-            exit $FOUND
-        fi
-        if $(! git diff --quiet ${BRANCH_TIP} ${BRANCH_BASE} -- \*\.cxx)
-        then
-            exit $FOUND
-        fi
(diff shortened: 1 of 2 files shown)


buildconfig/Jenkins/check_for_changes
```diff
@@ -46,6 +46,26 @@ case "$TYPE" in
 	    exit $NOTFOUND
 	fi
     ;;
+    dev-docs-only)
+	# FOUND=1 iff changes are limited to dev-docs only
+	# Find all changed files and grep for required type. -v inverts match so grep=0 means
+	# there are other changes besides this
+	if git diff --name-only ${BRANCH_TIP} ${BRANCH_BASE} -- | grep -q -E -v '^dev-docs/'; then
```
peterfpeterson: Why not
```shell
if $(! git diff --quiet ${BRANCH_TIP} ${BRANCH_BASE} -- dev-docs)
then
  exit $FOUND
fi
exit $NOTFOUND
```
Someone more thoughtful could probably make it a one-liner.

peterfpeterson (commented):
I *think* this does the right thing.

buildconfig/Jenkins/check_for_changes
peterfpeterson: Even shorter option (for the lines above too):
```shell
git diff --quiet ${BRANCH_TIP} ${BRANCH_BASE} -- dev-docs || return $FOUND
return $NOTFOUND
```
Then the things earlier could be single lines (for finding) that ends in `$NOTFOUND`. It's clearer to me, but maybe not others.

martyngigg: Thanks for these. The script looks a little cleaner now.

peterfpeterson (approved):
LGTM

Issue #22063, opened 2018-03-13
Umbrella Issue for Fixing Unreliable Tests 


martyngigg:
Some Unit/doc tests have been [sporadically failing](https://docs.google.com/spreadsheets/d/1qs81x3ZDDxvEu3H5Zg1KN8Qfu54dIVWKI2f3-zxFaFg/edit#gid=2043333561). These need to be addressed.

The list to look into:

* DownloadInstrumentTest.test_exec - 
* algorithms/AlignComponents-v1.rst - @martyngigg
* docs.algorithms/AlignComponents-v1.rotation - @martyngigg 
* mantidqt.test_jupyterconsole.test_jupyterconsole - 
* ICatTest_CatalogGetDataFiles - 
* Document: algorithms/IndirectDiffScan-v1 - @brandonhewer #22066
* DataObjectsTest.WorkspaceCreationTest - @martyngigg 

See the linked spreadsheet above for more information on the problems.

To pick up a failure please append your name to the appropriate line in the list above and reference this issue in the pull request.

martyngigg:
I think we've addressed these for now.

PR #22170, opened 2018-03-21
Fixes race condition in MatrixWorkspace::rebuildDetectorIDGroupings

martyngigg:
Description of work.

See commit message.

**To test:**

* Code review and ensure the tests pass.

Refs #22063 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/src/MatrixWorkspace.cpp
+++ b/Framework/API/src/MatrixWorkspace.cpp
@@ -2000,8 +2000,8 @@ void MatrixWorkspace::rebuildDetectorIDGroupings() {
   const auto &allDetIDs = detInfo.detectorIDs();
   const auto &specDefs = m_indexInfo->spectrumDefinitions();
   const auto size = static_cast<int64_t>(m_indexInfo->size());
-  std::atomic<bool> parallelException{false};
-  std::string error;
+  enum class ErrorCode { None, InvalidDetIndex, InvalidTimeIndex };
+  std::atomic<ErrorCode> errorValue(ErrorCode::None);
 #pragma omp parallel for
   for (int64_t i = 0; i < size; ++i) {
     auto &spec = getSpectrum(i);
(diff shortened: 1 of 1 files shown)


Framework/API/src/MatrixWorkspace.cpp
```diff
@@ -2014,22 +2014,29 @@ void MatrixWorkspace::rebuildDetectorIDGroupings() {
...
+        errorCode = 2;
       } else {
         detIDs.insert(allDetIDs[detIndex]);
       }
     }
     spec.setDetectorIDs(std::move(detIDs));
   }
-  if (parallelException)
-    throw std::invalid_argument(error);
+  if (parallelException) {
```
SimonHeybrock: Can we remove `parallelException`? Check for non-zero `errorCode` enough?

SimonHeybrock (changes requested):
Code looks good, but according to one of the related tests a wrong exception is thrown now. I had a look an cannot see why right now.

See comment below for possible cleanup.

Framework/API/src/MatrixWorkspace.cpp
martyngigg: Good point - that should work.

SimonHeybrock:
@quantumsteve If I am not mistaken, the parallel macros work only within algorithms and that is the reason for the hand-written OMP loop here.

SimonHeybrock (approved):

martyngigg:
The usual OMP macros expect 2 private member variables of Algorithm to be accessible - I'm not sure why the first macro can't just define them locally though. The main difference here is that an exception isn't being caught and rethrown we're just setting a flag to say what should be thrown later. The macros have a `try...catch` block to catch anything so I don't think we'd gain much in those cases.

martyngigg:
@quantumsteve I've made a small update to use an enum class rather than an `int` as you suggested to given some meaning to the codes.

PR #22156, opened 2018-03-20
Fix dev-docs-html target for old Sphinx versions

martyngigg:
Description of work.

See commit message.

**To test:**

* On Ubuntu 14.04 run `make dev-docs-html` and see that is it succesful.

*No issue*

**Release Notes** 
*Does not need to be in the release notes.*

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [x] Is the code of an acceptable quality?
- [x] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)?
- [ ] Are the unit tests small and test the class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [x] Do changes function as described? Add comments below that describe the tests performed?
- [x] Do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant (user and developer) documentation been added/updated?

Does everything look good? Mark the review as **Approve**. A member of `@mantidproject/gatekeepers` will take care of it.

PR #22174, opened 2018-03-21
Fix dev-docs-html target for fc26

peterfpeterson:
Fix call in `python -m sphinx` branch. The issue is how the `SPHINX_BUILD` variable is expanded in the custom target. Duplicating the `add_custom_target` call was the easiest solution.

The bug was accidentally introduced in #22156.

**To test:**

`cmake --build , --target dev-docs-html` on your platform of choice.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


martyngigg (approved):
We can live with the duplication to avoid cmake's quirky parsing rules.

PR #22194, opened 2018-03-25
SofQW3 segfault

### LINKED ISSUE #22191: SofQW3 segfault
There is an out-of-bounds error in the new `calcTrapezoidYIntersections` code introduced into `FractionalRebinning.cpp` in order to make the `SofQW3` and `Rebin2D` algorithms run faster. This leads to a segmentation fault which crashes Mantid (without an error message). 

### Expected behavior

`SofQW3` runs and gives a rebinned workspace

### Actual behavior

MantidPlot crashes without any error message.

```
ws = CreateSampleWorkspace(binWidth = 0.75, XMin = 0, XMax = 300, XUnit = 'DeltaE')
ws = ScaleX(ws, -150, "Add")
ws_sqw = SofQW3(ws, '0,10,10', 'Direct', 150, EAxisBinning='60,0.75,90')
```

### Platforms affected

All

mducle:
Fixes an out-of-bounds index error in `FractionalRebinning` which causes a segmentation fault when `SofQW3` is called with certain parameters.

**To test:**

Run the following script:

```
ws = CreateSampleWorkspace(binWidth = 0.75, XMin = 0, XMax = 300, XUnit = 'DeltaE')
ws = ScaleX(ws, -150, "Add")
ws_sqw = SofQW3(ws, '0,10,10', 'Direct', 150, EAxisBinning='60,0.75,90')
```

And check that no segfault occurs.

Fixes #22191. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/DataObjects/src/FractionalRebinning.cpp
+++ b/Framework/DataObjects/src/FractionalRebinning.cpp
@@ -203,7 +203,7 @@ void calcTrapezoidYIntersections(
   const double NaN = std::numeric_limits<double>::quiet_NaN();
   const double DBL_EPS = std::numeric_limits<double>::epsilon();
   std::vector<double> leftLim((nx + 1) * (ny + 1), NaN);
-  std::vector<double> rightLim((nx + 1)* (ny + 1), NaN);
+  std::vector<double> rightLim((nx + 1) * (ny + 1), NaN);
   auto x0_it = xAxis.begin() + x_start;
   auto x1_it = xAxis.begin() + x_end + 1;
   auto y0_it = yAxis.begin() + y_start;
--- a/Framework/DataObjects/src/FractionalRebinning.cpp
+++ b/Framework/DataObjects/src/FractionalRebinning.cpp
@@ -152,7 +152,7 @@ double polyArea(T &v1, T &v2, Ts &&... vertices) {
  * @param yAxis The output data vertical axis
  * @param inputQ The input quadrilateral
  * @param y_start The starting y-axis index
- * @param y_end The starting y-axis index
+ * @param y_end The ending y-axis index
  * @param x_start The starting x-axis index
  * @param x_end The ending x-axis index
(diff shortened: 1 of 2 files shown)


Framework/DataObjects/src/FractionalRebinning.cpp
```diff
@@ -185,8 +185,8 @@ void calcTrapezoidYIntersections(
     ur = V2D(xAxis[x_end], mTop * xAxis[x_end] + cTop);
   }
 
-  const size_t nx = x_end - x_start;
-  const size_t ny = y_end - y_start;
+  const size_t nx = x_end - x_start + 1;
+  const size_t ny = y_end - y_start + 1;
   const double ll_x(ll.X()), ll_y(ll.Y()), ul_y(ul.Y());
   const double lr_x(lr.X()), lr_y(lr.Y()), ur_y(ur.Y());
   // Check if there is only a output single bin and inputQ is fully enclosed
```
martyngigg: Is it correct that the next statement would now only fire if `xend==x_start` ans `y_end==y_start`?

mducle: Doh! I think I completely missed the point. I just looked at the loop in the general code (`calcGeneralIntersection`) and thought that it meant that when `[xy]_start==[xy]_end` meant there is only one bin width, and that I had misunderstood what `[xy]_end` meant. But actually the code before was right. 

What `[xy]_start == [xy]_end` means is that the *bin width* in that direction is zero, and this was what was causing the segmentation fault.

So, actually we can fix the segfault without changing so much of the code. I'll push another commit to do this...

martyngigg (commented):
@mducle Are there any extra tests that are worth adding here?

martyngigg (approved):
The extra protection avoids the crash when encountering a  bin with zero width. Good to go and is safe to include in the patch.

PR #20951, opened 2017-10-18
`SaveFocusedXYE` to write TEMP into the header

gvardany:
This PR modifies slightly how the saver writes the header part of the file for XYE format (fullprof number 10).
The request was to add the metadata value (most commonly temperature) corresponding to the diffractogram into the file. 
For temperature in particular, there exist a keyword `TEMP` that is parsed by fullprof, but only if the file starts with `XYDATA` keyword.
For the rest, the metadata information line should be commented with #.
Since the comment lines are restricted to 6, the other lines are squeezed to free a room for the new entries.
The test is modified accordingly.
See the link in the issue for more information.

**To test:**
This should be tested by someone who is familiar with XYE exports and has fullprof installed.
Make sure that your exports with this change are still readable by fullprof.

Fixes #20940 . 

**Release Notes** 

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)?
- [ ] Are the unit tests small and test the class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] Do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant (user and developer) documentation been added/updated?

Does everything look good? Mark the review as **Approve**. A member of `@mantidproject/gatekeepers` will take care of it.

PR #22205, opened 2018-03-26
Fix header for TOPAS files

peterfpeterson:
Turns out that the TOPAS format had no unit tests. This allowed
for changes introduced by #20951 to break those files.

This skips the first line of FULLPROF header if TOPAS format was requested and adds a unit test. which is mostly copied from one of the FULPROF tests.

**To test:**

A code review should be appropriate. 

**There is no associated issue.**

**Release Notes** 

This should go into the patch candidate so it does not currently have release notes.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


peterfpeterson:
The failing tests are related to an updated instrument geometry in `master` which shouldn't be seen here.

NickDraper (approved):
Minimal changes, well covered by unit tests

PR #22215, opened 2018-03-27
Fix bug when using StartLiveData through Python API

### LINKED ISSUE #22213: Listener properties are sometimes lost when using StartLiveData in python
### Expected behavior
Properties of the specified listener algorithm should be able to be set via `StartLiveData` in python. 

### Actual behavior
The listener algorithm sometimes see default values for such properties. "Sometimes" here means that it occurs for some properties and not others, but which it happens to is consistent.

For example using the following, where `InstrumentName` is a property of `KafkaEventListener`, results in `KafkaEventListener` getting the default empty string for `InstrumentName` instead of the specified value of `LARMOR`.
```python
StartLiveData(FromStartOfRun=True,
                       FromNow=False,
                       FromTime=False,
                       UpdateEvery=5,
                       Instrument='LARMOR',
                       Listener='KafkaEventListener',
                       Address='livedata.isis.cclrc.ac.uk:9092',
                       AccumulationMethod='Add',
                       PreserveEvents=True,
                       RunTransitionBehavior='Stop',
                       OutputWorkspace='outWS',
                       InstrumentName='LARMOR')
```

Use above command and set break point in `KafkaEventListener` where it uses `getProperty("InstrumentName")`, the value it gets is an empty string.

### Platforms affected
Observed on Ubuntu 16.04.

matthew-d-jones:
Fixes #22213 

*Based on release-v3.12 as candidate for inclusion in patch*

`StartLiveData` has some special treatment in `simpleapi.py` so that it sets some algorithm properties in a particular order. When the python code was updated to make use of dict `pop()` it resulted in the "Instrument" property not being removed from `kwargs` and thus being set twice, once before all other properties (correctly) and again later when `set_properties` is called. This second `setProperty` causes the Listener properties to be wiped. The fix is to always call `pop` in `handleSpecialProperty` whether or not the value of the property is required.

**To test:**

Depends on python dictionary order so not easy to test, please review code.

You could also try running the following:
```python
StartLiveData(FromStartOfRun=True,
                       FromNow=False,
                       FromTime=False,
                       UpdateEvery=5,
                       Instrument='LARMOR',
                       Listener='KafkaEventListener',
                       Address='livedata.isis.cclrc.ac.uk:9092',
                       AccumulationMethod='Add',
                       PreserveEvents=True,
                       RunTransitionBehavior='Stop',
                       OutputWorkspace='outWS',
                       InstrumentName='LARMOR')
```

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/PythonInterface/mantid/simpleapi.py
+++ b/Framework/PythonInterface/mantid/simpleapi.py
@@ -266,6 +266,10 @@ def StartLiveData(*args, **kwargs):
         try:
             if value is None:
                 value = kwargs.pop(name)
+            else:
+                # We don't need the value, but still need to remove from kwargs
+                # so that this property isn't set again later
+                kwargs.pop(name, None)
             algm.setProperty(name, value)
 
         except ValueError as ve:


Framework/PythonInterface/mantid/simpleapi.py
```diff
@@ -266,6 +266,10 @@ def handleSpecialProperty(name, value=None):
         try:
             if value is None:
                 value = kwargs.pop(name)
+            else:
+                # We don't need the value, but still need to remove from kwargs
+                # so that this property isn't set again later
+                kwargs.pop(name, None)
```
MikeHart85: Arguably no need for the `else:`... could just always do this one indent up (which is how it was before!). But it makes no difference either way. The comment definitely helps, I can see how it wasn't clear without it.

No changes needed.

MikeHart85 (approved):
Looks good, works as intended.

No changes needed. Good to merge.

PR #19320, opened 2017-04-07
Make ILiveListeners Algorithm Aware

OwenArnold:
Rather than having the LiveListenerFactories work with the ILiveListeners
as though they were just IPropertyManagers (updatePropertyValues). They
now call a specific `setAlgorithm` method. This currently just forwards to
`IPropertyManager::updatePropertyValues` for the `LiveListener` base class, but this can be overridden.

**To test:**

* Code review
* Ensure changes are fit for purpose
* Test sample listeners

Fixes #19319

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [x] Is the code of an acceptable quality?
- [x] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)? Is it well structured with small focussed classes/methods/functions?
- [x] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [x] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [x] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant documentation been added/updated?
- [ ] Is user-facing documentation written in a user-friendly manner?
- [ ] Has developer documentation been updated if required?

- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

PR #22229, opened 2018-03-28
Remove duplicate instrument name property from kafka event listener

### LINKED ISSUE #19342: Avoid duplicating StartLiveData's Instrument property in KafkaEventListener
`StartLiveData` has an `Instrument` property but it is not accessible to `KafkaEventListener` where it is needed to subscribe to the correct topics (data stream). The workaround for this was to have a second `InstrumentName` property in `KafkaEventListener`. Since Owen's changes in #19320 this can be cleaned up by overriding `setAlgorithm` in `KafkaEventListener`, something like this:
```
void KafkaEventListener::setAlgorithm(const API::IAlgorithm &callingAlgorithm) {
  this->updatePropertyValues(callingAlgorithm);
  // Get the instrument name from StartLiveData so we can sub to correct topics
  if (callingAlgorithm.existsProperty("Instrument")) {
    m_instrumentName = callingAlgorithm.getPropertyValue("Instrument");
  }
}
```

Don't forget to remove the `InstrumentName` property and handle the case in `KafkaEventListener::connect` that `m_instrumentName` wasn't set.

matthew-d-jones:
Removes "InstrumentName" property from the KafkaEventListener. Since the changes in #19320 the algorithm can instead access the "Instrument" property of StartLiveData.

**To test:**
Open Live Data Interface with ISIS as your default facility and select ZOOM as the instrument. There should not be an "InstrumentName" field in the bottom left of the interface and the algorithm should start without error using the following options:
"Starting Time": "Start of Run"
"At Run Transition": "Stop"
"Output Workspace": "outWs"
Leaving other settings as default is fine.

Fixes #19342.

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/LiveData/inc/MantidLiveData/Kafka/KafkaEventListener.h
+++ b/Framework/LiveData/inc/MantidLiveData/Kafka/KafkaEventListener.h
@@ -1,5 +1,5 @@
-#ifndef MANTID_LIVEDATA_ISISKAFKAEVENTLISTENER_H_
-#define MANTID_LIVEDATA_ISISKAFKAEVENTLISTENER_H_
+#ifndef MANTID_LIVEDATA_KAFKAEVENTLISTENER_H_
+#define MANTID_LIVEDATA_KAFKAEVENTLISTENER_H_
 //----------------------------------------------------------------------
 // Includes
 //----------------------------------------------------------------------
@@ -10,6 +10,11 @@
 //----------------------------------------------------------------------
 
 namespace Mantid {
+
+namespace API {
+class IAlgorithm;
+}
+
 namespace LiveData {
 class KafkaEventStreamDecoder;
 
@@ -38,8 +43,7 @@ class KafkaEventStreamDecoder;
  */
 class DLLExport KafkaEventListener : public API::LiveListener {
 public:
-  KafkaEventListener();
-  /// Destructor. Should handle termination of any socket connections.
+  KafkaEventListener() = default;
   ~KafkaEventListener() override = default;
 
   //----------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Framework/LiveData/inc/MantidLiveData/Kafka/KafkaEventListener.h
```diff
@@ -38,8 +43,7 @@ class KafkaEventStreamDecoder;
  */
 class DLLExport KafkaEventListener : public API::LiveListener {
 public:
-  KafkaEventListener();
-  /// Destructor. Should handle termination of any socket connections.
+  KafkaEventListener() = default;
```
OwenArnold: Very minor point (choose to ignore), but do we even need this line? As there isn't any other constructor defined, the compiler will make this anyway.

Framework/LiveData/src/Kafka/KafkaEventListener.cpp
```diff
@@ -13,21 +14,32 @@ namespace LiveData {
...
+    m_instrumentName = callingAlgorithm.getPropertyValue("Instrument");
+  } else {
+    g_log.error("KafkaEventListener requires Instrument property to be set in "
+                "calling algorithm");
+  }
 }
 
 /// @copydoc ILiveListener::connect
 bool KafkaEventListener::connect(const Poco::Net::SocketAddress &address) {
+  if (m_instrumentName.empty()) {
```
OwenArnold: What happens if the instrument name doesn't match a topic?

OwenArnold (commented):
Generally, looks ok to me. Few minor comments.

Is this particularly hard to add unit tests for?

Framework/LiveData/src/Kafka/KafkaEventListener.cpp
matthew-d-jones: This is checked by the consumer, an error is logged and the algorithm terminates.

matthew-d-jones:
I don't think any of the listener algorithms are tested, probably because injecting mocks for the network interface at this point is difficult. However `KafkaEventListener` itself does very little, `KafkaEventStreamDecoder` does all the work and is well tested.

OwenArnold (approved):

PR #22231, opened 2018-03-28
Fix build for glibc 2.27 Replace HUGE with numeric_limits

rosswhitfield:
[`HUGE` was removed from math.h in glibc 2.27](https://sourceware.org/ml/libc-announce/2018/msg00000.html). This changes to using `std::numeric_limits<float>::max()` which should be the same value as HUGE.

**To test:**
Code review should do as I don't believe any OSs are currently using glibc `2.27`. I found this testing Ubuntu 18.04 Beta which is using `2.27`.

**Release Notes** 
*Does not need to be in the release notes. Internal change only*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


quantumsteve (approved):
Verified that `HUGE==std::numeric_limits<float>::max()`.

Strange that it is stored in a `double`.

PR #22234, opened 2018-03-28
Updated if statement in ExtractMask to work with MaskWorkspaces

### LINKED ISSUE #22233: ExtractMask returning empty list of detectors when passed a masked workspace
The ExtractMask algorithm is currently returning an empty list of detectors when it is passed a mask workspace.

Matthew-Andrew:
This PR updates ExtractMask.cpp so that it detects masked pixels in a masked workspace to return to a non empty list

**To test:**

- Code Review

- Ensure that new system tests pass

Fixes #22233  

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


reimundILL:
Are release notes & documentation missing?

Framework/Algorithms/test/ExtractMaskTest.h
```diff
@@ -59,6 +60,67 @@ class ExtractMaskTest : public CxxTest::TestSuite {
...
+    AnalysisDataService::Instance().add(inputName, inputWS);
+    MaskWorkspace_sptr inputWSMask = runExtractMask(inputName);
+    std::vector<Mantid::detid_t> expectedDetectorList = {1,  6,  11, 21,
+                                                         24, 31, 38, 41};
+
+    std::vector<Mantid::detid_t> detectorList;
+
+    TS_ASSERT_THROWS_NOTHING(
+        detectorList = runExtractMaskReturnList(inputWSMask->getName()));
+    TS_ASSERT_EQUALS(detectorList, expectedDetectorList)
```
reimundILL: Spontaneously I would check sizes match and then loop over the elements of the vectors?

reimundILL: That's cool, I did not know it.

reimundILL:
I do tests with the script

``` Python
ws = CreateSampleWorkspace()
det = ws.getDetector(2).getID()  #102
ws.maskDetectors(det)
mask = ExtractMask(ws)
```

and see on master

screenshot from 2018-04-12 11-26-02

I hope this test makes sense, ... bank 2 has detector 201 masked. The mask workspace is the output workspace of the ExtractMask algorithm.

reimundILL:
It would be good to merge master at this point since the instrument view is otherwise not correct for rectangular detectors ...

reimundILL:
The changes made seem to not change the execution (outputs) of the ExtractMask algorithm.

reimundILL:
This is expected behavior, the changes look good.

reimundILL:
I cannot get the detector list in Python.

Running
```Python
mask, list = ExtractMask(ws)
```
executes fine but no list is returned. I would expect a list here.

Mantid-Matthew:
What script are you running ? I just tried

from mantid import simpleapi

ws = CreateSampleWorkspace()
det = ws.getDetector(2).getID()  #102
ws.maskDetectors(det)
mask, list = ExtractMask(ws)
print(list)

And got the result [201] on branch which I think is correct ?

reimundILL:
That's correct, you're right. I expected it to appear in MantidPlot as workspace somehow. I imagined a Python script user independent from a MantidPlot GUI user and that both should get access to the masked detector IDs. Like the ws that you request in Python but receive in MantidPlot.

This is my understanding when reading the release notes. I would add this list to the ADS or modify the release notes in terms of returned -> where.

Mantid-Matthew:
Just checked on master and list isn't returned to the ADS there so for consistency I will leave it as is and modify the release notes to make what it does clearer.

Would this be acceptable ?

reimundILL:
I think the documentation needs an update as well for the usage example. Just copy the code from the comments. It seems to be ok that the list is not in the ADS and the release note is then already fine.

vreimund (approved):
I am very happy about the documentation changes which will make this algorithm easy to use.
Super unit tests.

reimundILL:
Some more thoughts about this issue
- are you unit testing with a MaskWorkspace or a masked workspace?
- if detector IDs are unique, it would be nice to warn the user during execution, if several times the same ID is returned like now (several zeros)

Mantid-Matthew:
- For the new tests one is testing with MaskWorkspaces and one with a masked workspace. 

- Under what conditions is the algorithm returning several zeros ?

reimundILL:
Then I can test this PR successfully with this script:
```Python
ws = CreateSampleWorkspace()
det = ws.getDetector(2).getID() #102
ws.maskDetectors(det)
mask, list = ExtractMask(ws)
print(list)
mask2, list2 = ExtractMask(mask)
print(list2)
```

gives
[201]
[201]

My second question was only a concern since it was possible until know. If detector IDs are unique in Mantid, we're fine. But if some developer produces wrong code it may be easier to find out what happend. I only wondered if this could make sense.

Mantid-Matthew:
Yes that script tests the PR. The same script run on master will give you 
list = [201]
list2 = []
I think.

I think detector IDs are unique for any given instrument so the list should never have duplicates. If it does it implies there is something wrong in the code somewhere.

reimundILL:
You're right, see this output on master. Ok for me as well that no further warnings are outputted. Thank you!

PR #22249, opened 2018-04-04
Incorrect angle used in conversion to Q in RRO

### LINKED ISSUE #21742: Incorrect angle used in conversion to Q in RRO
In `ReflectometryReductionOne2`, when we do summation in lambda we use [GroupDetectors](https://github.com/mantidproject/mantid/blob/master/Framework/Algorithms/src/ReflectometryWorkflowBase2.cpp#L408) to sum in lambda across a set of detectors of interest. 

The resulting summed workspace has a single spectrum with detector angle taken from the centre of the detectors of interest. The range of detectors of interest is often centered on the scattering angle `twoTheta`, but not always, so the final detector angle is somewhat arbitrary. 

This angle gets used by `ConvertUnits` in the conversion to `Q` and therefore we get different X values depending on the set of detectors used. Instead, the correct angle to use in the conversion to `Q` is half the scattering angle `twoTheta`. Even when the range of detectors is notionally centered on the "twoTheta pixel", the pixel's centre may not be exactly at `twoTheta` anyway, so the angle is still slightly wrong.

There are a few approaches we could take, but from discussion with scientists we have agreed to correct the position of the detector group as an extra step in `ReflectometryReductionOne`. It should be corrected to the angle provided by the `ThetaIn` property, if this is given. If `ThetaIn` is not given, no correction should be done. Ideally this correction would be done in a separate algorithm because it would also be useful for ILL.

Note that this should only be done when summing in lambda. The key steps of the `ReflectometryReductionOne` algorithm when summing in lambda start [here](https://github.com/mantidproject/mantid/blob/master/Framework/Algorithms/src/ReflectometryReductionOne2.cpp#L606).  `makeDetectorWS` calls `GroupDetectors`. Then some normalisations by monitors and transmission runs are done. The new correction should be done after these normalisations, before the result of `makeIvsLam` is returned. The result of this function is then converted to Q using `ConvertUnits`.

Note that we have an algorithm called `SpecularReflectionPositionCorrect` which corrects detector positions, but this works on instrument component names so I don't think it can be used for a detector group. It may be useful to take a look at it, though.

## To test
```
import math

def printResults(name, thetaIn):
  ws=mtd[name]
  print "Expected theta:  ", thetaIn
  print "Actual theta:    ", ws.detectorTwoTheta(ws.getDetector(0)) * 180 / math.pi / 2

# OFFSPEC example - TOF workspace detectors are NOT in the correct position on load so ARE corrected by RROA
LoadISISNexus(Filename='OFFSPEC44684', OutputWorkspace='TOF_44684')
LoadISISNexus(Filename='OFFSPEC44683', OutputWorkspace='TRANS_44683')
ReflectometryReductionOneAuto(InputWorkspace='TOF_44684', ProcessingInstructions='390-410', ThetaIn=0.3, CorrectDetectors='1', DetectorCorrectionType='RotateAroundSample', WavelengthMin=1.2, WavelengthMax=14, I0MonitorIndex=1, MonitorBackgroundWavelengthMin=15, MonitorBackgroundWavelengthMax=20, MonitorIntegrationWavelengthMin=2, MonitorIntegrationWavelengthMax=14, FirstTransmissionRun='TRANS_44683', MomentumTransferStep=0.02, ScaleFactor=1, OutputWorkspaceBinned='IvsQ_binned_44684', OutputWorkspace='IvsQ_44684', OutputWorkspaceWavelength='IvsLam_44684')
printResults('IvsQ_binned_44684', 0.3)

# INTER example - TOF workspace detectors ARE in the correct position on load so are NOT corrected by RROA
LoadISISNexus(Filename='INTER00043583', OutputWorkspace='TOF_43583')
ReflectometryReductionOneAuto(InputWorkspace='TOF_43583', CorrectDetectors='0', AnalysisMode='MultiDetectorAnalysis', ProcessingInstructions='45-69', ThetaIn=1.3, MomentumTransferStep=-0.0001, ScaleFactor=1, OutputWorkspaceBinned='IvsQ_binned_43583', OutputWorkspace='IvsQ_43583', OutputWorkspaceWavelength='IvsLam_43583')
printResults('IvsQ_binned_43583', 1.3)
```

mantid-roman:
Use input property `ThetaIn` to convert to Q if summing in lambda

**To test:**

```
import numpy as np

def printResults(nameLam, nameQ, thetaIn):
  wsLam=mtd[nameLam]
  wsQ=mtd[nameQ]
  
  xLam = 4 * np.pi * np.sin(thetaIn * np.pi / 180) / wsLam.readX(0)[-1::-1]
  xQ = wsQ.readX(0)
  print np.all(xLam - xQ) == 0

# OFFSPEC example - TOF workspace detectors are NOT in the correct position on load so ARE corrected by RROA
LoadISISNexus(Filename='OFFSPEC44684', OutputWorkspace='TOF_44684')
LoadISISNexus(Filename='OFFSPEC44683', OutputWorkspace='TRANS_44683')
ReflectometryReductionOneAuto(InputWorkspace='TOF_44684',
    ProcessingInstructions='390-410',
    ThetaIn=0.3,
    CorrectDetectors='1',
    DetectorCorrectionType='RotateAroundSample',
    WavelengthMin=1.2,
    WavelengthMax=14,
    I0MonitorIndex=1,
    MonitorBackgroundWavelengthMin=15,
    MonitorBackgroundWavelengthMax=20,
    MonitorIntegrationWavelengthMin=2,
    MonitorIntegrationWavelengthMax=14,
    FirstTransmissionRun='TRANS_44683',
    MomentumTransferStep=0.02,
    ScaleFactor=1,
    OutputWorkspaceBinned='IvsQ_binned_44684',
    OutputWorkspace='IvsQ_44684',
    OutputWorkspaceWavelength='IvsLam_44684')
printResults('IvsLam_44684', 'IvsQ_44684', 0.3)

# INTER example - TOF workspace detectors ARE in the correct position on load so are NOT corrected by RROA
LoadISISNexus(Filename='INTER00043583', OutputWorkspace='TOF_43583')
ReflectometryReductionOneAuto(InputWorkspace='TOF_43583',
    CorrectDetectors='0',
    AnalysisMode='MultiDetectorAnalysis',
    ProcessingInstructions='45-69',
    ThetaIn=1.3,
    MomentumTransferStep=-0.0001,
    ScaleFactor=1,
    OutputWorkspaceBinned='IvsQ_binned_43583',
    OutputWorkspace='IvsQ_43583',
    OutputWorkspaceWavelength='IvsLam_43583')
printResults('IvsLam_43583', 'IvsQ_43583', 1.3)
```

Fixes #21742.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/src/ReflectometryReductionOne2.cpp
+++ b/Framework/Algorithms/src/ReflectometryReductionOne2.cpp
@@ -366,7 +366,8 @@ void ReflectometryReductionOne2::init() {
   // Angle correction type
   const std::vector<std::string> correctionType{"VerticalShift",
                                                 "RotateAroundSample"};
-  auto correctionTypeValidator = boost::make_shared<StringListValidator>(correctionType);
+  auto correctionTypeValidator =
+      boost::make_shared<StringListValidator>(correctionType);
   declareProperty(
       "DetectorCorrectionType", correctionType[0], correctionTypeValidator,
       "Whether detectors should be shifted vertically or rotated around the "
@@ -473,7 +474,8 @@ void ReflectometryReductionOne2::exec() {
     m_sum = false;
   }
   // Angle correction must be done if ThetaIn is set
-  m_correctAngleInLambda = !(*getProperty("ThetaIn")).isDefault() && !summingInQ();
+  m_correctAngleInLambda =
+      !(*getProperty("ThetaIn")).isDefault() && !summingInQ();
 
   // Create the output workspace in wavelength
   MatrixWorkspace_sptr IvsLam = makeIvsLam();
--- a/Framework/Algorithms/src/ReflectometryReductionOne2.cpp
+++ b/Framework/Algorithms/src/ReflectometryReductionOne2.cpp
@@ -7,7 +7,6 @@
 #include "MantidGeometry/Objects/BoundingBox.h"
 #include "MantidHistogramData/LinearGenerator.h"
 #include "MantidIndexing/IndexInfo.h"
-#include "MantidKernel/ListValidator.h"
 #include "MantidKernel/MandatoryValidator.h"
 #include "MantidKernel/StringTokenizer.h"
 #include "MantidKernel/Unit.h"
@@ -363,17 +362,6 @@ void ReflectometryReductionOne2::init() {
                       "ThetaIn", Mantid::EMPTY_DBL(), Direction::Input),
                   "Angle in degrees");
 
-  // Angle correction type
-  const std::vector<std::string> correctionType{"VerticalShift",
-                                                "RotateAroundSample"};
-  auto correctionTypeValidator =
-      boost::make_shared<StringListValidator>(correctionType);
-  declareProperty(
-      "DetectorCorrectionType", correctionType[0], correctionTypeValidator,
-      "Whether detectors should be shifted vertically or rotated around the "
-      "sample position if angle correction is applied.",
-      Direction::Input);
-
   // Processing instructions
   declareProperty(Kernel::make_unique<PropertyWithValue<std::string>>(
                       "ProcessingInstructions", "",
@@ -473,9 +461,6 @@ void ReflectometryReductionOne2::exec() {
     m_normaliseMonitors = false;
     m_sum = false;
   }
-  // Angle correction must be done if ThetaIn is set
-  m_correctAngleInLambda =
-      !(*getProperty("ThetaIn")).isDefault() && !summingInQ();
 
   // Create the output workspace in wavelength
   MatrixWorkspace_sptr IvsLam = makeIvsLam();
@@ -812,7 +797,10 @@ MatrixWorkspace_sptr ReflectometryReductionOne2::algorithmicCorrection(
 */
 MatrixWorkspace_sptr
 ReflectometryReductionOne2::convertToQ(MatrixWorkspace_sptr inputWS) {
-  if (m_correctAngleInLambda) {
+  bool const moreThanOneDetector = inputWS->getDetector(0)->nDets() > 1;
+  bool const shouldCorrectAngle =
+      !(*getProperty("ThetaIn")).isDefault() && !summingInQ();
+  if (shouldCorrectAngle && moreThanOneDetector) {
     MatrixWorkspace_sptr IvsQ = inputWS->clone();
     auto &XOut0 = IvsQ->mutableX(0);
     const auto &XIn0 = inputWS->x(0);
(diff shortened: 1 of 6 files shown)


Framework/Algorithms/src/ReflectometryReductionOne2.cpp
```diff
@@ -789,24 +790,37 @@ MatrixWorkspace_sptr ReflectometryReductionOne2::algorithmicCorrection(
...
+  bool const moreThanOneDetector = inputWS->getDetector(0)->nDets() > 1;
+  bool const shouldCorrectAngle =
+      !(*getProperty("ThetaIn")).isDefault() && !summingInQ();
+  if (shouldCorrectAngle && moreThanOneDetector) {
+    MatrixWorkspace_sptr IvsQ = inputWS->clone();
+    auto &XOut0 = IvsQ->mutableX(0);
+    const auto &XIn0 = inputWS->x(0);
+    double const theta = getProperty("ThetaIn");
+    double const factor = 4.0 * M_PI * sin(theta * M_PI / 180.0);
+    std::transform(XIn0.rbegin(), XIn0.rend(), XOut0.begin(),
```
gemmaguest: @mantid-roman As I mentioned, the Y values need reversing here too

Framework/Algorithms/src/ReflectometryReductionOne2.cpp
```diff
@@ -789,24 +790,37 @@ MatrixWorkspace_sptr ReflectometryReductionOne2::algorithmicCorrection(
...
-  auto convertUnits = this->createChildAlgorithm("ConvertUnits");
-  convertUnits->initialize();
-  convertUnits->setProperty("InputWorkspace", inputWS);
-  convertUnits->setProperty("Target", "MomentumTransfer");
-  convertUnits->setProperty("AlignBins", false);
-  convertUnits->execute();
-  MatrixWorkspace_sptr IvsQ = convertUnits->getProperty("OutputWorkspace");
-
-  return IvsQ;
+  bool const moreThanOneDetector = inputWS->getDetector(0)->nDets() > 1;
```
gemmaguest: Could we make this more generic and allow for more than one spectrum? In real usage there should only be one, but I think the algorithm in theory allows multiple detector groups which would result in multiple spectra.

gemmaguest (changes requested):
The changes on the whole look good. Just a couple of comments.

Framework/Algorithms/src/ReflectometryReductionOne2.cpp
mantid-roman: I would prefer checking and rejecting grouping to more than 1 group. For multiple groups we would need more than 1 `ThetaIn` but my understanding is that each run of an experiment has only one theta.

gemmaguest: Sure, that's probably better :+1:

gemmaguest (dismissed):

gemmaguest (approved):
This all looks good to me now :+1:

PR #22269, opened 2018-04-10
New Algorithm Categories Page in Documentation

### LINKED ISSUE #22254: Change Algorithm doc landing page to categories
This is to change the landing page for documentation to be the base category list

it will also:

1. contain a link to the full alphabetical list on the first landing page.
1. neaten up the list handling when there are few items to list.

keeeto:
This PR (referring to issue #22254) contains work to build a page which indexes the algorithm categories. The Algorithm link in the documentation now goes to the index of categories, rather than to the full list of all algorithms.

There is a block of introductory text on the Algorithm category index to explain what the page is, how categories are decided and approximately what one should expect to find in the documentation of an algorithm.

There is still a link to go to the list of all algorithms if desired.

**To test:**

Build Mantid. Build the html docs:

In `build/docs/` run : `python runsphinx_html.py`

* Check that the documentation builds.
* Check that the `Algorithms` link takes you to an index of algorithm categories.
* Check the intro text on `Algorithm Index` page.
* Check that linking from the `Algorithm Index` page works properly.
* Check that no other documentation has been affected e.g. `concepts/index.html`
 
Resolves #22254 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


docs/source/_templates/algorithm_categories.html
```diff
@@ -0,0 +1,21 @@
+<blockquote> Welcome to the Mantid Algorithm Categories page. This page provides a list of the categories into which the algorithms are split.
```
DanNixon: Is there a reason this is a block quote?

It looks inconsistent with other "next level down" pages (e.g. Training and Fitting Overview).

docs/source/index.rst
```diff
@@ -41,7 +41,7 @@ This is the documentation for Mantid |release|.
    :width: 200px
    :align: right
 
-* :ref:`Algorithms List`
+* `Algorithms List <algorithm%20index/Algorithms/Algorithms.html>`_
```
DanNixon: This should probably just say "Algorithms" now that it does not link to the full list.

DanNixon (changes requested):
The page appears but does not list the categories for me (testing on Ubuntu 16.04).

keeeto:
@DanNixon I did look at those comments, although now I see another one about blockquotes. I'll address that tomorrow and then resubmit. I'll give you a shout when its in.

DanNixon:
I'm tempted to say that that kind of font adjustment should be done in global styling (I haven't looked into it but there will be some CSS used for the template somewhere).

Another very minor thing (that I'm not sure is worth the effort to fix unless it is pretty simple), when you get to a category page that contains both subcategories and algorithms, having the algorithms called "Pages" is a little unintuitive (e.g. `[DOCS_ROOT]/algorithms/categories/Diffraction.html`).

keeeto:
@DanNixon you're probably right about defining the style in a CSS - I can do that for this PR.

For the category "Pages" - I know what you mean and have actually looked at this previously. The problem is that the template is used for all kinds of docs pages, eg Algorithms, Interfaces, Examples etc. Trying to find a general, yet expressive label that works for all of these is not easy. One solution is to define this heading separately based on the page name. This is probably not hard, but would take up more time than I think it is worth for this PR. I'd suggest creating a separate issue, which I will pick up afterwards.

docs/source/fitting/index.rst
```diff
@@ -6,7 +6,7 @@ Fitting Overview
 
 .. raw :: html
 
-    <blockquote>This is technical documentation relating to fitting capabilities in Mantid. The fitting 
+    <intro>This is technical documentation relating to fitting capabilities in Mantid. The fitting
```
DanNixon: I'd rather not use non standard element names.

This should either be a style that can be applied to any element as required, e.g.:
```css
.smaller_text {
  font-size: 80%;
}
```

Or (possibly the better solution) if the global font sizes are genuinely an issue then they should be changed everywhere.

DanNixon (changes requested):

docs/source/fitting/index.rst
keeeto: I'm not sure that global font size is an issue - or rather there are so many different lengths of list and description that there is no one-size-fits all solution. Now, that could be part of a larger issue in that we should not really have so many different scenarios, but that in turn requires major reworking of the docs structure, which is beyond this PR.

DanNixon:
The code changes look fine now.

I'm still not that sure the visual change to font size is really needed, to me the fact that there are now different font styles across different pages looks odd. However it doesn't bother me enough to not approve this PR.

DanNixon (dismissed):

keeeto:
@DanNixon - sorry to be a pain. But I think I have a much more acceptable solution to the line length problem. I have just set a maximum column width, which means that all font sizes can remain the same.

DanNixon (changes requested):
Sorry, but I don't like the latest changes.

Only the "body" wrapping appears to have changed, not the entire page.

This means the navigation/header bar is often wider than the content on the page.

The text also now overflows the viewport when it is less than 800px.

keeeto:
So, the nav bar extending the full page width is a conscious design decision as I think it looks worse when the navbar truncates to whitespace. I am open to change on this, but it should maybe be put to a few more people as it comes down to a subjective decision.

The over running text should not be happening, I thought I set the options to wrap text and all the pages I can see seem to do so. If you can give me an example it'd be useful to help fix it.

DanNixon:
I should be more clear, the controls on the nav bar are truncated (i.e. Mantid logo, links, search box). If you resize the browser you will see there is a point at which they no longer expand with page width.

When you reduce the browser width to less than 800px the text overruns as you have overridden the responsive style rules.

I'd say the best thing to do for now is remove the style changes from this PR so that the documentation changes can be merged and raise the style issues at a review meeting.

keeeto:
Ah yes I see that now. Agreed - the style changes belong in a separate PR and a review meeting seems the appropriate forum for discussion. We can also address your point about the "Pages" headings at that time. Done.

docs/sphinxext/mantiddoc/directives/categories.py
```diff
@@ -367,10 +371,37 @@ def create_category_pages(app):
...
+
+        template = CATEGORY_PAGE_TEMPLATE
+        top_context["text_page"] = "algorithm_categories.html"
+        top_context["subcategories"] = sorted(all_top_categories, key = lambda x: x.name)
+        #top_html_path_noext = ""
+        # Set up the positions in the toc tree
+        top_category_html_dir = posixpath.join(top_category.name.lower(), 'categories')
+        top_document_dir = posixpath.dirname(top_category_html_dir)
+        top_category_html_path_noext = posixpath.join(top_document_dir, 'index')
+        top_context['outpath'] = "../" + top_category_html_path_noext + '.html'
```
DanNixon: Is the `../` required here? Just tried this again and the top level category pages do not work as they miss the top level directory out of the URL.

DanNixon (changes requested):
Just had a final test and found a couple more issues.

Also I find there are more than just algorithms in the Algorithm Index page (e.g. *Calibrate* (fit functions), *DynamicPDF* (an interface)).

keeeto:
Hi @DanNixon - it looks like you've caught a pretty fundamental issue with our taxonomy of documents there. Well done for spotting it and pointing it out. I've talked to @NickDraper and we've decided that it is probably best to re-think how this Algorithms Categories page is going to be generated. Therefore, this PR is going to be closed and a new one opened. Thanks for all the reviewing on this, it has been very helpful!

PR #22276, opened 2018-04-11
Add force-overwrite option to GSAS-II installation scripts

josephframsay:
A force-overwrite option was added to the GSAS-II installation script. When enabled, this deletes the old installation and runs a fresh one

**To test:**

Instructions are for Windows - ENGINX primarily use Windows, so it would be best to test on that.

Install GSAS-II with one of either `install_gsas_vetted.bat` or `install_gsas_latest.bat` (if at ISIS you'll need to run as your 03 account, if somewhere else you might need to run with admin privileges). Make a note of where GSAS-II was installed (unless you used the `-d` flag it will be in `C:\g2conda\GSASII`). Run either of the scripts again (it doesn't have to be the same one) with the `-f` flag. Go to the installation directory and verify that everything from before was overwritten (look at the 'Date Modified' field)

As a bonus, make sure the installed GSAS-II actually works - you can run it from `<wherever you installed GSAS-II>/g2conda/GSASII/RunGSASII.bat`

No issue associated with this PR

**Release Notes** 
No-one uses the script as yet, so improvements to it don't need to go in the release notes

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/scripts/GSAS-II/install_gsas_common.bat
+++ b/scripts/GSAS-II/install_gsas_common.bat
@@ -1,25 +1,31 @@
+setlocal enableextensions enabledelayedexpansion
+
 @echo off
 @echo Running install_gsas_latest.bat
 @echo Run with -b flag for non-interactive (quiet) mode
 @echo (don't pause after installation and install in Python user site package directory)
 
+:: find python
 @set DEV_PYTHON_EXE=%~dp0..\..\external\src\ThirdParty\lib\python2.7\python.exe
 @set RELEASE_PYTHON_EXE=%~dp0..\..\bin\python.exe
+@set PATH_PYTHON_EXE=
+python --version 2>nul
+if %ERRORLEVEL%==0 @set PATH_PYTHON_EXE=python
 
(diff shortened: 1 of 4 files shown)


martyngigg (changes requested):
The changes made here look sensible. I have a couple of other suggestions for the related scripts to make them easier to use.

The main one is that from a command line I have to be in the `scripts\GSAS-II` directory to run them or I get an error. In the `.bat` scripts such as [here](https://github.com/mantidproject/mantid/blob/master/scripts/GSAS-II/install_gsas_common.bat#L17) the files are referenced with a relative path to the current working directory. Batch scripts are able to use the `%~dp0` prefix to expand to the directory of the current file, we could use this to rewrite the above line as
```
%DEV_PYTHON_EXE% %~dp0install_gsas_proxy.py %*
```
and then similarly in all of the other .bat files. This way you can open a command prompt and then run `scripts\GSAS-II\install_gsas_latest.bat` for example.

martyngigg (approved):

PR #22281, opened 2018-04-12
ISIS Powder PEARL - revert instrument settings to default after every method call

### LINKED ISSUE #22280: ISIS Powder - PEARL object settings should not persist after method calls
Currently (on all ISIS Powder instruments), any settings which are changed in a call to either `focus`, `create_vanadium` or `create_cal` (for PEARL) are taken as the new defaults, and apply to any future method calls. PEARL don't like this behaviour, as they fiddle with their settings all the time, and it gets confusing to keep track of what settings they're operating with, especially as they tend to run their scripts in a non-linear way.

Instead, they want any settings given in the constructor (either explicitly or by a config file) to be taken as the defaults. Any settings changed in a method call should apply only for the duration of that method, and revert back to default once the method has terminated

josephframsay:
The way the PEARL object handles its settings was changed. Now, instead of every method call updating the instrument settings with its parameters, any settings given in the constructor are taken as defaults, and settings changed in a method call (ie `focus`, `create_vanadium` or `create_cal`) are only applied during execution of that method. On termination, the default settings are restored

**To test:**

If not reviewing from ISIS, you'll need to change the second line and do some setup locally - ping me a message for an explanation of how to set up ISIS_Powder directories, as it can be fiddly.

```
from isis_powder import Pearl

pearl = Pearl(config_file=r"\\olympic\Babylon5\Public\JoeRamsay\PEARL\pearl_shared_config.yaml")

pearl.focus(run_number="99504")
pearl.focus(run_number="99506", tt_mode="tt88") # Should show an warning about replacing this parameter
pearl.focus(run_number="99507")

print(pearl._inst_settings.tt_mode)
```

You should get 3 GroupWorkspaces - `PEARL99504_tt70-Results-D-Grp`, `PEARL99506_tt88-Results-D-Grp` and `PEARL99507_tt70-Results-D-Grp`. The final line should print `tt88`.

Fiddle with any parameters you like, and make sure that any that you change are reverted back to the defaults after that line has finished. Also try running likes in non-sequential order (by highlighting and `Ctrl-Enter`) - make sure that the correct settings are still used

Fixes #22280 

**Release Notes** 
[Here](https://github.com/mantidproject/mantid/compare/22280_isis_powder_pearl_default_settings?expand=1#diff-b2bd2540c595f643521322a749fbd5dc)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


scripts/Diffraction/isis_powder/pearl.py
```diff
@@ -24,9 +27,11 @@ def __init__(self, **kwargs):
     def focus(self, **kwargs):
         self._switch_long_mode_inst_settings(kwargs.get("long_mode"))
         self._inst_settings.update_attributes(kwargs=kwargs)
-        return self._focus(run_number_string=self._inst_settings.run_number,
-                           do_absorb_corrections=self._inst_settings.absorb_corrections,
-                           do_van_normalisation=self._inst_settings.van_norm)
+        focused_run = self._focus(run_number_string=self._inst_settings.run_number,
```
martyngigg: I can see that this makes sense as the additional arguments should only apply for this function call. Could the approach of having to remember to restore the settings afterwards be error prone though? What if `self._focus` throws an exception and the final line doesn't run?

Could we use a decorator or a context manager? For example a context manager would look like:

```python
def focus(self, **kwargs):
    with restore_defaults_post_process():
         self._switch_long_mode_inst_settings(kwargs.get("long_mode"))
         self._inst_settings.update_attributes(kwargs=kwargs)
         return self._focus(run_number_string=self._inst_settings.run_number, ...
```
or a decorator (but I'm not sure how this plays with exceptions)

```python
@restore_defaults_post_process
def focus(self, **kwargs):
     self._switch_long_mode_inst_settings(kwargs.get("long_mode"))
     self._inst_settings.update_attributes(kwargs=kwargs)
     return self._focus(run_number_string=self._inst_settings.run_number, ...
```

docs/source/techniques/ISISPowder-Tutorials.rst
```diff
@@ -210,6 +210,10 @@ found for each individual instrument in the reference document:
 
 How objects hold state in ISIS Powder
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. warning:: This is NOT relevant for PEARL. PEARL scientists should
```
martyngigg: Is there now a concern that we have classes deriving from the same base interface but behaving differently? I actually think the PEARL behaviour sounds more sensible - is there any hope to harmonise this?

josephframsay: This makes me a bit uncomfortable as well. I think that scripts from other instruments rely on the old behaviour (we made some changes to that effect in #21420), so I'm not sure how much I want to disrupt other instruments' workflows? I'm sort of in two minds...

scripts/Diffraction/isis_powder/pearl.py
josephframsay: Ooh yes, I thought there must be a nicer way of doing it

martyngigg (approved):

rosswhitfield:
The `SystemTests.ISIS_PowderPearlTest.*` are now failing with python 3.
http://builds.mantidproject.org/view/Master%20Pipeline/job/master_systemtests-ubuntu-16.04-python3/420/testReport/

josephframsay:
Thanks @rosswhitfield, there's now a PR on this at #22382

PR #22300, opened 2018-04-17
Small Mesh Object Fix in Validity Check

### LINKED ISSUE #22299: Bug in MeshObject::hasValisShape check
`MeshObject::hasValidShape` currently checks for at least 4 vertices and 4 triangular faces. However 4 vertices would correspond to 2 triangular faces.

LamarMoore:
**DO NOT MERGE**
Fixes a bug in `MeshObject::hasValidShape`

**To test:**

Code Review should suffice.

Fixes #22299 . 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Geometry/src/Objects/MeshObject.cpp
+++ b/Framework/Geometry/src/Objects/MeshObject.cpp
@@ -58,8 +58,8 @@ const Kernel::Material MeshObject::material() const { return m_material; }
 */
 bool MeshObject::hasValidShape() const {
   // May enclose volume if there are at
-  // at least 4 triangles and 4 vertices
-  return (numberOfTriangles() >= 4 && numberOfVertices() >= 4);
+  // at least 2 triangles and 4 vertices
+  return (numberOfTriangles() >= 2 && numberOfVertices() >= 4);
 }
 
 /**


Framework/Geometry/src/Objects/MeshObject.cpp
```diff
@@ -58,8 +58,8 @@ const Kernel::Material MeshObject::material() const { return m_material; }
 */
 bool MeshObject::hasValidShape() const {
   // May enclose volume if there are at
-  // at least 4 triangles and 4 vertices
-  return (numberOfTriangles() >= 4 && numberOfVertices() >= 4);
+  // at least 2 triangles and 4 vertices
+  return (numberOfTriangles() >= 2 && numberOfVertices() >= 4);
```
KarlPalmen: 4 triangles & 4 vertices is the simplest MeshObject that can enclose a volume. Can you please explain why the number of triangles has been reduced to 2 while the number of vertices remains 4?

KarlPalmen: If this fixes a bug, how does one produce the bug?

LamarMoore: @KarlPalmen I believe this would constitute a more robust check for a closed shape https://en.wikipedia.org/wiki/Euler_characteristic. We load a completely planar mesh with >> 4 faces and vertices and the check passes. This would obviously cause problems since it does not enclose a volume. I think the current check is insufficient.

KarlPalmen: What I see above is a less robust check. 

It could be followed by further checks such as the Euler characteristic check, although even that would not work for a surface enclosing multiple volumes.
I decided not the put a more robust check in, because it would require considerable effort. I did consider the Euler characteristic check. It would require determining and counting the edges. Also on the way one could check each edge has the 2 vertices in opposite order in the two triangles that form the edge. Another less robust possibility is to check the the volume is positive.

You are welcome to add such checks, but there impact on performance needs testing.

KarlPalmen:
I don't think will be ready for release 3.13 .

OwenArnold:
@LamarMoore I suggest closing this in light of the `MeshObject2D` that is being worked on.

OwenArnold:
This has subsequently been fixed via `MeshObject2D`. See master.

PR #22302, opened 2018-04-17
Eliminate Ranges in BB calculations

### LINKED ISSUE #21562: Eliminate Detector Exclusion Ranges from ComponentInfo::boundingBox
Deals with removing detector exclusion ranges mentioned here #21506. Currently, unique exclusion ranges are stored in a sorted container while non-detector bounding boxes are treated and then these are separately enumerated over so that detector bounding boxes can be treated. This is a little messy as it involves passing a `map` around to add to and store the exclusion ranges. 

@SimonHeybrock has suggested that the algorithm could be refactored into:
```
componentInfo::boundingBox(size_t compIndex) {
if(isDetector(compIndex))
  return growFromSingleDetector(compIndex);
auto compRange = componentRangeInSubtree(compIndex);
auto detRange = detectorRangeInSubtree(compIndex);
auto compEnd = compRange.end();
auto detEnd = detRange.end();
auto comp = compRange.begin();
auto det = detRange.begin(); 
while (comp != compEnd || det != detEnd) {
  auto dets = detectorRangeInSubtree(*comp);
  while (det != detEnd && (comp == compEnd || det < *dets.begin())) {
    growFromSingleDetector();
  }
  if (comp == compEnd)
    continue;
  if (compFlag == Tube) {
    growFromTube();
    det += dets.size();
    ++comp;
  } else if (...) {
  }
}
}
```

OwenArnold:
Fixes #21562

**To test:**

As far as I can tell with some comparative testing, this does not degrade performance, but if the master performance tests indicate otherwise (post merge) then this PR should be reverted.

**Release Notes** 

No update to release notes required

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Geometry/src/Instrument/ComponentInfo.cpp
+++ b/Framework/Geometry/src/Instrument/ComponentInfo.cpp
@@ -295,30 +295,16 @@ double ComponentInfo::solidAngle(const size_t componentIndex,
  * @param index : Index of the component to get the bounding box for
  * @param reference : Reference bounding box (optional)
  * @param mutableBB : Output bounding box. This will be grown.
- * @param mutableDetExclusions : Output detector exclusions to append to. These
- * are ranges of detector indices that we do NOT need to consider for future
- * bounding box calculations for detectors.
- * @param mutableIterator : Iterator to the next non-detector component.
  */
-template <typename IteratorT>
 void ComponentInfo::growBoundingBoxAsRectuangularBank(
     size_t index, const Geometry::BoundingBox *reference,
-    Geometry::BoundingBox &mutableBB,
-    std::map<size_t, size_t> &mutableDetExclusions,
-    IteratorT &mutableIterator) const {
+    Geometry::BoundingBox &mutableBB) const {
 
   auto panel = quadrilateralComponent(index);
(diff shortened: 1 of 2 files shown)


Framework/Geometry/src/Instrument/ComponentInfo.cpp
```diff
@@ -457,32 +392,25 @@ BoundingBox ComponentInfo::boundingBox(const size_t componentIndex,
...
-    } else {
-      // General case
-      absoluteBB.grow(componentBoundingBox(index, reference));
-      ++compIterator;
+  const auto compFlag = componentType(componentIndex);
+  if (hasSource() && componentIndex == source()) {
+    // Do nothing. Source is not considered part of the beamline for bounding
+    // box calculations.
+  } else if (compFlag == Beamline::ComponentType::Unstructured) {
+    for (const auto &childIndex : this->children(componentIndex)) {
```
SimonHeybrock: Is my understanding correct here: This `for` loop is essentially replacing the old `while` that used to be outside the `if/else` block?

SimonHeybrock (approved):
Code is much simpler and easier to understand now. Existing tests should be covering this cleanup. 👍

PR #22312, opened 2018-04-20
Architecture for Muon Analysis version 2

### LINKED ISSUE #20941: Create Full architecture for FDA
This is a PR to create the general architecture for the FDA GUI. To allow this a simple pair of dummy widgets will be made (to later be replaced by the data loader #20943 and the help button #20944).

AnthonyLim23:
Muon Analysis version 2 will be a fully MVP code. This PR is for the basic architecture of the GUI, this includes a load widget at the top, some tabs in the centre and a help widget at the bottom. To demonstrate the GUI we are only using dummy widgets (button or a label). 

This does not include any tests as the only part of the code we will keep is the dock_widget and dock_view, which just wrap a pyqt widget (so we don't need to check it works).

**To test:**

To have muon analysis 2 as an interface you will need add the following to https://github.com/mantidproject/mantid/blob/master/Framework/Properties/Mantid.properties.template#L23    `Muon/Muon_Analysis_2.py`

Then build mantid

Once it has built open mantid and open Muon Analysis 2
A GUI will appear, it will have labels at the top and bottom, then some tabs in the centre. 
If you click the square in the top right corner of the tab it should be ejected.
Eject 2 of the 3 tabs and position them so you can see them all
Press one of the buttons and the label on tab two will change (depending on which button you pressed the label will be different)
Make sure you can put the tabs back into the GUI (pressing the square in the corner should do the trick)

Eject all of the tabs
Close the GUI main window
All of the tabs should also close

Fixes #20941  

**Release Notes** 

---
Will not be available to the user, so not in the release notes
#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


scripts/Muon/dummy_label_widget.py
```diff
@@ -0,0 +1,25 @@
...
+
+    """
+    """
+
+    def __init__(self, name, parent=None):
+        view = dummy_label_view.DummyLabelView(name, parent)
+        model = None
+        self.presenter = dummy_label_presenter.DummyLabelPresenter(view, model)
+
+    def getPresneter(self):
```
martyngigg: A typo: `getPresneter` -> `getPresenter`.

Can we consider using [property accessors](https://docs.python.org/2/library/functions.html?highlight=property#property) for these types of thing and just having them acessed `.widget` and `.presenter` without the `get`prefix? It's slightly more Pythonic that way.

scripts/Muon/dummy_presenter.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class DummyPresenter(object):
+    """
+    """
+    def __init__(self,view,model):
+        self.view=view
+        self.model=model
+
+    def getWidget(self):
```
martyngigg: Similar comment as above.

scripts/Muon/dummy_presenter.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import (absolute_import, division, print_function)
+
+
+class DummyPresenter(object):
+    """
```
martyngigg: Can we remove the empty doc strings or at least add something to it?

scripts/Muon_Analysis_2.py
```diff
@@ -0,0 +1,52 @@
...
+from Muon import dummy_label_widget
+from Muon import dock_widget
+
+
+class MuonAnalysis2Gui(QtGui.QMainWindow):
+
+    def __init__(self, parent=None):
+        super(MuonAnalysis2Gui, self).__init__(parent)
+
+        loadWidget = dummy_label_widget.DummyLabelWidget("Load dummy", self)
```
martyngigg: Can you help me understand why we wouldn't use the Qt designer for a lot of the layout work?

scripts/Muon/dummy_label_widget.py
AnthonyLim23: I never knew about this. I will change the code, I agree it looks much neater

scripts/Muon_Analysis_2.py
AnthonyLim23: Personally I prefer doing it through the code, as I find it easier to understand. It also keeps everything in one place. I feel that it shouldn't make too much difference as we are primarily just adding widgets

martyngigg: Ok. I would say that one thing to consider if things get much more complicated is that maintenance in the long term can be easier with a `.ui` files as changing the layout is easier with just dragging things around.

AnthonyLim23:
yes, that is true. But at that point I think it should be multiple widgets (each tab will be its on widget and will usually be a pair of tables and some buttons)

AnthonyLim23:
MantidPlotProjectSerialiseTest keeps failing on windows and I am not sure why. @martyngigg do you have any ideas?

AnthonyLim23:
Is there anything else that I need to change for this PR?

martyngigg (commented):
I think we should have a bit more structure to the code in `scripts/Muon`. We have the `maxent` module but then everything else is just in the top-level directory and it's hard to know what goes together.  Maybe at minimum have a `gui` module with all of the GUI-related code?

AnthonyLim23:
Origianlly the MaxEnt folder was not going to exist. In the long term it will be purged. But I have considered having either folder for each major widget (in this case FFT and MaxEnt would be together) or folders for the presenters, models and views. What do you think?

martyngigg (commented):
What is the plan for the remainder of the GUI code that is in the top-level `Muon` folder?

AnthonyLim23:
I need to refactor it to match the new layout and as I do so it will move into the new folder structure.

martyngigg:
I see, that's fine.

martyngigg (approved):

AnthonyLim23:
@mantidproject/gatekeepers could we please merge this?

PR #22343, opened 2018-04-27
Stitch1D: input point data with optional dx values

### LINKED ISSUE #22197: New `OutputWorkspace` option for `Stitch1D` algorithm
For some reflectometry users it is required to stitch 1D single point data together.

This can be achieved by running `ConjoinXRuns` and `SortXAxis` consecutively for point data. And no rebinning is necessary. Some inconveniences are
- `ConjoinXRuns` needs input workspace to be in the ADS
- `SortXAxis` is a Python algorithm

The original behavior of the algorithm will not change, i.e. histograms will be treated as before.

Related to this
- Add input validation for either histogram or point data workspaces

Unrelated to this
- Update doxygen page

vreimund:
For some reflectometry users it is required to stitch 1D single point data together without loosing points.
Stitch1D offers much properties and scaling, while rebinning and weighted or unweighted mean are not desired. Please read the documentation.

This can be achieved by running `ConjoinXRuns` and `SortXAxis` consecutively for point data. Some inconveniences are
- `ConjoinXRuns` needs input workspace to be in the ADS, the output workspace property is `API::Workspace` and not `API::MatrixWorkspace` (I really like to change that since the return is a MatrixWorkspace)
- `SortXAxis` is a Python algorithm

The original behavior of the algorithm will not change, i.e. histograms will be treated as before.

*Stitch1D*
- At some point I noticed that the input workspaces got modified and considered constness hopefully everywhere where needed, added unit tests for this
- SortXAxis in c++ using map since sorted by key. The current implementation is based on multimap and performance not tested since not crucial at the moment.
- Add input validation for either histogram or point data workspaces, ... (via `RunCombinationHelper`)
- Update diagram for histogram workspaces (unweighted mean)
- Update doxygen page
- Build documentation without warnings

*Stitch1DMany*
- Input validation via `RunCombinationHelper`
- Avoid crash if next to the input workspaces only `UseManualScaleFactor` is given for input matrix workspaces not grouped
- In test: used existing helper for creating and grouping workspaces everywhere
- All tests can be run independently now
- Did not use `checkGroups` and `processGroups` any more, which may result in a different algorithm history (? the original test showed always single calls of `Stitch1DMany` but maybe not the full history was checked, size check was missing for example to show how many history entries were there) revealing the recursive call of `Stitch1DMany`

It may be worth to improve the documentation of `Stitch1DMany`. It can be described simpler and clearer. The points listed in workflow do not distinguish well between groups or non groups. The diagram does not show clearly the recursion. This would, however, help understanding.

**To test:**

In general, the existing behavior should not have changed but missing behavior added and some exisiting problems handled.

For example:
```Python
ws1 = CreateWorkspace(DataX=[1.,2.,3.], DataY=[1.,2.,3.], DataE=[3.,2.,1.], Dx=[3.,4.,3.])
ws2 = CreateWorkspace(DataX=[1.5,2.5,3.5], DataY=[5.,6.,7.], DataE=[8.,9.,10.], Dx=[9.,8.,9.])
stitched = Stitch1D(ws1, ws2, UseManualScaleFactor=1)

stitched_as_well = Stitch1DMany(InputWorkspaces='ws1,ws2', UseManualScaleFactors=True, ManualScaleFactors='1')

CompareWorkspaces('stitched', 'stitched_as_well')
```

Fixes #22197.

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


reimundILL:
I will open a new issue for the `Stitch1DMany` algorithm since:
- no input validation if point data or histogram data (it works for point data now!)
- update of documentation
- segfaulting when giving manual scale factors (a list)
- tooltips not fully visible (this I will be added to this PR)
- maybe more

reimundILL:
The following modifications are not functional changes of `Stitch1DMany`, only the now allowed empty bool for `Params` input validation is different and intended for point data users. If `Params` is given can be checked later for histogram users (not done yet).

However, on master I find this behavior which needs attention in my opinion: (this code originates partly from a unit test and quickly translated to Python):
```Python
import numpy
x = [0.1,0.2,0.3,0.4,0.5,0.6,0.7,0.8,0.9,1.0,1.1,0.1,0.2,0.3,0.4,0.5,0.6,0.7,0.8,0.9,1.0,1.1]
y = [1.,1.,1.,1.,1.,1.,1.,1.,1.,1.,1.,1.,1.,1.,1.,1.,1.,1.,1.,1.,1.,1.]
ws1 = CreateWorkspace(x,y,y,2)
y = 2. * numpy.ones(20)
ws2 = CreateWorkspace(x,y,y,2)
g1 = GroupWorkspaces([ws1, ws2])
y = 3. * numpy.ones(20)
x = [0.8,0.9,1.0,1.1,1.2,1.3,1.4,1.5,1.6,1.7,1.8,0.8,0.9,1.,1.1,1.2,1.3,1.4,1.5,1.6,1.7,1.8]
ws3 = CreateWorkspace(x,y,y,2)
y = 4. * numpy.ones(20)
ws4 = CreateWorkspace(x,y,y,2)
g2 = GroupWorkspaces([ws3, ws4])
y = 5. * numpy.ones(20)
x = [1.6,1.7,1.8,1.9,2.,2.1,2.2,2.3,2.4,2.5,2.6,1.6,1.7,1.8,1.9,2.,2.1,2.2,2.3,2.4,2.5,2.6]
ws5 = CreateWorkspace(x,y,y,2)
y = 6. * numpy.ones(20)
ws6 = CreateWorkspace(x,y,y,2)
g3 = GroupWorkspaces([ws5, ws6])
s = Stitch1DMany(InputWorkspaces="g1, g2, g3", Params="0.1, 0.1, 2.6")

```

and get the error

```Python
ValueError: The workspace "s_ws1_ws3_ws5" is not in the workspace list.
```

Actually, the ADS has now `g1`, `g2`, `g3`, and `s_ws2_ws4_ws6` available.
This is on master and on my branch the same error message.

reimundILL:
This causes a crash:
```Python
ws1 = CreateWorkspace(DataX=[1.,2.,3.], DataY=[1.,2.,3.], DataE=[3.,2.,1.], Dx=[3.,4.,3.])
ws2 = CreateWorkspace(DataX=[1.5,2.5,3.5], DataY=[5.,6.,7.], DataE=[8.,9.,10.], Dx=[9.,8.,9.])
stitched = Stitch1DMany(InputWorkspaces='ws1,ws2', UseManualScaleFactors=True)
```
I think this is master behavior as well and needs better input validation in `Stitch1DMany`: `ManualScaleFactors` not set

ianbush (changes requested):
This is looking good, the new features are there and seem to be working as expected, I have added a few comments on the code.

I have done a bit of functional testing, I would be reasonably sure this does not break anything existing as all the tests pass, but it would be good if someone else who knows reflectometry could comment - perahps @mantid-roman or @gemmaguest ? I think @soininen had some ILL examples perhaps he could test this with too.

Framework/Algorithms/test/Stitch1DManyTest.h
```diff
@@ -147,182 +153,175 @@ class Stitch1DManyTest : public CxxTest::TestSuite {
...
   void test_throws_with_too_few_workspaces() {
+    createUniformWorkspace(0.1, 0.1, 1., 2., "ws1");
     Stitch1DMany alg;
     alg.setChild(true);
     alg.initialize();
     alg.setProperty("InputWorkspaces", "ws1");
     alg.setProperty("Params", "0.1, 0.1, 1.8");
-    alg.setPropertyValue("OutputWorkspace", "outws");
+    alg.setProperty("OutputWorkspace", "outws");
     TS_ASSERT_THROWS(alg.execute(), std::runtime_error);
```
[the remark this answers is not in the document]
reimundILL: **Good point**, I thought about it earlier (sorry for the long answer):

This was the intention to be more clear in `ConjoinXRunsTest` but I am unhappy now about this test: it tests input validation that always throws `std::runtime_error` with message `Some invalid properties found`. Let's say less informative?

To the contrary, if the algorithm throws, it is in my opinion important to test if exactly this error throws with the expected message. That is why I think in most code input validation messages do not get checked (?). On the other hand side, we can be really clear that input validation throws and nothing else -> what else can throw here? If the algorithm itself does not throw `std::runtime_error`s there is less possibility and it can only be input validation? However, `Stitch1D` throws `std:runtime_errors`.

Finally, `Some invalid properties found` does not tell us if it really is the `too_few_workspaces` thing the unit test wants to test. (Can this be improved?). Only the developer knows this.

Do you think I should do `TS_ASSERT_THROWS_EQUALS`? This influences all my future testing ...

Framework/Algorithms/test/Stitch1DManyTest.h
```diff
@@ -1200,36 +1124,66 @@ class Stitch1DManyTest : public CxxTest::TestSuite {
...
+    Stitch1DMany alg;
+    alg.initialize();
+    alg.setProperty("InputWorkspaces", "ws1, ws2");
+    alg.setProperty("UseManualScaleFactors", "1");
+    alg.setProperty("ManualScaleFactors", "1");
+    alg.setProperty("OutputWorkspace", "ws");
+    alg.execute();
+    TS_ASSERT(alg.isExecuted());
+    Mantid::API::AnalysisDataService::Instance().clear();
+  }
```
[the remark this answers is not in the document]
reimundILL: Thanks, I forgot to test!

Framework/Algorithms/src/Stitch1D.cpp
```diff
@@ -356,39 +374,109 @@ MatrixWorkspace_sptr Stitch1D::integration(MatrixWorkspace_sptr &input,
...
+  Mantid::API::AnalysisDataService::Instance().remove(in2);
+  API::Workspace_sptr ws = conjoinX->getProperty("OutputWorkspace");
+  return boost::dynamic_pointer_cast<Mantid::API::MatrixWorkspace>(ws);
+}
+
+/** Sort x axis
+ @param ws :: Input workspace
+ @return A shared pointer to the resulting MatrixWorkspace
+ */
+void Stitch1D::sortXAxis(MatrixWorkspace_sptr &ws) {
```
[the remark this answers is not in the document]
reimundILL: Created an issue https://github.com/mantidproject/mantid/issues/22515 and I will move it.

Framework/Algorithms/test/Stitch1DManyTest.h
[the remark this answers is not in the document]
ianbush: If you just get `Some invalid properties found` I agree the extra test is less useful. It still tells you something - that the validation failed and not, say, something later in the execution.

I think it should be your decision if you think it is useful to add the extra asserts. Perhaps not here, but for the `Stitch1DTests` which use `RunCombinationHelpers`. If `TS_ASSERT_THROWS` is true then there should be no serious problems, but I have seen cases where an algorithm throws for an unexpected reason - which means unexpected behaviour in the code somewhere and a less helpful message to the user.

soininen:
I'll try this out with ILL reflectometry reduction.

soininen:
When the input workspaces are histograms but have DX, I get an error message:
```
Invalid value for RHSWorkspace: Workspace Rup-histo-4.5 is not compatible: spectra must have either Dx values or not; 
```

This is a bit confusing. Maybe it should say:
```
Invalid value for RHSWorkspace: Workspace Rup-histo-4.5 is not compatible: histogram data with DX values is not supported.
```

soininen:
Tested `Stitch1DMany` with ILL's point data with DX and the results look good. The figure below shows three stitched workspaces for two different polarizations. I checked some individual values by hand and at the overlapping regions the X and DX values match with the original workspaces. Great work!

stitched

ianbush (approved):
This looks good to go :+1: .

Framework/Algorithms/src/Stitch1D.cpp
```diff
@@ -1,45 +1,95 @@
...
+        vecdx[l] = (++it)->second;
+      ++l;
+      ++it;
+    }
+    auto x = make_cow<HistogramX>(std::move(vecx));
+    auto y = make_cow<HistogramY>(std::move(vecy));
+    auto e = make_cow<HistogramE>(std::move(vece));
+    ws->setSharedX(i, x);
+    ws->setSharedY(i, y);
+    ws->setSharedE(i, e);
```
SimonHeybrock: Using `make_cow` and `setSharedXYE` is not helping here since it is not reused for different workspace indices, i.e., there will not be any sharing. I would suggest to directly use the accessors like `mutableX`` etc., instead of using temporary vectors `vecx` etc., avoiding allocation overheads.

Framework/Algorithms/src/Stitch1D.cpp
```diff
@@ -1,45 +1,95 @@
...
+    Mantid::MantidVec vecy(ws_size);
+    Mantid::MantidVec vece(ws_size);
+    Mantid::MantidVec vecdx(ws_size);
+    int l = 0;
+    auto it = sorter.cbegin();
+    while (it != sorter.cend()) {
+      vecx[l] = it->first;
+      vecy[l] = it->second;
+      vece[l] = (++it)->second;
+      if (ws->hasDx(i) && !ws->isHistogramData())
```
SimonHeybrock: Why are you not handling `Dx` for histogram data?

reimundILL: For stitching histogram data, the weighted mean will be computed for the overlap region. This changes the number of bins and bin edges in this overlap region and then it is not easy to introduce new `Dx` values. My feeling here is that we need to think about this more deeply. At ILL, point data are used for reflectometry and thus, our scientists may not demand further `Dx` handling.
Histogram data having `Dx` can be converted to point data and then handled, if desired.

Framework/Algorithms/src/Stitch1D.cpp
reimundILL: Sounds reasonable.

Framework/Algorithms/src/Stitch1D.cpp
SimonHeybrock: Ok makes sense. Could you put a small comment explaining this in the code, for future reference?

reimundILL: Yes, no problem

PR #22362, opened 2018-05-01
 workspace documentation

### LINKED ISSUE #22318: Workspace Documentation Needs Updating
FRM2 Nicos team complained that they could not independently understand or navigate the information.

I suggest:

1. Separating out the c++ (even the implied c++) stuff from the documentation
1. Making it very clear what the concrete workspace types are and what they offer (how they are structured) etc. No need to mention base classes.
1. Avoid describing lots about the ADS (how data is stored), and instead focus on the structure of each data structure itself when it comes to python description.

OwenArnold:
Fix workspace documentation as per the external user request. I have focused effort on *Workspace*, *MatrixWorkspace* and *EventWorkspace* to hopefully make things clearer. My objectives are described in the original issue

Fixes #22318

**Tester**

The tester should review the documentation and provide feedback for changes suggested. Ideally the tester should be very familiar with all the concepts discussed.


dev-docs/source/EventWorkspaceDev.rst
```diff
@@ -0,0 +1,93 @@
...
+Event Workspace Development
+===========================
+
+.. contents::
+  :local:
+
+The following information will be useful to you if you want to write an
+:ref:`algorithm` that is :ref:`Event Workspace <EventWorkspace>` aware.
+
+Individual Neutron Event Data (TofEvent)
```
peterfpeterson: It may be useful to mention what header ([Framework/Types/inc/MantidTypes/Event/TofEvent.h](https://github.com/mantidproject/mantid/blob/master/Framework/Types/inc/MantidTypes/Event/TofEvent.h)) has these constructs so people can find more information. There are actually a couple of different types of "events" including `TofEvent`, `WeightedEvent`, and `WeightedEventNoTime` which all have different number of fields. They are all implemented as `struct` rather than classes.

dev-docs/source/EventWorkspaceDev.rst
```diff
@@ -0,0 +1,93 @@
...
+########################################
+
+The TofEvent class holds information for each neutron detection event
+data:
+
+-  PulseTime: An absolute time of the pulse that generated this neutron.
```
peterfpeterson: Actually stored as a `Kernel::DateAndTime` object which internally stores it as you say.

dev-docs/source/EventWorkspaceDev.rst
```diff
@@ -0,0 +1,93 @@
...
+   Note that this field can be converted to other units, e.g. d-spacing.
+
+Lists of Events (EventList)
+###########################
+
+-  The EventList class consists of a list of TofEvent's. The order of
+   this list is not significant, since various algorithms will resort by
+   time of flight or pulse time, as needed.
+
+-  Also contained in the EventList is a std::set of detector ID's. This
```
peterfpeterson: I believe this is a requirement of `Spectrum` which `EventList` inherits from which may be worth mentioning.

dev-docs/source/EventWorkspaceDev.rst
```diff
@@ -0,0 +1,93 @@
...
+   tracks which detector(s) were hit by the events in
+   the list.
+
+-  The histogram bins (X axis) are also stored in EventList. The Y and E
```
peterfpeterson: I think the sentence should end "as they are calculated on demand in the MRU (below)."

dev-docs/source/EventWorkspaceDev.rst
```diff
@@ -0,0 +1,93 @@
...
+
+An Event Workspace contains a list of the 100 most-recently used
+histograms, a MRUList. This MRU caches the last histogram
+data generated for fastest display.
+
+A note about workspace index / spectrum number / detector ID
+############################################################
+
+For event workspaces there is no benefit, and only a drawback to grouping detectors in hardware, therefore most of the loading algorithms for event data **match** the workspace index and spectrum number
+in the Event Workspace. Therefore, in an Event Workspace, the two numbers
```
peterfpeterson: "... the two numbers will often be the same..."

dev-docs/source/EventWorkspaceDev.rst
```diff
@@ -0,0 +1,93 @@
...
+
+Workspace2D compatibility
+#########################
+
+Event Workspace is designed to be able to be read (but not written to)
+like a :ref:`MatrixWorkspace <MatrixWorkspace>`. By default, if an algorithm
+performs an operation and outputs a new workspace, the
+WorkspaceFactory will create a :ref:`Workspace2D` *copy*
+of your Event Workspace's histogram representation. If you attempt to
+change an Event Workspace's Y or E data in place, you will get an error
```
peterfpeterson: I recall this generating an exception that gets converted to an error message by the `IAlgorithm` framework.

docs/source/concepts/EventWorkspace.rst
```diff
@@ -210,91 +211,7 @@ Please note these should only be done as part of a Python Algorithm, otherwise t
...
-This function will return *false* (thereby disabling parallelization) if
-any of the event lists are unsorted.
-
-You can go around this by forcing the parallel loop with a plain
-PARALLEL\_FOR() macro. **Make sure you do not read from the same
-spectrum in parallel!**
-
-.. include:: WorkspaceNavigation.txt
-
+See the Event Workspace Development documentation
```
peterfpeterson: Make this a plain link to the developer docs. Since we aren't including the inter-sphinx there isn't another way.

docs/source/concepts/MatrixWorkspace.rst
```diff
@@ -8,57 +8,53 @@ Matrix Workspace
...
+-  Usable geometry may be present, for example to determine the position of each detector, see :ref:`here <InstrumentAccessLayers>` for more information
+-  Spectra - detector mapping, and other spectra level information see :ref:`here <InstrumentAccessLayers>`.
+-  :ref:`Histogram <HistogramData>` information
 -  A list of 'masked' bins
 
 Working with Matrix Workspaces in Python
 ----------------------------------------
 
-MatrixWorkspace is an abstract description of an specific workspace implementation. It provides access to a common way of accessing the data for a 2D workspace without needing to know the details of how that data is actually stored.
+As alluded to above, Matrix Workspace is an abstraction. Matrix Workspace provides access to the majority of the common information for all 2D workspaces in its family, without needing to know all the details of how that data is actually stored.
```
peterfpeterson: Should be "MatrixWorkspace" rather than "Matrix Workspace".

docs/source/concepts/Workspace.rst
```diff
@@ -10,131 +10,38 @@ Workspace
...
-    The variable myWS now points to the workspace called MyNewWorkspace
-    MyNewWorkspace has been created that also points to the workspace called MyNewWorkspace
-    myOtherWS now points to the workspace called myOtherWS
+- The type identifier (id) of the workspace (see below), unique to every type of workspace 
+- The name of the workspace, unique to every instance of a workspace even those of the same type
+- The history of :ref:`algorithms <algorithm>` that were run to form and manipulate the workspace to its current state
 
-Workspace Properties
-####################
+.. tip:: In `MantidPlot <http://www.mantidproject.org/MantidPlot:_Help>`__ the data from the workspaces can be graphically viewed, inspected, and plotted in many ways.
```
peterfpeterson: Isn't there a place in the user docs to point people at instead? I'm surprised that it wasn't moved over, but cannot find it.

peterfpeterson (changes requested):
Overall a nice re-organization with good additions. Some minor points on particular sections and two general comments:
1. You should do is markup parameters/functions/etc with double back-ticks. This will aid in readability and make it that you shouldn't need to escape `_`. It is an issue in multiple parts of the documentation that should be fixed as they are discovered.
2. Various places have "Matrix Workspace" rather than "MatrixWorkspace". It should not have a space since it is an actual classname and should be escaped in back-ticks.

docs/source/concepts/MatrixWorkspace.rst
OwenArnold: @peterfpeterson I deliberately changed from`MatrixWorkspace`, to `Matrix Workspace` but I think we should agree a common way that we present "concepts" in plain english to users who are not necessarily going to be interested in our `c++` or `python` conventions. `MatrixWorkspace` feels less natural

peterfpeterson (approved):
Works for me.

docs/source/concepts/MatrixWorkspace.rst
SimonHeybrock: I am with Pete here -- `Matrix Workspace` does not make any sense in English (or at least no one will see the connection to what our `MatrixWorkspace` is). If we add spaces to many of our class names it will be a big mess, no one will know what to search for, and it is not clear to anyone which concept is a class and which is not.

PR #22367, opened 2018-05-02
Engineering GUI implement new model for EnggVanadiumCorrections

### LINKED ISSUE #17424: Engg:: Vanadium corrections are always performed
In `loadOrCalcVanadiumWorkspaces` - `findPrecalcVanadiumCorrFilenames` appears to fail to find pre calculated vanadium workspaces. This means we are always forced to rerun `vanadiumCorrections` which significantly slows down calibration.

This affects debug more than release as it only adds 10-20 seconds vs 10 minutes in debug.

josephframsay:
A model for generating and caching the vanadium corrections workspaces (calculated by EnggVanadiumCorrections and used for calibration and focusing) was implemented and integrated into the main presenter of the ED GUI. The intended behaviour was always to cache the workspaces so that we don't need to run EnggVanadiumCorrections (which is slow) for every calibration or focus, but the caching didn't actually appear to happen. The new model takes care of the algorithm and the cache, and also gets unit tests on the behaviour.

**To test:**

There's some sample data [here](https://github.com/mantidproject/mantid/files/1967047/EnggCalibrateAndFocus.zip)

Open up the GUI and go to the Calibration tab. Under **Make new calibration**, select the 280329 run for Vanadium and 280336 for Calibration sample. Click **Calibrate**, and go and make a coffee for a few minutes. Once this is complete, make sure that the two workspaces were saved out (you can see where they were saved in the Calibration field in the Settings tab).

Now go to the Focus tab and select the 281188 run and click **Focus**. Make sure that EnggVanadiumCorrections isn't run by looking in the Algorithm Progress window.

Repeat any of the above steps, and EnggVanadiumCorrections still shouldn't be run

Back in the Settings tab, tick **Force recalculate all existing Vanadium files**. Repeat one of the above steps and make sure EnggVanadiumCorrections is run again, and the files are overwritten

Fixes #17424

**Release Notes** 

Doesn't need to go in the release notes, as it's mostly a refactoring change, and I don't think the scientists were aware that the cache didn't work in the first place

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/qt/scientific_interfaces/EnggDiffraction/EnggDiffractionPresenter.cpp
+++ b/qt/scientific_interfaces/EnggDiffraction/EnggDiffractionPresenter.cpp
@@ -1076,8 +1076,8 @@ void EnggDiffractionPresenter::doCalib(const EnggDiffCalibSettings &cs,
   m_vanadiumCorrectionsModel->setCurrentInstrument(m_view->currentInstrument());
   const auto vanadiumCorrectionWorkspaces =
       m_vanadiumCorrectionsModel->fetchCorrectionWorkspaces(vanNo);
-  const auto vanIntegWS = vanadiumCorrectionWorkspaces.first;
-  const auto vanCurvesWS = vanadiumCorrectionWorkspaces.second;
+  const auto &vanIntegWS = vanadiumCorrectionWorkspaces.first;
+  const auto &vanCurvesWS = vanadiumCorrectionWorkspaces.second;
 
   try {
     auto load = Mantid::API::AlgorithmManager::Instance().create("Load");
@@ -1731,8 +1731,8 @@ void EnggDiffractionPresenter::doFocusing(const EnggDiffCalibSettings &cs,
   const auto vanadiumCorrectionWorkspaces =
       m_vanadiumCorrectionsModel->fetchCorrectionWorkspaces(
           m_view->currentVanadiumNo());
-  const auto vanIntegWS = vanadiumCorrectionWorkspaces.first;
-  const auto vanCurvesWS = vanadiumCorrectionWorkspaces.second;
+  const auto &vanIntegWS = vanadiumCorrectionWorkspaces.first;
+  const auto &vanCurvesWS = vanadiumCorrectionWorkspaces.second;
 
   const std::string inWSName = "engggui_focusing_input_ws";
   const std::string instStr = m_view->currentInstrument();
--- a/qt/scientific_interfaces/EnggDiffraction/EnggDiffractionPresenter.cpp
+++ b/qt/scientific_interfaces/EnggDiffraction/EnggDiffractionPresenter.cpp
@@ -8,6 +8,7 @@
 #include "MantidKernel/StringTokenizer.h"
 #include "MantidQtWidgets/Common/PythonRunner.h"
 #include "EnggDiffractionPresWorker.h"
+#include "EnggVanadiumCorrectionsModel.h"
 #include "IEnggDiffractionView.h"
 
 #include <algorithm>
@@ -57,7 +58,10 @@ std::string EnggDiffractionPresenter::g_sumOfFilesFocus = "";
 EnggDiffractionPresenter::EnggDiffractionPresenter(IEnggDiffractionView *view)
     : m_workerThread(nullptr), m_calibFinishedOK(false),
       m_focusFinishedOK(false), m_rebinningFinishedOK(false), m_view(view),
-      m_viewHasClosed(false) {
+      m_viewHasClosed(false),
+      m_vanadiumCorrectionsModel(
+          boost::make_shared<EnggVanadiumCorrectionsModel>(
+              m_view->currentCalibSettings(), m_view->currentInstrument())) {
   if (!m_view) {
     throw std::runtime_error(
         "Severe inconsistency found. Presenter created "
@@ -1059,8 +1063,6 @@ void EnggDiffractionPresenter::doCalib(const EnggDiffCalibSettings &cs,
                                        const std::string &ceriaNo,
                                        const std::string &outFilename,
                                        const std::string &specNos) {
-  ITableWorkspace_sptr vanIntegWS;
-  MatrixWorkspace_sptr vanCurvesWS;
   MatrixWorkspace_sptr ceriaWS;
 
   // Append current instrument name if numerical only entry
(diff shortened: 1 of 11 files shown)


soininen:
Testing this with the provided test data. `EnggVanadiumCorrections` indeed gets omitted when running the steps the second time. However, the forced recalculation doesn't seems to work: after checking the box in the Settings tab, `EnggVanadiumCorrections` still won't run and the two files `ENGINX00280329_precalculated_vanadium_run_bank_curves.nxs` and `ENGINX00280329_precalculated_vanadium_run_integration.nxs` don't get updated. In fact, if I delete these files I get
```
EnggDiffraction GUI: starting new focusing. This may take some seconds... 
Generating new focusing workspace(s) and file(s) into this directory: 
Generating new focused file (bank 1) for run 281188 into: ENGINX_281188_focused_bank_1.nxs
The focusing failed. Some input properties were not valid. See log messages for details. Error: Invalid value for property Filename (list of str lists) "/users/soininen/ENGINX00280329_precalculated_vanadium_run_bank_curves.nxs": File "/users/soininen/ENGINX00280329_precalculated_vanadium_run_bank_curves.nxs" not found
Generating new focused file (bank 2) for run 281188 into: ENGINX_281188_focused_bank_2.nxs
The focusing failed. Some input properties were not valid. See log messages for details. Error: Invalid value for property Filename (list of str lists) "/users/soininen/ENGINX00280329_precalculated_vanadium_run_bank_curves.nxs": File "/users/soininen/ENGINX00280329_precalculated_vanadium_run_bank_curves.nxs" not found
The focusing did not finish correctly. Check previous log messages for details
```

when running Focus and
```
EnggDiffraction GUI: starting new calibration. This may take a few seconds... 
Generating new calibration file: ENGINX_280329_280336_all_banks.prm
The calibration calculations failed. Some input properties were not valid. See log messages for details. 
The calibration did not finish correctly. Please check previous log messages for details.
```

when calibrating.

Here is how the interface looks for me:
engg-calibration

engg-focus

engg-settings

Am I doing something wrong here?

josephframsay:
@soininen nope, I missed both of those - nice one! This should be ready for test again

qt/scientific_interfaces/EnggDiffraction/EnggDiffractionPresenter.cpp
```diff
@@ -1059,19 +1063,21 @@ void EnggDiffractionPresenter::doCalib(const EnggDiffCalibSettings &cs,
...
-  loadOrCalcVanadiumWorkspaces(vanFileHint, cs.m_inputDirCalib, vanIntegWS,
-                               vanCurvesWS, cs.m_forceRecalcOverwrite, specNos);
+  // TODO: the settings tab should emit a signal when these are changed, on
+  // which the vanadium corrections model should be updated automatically
+  m_vanadiumCorrectionsModel->setCalibSettings(cs);
+  m_vanadiumCorrectionsModel->setCurrentInstrument(m_view->currentInstrument());
+  const auto vanadiumCorrectionWorkspaces =
+      m_vanadiumCorrectionsModel->fetchCorrectionWorkspaces(vanNo);
+  const auto vanIntegWS = vanadiumCorrectionWorkspaces.first;
+  const auto vanCurvesWS = vanadiumCorrectionWorkspaces.second;
```
soininen: Not that it matters much performance-wise, but as far as I can see, `vanIntegWS` and `vanCurvesWS` could be references avoiding copying the shared pointers.

qt/scientific_interfaces/EnggDiffraction/EnggDiffractionPresenter.cpp
```diff
@@ -1718,20 +1724,15 @@ void EnggDiffractionPresenter::doFocusing(const EnggDiffCalibSettings &cs,
...
-
-  loadOrCalcVanadiumWorkspaces(vanFileHint, cs.m_inputDirCalib, vanIntegWS,
-                               vanCurvesWS, cs.m_forceRecalcOverwrite, "");
+  m_vanadiumCorrectionsModel->setCalibSettings(cs);
+  m_vanadiumCorrectionsModel->setCurrentInstrument(m_view->currentInstrument());
+  const auto vanadiumCorrectionWorkspaces =
+      m_vanadiumCorrectionsModel->fetchCorrectionWorkspaces(
+          m_view->currentVanadiumNo());
+  const auto vanIntegWS = vanadiumCorrectionWorkspaces.first;
+  const auto vanCurvesWS = vanadiumCorrectionWorkspaces.second;
```
soininen: See the above comment about references, maybe it applies here as well?

qt/scientific_interfaces/EnggDiffraction/EnggVanadiumCorrectionsModel.cpp
```diff
@@ -0,0 +1,190 @@
...
+boost::optional<Mantid::API::ITableWorkspace_sptr>
+EnggVanadiumCorrectionsModel::fetchCachedIntegratedWorkspace(
+    const std::string &vanadiumRunNumber) const {
+  const auto filename = generateIntegratedFilename(vanadiumRunNumber);
+  const Poco::File inputFile(filename);
+  if (inputFile.exists()) {
+    return loadTableWorkspace(filename, INTEGRATED_WORKSPACE_NAME);
+  }
+
+  return boost::none;
```
soininen: Why not just return a `nullptr`?

qt/scientific_interfaces/EnggDiffraction/EnggVanadiumCorrectionsModel.cpp
```diff
@@ -0,0 +1,190 @@
...
+  const auto integratedFilename = normalisedRunName + filenameSuffix;
+
+  Poco::Path inputDir(m_calibSettings.m_inputDirCalib);
+  inputDir.append(integratedFilename);
+  return inputDir.toString();
+}
+
+std::string EnggVanadiumCorrectionsModel::generateVanadiumRunName(
+    const std::string &vanadiumRunNumber) const {
+  const static size_t normalisedRunNumberLength = 8;
```
soininen: How about `constexpr size_t`? Or is this something that could be read from the facility info?

soininen (changes requested):
The interface works now as promised: `EnggVanadiumCorrections` is only run when needed or specifically requested. The code changes look good and are covered by some unit tests. A few minor comments below.

I did notice, though, that MantidPlot crashes due to unhandled expection if the processing of the interface's algorithms is canceled from the Algorithm process dialog. This should perhaps be investigated and fixed under another issue.

josephframsay:
Ah, I always forget my `constexpr`s. The problem with MantidPlot crashing is present across the whole GUI except for the GSAS tab - fixing it is way way way out of scope of this PR. I've put in an issue at #22492

soininen (approved):

PR #22386, opened 2018-05-04
Prevent crash - missing error handling in `RunCombinationHelper`

### LINKED ISSUE #22383: `ConjoinXRuns` segfaults when giving a `TableWorkspace` input etc (input validation fails)
`ConjoinXRuns` should only receive MatrixWorkspaces or groups of MatrixWorkspaces.

I investigated the input validation for Stitch1DMany that has similar input requirements. And found some differences, then tested the input validation of ConjoinXRuns with a TableWorkspace  and it crashes MantidPlot. The auto generated GUI puts all workspaces in the ADS to the selection list. Then, it is easy to select a wrong workspace type.

Try something like:
ws1 = CreateSampleWorkspace()
ws2 =CreateEmptyTableWorkspace()
ConjoinXRuns(ws1, ws2)

vreimund:
`RunCombinationHelper` did not check, if the input workspaces represent either a GroupWorkspace or alternatively a MatrixWorkspace. This affects today the algorithms `ConjoinXRuns`, `MergeRuns`, and `SumOverlappingTubes`.

Now, it throws a std::runtime_error that need to be catched during input validation in `ConjoinXRuns` (otherwise the GUI freezes). It may be ok for `MergeRuns` and `SumOverlappingTubes` to just throw an error.

If the input workspace is not a group, I added the check if it is a MatrixWorkspace instead. It may be enough to add the workspace name to the std::vector as implemented before.
In `ConjoinXRuns`, I think it is not good to get the name of a workspace if the corresponding pointer does not exist.

**To test:**

```Python
CreateSampleWorkspace(OutputWorkspace='ws')
CreateEmptyTableWorkspace(OutputWorkspace='table')
out = ConjoinXRuns(mtd["table"], mtd["ws"])
```

Fixes #22383.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/test/RunCombinationHelperTest.h
+++ b/Framework/Algorithms/test/RunCombinationHelperTest.h
@@ -40,6 +40,20 @@ public:
     m_testee.setReferenceProperties(m_reference);
   }
 
+  void testUnwraping_throws() {
+
+    MatrixWorkspace_sptr ws1 = create2DWorkspace(2, 3);
+    storeWS("ws1", ws1);
+
+    TS_ASSERT_THROWS_EQUALS(
+        m_testee.unWrapGroups(std::vector<std::string>{"ws1", "ws?"}),
+        const std::exception &e, std::string(e.what()),
+        "Unable to find workspace type with name 'ws?': data service  search "
+        "object ws?");
+
+    removeWS("ws1");
+  }
+
   void testUnwraping() {
 
     MatrixWorkspace_sptr ws1 = create2DWorkspace(2, 3);
--- a/Framework/Algorithms/test/RunCombinationHelperTest.h
+++ b/Framework/Algorithms/test/RunCombinationHelperTest.h
@@ -27,7 +27,6 @@ using namespace Mantid::HistogramData;
 using namespace Mantid::Kernel;
 using namespace WorkspaceCreationHelper;
 
-
 class RunCombinationHelperTest : public CxxTest::TestSuite {
 public:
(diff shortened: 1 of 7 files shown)


Framework/Algorithms/test/RunCombinationHelperTest.h
```diff
@@ -40,6 +40,20 @@ class RunCombinationHelperTest : public CxxTest::TestSuite {
...
+
+    TS_ASSERT_THROWS_EQUALS(
+        m_testee.unWrapGroups(std::vector<std::string>{"ws1", "ws?"}),
+        const std::runtime_error &e, std::string(e.what()),
+        "Unable to find workspace type with name 'ws?': data service  search "
+        "object ws?");
+
+    removeWS("ws1");
+  }
+
```
josephframsay: Could we also test that trying to conjoin invalid workspaces such as tables throws?

josephframsay (changes requested):
This is a nice fix - the changes look good, and the crash has disappeared in all 3 algorithms. Just one comment about units tests

Framework/Algorithms/test/RunCombinationHelperTest.h
reimundILL: That's true, the std::runtime error with message "The input ... is neither a WorkspaceGroup nor a MatrixWorkspace" should be tested instead.

josephframsay (approved):

josephframsay:
All looks good to go now!

PR #22396, opened 2018-05-08
Exclude and IgnoreInvalidData Properties for PlotPeakByLogValue and QENSFitSequential

### LINKED ISSUE #22387: Indirect/Fitting - 'Exclude' not available in PlotPeakByLogValue and QENSFitSequential
PlotPeakByLogValue and as a result, QENSFitSequential, do not provide an option 'Exclude' for excluding a region from the fit.

This value of a new 'Exclude' property should be passed to the Exclude property of the 'Fit' algorithm.

brandonhewer:
The PlotPeakByLogValue (and as a result the QENSFitSequential algorithm) does not have an 'Exclude' or 'IgnoreInvalidData' property for excluding a region from the fit and ignoring NaNs, 0s and Infinite values.

This PR adds the 'Exclude' and 'IgnoreInvalidData' properties to the PlotPeakByLogValue algorithm.

**To test:**
1. Run the PlotPeakByLogValue unit tests and ensure they pass.

Fixes #22387. 

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


josephframsay (changes requested):
Nice changes, and tests pass fine. Do we also need to include the change in the test for QENSFitSequential?

Also for forwarding `Exclude` and `IgnoreInvalidData`, could you use `Property.setValueFromProperty`? I don't really know if this would be advantageous or not - I suppose it would maybe make it make it a bit clearer that the property is just being forwarded, and possibly save the parent algorithm having to extract the value from the property?

Might it also be worth declaring these with `copyProperty` a la [this](https://github.com/mantidproject/mantid/blob/master/Framework/WorkflowAlgorithms/src/LoadEventAndCompress.cpp#L55), so that if they ever change in `Fit`, the change applies to your algorithm also?

brandonhewer:
@josephframsay 
I'm not sure setValueFromProperty makes sense from the context of a PropertyManager (since Property objects are inaccessible -- encapsulated).

The copyProperty may be a reasonable option to add to QENSFitSequential, however will not be possible to do within PlotPeakByLogValue for the Exclude property. This is because this property is actually created dynamically after the InputWorkspace property is set and not at the point the init() method is called.

josephframsay:
@brandonhewer ah yes, right you are. in that case, this looks good to go

josephframsay (approved):

josephframsay:
@brandonhewer ah yes, right you are. in that case, this looks good to go

PR #22407, opened 2018-05-09
Propagate input workspace properties properly in LineProfile

### LINKED ISSUE #21964: LineProfile: sample logs from input workspace
The output workspace of the `LineProfile` algorithm should include the sample logs of the input workspace.

### LINKED ISSUE #22103: LineProfile doesn't propagate the distribution flag
The distributionness of `LineProfile` output workspaces is always False independently of the input. The output should rather have the same data type as the input.

### LINKED ISSUE #22104: LineProfile should conserve the instrument, sample logs and history
The output of `LineProfile` is too clean: it should rather include the instrument, sample logs and history of the input workspace.

soininen:
This PR improves the `LineProfile` algorithm which now uses the InputWorkspace as the parent workspace for the OutputWorkspace preserving the instrument, sample logs and history.

**To test:**

Grab a Workspace2D from somewhere and test the algorithm with it. Check that the output workspace has the sample logs and instrument of the input and that the history is correct.

Fixes #22104. Fixes #22103. Fixes #21964.

**Release Notes** 

Framework release notes were updated accordingly.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Algorithms/test/LineProfileTest.h
```diff
@@ -390,6 +391,153 @@ class LineProfileTest : public CxxTest::TestSuite {
...
+    const BinEdges edges{{0., 0.1, 1.1, 11.1}};
+    const Frequencies frequencies{3., 2., 1.};
+    const Histogram histogram(edges, frequencies);
+    MatrixWorkspace_sptr inputWS = create<Workspace2D>(nHist, histogram);
+    LineProfile alg;
+    // Don't put output in ADS by default
+    alg.setChild(true);
+    alg.setRethrows(true);
+    TS_ASSERT_THROWS_NOTHING(alg.initialize())
+    TS_ASSERT(alg.isInitialized())
```
reimundILL: I like your testing, one question: test_Init() already checks generally if the algorithm is initialized after initialization. You decided to (always?) check this again. Is this needed?
Maybe test_Init() can check initialization for child algorithms as well, if needed?

vreimund (approved):
Very good documentation & testing. I run all documentation tests and verified SampleLogs, history, instrument. The instrument view shows the detector now in dark gray color since line profile. I tested distributions and non-distributions and results are as expected. Some figures for the first documentation example:

lineprofile1

lineprofile2

Framework/Algorithms/src/LineProfile.cpp
```diff
@@ -87,6 +89,39 @@ struct IndexLimits {
...
+ * @param ws The parent workspace.
+ * @param Xs Profile's X values.
+ * @param Ys Profile's Y values.
+ * @param Es Profile's E values.
+ * @return A single histogram profile workspace.
+ */
+Workspace2D_sptr makeOutput(const MatrixWorkspace &ws,
+                            const LineDirection direction,
+                            std::vector<double> &&Xs, std::vector<double> &&Ys,
+                            std::vector<double> &&Es) {
```
SimonHeybrock: Why use an rvalue reference if you are not moving from it below?

Framework/Algorithms/src/LineProfile.cpp
```diff
@@ -87,6 +89,39 @@ struct IndexLimits {
...
+  Workspace2D_sptr outWS;
+  const bool histogramOutput = direction == LineDirection::horizontal
```
SimonHeybrock: Have a look at `HistogramBuilder`, I think you might be able to use it to avoid distinguishing all those cases here and below. Just build the histogram and use it with a single `create` call.

Framework/Algorithms/src/LineProfile.cpp
```diff
@@ -87,6 +89,39 @@ struct IndexLimits {
...
+                                   ? ws.isHistogramData()
+                                   : Xs.size() > Ys.size();
```
SimonHeybrock: I am not sure I understand these cases. If `Xs.size() == Ys.size()` but `direction == LineDirection::horizontal` you will be attempting to create a histogram output from it, but the `Histogram` constructor will throw, I think?

Framework/Algorithms/src/LineProfile.cpp
```diff
@@ -206,12 +243,18 @@ void profile(std::vector<double> &Xs, std::vector<double> &Ys,
...
+      double y = ws.y(iVert)[iHor];
       if ((ignoreNans && std::isnan(y)) || (ignoreInfs && std::isinf(y))) {
         continue;
       }
+      double e = ws.e(iVert)[iHor];
+      if (convertFromDistribution) {
+        const auto binWidth =
+            std::abs(ws.x(iVert)[iHor + 1] - ws.x(iVert)[iHor]);
+        y *= binWidth;
+        e *= binWidth;
```
SimonHeybrock: Have a look at `HistogramIterator` (or rather `HistogramItem` for the accessors). It provides thinks like bin width, might be simpler to use here.

docs/source/algorithms/LineProfile-v1.rst
```diff
@@ -24,6 +24,13 @@ By default, the profile is calculated as an average over the line width. This be
 'Sum'
     Sum the values, weighting them by :math:`n / n_{tot}` where :math:`n` is the number of summed data points (excluding special values if *IgnoreNans* or *IgnoreInfs* is set) and :math:`n_{tot}` is the total number of data points (including special values).
 
+Profiles over distributions
+---------------------------
+
+Horizontal profiles over distribution data (Y divided by bin width) will be distributions as well.
+
+However, vertical profiles over distribution data will be non-distributions, that is, the Y values will be multiplied by the X bin width before summation or averaging.
```
SimonHeybrock: I am not entirely convinced that this makes sense. Can you explain the reasoning behind this? I agree that the output would not be a distribution, but does that really imply that we need to multiply by the X bin width?

SimonHeybrock (changes requested):
Some comments/questions below. :+1: :+1: for adding a good number of new unit tests!

Framework/Algorithms/src/LineProfile.cpp
soininen: `BinEdges` and `VectorOf` have move constructors. Do I still need to `move` manually? I'd thought `move` only casts to rvalue if it is possible.

SimonHeybrock: I think so, have a look: https://stackoverflow.com/questions/31213539/c-stdmove-with-rvalue-reference-is-not-moving-contents.

Framework/Algorithms/src/LineProfile.cpp
soininen: In the case `Xs.size() == Ys.size()` and `LineDirection::horizontal`, `ws.isHistogramData()` should return `false`, right? So the output would be point data. Anyway, I think I can rewrite this using `HistogramBuilder` as you suggested (thanks for the hint!), so this doesn't matter.

docs/source/algorithms/LineProfile-v1.rst
soininen: The line profile is a single histogram, and in the case of vertical profiles, the cut is transposed. The problem with vertical line profiles from distributions is that if the bin width normalization is kept, there is no flag or similar in Mantid that could tell that the bins in the output workspace are normalized by the bin 'height'. That is why I opted to unnormalize the bins. I am also unsure if this is the right thing to do, so let me know if you better ideas.

SimonHeybrock: I am not sure either.... but the bin 'height' is not available in the output workspace, i.e., in the best case (constant input bin width) it contains an unknown normalization factor, or in the worst case (variable input bin width) each output value will contain a different unknown factor that is impossible to deduce.

If we convert to input to a distribution, at least the value make sense, just the unit is unknown?

@martyngigg Any thoughts on this?

soininen: The 'height' is actually available, as the vertical axis is a `BinEdgeAxis`. The algorithm doesn't accept variable bin widths as input.

SimonHeybrock: Ok, with `CommonBinsValidator` this seems to make sense.

martyngigg: Agreed that with common bins this makes sense as the best we can achieve with the current strategy for tracking normalization.

docs/source/algorithms/LineProfile-v1.rst
```diff
@@ -29,7 +29,17 @@ Profiles over distributions
 
 Horizontal profiles over distribution data (Y divided by bin width) will be distributions as well.
 
-However, vertical profiles over distribution data will be non-distributions, that is, the Y values will be multiplied by the X bin width before summation or averaging.
+However, vertical profiles over distribution data will be non-distributions, albeit the data is normalized by the profile width. To remove the normalization, it is possible to multiply the data by the total width of included bins. This information is contained in the vertical axis:
```
SimonHeybrock: Thanks for explaining/documenting this!

SimonHeybrock (approved):

Issue #22406, opened 2018-05-09
Event Listeners do not update X values to reflect range of TOF in events


MikeHart85:
Issue encountered while working on Kafka Event Listener, but appears to apply to all Event Listeners.

### Expected behavior

When using a LiveListener to record event data in Add or Replace mode, the resulting single-bin workspace should have sensible minimum and maximum bin boundaries as its X values for every spectrum, such that no events in the workspace have a TOF outside of these boundaries.

### Actual behavior

The bin boundaries in the X values for every spectrum are always 0 and a tiny floating point number (probably DBL_MIN). They are not updated as new events are added or the workspace is replaced.

### Steps to reproduce the behavior

Run `FakeISISEventDAE` with default arguments from sidebar.

Then execute the following in the console:
```python
>>> ConfigService.setFacility('TEST_LIVE')
>>> StartLiveData(Instrument="ISIS_Event", OutputWorkspace="test", UpdateEvery=1, PreserveEvents=True, AccumulationMethod="Replace")
```
Open the Instrument View and observe the Time-of-flight range on the bottom, or view the workspace data and observe the X values.

PR #22413, opened 2018-05-09
Re #22406 Hack to avoid performance issues with bulk updating X values

MikeHart85:
Workaround for #22406.

**Description of work**

When in PreserveEvents mode, EventListeners were creating workspaces that had incorrect bin boundaries in their X values. This resulted in some algorithms not being able to work with the workspaces as expected, inconsistent behaviour in the Instrument View, etc.

While attempting to address this issue, it became apparent that there was no good way to bulk update X values for many spectra without incurring unreasonable performance costs. This will need to be handled in a separate issue.

As a temporary solution for this problem, I've implemented a quick and dirty workaround in this pull request so that work can progress on another issue:

Since all spectra share the same HistogramX instance via the copy-on-write mechanism, and this is guaranteed to be the case during the live listening process, we can update them all if we by-pass the copy and force a write to the single shared instance.

This should be replaced with something more robust once a better solution is found.

**To test:**

Run `FakeISISEventDAE` with default arguments from sidebar.

Then execute the following in the console:
```python
>>> ConfigService.setFacility('TEST_LIVE')
>>> StartLiveData(Instrument="ISIS_Event", OutputWorkspace="test", UpdateEvery=1, PreserveEvents=True, AccumulationMethod="Replace")
```
Open the Instrument View and observe the Time-of-flight range on the bottom, or view the workspace data and observe the X values.

To check performance, run with Kafka event streaming docker stack.

**Release Notes** 

Release notes updated.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/LiveData/src/LoadLiveData.cpp
+++ b/Framework/LiveData/src/LoadLiveData.cpp
@@ -500,6 +500,23 @@ void LoadLiveData::exec() {
     // Default to Add.
     this->addChunk(processed);
 
+  // If m_accumWS is an EventWorkspace, we have to update the bin boundaries
+  if (EventWorkspace_sptr accumEvent =
+          boost::dynamic_pointer_cast<EventWorkspace>(m_accumWS)) {
+    // Calculate fresh min and max based on all events in WS
+    double tofmin, tofmax;
+    accumEvent->getEventXMinMax(tofmin, tofmax);
+
+    // Ugly hack to update all X values in a performant way.
+    // There is currently no way to set X values for all spectra without
+    // breaking sharing of bin edges and paying an excessive performance cost.
+    // This hack depends on the premise that all spectra have shared bin edges
+    // (which should hold true at this point in the code).
+    auto &cowX =
+        *const_cast<HistogramData::HistogramX *>(accumEvent->sharedX(0).get());
+    cowX = std::vector<double>{tofmin, tofmax};
+  }
+
   // At this point, m_accumWS is set.
 
   if (this->hasPostProcessing()) {
--- a/docs/source/release/v3.13.0/framework.rst
+++ b/docs/source/release/v3.13.0/framework.rst
@@ -39,6 +39,7 @@ Improved
 
 - :ref:`ConjoinXRuns <algm-ConjoinXRuns>` joins Dx errors if present
 - The algorithm :ref:`SortXAxis <algm-SortXAxis>` has a new input option that allows ascending (default) and descending sorting. Furthermore, Dx values will be considered if present. The documentation needed to be corrected.
+- :ref:`Live Data <algm-StartLiveData>` for events in PreserveEvents mode now produces workspaces that have reasonable bin boundaries.
 
 New
 ###


docs/source/release/v3.13.0/framework.rst
```diff
@@ -39,6 +39,7 @@ Improved
 
 - :ref:`ConjoinXRuns <algm-ConjoinXRuns>` joins Dx errors if present
 - The algorithm :ref:`SortXAxis <algm-SortXAxis>` has a new input option that allows ascending (default) and descending sorting. Furthermore, Dx values will be considered if present. The documentation needed to be corrected.
+- :ref:`Live Data <algm-StartLiveData>` for events in PreserveEvents mode now produces workspaces that have reasonable bin boundaries.
```
OwenArnold: could you swap out the word "reasonable" for "that encompass the total x-range (TOF) for all events across all spectra" or something like that?

Framework/LiveData/src/LoadLiveData.cpp
```diff
@@ -500,6 +500,23 @@ void LoadLiveData::exec() {
...
+  if (EventWorkspace_sptr accumEvent =
+          boost::dynamic_pointer_cast<EventWorkspace>(m_accumWS)) {
+    // Calculate fresh min and max based on all events in WS
+    double tofmin, tofmax;
+    accumEvent->getEventXMinMax(tofmin, tofmax);
+
+    // Ugly hack to update all X values in a performant way.
+    // There is currently no way to set X values for all spectra without
+    // breaking sharing of bin edges and paying an excessive performance cost.
+    // This hack depends on the premise that all spectra have shared bin edges
```
OwenArnold: Reword: "The success of this hack is possible because all spectra have shared bin edges." Current wording kind of implies that it might not (or only sometimes work)

Framework/LiveData/src/LoadLiveData.cpp
```diff
@@ -500,6 +500,23 @@ void LoadLiveData::exec() {
...
     this->addChunk(processed);
 
+  // If m_accumWS is an EventWorkspace, we have to update the bin boundaries
+  if (EventWorkspace_sptr accumEvent =
+          boost::dynamic_pointer_cast<EventWorkspace>(m_accumWS)) {
+    // Calculate fresh min and max based on all events in WS
+    double tofmin, tofmax;
+    accumEvent->getEventXMinMax(tofmin, tofmax);
+
+    // Ugly hack to update all X values in a performant way.
```
OwenArnold: You should reference the issue we discussed that you should create for the bin edges problem in the comment too. You have profiling information that we don't want to loose, and want to connect up with this hack.

Framework/LiveData/src/LoadLiveData.cpp
```diff
@@ -500,6 +500,23 @@ void LoadLiveData::exec() {
...
+    // There is currently no way to set X values for all spectra without
+    // breaking sharing of bin edges and paying an excessive performance cost.
+    // This hack depends on the premise that all spectra have shared bin edges
+    // (which should hold true at this point in the code).
+    auto &cowX =
+        *const_cast<HistogramData::HistogramX *>(accumEvent->sharedX(0).get());
```
OwenArnold: do you think the zeroth index needs any explanation? zeroth is as good as any. Also is it possible for the workspace to have zero spectra? in which case the line above may cause a segfault.

OwenArnold (changes requested):
The fix is the best option we have, I have a few minor points that need to be cleared up.

MikeHart85:
Closing this PR as it will be replaced by a better fix the problem.

PR #19972, opened 2017-07-06
Reference frame: scattering angle sign defining axis

gvardany:
This PR adds a possibility in the IDF to set the axis defining the 2theta sign.
Until now, 2theta sign was the sign of the coordinate on the axis pointing up.
I. e. normally the pointing up axis is y, so negative y would mean negative 2theta.
For D20 at ILL, negative x should mean negative 2theta.
This adds the possibility to customise the desired behaviour, without altering the previous default one.

Merge the corresponding amend in geometry [PR #96](https://github.com/mantidproject/mantidgeometry/pull/96)

**To test:**

0. Code review & tests pass.
1. Load `ILL/D20/967100.nxs` from unit tests.
2. ConvertSpectrumAxis to signed_theta (in Direct mode, leave EFixed blank).
3. Resulting spectrum axis should span from about -6 to 147 degrees.

Fixes #19963 . 

**Release Notes** 

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [x] Is the code of an acceptable quality?
- [x] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)?
- [x] Are the unit tests small and test the class in isolation?
- [x] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [x] Do changes function as described? Add comments below that describe the tests performed?
- [ ] Do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant (user and developer) documentation been added/updated?

Does everything look good? Mark the review as **Approve**. A member of `@mantidproject/gatekeepers` will take care of it.

PR #22419, opened 2018-05-11
Fix signed two theta not respecting the IDF

### LINKED ISSUE #22415: MatrixWorkspace.detectorSignedTwoTheta does not respect the theta-sign setting of IDF
The theta-sign axis is the axis pointing up irrespectively of what the IDF says.

soininen:
This PR fixes a bug in `MatrixWorkspace::detectorSignedTwoTheta()` where the axis pointing up was accidentally used as the reference axis when determining the sign of a detector's 2theta.

**To test:**

Choose some instrument where the signed 2theta makes sense. One example is D17 where the theta-sign axis should be x. Make sure the `theta-sing` axis is defined in the IDF. The file should contain something like this:
```xml
  <defaults>
    <length unit="metre"/>
    <angle unit="degree"/>
    <reference-frame>
      <along-beam axis="z"/>
      <pointing-up axis="y"/>
      <handedness val="right"/>
      <theta-sign axis="x"/>
    </reference-frame>
  </defaults>
```

In MantidPlot, load the instrument using `LoadEmptyInstrument`. Check the signed two theta in Python:
```python
ws = mtd['ws']
ws.detectorSignedTwoTheta(ws.getDetector(33))
```

Make sure the signs are correct, for example for D17, detector 0 should have a positive signed two theta and detector 250 negative when the theta-sing axis is x.

Fixes #22415 . 

**Release Notes** 

The fix is mentioned in the Framework release notes.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/src/MatrixWorkspace.cpp
+++ b/Framework/API/src/MatrixWorkspace.cpp
@@ -803,9 +803,8 @@ MatrixWorkspace::detectorSignedTwoTheta(const Geometry::IDetector &det) const {
         "Source and sample are at same position!");
   }
   // Get the instrument up axis.
-  const V3D &instrumentUpAxis =
-      instrument->getReferenceFrame()->vecPointingUp();
-  return det.getSignedTwoTheta(samplePos, beamLine, instrumentUpAxis);
+  const V3D &thetaSignAxis = instrument->getReferenceFrame()->vecThetaSign();
+  return det.getSignedTwoTheta(samplePos, beamLine, thetaSignAxis);
 }
 
 /** Returns the 2Theta scattering angle for a detector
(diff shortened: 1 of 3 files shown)


AndreiSavici:
If you had the z and y direction, and the handedness, you know the x axis direction, so the sign of `detectorSignedTwoTheta()` should be uniquely determined. No need to add `<theta-sign axis="x"/>`

soininen:
@AndreiSavici There are always two options for the theta sign axis. By default, theta sign axis is the axis pointing 'up', which is the y axis in the standard Mantid reference frame. However, in some cases (like the D17 reflectometer) we want the axis to be 'the third axis', that is x.

SimonHeybrock (approved):
This now matches https://github.com/mantidproject/mantid/blob/master/Framework/Geometry/src/Instrument/DetectorInfo.cpp#L192, apparently `MatrixWorkspace` had not been updated when this change was made in https://github.com/mantidproject/mantid/pull/19972.

Therefore 👍 .

PR #18315, opened 2016-12-19
Fixed locale issues in MantidTable

soininen:
This PR fixes locale issues when working with TableWorkspaces in MantidPlot.

`MantidTable` writes numerical data to the table cells using STL streams which default to the 'C' locale. When the data is read from the cells, MantidPlot's `Table` converts text back to numerical data using the system locale via functionality in Qt. If the system locale has different numeric format (`LC_NUMERIC` in linuxes) than the 'C' locale, the text written by `MantidTable` cannot be converted back to numbers by `Table`.

**To test:**
1. Start MantidPlot using a locale with different number format compared to the 'C' locale, for example
`LC_NUMERIC=fr_FR.UTF-8 MantidPlot`
might work on some linuxes.
2. Have a TableWorkspace with non-integral values. Use for example this script:
```python
table = CreateEmptyTableWorkspace()
table.addColumn('double', 'x_vals')
table.setPlotType(0, 1)
table.addColumn('double', 'y_vals')
table.addRow([0.23, -0.42])
table.addRow([2.3, -4.2])
table.addRow([23, -43])
table.addRow([66, -55])
```
3. Show data.
4. Plot the 'y_vals' column by right-clicking the column title, and selecting one of the plotting options from the popup.
5. Without these changes, only the last two rows are plotted since they don't contain decimal separators.

Fixes #18275 

*Release notes were updated accordingly.*

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [x] Is the code of an acceptable quality?
- [x] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)? Is it well structured with small focussed classes/methods/functions?
- [ ] Are there unit/system tests in place? Are the unit tests small and test the a class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [x] Do changes function as described? Add comments below that describe the tests performed?
- [ ] How do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant documentation been added/updated?
- [ ] Is user-facing documentation written in a user-friendly manner?
- [ ] Has developer documentation been updated if required?

- Does everything look good? Comment with the ship it emoji but don't merge. A member of `@mantidproject/gatekeepers` will take care of it.

PR #22422, opened 2018-05-14
Improve handling of TableColumns with non-numerical entries

### LINKED ISSUE #22394: TableWorkspace deletes comma separated entries of the form 1,2,3,4
### Expected behavior

A TableWorkspace is created using comma separation to denote several items (in this case detectors). i.e.

[1,2,3,4]

The cells of the table should not change their values. This syntax is used for example in the MuonGroupDetectors (v1) algorithm.

### Actual behavior

When 'Show Data' is run on the TableWorkspace the first time the data appears as entered. Showing the data a second time causes the comma separated values to be replaced by zero.

In the Python Script Window : 

`grouping = CreateEmptyTableWorkspace()`
`grouping.addColumn('vector_int', 'detectors')`
`grouping.addRow([[1,2,3,4]])`

josephframsay:
Fixes a bug where the cell values of a VectorColumn were reset to 0 if the table was viewed. The check that column data is numerical in the handler `MantidTable::cellEdited` was improved to cover all column types. This means the handler doesn't try to interpret the text value as a number (which it was doing before), and instead just reads it as a string

**To test:**

```
t = CreateEmptyTableWorkspace()
t.addColumn("vector_int", "r")
t.addRow([[1,2,3]])
```

Double-click on `t` to view it, then close the table view. Double-click to view again, and the row should still be `1,2,3`, not 0.

Fixes #22394 

**Release Notes** 

[Here](https://github.com/mantidproject/mantid/commit/4c6b211b96a1c127836085041ab5f0ed1b5a6e8f#diff-d41279afe39cfbf234c3ca92983d1c5c)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/inc/MantidAPI/Column.h
+++ b/Framework/API/inc/MantidAPI/Column.h
@@ -100,6 +100,9 @@ public:
   /// Specialized type check
   virtual bool isBool() const = 0;
 
+  /// Are elements of the column interpretable as a number?
+  virtual bool isNumber() const = 0;
+
   /// Must return overall memory size taken by the column.
   virtual long int sizeOfData() const = 0;
 
--- a/Framework/API/test/FunctionDomainGeneralTest.h
+++ b/Framework/API/test/FunctionDomainGeneralTest.h
@@ -29,6 +29,7 @@ public:
   }
   /// Specialized type check
   bool isBool() const override { return false; }
+  bool isNumber() const override { return false; }
   /// Must return overall memory size taken by the column.
   long int sizeOfData() const override {
     throw std::logic_error("Not implemented");
--- a/Framework/DataObjects/inc/MantidDataObjects/PeakColumn.h
+++ b/Framework/DataObjects/inc/MantidDataObjects/PeakColumn.h
@@ -47,6 +47,8 @@ public:
   /// Specialized type check
   bool isBool() const override;
 
+  bool isNumber() const override;
+
   /// Must return overall memory size taken by the column.
   long int sizeOfData() const override;
 
--- a/Framework/DataObjects/inc/MantidDataObjects/TableColumn.h
+++ b/Framework/DataObjects/inc/MantidDataObjects/TableColumn.h
@@ -127,7 +127,7 @@ public:
   /// Type check
   bool isBool() const override { return typeid(Type) == typeid(API::Boolean); }
   bool isNumber() const override {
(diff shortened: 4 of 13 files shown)


reimundILL:
On master I received an unhandled exception when running the test script and viewing `t` (Ubuntu). With the changes it works now!

reimundILL:
Please consider to update the documentation of TableWorkspaces (Concepts). For example, I do not see the `vector_int` description.
I tried for example with `vector_double`:
```Python
t = CreateEmptyTableWorkspace()
t.addColumn("vector_double", "r")
t.addRow([[1.,2.2,3.3]])
```
result is 1,1,2,2,3,3.

josephframsay:
@reimundILL could you describe the steps you went through to get the problem with `vector_double`? I can't replicate this on either Ubuntu or Windows

reimundILL:
Thanks for the documentation update!
I only executed the lines from my last comment in the script interpreter. This may be a different issue since the result is on master as well. When I click on `t` the second time, I get `1,2`. The first view showed `1,2,2,3,3`. (Maybe I reported this wrong in the previous comment)

Here is the script:
```Python
t = CreateEmptyTableWorkspace()
t.addColumn("vector_double", "r")
t.addRow([[1.,2.2,3.3]])
```
If you can replicate this, it may be worth opening a new issue for this case. If only I see this behavior, this would be really strange ...

reimundILL:
Well, maybe it's our French dots and comma uses which are different. @soininen spend some time fixing related issues.

vreimund (approved):
This fixes a bug for `vector_int` table values cleanly and nicely. Tests and documentation were added. I tested on Ubuntu 16.04 where the original behavior for this issue was an unhandled exception.

I think there is still a related issue for `vector_double` table columns:
With the changes applied here, the script
```Python
t = CreateEmptyTableWorkspace()
t.addColumn("vector_double", "r")
t.addRow([[1.,2.2,3.3]])
```
produces now stable `1,2,2,3,3` and the reason may be or not misinterpretation of dots and commas on French machines. On master, I got first `1,2,2,3,3` and then stable `1,2`.

Something about locale : https://github.com/mantidproject/mantid/pull/18315 but likely unrelated.

josephframsay:
Ah yes, I can reproduce this if I change my number format to French!

PR #22425, opened 2018-05-14
LoadLiveData should adjust bin boundaries to fit all events when in PreserveEvents mode

### LINKED ISSUE #22406: Event Listeners do not update X values to reflect range of TOF in events

MikeHart85:
Fixes #22406.

**Description of work**

When in PreserveEvents mode, EventListeners were creating workspaces that had incorrect bin boundaries in their X values. This resulted in some algorithms not being able to work with the workspaces as expected, inconsistent behaviour in the Instrument View, etc.

This PR resolves this by introducing a `void IEventWorkspace::updateAllX(void)` method, which replaces HistogramX for all spectra with a single bin that has bin boundaries which are set to the minimum and maximum TOF based on all events. This method is then called in LoadLiveData every time new data is processed.

This occurs after the pre-processing step and overwrites any user-defined rebinning that occurred there. If a user wants custom binning on live data output, this should be done as a post-processing step instead.

**To test:**

Run `FakeISISEventDAE` with default arguments from sidebar.

Then execute the following in the console:
```python
>>> ConfigService.setFacility('TEST_LIVE')
>>> StartLiveData(Instrument="ISIS_Event", OutputWorkspace="test", UpdateEvery=1, PreserveEvents=True, AccumulationMethod="Replace")
```
Open the Instrument View and observe the Time-of-flight range on the bottom, or view the workspace data and observe the X values.

@matthew-d-jones Maximum event throughput performance should also be tested since we saw it dip very low while trying solutions for this. It looks fine to me now, but maybe you can double check.

**Release Notes** 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


MikeHart85:
Some questions I have for reviewers:

- Should `setAllX` and `updateAllX` be part of `IEventWorkspace` or just `EventWorkspace`? `setAllX` was only in `EventWorkspace` before, but that seemed inconsistent to me so I added it to `IEventWorkspace` along with `updateAllX`.
- Should the `tofmin`/`tofmax` sanitation happen in `updateAllX` or higher up? Does it make sense for `getTofMin`, `getEventXMinMax`, etc to return values like `numeric_limits<double>::min()` and other values that do not pass the HistogramX validator?
- Is `updateAllX` a good name? `fitAllX`? `refitAllX`? `rebinAllX`?

Framework/DataObjects/src/EventWorkspace.cpp
```diff
@@ -580,6 +580,21 @@ void EventWorkspace::setAllX(const HistogramData::BinEdges &x) {
...
+ * all events.
+ */
+void EventWorkspace::updateAllX() {
+  double tofmin, tofmax;
+  getEventXMinMax(tofmin, tofmax);
+
+  // Sanitize TOF min/max to ensure it always passes HistogramX validation.
+  // They would be invalid when number of events are 0 or 1, for example.
+  tofmin = tofmin < tofmax ? tofmin : 0;
+  tofmax = std::max(tofmax, tofmin + DBL_MIN);
```
matthew-d-jones: Should `std::numeric_limits<double>::min()` be used, or is `DBL_MIN` preferred in Mantid?

matthew-d-jones: This validation seems sensible to me.

matthew-d-jones (commented):
I've just had a quick look through, my only other comment is that the method name could indicate that it creates a single bin, `updateSingleXBin` maybe?

docs/source/release/v3.13.0/framework.rst
```diff
@@ -42,6 +42,7 @@ Improved
 
 - :ref:`ConjoinXRuns <algm-ConjoinXRuns>` joins Dx errors if present
 - The algorithm :ref:`SortXAxis <algm-SortXAxis>` has a new input option that allows ascending (default) and descending sorting. Furthermore, Dx values will be considered if present. The documentation needed to be corrected.
+- :ref:`Live Data <algm-StartLiveData>` for events in PreserveEvents mode now produces workspaces that have bin boundaries which encompass the total x-range (TOF) for all events across all spectra.
```
OwenArnold: I think that we need to introduce the behaviour change into the concepts documentation or similar, not just into the release notes.

OwenArnold (changes requested):
Looks good. See comment about documentation. Also missing some clone test, as discussed in the MatrixWorkspaceTest.

Framework/DataObjects/src/EventWorkspace.cpp
martyngigg: `std::numeric_limits<double>::min()` would be preferable.

martyngigg: I also agree with Matthew that it would be nice to indicate that this will completely disregard the binning that is currently set and reset it to a single bin.  Something like `.resetAllXToSingleBin()`?

martyngigg (changes requested):
This all looks sensible. Can we also update the `StartLiveData` documentation to provide a comment/warning that using `Rebin` for `EventWorkspaces` in the `Processing` step will be ignored?

Framework/DataObjects/src/EventWorkspace.cpp
MikeHart85: `numeric_limits<double>::min()` is the smallest (largest?) negative number a double can represent, `DBL_MIN` is the smallest positive normal double.

So this is meant ensure that: `tofmin < tofmax` and `tofmax >= tofmin + DBL_MIN`.

If `tofmin = <double>::max()` and `tofmax = <double>::min()` (which `getEventXMinMax` will produce if there are no events), this will force them to `0` and `DBL_MIN` respectively.

If `tofmin = tofmax` (which `getEventXMinMax` will produce if there is exactly one event), this will leave `tofmin` alone by add `DBL_MIN` to `tofmax` to satisfy the distance requirement.

Should I make the comment in the code more detailed to explain this better?

matthew-d-jones: I think `numeric_limits<double>::lowest()` is `-1*numeric_limits<double>::max()`
`numeric_limits<double>::min()` should be equivalent to `DBL_MIN`

MikeHart85: You're right, my bad!

martyngigg:
The code changes look good now. Once the documentation updates are in place I think this is good to go.

Framework/LiveData/src/LoadLiveData.cpp
```diff
@@ -391,6 +391,32 @@ Workspace_sptr LoadLiveData::appendMatrixWSChunk(Workspace_sptr accumWS,
...
+ *
+ * Ensures bin boundaries encompass all events currently in the workspace.
+ * This will overwrite any rebinning that was previously.
+ *
+ * Input should be an EventWorkspace or WorkspaceGroup containing
+ * EventWorkspaces. Any other workspace types are ignored.
+ *
+ * @param workspace :: Workspace that will have its bins reset
+ */
+void LoadLiveData::resetAllXToSingleBin(API::Workspace_sptr workspace) {
```
OwenArnold: Minor point, but argument need not be a `shared_ptr<Workspace>` could just be a `Workspace &`. Just clears up the semantics a little since we are doing nothing with the shared_ptr aspects here. The code below could be changed to a regular `dynamic_cast` in that case.

MikeHart85: Would a `Workspace_sptr&` be more appropriate, since both `chunkWS` and `m_accumWS` are already `Workspace_sptr`? Or would `Workspace&` still be better here?

MikeHart85:
This seems mostly done, just needs performance testing and there is one remaining issue to resolve:

@matthew-d-jones raised the point that if we don't fix the bins before the `Process` step, then they would remain uninitialized for any algorithm or script in that step. This would be inconsistent and could be very confusing. So I've factored out the fix into a method and now call it twice, once for the `chunkWS` with the newly added events, and once on `m_accumWS` after the events have been added.

This means that in Replace mode we end up calling it twice on the same workspace. This is a bit wasteful... but if we don't call it the second time then Rebinning done in the Process step would stick in Replace mode but get overwritten in Add mode... which would be worse, I think? So this is probably fine.

However, fixing the bins on the new events chunk has caused `test_ProcessToMDWorkspace_and_Add` in `LoadLiveDataTest` to fail in a strange way. I'm having trouble figuring out why because I don't fully understand how it is supposed to work.

Without the fix (nonsensical bin boundaries), the test always produces two MDWorkspaces with `getNPoints() == 200` and `400`. With the fix (correct bin boundaries) it fails because `getNPoints()` returns randomly lower values like `198` and `392`... always a few short.

I've tried enlarging the bin boundaries. First by `DBL_MIN`, which didn't seem to change anything. Then `1.0`, which still failed most of the time but succeeded occasionally. Then `10.0`, which still fails mostly but succeeds a bit more often.

It seems to be going wrong in `ConvToMDEventsWS::convertEventList`, where...
```cpp
    if (!m_QConverter->calcMatrixCoord(val, locCoord, signal, errorSq))
      continue; // skip ND outside the range
```
... is hitting the `continue`. Since this doesn't happen when the bin boundaries are wrong, but does happen when they are correct, it seems like either there's a mistake here. Either in this behaviour or in what the test is expecting. But I'm not sure which, and the code in both is quite old (4-6 years).

Any ideas or suggestions who to ask?

martyngigg:
@MikeHart85 I was surprised that the values seem to change - I would have expected them to still be consistently the same and wrong. I think this problem is coming from the [`TestDataListener`](https://github.com/mantidproject/mantid/blob/master/Framework/LiveData/test/TestDataListener.cpp#L25) where it seeds the random number generator with the current time. I think it would be better if we could optionally fix a seed to produce a known sequence of events here.

You're correct in that the form of the [`extractDataImpl`](https://github.com/mantidproject/mantid/blob/master/Framework/LiveData/test/TestDataListener.cpp#L119) for the test listener should lead us to expect 200 events in the first output workspace - perhaps @abuts might be able to point us at the parts of ConvertToMD that deal with computing the expected extents for the new MD workspace? I presume that is where it is falling over...

Framework/LiveData/src/LoadLiveData.cpp
```diff
@@ -391,6 +391,41 @@ Workspace_sptr LoadLiveData::appendMatrixWSChunk(Workspace_sptr accumWS,
...
+ *
+ * Ensures bin boundaries encompass all events currently in the workspace.
+ * This will overwrite any rebinning that was previously done.
+ *
+ * Input should be an EventWorkspace or WorkspaceGroup containing
+ * EventWorkspaces. Any other workspace types are ignored.
+ *
+ * @param workspace :: Workspace(Group) that will have its bins reset
+ */
+void LoadLiveData::resetAllXToSingleBin(API::Workspace &workspace) {
```
OwenArnold: I think we are dealing with minor points here, that I definitely wouldn't want to see delaying things. However, as per your comments on slack... since the `shared_ptr` is nullable, `API::Workspace * const workspace` might be a better signature for the input, and allow you to avoid the try-catch provided below.

MikeHart85:
The build is failing due to the issue with `ConvToMDEventsWS::convertEventList` described above. This isn't a problem caused by anything added by this PR, but rather a separate issue brought to light as a result. It will need further investigating, but currently I'm tied up working on something else.

Additionally, before this can be merged, @matthew-d-jones will need to run performance tests to confirm that the changes here are sufficiently performant.

martyngigg:
@MikeHart85 Can I move this back to 3.14?

matthew-d-jones:
I am happy that the testing Mike has done demonstrates that the performance after these changes is at least fine for ISIS use, so I would be happy for it to be merged. I will do more rigorous performance testing as soon as soon as I am able, this is likely to be the week of 16th July.

MikeHart85:
Tracked down what was going wrong together with @abuts.

Events were being discarded in `ConvToMDEventsWS::convertEventList` here...

```c++
    if (!m_QConverter->calcMatrixCoord(val, locCoord, signal, errorSq))
      continue; // skip ND outside the range
```

... or in `MDBoxBase::addEvents` here:

```c++
      if (extents[d].outside(x)) {
        badEvent = true;
        break;
      }
```

In both cases it looks like a matter of rounding errors on the calculated valid extents box.

This test didn't fail before because, due to the incorrect bin boundaries, `ConvertToDiffractionMDWorkspace3::calculateExtentsFromData` was returning default min and max values of `-50` and `50` respectively in each dimension. With the corrected bin boundaries, it was now producing min and max values between `-1.0` and `1.0`.

Those min and max values are much more sensible. However, they fit the events exactly and further calculations introduce rounding errors. This caused events near the edges to be discarded later.

In the above change I widened the calculated extents by 5% to leave room for error. I tested this and other margins my machine by running `LoadLiveDataTest::test_ProcessToMDWorkspace_and_Add` repeatedly in a loop, with the following results:

- 1% margin: Failed after 63 successful attempts
- 2% margin: Failed after 498 successful attempts
- 5% margin: Succeeded 11822+ times without failure

However, the tests here appear to still be failing consistently with a 5% margin. Not sure what is going on. Perhaps it's because I was testing with the Debug build. I'll investigate further in Release mode.

MikeHart85:
No, the test still succeeds 1400+ times on my machine (Ubuntu) in Release mode. Very strange.

MikeHart85:
Increasing error margin to 10% just to see if this will make a difference on the Jenkins servers. I'm not liking this inconsistency between machines though.

NickDraper:
Given the that to get this ready for release will mean touching some of the core code, then I suggest we move this out to the next release.  This does not in any way stop you from working on it, it just isolates the release from any unanticipated side effects of the changes you need to make.

matthew-d-jones:
I tested performance using Kafka running in docker on localhost and prepopulated with a run of data containing 1e6 events per message (1e7 events/s at 10 pulses/s) and 3000 messages (5 mins of data).
To consume this data the total run time reported by MonitorLiveData was:
```
On branch 22406:                       On master:
4 min 23 s                             4 min 24 s
4 min 19 s                             4 min 25 s
4 min 24 s                             4 min 20 s
```
I conclude that any performance change is insignificant.

MikeHart85:
Rebased for clang-format 5.0 update, make sure you checkout a clean copy of the branch if you had one already.

abuts:
The change rejecting event at x>min is extremely dangerous and would lead to a difficult to detect crashes. 
There is piece of code, which calculates the index of an event in a limited size box using ind = x/bin_size where bin_size = Dx/capacity and if x== bin_size index would be capacity+1.

As it is usually 3-4D boxes, mapped as 1D array, most of such events would go into next dimension except once a while a point in last dimension comes and causes out-of memory. 

I would not recommend changes as in #5206e66 
``
  if (Coord[1] < dim_min[1] || Coord[1] > dim_max[1])
    return false;
``
and would keep the older version: 
``
  if (Coord[1] < dim_min[1] || Coord[1] >= dim_max[1])
    return false;
``

abuts:
It looks like a unit tests is introduced to check the problem I have outlined.

MikeHart85:
Thanks for the comments @abuts, I was unsure about the consequences of making that change and hoping to get your feedback on that!

The reason I made that change is that a `min < x <= max` comparison causes events on the max outer edge to be lost:

mdconv_01

Whereas if we change the logic to `min < x < max`, all events stay inside (although those on an inner edge end up in a different box):

mdconv_02

This wasn't a visible problem before, because the outer box was set to huge extents (+/- 50). But with the changes in this PR, it now wraps around events very closely, so it's very likely for an event to sit on a max outer edge.

Is this the index calculating function you are talking about: [MDGridBox::calculateChildIndex](https://github.com/mantidproject/mantid/blob/6a066315935149d361e349e639467b5abad3fa4a/Framework/DataObjects/inc/MantidDataObjects/MDGridBox.tcc#L1734-L1742)?

This function is called in two places: [MDGridBox::addEvent](https://github.com/mantidproject/mantid/blob/6a066315935149d361e349e639467b5abad3fa4a/Framework/DataObjects/inc/MantidDataObjects/MDGridBox.tcc#L1637-L1649) and [MDGridBox::addEventUnsafe](https://github.com/mantidproject/mantid/blob/6a066315935149d361e349e639467b5abad3fa4a/Framework/DataObjects/inc/MantidDataObjects/MDGridBox.tcc#L1668-L1680).

Both of these have this code block right after calling it:

```c++
  size_t cindex = calculateChildIndex(event);

  // We can erroneously get cindex == numBoxes for events which fall on the
  // upper boundary of the last child box, so add these events to the last box
  if (cindex == numBoxes)
    cindex = numBoxes - 1;
```

Does this prevent the issue you are talking about? If yes, I think this check and correction should be moved inside `calculateChildIndex`, at the end, to ensure it's always done when calling it.

Actually, this is also the (hopefully final) place where I am still losing events sometimes. Very rarely, `calculateChildIndex` will return, for example, `cindex = 12` event though `numBoxes == 8`. So this even bypasses this safety mechanism, because it's only checking for `cindex == numBoxes`.

I was thinking of just changing the comparison to `cindex >= numBoxes`, but it would be much better to figure out why it would return 12 to begin with and fix that. Could it be this is because an event is sitting exactly on the `max` of every dimension?

abuts:
> "@MikeHard85 The reason I made that change is that a min < x <= max comparison causes events on the max outer edge to be lost"  -- the inconsistency between 1 and 2 boxes are obvious. It does not look a problem at this state, but when you would try to do more complex operation, (e.g. combining different cuts) this will come to haunt you.  I would better not to do this change. 

As there is unit test to check limit rejection, I suspect that calculateChildIndex is not the only place where this assumption is used. ( I have not written this test)

"Actually, this is also the (hopefully final) place where I am still losing events sometimes. Very rarely, calculateChildIndex will return, for example, cindex = 12 Could it be this is because an event is sitting exactly on the max of every dimension?" 

It looks like this is exactly what happens. (for 2x2x2 box, the number will be higher for bigger box)

The change would allow you to fix this particular problem but would cause different and difficult to detect problem in a future. I would avoid making this change.

MikeHart85:
I've change it back to `Coord[x] >= dim_max[x]` and removed the `4 * FLT_EPSILON` margins just above (which where having the same and even bigger effect). 

And I've reinstated the margins in `calculateExtentsFromData` that we put it. I had removed them because even at 10% they weren't big enough before... but now, with all the other fixes, even `1 - 1e-5` multipliers seem to be enough (at least when I tested it locally):

```c++
  for (size_t i = 0; i < maxVal.size(); ++i) {
    minVal[i] *= (1 - 1.e-5 * boost::math::sign(minVal[i]));
    if (fabs(minVal[i]) < 1.e-5)
      minVal[i] = -1.e-5;
    maxVal[i] *= (1 + 1.e-5 * boost::math::sign(maxVal[i]));
    if (fabs(maxVal[i]) < 1.e-5)
      maxVal[i] = 1.e-5;
  }
```

I could try even smaller margins, but I think this should be fine since we're just expanding the extents of the outer box?

abuts:
1.e-5 should be reasonable error margin for single precision. If it works, I would keep it as it is. I am highly surprised why 10% would be necessary -- this makes me suspect something wrong elsewhere.

MikeHart85:
Ah, looks like that unit test failure was caused by a different place where I changed `x >= max`. That might be why you didn't recognize it.

Yes, the 10% was just to test if a margin that big would be enough, and it wasn't... so I removed it and went to look for problems in other places. But it looks like a small margin is needed in that place after all, to prevent losing event on the outer max edge.

I've run the test locally a few thousand times now, and it looks like it works well like this.

Framework/LiveData/src/LoadLiveData.cpp
```diff
@@ -277,18 +272,49 @@ void LoadLiveData::addMatrixWSChunk(const std::string &algoName,
...
+  ws_names_to_merge.append(chunkName);
+
+  IAlgorithm_sptr alg = this->createChildAlgorithm("MergeMD");
+  alg->setPropertyValue("InputWorkspaces", ws_names_to_merge);
+  alg->executeAsChildAlg();
+
+  // No longer needed in ADS
+  AnalysisDataService::Instance().remove(chunkName);
+
   if (!alg->isExecuted()) {
```
OwenArnold: I know you didn't add this, but it would probably be best if we propagated the original exception. That could be done by setting `Algorithms::rethrows(true)`. Not convinced we should do that as part of the PR though.

Framework/MDAlgorithms/src/ConvertToDiffractionMDWorkspace3.cpp
```diff
@@ -114,6 +114,18 @@ void ConvertToDiffractionMDWorkspace3::calculateExtentsFromData(
   std::replace(maxVal.begin(), maxVal.end(), INF, DEFAULT_BOUND);
   std::replace(minVal.begin(), minVal.end(), MAX_DBL, -DEFAULT_BOUND);
   std::replace(maxVal.begin(), maxVal.end(), -MAX_DBL, DEFAULT_BOUND);
+
+  // Increases bounds by a small margin and ensures they aren't too close to
+  // zero. This is to prevent events from being incorrectly discarded as out
+  // of bounds by calcMatrixCoord functions.
```
OwenArnold: You pointed out a related bug with index calculation. Comment should include link/note to the `MDBox` box index calculation code error that is coupled with events on upper boundaries.

OwenArnold (changes requested):

DanNixon (commented):
Looks good to me. To me expanding the MD extents seems like the least disruptive solution.

Framework/DataObjects/src/EventWorkspace.cpp
```diff
@@ -590,8 +590,12 @@ void EventWorkspace::resetAllXToSingleBin() {
...
   // Sanitize TOF min/max to ensure it always passes HistogramX validation.
   // They would be invalid when number of events are 0 or 1, for example.
-  tofmin = tofmin < tofmax ? tofmin : 0;
-  tofmax = std::max(tofmax, tofmin + std::numeric_limits<double>::min());
+  if (tofmin > tofmax) {
+    tofmin = 0;
+    tofmax = std::numeric_limits<double>::min();
+  } else if (tofmin == tofmax) {
+    tofmax += std::numeric_limits<double>::min();
+  }
```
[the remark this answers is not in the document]
MikeHart85: This isn't quite the same because this change would alter tofmax in the case `tofmax > tofmin`, when it should stay unaltered.

Also, somewhat odd:
```c++ 
std::numeric_limits<double>::min() == 2.22507e-308
```
... but:
```c++
std::nexttoward(0.0, std::numeric_limits<double>::max()) == 4.94066e-324
```

Framework/LiveData/src/LoadLiveData.cpp
MikeHart85: Looking into this a bit I found that these algorithms would rethrow anyway, because they are child algorithms. So that condition and bit of code was dead anyway (every path to isExecuted == false that I could find also led to a throw); I've simply removed it.

Framework/DataObjects/src/EventWorkspace.cpp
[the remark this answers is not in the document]
DanNixon: `std::nexttoward` includes subnornal values, so

```cpp
std::nexttoward(0.0, std::numeric_limits<double>::max()) == std::numeric_limits<double>::denorm_min()
```

OwenArnold (approved):

martyngigg (approved):
This looks good to me now too. I agree that extending the extents slightly makes the most sense.

Issue #22433, opened 2018-05-16
Remove fit benchmarking code


Anders-Markvardsen:
Relevant fit benchmarking code has been moved to mantidproject/fitbenchmarking. In this issue remove this code from this repository, thereby completing the move this tool.

It has been agreed with senior developers of Mantid it is better to continue the developement of this tool in a separate repository, where e.g. separate testing of versions of mantid on the full set of fit benchmarking problem, which currently takes 3 hours, can be more easier accommodated.

Anders-Markvardsen:
This was fixed in #22437

PR #22437, opened 2018-05-17
remove fit benchmarking files

Anders-Markvardsen:
See #22433 

**To test:**

* Check that all tests are passing
* check that all that needs moving has been moved to https://github.com/mantidproject/fitbenchmarking

No release note required


AnthonyLim23 (approved):
The only thing that I can see that is missing from the new folder are the Muon tests (but they were never in master). Everything else looks good.

PR #22449, opened 2018-05-21
Error reporting: Display reporter on application segfaults

### LINKED ISSUE #22293: Refactor error_report dialog window to be more self contained.

Matthew-Andrew:
**Description of work**

This PR updates the ErrorReporter dialog so that it catches hard crashes to desktop as well as errors which reach the last chance error handler.

**To test:**

* These tests need to be done on a package version of mantid rather than a local build.
* It also needs to be tested on all OS's

* Open mantid
* Turn usage reporting on/restart mantid
* Run the segfault algorithm and check that the error report dialog appears.
* Check that no report is sent if you select no report
* Check that nothing appears if usage reporting is turned off.

Fixes #22293  . 

**Release Notes** 
Has been added to release notes.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


buildconfig/CMake/DarwinSetup.cmake
```diff
@@ -100,6 +99,14 @@ set ( CMAKE_INSTALL_PREFIX "" )
 set ( CPACK_PACKAGE_EXECUTABLES MantidPlot )
 set ( INBUNDLE MantidPlot.app/ )
 
+# Copy the launcher script to the correct location
+configure_file ( ${CMAKE_MODULE_PATH}/Packaging/osx/Mantid_osx_launcher 
+                 ${CMAKE_BINARY_DIR}/bin/MantidPlot.app/Contents/MacOS/Mantid_osx_launcher COPYONLY )
+# Copy the error dialog script to the correct location
+configure_file ( ${CMAKE_MODULE_PATH}/Packaging/osx/error_dialog_app.py.in
```
Mantid-Matthew: @keeeto do we still need error_dialog_app.py.in ?

scripts/ErrorReporter/error_dialog_app.py
```diff
@@ -0,0 +1,30 @@
...
+
+import mantid # noqa
+from ErrorReporter import resources # noqa
+from mantid.kernel import UsageService # noqa
+from ErrorReporter.error_report_presenter import ErrorReporterPresenter # noqa
+from ErrorReporter.errorreport import CrashReportPage # noqa
+
+
+def main():
+    if not UsageService.isEnabled():
```
martyngigg: There's a minor issue here that if you have usage reports disabled and the crash happens the exit code of the process is `0`. I think we need to make sure main always returns an integer and then main gets called as `sys.exit(main())` from below.

scripts/ErrorReporter/error_dialog_app.py
```diff
@@ -0,0 +1,30 @@
...
+        return
+    app = QtGui.QApplication(sys.argv)
+    form = CrashReportPage(show_continue_terminate=False)
+    ErrorReporterPresenter(form, command_line_args.exit_code)
+    app.exec_()
```
martyngigg: I think `return app.exec_()` is then what we want here also.

martyngigg (changes requested):
This looks really good now. I think we just need to fix up the exit code situation from the app and then it's good to go.

martyngigg (approved):
This looks good - let's get it merged and out in the wild.

PR #22452, opened 2018-05-21
Muon - ApplyMuonDetectorGrouping Algorithm

TWJubb:
Added an algorithm to group detectors within a workspace or workspace group, for use in the Muon Analysis GUI

**To test:**

Open the Muon Analysis Interface (Interfaces->Muon->Muon Analysis), select the EMU instrument and Load the current run. Now open the algorithm Data Handling->Grouping->ApplyMuonDetectorGrouping. The Workspaces should be filled as MuonAnalysis and EMU<current run number>. Call the groupName "test", and the Grouping 1,2,3. Set the plotType to Counts. Leave all the other boxes.

When the algorithm is run, two new workspaces should be created in the EMU<current run number> workspace group. The workspace whose name ends in "_Raw" should have y-values equal to the sum of the first three histograms in the MuonAnalysis workspace. The errors should be added in quadrature.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


AnthonyLim23:
looking at the errors it is not finding the algorithms correctly

martyngigg:
The build error on gcc & clang is due to a weird edge case with `getProperty` - you can't use it to assign to a `std::string` object that has already been created, i.e.

```
m_groupName = getProperty(...)
```
doesn't work because of a compiler abiguity. In this case though as you know you want a string you can just use `getPropertyValue` instead.

AnthonyLim23:
thanks @martyngigg. I think we may potentially be better off getting rid of the private member values, it depends on how they are used

Framework/Muon/src/ApplyMuonDetectorsGrouping.cpp
```diff
@@ -0,0 +1,350 @@
...
+  options.subtractedPeriods = m_subtractedPeriods;
+  options.timeZero = m_timeZero;
+  options.loadedTimeZero = m_loadedTimeZero;
+  options.timeLimits.first = m_Xmin;
+  options.timeLimits.second = m_Xmax;
+  options.rebinArgs = isRaw ? "" : m_rebinArgs;
+  options.plotType = m_plotType;
+  options.groupPairName = m_groupName;
+
+  IAlgorithm_sptr alg =
```
martyngigg: Is it worth considering having a method `runMuonProcess` that ran the algorithm and returned the workspace that could encapsulate the code from here to the end of this function? It would only have to accept a `MuonAnalysisOptions` type.

Framework/Muon/src/ApplyMuonDetectorsGrouping.cpp
```diff
@@ -0,0 +1,350 @@
...
+      AlgorithmManager::Instance().createUnmanaged("MuonProcess");
```
martyngigg: Here you can use a method called `createChildAlgorithm()` provided by the `Algorithm` base class. It creates the unmanaged algorithm, initializes it, calls `setChild(true)` and also sets a dummy name on the output workspaces types.

Framework/Muon/inc/MantidMuon/ApplyMuonDetectorsGrouping.h
```diff
@@ -0,0 +1,101 @@
...
+  /// Execution code
+  void exec() override;
+  ///  Store the input properties as private member variables
+  WorkspaceGroup_sptr getUserInput(Workspace_sptr &inputWS,
+                                   WorkspaceGroup_sptr &groupedWS);
+  /// Creates and analyses a workspace, if isRaw does not rebin.
+  Workspace_sptr createAnalysisWorkspace(Workspace_sptr &inputWS, bool isRaw);
+  /// Sets algorithm properties according to options.
+  void
+  setProcessAlgorithmProperties(IAlgorithm_sptr alg,
```
martyngigg: Similarly to the above comment I would use `Algorithm&` as the argument type to indicate that the function does nothing with managing the memory.

AnthonyLim23:
@martyngigg should the documentation for this go into algorithms?

martyngigg (changes requested):
This is great start. I have some initial observations from a first look over the code that are mentioned in line.

In addition I don't think the `Framework/Muon/inc/MantidMuon/CMakeLists.txt` file is needed? Can we `git rm` that?

martyngigg:
@AnthonyLim23 Yes, the documentation should go in to the algorithms section. Keith is currently make this section of the user documentation more navigable so as long as the categories for this are Muon we should be fine.

Framework/Muon/inc/MantidMuon/ApplyMuonDetectorsGrouping.h
```diff
@@ -0,0 +1,101 @@
...
+               Muon::AnalysisOptions &options);
+  /// Clip Xmin/Xmax to the range in the input WS
+  void clipXRangeToWorkspace(const API::WorkspaceGroup &ws,
+                             Muon::AnalysisOptions &options);
+  /// Creates and analyses a workspace, if isRaw does not rebin.
+  API::Workspace_sptr
+  createAnalysisWorkspace(const API::Workspace_sptr &inputWS, bool noRebin,
+                          Muon::AnalysisOptions &options);
+  /// Sets algorithm properties according to options.
+  void setMuonProcessAlgorithmProperties(API::IAlgorithm_sptr alg,
```
AnthonyLim23: add a `&` to alg

Framework/Muon/inc/MantidMuon/ApplyMuonDetectorsGrouping.h
```diff
@@ -0,0 +1,101 @@
...
+                                         const AnalysisOptions &options) const;
+
+  /// Set algorithm properties (input workspace, and period properties)
+  /// according to the given options. For use with
+  /// MuonProcess.
+  void
+  setMuonProcessPeriodProperties(API::IAlgorithm_sptr alg,
```
AnthonyLim23: see above

Framework/Muon/inc/MantidMuon/MuonAlgorithmHelper.h
```diff
@@ -0,0 +1,78 @@
...
+#include "MantidAPI/WorkspaceGroup.h"
+#include "MantidKernel/System.h"
+
+#include <string>
+#include <vector>
+
+namespace Mantid {
+namespace Muon {
+/// Types of entities we are dealing with
+enum ItemType { Pair, Group };
```
AnthonyLim23: could we add a `const `here?

Framework/Muon/inc/MantidMuon/MuonAlgorithmHelper.h
```diff
@@ -0,0 +1,78 @@
...
+
+/// Possible plot types users might request
+enum PlotType { Asymmetry, Counts, Logarithm };
```
AnthonyLim23: same as above

Framework/Muon/inc/MantidMuon/MuonAlgorithmHelper.h
```diff
@@ -0,0 +1,78 @@
...
+};
+
+/// Whether multiple fitting is enabled or disabled
+enum class MultiFitState { Enabled, Disabled };
+} // namespace Muon
+
+namespace MuonAlgorithmHelper {
+
+/// Returns a first period MatrixWorkspace in a run workspace
+Mantid::API::MatrixWorkspace_sptr firstPeriod(API::Workspace_sptr ws);
```
AnthonyLim23: add a &

Framework/Muon/src/ApplyMuonDetectorsGrouping.cpp
```diff
@@ -0,0 +1,379 @@
...
+                  "IDs or hyphenated ranges of IDs.",
+                  Direction::Input);
+
+  declareProperty(
+      "AnalysisType", "Counts",
+      boost::make_shared<Kernel::ListValidator<std::string>>(g_analysisTypes),
+      "The type of analysis to perform on the spectra.", Direction::Input);
+
+  declareProperty(
+      "TimeMin", 0.0,
```
AnthonyLim23: change the default to 0.1, that is typically when first good data is

Framework/Muon/src/ApplyMuonDetectorsGrouping.cpp
```diff
@@ -0,0 +1,379 @@
...
+      "Start time for the data in ms. Only used with the asymmetry analysis.",
+      Direction::Input);
+  setPropertySettings("TimeMin",
+                      make_unique<Kernel::EnabledWhenProperty>(
+                          "AnalysisType", Kernel::IS_EQUAL_TO, "Asymmetry"));
+
+  declareProperty(
+      "TimeMax", 30.0,
```
AnthonyLim23: I think the time is typically 32

Framework/Muon/src/ApplyMuonDetectorsGrouping.cpp
```diff
@@ -0,0 +1,379 @@
...
+      Direction::Input);
+  setPropertySettings("TimeMax",
+                      make_unique<Kernel::EnabledWhenProperty>(
+                          "AnalysisType", Kernel::IS_EQUAL_TO, "Asymmetry"));
+
+  declareProperty("RebinArgs", emptyString,
+                  "Rebin arguments. No rebinning if left empty.",
+                  Direction::Input);
+
+  declareProperty("TimeOffset", 0.0,
```
AnthonyLim23: if we change the t_min default does this need to change?

Framework/Muon/src/ApplyMuonDetectorsGrouping.cpp
```diff
@@ -0,0 +1,379 @@
...
+    const Muon::AnalysisOptions &options, const std::string &groupWSName) {
+
+  Muon::DatasetParams params;
+  // don't fill in instrument, runs, periods; not required.
+  params.label = groupWSName;
+  params.itemType = Muon::ItemType::Group;
+  params.itemName = options.groupPairName;
+  params.plotType = options.plotType;
+  params.version = 1;
+  const std::string wsName = generateWorkspaceName(params);
```
AnthonyLim23: could just return this line

Framework/Muon/src/ApplyMuonDetectorsGrouping.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+  // Store all the options needed for MuonProcess
+  Grouping grouping;
+  grouping.description = "no description";
+  grouping.groupNames.emplace_back(this->getPropertyValue("groupName"));
+  grouping.groups.emplace_back(this->getPropertyValue("Grouping"));
+
+  options.grouping = grouping;
+  options.summedPeriods = this->getPropertyValue("SummedPeriods");
+  options.subtractedPeriods = this->getPropertyValue("SubtractedPeriods");
```
AnthonyLim23: I think you do not need `this`

Framework/Muon/src/ApplyMuonDetectorsGrouping.cpp
```diff
@@ -0,0 +1,379 @@
...
+  // Find index of a name in a collection
+  const auto indexOf = [](const std::string &name,
+                          const std::vector<std::string> &collection) {
+    return std::distance(collection.begin(),
+                         std::find(collection.begin(), collection.end(), name));
+  };
+  if (isContainedIn(options.groupPairName, options.grouping.groupNames)) {
+    // Group
+    std::string outputType;
+    switch (options.plotType) {
```
AnthonyLim23: can we rewrite this to remove the breaks?

Framework/Muon/src/MuonAlgorithmHelper.cpp
```diff
@@ -0,0 +1,241 @@
...
+  label << std::setw(zeroPadding) << std::setfill('0') << std::right;
+
+  for (auto range : ranges) {
+    std::string firstRun = std::to_string(range.first);
+    std::string lastRun = std::to_string(range.second);
+    label << firstRun;
+    if (range.second != range.first) {
+      // Remove the common part of the first and last run, so we get e.g.
+      // "12345-56" instead of "12345-12356"
+      for (size_t i = 0; i < firstRun.size() && i < lastRun.size(); ++i) {
```
AnthonyLim23: could we rewrite this to remove the break? Would a while loop work?

Framework/Muon/src/MuonAlgorithmHelper.cpp
```diff
@@ -0,0 +1,241 @@
...
+std::vector<std::pair<int, int>>
+findConsecutiveRuns(const std::vector<int> &runs) {
+  // Groups to output
+  std::vector<std::pair<int, int>> ranges;
+  // Sort the vector to begin with
+  std::vector<int> runNumbers(runs); // local copy
+  std::sort(runNumbers.begin(), runNumbers.end());
+
+  // Iterate through vector looking for consecutive groups
+  auto a = runNumbers.begin();
```
AnthonyLim23: could we have better variable names than `a` and `b `please

Framework/Muon/test/ApplyMuonDetectorsGroupingTest.h
```diff
@@ -0,0 +1,263 @@
...
+    delete suite;
+  };
+
+  void test_Init() {
+    ApplyMuonDetectorGrouping alg;
+    TS_ASSERT_THROWS_NOTHING(alg.initialize());
+    TS_ASSERT(alg.isInitialized());
+  };
+
+  void test_output_produced() {
```
AnthonyLim23: `test_outputProduced`

Framework/Muon/test/ApplyMuonDetectorsGroupingTest.h
```diff
@@ -0,0 +1,263 @@
...
+
+    auto wsOut = wsGroup->getItem("inputGroup; Group; test; Counts; #1_Raw");
+
+    CompareWorkspaces algCompare;
+    TS_ASSERT_THROWS_NOTHING(algCompare.initialize());
+    TS_ASSERT(algCompare.isInitialized());
+    TS_ASSERT_THROWS_NOTHING(
+        algCompare.setProperty("Workspace1", ws->getName()));
+    TS_ASSERT_THROWS_NOTHING(
+        algCompare.setProperty("Workspace2", wsOut->getName()));
```
AnthonyLim23: CompareWorkspaces doesnt compare data by default http://docs.mantidproject.org/nightly/algorithms/CompareWorkspaces-v1.html

Framework/Muon/test/ApplyMuonDetectorsGroupingTest.h
```diff
@@ -0,0 +1,263 @@
...
+    TS_ASSERT_DELTA(wsOut->readE(0)[0], 3.162, 0.001);
+    TS_ASSERT_DELTA(wsOut->readE(0)[0], 3.162, 0.001);
+    TS_ASSERT_DELTA(wsOut->readE(0)[0], 3.162, 0.001);
+
+    TS_ASSERT(wsGroup);
+    if (wsGroup) {
+      TS_ASSERT_EQUALS(wsGroup->getNumberOfEntries(), 2);
+    }
+  }
+
```
AnthonyLim23: can we have a test that does asymmetry and one that looks at multiple period data

Framework/Muon/inc/MantidMuon/MuonAlgorithmHelper.h
```diff
@@ -0,0 +1,78 @@
...
+#include "MantidAPI/WorkspaceGroup.h"
+#include "MantidKernel/System.h"
+
+#include <string>
+#include <vector>
+
+namespace Mantid {
+namespace Muon {
+/// Types of entities we are dealing with
+const enum ItemType { Pair, Group };
```
martyngigg: A name for these types is great. C++11 introduced the concept of ["Scoped Enumerations](http://en.cppreference.com/w/cpp/language/enum) that makes them even better and these should generally be preferred these days. Can we use these here?

Framework/Muon/inc/MantidMuon/MuonAlgorithmHelper.h
```diff
@@ -0,0 +1,78 @@
...
+
+/// Possible plot types users might request
+const enum PlotType { Asymmetry, Counts, Logarithm };
```
martyngigg: Same question as above.

Framework/Muon/inc/MantidMuon/MuonAlgorithmHelper.h
```diff
@@ -0,0 +1,78 @@
...
+  std::pair<double, double> timeLimits; /// Min, max X values
+  std::string rebinArgs;          /// Arguments for rebin (empty to not rebin)
+  std::string groupPairName;      /// Name of group or pair to use
+  Mantid::API::Grouping grouping; /// Grouping to use
+  PlotType plotType = {};         /// Type of analysis to perform
+  explicit AnalysisOptions() {}
+};
+
+/// Whether multiple fitting is enabled or disabled
+enum class MultiFitState { Enabled, Disabled };
```
martyngigg: Same question regarding the `enum` as above.

Framework/Muon/inc/MantidMuon/MuonAlgorithmHelper.h
```diff
@@ -0,0 +1,78 @@
...
+  std::string summedPeriods;            /// Set of periods to sum
+  std::string subtractedPeriods;        /// Set of periods to subtract
+  double timeZero = 0;                  /// Value to use for t0 correction
+  double loadedTimeZero = 0;            /// Time zero from data file
+  std::pair<double, double> timeLimits; /// Min, max X values
+  std::string rebinArgs;          /// Arguments for rebin (empty to not rebin)
+  std::string groupPairName;      /// Name of group or pair to use
+  Mantid::API::Grouping grouping; /// Grouping to use
+  PlotType plotType = {};         /// Type of analysis to perform
+  explicit AnalysisOptions() {}
```
martyngigg: Is this constructor necessary?

Framework/Muon/inc/MantidMuon/ApplyMuonDetectorsGrouping.h
```diff
@@ -0,0 +1,101 @@
...
+  int version() const override { return (1); }
+  /// Algorithm's category for identification
+  const std::string category() const override {
+    return "DataHandling\\Grouping;Transforms\\Grouping";
+  }
+  /// Algorithm's summary for identification
+  const std::string summary() const override {
+    return "Group several muon detector IDs together and perform an analysis "
+           "(either counts or asymmetry).";
+  }
```
AnthonyLim23: should add the `seeAlso `function. For now I suppose it will be just `MuonProcess`, but will include both of the new algorithms once they are complete. see https://github.com/mantidproject/mantid/blob/0e7d445f4574ad5ac3266d0a48b3f1b4409ac3c5/Framework/Algorithms/inc/MantidAlgorithms/MaxEnt.h#L46 for an example

Framework/Muon/src/MuonAlgorithmHelper.cpp
```diff
@@ -0,0 +1,241 @@
...
+ */
+MatrixWorkspace_sptr firstPeriod(Workspace_sptr ws) {
+  if (auto group = boost::dynamic_pointer_cast<WorkspaceGroup>(ws)) {
+    return boost::dynamic_pointer_cast<MatrixWorkspace>(group->getItem(0));
+  } else {
+    return boost::dynamic_pointer_cast<MatrixWorkspace>(ws);
+  }
+}
+
+std::string getRunLabel(const Mantid::API::Workspace_sptr &ws) {
```
brandonhewer: Should this be `Workspace_const_sptr`?

Framework/Muon/src/MuonAlgorithmHelper.cpp
```diff
@@ -0,0 +1,241 @@
...
+    zeroPadding = 3;
+  }
+
+  // Begin string output with full label of first run
+  std::ostringstream label;
+  label << instrument;
+  label << std::setw(zeroPadding) << std::setfill('0') << std::right;
+
+  for (auto range : ranges) {
+    std::string firstRun = std::to_string(range.first);
```
brandonhewer: This might make sense as its own function, `std::string createStringFromRange(const std::pair<int, int> &range)`.
This would help simplify this entire code to something similar to:
```
std::ostream_iterator<std::string> addRunIterator(label,", ");
auto runStartIterator = boost::make_transform_iterator(ranges.begin(), createStringFromRange);
auto runEndIterator = boost::make_transform_iterator(ranges.end(), createStringFromRange);
std::copy(runStartIterator, runEndIterator, addRunIterator);
```

Note this specific idea would require including <boost/iterator/transform_iterator.hpp>

brandonhewer (commented):
Just some general comments

Framework/Muon/inc/MantidMuon/ApplyMuonDetectorGrouping.h
```diff
@@ -0,0 +1,119 @@
...
+  void init() override;
+  /// Execution code
+  void exec() override;
+  /// Return the algorithm properties in a struct
+  Muon::AnalysisOptions getUserInput();
+  /// Clip Xmin/Xmax to the range in the input WS
+  void clipXRangeToWorkspace(const API::WorkspaceGroup &ws,
+                             Muon::AnalysisOptions &options);
+  /// Creates and analyses a workspace, if noRebin does not rebin.
+  API::Workspace_sptr createAnalysisWorkspace(API::Workspace_sptr inputWS,
```
AnthonyLim23: you could add a` &` to input workspace

Framework/Muon/inc/MantidMuon/ApplyMuonDetectorGrouping.h
```diff
@@ -0,0 +1,119 @@
...
+  /// Sets algorithm properties according to options.
+  void setMuonProcessAlgorithmProperties(API::IAlgorithm &alg,
+                                         const AnalysisOptions &options) const;
+  /// Set algorithm properties (input workspace, and period properties)
+  /// according to the given options. For use with
+  /// MuonProcess.
+  void
+  setMuonProcessPeriodProperties(API::IAlgorithm &alg,
+                                 API::Workspace_sptr inputWS,
+                                 const Muon::AnalysisOptions &options) const;
```
AnthonyLim23: same as above

Framework/Muon/inc/MantidMuon/ApplyMuonDetectorGrouping.h
```diff
@@ -0,0 +1,119 @@
...
+  void setMuonProcessAlgorithmTimeProperties(
+      IAlgorithm &alg, const Muon::AnalysisOptions &options) const;
+  /// Generate the name of the new workspace
+  const std::string getNewWorkspaceName(const Muon::AnalysisOptions &options,
+                                        const std::string &groupWSName);
+  /// Add the grouped counts to a workspace named "MuonAanalysisGrouped" if its
+  /// in the ADS
+  void addCountsToMuonAnalysisGrouped(API::Workspace_sptr inputWS,
+                                      const Muon::AnalysisOptions options);
+  /// If "MuonAnalysisGrouped" in ADS and a grouped workspace return the
```
AnthonyLim23: these two too

Framework/Muon/src/ApplyMuonDetectorGrouping.cpp
```diff
@@ -0,0 +1,503 @@
...
+
+namespace Mantid {
+namespace Muon {
+
+// Register the algorithm into the AlgorithmFactory
+DECLARE_ALGORITHM(ApplyMuonDetectorGrouping)
+
+void ApplyMuonDetectorGrouping::init() {
+
+  std::string emptyString("");
```
martyngigg: I would just use `""` directly in each space. This actually saves the compiler having to copy this each time.

Framework/Muon/inc/MantidMuon/MuonAlgorithmHelper.h
martyngigg: I don't think this is required here?

Framework/Muon/src/ApplyMuonDetectorGrouping.cpp
```diff
@@ -0,0 +1,503 @@
...
+    return Mantid::Muon::PlotType::Counts;
+  }
+}
+
+/**
+ * Convert the input workspace into a workspace group if e.g. it has only a
+ * single period otherwise leave it alone.
+ */
+Mantid::API::WorkspaceGroup_sptr
+convertInputWStoWSGroup(Mantid::API::Workspace_sptr inputWS) {
```
AnthonyLim23: could add &

Framework/Muon/src/ApplyMuonDetectorGrouping.cpp
```diff
@@ -0,0 +1,503 @@
...
+#include "MantidAPI/AlgorithmManager.h"
+#include "MantidAPI/MatrixWorkspace.h"
+#include "MantidAPI/Workspace.h"
+#include "MantidAPI/WorkspaceGroup.h"
+#include "MantidDataObjects/TableWorkspace.h"
+#include "MantidKernel/EnabledWhenProperty.h"
+#include "MantidKernel/ListValidator.h"
+
+#include <algorithm>
+#include <ctype.h>
```
martyngigg: In C++ we should avoid using the `.h` variants of headers. Is this for `isalnum`? If so I think the correct header is `<cctype>`

Framework/Muon/src/ApplyMuonDetectorGrouping.cpp
```diff
@@ -0,0 +1,503 @@
...
+  Workspace_sptr wsTemp = alg->getProperty("OutputWorkspace");
+  MatrixWorkspace_sptr wsOut =
+      boost::dynamic_pointer_cast<MatrixWorkspace>(wsTemp);
+  AnalysisDataService::Instance().addOrReplace("wsTemp", wsOut);
+
+  std::vector<std::string> names = getMuonAnalysisGroupedWorkspaceNames();
+
+  for (auto &name : names) {
+    IAlgorithm_sptr algAppend =
+        Algorithm::createChildAlgorithm("AppendSpectra");
```
AnthonyLim23: should this be a child algorithm?

Framework/Muon/src/ApplyMuonDetectorGrouping.cpp
```diff
@@ -0,0 +1,503 @@
...
+    // auto ws = AnalysisDataService::Instance().retrieve(name);
```
martyngigg: Can we remove the commented out code, please?

Framework/Muon/src/ApplyMuonDetectorGrouping.cpp
```diff
@@ -0,0 +1,503 @@
...
+    algAppend->setPropertyValue("InputWorkspace1", name);
+    algAppend->setPropertyValue("InputWorkspace2", wsOut->getName());
+    // alg->setPropertyValue("OutputWorkspace", wsAppended->getName());
+
+    algAppend->execute();
+    MatrixWorkspace_sptr wsAppended = algAppend->getProperty("OutputWorkspace");
+
+    if (AnalysisDataService::Instance().doesExist(name)) {
+      IAlgorithm_sptr deleteAlg =
+          AlgorithmManager::Instance().create("DeleteWorkspace");
```
AnthonyLim23: should this be  a child?

Framework/Muon/src/ApplyMuonDetectorGrouping.cpp
```diff
@@ -0,0 +1,503 @@
...
+  auto wsGroup = boost::make_shared<WorkspaceGroup>();
+  AnalysisDataService::Instance().addOrReplace("MuonAnalysisGrouped", wsGroup);
+  MuonAlgorithmHelper::groupWorkspaces("MuonAnalysisGrouped", names);
+}
+
+void ApplyMuonDetectorGrouping::deleteWorkspaceIfExists(
+    const std::string &name) {
+  if (AnalysisDataService::Instance().doesExist(name)) {
+    IAlgorithm_sptr deleteAlg =
+        AlgorithmManager::Instance().create("DeleteWorkspace");
```
AnthonyLim23: child?

Framework/Muon/src/ApplyMuonDetectorGrouping.cpp
```diff
@@ -0,0 +1,503 @@
...
+    deleteAlg->setPropertyValue("Workspace", name);
+    deleteAlg->execute();
+  }
+}
+
+std::vector<std::string>
+ApplyMuonDetectorGrouping::getMuonAnalysisGroupedWorkspaceNames() {
+
+  std::vector<std::string> names;
+  names.clear();
```
martyngigg: `names` will be guaranteed to be empty so `clear` shouldn't be necessary?

Framework/Muon/src/ApplyMuonDetectorGrouping.cpp
```diff
@@ -0,0 +1,503 @@
...
+
+  auto &ads = AnalysisDataService::Instance();
```
martyngigg: :+1:

Framework/Muon/src/MuonAlgorithmHelper.cpp
```diff
@@ -0,0 +1,272 @@
...
+  int zeroPadding;
+  try {
+    zeroPadding = ConfigService::Instance()
+                      .getInstrument(instrument)
+                      .zeroPadding(ranges.begin()->first);
+  } catch (const Mantid::Kernel::Exception::NotFoundError &) {
+    // Old muon instrument without an IDF - default to 3 zeros
+    zeroPadding = 3;
+  }
+
```
AnthonyLim23: could we add a general catch (if it throws something else)

```
catch(...){
 do something sensible.... 
}
```

Framework/Muon/src/ApplyMuonDetectorGrouping.cpp
```diff
@@ -0,0 +1,503 @@
...
+  alg.setProperty("CropWorkspace", false);
+  if (!options.rebinArgs.empty()) {
+    alg.setProperty("RebinParams", options.rebinArgs);
+  }
+  setMuonProcessAlgorithmGroupingProperties(alg, options);
+  setMuonProcessAlgorithmOutputTypeProperty(alg, options);
+  setMuonProcessAlgorithmTimeProperties(alg, options);
+}
+
+/**
```
martyngigg: I would be tempted to move this nearer `init`, maybe underneath, so it's  init then basically validate inputs are closer together.

martyngigg (commented):
Some very minor comments code wise, other than that I'm happy

Framework/Muon/inc/MantidMuon/ApplyMuonDetectorGrouping.h
TWJubb: I'm avoiding using references to pointers

Framework/Muon/test/ApplyMuonDetectorGroupingTest.h
```diff
@@ -0,0 +1,777 @@
...
+    TS_ASSERT_THROWS_NOTHING(alg.setProperty("RebinArgs", emptyString));
+    TS_ASSERT_THROWS_NOTHING(alg.setProperty("TimeOffset", 0.0));
+    TS_ASSERT_THROWS_NOTHING(
+        alg.setProperty("SummedPeriods", std::to_string(1)));
+    TS_ASSERT_THROWS_NOTHING(alg.setProperty("SubtractedPeriods", emptyString));
+    TS_ASSERT_THROWS_NOTHING(alg.execute());
+    TS_ASSERT(alg.isExecuted());
+
+    TS_ASSERT(wsGroup);
+    if (wsGroup) {
```
AnthonyLim23: this is not needed as the above would throw if it were not present

Framework/Muon/test/ApplyMuonDetectorGroupingTest.h
```diff
@@ -0,0 +1,777 @@
...
+    TS_ASSERT_THROWS_NOTHING(alg.execute());
+    TS_ASSERT(alg.isExecuted());
+
+    TS_ASSERT(wsGroup->getItem("inputGroup; Group; test; Counts; #1"));
+    TS_ASSERT(wsGroup->getItem("inputGroup; Group; test; Counts; #1_Raw"));
+
+    auto wsOut = boost::dynamic_pointer_cast<MatrixWorkspace>(
+        wsGroup->getItem("inputGroup; Group; test; Counts; #1_Raw"));
+
+    TS_ASSERT_DELTA(wsOut->readX(0)[0], 0.000, 0.001);
```
AnthonyLim23: in future you could define a variable for the delta

docs/source/algorithms/ApplyMuonDetectorGrouping-v1.rst
```diff
@@ -0,0 +1,58 @@
...
+.. note::
+
+   For examples of applying custom dead times, please refer to :ref:`algm-ApplyDeadTimeCorr`
+   documentation.
+
+   For examples of applying custom grouping, please refer to :ref:`algm-MuonGroupDetectors`
+   documentation.
+
+.. categories::
+
```
AnthonyLim23: could we have some example python code?

docs/source/algorithms/ApplyMuonDetectorGrouping-v1.rst
```diff
@@ -0,0 +1,58 @@
...
+The workspaces which hold the analysed data are stored in the *WorkspaceGroup*, using a unique name. From the following variables
+
+#. The *InputWorkspaceGroup* name is "inputGroup"
+#. *analysisType* is "Counts"
+#. *groupName* is "test"
+
+then the name would be "inputGroup; Group; test; Counts; #1". A second workspace is also created with the same name, but ending in "#1_Raw", the only difference being that the *RebinArgs* are not applied to the latter workspace.
+
+An option exists to apply dead time correction to the counts if *ApplyDeadTimeCorrection* is true, requiring a *DeadTimeTable* with the format as in :ref:`algm-DeadTimeCorr`.
```
AnthonyLim23: http://docs.mantidproject.org/nightly/algorithms/CalMuonDeadTime-v1.html

Framework/Muon/test/ApplyMuonDetectorGroupingTest.h
TWJubb: In this case that's quite difficult as I'm using a delta which is 3 sig figs compared to the expected value (which, depending on the axis, varies by several orders of magnitude). So it was just quicker to hard code them.

Framework/Muon/src/ApplyMuonDetectorGrouping.cpp
TWJubb: For some reason I get a link error (unresolved external symbol) when using "" directly

martyngigg: Okay - don't worry then. It's not worth spending too much time on it.

AnthonyLim23:
@martyngigg  I am happy that it does what its supposed to

martyngigg:
@AnthonyLim23 Thanks. Can you mark the review as approved?

martyngigg:
@TWJubb The only thing left is to fix the documentation tests and then this is good to go.

AnthonyLim23 (approved):
looks good and works as expected

martyngigg (approved):

PR #22459, opened 2018-05-22
Frequency domain analysis clean up

### LINKED ISSUE #22399: Refactor FDA to match new layout
Make the FDA match the new layout (model, view, presenter and widget classes) of the new architecture for muon analysis 2

AnthonyLim23:
**Description of work**
This is a clean up of the frequency domain analysis GUI to match the new structure. A few other minor improvements have also been made. 

To have muon analysis 2 as an interface you will need add the following to https://github.com/mantidproject/mantid/blob/master/Framework/Properties/Mantid.properties.template#L23    `Muon/Muon_Analysis_2.py`

**To test:**
Open muon analysis 2 and check it exists. Then you can close it

Open Muon Analysis (v1) and load some data
Close Muon Analysis
Open Frequency Domain analysis and run an FFT 
A workspace will be added to the workspace group, check it has a history
Change the combo box at the top to maxent
Run maxent and check it produces an output

The tables in both maxent and FFT mode should now also be rescalable 

Fixes #22399. 

**Release Notes** 

---
Does not need to be in release notes
#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


scripts/Muon/GUI/FrequencyDomainAnalysis/FFT/fft_model.py
```diff
@@ -0,0 +1,155 @@
...
+            self.FFT = inputs["FFT"]
+        else:
+            self.FFT = None
+        self.model.setRun(inputs["Run"])
+
+    def execute(self):
+        """
+        runs the relevant parts of the FFT and the preprocessing
+        """
+        if self.phaseTable is not None:
```
TWJubb: Is this just the same as `if self.phaseTable`?

scripts/Muon/GUI/FrequencyDomainAnalysis/FFT/fft_model.py
```diff
@@ -0,0 +1,155 @@
...
+    def __init__(self):
+        self.name = "FFT"
+        self.alg = None
+
+    def cancel(self):
+        if self.alg is not None:
+            self.alg.cancel()
+
+    def setRun(self, run):
+        self.runName = run
```
TWJubb: Just a general question; is it normal to have member variables of the class defined within member functions? It makes it hard to understand what information the object carries around. There are also many occasions where a member variable is set to `None` if not in a list of inputs; would it be easier to just declare them all `None` in the `__init__`?

scripts/Muon/GUI/FrequencyDomainAnalysis/FFT/fft_model.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def preAlg(self, preInputs):
```
TWJubb: This name confused me... is it implying that this algorithm must be performed before FFT?

scripts/Muon/GUI/FrequencyDomainAnalysis/FFT/fft_model.py
```diff
@@ -0,0 +1,155 @@
...
+        self.alg.setAlwaysStoreInADS(False)
+        for name, value in iteritems(FFTInputs):
+            self.alg.setProperty(name, value)
+        self.alg.execute()
+        mantid.AnalysisDataService.addOrReplace(
+            FFTInputs["OutputWorkspace"],
+            self.alg.getProperty("OutputWorkspace").value)
+
+        ws = self.alg.getPropertyValue("OutputWorkspace")
+        group = mantid.AnalysisDataService.retrieve(self.runName)
```
TWJubb: Is it worth checking that the workspace exists when querying the ADS, just in case something strange happened or the naming changed from the GUI?

scripts/Muon/GUI/FrequencyDomainAnalysis/FFT/fft_presenter.py
```diff
@@ -3,25 +3,35 @@
...
     This class links the FFT model to the GUI
     """
-    def __init__(self,view,alg,load):
-        self.view=view
-        self.alg=alg
-        self.load=load
+
+    def __init__(self, view, alg, load):
+        self.view = view
+        self.alg = alg
```
TWJubb: Why `alg` and not `FFTmodel` or something similar?

scripts/Muon/GUI/FrequencyDomainAnalysis/FFT/fft_presenter.py
```diff
@@ -31,21 +41,25 @@ def deactivate(self):
         self.view.deactivateButton()
 
     def getWorkspaceNames(self):
-        final_options=self.load.getWorkspaceNames()
+        final_options = self.load.getWorkspaceNames()
         self.view.addItems(final_options)
 
-    #functions
+    # functions
     def phaseCheck(self):
```
TWJubb: This function is doing more than just checking isn't it, by calling `setPhaseBox`. So is there a more descriptive name.

scripts/Muon/GUI/FrequencyDomainAnalysis/MaxEnt/maxent_widget.py
```diff
@@ -0,0 +1,29 @@
...
+        model = MaxEntWrapper(maxEnt)
+        self.pres = MaxEntPresenter(view, model, load)
+
+    @property
+    def presenter(self):
+        return self.pres
+
+    @property
+    def widget(self):
+        return self.presenter.widget
```
TWJubb: Should this be self.pres.widget?

scripts/Muon/GUI/FrequencyDomainAnalysis/MaxEnt/maxent_presenter.py
```diff
@@ -23,6 +23,10 @@ def __init__(self, view, alg, load):
         self.view.cancelSignal.connect(self.cancel)
         self.view.phaseSignal.connect(self.handlePhase)
 
+    @property
+    def widget(self):
+        return self.view
+
     # functions
     def getWorkspaceNames(self):
```
TWJubb: This functions seems to do more than getting workspace names.

scripts/Muon/GUI/FrequencyDomainAnalysis/FFT/fft_model.py
AnthonyLim23: the check is done in the view

scripts/Muon/GUI/FrequencyDomainAnalysis/FFT/fft_presenter.py
AnthonyLim23: its actually a thread model

scripts/Muon/GUI/FrequencyDomainAnalysis/FFT/fft_presenter.py
AnthonyLim23: if its not there then we are not allowing users to decide not to produce one

scripts/Muon/GUI/FrequencyDomainAnalysis/MaxEnt/maxent_widget.py
AnthonyLim23: the others should be presenter

scripts/Muon/GUI/FrequencyDomainAnalysis/MaxEnt/maxent_presenter.py
AnthonyLim23: it makes sure that the user does not use less data points then are available, since this is view we update it with adding the workspaces to the view

scripts/Muon/GUI/FrequencyDomainAnalysis/TransformSelection/transform_selection_widget.py
```diff
@@ -0,0 +1,26 @@
...
+        view = self.widget
+        view.changeMethodSignal.connect(slot)
+
+    @property
+    def presenter(self):
+        return self.pres
+
+    @property
+    def widget(self):
+        return self.presenter.widget
```
TWJubb: Should this also be `self.pres`?

AnthonyLim23:
they have splitters

scripts/Muon/GUI/FrequencyDomainAnalysis/MaxEnt/maxent_widget.py
AnthonyLim23: apparently `presenter` breaks it. So changing them all to `pres`

scripts/Muon/GUI/FrequencyDomainAnalysis/FFT/fft_model.py
AnthonyLim23: yes

scripts/Muon/GUI/FrequencyDomainAnalysis/FFT/fft_model.py
AnthonyLim23: some OS can be fussy, so its safer

scripts/Muon/GUI/FrequencyDomainAnalysis/FFT/fft_model.py
AnthonyLim23: This is important for how it works later on

scripts/Muon/GUI/FrequencyDomainAnalysis/FFT/fft_model.py
martyngigg: This style is actually recommended by the Python PEP-8 style guide: https://www.python.org/dev/peps/pep-0008/#programming-recommendations for comparisons with singleton objects. I don't think it should be OS specific...

scripts/Muon/GUI/FrequencyDomainAnalysis/FFT/fft_model.py
martyngigg: I tend to agree with Tom here. I think it's clearer to define all of the attributes used by the class, even if they are set to None, in the class `__init__` so you can see what data the class deals with.

scripts/Muon/GUI/FrequencyDomainAnalysis/FFT/fft_model.py
martyngigg: Is the possiblity that `retrieve` could raise an exception if the workspace had disappeared, by being deleted/renamed by another task, taken into account somewhere?

scripts/Muon/GUI/FrequencyDomainAnalysis/MaxEnt/maxent_widget.py
martyngigg: Why does `self.presenter` break it? If we don't understand that then are we not in danger of some code behaving in a way we don't understand?

scripts/Muon/GUI/Common/dummy/dummy_widget.py
```diff
@@ -1,8 +1,8 @@
 from __future__ import (absolute_import, division, print_function)
 
 
-from Muon.GUI.dummy.dummy_view import DummyView
-from Muon.GUI.dummy.dummy_presenter import DummyPresenter
+from Muon.GUI.Common.dummy.dummy_view import DummyView
+from Muon.GUI.Common.dummy.dummy_presenter import DummyPresenter
 
 
 class DummyWidget(object):
```
martyngigg: GitHub won't let me comment inline below here but there is a typo on line 17 in `DummyWidget.` It reads `presneter` and not `presenter`

martyngigg:
I've just noticed that this branch has `googletest-src` and `python-xmlrunner-src` entries in the root of the repository - can we remove those I think they have been committed by accident?

scripts/Muon/GUI/FrequencyDomainAnalysis/FFT/fft_model.py
AnthonyLim23: if any exceptions are created it is caught in the thread model, we could add a check if its not none

martyngigg: :+1:

scripts/Muon/GUI/FrequencyDomainAnalysis/FFT/fft_model.py
AnthonyLim23: an easy way to do it without duplicating code would be to just call `loadData` at the end of init with an empty dictionary

scripts/Muon/GUI/FrequencyDomainAnalysis/FFT/fft_model.py
AnthonyLim23: I made a mistake in the presenter for FFT, as it did not set up the thread model correctly. Fixing that means if a workspaces is missing a nice pop up appears with an error message (no other code changes needed)

martyngigg (approved):
I'm happy for these changes to go in as they are and for the FFT GUI to still be iteratively improved as we go.

PR #22469, opened 2018-05-24
Update to EstimateMuonAsymmetryFromCounts

### LINKED ISSUE #22464: Mistake in normalisation estimate
In issue #19702 we updated the method for calculating the normalisation constant. However, there was a mistake in the maths and the denominator is missing a factor of 

- exp(-t_0/tau) - exp(-t_N/tau)

### LINKED ISSUE #22465: EstimateMuonAsymmetryFromCounts : First bin in data not included in normalization.
### Expected behavior

In ``EstimateMuonAsymmetryFromCounts`` algorithm, I feed in a histogram with bin edges (0.0,0.1,0.2,...). The calculation of the asymmetry requires a summation of the counts from all of the bins between a start and end time. If I set my start time to be equal to the lowest bin edge (i.e. t_0 = 0.0), then I expect that the counts from the first bin (0.0-0.1) are included in the summation. Currently that is not the case

### Actual behavior

https://github.com/mantidproject/mantid/blob/master/Framework/Algorithms/src/MuonAsymmetryHelper.cpp#L87

If the first bin edge of dataX is equal to startX, the summation will not include the first bin. This is caused by the ```startIndexFromTime``` function

https://github.com/mantidproject/mantid/blob/master/Framework/Algorithms/src/MuonAsymmetryHelper.cpp#L106

which uses ```std::upper_bound```, which uses a 'greater than' rather than 'greater than or equal to' comparison.

### LINKED ISSUE #22471: EstimateMuonAsymmetry
Update EstimateMuonAsymmetryFromCounts to include updating/creating a table containing the norm and to have an option to output the un-normalized results.

AnthonyLim23:
The calculation of the muon asymmetry is different between the doc and code. This difference is from replacing the denominator with an integral, then estimating it with the trapezium rule. 

If the `startX` is the same as a bin edge, then the bin is missing (which is wrong) from the calculation.

General update to the input/output and integration into Muon analysis

The code was missing a subtraction for the two edge cases. 

Note that the changes to muon process test are because it uses this algorithm to produce its result. These changes have also impacted CaclulcateMuonAsymmetry, but in issue #22490 the algorithm is getting a big rewrite.

**To test:**
Open Muon Analysis and load `musr 62260`
The `MuonAnalysisTFNormalizations` will contain the normalization value and the workspace name
Go to the data analysis tab
Add a flat background and a GaussOsc function
Change the frequency value to 1.3
Do a fit
The flat background should be small but non zero

Remove the flat background
Do a TF Asymmetry fit
The `MuonAnalysisTFNormalizations` table should be updated

Go to the settings tab and turn on Multiple fitting mode
Go to data analysis
Check the table now contains the information for all 4 groups

Run the below script and you will get the same result for all 3 outputs -> the bin edges are included correctly.
```
import math
import numpy as np

tab.addColumn('double', 'norm')
tab.addColumn('str', 'name')
tab.addColumn('str', 'method')

tab.addRow([1.,"Run;;Group;;a;;Asym;;#1","Estimate"])

y = [100, 150, 50, 10, 5]
x = [1,2,3,4,5,6]
input = CreateWorkspace(x,y)
run = input.getRun()
run.addProperty("goodfrm","10","None",True)
output=EstimateMuonAsymmetryFromCounts(InputWorkspace=input,NormalizationTable=tab,spectra=0,StartX=1,EndX=5)
print("Asymmetry:  {}".format(['{0:.2f}'.format(value) for value in output.readY(0)]))

output=EstimateMuonAsymmetryFromCounts(InputWorkspace=input,NormalisationTable=tab,spectra=0,StartX=.9,EndX=5.)
print("Asymmetry:  {}".format(['{0:.2f}'.format(value) for value in output.readY(0)]))

output=EstimateMuonAsymmetryFromCounts(InputWorkspace=input,NormalisationTable=tab,spectra=0,StartX=1,EndX=5.1)
print("Asymmetry:  {}".format(['{0:.2f}'.format(value) for value in output.readY(0)]))
```

Fixes #22464. 
Fixes #22465 
Fixes #22471
**Release Notes** 

In release notes
---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


AnthonyLim23:
checks to run include:

Open group, period
single fit show correct norm
multi fit show correct norm
does having binning and then selecting raw give correct norm
does load all ... work

AnthonyLim23:
Got to update release notes and check tests. Need to add example and test for table norms and unnorm workspace

Framework/Algorithms/src/MuonAsymmetryHelper.cpp
```diff
@@ -106,12 +117,12 @@ double estimateNormalisationConst(const HistogramData::Histogram &histogram,
...
   auto upper =
-      std::upper_bound(xData.rawData().begin(), xData.rawData().end(), startX);
-  return std::distance(xData.rawData().begin(), upper);
+      std::lower_bound(xData.rawData().begin(), xData.rawData().end(), startX);
+  return std::distance(xData.rawData().begin(), upper+1);
 }
 /**
 * find the first index in bin edges that is after
-* the endtime.
+* the endtim e.
```
AnthonyLim23: need to fix typo

MatthewBowles (changes requested):
The MuonAnalysisTFNormalizations table did not change for me after the fit, and running the script gives an error:
```
RuntimeError: Unknown property search object NormalisationTable
  at line 16 in 'New script'
```

AnthonyLim23:
ah, thats because I forgot to update the example in this PR. swap the s for a z.

Framework/Algorithms/src/EstimateMuonAsymmetryFromCounts.cpp
```diff
@@ -132,6 +156,23 @@ void EstimateMuonAsymmetryFromCounts::exec() {
   // Do the specified spectra only
   int specLength = static_cast<int>(spectra.size());
   std::vector<double> norm(specLength, 0.0);
+
+  double normConst = getProperty("NormalizationIn");
+  for (int j = 0; j < specLength; j++) {
+    if (normConst == 0) {
```
MatthewBowles: This can be changed so the if statements are not inside the loop, it's better not to do the check every time.

qt/scientific_interfaces/Muon/MuonAnalysis.cpp
```diff
@@ -449,20 +449,28 @@ std::string MuonAnalysis::addItem(ItemType itemType, int tableRow,
...
+  // Create workspace and a raw (unbinned) version of it
+  auto ws = createAnalysisWorkspace(itemType, tableRow, plotType, wsName);
+  if (ads.doesExist("tmp_unNorm")) {
+    ads.rename("tmp_unNorm", wsName + unnorm);
+    wsNames.push_back(wsName + unnorm);
+  }
 
+  auto wsRaw =
+      createAnalysisWorkspace(itemType, tableRow, plotType, wsRawName, true);
+  if (ads.doesExist("tmp_unNorm")) {
```
MatthewBowles: Not sure I understand this - do we need to do this check twice? If so, put it in a function.

MatthewBowles (changes requested):
Just a couple of things:

Framework/Algorithms/src/EstimateMuonAsymmetryFromCounts.cpp
MatthewBowles: I'd suggest either store the bool beforehand or use two loops, one for the `if` and one for the `else`.

qt/scientific_interfaces/Muon/MuonAnalysis.cpp
AnthonyLim23: its because tmp_unNorm is produced by createAnalysisWorkspace (the first one is binned and the second is raw/unbinned).

MatthewBowles: If we know it's being produced then why do we need the test, can it fail?

AnthonyLim23: yes, it only passes in the case of asymmetry and groups

MatthewBowles: Ok. I'd function it anyway, just so we don't have three lines of code repeated right after each other.

MatthewBowles (approved):

PR #22496, opened 2018-05-30
New algorithm: ReflectometrySumInQ

### LINKED ISSUE #22315: Coherent or SumInQ mode for ReflectometryILLSumForeground
ILL's reflectometry reduction workflow is missing the coherent summation method, a.k.a SumInQ mode. This should be added to the `ReflectometryILLSumForeground` algorithm. An implementation exists in `ReflectometryReductionOne` which could be extracted into a separate algorithm and shared between the reductions.

soininen:
**Description of work**

This PR introduces a new algorithm, `ReflectometrySumInQ`. The algorithm maps the intensities from the input workspace onto a single reference 2theta by preserving the momentum transfer of each bin. The implementation follows closely the *SumInQ* method in `ReflectometryReductionOne` the main differences being that `ReflectometrySumInQ` produces only a single summed histogram as the output and that the 2theta size of the detectors is taken from the averages between the 2thetas of neighboring detectors instead of from their bounding boxes.

As part of this PR, some duplicated code in `ReflectometryReductionOne` and `GroupDetectors` was combined into `Kernel::String` as it was initially thought that the code would also be needed by `ReflectometrySumInQ`.

**To test:**

With facility set to ILL and unit test data included in the data search directories, the following script could provide a starting point for testing:
```python
ws = Load('ILL/D17/317370')
sum = ReflectometrySumInQ(
    InputWorkspace=ws,
    InputWorkspaceIndexSet=[200, 201, 202, 203, 204, 205],
    BeamCentre=202,
    WavelengthMin=0.1,
    WavelengthMax=14.)
```

Partially implements #22315.

**Release Notes** 

The reflectometry release notes were updated accordingly.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Algorithms/src/ReflectometrySumInQ.cpp
```diff
@@ -0,0 +1,552 @@
...
+  const HistogramData::Histogram modelHistogram(std::move(bins),
+                                                std::move(counts));
+  // Create the output workspace
+  API::MatrixWorkspace_sptr outputWS =
+      DataObjects::create<DataObjects::Workspace2D>(detectorWS, 1,
+                                                    std::move(modelHistogram));
+
+  // Set the detector ID from the twoThetaR detector.
+  auto &outSpec = outputWS->getSpectrum(0);
+  // TODO: Handle grouped detectors correctly.
```
gemmaguest: Just to check: was this `TODO` left in intentionally?

Framework/Algorithms/src/ReflectometrySumInQ.cpp
```diff
@@ -0,0 +1,552 @@
...
+  // For bLambda, use the average bin size for this spectrum
+  const auto minThetaEdges = detectorWS.binEdges(twoThetaMin.first);
+  double bLambda =
+      (minThetaEdges[minThetaEdges.size() - 1] - minThetaEdges[0]) /
+      static_cast<int>(minThetaEdges.size());
+  MinMax lambdaRange(lambdaMax - bLambda / 2., lambdaMax + bLambda / 2.);
+  auto r = projectedLambdaRange(lambdaRange, twoThetaMinRange, refAngles);
+  wavelengthRange.max = r.max;
+  const auto twoThetaMaxRange = twoThetaWidth(twoThetaMax.first, spectrumInfo);
+  const auto maxThetaEdges = detectorWS.x(twoThetaMax.first);
```
gemmaguest: Should this use `binEdges()` rather than `x()`, as above?

gemmaguest:
This looks good to me :+1: Just a couple of in-line queries above.
- The results for the ILL example look sensible.
- It produces very similar results to `ReflectometryReductionOne` for the usage example.
- Handles invalid inputs well

Framework/Algorithms/src/ReflectometrySumInQ.cpp
soininen: It was left in out of laziness. But now I think I've found an elegant way of getting rid of it.

gemmaguest (approved):
This all looks good now :+1:

SimonHeybrock:
There is a conflict now.

PR #22498, opened 2018-05-30
Workbench explicit overplot

martyngigg:
**Description of work**

Feedback from the January 2018 user meeting gave universal disapproval for the active/hold facility of plot management. Users wanted to decide at the point of plotting the workspace whether it should open in a new figure or overplot on the existing one.

This work removes the active/hold functionality and adds separate items to overplot on to existing figures if they exist.

**To test:**

Start the workbench and either load some data or create some fake data with `CreateSampleWorkspace`. Examine the plot menu of the workspace and check you are happy with the layout and text.

Try all of the options. The behaviour is intended to be as follows:
* plot will always open a new figure
* overplot will add to an existing active figure if one exists and create a new one if one doesn't exist
  * if two or more plot windows are open then clicking on the window makes it active and the next overplot request will go there.

*No associated issue*

**Release Notes** 

*Does not need to be in the release notes. The workbench will be advertised separately.*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/qt/applications/workbench/workbench/plotting/functions.py
+++ b/qt/applications/workbench/workbench/plotting/functions.py
@@ -68,6 +68,17 @@ def _validate_pcolormesh_inputs(workspaces):
         raise_if_not_sequence(workspaces, 'Workspaces')
 
 
+def current_figure_or_none():
+    """If an active figure exists then return it otherwise return None
+
+    :return: An active figure or None
+    """
+    if len(plt.get_fignums()) > 0:
+        return plt.gcf()
+    else:
+        return None
+
+
 def plot(workspaces, spectrum_nums=None, wksp_indices=None, errors=False,
          overplot=False):
     """
--- a/qt/applications/workbench/workbench/plotting/functions.py
+++ b/qt/applications/workbench/workbench/plotting/functions.py
@@ -68,7 +68,8 @@ def _validate_pcolormesh_inputs(workspaces):
         raise_if_not_sequence(workspaces, 'Workspaces')
 
 
-def plot(workspaces, spectrum_nums=None, wksp_indices=None, errors=False):
+def plot(workspaces, spectrum_nums=None, wksp_indices=None, errors=False,
+         overplot=False):
     """
     Create a figure with a single subplot and for each workspace/index add a
     line plot to the new axes. show() is called before returning the figure instance. A legend
@@ -78,6 +79,7 @@ def plot(workspaces, spectrum_nums=None, wksp_indices=None, errors=False):
     :param spectrum_nums: A list of spectrum number identifiers (general start from 1)
     :param wksp_indices: A list of workspace indexes (starts from 0)
     :param errors: If true then error bars are added for each plot
+    :param overplot: If true then overplot over the current figure if one exists
     :returns: The figure containing the plots
     """
     # check inputs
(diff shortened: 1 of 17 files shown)


qt/applications/workbench/workbench/plotting/globalfiguremanager.py
```diff
@@ -63,8 +70,14 @@ def destroy(cls, num):
         manager = cls.figs[num]
         manager.canvas.mpl_disconnect(manager._cidgcf)
 
-        if cls._active.num == num:
-            cls._active = None
+        # There must be a good reason for the following careful
+        # rebuilding of the activeQue; what is it?
```
ianbush: I think this comment is from `_pylab_helpers`, but if I understand this correctly it means we can do a sort by last active in the plot selector.

ianbush (changes requested):
This looks good, I do prefer this over the previous active/hold buttons.

One issue - you can overplot on a colorfill plot. This gives an interesting but not very useful plot :smile:. If possible it would be nice to disable the menu options if the last plot was a colorfill.

At some point it would be good to get this code under test, but that is for another PR.

qt/applications/workbench/workbench/plotting/functions.py
```diff
@@ -86,17 +88,25 @@ def plot(workspaces, spectrum_nums=None, wksp_indices=None, errors=False):
...
     plot_fn = ax.errorbar if errors else ax.plot
     for ws in workspaces:
         for num in nums:
             plot_fn(ws, **{kw: num})
 
     ax.legend()
-    ax.set_title(workspaces[0].name())
+    title = workspaces[0].name()
+    ax.set_title(title)
+    fig.canvas.set_window_title(title)
```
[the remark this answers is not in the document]
martyngigg: Good point. Matplotlib stores the figure number so we could just use that and append it? The manager requires it to be unique I think...

qt/applications/workbench/workbench/plotting/globalfiguremanager.py
martyngigg: I think you are correct and this will probably be welcome.

martyngigg:
I've added a note to the project board to add tests for the global figure manager and plotting functions. Thanks for reminding me as it got lost in the dash to prepare for the user meeting.

martyngigg:
Thanks for picking up the colorfill oddity. It turns out that MantidPlot did this by accident and I replicated it but we agree that it's not useful and if you drag on to an existing 2D plot it should replace it. Would you agree?

martyngigg:
@ianbush The current code structure of the workspace widget is not playing nicely with this. Our valiant attempts at sharing code mean that the menu generation is really quite disconnected from the rest of the python code controlling the plotting. I have opened #22504 to address this. How would you feel about a simple pop dialog that says you can't overplot a 1D plot on a colorfill plot and asks you to select a new one for now?

qt/applications/workbench/workbench/plugins/workspacewidget.py
```diff
@@ -106,8 +107,14 @@ def _do_plot_spectrum(self, names, errors, overplot):
         :param names: A list of workspace names
         :param errors: If true then error bars will be plotted on the points
         :param overplot: If true then the add to the current figure if one
-                         exists
+                         exists and it is a compatible figure
         """
+        if overplot:
```
ianbush: :+1:

ianbush (approved):
Changes look good now. Plot naming can be improved in #22516 and menu behaviour after #22504.

PR #22503, opened 2018-05-30
Fix right-click segfault in new workbench

peterfpeterson:
**Description of work**

The error appears to be solved in newer versions of sip / python-qt5,
but that isn't what is available on rhel7. This removes the ability to
have a different `MantidDisplayBase` from the workbench, but it doesn't
appear to be a necessary abstraction for now.

**To test:**

Build the new workbench on RHEL7 and try to right-click plot a workspace. If you can, then this did it.

*Does not need to be in the release notes* because it fixes a bug in the new workbench that nobody can use yet.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


AndreiSavici (approved):
I can right click on workspaces. Running SumSpectra causes segmentation fault (only on RHEL7). Works on Ubuntu 18.04

PR #22525, opened 2018-06-04
Fix path finding for cxxtestgen

samueljackson92:
**Description of work.**

This prevents cmake from finding a cxxtestgen installed on the system
path rather than the one shipped with Mantid. For example, I had cxxtest installed via homebrew and cmake found the includes for that before the mantid one.

**To test:**

 - Configuring a fresh build with mantid should work as before.
 - They only known way to reproduce this is to install cxxtest somewhere on your path and then run a fresh cmake. For example I installed cxxtest via homebrew which first showed this issue. 

*No associated issue*

- [ ] Yes
- [x] No

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


martyngigg (approved):
Changes are good. We only want to find our internal copy of cxxtest.

PR #22448, opened 2018-05-18
Enable FitPeaks to "observe" more profiles and export effective parameters

wdzhou:
Enable FitPeaks to observe peak width for more peak profile types at least including Lorentzian, PseudoVoigt and Voigt.  Besides, FitPeaks shall be able to export effective peak parameters alternative to raw peak parameters.

**To test:**

1. Check the new added unit test
2. Make sure all the unit tests and system tests are passed
3. Use FitPeaks to fit some Voigt/PseudoVoigt/Lorentzian peaks
4. Check the output with effective peak parameters

Fixes #22435.

*Does not need to be in the release notes, because the newly added features are standard features of FindPeaks, which FitPeaks is designed to replace.*

---

#### Reviewer ####

Please comment on the following ([full description](http://www.mantidproject.org/Individual_Ticket_Testing)):

##### Code Review #####

- [ ] Is the code of an acceptable quality?
- [ ] Does the code conform to the [coding standards](http://www.mantidproject.org/Coding_Standards)?
- [ ] Are the unit tests small and test the class in isolation?
- [ ] If there are changes in the release notes then do they describe the changes appropriately?

##### Functional Tests #####

- [ ] Do changes function as described? Add comments below that describe the tests performed?
- [ ] Do the changes handle unexpected situations, e.g. bad input?
- [ ] Has the relevant (user and developer) documentation been added/updated?

Does everything look good? Mark the review as **Approve**. A member of `@mantidproject/gatekeepers` will take care of it.

PR #22532, opened 2018-06-04
PDCalibration peak shape

### LINKED ISSUE #22520: PDCalibration issues
I'm trying to use PDCalibration but run into many small issues that make it hard to use, mainly documentation.

* `InputWorkspace` property description refers to `SignalFile` that no longer exist.
* `PeakWidthPercent` property description says
    > The estimated peak width as a percentage of the d-spacing of the center of the peak.
    
    But I think this is actually percentage of the `tof` of the center of the peak?
* The `PositionTolerance` property does nothing. Also the descriptions end with
    > If there is only one value given, then 

   Then? Then what‽
* The `ConstrainPeakPositions` property does nothing. I think it should set the `ConstrainPeakPositions` option of the call to `FitPeaks`?
* Since now it uses `FitPeaks` instead of `FindPeaks` the documentation should be updated to say this.

peterfpeterson:
Modify `PDCalibration` to correctly allow for multiple peak shape functions. 

*Does not need to be reported to anybody*

**To test:**

Look over the code and documentation. Everything should still work as expected.

Fixes #22520 as well.

~~Do not merge until #22448 is.~~

*This is already covered in a previous set of release notes.*

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/Algorithms/src/FitPeaks.cpp
+++ b/Framework/Algorithms/src/FitPeaks.cpp
@@ -1155,11 +1155,7 @@ void FitPeaks::calculateFittedPeaks(std::vector<
 
     for (size_t ipeak = 0; ipeak < m_numPeaksToFit; ++ipeak) {
       // get and set the peak function parameters
-      size_t row_index =
-          (static_cast<size_t>(iws) - m_startWorkspaceIndex) * m_numPeaksToFit +
-          ipeak;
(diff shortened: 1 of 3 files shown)


rosswhitfield (approved):
Gaussian and Lorentzian peak functions work well for me. The other options not at all but they may work in other cases.

PR #22538, opened 2018-06-05
Fix order of merging in MergeRuns

### LINKED ISSUE #22402: LoadAndMerge: strange behavior when trying to merge IN4 workspaces with different wavelengths
The issue arises when changing the order of the input runs in the following script (have to be at ILL):
```python
mantid.config.appendDataSearchDir(
   '/net4/serdon/illdata/processed/181/in6/exp_4-04-492/rawdata/')

merged = LoadAndMerge(Filename='197163+197164+197165', LoaderName='LoadILLTOF')
```

Run 197164 is incompatible with the other runs due to `monitor1.time` sample log being different.

`monitor1.time` for the above runs:

197163: 3.5

197164: 5.4

197165: 3.5

Regardless of the order of the runs, the output workspace has `monitor1.time` 5.4. However, the algorithm prints an error about not being able to merge once or twice, depending on the order of runs.

The expected behavior would be to pick the first run as the 'master' run against which the rest of the runs are compared.

soininen:
**Description of work.**

This PR removes the confusing sorting of input workspaces according to their X values from `MergeRuns`. Now the workspaces are handled in the order they are listed in the *InputWorkspace* property, as one would expect.

**Report to:** Björn Fåk.

**To test:**

Check that the `merged` workspace in the following has `ws_number == 2.` in the sample logs:

```python
Xs = [0., 1.]
Ys = [0.]
ws1 = CreateWorkspace(DataX=Xs, DataY=Ys)
ws1.mutableRun().addProperty('ws_number', 1., True)
Xs = [1., 2.]
ws2 = CreateWorkspace(DataX=Xs, DataY=Ys)
ws2.mutableRun().addProperty('ws_number', 2., True)
merged = MergeRuns(['ws2', 'ws1'])
```

Fixes #22402.

- [x] Yes
- [ ] No

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Algorithms/inc/MantidAlgorithms/MergeRuns.h
```diff
@@ -128,18 +128,20 @@ class DLLExport MergeRuns : public API::MultiPeriodGroupAlgorithm {
...
-  void noOverlapParams(const HistogramData::HistogramX &X1,
-                       const HistogramData::HistogramX &X2,
-                       std::vector<double> &params) const;
-  void intersectionParams(const HistogramData::HistogramX &X1, int64_t &i,
-                          const HistogramData::HistogramX &X2,
-                          std::vector<double> &params) const;
-  void inclusionParams(const HistogramData::HistogramX &X1, int64_t &i,
-                       const HistogramData::HistogramX &X2,
-                       std::vector<double> &params) const;
+  std::pair<bool, std::vector<double>> checkRebinning();
```
josephframsay: Probably just a matter of taste, but calls to `checkRebinning` might look a bit nicer if you returned a `boost::optional<std::vector<double>>` here, which would avoid the annoying `std::tie` at the other end - you could rename the method something like `getRebinParamsIfNeeded` to make it clear what's going on (plus the return value could be `const`)

Framework/Algorithms/src/MergeRuns.cpp
```diff
@@ -476,47 +457,91 @@ bool MergeRuns::validateInputsForEventWorkspaces(
...
 
+/** Checks if the workspaces need to be rebinned and if so, returns the
+ *  rebinning parameters for the Rebin algorithm.
+ *  @return :: A bool signifying if rebinning is needed and the parameters as a
+ *pair.
+ */
+std::pair<bool, std::vector<double>> MergeRuns::checkRebinning() {
+  const std::string rebinBehaviour = getProperty("RebinBehaviour");
+  const std::string sampleLogsFailBehaviour = getProperty("FailBehaviour");
+  std::list<MatrixWorkspace_sptr> inputsSortedByX{m_inMatrixWS};
```
josephframsay: Is there any reason that this couldn't be a `std::vector` instead?

Framework/Algorithms/src/MergeRuns.cpp
```diff
@@ -558,34 +583,36 @@ void MergeRuns::noOverlapParams(const HistogramX &X1, const HistogramX &X2,
...
 void MergeRuns::intersectionParams(const HistogramX &X1, int64_t &i,
                                    const HistogramX &X2,
-                                   std::vector<double> &params) const {
+                                   std::vector<double> &params) {
   // First calculate the number of bins in each workspace that are in the
   // overlap region
-  int64_t overlapbins1, overlapbins2;
-  overlapbins1 = X1.size() - i;
-  for (overlapbins2 = 0; X2[overlapbins2] < X1.back(); ++overlapbins2) {
+  int64_t const overlapbins1 = X1.size() - i;
```
josephframsay: Could this just be `auto`?

Framework/Algorithms/src/MergeRuns.cpp
```diff
@@ -558,34 +583,36 @@ void MergeRuns::noOverlapParams(const HistogramX &X1, const HistogramX &X2,
...
+  const auto iterX2 = std::lower_bound(X2.cbegin(), X2.cend(), X1.back());
+  if (iterX2 == X2.end()) {
+    throw std::runtime_error("MergerRuns::intersectionParams: no intersection "
+                             "between the histograms.");
   }
-
+  int64_t const overlapbins2 = std::distance(X2.cbegin(), iterX2);
```
josephframsay: Similarly here `auto`?

josephframsay (changes requested):
These changes are nice - improvements to code quality and documentation in particular I'm a fan of. The changes to the PEARL output are fairly negligible so I'm happy to let that through, however (and I know this is a massive pain) I'd rather the reference files were updated, as opposed to the tolerance of the tests increased. If you're not able to run the PEARL reduction locally I can send you the focused output, or even just do the update myself if that's easier?

Apart from that and a few really nit-picky points, everything looks good to me!

Framework/Algorithms/inc/MantidAlgorithms/MergeRuns.h
soininen: Forgot completely about `boost::optional`! I was trying to use `std::optional` here but it's C++17.

soininen:
@josephframsay Updating the system test reference files is the way to go also in my opinion. Changing the tolerances was an act of desperation on my part. I would highly appreciate it if you could update the reference files straight here in this PR. I think I could run the reduction here at ILL, but I would be more confident if the references were produced by someone who knows what they are doing.

josephframsay (approved):

soininen:
Thanks a lot for updating the system test references!

josephframsay:
@soininen no problem!

PR #22558, opened 2018-06-08
Add adjustments to reconstructed data in Maxent

### LINKED ISSUE #21577: MaxEnt: allow calculated data to be modified and one reconstruction for all spectra
Extend this algorithm to allow the calculated data (from the Fourier transform of the image) to be modified before comparing with the observed data in the chi-squared expression.

Introduce two three algorithm properties:

* DataConstAdj. Complex array with dimensions: number of spectra times number of data 
* DataLinearAdj. Complex array with dimensions: number of spectra times number of data 
* PerSpectrumReconstruction. Boolean which default to true

Using the notation in MaxEnt then d_m^c is the calculated data for the m'th data point. To refer to the m'th data point for the k'th spectrum then refer to this as d_k(m)^c. Then adjust these as follows before calculating chi-squared:

modified_d_k(m)^c = DataConstAdj[k,m] + d_k(m)^c * DataLinearAdj[k,m]

When PerSpectrumReconstruction = True the behaviour is as is currently the case. When PerSpectrumReconstruction = False one image is used to fit against all spectra simultaniously, for an example of this see chi-squared expression for MuonMaxent. The chi^2 value is then:

chi^2 = \sum_{k,m} \frac{ ( d_k(m) - modified_d_k(m) )^2}{\sigma_k(m)^2 }

The reconstructed workspace (time domain) should always return N spectra.

KarlPalmen:
**Description of work.**

**To test:**

Run the algorthm with the new properties DataLinearAdj, DataConstAdj & PerSpectrumReconstruction and check it behaves as expected. Also build and check the documentation of Maxent.

I suggest running variations of the two usage examples added to the documentation to the documentation to do the testing.

Fixes #21577 

- [x] Yes
- [ ] No
 

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -229,17 +251,54 @@ std::map<std::string, std::string> MaxEnt::validateInputs() {
...
                                  "workspace must be even for complex data";
+    if (!complex)
+      nHistograms *=
+          2; // Double number of histograms to compare with adjustments.
   }
 
+  // Check linear adjustments, we expect and even number of histograms
+  // and if any, they must be sufficient for all spectra in input workspace,
+  // if per spectrum reconstruction is done.
+  bool psr = getProperty("perSpectrumReconstruction");
```
AnthonyLim23: can we have a more descriptive name than psr?

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -350,19 +420,42 @@ void MaxEnt::exec() {
...
     std::vector<double> image(npoints, background);
 
-    std::vector<double> data;
-    std::vector<double> errors;
+    std::vector<double> data(dataLength, 0.0);
+    std::vector<double> errors(dataLength, 0.0);
     if (complexData) {
-      data = toComplex(inWS, spec, false);  // false -> data
-      errors = toComplex(inWS, spec, true); // true -> errors
+      data = toComplex(inWS, spec, false, sumSpectra);  // false -> data
```
AnthonyLim23: I don't understand these comments. Could you please elaborate?

Framework/Algorithms/test/MaxEntTest.h
```diff
@@ -110,6 +110,51 @@ class MaxEntTest : public CxxTest::TestSuite {
...
+    alg->setPropertyValue("MaxIterations", "1");
+    alg->setProperty("DataLinearAdj", ws);
+    alg->setProperty("DataConstAdj", ws);
+    alg->setProperty("PerSpectrumReconstruction", false);
+    alg->setPropertyValue("ReconstructedImage", "image");
+    alg->setPropertyValue("ReconstructedData", "data");
+    alg->setPropertyValue("EvolChi", "evolChi");
+    alg->setPropertyValue("EvolAngle", "evolAngle");
+
+    TS_ASSERT_THROWS_NOTHING(alg->execute());
```
AnthonyLim23: you forgot to test the output ws dimensions

Framework/Algorithms/test/MaxEntTest.h
```diff
@@ -110,6 +110,51 @@ class MaxEntTest : public CxxTest::TestSuite {
...
+    alg->setProperty("ComplexData", true);
+    alg->setPropertyValue("MaxIterations", "1");
+    alg->setProperty("DataLinearAdj", ws);
+    alg->setProperty("DataConstAdj", ws);
+    alg->setPropertyValue("ReconstructedImage", "image");
+    alg->setPropertyValue("ReconstructedData", "data");
+    alg->setPropertyValue("EvolChi", "evolChi");
+    alg->setPropertyValue("EvolAngle", "evolAngle");
+
+    TS_ASSERT_THROWS_NOTHING(alg->execute());
```
AnthonyLim23: you forgot to test the output workspace dimensions

Framework/Algorithms/src/MaxEnt.cpp
KarlPalmen: It determines whether the complex data is taken from the errors or the Y values. The comments could be improved. That was how they were before I added the 4th argument.

Framework/Algorithms/test/MaxEntTest.h
```diff
@@ -128,9 +173,85 @@ class MaxEntTest : public CxxTest::TestSuite {
...
+    alg->setProperty("InputWorkspace", ws);
+    alg->setProperty("ComplexData", false);
+    alg->setPropertyValue("MaxIterations", "1");
+    alg->setProperty("DataLinearAdj", ws);
+    alg->setPropertyValue("ReconstructedImage", "image");
+    alg->setPropertyValue("ReconstructedData", "data");
+    alg->setPropertyValue("EvolChi", "evolChi");
+    alg->setPropertyValue("EvolAngle", "evolAngle");
+
+    TS_ASSERT_THROWS_ANYTHING(alg->execute());
```
AnthonyLim23: we can use `TS_ASSERT_THROWS(alg->execute(), ...) ` here, where ... is the type of error thrown.

Framework/Algorithms/src/MaxEnt.cpp
KarlPalmen: I haven't thought of one not almost as long as "perSpectrumReconstruction" and I thought "psr" would be OK for a short scope.

AnthonyLim23:
could you please add some step by step instructions?

KarlPalmen:
@AnthonyLim23 Did you intend to request changes?

AnthonyLim23:
no, I pressed the wrong button. I want to test it first

KarlPalmen:
I've added a suggestion to the test instructions, which I hope is helpful.

AnthonyLim23:
I cannot get the algorithm to run from the algorithm dialog. I loaded MUSR00062260 then cropped it from 0 to 16. I then passed it to maxent and nothing happens

KarlPalmen:
When I tried it I got the message 
```
Error in execution of algorithm MaxEnt:
Input data must have non-zero errors.
```
The input workspace must all its error values non-zero, because the algorithm uses each error value as a unit for the Maximum Entropy fitting. I do see that the error message is not precise enough and could be interpreted as ``Input data must have some non-zero errors``.

AnthonyLim23:
I did not even get that. I will not get time to try again today. I can carry on looking at this on Wednesday.

Framework/Algorithms/src/MaxEnt.cpp
Anders-Markvardsen: Should perSpectrumReconstruction be PerSpectrumReconstruction?

AnthonyLim23:
Got it working with some muon data.

Load MUSR 62260
Use RemoveExpDecay
Then crop the output from 0.1 to 10
This output will work in maxent.

I have a few concerns:
1. If you run with ReconstructPerSpectrum set to false, then the reconstructed data is 2 orders of magnitude larger than the original data. 
2. In the case from 1 should we not get a single frequency spectrum but multiple reconstructed data (which differ by a phase)?
3. I am unclear on what DataLinAdj and DataConstAdj are suppose to do. 

On a side note, I have managed to make it crash a couple of times, but not in a reproducible way.

KarlPalmen:
@AnthonyLim23 About concern 1, the workspace has 64 spectra so if PerSpectrumReconstruction is false 64 spectra would be added together leading to the creconstructed data being almost 2 orders of maginute greater than the original data.

KarlPalmen:
I don't understand @AnthonyLim23 concern 2. Would the description of the issue #21577 need changing to ensure it is met?

AnthonyLim23:
@KarlPalmen my first two points are actually related. By adding up all of the reconstructed data we are producing an unphysical output. If you want to check if the reconstructed data (time domain) is correct you will always need one spectrum for every input spectrum.

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -350,19 +421,43 @@ void MaxEnt::exec() {
...
+          for (size_t i = 0; i < numBins; i++) {
+            data[i] += inWS->y(s).rawData()[i];
+            errors[i] += inWS->e(s).rawData()[i];
+          }
+        }
+      } else {
+        data = inWS->y(spec).rawData();
+        errors = inWS->e(spec).rawData();
+      }
+    }
```
AnthonyLim23: am I right in thinking that this means if you sum the spectra then the chi^2 is a comparison between the summed real data and the summed calculated results (time domain)?

KarlPalmen: Yes. I've taken to be effectively a one spectrum comparison on the sum of all the spectra.

AnthonyLim23:
I have had a talk to @Anders-Markvardsen  and updated the issue to be more explicit. We should be calculating N reconstructed spectra every time. If we have one frequency spectrum then we only need one chi^2 calculation, but it must sum across the data points and the spectra.

docs/source/algorithms/MaxEnt-v1.rst
```diff
@@ -526,6 +527,183 @@ Output:
 
 .. figure:: ../images/MaxEntResolutionFactor2.png
    :align: center
+   
+Adjusting the Reconstructed Data
+--------------------------------
+The reconstructed data can be adjusted by multiplying by the *DataLinearAdj* property and then adding the *DataConstAdj* property. 
+Each of these properties is a workspace with complex Y values, just like complex input workspaces. Each Y value is then applied to 
+the corresponding Y values in the reconstructed data at each iteration.
```
AnthonyLim23: I think this would be clearer with an equation

docs/source/algorithms/MaxEnt-v1.rst
```diff
@@ -526,6 +527,183 @@ Output:
...
+   Y = []
+   E = []
+   N = 200
+   w = 3
+   for s in range(0,3):
+      seed(0)
+      for i in range(0,N):
+        x = 2*pi*i/N
+        X.append(x)
+        Y.append(cos(w*x)+(random()-0.5)*0.3)
```
AnthonyLim23: to demonstrate the linear adjustment would it not be better to create a workspace that is offset by a vector of constants (e.g. `cos(w*x)+x` ) then have the `Xlin` as having `Ylin.append(X[i])`? Similarly for multiply

docs/source/algorithms/MaxEnt-v1.rst
```diff
@@ -526,6 +527,183 @@ Output:
...
+   CreateWorkspace(OutputWorkspace='inputws',DataX=X,DataY=Y,DataE=E,NSpec=3)
+
+   # Construct linear adjustment workspace (real = 1, imaginary linear)
+   # no adjustment on first spectrum, double adjustment on third spectrum.
+   Zeroes = []
+   Xlin = []
+   Ylin = []
+   Magnitude = - 0.1
+   # Real values
+   for i in range(0,N):
```
AnthonyLim23: this can be written as `for xValue in X:`

docs/source/algorithms/MaxEnt-v1.rst
KarlPalmen: for ``Xlin``?

AnthonyLim23: line 659

Framework/Algorithms/test/MaxEntTest.h
```diff
@@ -279,6 +279,30 @@ class MaxEntTest : public CxxTest::TestSuite {
...
+    // in the input workspace even though images are summed.
+    alg->setProperty("DataLinearAdj", ws1);
+    alg->setProperty("DataConstAdj", ws1);
+    alg->setProperty("PerSpectrumReconstruction", false);
+    alg->setPropertyValue("ReconstructedImage", "image");
+    alg->setPropertyValue("ReconstructedData", "data");
+    alg->setPropertyValue("EvolChi", "evolChi");
+    alg->setPropertyValue("EvolAngle", "evolAngle");
+
+    TS_ASSERT_THROWS_ANYTHING(alg->execute());
```
AnthonyLim23: we should be catching the type of exception thrown

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -277,22 +336,29 @@ void MaxEnt::exec() {
...
   // this will throw an exception
   size_t npoints = inWS->blocksize() * resolutionFactor;
   // Number of X bins
   const size_t npointsX = inWS->isHistogramData() ? npoints + 1 : npoints;
+  // Linear adjustment of calculated data
+  MatrixWorkspace_const_sptr dataLinearAdj = getProperty("DataLinearAdj");
+  // Constant adjustment of calculated data
+  MatrixWorkspace_const_sptr dataConstAdj = getProperty("DataConstAdj");
+  // Add spectra in reconstruction if false
+  const bool sumSpectra = !getProperty("PerSpectrumImage");
```
Anders-Markvardsen: This line is very hard to read. Please keep variable names as close as possible to the property names to help the person reading the code in this case: const bool perSpectrumImage = getProperty("PerSpectrumImage");

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -350,19 +421,43 @@ void MaxEnt::exec() {
...
     if (complexData) {
-      data = toComplex(inWS, spec, false);  // false -> data
-      errors = toComplex(inWS, spec, true); // true -> errors
+      data = toComplex(inWS, spec, false, sumSpectra); // 3rd arg false -> data
+      errors =
+          toComplex(inWS, spec, true, sumSpectra); // 3rd arg true -> errors
     } else {
-      data = inWS->y(spec).rawData();
-      errors = inWS->e(spec).rawData();
+      if (sumSpectra) {
```
Anders-Markvardsen: When PerSpectrumImage = False why are the data adjusted?

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -204,6 +204,29 @@ void MaxEnt::init() {
...
+          "DataConstAdj", "", Direction::Input, PropertyMode::Optional,
+          boost::make_shared<EqualBinSizesValidator>(errorLevel, warningLevel)),
+      "Adjusts the calculated data by adding to each value the corresponding Y "
+      "value of this workspace. "
+      "If DataLinearAdj is also specified, this addition is done after its "
+      "multiplication. "
+      "The data in this workspace is complex in the same manner as complex "
+      "input data.");
+  declareProperty(
+      "PerSpectrumImage", true,
```
Anders-Markvardsen: As discussed please change the name of this property to PerSpectrumReconstruction

KarlPalmen:
Even though there remains a chance of this being complete by release 3.13, I consider the risk of disruption to Maxent usage by these changes to be considerable.

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -939,55 +940,66 @@ void MaxEnt::populateDataWS(MatrixWorkspace_const_sptr &inWS, size_t spec,
...
-  outWS->setSharedX(nspec + spec, outWS->sharedX(spec));
-  outWS->mutableY(nspec + spec) = std::move(YI);
-  outWS->setSharedE(nspec + spec, outWS->sharedE(spec));
+    outWS->mutableX(specA) = std::move(X);
+    outWS->mutableY(specA) = std::move(YR);
+    outWS->mutableE(specA) = std::move(E);
+    auto nH = outWS->getNumberHistograms();
+    outWS->mutableY(nspec + specA) = std::move(YI);
+    outWS->setSharedX(nspec + specA, outWS->sharedX(spec));
+    outWS->setSharedE(nspec + specA, outWS->sharedE(spec));
```
KarlPalmen: This I cannot debug, because MVS debugger stops showing all variable values.

docs/source/algorithms/MaxEnt-v1.rst
KarlPalmen: I would like to try this, but do not understand what you are suggesting.

Framework/Algorithms/test/MaxEntTest.h
KarlPalmen: Can you give a useful example for `alg->execute()`?

Framework/Algorithms/test/MaxEntTest.h
KarlPalmen: I've tried this but the algorithm class changes the error type to something else and so I doubt it would be useful.

KarlPalmen:
This issue is ready to review again.

docs/source/algorithms/MaxEnt-v1.rst
```diff
@@ -526,6 +541,202 @@ Output:
...
+   # Imaginary values
+   for s in range(0,3):
+       for i in range(0,N):
+           X.append(X[i])
+           Ylin.append(s*Magnitude*X[i])
+           Zeroes.append(0)
+        
+   CreateWorkspace(OutputWorkspace='linadj', DataX=X, DataY=Ylin, DataE=Zeroes, NSpec=6)
+
+   # Construct linear adjustment workspace (real = 0, imaginary linear)
```
Anders-Markvardsen: Do you mean 'Construct const adjustment workspace' here?

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -204,6 +205,30 @@ void MaxEnt::init() {
...
+      "The data in this workspace is complex in the same manner as complex "
+      "input data.");
+  declareProperty(
+      make_unique<WorkspaceProperty<>>(
+          "DataConstAdj", "", Direction::Input, PropertyMode::Optional,
+          boost::make_shared<EqualBinSizesValidator>(errorLevel, warningLevel)),
+      "Adjusts the calculated data by adding to each value the corresponding Y "
+      "value of this workspace. "
+      "If DataLinearAdj is also specified, this addition is done after its "
+      "multiplication. "
```
Anders-Markvardsen: Why have you added the sentence 'If DataLinearAdj is also specified, this addition is done after its "
+      "multiplication' ?

Anders-Markvardsen (dismissed):
Initial functionality tests worked. A few question about the documentation, but otherwise reads clear

Framework/Algorithms/src/MaxEnt.cpp
KarlPalmen: I needed to state the order of the two operations multiplication and addition, because they do not commute.

Anders-Markvardsen: I see what you mean now. However, you have in the documentation a nice equation showing how these properties are applied. Hence, I would argue that either remove this sentence or add 'See equation in documentation for how DataLinearAdj and DataConstAdj are applied'. What do you think

KarlPalmen: I think it should stay in. It saves the reader from having to go to the equation in the documentation.

Anders-Markvardsen: yes in theory:-) but only a few of our users will have the mathematically background to instantly understand what it means

docs/source/algorithms/MaxEnt-v1.rst
```diff
@@ -526,6 +541,202 @@ Output:
...
+   
+.. figure:: ../images/MaxEntAdjust.png
+   :align: center
+   
+   The first plot shows the reconstructed data and the second plot the corresponding image.
+   The first three spectra plotted are the real parts of the 3 spectra
+   and the other three spectra are the imaginary parts.
+   
+This can also be done with the spectra of the data effectively concatenated and converted into a single image.
+The reconstructed data is then obtained by converting the image back to data copied for each spectrum and applying the adjustments.
```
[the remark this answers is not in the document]
Anders-Markvardsen: Could you update the rst accordingly please?

Framework/Algorithms/src/MaxEnt.cpp
KarlPalmen: but they will understand the equation.

KarlPalmen: I've added the "See equation ...".

Framework/Algorithms/inc/MantidAlgorithms/MaxEnt/MaxentCalculator.h
```diff
@@ -5,13 +5,12 @@
...
 #include "MantidAlgorithms/MaxEnt/MaxentEntropy.h"
 #include "MantidAlgorithms/MaxEnt/MaxentTransform.h"
-#include <memory>
 
 namespace Mantid {
 namespace Algorithms {
 
-using MaxentEntropy_sptr = std::shared_ptr<MaxentEntropy>;
-using MaxentTransform_sptr = std::shared_ptr<MaxentTransform>;
+// using MaxentEntropy_sptr = boost::shared_ptr<MaxentEntropy>;
```
Anders-Markvardsen: Should these two lines be deleted?

Framework/Algorithms/inc/MantidAlgorithms/MaxEnt.h
```diff
@@ -63,7 +63,7 @@ class DLLExport MaxEnt : public API::Algorithm {
   void exec() override;
   /// Returns spectrum 'spec' as a complex vector
   std::vector<double> toComplex(API::MatrixWorkspace_const_sptr &inWS,
-                                size_t spec, bool errors);
+                                size_t spec, bool errors, bool concatSpec);
```
AnthonyLim23: should be ``concatenatedSpectra`` for consistency

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -373,11 +481,13 @@ void MaxEnt::exec() {
     Progress progress(this, 0.0, 1.0, nIter);
 
     // Run maxent algorithm
+    bool stop = false;
```
AnthonyLim23: I dont think stop is the best name, I would assume it has more to do with a user cancelling the algorithm than it having converged before the maximum number of iterations.

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -446,33 +563,50 @@ void MaxEnt::exec() {
...
 */
 std::vector<double> MaxEnt::toComplex(API::MatrixWorkspace_const_sptr &inWS,
-                                      size_t spec, bool errors) {
+                                      size_t spec, bool errors,
+                                      bool concatSpec) {
   const size_t numBins = inWS->y(0).size();
-  std::vector<double> result(numBins * 2);
+  size_t nSpec = inWS->getNumberHistograms() / 2;
+  std::vector<double> result;
+  result.reserve(numBins);
```
AnthonyLim23: should this still be ```result.reserve(numBins*2)```?

Framework/Algorithms/src/MaxEnt.cpp
```diff
@@ -797,57 +932,86 @@ void MaxEnt::populateImageWS(MatrixWorkspace_const_sptr &inWS, size_t spec,
...
-  }
+  // Loop over each spectrum being analyzed - one spectrum unless concatenated
+  for (size_t specA = spec; specA < spec + nSpecAnalyzed; specA++) {
 
-  // Y values
-  if (complex) {
-    for (int i = 0; i < npoints; i++) {
-      YR[i] = result[2 * i];
-      YI[i] = result[2 * i + 1];
+    MantidVec X(spectrumLengthX);
```
AnthonyLim23: should these not be the ```binEdges``` and ```counts``` or ```HistogramX``` etc. data types? http://docs.mantidproject.org/v3.9.2/concepts/HistogramData.html

Framework/Algorithms/src/MaxEnt/MaxentTransformMultiFourier.cpp
```diff
@@ -0,0 +1,121 @@
...
+    for (size_t i = 0; i < dataOneSpec.size(); i++) {
+      data.emplace_back(dataOneSpec[i]);
+    }
+  }
+
+  // Apply adjustments (we assume there are sufficient adjustments supplied)
+  double dataR = 0.123456789;
+  double dataI = 0.987654321;
+  if (!m_linearAdjustments.empty() && !m_constAdjustments.empty()) {
+    for (size_t i = 0; i < data.size(); i++) {
```
AnthonyLim23: I thought the data was stored as the first N were the real part and the second N are the imaginary part. If that is the case, then I am not convinced by the else statement as the ```dataR``` and ```dataI``` will have the values from the Nth real data

KarlPalmen: This is not the case here. We are dealing with a single (concatenated) spectrum arranged (real1, imaginary1, real2, imaginary2, etc.) as in MaxentTransformFourier.

Framework/Algorithms/src/MaxEnt.cpp
KarlPalmen: I haven't seen these data types used here before. I was just using the data types already in use.

Framework/Algorithms/src/MaxEnt/MaxentTransformMultiFourier.cpp
AnthonyLim23: can we have comment stating this?

Framework/Algorithms/test/MaxEnt/MaxentTransformMultiFourierTest.h
```diff
@@ -0,0 +1,232 @@
...
+    MaxentTransformMultiFourier transformMF(dataSpaceMF, imageSpaceMF, 3);
+    MaxentSpace_sptr dataSpaceF = boost::make_shared<MaxentSpaceComplex>();
+    MaxentSpace_sptr imageSpaceF = boost::make_shared<MaxentSpaceComplex>();
+    MaxentTransformFourier transformF(dataSpaceF, imageSpaceF);
+
+    std::vector<double> realDataMF = {1, -1, 1,  -1, 1,  -1, 1,  -1,
+                                      2, 2,  -2, -2, 2,  2,  -2, -2,
+                                      4, 4,  4,  4,  -4, -4, -4, -4};
+    std::vector<double> realDataF = {7, 5, 3, 1, -1, -3, -5, -7};
+
```
AnthonyLim23: sorry, why should these be the same?

Framework/Algorithms/test/MaxEnt/MaxentTransformMultiFourierTest.h
```diff
@@ -0,0 +1,232 @@
...
+                          1e-4);
+          TS_ASSERT_DELTA(result[i + 2 * n],
+                          linearAdjustments[i + 2 * n] * resultF[i - 1] +
+                              constAdjustments[i + 2 * n],
+                          1e-4);
+        }
+      }
+    }
+  }
+};
```
AnthonyLim23: unless I have missed something, it looks like the values for transform is never tested as it is always compared against another transform. So if the alg changes it  could still pass even if it gives a bad answer.

Framework/Algorithms/test/MaxEntTest.h
```diff
@@ -110,6 +110,80 @@ class MaxEntTest : public CxxTest::TestSuite {
...
+    TS_ASSERT_EQUALS(data->getNumberHistograms(), nHist);
+    TS_ASSERT_EQUALS(image->getNumberHistograms(), 2);
+    TS_ASSERT_EQUALS(chi->getNumberHistograms(), 1);
+    TS_ASSERT_EQUALS(angle->getNumberHistograms(), 1);
+
+    TS_ASSERT_EQUALS(data->blocksize(), nBins);
+    TS_ASSERT_EQUALS(image->blocksize(), nBins);
+    TS_ASSERT_EQUALS(chi->blocksize(), 1);
+    TS_ASSERT_EQUALS(angle->blocksize(), 1);
+  }
```
AnthonyLim23: we should also check the values when using the new adjustments

Framework/Algorithms/test/MaxEnt/MaxentTransformMultiFourierTest.h
KarlPalmen: Which alg?

docs/source/algorithms/MaxEnt-v1.rst
AnthonyLim23: you would have ```offset = [1,2,3,4]``` then your y data would be ```y[i] = cos(w*x[i])+offset``` Then the ```DataLinearAdj``` would just be a workspace with the y values from offset

Framework/Algorithms/test/MaxEnt/MaxentTransformMultiFourierTest.h
AnthonyLim23: your ```dataToImage``` method is the one that stood out. There might be others too

AnthonyLim23:
Can we add this to the release notes

KarlPalmen:
@AnthonyLim23 You've accidentally closed this pull request.

KarlPalmen:
I'll add to the release notes last. This will require a merger from master of the release notes.

Framework/Algorithms/test/MaxEnt/MaxentTransformMultiFourierTest.h
KarlPalmen: Please explain in more detail.

KarlPalmen: Perhaps, that is because I comparing with the effects of the fourrier transform. I'll have look into this.
I have not tested the Fourrier transform here, because it is tested by MaxentTransformFourier.

Framework/Algorithms/test/MaxEnt/MaxentTransformMultiFourierTest.h
KarlPalmen: This needs explaining in more detail.

KarlPalmen: If either the `dataToImage` or `imageToData` methods of `MaxentTransformMultiFourier` were to give bad answer, it would show up against a comparison with the same method of `MaxentTransformFourier`. If you think I'm wrong here, please give a senario where a bad result would not cause the test to fail.

AnthonyLim23: could you please give me a link to the line of code you think check the transform values?

Framework/Algorithms/test/MaxEntTest.h
AnthonyLim23: we should also check the simple case

Framework/Algorithms/test/MaxEnt/MaxentTransformMultiFourierTest.h
KarlPalmen: I only compare against the values given by the plain fourier transform.

Framework/Algorithms/test/MaxEntTest.h
KarlPalmen: Why?

AnthonyLim23: if this one fails we know the bug is in adjustment. If ```test_adjustments_three_spectra_together``` fails and the simple case does not then we know the bug is in the combining of the spectra. At present we would not know if its the combining or the adjustment.

KarlPalmen: I don't see any sense in testing values in tests where the maximumum number of iterations has been set to 1.

AnthonyLim23: could we have a test (or system test) with more iterations? Does ```test_adjustments_three_spectra_together ``` just have 1 iteration too?

KarlPalmen: Doesn't test `test_three_spectra_together` take care of the case where the spectra are just combined and there are no adjustments?

KarlPalmen: I've got tests for adjustments too.

KarlPalmen: Do you want a test that explicitly tests the arithmetic of the adjustments rather than showing that the fitting works with adjustments and gives certain data values not explicitly determined by adjustment arithmetic?

KarlPalmen: I note that MaxentTransformMultiFourier does explicitly test the adjustment arithmestic (but with real and complex adjustmests separately to keep it simple). So any such test in Maxent would merely confirm the the adjusted values were delivered correctly to the data workspace.

AnthonyLim23: that is still useful information.

KarlPalmen: So I I'll go ahead with a cursory test of Maxent the explicitly tests the arithmetic.

Framework/Algorithms/src/MaxEnt.cpp
KarlPalmen: I've created issue #23106 to deal with this.

Framework/Algorithms/test/MaxEntTest.h
KarlPalmen: The test took most of the day to write, but has shown a problem with to `toComplex` function in `MaxEnt`. It interleaves the spectra instead of concatenating them as required. I'll be away until Wednesday next week, then I'll fix it.

KarlPalmen:
The issue is ready for reviewing & testing again.

Framework/Algorithms/src/MaxEnt/MaxentTransformMultiFourier.cpp
KarlPalmen: Cooments added in argument descriptions here and in `MaxentTransformFourier.cpp`.

Framework/Algorithms/test/MaxEntTest.h
KarlPalmen: Work on this has been completed and the issue is ready for testing again.

Framework/Algorithms/test/MaxEntTest.h
```diff
@@ -1073,6 +1132,35 @@ class MaxEntTest : public CxxTest::TestSuite {
     TS_ASSERT_EQUALS(data->readX(0), ws->readX(0));
   }
 
+  MatrixWorkspace_sptr createWorkspaceWithYValues(size_t nHist, size_t length, std::vector<double> const &YVal) {
+
+    size_t nPts = length * nHist;
+    TS_ASSERT_EQUALS(nPts, YVal.size());
+    MantidVec X(nPts);
+    MantidVec Y(nPts);
+    MantidVec E(nPts);
```
AnthonyLim23: could we please use the new histogram data types

KarlPalmen: This is covered by issue #23106

AnthonyLim23:
there is an error in ``MaxentCalculatorTest``

AnthonyLim23:
@Anders-Markvardsen this seems to be ok now

Anders-Markvardsen (approved):
Requested modifications made to code and test examples. Looks OK to me also

PR #22565, opened 2018-06-12
Indirect - ISISEnergyTransfer - Custom Detector Grouping Option

MatthewBowles:
Replaces default option for detector grouping in the `ISISEnergyTransfer' interface with a 'Custom' option that allows specific spectra or ranges to be entered.
Also fixes the 'Group' option which didn't work.

**Report to:**  @brandonhewer
 
**To test:**
- Navigate to Interfaces -> Indirect -> Data Reduction
- Enter a file to load e.g. `IRS26176`
- In Detector Grouping, check 'Custom' is selected by default
- Check the line edit box next to 'Custom' is only present when 'Custom' is selected
- Enter spectra and/or a range in this box: `1, 2, 4-7`
- The resulting workspace should contain only the spectra specified (+ spectra min): `55, 56, 58-61`

Group option:
- Run again using the 'Group' option in Detector grouping (with any number of groups) and check there are no errors. Result workspace should contain all spectra (54-104).

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/docs/source/release/v3.13.0/indirect_inelastic.rst
+++ b/docs/source/release/v3.13.0/indirect_inelastic.rst
@@ -15,7 +15,7 @@ Data Reduction Interfaces
 -------------------------
 
 - Added 'Sum Files' checkbox to ISIS Calibration, to sum a specified range of input files on load.
-- Detector grouping in ISISEnergyTransfer:  add custom grouping method to allow specific spectra or ranges, and
+- Detector grouping in ISISEnergyTransfer:  added custom grouping method to allow specific spectra or ranges, and
   the 'groups' method now includes all spectra including remainder.
 
 
--- a/docs/source/release/v3.13.0/indirect_inelastic.rst
+++ b/docs/source/release/v3.13.0/indirect_inelastic.rst
@@ -15,6 +15,7 @@ Data Reduction Interfaces
 -------------------------
 
 - Added 'Sum Files' checkbox to ISIS Calibration, to sum a specified range of input files on load.
+- Add custom method to specify detector grouping in ISISEnergyTransfer as individual spectra or a range.
 
 Algorithms
 ----------
--- a/docs/source/release/v3.13.0/indirect_inelastic.rst
+++ b/docs/source/release/v3.13.0/indirect_inelastic.rst
@@ -15,7 +15,9 @@ Data Reduction Interfaces
 -------------------------
 
 - Added 'Sum Files' checkbox to ISIS Calibration, to sum a specified range of input files on load.
-- Add custom method to specify detector grouping in ISISEnergyTransfer as individual spectra or a range.
+- Detector grouping in ISISEnergyTransfer:  add custom grouping method to allow specific spectra or ranges, and
+  the 'groups' method now includes all spectra including remainder.
+
 
 Algorithms
 ----------
--- a/qt/scientific_interfaces/Indirect/ISISEnergyTransfer.cpp
+++ b/qt/scientific_interfaces/Indirect/ISISEnergyTransfer.cpp
@@ -48,7 +48,7 @@ ISISEnergyTransfer::ISISEnergyTransfer(IndirectDataReduction *idrUI,
   mappingOptionSelected(m_uiForm.cbGroupingOptions->currentText());
 
   // Add validation to custom detector grouping
-  QRegExp re("([0-9]+[-]?[0-9]*,[ ])*[0-9]+[-]?[0-9]*");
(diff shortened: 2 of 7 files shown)


qt/scientific_interfaces/Indirect/ISISEnergyTransfer.cpp
```diff
@@ -273,9 +273,7 @@ void ISISEnergyTransfer::run() {
       createMapFile(m_uiForm.cbGroupingOptions->currentText());
```
KarlPalmen: I get a compilation error here concerning type conversion between QString and const std::string.

qt/scientific_interfaces/Indirect/ISISEnergyTransfer.cpp
```diff
@@ -453,40 +451,36 @@ ISISEnergyTransfer::createMapFile(const QString &groupType) {
...
+    return std::make_pair("Custom", createDetectorGroupingString());
   } else if (groupType == "Default")
-      return qMakePair(QString("IPF"), QString());
+      return std::make_pair("IPF", "");
     else if (groupType == "Custom")
-      return qMakePair(QString("Custom"), m_uiForm.leCustomGroups->text());
+      return std::make_pair("Custom", m_uiForm.leCustomGroups->text().toStdString());
     else {
     // Catch All and Individual
     return qMakePair(groupType, QString());
```
KarlPalmen: I get a compilation error here about conversion between QPair and std::pair here.

KarlPalmen:
I don't see any "custom" in the Detector Grouping. I see "default" selected by default.

KarlPalmen:
I had my default facility set to ILL for previous test. Setting it back to ISIS and instrument to IRIS made custom appear as described.

MatthewBowles:
That's weird. It's clearly changed [here](https://github.com/mantidproject/mantid/pull/22565/files#diff-48cd8b084a7a38294e273ca24e745a01R218).

KarlPalmen:
The crash occurred again on retry. I used `irs26176_graphite002_red.nxs` .

KarlPalmen:
Looks like you may be assigning an incorrect type as a `minimum spectrum` property in this case and this exception is not caught.

SpencerHowells:
Data reduction uses a RAW file NOT a *_red.nxs file. Typing the run number is the method of entry.

SpencerHowells:
Also, this interface is not used by ILL - they have their own version.

SpencerHowells:
Further to the ILL version, if the default Facility is set to ILL then launching Data Reduction will give a GUI specific to ILL. The aim is to do this for all Facilities which want to use this GUI.

KarlPalmen:
If typing a run number is the only method of entry, I would not expect a file browser to be available. If one does want to support file selection, then the set of file types needs to be limited appropeiately (e.g to .raw). This may be beyond the scope of thjis issue.

KarlPalmen:
I got the desired functionality after selecting the raw file for run 57093, but the workspace only had 'nan' in it, which might not be related to this issue. If I just type a number, I get invalid runs.

KarlPalmen:
I got numeric data in the workspace from run 57907 (selected as raw file).

KarlPalmen:
If you can easily change the default file type in the file browser to "*.raw", please do so (assuming only 'raw' file are valid).

KarlPalmen:
I don't see any update of the documentation for the Indirect interface.

KarlPalmen:
I'll approve this change, when I see a good documentation update.

SpencerHowells:
When Browsing you can select a RAW file - we usually test on iris26176.raw.

SpencerHowells:
Browse filters on *.nxs, *.raw & *.sav which are all raw files. The nxs file would be eg iris26176.nxs.

docs/source/release/v3.13.0/indirect_inelastic.rst
```diff
@@ -15,6 +15,7 @@ Data Reduction Interfaces
 -------------------------
 
 - Added 'Sum Files' checkbox to ISIS Calibration, to sum a specified range of input files on load.
+- Add custom method to specify detector grouping in ISISEnergyTransfer as individual spectra or a range.
```
KarlPalmen: Should it not be `Added` instead of `Add` as in previous item?

KarlPalmen (changes requested):

KarlPalmen (approved):

SimonHeybrock:
@brandonhewer Merged.

PR #22575, opened 2018-06-13
Gracefully abort if ws doesn't load during project loading

### LINKED ISSUE #22540: Loading projects with missing nexus files throws
### Expected behavior
Load a project file when the associated nexus file is missing should be handled gracefully

### Actual behavior
An exception is thrown which is not caught

- Save a new project file out with at least one workspace associated
- Delete the nexus file associated with that project
- Attempt to load the saved project

### Platforms affected
All

DavidFair:
**Description of work.**
Adds checks and gracefully stops project loading if a workspace could not be loaded for any reason. 
This will prevent Mantid getting an unhandled exception if the associated `.nxs` file has been deleted/moved.

**To test:**
- Create a sample workspace
- Open a plot
- Save the project file
- Go to project file location, delete the associated .nxs file
- Attempt to load project
- Check it is handled gracefully

Fixes #22540 

- [ ] Yes
- [x] No

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/MantidPlot/src/ApplicationWindow.cpp
+++ b/MantidPlot/src/ApplicationWindow.cpp
@@ -4670,12 +4670,11 @@ ApplicationWindow *ApplicationWindow::openProject(const QString &filename,
   // Open as a top level folder
   ProjectSerialiser serialiser(this);
   try {
-	  serialiser.load(lines, fileVersion);
-  }
-  catch (std::runtime_error &e) {
-	  g_log.error(e.what());
-	  // We failed to load - bail out
-	  return this;
+    serialiser.load(lines, fileVersion);
+  } catch (std::runtime_error &e) {
+    g_log.error(e.what());
+    // We failed to load - bail out
+    return this;
   }
 
   if (d_loaded_current)
@@ -14157,12 +14156,11 @@ Folder *ApplicationWindow::appendProject(const QString &fn,
   ProjectSerialiser serialiser(this);
 
   try {
-	  serialiser.load(lines, fileVersion);
-  }
-  catch (std::runtime_error &e) {
-	  g_log.error(e.what());
-	  // We failed to load - bail out
-	  return nullptr;
+    serialiser.load(lines, fileVersion);
+  } catch (std::runtime_error &e) {
+    g_log.error(e.what());
+    // We failed to load - bail out
+    return nullptr;
   }
 
   // Restore the selected folder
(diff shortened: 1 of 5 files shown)


MantidPlot/src/ProjectSerialiser.cpp
```diff
@@ -45,6 +46,19 @@ ProjectSerialiser::ProjectSerialiser(ApplicationWindow *window, Folder *folder)
     : window(window), m_currentFolder(folder), m_windowCount(0),
       m_saveAll(true) {}
 
+bool MantidQt::API::ProjectSerialiser::checkAllWorkspacesInAds(
```
josephframsay: Could this use `std::all_of`? Something like
```
const auto &ADS = Mantid::API::AnalysisDataService::Instance();
return std::all_of(expectedWsNames.begin(), expectedWsNames.end(),
                   [&](const std::string &wsName){ return ADS.doesExist(wsName); });

File: MantidPlot/src/ProjectSerialiser.cpp [hunk 2]
```diff
@@ -708,7 +726,20 @@ void ProjectSerialiser::populateMantidTreeWidget(const QString &lines) {
       }
     } else // ...not a group so just load the workspace
     {
-      loadWsToMantidTree((*line).toStdString());
+      std::string wsName = line->toStdString();
```
josephframsay: `const auto`?

MantidPlot/src/ApplicationWindow.cpp
```diff
@@ -14148,7 +14154,14 @@ Folder *ApplicationWindow::appendProject(const QString &fn,
...
   // Open folders
   ProjectSerialiser serialiser(this);
-  serialiser.load(lines, fileVersion);
+
+  try {
+    serialiser.load(lines, fileVersion);
+  } catch (std::runtime_error &e) {
+    g_log.error(e.what());
+    // We failed to load - bail out
+    return nullptr;
```
josephframsay: I don't understand why this method needs to return anything - it doesn't ever seem to get assigned to, and it only returns `nullptr`. Am I missing something, or could it be changed to `void`?

josephframsay (changes requested):
Changes are nice, and the test works as expected. Is there a way for any of this to be unit tested? Possibly  `checkAllWorkspacesInAds` could to be a method on the ADS itself, and then at least we'd get to unit test that (plus I bet it's got use cases outside of this example)

DavidFair:
That's a really good idea. I'll add the method to the ADS, plus unit testing.

MantidPlot/src/ApplicationWindow.cpp
DavidFair: I'm matching the existing returns for this method. So I'm not sure why we are returning handles, only the fact that we need to (unless we want to change multiple callers in ApplicationWindow)

josephframsay (approved):

josephframsay:
Pending build servers, now looks good to go

PR #22578, opened 2018-06-14
New algorithm: SaveGDA

### LINKED ISSUE #22468: Gem texture focusing step 2: save GDA format
#Once #22456 is done we can move onto step 2, which means saving to a new file format, the `.gda` (or RALF) format

josephframsay:
A new algorithm has been implemented to save a focused diffraction workspace to the MAUD-readable GDA format.

This algorithm has a few quirks to it, that could be with being explained (they're also cleared up in the documentation). The big one is that GDA is a TOF format, but the TOF values that SaveGDA writes out are not necessarily the actual recorded ones.

This is because of the following. MAUD reads in a TOF file and converts it to D using a calibration file, similar to a GSAS calibration file (ie containing DIFC conversion factors). The use case for this is a 160-bank workspace. We can't save the actual TOF output, as it is not practical to produce a calibration file for this many banks - there just isn't enough data in each of the banks for such runs. Instead, we use the 6-bank calibration file and essentially manually undo AlignDetectors. Since conversion is done with the same parameters when going from D to TOF (in Mantid) as when going back (in MAUD), the end result is still correct.

After this algorithm is merged, the final step will be to create the MAUD calibration file as part of the texture reduction - it will be at this point that we'll be able to make sure that the same conversion factors are used for both conversions.

**To test:**

The GDA format has been validated by scientists, so code review and a quick crack at breaking it should be fine

Test data is 
[PR22578.zip](https://github.com/mantidproject/mantid/files/2102017/PR22578.zip)

Make sure the following runs OK

```
input_ws = Load(Filename="GEM61785_focused_texture_d.nxs")
SaveGDA(InputWorkspace=input_ws,
        GSASParamFile="GEM_PF1_PROFILE.IPF",
        GroupingScheme=[1] * 3 + [2] * 8 + [3] * 20 + [4] * 42 + [5] * 52 + [6] * 35,
        Filename="/somewhere.gda")
```

Closes #22468

[Yes](https://github.com/mantidproject/mantid/pull/22578/commits/3f8e311de4ec94a099655d93baacde699f3e66c0)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/DataHandling/inc/MantidDataHandling/SaveGDA.h
```diff
@@ -0,0 +1,52 @@
+#ifndef MANTID_DATAHANDLING_SAVEGDA_H_
+#define MANTID_DATAHANDLING_SAVEGDA_H_
+
+#include "MantidAPI/Algorithm.h"
+#include "MantidDataHandling/DllConfig.h"
+
+#include "MantidAPI/MatrixWorkspace_fwd.h"
+
+#include <unordered_map>
```
DavidFair: Your are missing `std::string` despite using it in this header.

Framework/DataHandling/src/SaveGDA.cpp
```diff
@@ -0,0 +1,260 @@
...
+namespace { // helper functions
+
+const int POINTS_PER_LINE = 4;
+
+double mean(const std::vector<double> &values) {
+  return std::accumulate(values.begin(), values.end(), 0.0) /
+         static_cast<double>(values.size());
+}
+
+double computeAverageDeltaTByT(const HistogramData::HistogramX &tValues) {
```
DavidFair: Can we add a comment explaining what a *DeltaTByT* is

Framework/DataHandling/src/SaveGDA.cpp
```diff
@@ -0,0 +1,260 @@
...
+                           });
+  // First element is just first element of tValues, so remove it
+  deltaTByT.erase(deltaTByT.begin());
+  return mean(deltaTByT);
+}
+
+std::string generateBankHeader(int bank, int minT, size_t numberBins,
+                               double deltaTByT) {
+  std::stringstream stream;
+  const auto numberLines = (int)std::ceil((double)numberBins / POINTS_PER_LINE);
```
DavidFair: Could we change this to `size_t`. The number cannot be negative, however I can see users overflowing the int if they use a corrupt workspace for example

Framework/DataHandling/src/SaveGDA.cpp
```diff
@@ -0,0 +1,260 @@
...
+boost::optional<std::vector<std::string>>
+getParamLinesFromGSASFile(const std::string &paramsFilename) {
+  std::ifstream paramsFile;
+  paramsFile.open(paramsFilename);
+
+  if (paramsFile.is_open()) {
+    std::vector<std::string> paramLines;
+    std::string line;
+    while (std::getline(paramsFile, line)) {
+      if (line.find("ICONS") != std::string::npos) {
```
DavidFair: Comment explaining why ICONS is significant

Or even better a comment and a variable name e.g.:

```
// Icons is used as a delimiter in foo
const std::string delimiter = "ICONS"

if (line.find(delimiter) ...)
```

Framework/DataHandling/src/SaveGDA.cpp
```diff
@@ -0,0 +1,260 @@
...
+    const auto &bankCalibParams = calibParams[groupingScheme[i] - 1];
+
+    std::vector<double> tofScaled;
+    tofScaled.reserve(d.size());
+    std::transform(d.begin(), d.end(), std::back_inserter(tofScaled),
+                   [&bankCalibParams](const double dVal) {
+                     return (dVal * bankCalibParams.difa +
+                             dVal * dVal * bankCalibParams.difc +
+                             bankCalibParams.tzero) *
+                            32;
```
DavidFair: Document this magic value in a variable, even if it itself is magic (does that make sense?)

Framework/DataHandling/test/SaveGDATest.h
```diff
@@ -0,0 +1,308 @@
...
+      // each point has 3 space-separated items on a line
+      numPoints += static_cast<int>(lineItems.size()) / 3;
+      numLines++;
+
+      for (size_t i = 0; i < lineItems.size(); i += 3) {
+        TOFs.emplace_back(std::stod(lineItems[i]));
+      }
+    }
+
+    const auto expectedNumPoints = std::stoi(headerItems[2]);
```
DavidFair: Check the length of the array before accessing into it. An assert failing is far easy to diagnose than an exception violation in a unit test

Framework/DataHandling/test/SaveGDATest.h
```diff
@@ -0,0 +1,308 @@
...
+    TS_ASSERT_EQUALS(intensity[103], 1);
+    TS_ASSERT_EQUALS(error[103], 34);
+
+    TS_ASSERT_EQUALS(tof[123], 100725);
+    TS_ASSERT_EQUALS(intensity[123], 1);
+    TS_ASSERT_EQUALS(error[123], 35);
+
+    TS_ASSERT_EQUALS(tof[3000], 239053);
+    TS_ASSERT_EQUALS(intensity[3000], 800);
+    TS_ASSERT_EQUALS(error[3000], 894);
```
DavidFair: Same for all of these arrays, what happens if there are not 3000 elements (due to a regression in the original algorithm)

docs/source/algorithms/SaveGDA-v1.rst
```diff
@@ -0,0 +1,118 @@
...
+
+   # Print the header and the 4 lines from the middle of the file
+   # rstrip the header just to make the doctest script happy
+   print(file_contents[0].rstrip())
+   for i in range(100, 104):
+       print(file_contents[i])
+
+.. testcleanup:: SaveGDA
+
+   os.remove(output_file)
```
DavidFair: You'll want to remove the ws from the ADS too, otherwise it starts to fill as each doc test finishes

DavidFair (changes requested):
Some minor changes required, but otherwise a good clean implementation

Framework/DataHandling/test/SaveGDATest.h
josephframsay: This is already taken care of in lines 225 to 228

DavidFair (approved):

DavidFair:
Everything looks good, the file was correctly saved on my machine and the new unit tests passed

peterfpeterson (changes requested):
`SaveBankScatteringAngles` is not listed with the other algorithms in the documentation. It should only need `.. categories::` near the bottom of the file to fix that.

josephframsay:
@peterfpeterson I think this is ready for review again - were there any other changes you wanted made?

peterfpeterson (approved):

PR #22599, opened 2018-06-18
GEM texture mode step 3: save MAUD calibration file and reverse bank numbers

### LINKED ISSUE #22526: GEM texture focusing step 3: reverse bank numbers and save .maud
As the final step of GEM texture focusing (following #22456 and #22468), saving a `.maud` instrument parameters file should be implemented. An example of one of these is 
[here](https://github.com/mantidproject/mantid/files/2067651/GEM61785_MAUD_func2.zip), corresponding to the `GEMmaudstr` sections of Saurabh's GEM script [here](https://github.com/mantidproject/mantid/files/2067656/GEM2MAUD_func2.zip).

Note that Saurabh's script has bank numbers in reverse order (ie the first bank saved has ID 160, the second 159 and so on). At present `SaveGDA` also follows this convention, but this is only for compatibility with the existing `.maud` files, and this PR should correct the reversal (essentially revert [this](https://github.com/mantidproject/mantid/pull/22463/commits/d407b9e92c2e5a3b3e0d8272244fe74f19a56e89) commit and give banks ascending bank IDs in the new script)

josephframsay:
A new algorithm, `SaveGEMMAUDParamFile` was implemented. This algorithm reads various calibration parameters needed by MAUD from a focused workspace and a GSAS calibration file, and outputs them to a MAUD-readable format.

The new algorithm is incorporated into the `isis_powder` routines for GEM.

**To test:**

As always, if not reviewing from ISIS, you'll need to change the second line and do some setup locally - ping me a message for an explanation of how to set up ISIS_Powder directories, as it can be fiddly.

The `.maud` format has been validated by GEM scientists, so just make sure the following runs OK, and the ``.maud`` file is saved in ``\\olympic\Bablyon5\Public\JoeRamsay\GEM\Output\cycle_18_1\GEM``

```
from isis_powder import Gem

gem = Gem(config_file=r"\\olympic\Babylon5\Public\JoeRamsay\GEM\Gem_config_example.yaml",
          vanadium_normalisation=False)

gem.focus(run_number=84734,
          texture_mode=True)
```

Try setting `save_maud_calib` to ``False`` in the `gem.focus` line - the file shouldn't be saved. Have a go a breaking it and code review.

Fixes #22526 

- [x] Yes
- [ ] No

[Here](https://github.com/mantidproject/mantid/commit/bd6fbba6b7b709f200e08eba464f3c4bed67d272)

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


Framework/DataHandling/src/SaveGDA.cpp
```diff
@@ -179,9 +178,8 @@ void SaveGDA::exec() {
     const auto numPoints =
         std::min({tofScaled.size(), intensity.size(), error.size()});
 
-    const auto header =
-        generateBankHeader(numBanks - i, (int)std::round(tofScaled[0]),
-                           numPoints, averageDeltaTByT);
+    const auto header = generateBankHeader(i + 1, (int)std::round(tofScaled[0]),
```
samueljackson92: Don't use C style casts. Use `static_cast<int>`

Framework/PythonInterface/plugins/algorithms/SaveGEMMAUDParamFile.py
```diff
@@ -0,0 +1,204 @@
...
+                alpha_ones.append(float(line_items[4]))
+                beta_zeros.append(float(line_items[5]))
+                beta_ones.append(float(line_items[6]))
+            elif re.match(self.PROFILE_COEFFS_LINE_2, line):
+                sigma_zeros.append(float(line_items[3]))
+                sigma_ones.append(float(line_items[4]))
+                sigma_twos.append(float(line_items[5]))
+
+        return distances, difas, difcs, tzeros, alpha_zeros, alpha_ones, \
+            beta_zeros, beta_ones, sigma_zeros, sigma_ones, sigma_twos
```
samueljackson92: Could we maybe tidy up some of this by sticking all the entries in a `dict` instead of a tuple? It feels like the more natural data structure.

I don't think ordering is particular important here, but we could use an ordered dict if that is necessary.  You can also use a default dict to initialise the entires to a list by default instead of checking where a key exists yet.

Framework/PythonInterface/plugins/algorithms/SaveGEMMAUDParamFile.py
```diff
@@ -0,0 +1,204 @@
...
+        corresponding bank
+        """
+        return (bank_param_list[grouping_scheme[spec_num] - 1] for spec_num in spectrum_numbers)
+
+    def _find_isis_powder_dir(self):
+        script_dirs = config["pythonscripts.directories"].split(";")
+        diffraction_dir = None
+        for dir in script_dirs:
+            if "Diffraction" in dir:
+                diffraction_dir = dir
```
samueljackson92: Could we just use `filter` here and take the last one?

samueljackson92:
```python
script_dirs = config["pythonscripts.directories"].split(";")
script_dirs = filter(lambda directory: "Diffraction" in directory, script_dirs)
diffraction_dir = script_dirs.pop() if len(script_dirs) > 0 else None
```

samueljackson92: Are we sure we always want the last directory?

docs/source/algorithms/SaveGDA-v1.rst
```diff
@@ -113,7 +113,7 @@ Output:
 
 .. testoutput:: SaveGDA
 
-    BANK 4 4246  1062 RALF  27388  96  27388 0.001 ALT
+    BANK 1 4246  1062 RALF  27388  96  27388 0.001 ALT
```
samueljackson92: Why did the output of this algorithm change?

samueljackson92 (changes requested):
Changes seem to work fine. A few comments from code review

docs/source/algorithms/SaveGDA-v1.rst
josephframsay: In the old GEM scripts bank numbers went from high to low in the GDA and MAUD files (and they need to be in the same order in both files), so when we could only generate the GDA from Mantid we kept them in descending order so GEM could use the GDA with their old MAUD file. GEM prefer them to be ascending, so now that we can generate both files from Mantid, which happened in this PR, the bank numbers can be reversed in the GDA

samueljackson92: Ok. Sounds good.

samueljackson92: But does it mean we're no longer backwards compatible?

josephframsay: It means that a GDA made with our scripts won't work with a MAUD made with theirs, and vice versa. Previously both files were made with scientist-written scripts that made no mention of Mantid, so it's not a case of not being backwards compatible with ourselves. The scientists also understand that this is the case and are happy with it

Framework/PythonInterface/plugins/algorithms/SaveGEMMAUDParamFile.py
josephframsay: Good point - I've updated it to use the first one for which the template file actually exists

samueljackson92 (approved):

PR #22606, opened 2018-06-18
LoadEventNexus duplicate code

peterfpeterson:
While helping out people loading data from LENS, I discovered that `LoadEventNexus` had an independent implementation of loading event monitors which gave different results from `LoadNexusMonitors`. Talking with developers at SNS, we decided to change the parameter names to clarify whether it was event or histogram monitors that were being loaded. The result is having `LoadNexusMonitors` always used from `LoadEventNexus` and documentation cleanup. The tests that were removed were verifying that broken behavior continued. There is much better coverage in `LoadNexusMonitors` already so the likelihood of new bugs being introduced should actually be lower.

**To test:**

Try loading your favorite nexus files with monitors and see if things still work. There were some CORELLI files that had to be tried with a particular (and strange) set of parameters in `LoadEventNexus`. Ticking the `LoadMonitors` option should be enough now.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####


martyngigg:
@peterfpeterson This has picked up a conflict.

Framework/DataHandling/src/LoadEventNexus.cpp
```diff
@@ -210,27 +185,11 @@ void LoadEventNexus::init() {
...
-                  "will be loaded.");
-
-  declareProperty("LoadHistoMonitors", true,
-                  "Load histogram monitor in NeXus file both event monitor and "
-                  "histogram monitor found in NeXus file."
-                  "If both of LoadEventMonitor and LoadHistoMonitor are true, "
-                  "or both of them are false,"
-                  "then it is in the auto mode such that any existing monitor "
-                  "will be loaded.");
+  std::vector<std::string> options{"", "Events", "Histogram"};
```
martyngigg: So this change will break user scripts - should we provide a deprecation warning in the Python layer?

peterfpeterson: In common usage, people use the default which is "figure it out." It is only in the case of both histogram *and* events being in the file that the user has to intervene.

Framework/PythonInterface/mantid/simpleapi.py
```diff
@@ -1075,6 +1075,17 @@ def algorithm_wrapper(*args, **kwargs):
...
             del kwargs["CoordinatesToUse"]
 
+        # a change in parameters should get a better error message
+        if algm.name() in ['LoadEventNexus', 'LoadNexusMonitors']:
+            for propname in ['MonitorsAsEvents', 'LoadEventMonitors', 'LoadHistoMonitors']:
+                if propname in kwargs:
+                    suggest = 'LoadOnly'
+                    if algm.name() == 'LoadEventNexus':
+                        suggest = 'MonitorsLoadOnly'
+                    msg = 'Deprecated property "{}" in {}. Use "{}" instead'.format(propname,
```
martyngigg: :+1:

martyngigg (approved):
Refactoring improves the situation regarding loading the monitors code.

The clean up of the `LoadEventNexus` properties provides much better clarity over what happens when you load monitors with an event file. :+1:

PR #22615, opened 2018-06-19
Sans renames workspace groups correctly

### LINKED ISSUE #18933: Make name of WorkspaceGroup and child workspaces consistent for sliced reductions.
Sarah wrote:
> As discussed before lunch…could you look please into the naming scheme used for saved files when we use the batch script with time slicing?
 
>As you can see in the attached the current scheme uses the output name for the grouped workspace but not the individual workspaces making up that group (which retains the original SANS Sample number).  This therefore means that the saved files also have the original SANS Sample number as their filename as it is the individual workspaces that are written out.
 
>For me the preferred naming scheme (if you look at the attached) would be:
 
> SS0P4LkinSemiShort_rear_1D_1.75_16.5_t0.00_T60.00

image

The naming scheme that Sarah suggests replaces the run number with the user-provided file/workspace name and leaves the suffix list as it is. This makes sense, since in the case of nested scans of time, wavelength and/or angle-masking,  we won't create duplicate names (which would be overwritten by each other otherwise). 
The change has to happen [here](https://github.com/mantidproject/mantid/blob/master/scripts/SANS/ISISCommandInterface.py#L669:L681)

Matthew-Andrew:
**Description of work.**

This changes how the workspaces within workspace groups are renamed in the sans reduction. The individual members of the group are now renamed in line with the group itself.

**Report to:** sarah.rogers@stfc.ac.uk 

**To test:**

The test files for this PR are available from //olympic/babylon5/Scratch/Matthew Andrew/SANS2DSAMPLE

* Open the old Isis Sans GUI,
* Select SANS2D as the instrument
* Load the maskfile from the folder (this is the *.txt file)
* Select Batch mode and load a batch file (this is the *.csv file)
* Type 1:1:3 into the event slices box. This is specifying the event slices you want to reduce and in this case between 1 and 3 in steps of 1
*  Click 1D Reduce
* At the end of the reduction confirm that there is a workspace group called line_1_reduction_rear which contains line_1_reduction_t1_T2_rear and line_1_reduction_t2_T3_rear.

Fixes #18933. 

Release notes are here #22628
- [x] Yes
- [ ] No

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Testing/SystemTests/tests/analysis/LOQReductionGUI.py
+++ b/Testing/SystemTests/tests/analysis/LOQReductionGUI.py
@@ -26,4 +26,4 @@ class LOQMinimalBatchReduction(stresstesting.MantidStressTest):
         # when overlaying the two options they overlap very well
         self.tolerance = 1.0e+1
         self.disableChecking.append('Instrument')
-        return 'first_time_merged', 'LOQReductionMergedData.nxs'
+        return 'first_time_merged_1D_2.2_10.0', 'LOQReductionMergedData.nxs'
--- a/Testing/SystemTests/tests/analysis/SANS2DReductionGUI.py
+++ b/Testing/SystemTests/tests/analysis/SANS2DReductionGUI.py
@@ -48,7 +48,7 @@ class SANS2DMinimalBatchReduction(stresstesting.MantidStressTest):
 
     def validate(self):
         self.disableChecking.append('Instrument')
-        return "trans_test_rear","SANSReductionGUI.nxs"
+        return "trans_test_rear_1D_1.5_12.5","SANSReductionGUI.nxs"
 
 
 class SANS2DMinimalSingleReduction(SANS2DMinimalBatchReduction):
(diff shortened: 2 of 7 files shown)


martyngigg (commented):
Looking at the original issue the suggested naming scheme suggested as an example:
```
SS0P4LkinSemiShort_rear_1D_1.75_16.5_t0.00_T60.00
```

With the changes I get a group called `line_1_reduction_rear` and the sub members:
* `line_1_reduction_t1.00_T2.00_rear`
* `line_1_reduction_t2.00_T3.00_rear`

Should the `_rear` be after the `line_1_reduction` followed by the `t1....` or have I misunderstood the intention?

keeeto:
The OSX issue is also present on master branch and older release versions of Mantid. Therefore it is not related to this PR. I have created a separate issue #22702.

martyngigg:
@Mantid-Matthew The GUI output looks good but it seems that some system tests are now failing

martyngigg:
@Mantid-Matthew There still seems to be some system tests failing I'm afraid.

martyngigg (approved):
This looks good now.

PR #22617, opened 2018-06-19
(Topic Branches) Recovery background thread

DavidFair:
**Note this is a merge between topic branches, not into master**

**Description of work.**
Adds a saving loop into a background thread for project recovery. The details of this are below.

Still pending:
- Error handling (stopping the mechanism if something goes wrong...etc.)
- Configuration Options
- Filename generation
- Loading

*Details:* 
- We use a `std::thread` and a  `std::conditional_variable` with a timeout to implement the background thread without polling. If the mutex protected changes (when the flag to indicate thread exit) the thread will be woken. Similarly if the timeout occurs the exit flag is checked and if false, we trigger saving.

To save the Qt side of things we must emit a signal to the event queue in the main thread, this is later processed by ApplicationWindow to trigger project saving. One caveat is if multiple actions are queued (for example 100 window loads) we must wait for them to parse first. In other ways there is no to pre-empt a large number of tasks for the GUI side of things

**To test:**
- Open Mantid
- Set the logger to information, so you can see each time it triggers 
- Debug logging will show details from the Python script too if required
- Go to %appdata%/recovery for saved files

- [ ] Yes
- [x] No

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

### DIFF
--- a/Framework/API/inc/MantidAPI/FileProperty.h
+++ b/Framework/API/inc/MantidAPI/FileProperty.h
@@ -86,6 +86,9 @@ public:
   /// Returns the main file extension that's used
   std::string getDefaultExt() const { return m_defaultExt; }
 
+  /// Returns the path to the users home directory
+  static std::string FileProperty::getHomePath();
+
   // Unhide the PropertyWithValue assignment operator
   using Kernel::PropertyWithValue<std::string>::operator=;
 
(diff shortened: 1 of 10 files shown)


keeeto:
@DavidFair One issue I notice with this is that if I delete a workspace after it's history has been written, then the workspace history is still in the recovery folder.

I guess that this will be taken care of with your idea for time-stamped folders.

